PR #598, opened 2026-04-26
feat: add federation domain rules

llun:
## Summary
- Add Mastodon-like domain block and allow rules with storage, admin APIs, and a federation admin UI.
- Enforce federation policy for inbound ActivityPub, remote fetches, follows, actor recording, and outbound delivery fanout.
- Add import support for the Oliphant unified tier 0 known blocklist source and expose public instance domain blocks.

## Verification
- yarn prettier
- yarn lint
- yarn build
- yarn test
- Browser check: verified https://activities.local/admin/federation with block/allow create flows and accessible delete actions.

### DIFF
--- a/app/(timeline)/admin/federation/actions.ts
+++ b/app/(timeline)/admin/federation/actions.ts
@@ -15,6 +15,7 @@ import {
 } from '@/lib/services/federation/domainRules'
 import { DomainBlockSeverity } from '@/lib/types/database/operations'
 import { getAdminFromSession } from '@/lib/utils/getAdminFromSession'
+import { logger } from '@/lib/utils/logger'
 
 const ADMIN_FEDERATION_PATH = '/admin/federation'
 
@@ -108,7 +109,12 @@ export async function importKnownDomainBlocklistAction(formData: FormData) {
     created = result.created
     updated = result.updated
     skipped = result.skipped
-  } catch {
+  } catch (error) {
+    logger.error({
+      message: 'Failed to import known domain blocklist',
+      sourceId,
+      error
+    })
     redirectWithStatus('import-failed')
   }
 
--- a/app/(timeline)/admin/federation/actions.ts
+++ b/app/(timeline)/admin/federation/actions.ts
@@ -1,0 +1,108 @@
+'use server'
+
+import { revalidatePath } from 'next/cache'
+import { redirect } from 'next/navigation'
+
+import { getDatabase } from '@/lib/database'
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import {
+  KnownDomainBlocklistSourceId,
+  fetchKnownDomainBlocklist
+} from '@/lib/services/federation/blocklistSources'
+import {
+  DEFAULT_DOMAIN_BLOCK_SEVERITY,
+  normalizeDomain
+} from '@/lib/services/federation/domainRules'
+import { DomainBlockSeverity } from '@/lib/types/database/operations'
+import { getAdminFromSession } from '@/lib/utils/getAdminFromSession'
+
+const ADMIN_FEDERATION_PATH = '/admin/federation'
+
+const getAdminDatabase = async () => {
+  const database = getDatabase()
+  if (!database) throw new Error('Failed to load database')
+
+  const session = await getServerAuthSession()
+  const admin = await getAdminFromSession(database, session)
+  if (!admin) redirect('/')
+
+  return database
+}
+
+const redirectWithStatus = (status: string): never => {
+  revalidatePath(ADMIN_FEDERATION_PATH)
+  redirect(`${ADMIN_FEDERATION_PATH}?status=${encodeURIComponent(status)}`)
+}
+
+const getText = (formData: FormData, key: string): string =>
+  String(formData.get(key) ?? '').trim()
+
+const getCheckbox = (formData: FormData, key: string): boolean =>
+  formData.get(key) === 'on'
+
+const getSeverity = (value: string): DomainBlockSeverity =>
+  value === 'noop' || value === 'silence' || value === 'suspend'
+    ? value
+    : DEFAULT_DOMAIN_BLOCK_SEVERITY
+
+export async function createDomainBlockAction(formData: FormData) {
+  const database = await getAdminDatabase()
+  const domain = normalizeDomain(getText(formData, 'domain'))
+  const normalizedDomain = domain ?? redirectWithStatus('invalid-block-domain')
+
+  await database.createDomainBlock({
+    domain: normalizedDomain,
+    severity: getSeverity(getText(formData, 'severity')),
+    rejectMedia: getCheckbox(formData, 'rejectMedia'),
+    rejectReports: getCheckbox(formData, 'rejectReports'),
+    privateComment: getText(formData, 'privateComment') || null,
+    publicComment: getText(formData, 'publicComment') || null,
+    obfuscate: getCheckbox(formData, 'obfuscate'),
+    source: null
+  })
+
+  redirectWithStatus('block-saved')
+}
+
+export async function deleteDomainBlockAction(formData: FormData) {
+  const database = await getAdminDatabase()
+  const id = getText(formData, 'id')
+  if (id) await database.deleteDomainBlock(id)
+
+  redirectWithStatus('block-deleted')
+}
+
+export async function createDomainAllowAction(formData: FormData) {
+  const database = await getAdminDatabase()
+  const domain = normalizeDomain(getText(formData, 'domain'))
+  const normalizedDomain = domain ?? redirectWithStatus('invalid-allow-domain')
+
+  await database.createDomainAllow({ domain: normalizedDomain })
+
+  redirectWithStatus('allow-saved')
+}
+
+export async function deleteDomainAllowAction(formData: FormData) {
+  const database = await getAdminDatabase()
+  const id = getText(formData, 'id')
+  if (id) await database.deleteDomainAllow(id)
+
+  redirectWithStatus('allow-deleted')
+}
+
+export async function importKnownDomainBlocklistAction(formData: FormData) {
+  const database = await getAdminDatabase()
+  const source = KnownDomainBlocklistSourceId.safeParse(
+    getText(formData, 'source')
+  )
+  const sourceId = source.success
+    ? source.data
+    : redirectWithStatus('invalid-source')
+
+  const blocks = await fetchKnownDomainBlocklist(sourceId)
+  const result = await database.importDomainBlocks({ blocks })
+
+  redirectWithStatus(
+    `imported-${result.created}-${result.updated}-${result.skipped}`
+  )
+}
--- a/app/(timeline)/admin/federation/actions.ts
+++ b/app/(timeline)/admin/federation/actions.ts
@@ -7,7 +7,7 @@ import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import {
   KnownDomainBlocklistSourceId,
-  fetchKnownDomainBlocklist
+  downloadKnownDomainBlocklist
 } from '@/lib/services/federation/blocklistSources'
 import {
   DEFAULT_DOMAIN_BLOCK_SEVERITY,
@@ -103,7 +103,7 @@ export async function importKnownDomainBlocklistAction(formData: FormData) {
   let updated = 0
   let skipped = 0
   try {
-    const blocks = await fetchKnownDomainBlocklist(sourceId)
+    const blocks = await downloadKnownDomainBlocklist(sourceId)
     const result = await database.importDomainBlocks({ blocks })
     created = result.created
     updated = result.updated
--- a/app/(timeline)/admin/federation/actions.ts
+++ b/app/(timeline)/admin/federation/actions.ts
@@ -99,10 +99,18 @@ export async function importKnownDomainBlocklistAction(formData: FormData) {
     ? source.data
     : redirectWithStatus('invalid-source')
 
-  const blocks = await fetchKnownDomainBlocklist(sourceId)
-  const result = await database.importDomainBlocks({ blocks })
+  let created = 0
+  let updated = 0
+  let skipped = 0
+  try {
+    const blocks = await fetchKnownDomainBlocklist(sourceId)
+    const result = await database.importDomainBlocks({ blocks })
+    created = result.created
+    updated = result.updated
+    skipped = result.skipped
+  } catch {
+    redirectWithStatus('import-failed')
+  }
 
-  redirectWithStatus(
-    `imported-${result.created}-${result.updated}-${result.skipped}`
-  )
+  redirectWithStatus(`imported-${created}-${updated}-${skipped}`)
 }
--- a/app/(timeline)/admin/federation/page.tsx
+++ b/app/(timeline)/admin/federation/page.tsx
@@ -1,4 +1,12 @@
-import { Download, ShieldBan, ShieldCheck, Trash2 } from 'lucide-react'
+import {
+  ChevronLeft,
+  ChevronRight,
+  Download,
+  ShieldBan,
+  ShieldCheck,
+  Trash2
+} from 'lucide-react'
+import Link from 'next/link'
 import { redirect } from 'next/navigation'
 
 import {
@@ -18,7 +26,7 @@ import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import { KNOWN_DOMAIN_BLOCKLIST_SOURCES } from '@/lib/services/federation/blocklistSources'
-import { toPublicDomainBlock } from '@/lib/services/federation/domainRules'
+import { cn } from '@/lib/utils'
 import { getAdminFromSession } from '@/lib/utils/getAdminFromSession'
 
 export const dynamic = 'force-dynamic'
@@ -38,6 +46,13 @@ const STATUS_MESSAGES: Record<string, string> = {
   'import-failed': 'Unable to import the blocklist'
 }
 
+const ERROR_STATUSES = new Set([
+  'invalid-block-domain',
+  'invalid-allow-domain',
+  'invalid-source',
+  'import-failed'
+])
+
 const ADMIN_FEDERATION_PAGE_SIZE = 100
 
 const getStatusMessage = (status?: string): string | null => {
@@ -50,6 +65,87 @@ const getStatusMessage = (status?: string): string | null => {
   return `Imported ${match[1]} new block${match[1] === '1' ? '' : 's'}, updated ${match[2]}, skipped ${match[3]}`
 }
 
+const getOffset = (value?: string): number => {
+  const offset = Number(value ?? 0)
+  return Number.isInteger(offset) && offset > 0 ? offset : 0
+}
+
+const getPaginationHref = ({
+  blockOffset,
+  allowOffset
+}: {
+  blockOffset: number
+  allowOffset: number
+}): string => {
+  const params = new URLSearchParams()
+  if (blockOffset > 0) params.set('blockOffset', String(blockOffset))
+  if (allowOffset > 0) params.set('allowOffset', String(allowOffset))
+
+  const query = params.toString()
+  return query ? `/admin/federation?${query}` : '/admin/federation'
+}
+
+const getNextOffset = (offset: number, count: number): number => offset + count
+
+const getPreviousOffset = (offset: number): number =>
+  Math.max(0, offset - ADMIN_FEDERATION_PAGE_SIZE)
+
+const getPaginationLabel = (
+  noun: string,
+  offset: number,
+  count: number,
+  total: number
+): string => {
+  if (count === 0) return `Showing 0 of ${total} ${noun}`
+  return `Showing ${offset + 1}-${offset + count} of ${total} ${noun}`
+}
+
+const PaginationControls = ({
+  label,
+  previousHref,
+  nextHref,
+  hasPrevious,
+  hasNext
+}: {
+  label: string
+  previousHref: string
+  nextHref: string
+  hasPrevious: boolean
+  hasNext: boolean
+}) => (
+  <div className="mt-4 flex flex-col gap-3 sm:flex-row sm:items-center sm:justify-between">
+    <p className="text-sm text-muted-foreground">{label}</p>
+    <div className="flex gap-2">
+      {hasPrevious ? (
+        <Button asChild variant="outline" size="sm">
+          <Link href={previousHref}>
+            <ChevronLeft className="h-4 w-4" />
+            Previous
+          </Link>
+        </Button>
+      ) : (
+        <Button disabled variant="outline" size="sm">
+          <ChevronLeft className="h-4 w-4" />
+          Previous
+        </Button>
+      )}
+      {hasNext ? (
+        <Button asChild variant="outline" size="sm">
+          <Link href={nextHref}>
+            Next
+            <ChevronRight className="h-4 w-4" />
+          </Link>
+        </Button>
+      ) : (
+        <Button disabled variant="outline" size="sm">
+          Next
+          <ChevronRight className="h-4 w-4" />
+        </Button>
+      )}
+    </div>
+  </div>
+)
+
 const Page = async ({ searchParams }: Props) => {
   const database = getDatabase()
   if (!database) throw new Error('Failed to load database')
@@ -58,16 +154,35 @@ const Page = async ({ searchParams }: Props) => {
   const admin = await getAdminFromSession(database, session)
   if (!admin) return redirect('/')
 
-  const [{ status }, blocks, allows, stats] = await Promise.all([
-    searchParams,
-    database.getDomainBlocks({ limit: ADMIN_FEDERATION_PAGE_SIZE }),
-    database.getDomainAllows({ limit: ADMIN_FEDERATION_PAGE_SIZE }),
+  const params = await searchParams
+  const { status } = params
+  const blockOffset = getOffset(params.blockOffset)
+  const allowOffset = getOffset(params.allowOffset)
+
+  const [blocks, allows, stats] = await Promise.all([
+    database.getDomainBlocks({
+      limit: ADMIN_FEDERATION_PAGE_SIZE,
+      offset: blockOffset
+    }),
+    database.getDomainAllows({
+      limit: ADMIN_FEDERATION_PAGE_SIZE,
+      offset: allowOffset
+    }),
     database.getDomainFederationRuleStats()
   ])
   const statusMessage = getStatusMessage(status)
+  const isErrorStatus = status ? ERROR_STATUSES.has(status) : false
   const sourceCounts = new Map(Object.entries(stats.sourceCounts))
   const config = getConfig()
   const federationMode = config.federationMode ?? 'open'
+  const hasPreviousBlocks = blockOffset > 0
+  const hasNextBlocks =
+    blocks.length > 0 &&
+    getNextOffset(blockOffset, blocks.length) < stats.blocks
+  const hasPreviousAllows = allowOffset > 0
+  const hasNextAllows =
+    allows.length > 0 &&
+    getNextOffset(allowOffset, allows.length) < stats.allows
 
   return (
     <div className="space-y-6">
@@ -81,7 +196,14 @@ const Page = async ({ searchParams }: Props) => {
       </div>
 
       {statusMessage && (
-        <div className="rounded-lg border border-green-200 bg-green-50 px-4 py-3 text-sm text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100">
+        <div
+          className={cn(
+            'rounded-lg border px-4 py-3 text-sm',
+            isErrorStatus
+              ? 'border-red-200 bg-red-50 text-red-900 dark:border-red-900 dark:bg-red-950 dark:text-red-100'
+              : 'border-green-200 bg-green-50 text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100'
+          )}
+        >
           {statusMessage}
         </div>
       )}
@@ -176,6 +298,14 @@ const Page = async ({ searchParams }: Props) => {
                 rows={2}
               />
             </div>
+            <div className="space-y-1">
+              <Label htmlFor="block-private-comment">Private comment</Label>
+              <Textarea
+                id="block-private-comment"
+                name="privateComment"
+                rows={2}
+              />
+            </div>
             <div className="grid gap-3 sm:grid-cols-3">
               <Label>
                 <Checkbox name="rejectMedia" />
@@ -213,54 +343,59 @@ const Page = async ({ searchParams }: Props) => {
 
       <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
         <h2 className="mb-4 text-lg font-semibold">Domain Blocks</h2>
-        {stats.blocks > blocks.length && (
-          <p className="mb-3 text-sm text-muted-foreground">
-            Showing first {blocks.length} of {stats.blocks} blocked domains.
-          </p>
-        )}
         <div className="space-y-2">
           {blocks.length === 0 ? (
             <p className="text-sm text-muted-foreground">No domains blocked</p>
           ) : (
-            blocks.map((block) => {
-              const publicBlock = toPublicDomainBlock(block)
-              return (
-                <div
-                  key={block.id}
-                  className="flex flex-col gap-3 rounded-lg border p-4 sm:flex-row sm:items-center sm:justify-between"
-                >
-                  <div className="min-w-0">
-                    <p className="truncate font-medium">{block.domain}</p>
-                    <p className="text-sm text-muted-foreground">
-                      {block.severity}
-                      {publicBlock.comment ? ` - ${publicBlock.comment}` : ''}
-                    </p>
-                  </div>
-                  <form action={deleteDomainBlockAction}>
-                    <input type="hidden" name="id" value={block.id} />
-                    <Button
-                      type="submit"
-                      variant="ghost"
-                      size="icon"
-                      aria-label={`Delete block for ${block.domain}`}
-                    >
-                      <Trash2 className="h-4 w-4" />
-                    </Button>
-                  </form>
+            blocks.map((block) => (
+              <div
+                key={block.id}
+                className="flex flex-col gap-3 rounded-lg border p-4 sm:flex-row sm:items-center sm:justify-between"
+              >
+                <div className="min-w-0">
+                  <p className="truncate font-medium">{block.domain}</p>
+                  <p className="text-sm text-muted-foreground">
+                    {block.severity}
+                    {block.publicComment ? ` - ${block.publicComment}` : ''}
+                  </p>
                 </div>
-              )
-            })
+                <form action={deleteDomainBlockAction}>
+                  <input type="hidden" name="id" value={block.id} />
+                  <Button
+                    type="submit"
+                    variant="ghost"
+                    size="icon"
+                    aria-label={`Delete block for ${block.domain}`}
+                  >
+                    <Trash2 className="h-4 w-4" />
+                  </Button>
+                </form>
+              </div>
+            ))
           )}
         </div>
+        <PaginationControls
+          label={getPaginationLabel(
+            'blocked domains',
+            blockOffset,
+            blocks.length,
+            stats.blocks
+          )}
+          previousHref={getPaginationHref({
+            blockOffset: getPreviousOffset(blockOffset),
+            allowOffset
+          })}
+          nextHref={getPaginationHref({
+            blockOffset: getNextOffset(blockOffset, blocks.length),
+            allowOffset
+          })}
+          hasPrevious={hasPreviousBlocks}
+          hasNext={hasNextBlocks}
+        />
       </section>
 
       <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
         <h2 className="mb-4 text-lg font-semibold">Domain Allows</h2>
-        {stats.allows > allows.length && (
-          <p className="mb-3 text-sm text-muted-foreground">
-            Showing first {allows.length} of {stats.allows} allowed domains.
-          </p>
-        )}
         <div className="space-y-2">
           {allows.length === 0 ? (
             <p className="text-sm text-muted-foreground">No domains allowed</p>
@@ -286,6 +421,24 @@ const Page = async ({ searchParams }: Props) => {
             ))
           )}
         </div>
+        <PaginationControls
+          label={getPaginationLabel(
+            'allowed domains',
+            allowOffset,
+            allows.length,
+            stats.allows
+          )}
+          previousHref={getPaginationHref({
+            blockOffset,
+            allowOffset: getPreviousOffset(allowOffset)
+          })}
+          nextHref={getPaginationHref({
+            blockOffset,
+            allowOffset: getNextOffset(allowOffset, allows.length)
+          })}
+          hasPrevious={hasPreviousAllows}
+          hasNext={hasNextAllows}
+        />
       </section>
     </div>
   )
--- a/app/(timeline)/admin/federation/page.tsx
+++ b/app/(timeline)/admin/federation/page.tsx
@@ -34,7 +34,8 @@ const STATUS_MESSAGES: Record<string, string> = {
   'allow-deleted': 'Domain allow deleted',
   'invalid-block-domain': 'Enter a valid domain to block',
   'invalid-allow-domain': 'Enter a valid domain to allow',
-  'invalid-source': 'Choose a known blocklist source'
+  'invalid-source': 'Choose a known blocklist source',
+  'import-failed': 'Unable to import the blocklist'
 }
 
 const ADMIN_FEDERATION_PAGE_SIZE = 100
--- a/app/(timeline)/admin/federation/page.tsx
+++ b/app/(timeline)/admin/federation/page.tsx
@@ -1,0 +1,283 @@
+import { Download, ShieldBan, ShieldCheck, Trash2 } from 'lucide-react'
+import { redirect } from 'next/navigation'
+
+import {
+  createDomainAllowAction,
+  createDomainBlockAction,
+  deleteDomainAllowAction,
+  deleteDomainBlockAction,
+  importKnownDomainBlocklistAction
+} from '@/app/(timeline)/admin/federation/actions'
+import { Button } from '@/lib/components/ui/button'
+import { getConfig } from '@/lib/config'
+import { getDatabase } from '@/lib/database'
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { KNOWN_DOMAIN_BLOCKLIST_SOURCES } from '@/lib/services/federation/blocklistSources'
+import { toPublicDomainBlock } from '@/lib/services/federation/domainRules'
+import { getAdminFromSession } from '@/lib/utils/getAdminFromSession'
+
+export const dynamic = 'force-dynamic'
+
+interface Props {
+  searchParams: Promise<Record<string, string | undefined>>
+}
+
+const STATUS_MESSAGES: Record<string, string> = {
+  'block-saved': 'Domain block saved',
+  'block-deleted': 'Domain block deleted',
+  'allow-saved': 'Domain allow saved',
+  'allow-deleted': 'Domain allow deleted',
+  'invalid-block-domain': 'Enter a valid domain to block',
+  'invalid-allow-domain': 'Enter a valid domain to allow',
+  'invalid-source': 'Choose a known blocklist source'
+}
+
+const getStatusMessage = (status?: string): string | null => {
+  if (!status) return null
+  if (STATUS_MESSAGES[status]) return STATUS_MESSAGES[status]
+
+  const match = /^imported-(\d+)-(\d+)-(\d+)$/.exec(status)
+  if (!match) return null
+
+  return `Imported ${match[1]} new block${match[1] === '1' ? '' : 's'}, updated ${match[2]}, skipped ${match[3]}`
+}
+
+const Page = async ({ searchParams }: Props) => {
+  const database = getDatabase()
+  if (!database) throw new Error('Failed to load database')
+
+  const session = await getServerAuthSession()
+  const admin = await getAdminFromSession(database, session)
+  if (!admin) return redirect('/')
+
+  const [{ status }, blocks, allows] = await Promise.all([
+    searchParams,
+    database.getDomainBlocks({ limit: 10_000 }),
+    database.getDomainAllows({ limit: 10_000 })
+  ])
+  const statusMessage = getStatusMessage(status)
+  const sourceCounts = new Map<string, number>()
+  blocks.forEach((block) => {
+    if (!block.source) return
+    sourceCounts.set(block.source, (sourceCounts.get(block.source) ?? 0) + 1)
+  })
+  const config = getConfig()
+  const federationMode = config.federationMode ?? 'open'
+
+  return (
+    <div className="space-y-6">
+      <div>
+        <h1 className="text-2xl font-bold">Federation</h1>
+        <p className="text-sm text-muted-foreground">
+          {federationMode === 'allowlist'
+            ? 'Limited federation mode'
+            : 'Open federation mode'}
+        </p>
+      </div>
+
+      {statusMessage && (
+        <div className="rounded-lg border border-green-200 bg-green-50 px-4 py-3 text-sm text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100">
+          {statusMessage}
+        </div>
+      )}
+
+      <div className="grid gap-4 md:grid-cols-3">
+        <div className="rounded-xl border bg-background/80 p-5 shadow-sm">
+          <div className="flex items-center gap-3">
+            <ShieldBan className="h-5 w-5 text-destructive" />
+            <div>
+              <p className="text-sm text-muted-foreground">Blocked domains</p>
+              <p className="text-2xl font-bold">{blocks.length}</p>
+            </div>
+          </div>
+        </div>
+        <div className="rounded-xl border bg-background/80 p-5 shadow-sm">
+          <div className="flex items-center gap-3">
+            <ShieldCheck className="h-5 w-5 text-green-600" />
+            <div>
+              <p className="text-sm text-muted-foreground">Allowed domains</p>
+              <p className="text-2xl font-bold">{allows.length}</p>
+            </div>
+          </div>
+        </div>
+        <div className="rounded-xl border bg-background/80 p-5 shadow-sm">
+          <p className="text-sm text-muted-foreground">Shared-list entries</p>
+          <p className="text-2xl font-bold">
+            {blocks.filter((block) => block.source).length}
+          </p>
+        </div>
+      </div>
+
+      <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
+        <div className="mb-4 flex items-center justify-between gap-4">
+          <div>
+            <h2 className="text-lg font-semibold">Known Blocklist</h2>
+            <p className="text-sm text-muted-foreground">
+              Import a Mastodon-compatible CSV source.
+            </p>
+          </div>
+        </div>
+        <div className="space-y-3">
+          {KNOWN_DOMAIN_BLOCKLIST_SOURCES.map((source) => (
+            <form
+              key={source.id}
+              action={importKnownDomainBlocklistAction}
+              className="flex flex-col gap-3 rounded-lg border p-4 sm:flex-row sm:items-center sm:justify-between"
+            >
+              <input type="hidden" name="source" value={source.id} />
+              <div>
+                <p className="font-medium">{source.name}</p>
+                <p className="text-sm text-muted-foreground">
+                  {sourceCounts.get(source.id) ?? 0} imported entries
+                </p>
+              </div>
+              <Button type="submit" className="sm:w-auto">
+                <Download className="h-4 w-4" />
+                Import
+              </Button>
+            </form>
+          ))}
+        </div>
+      </section>
+
+      <section className="grid gap-6 lg:grid-cols-2">
+        <div className="rounded-xl border bg-background/80 p-5 shadow-sm">
+          <h2 className="mb-4 text-lg font-semibold">Add Domain Block</h2>
+          <form action={createDomainBlockAction} className="space-y-4">
+            <label className="block">
+              <span className="mb-1 block text-sm font-medium">Domain</span>
+              <input
+                required
+                name="domain"
+                placeholder="example.social"
+                className="w-full rounded-lg border bg-background px-3 py-2 text-sm"
+              />
+            </label>
+            <label className="block">
+              <span className="mb-1 block text-sm font-medium">Severity</span>
+              <select
+                name="severity"
+                defaultValue="suspend"
+                className="w-full rounded-lg border bg-background px-3 py-2 text-sm"
+              >
+                <option value="suspend">Suspend</option>
+                <option value="silence">Silence</option>
+                <option value="noop">Noop</option>
+              </select>
+            </label>
+            <label className="block">
+              <span className="mb-1 block text-sm font-medium">
+                Public comment
+              </span>
+              <textarea
+                name="publicComment"
+                rows={2}
+                className="w-full rounded-lg border bg-background px-3 py-2 text-sm"
+              />
+            </label>
+            <div className="grid gap-3 sm:grid-cols-3">
+              <label className="flex items-center gap-2 text-sm">
+                <input name="rejectMedia" type="checkbox" />
+                Reject media
+              </label>
+              <label className="flex items-center gap-2 text-sm">
+                <input name="rejectReports" type="checkbox" />
+                Reject reports
+              </label>
+              <label className="flex items-center gap-2 text-sm">
+                <input name="obfuscate" type="checkbox" />
+                Obfuscate
+              </label>
+            </div>
+            <Button type="submit">Save block</Button>
+          </form>
+        </div>
+
+        <div className="rounded-xl border bg-background/80 p-5 shadow-sm">
+          <h2 className="mb-4 text-lg font-semibold">Add Domain Allow</h2>
+          <form action={createDomainAllowAction} className="space-y-4">
+            <label className="block">
+              <span className="mb-1 block text-sm font-medium">Domain</span>
+              <input
+                required
+                name="domain"
+                placeholder="trusted.social"
+                className="w-full rounded-lg border bg-background px-3 py-2 text-sm"
+              />
+            </label>
+            <Button type="submit">Save allow</Button>
+          </form>
+        </div>
+      </section>
+
+      <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
+        <h2 className="mb-4 text-lg font-semibold">Domain Blocks</h2>
+        <div className="space-y-2">
+          {blocks.length === 0 ? (
+            <p className="text-sm text-muted-foreground">No domains blocked</p>
+          ) : (
+            blocks.map((block) => {
+              const publicBlock = toPublicDomainBlock(block)
+              return (
+                <div
+                  key={block.id}
+                  className="flex flex-col gap-3 rounded-lg border p-4 sm:flex-row sm:items-center sm:justify-between"
+                >
+                  <div className="min-w-0">
+                    <p className="truncate font-medium">{block.domain}</p>
+                    <p className="text-sm text-muted-foreground">
+                      {block.severity}
+                      {publicBlock.comment ? ` - ${publicBlock.comment}` : ''}
+                    </p>
+                  </div>
+                  <form action={deleteDomainBlockAction}>
+                    <input type="hidden" name="id" value={block.id} />
+                    <Button
+                      type="submit"
+                      variant="ghost"
+                      size="icon"
+                      aria-label={`Delete block for ${block.domain}`}
+                    >
+                      <Trash2 className="h-4 w-4" />
+                    </Button>
+                  </form>
+                </div>
+              )
+            })
+          )}
+        </div>
+      </section>
+
+      <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
+        <h2 className="mb-4 text-lg font-semibold">Domain Allows</h2>
+        <div className="space-y-2">
+          {allows.length === 0 ? (
+            <p className="text-sm text-muted-foreground">No domains allowed</p>
+          ) : (
+            allows.map((allow) => (
+              <div
+                key={allow.id}
+                className="flex items-center justify-between gap-3 rounded-lg border p-4"
+              >
+                <p className="min-w-0 truncate font-medium">{allow.domain}</p>
+                <form action={deleteDomainAllowAction}>
+                  <input type="hidden" name="id" value={allow.id} />
+                  <Button
+                    type="submit"
+                    variant="ghost"
+                    size="icon"
+                    aria-label={`Delete allow for ${allow.domain}`}
+                  >
+                    <Trash2 className="h-4 w-4" />
+                  </Button>
+                </form>
+              </div>
+            ))
+          )}
+        </div>
+      </section>
+    </div>
+  )
+}
+
+export default Page
--- a/app/(timeline)/admin/federation/page.tsx
+++ b/app/(timeline)/admin/federation/page.tsx
@@ -289,6 +289,10 @@ const Page = async ({ searchParams }: Props) => {
                 <option value="silence">Silence</option>
                 <option value="noop">Noop</option>
               </Select>
+              <p className="text-xs text-muted-foreground">
+                Only Suspend rejects federation. Silence and Noop are stored for
+                Mastodon-compatible metadata.
+              </p>
             </div>
             <div className="space-y-1">
               <Label htmlFor="block-public-comment">Public comment</Label>
--- a/app/(timeline)/admin/federation/page.tsx
+++ b/app/(timeline)/admin/federation/page.tsx
@@ -9,6 +9,11 @@ import {
   importKnownDomainBlocklistAction
 } from '@/app/(timeline)/admin/federation/actions'
 import { Button } from '@/lib/components/ui/button'
+import { Checkbox } from '@/lib/components/ui/checkbox'
+import { Input } from '@/lib/components/ui/input'
+import { Label } from '@/lib/components/ui/label'
+import { Select } from '@/lib/components/ui/select'
+import { Textarea } from '@/lib/components/ui/textarea'
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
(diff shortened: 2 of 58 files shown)


llun:
@codex revew

lib/database/sql/admin.ts
```diff
@@ -246,5 +327,204 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+    return findMatchingDomainRule(normalized, blocks)
+  },
+
+  async getDomainAllowForDomain(domain: string) {
+    const normalized = normalizeDomain(domain)
+    if (!normalized) return null
+
+    const allows = await this.getDomainAllows({ limit: 10_000 })
+    return findMatchingDomainRule(normalized, allows)
+  },
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: domain block and allow lookups now run targeted SQL candidate matching instead of fetching 10k rows into memory. I kept null returns because the AdminDatabase interface and existing callers/tests use null for optional database results.

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,84 @@
...
+export const isDomainAllowed = async (
+  database: Database,
+  value: string
+): Promise<boolean> => {
+  const domain = getDomainFromUrl(value)
+  if (!domain) return false
+  if (isLocalFederationDomain(domain)) return true
+
+  const allows = await database.getDomainAllows()
+  return Boolean(findMatchingDomainRule(domain, allows))
```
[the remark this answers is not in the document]
llun: Addressed in 67bc20ca: allowlist enforcement now uses getDomainAllowForDomain, so it is not affected by getDomainAllows pagination.

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,84 @@
...
+): Promise<string[]> => {
+  const uniqueUrls = [...new Set(urls)]
+  const allowed = await Promise.all(
+    uniqueUrls.map(async (url) =>
+      (await canFederateWithDomain(database, url)) ? url : null
+    )
+  )
+
+  return allowed.filter((url): url is string => url !== null)
+}
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: filterFederatedUrls now groups by unique domain and performs one policy lookup per domain, with a regression test.

app/api/v1/admin/domain_blocks/schema.ts
```diff
@@ -0,0 +1,48 @@
+import { z } from 'zod'
+
+import { normalizeDomain } from '@/lib/services/federation/domainRules'
+import { DomainBlockSeverity } from '@/lib/types/database/operations'
+
+const Booleanish = z.union([z.boolean(), z.string()]).transform((value) => {
+  if (typeof value === 'boolean') return value
+  return value.toLowerCase() === 'true' || value === '1'
+})
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: Booleanish now accepts checkbox-style "on", with API coverage.

app/api/v1/admin/domain_allows/route.ts
```diff
@@ -0,0 +1,62 @@
...
+      allowedMethods: CORS_HEADERS,
+      data: allows.map(toAdminDomainAllow)
+    })
+  })
+)
+
+export const POST = traceApiRoute(
+  'adminCreateDomainAllow',
+  AdminApiGuard(CORS_HEADERS, async (req: NextRequest, { database }) => {
+    const parsed = DomainAllowRequest.safeParse(await readRequestData(req))
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: admin domain allow/block/create/update/import handlers catch invalid request bodies and return 400 via apiResponse.

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,283 @@
...
+              </label>
+              <label className="flex items-center gap-2 text-sm">
+                <input name="rejectReports" type="checkbox" />
+                Reject reports
+              </label>
+              <label className="flex items-center gap-2 text-sm">
+                <input name="obfuscate" type="checkbox" />
+                Obfuscate
+              </label>
+            </div>
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: the admin federation page uses shared Input, Checkbox, Label, Textarea, and a native Select wrapper.

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,283 @@
...
+  if (!database) throw new Error('Failed to load database')
+
+  const session = await getServerAuthSession()
+  const admin = await getAdminFromSession(database, session)
+  if (!admin) return redirect('/')
+
+  const [{ status }, blocks, allows] = await Promise.all([
+    searchParams,
+    database.getDomainBlocks({ limit: 10_000 }),
+    database.getDomainAllows({ limit: 10_000 })
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: the page now fetches only the first 100 rows and gets totals/source counts from a DB aggregate instead of rendering up to 10k entries.

lib/database/sql/admin.ts
```diff
@@ -246,5 +327,204 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+
+    return row ? toDomainAllow(row) : null
+  },
+
+  async getDomainBlockForDomain(domain: string) {
+    const normalized = normalizeDomain(domain)
+    if (!normalized) return null
+
+    const blocks = await this.getDomainBlocks({ limit: 10_000 })
+    return findMatchingDomainRule(normalized, blocks)
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: block lookup now uses targeted SQL candidate matching for exact, suffix, wildcard, and global rules.

lib/database/sql/admin.ts
```diff
@@ -246,5 +327,204 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+  },
+
+  async getDomainAllowForDomain(domain: string) {
+    const normalized = normalizeDomain(domain)
+    if (!normalized) return null
+
+    const allows = await this.getDomainAllows({ limit: 10_000 })
+    return findMatchingDomainRule(normalized, allows)
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: allow lookup now uses targeted SQL candidate matching for exact, suffix, wildcard, and global rules.

app/api/v1/admin/domain_blocks/schema.ts
```diff
@@ -0,0 +1,48 @@
+import { z } from 'zod'
+
+import { normalizeDomain } from '@/lib/services/federation/domainRules'
+import { DomainBlockSeverity } from '@/lib/types/database/operations'
+
+const Booleanish = z.union([z.boolean(), z.string()]).transform((value) => {
+  if (typeof value === 'boolean') return value
+  return value.toLowerCase() === 'true' || value === '1'
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: Booleanish accepts "on" and the API test covers checkbox-style values.

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,130 @@
...
+  fields.push(current)
+  return fields
+}
+
+export const parseDomainBlockCsv = (
+  csv: string,
+  source: string
+): ImportDomainBlockParams[] => {
+  const lines = csv
+    .split(/\r?\n/)
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: CSV parsing now operates at record level and supports quoted fields containing newlines, with coverage.

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,82 @@
...
+  value: string
+): Promise<boolean> => {
+  const domain = getDomainFromUrl(value)
+  if (!domain) return false
+  if (isLocalFederationDomain(domain)) return true
+
+  return Boolean(await database.getDomainAllowForDomain(domain))
+}
+
+export const canFederateWithDomain = async (
```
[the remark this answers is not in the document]
llun: Addressed in 67bc20ca: isDomainAllowed uses getDomainAllowForDomain instead of paginated getDomainAllows.

lib/database/sql/admin.ts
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: getDomainBlockForDomain now performs SQL candidate matching instead of a full in-memory scan.

lib/database/sql/admin.ts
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91: getDomainAllowForDomain now performs SQL candidate matching instead of a full in-memory scan.

lib/database/sql/admin.ts
```diff
@@ -246,5 +327,204 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+            .update(buildDomainBlockUpdate(block, now))
+          updated++
+          continue
+        }
+
+        await trx('domain_federation_rules').insert(
+          buildDomainBlockInsert(block, now)
+        )
+        created++
+      }
```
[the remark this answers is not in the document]
llun: Addressed in 4d0f47f7: importDomainBlocks now normalizes/dedupes domains, fetches existing rows in chunks, and uses chunked onConflict upserts instead of per-row SELECT/update/insert loops. Added a batch import regression test.

lib/services/federation/domainPolicy.ts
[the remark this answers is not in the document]
llun: Addressed in 67bc20ca: allowlist mode now checks getDomainAllowForDomain directly, avoiding the paginated allow-list bug.

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const filterFederatedUrls = async (
+  database: Database,
+  urls: string[]
+): Promise<string[]> => {
+  const uniqueUrls = [...new Set(urls)]
+  const allowed = await Promise.all(
+    uniqueUrls.map(async (url) =>
+      (await canFederateWithDomain(database, url)) ? url : null
+    )
```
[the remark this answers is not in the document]
llun: Addressed in 4d65ed91 and strengthened by 4d0f47f7: fanout filtering is grouped by unique domain, rule lookups are SQL targeted, and imports use chunked upserts.

llun:
@codex revew

lib/database/sql/admin.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
```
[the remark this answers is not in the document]
llun: Addressed in 9b33643a: admin SQL now imports randomUUID from node:crypto.

lib/database/sql/admin.ts
```diff
@@ -52,6 +64,98 @@ const toDomainActor = (row: SQLActor): Actor => {
...
+    chunks.push(items.slice(index, index + size))
+  }
+
+  return chunks
+}
+
+const buildDomainBlockInsert = (
+  params: CreateDomainBlockParams,
+  now: Date,
+  id = crypto.randomUUID()
```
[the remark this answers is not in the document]
llun: Addressed in 9b33643a: buildDomainBlockInsert now uses the imported randomUUID helper.

lib/database/sql/admin.ts
```diff
@@ -246,5 +350,281 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+    const existing = await database<SQLDomainFederationRule>(
+      'domain_federation_rules'
+    )
+      .where({ type: 'allow', domain: normalized })
+      .first()
+    if (existing) return toDomainAllow(existing)
+
+    const now = new Date()
+    const row = {
+      id: crypto.randomUUID(),
```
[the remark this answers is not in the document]
llun: Addressed in 9b33643a: createDomainAllow now uses the imported randomUUID helper.

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export const parseDomainBlockCsv = (
+  csv: string,
+  source: string
+): ImportDomainBlockParams[] => {
+  const [headerFields, ...rows] = parseCsvRecords(csv)
+  if (!headerFields) return []
+
+  const headers = headerFields.map((field) => field.trim().replace(/^#/, ''))
+  const domainIndex = headers.indexOf('domain')
```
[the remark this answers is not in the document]
llun: Addressed in 9b33643a: CSV headers are normalized case-insensitively before lookup, with the parser test updated to cover capitalized headers.

llun:
@codex revew

lib/services/federation/domainRules.ts
```diff
@@ -0,0 +1,92 @@
...
+    ? trimmed
+    : `https://${trimmed}`
+
+  try {
+    const url = new URL(withScheme)
+    return url.hostname.replace(/\.$/, '') || null
+  } catch {
+    return null
+  }
+}
```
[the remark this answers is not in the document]
llun: Addressed in 3771016c: normalizeDomain now preserves and normalizes wildcard domains like *.example.social, with coverage.

app/api/v1/admin/domain_blocks/route.ts
```diff
@@ -0,0 +1,82 @@
...
+  HttpMethod.enum.GET,
+  HttpMethod.enum.POST
+]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+export const GET = traceApiRoute(
+  'adminListDomainBlocks',
+  AdminApiGuard(CORS_HEADERS, async (req: NextRequest, { database }) => {
+    const blocks = await database.getDomainBlocks({ limit: 10_000 })
```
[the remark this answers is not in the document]
llun: Addressed in 3771016c: admin domain block listing now accepts clamped limit/offset pagination and returns pagination headers.

app/api/v1/admin/domain_allows/route.ts
```diff
@@ -0,0 +1,75 @@
...
+  HttpMethod.enum.GET,
+  HttpMethod.enum.POST
+]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+export const GET = traceApiRoute(
+  'adminListDomainAllows',
+  AdminApiGuard(CORS_HEADERS, async (req: NextRequest, { database }) => {
+    const allows = await database.getDomainAllows({ limit: 10_000 })
```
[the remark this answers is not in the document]
llun: Addressed in 3771016c: admin domain allow listing now accepts clamped limit/offset pagination and returns pagination headers.

app/(timeline)/admin/federation/actions.ts
```diff
@@ -0,0 +1,108 @@
...
+  const database = await getAdminDatabase()
+  const source = KnownDomainBlocklistSourceId.safeParse(
+    getText(formData, 'source')
+  )
+  const sourceId = source.success
+    ? source.data
+    : redirectWithStatus('invalid-source')
+
+  const blocks = await fetchKnownDomainBlocklist(sourceId)
+  const result = await database.importDomainBlocks({ blocks })
```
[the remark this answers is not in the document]
llun: Addressed in 3771016c: blocklist import server action catches fetch/import failures and redirects with a user-friendly import-failed status.

app/api/v1/admin/domain_blocks/import/route.ts
```diff
@@ -0,0 +1,64 @@
...
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_422,
+        responseStatusCode: HTTP_STATUS.UNPROCESSABLE_ENTITY
+      })
+    }
+
+    const blocks = await fetchKnownDomainBlocklist(parsed.data)
+    const result = await database.importDomainBlocks({ blocks })
```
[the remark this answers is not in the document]
llun: Addressed in 3771016c: the import API wraps fetch/import failures and returns a generic bad request response via apiResponse.

app/api/v1/admin/domain_blocks/[id]/route.ts
```diff
@@ -0,0 +1,130 @@
...
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_404,
+        responseStatusCode: HTTP_STATUS.NOT_FOUND
+      })
+    }
+
+    return apiResponse({
+      req,
+      allowedMethods: CORS_HEADERS,
+      data: {}
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/admin/domain_blocks/[id]/route.ts]: Addressed in 3771016c: domain block DELETE now returns the deleted block, matching the allow delete behavior.

llun:
@codex revew

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const config = getConfig()
+  const localDomains = config.allowActorDomains?.length
+    ? config.allowActorDomains
+    : [config.host]
+
+  return localDomains.some((localDomain) =>
+    domainMatchesRule(domain, localDomain)
+  )
+}
```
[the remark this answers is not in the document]
llun: Addressed in fc1026a3: local federation domain checks now require exact host/allow-domain matches unless a wildcard is explicitly configured, with regression coverage for evil.local.test.

app/api/v1/instance/domain_blocks/route.ts
```diff
@@ -1,5 +1,29 @@
...
+  if (!database) {
+    return apiResponse({
+      req: req as NextRequest,
+      allowedMethods: CORS_HEADERS,
+      data: { error: 'Database unavailable' },
+      responseStatusCode: HTTP_STATUS.INTERNAL_SERVER_ERROR
+    })
+  }
+
+  const blocks = await database.getDomainBlocks({ limit: 10_000 })
```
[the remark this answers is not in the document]
llun: Addressed in fc1026a3: the public domain blocks API now supports bounded limit/offset pagination and returns pagination headers.

app/api/v1/admin/domain_allows/schema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const readRequestData = async (req: Request) => {
+  const contentType = req.headers.get('content-type') ?? ''
+  if (contentType.includes('application/json')) {
+    return req.json()
+  }
+
+  const formData = await req.formData()
+  return Object.fromEntries(formData.entries())
+}
```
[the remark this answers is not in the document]
llun: Addressed in fc1026a3: admin domain allow/block routes now use the shared getRequestBody helper instead of duplicated schema-local body parsing.

llun:
@codex revew

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,101 @@
...
+        .map((url) => getDomainFromUrl(url))
+        .filter((domain): domain is string => domain !== null)
+    )
+  ]
+  const domainResults = await Promise.all(
+    domains.map(async (domain) => ({
+      domain,
+      allowed: await canFederateWithDomain(database, domain)
+    }))
+  )
```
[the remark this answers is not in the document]
llun: Addressed in 8ef0a3e8: fanout filtering now uses batched domain rule lookups. The SQL layer fetches candidate rules for all domains with WHERE IN, and filterFederatedUrls uses one batched block lookup plus one batched allow lookup only in allowlist mode.

llun:
@codex revew

lib/database/sql/admin.ts
```diff
@@ -246,5 +394,311 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+    const normalized = normalizeDomain(domain)
+    if (!normalized) return null
+
+    const row = await database<SQLDomainFederationRule>(
+      'domain_federation_rules'
+    )
+      .where('type', 'block')
+      .whereIn('domain', getDomainRuleCandidates(normalized))
+      .orderByRaw('length(domain) DESC')
+      .orderByRaw("case when domain like '*.%' then 1 else 0 end asc")
```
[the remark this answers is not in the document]
llun: Already handled in the current code: the single-domain SQL lookup orders by length and then prioritizes non-wildcard domains over wildcard domains with `case when domain like '*.%' then 1 else 0 end asc`; the new batch path uses the same exact-over-wildcard comparator. No code change needed.

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    field += char
+  }
+
+  if (field || fields.length) {
+    pushRecord()
+  }
+
+  return records
+}
```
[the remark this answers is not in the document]
llun: Acknowledged, but not taking a dependency for this PR. The parser is intentionally narrow for Mastodon-compatible blocklist CSV and already has coverage for escaped quotes, quoted commas, quoted newlines, CRLF handling, and case-insensitive headers. A future parser-library swap can be considered separately if blocklist formats expand.

app/api/v1/instance/domain_blocks/route.ts
```diff
@@ -1,5 +1,54 @@
...
+  const { limit, offset } = parsedParams.data
+  const [blocks, stats] = await Promise.all([
+    database.getDomainBlocks({ limit, offset }),
+    database.getDomainFederationRuleStats()
+  ])
+
+  return apiResponse({
+    req: req as NextRequest,
+    allowedMethods: CORS_HEADERS,
+    data: blocks.map(toPublicDomainBlock),
```
llun: **Bug: Public endpoint exposes `silence` and `noop` blocks**

`getDomainBlocks()` returns all blocks regardless of severity. The public instance domain blocks endpoint should typically only surface `suspend`-severity blocks — `silence`/`noop` blocks are admin-level information and leaking them tells outsiders which domains are deprioritised rather than fully suspended.

```ts
const blocks = await database.getDomainBlocks({ limit, offset })
// Filter to only expose suspend-severity to the public:
const publicBlocks = blocks.filter((b) => b.severity === 'suspend')
data: publicBlocks.map(toPublicDomainBlock),
```

The `X-Total-Count` header should also reflect only the suspend count, so `getDomainFederationRuleStats()` (or a separate `suspendBlocks` stat) would need to track that separately.

---

lib/database/sql/admin.ts
```diff
@@ -246,5 +394,311 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+        .update(buildDomainBlockUpdate(params, now))
+      const updated = await this.getDomainBlockById(existing.id)
+      if (!updated) throw new Error('Failed to update domain block')
+      return updated
+    }
+
+    await database('domain_federation_rules').insert(row)
+    const created = await this.getDomainBlockById(row.id)
+    if (!created) throw new Error('Failed to create domain block')
+    return created
```
llun: **Bug: Non-atomic upsert — race condition under concurrent requests**

The read-then-write pattern here is not atomic. Two concurrent `createDomainBlock` calls for the same domain can both find no existing row and both attempt to `INSERT`, causing a unique constraint violation on `(type, domain)`.

`importDomainBlocks` already solves this correctly with `onConflict(['type', 'domain']).merge(...)`. `createDomainBlock` should use the same approach:

```ts
async createDomainBlock(params) {
  const now = new Date()
  const row = buildDomainBlockInsert(params, now)
  await database('domain_federation_rules')
    .insert(row)
    .onConflict(['type', 'domain'])
    .merge([
      'severity', 'rejectMedia', 'rejectReports',
      'privateComment', 'publicComment', 'obfuscate', 'source', 'updatedAt'
    ])
  const result = await this.getDomainBlockById(row.id)
  if (!result) throw new Error('Failed to upsert domain block')
  return result
}
```

---

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,294 @@
...
+        <h1 className="text-2xl font-bold">Federation</h1>
+        <p className="text-sm text-muted-foreground">
+          {federationMode === 'allowlist'
+            ? 'Limited federation mode'
+            : 'Open federation mode'}
+        </p>
+      </div>
+
+      {statusMessage && (
+        <div className="rounded-lg border border-green-200 bg-green-50 px-4 py-3 text-sm text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100">
```
llun: **UI: Error statuses use the same green success styling**

Error statuses (`import-failed`, `invalid-block-domain`, `invalid-allow-domain`, `invalid-source`) are shown with the same `border-green-200 bg-green-50 text-green-900` classes as success messages. They should use a red/destructive variant.

Suggested approach — differentiate by status type:

```tsx
const isError = status?.startsWith('invalid') || status === 'import-failed'

{statusMessage && (
  <div className={cn(
    "rounded-lg border px-4 py-3 text-sm",
    isError
      ? "border-red-200 bg-red-50 text-red-900 dark:border-red-900 dark:bg-red-950 dark:text-red-100"
      : "border-green-200 bg-green-50 text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100"
  )}>
    {statusMessage}
  </div>
)}
```

---

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,294 @@
...
+              <Label>
+                <Checkbox name="rejectReports" />
+                Reject reports
+              </Label>
+              <Label>
+                <Checkbox name="obfuscate" />
+                Obfuscate
+              </Label>
+            </div>
+            <Button type="submit">Save block</Button>
```
llun, line 327: **UI: `privateComment` field missing from the form**

`createDomainBlockAction` hardcodes `privateComment: null` (line 58 of `actions.ts`), but the admin API exposes it as a writable field. The UI form has `publicComment` but no corresponding `privateComment` textarea.

Either add the field to the form and wire it through the action, or explicitly document that private comments can only be set via the API. Leaving it silently as `null` in the UI while the API supports it is inconsistent.

---

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,294 @@
...
+          </form>
+        </div>
+      </section>
+
+      <section className="rounded-xl border bg-background/80 p-5 shadow-sm">
+        <h2 className="mb-4 text-lg font-semibold">Domain Blocks</h2>
+        {stats.blocks > blocks.length && (
+          <p className="mb-3 text-sm text-muted-foreground">
+            Showing first {blocks.length} of {stats.blocks} blocked domains.
+          </p>
```
llun: **UI: No pagination controls — first 100 entries only**

The truncation notice appears when there are more than 100 blocks/allows, but there are no "Next page" / "Load more" controls. An instance that imports the Oliphant tier 0 list gets thousands of entries — all but the first 100 are invisible and unreachable from the UI.

Either add pagination (e.g. `?page=` query param) or increase the limit and add a virtual/windowed list. At minimum the truncation notice should include a link to the admin API for programmatic access.

---

app/(timeline)/admin/federation/page.tsx
```diff
@@ -0,0 +1,294 @@
...
+        )}
+        <div className="space-y-2">
+          {blocks.length === 0 ? (
+            <p className="text-sm text-muted-foreground">No domains blocked</p>
+          ) : (
+            blocks.map((block) => {
+              const publicBlock = toPublicDomainBlock(block)
```
llun: **Nit: `toPublicDomainBlock` called just to access `publicComment`**

```tsx
const publicBlock = toPublicDomainBlock(block)
// ...
{publicBlock.comment ? ` - ${publicBlock.comment}` : ''}
```

In an admin page `block.publicComment` is directly accessible. Using the public-facing transformer here is confusing — it also runs the obfuscation logic on `block.domain` inside the transformer, even though the result is discarded. Prefer `block.publicComment` directly.

---

lib/services/guards/AdminApiGuard.ts
```diff
@@ -0,0 +1,45 @@
...
+    const admin = await getAdminFromSession(database, session)
+    if (!admin) {
+      return apiResponse({
+        req,
+        allowedMethods,
+        data: { error: 'Forbidden' },
+        responseStatusCode: HTTP_STATUS.FORBIDDEN
+      })
+    }
+
```
llun, line 51: **API auth: session-based guard makes the admin API unusable by external clients**

`AdminApiGuard` authenticates via `getServerAuthSession()` (cookie session), which works in a browser but not for programmatic API clients. Mastodon-compatible admin API clients authenticate with an OAuth bearer token that has admin scope.

The existing `OAuthGuard` in this project handles bearer tokens. The admin API routes should either:
1. Accept an admin-scoped OAuth token, or
2. Accept both (session OR admin bearer token)

As-is, `GET /api/v1/admin/domain_blocks` will always return 403 for any client that sends `Authorization: Bearer <token>` without a session cookie, making the endpoints Mastodon API-incompatible.

---

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,148 @@
...
+    (item) => item.id === sourceId
+  )
+  if (!source) throw new Error('Unknown domain blocklist source')
+
+  const response = await fetchImpl(source.url)
+  if (!response.ok) {
+    throw new Error(`Failed to fetch ${source.name}`)
+  }
+
+  const csv = await response.text()
```
llun: **No timeout or size limit on external fetch**

`fetchImpl(source.url)` has no timeout and `response.text()` has no size cap. A slow or unexpectedly large response from Codeberg (or any future source) could either hang the server action indefinitely or cause memory pressure.

```ts
const response = await fetchImpl(source.url, {
  signal: AbortSignal.timeout(30_000) // 30s max
})
if (!response.ok) throw new Error(`Failed to fetch ${source.name}`)

// Guard against unexpectedly large payloads (e.g. 10 MB)
const MAX_BYTES = 10 * 1024 * 1024
const contentLength = Number(response.headers.get('content-length') ?? 0)
if (contentLength > MAX_BYTES) throw new Error('Blocklist response too large')

const csv = await response.text()
if (csv.length > MAX_BYTES) throw new Error('Blocklist response too large')
```

---

app/api/v1/instance/domain_blocks/route.ts
```diff
@@ -1,5 +1,54 @@
...
-export const GET = traceApiRoute('getInstanceDomainBlocks', async () => {
-  return Response.json([])
+const CORS_HEADERS = [HttpMethod.enum.GET]
+const DomainBlockListQueryParams = z.object({
+  limit: z.coerce.number().int().min(1).max(1000).default(100),
+  offset: z.coerce.number().int().min(0).default(0)
+})
+
+export const GET = traceApiRoute('getInstanceDomainBlocks', async (req) => {
+  const database = getDatabase()
```
llun: **Nit: Repeated `req as NextRequest` casts suggest a type mismatch**

`traceApiRoute`'s callback receives a `NextRequest` in every other route in this project — they call `apiResponse({ req, ... })` without casting. Here every call does `req as NextRequest`, which means the inferred type of `req` is something else (likely `Request`). Check whether `traceApiRoute` needs an explicit generic or overload for this non-guarded route signature, rather than suppressing the mismatch with casts.

---

llun (commented):
## Review: feat: add federation domain rules

Good overall structure — the layering (migration → DB mixin → service → API → UI) is clean and the batch-optimised `filterFederatedUrls` path is well thought-out. The test coverage for the new service files is solid. A few issues need addressing before merge.

### Bugs

1. **Public endpoint leaks `silence`/`noop` blocks** — `GET /api/v1/instance/domain_blocks` returns all blocks regardless of severity. Only `suspend` blocks should be visible to the public; silence/noop are admin-level metadata. (Inline comment on `instance/domain_blocks/route.ts:47`.)

2. **Race condition in `createDomainBlock`** — read-then-write is not atomic; concurrent creates for the same domain will hit the unique constraint. Should use `onConflict().merge()` like `importDomainBlocks` already does. (Inline comment on `admin.ts:546-564`.)

### UI issues

3. **Error statuses shown in green success styling** — `import-failed`, `invalid-block-domain`, `invalid-allow-domain`, `invalid-source` all render with the green banner. Need a red/destructive variant. (Inline comment on `page.tsx:84`.)

4. **`privateComment` silently always `null` in the UI form** — the action hard-codes it to `null` even though the admin API exposes it as writable. Either add the field to the form or document the omission. (Inline comment on `page.tsx:149-193`.)

5. **No pagination past the first 100 entries** — large instances (especially after importing Oliphant tier 0) will have hundreds of hidden blocks with no way to navigate to them. (Inline comment on `page.tsx:216-219`.)

### Code quality

6. **`AdminApiGuard` uses session auth** — the new `/api/v1/admin/*` routes are unreachable by OAuth bearer-token clients, making them Mastodon API-incompatible. Should accept admin-scoped tokens via `OAuthGuard` or a combined check. (Inline comment on `AdminApiGuard.ts:32-43`.)

7. **No timeout or size limit on the external blocklist fetch** — `fetchKnownDomainBlocklist` can hang or consume unbounded memory. (Inline comment on `blocklistSources.ts:141-146`.)

8. **`toPublicDomainBlock` used in admin context just to get `publicComment`** — use `block.publicComment` directly. (Inline comment on `page.tsx:226`.)

9. **Repeated `req as NextRequest` casts** in `instance/domain_blocks/route.ts` — suggests a type-level mismatch worth fixing at the source. (Inline comment on `route.ts:17`.)

---

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,148 @@
...
+export const fetchKnownDomainBlocklist = async (
+  sourceId: KnownDomainBlocklistSourceId,
+  fetchImpl: FetchLike = fetch
+): Promise<ImportDomainBlockParams[]> => {
+  const source = KNOWN_DOMAIN_BLOCKLIST_SOURCES.find(
+    (item) => item.id === sourceId
+  )
+  if (!source) throw new Error('Unknown domain blocklist source')
+
+  const response = await fetchImpl(source.url)
```
llun: If this is backend service, don't use fetch but got

app/api/v1/instance/domain_blocks/route.ts
llun: Fixed in f623c30f: the public instance domain blocks endpoint now requests only suspend rules via the severity filter, and X-Total-Count uses the new suspendBlocks stat. Added route and SQL coverage for this.

lib/database/sql/admin.ts
llun: Fixed in f623c30f: createDomainBlock now uses insert(...).onConflict(["type", "domain"]).merge(...) and fetches the upserted row by type/domain. I also made createDomainAllow conflict-safe for the same table constraint and added SQL upsert coverage.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in f623c30f: error statuses now render with the red/destructive banner classes. I verified the import-failed state in the browser.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in f623c30f: added the privateComment textarea to the admin form. The server action was already reading privateComment from the form data, so the UI now matches the action/API behavior.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in f623c30f: added blockOffset/allowOffset pagination controls with range labels for both domain blocks and domain allows. Verified the controls in the browser.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in f623c30f: removed toPublicDomainBlock from the admin page and read block.publicComment directly.

lib/services/guards/AdminApiGuard.ts
llun: Not changing this in this PR. The project currently defines only general OAuth scopes such as read/write/follow/push and does not define or advertise an admin OAuth scope. Accepting the existing bearer tokens for admin routes would widen privilege semantics; adding scoped admin OAuth support should be a separate auth design change. Keeping these first-party admin routes on the session-based admin guard here.

lib/services/federation/blocklistSources.ts
llun: Fixed in f623c30f: known blocklist downloads now use the got-backed request helper with a 30s request timeout, no retries, content-length/body-size caps, and tests for oversized responses.

app/api/v1/instance/domain_blocks/route.ts
llun: Fixed in f623c30f: the trace callback is explicitly typed as NextRequest and the req as NextRequest casts are gone.

lib/services/federation/blocklistSources.ts
llun: Fixed in f623c30f: replaced the direct fetch path with the existing got-backed request helper for backend blocklist downloads.

llun:
@codex review

llun (commented):
## Review: feat: add federation domain rules

Good overall structure — the layering (migration → DB mixin → service → API → UI) is clean and the batch-optimised `filterFederatedUrls` path is well thought-out. The test coverage for the new service files is solid. A few issues need addressing before merge.

### Bugs

1. **Public endpoint leaks `silence`/`noop` blocks** — `GET /api/v1/instance/domain_blocks` returns all blocks regardless of severity. Only `suspend` blocks should be visible to the public; silence/noop are admin-level metadata. (Inline comment on `instance/domain_blocks/route.ts:47`.)

2. **Race condition in `createDomainBlock`** — read-then-write is not atomic; concurrent creates for the same domain will hit the unique constraint. Should use `onConflict().merge()` like `importDomainBlocks` already does. (Inline comment on `admin.ts:546-564`.)

### UI issues

3. **Error statuses shown in green success styling** — `import-failed`, `invalid-block-domain`, `invalid-allow-domain`, `invalid-source` all render with the green banner. Need a red/destructive variant. (Inline comment on `page.tsx:84`.)

4. **`privateComment` silently always `null` in the UI form** — the action hard-codes it to `null` even though the admin API exposes it as writable. Either add the field to the form or document the omission. (Inline comment on `page.tsx:149-193`.)

5. **No pagination past the first 100 entries** — large instances (especially after importing Oliphant tier 0) will have hundreds of hidden blocks with no way to navigate to them. (Inline comment on `page.tsx:216-219`.)

### Code quality

6. **`AdminApiGuard` uses session auth** — the new `/api/v1/admin/*` routes are unreachable by OAuth bearer-token clients, making them Mastodon API-incompatible. Should accept admin-scoped tokens via `OAuthGuard` or a combined check. (Inline comment on `AdminApiGuard.ts:32-43`.)

7. **No timeout or size limit on the external blocklist fetch** — `fetchKnownDomainBlocklist` can hang or consume unbounded memory. (Inline comment on `blocklistSources.ts:141-146`.)

8. **`toPublicDomainBlock` used in admin context just to get `publicComment`** — use `block.publicComment` directly. (Inline comment on `page.tsx:226`.)

9. **Repeated `req as NextRequest` casts** in `instance/domain_blocks/route.ts` — suggests a type-level mismatch worth fixing at the source. (Inline comment on `route.ts:17`.)

app/api/v1/instance/domain_blocks/route.ts
llun: **Bug: Public endpoint leaks `silence`/`noop` blocks**

`getDomainBlocks()` returns all blocks regardless of severity. The public instance domain blocks endpoint should typically only surface `suspend`-severity blocks — `silence`/`noop` blocks are admin-level information and leaking them tells outsiders which domains are deprioritised rather than fully suspended.

```ts
const blocks = await database.getDomainBlocks({ limit, offset })
// Filter to only expose suspend-severity to the public:
const publicBlocks = blocks.filter((b) => b.severity === 'suspend')
data: publicBlocks.map(toPublicDomainBlock),
```

The `X-Total-Count` header should also reflect only the suspend count, so `getDomainFederationRuleStats()` (or a separate `suspendBlocks` stat) would need to track that separately.

lib/database/sql/admin.ts
llun: **Bug: Non-atomic upsert — race condition under concurrent requests**

The read-then-write pattern here is not atomic. Two concurrent `createDomainBlock` calls for the same domain can both find no existing row and both attempt to `INSERT`, causing a unique constraint violation on `(type, domain)`.

`importDomainBlocks` already solves this correctly with `onConflict(["type", "domain"]).merge(...)`. `createDomainBlock` should use the same approach:

```ts
async createDomainBlock(params) {
  const now = new Date()
  const row = buildDomainBlockInsert(params, now)
  await database('domain_federation_rules')
    .insert(row)
    .onConflict(['type', 'domain'])
    .merge([
      'severity', 'rejectMedia', 'rejectReports',
      'privateComment', 'publicComment', 'obfuscate', 'source', 'updatedAt'
    ])
  const result = await this.getDomainBlockById(row.id)
  if (!result) throw new Error('Failed to upsert domain block')
  return result
}
```

app/(timeline)/admin/federation/page.tsx
llun: **UI: Error statuses use the same green success styling**

Error statuses (`import-failed`, `invalid-block-domain`, `invalid-allow-domain`, `invalid-source`) are shown with the same `border-green-200 bg-green-50 text-green-900` classes as success messages. They should use a red/destructive variant.

Suggested approach — differentiate by status type:

```tsx
const isError = status?.startsWith('invalid') || status === 'import-failed'

{statusMessage && (
  <div className={cn(
    "rounded-lg border px-4 py-3 text-sm",
    isError
      ? "border-red-200 bg-red-50 text-red-900 dark:border-red-900 dark:bg-red-950 dark:text-red-100"
      : "border-green-200 bg-green-50 text-green-900 dark:border-green-900 dark:bg-green-950 dark:text-green-100"
  )}>
    {statusMessage}
  </div>
)}
```

app/(timeline)/admin/federation/page.tsx
llun, line 327: **UI: `privateComment` field missing from the form**

`createDomainBlockAction` hardcodes `privateComment: null` (line 58 of `actions.ts`), but the admin API exposes it as a writable field. The UI form has `publicComment` but no corresponding `privateComment` textarea.

Either add the field to the form and wire it through the action, or explicitly document that private comments can only be set via the API. Leaving it silently as `null` in the UI while the API supports it is inconsistent.

app/(timeline)/admin/federation/page.tsx
llun: **UI: No pagination controls — first 100 entries only**

The truncation notice appears when there are more than 100 blocks/allows, but there are no "Next page" / "Load more" controls. An instance that imports the Oliphant tier 0 list gets thousands of entries — all but the first 100 are invisible and unreachable from the UI.

Either add pagination (e.g. `?page=` query param) or increase the limit and add a virtual/windowed list. At minimum the truncation notice should include a link to the admin API for programmatic access.

app/(timeline)/admin/federation/page.tsx
llun: **Nit: `toPublicDomainBlock` called just to access `publicComment`**

```tsx
const publicBlock = toPublicDomainBlock(block)
// ...
{publicBlock.comment ? ` - ${publicBlock.comment}` : ""}
```

In an admin page `block.publicComment` is directly accessible. Using the public-facing transformer here is confusing — it also runs the obfuscation logic on `block.domain` inside the transformer, even though the result is discarded. Prefer `block.publicComment` directly.

lib/services/guards/AdminApiGuard.ts
llun, line 51: **API auth: session-based guard makes the admin API unusable by external clients**

`AdminApiGuard` authenticates via `getServerAuthSession()` (cookie session), which works in a browser but not for programmatic API clients. Mastodon-compatible admin API clients authenticate with an OAuth bearer token that has admin scope.

The existing `OAuthGuard` in this project handles bearer tokens. The admin API routes should either:
1. Accept an admin-scoped OAuth token, or
2. Accept both (session OR admin bearer token)

As-is, `GET /api/v1/admin/domain_blocks` will always return 403 for any client that sends `Authorization: Bearer <token>` without a session cookie, making the endpoints Mastodon API-incompatible.

lib/services/federation/blocklistSources.ts
llun: **No timeout or size limit on external fetch**

`fetchImpl(source.url)` has no timeout and `response.text()` has no size cap. A slow or unexpectedly large response from Codeberg (or any future source) could either hang the server action indefinitely or cause memory pressure.

```ts
const response = await fetchImpl(source.url, {
  signal: AbortSignal.timeout(30_000) // 30s max
})
if (!response.ok) throw new Error(`Failed to fetch ${source.name}`)

// Guard against unexpectedly large payloads
const MAX_BYTES = 10 * 1024 * 1024
const contentLength = Number(response.headers.get('content-length') ?? 0)
if (contentLength > MAX_BYTES) throw new Error('Blocklist response too large')

const csv = await response.text()
if (csv.length > MAX_BYTES) throw new Error('Blocklist response too large')
```

app/api/v1/instance/domain_blocks/route.ts
llun: **Nit: Repeated `req as NextRequest` casts suggest a type mismatch**

`traceApiRoute`'s callback receives a `NextRequest` in every other route in this project — they call `apiResponse({ req, ... })` without casting. Here every call does `req as NextRequest`, which means the inferred type of `req` is something else (likely `Request`). Check whether `traceApiRoute` needs an explicit generic or overload for this non-guarded route signature, rather than suppressing the mismatch with casts.

llun (commented):
## Review: feat: add federation domain rules (updated)

Reviewing the current tip of the branch (`581e97c`). The earlier round of feedback has been addressed well — error/success styling, pagination, `privateComment` in the UI form, the atomic `createDomainBlock` upsert, the public endpoint severity filter, the fetch timeout/size limit, and the `req as NextRequest` casts are all fixed. Good work on those.

Four issues remain that should be resolved before merge.

### Bugs / behaviour gaps

1. **`silence` and `noop` blocks are stored and advertised but have no enforcement** — `canFederateWithDomain` only suspends on `severity === "suspend"`; silence and noop do nothing at the federation layer. Admins creating or importing these severity levels will see them saved and returned by the API but find that activities still flow normally. Either wire up the silence effect (hide from timelines/search) or document the limitation clearly — right now the UI gives no indication that only `suspend` actually blocks anything. (Inline comment on `domainPolicy.ts:64`.)

2. **An exact-domain block also silently covers all its subdomains** — `getDomainRuleCandidates("sub.example.com")` includes `example.com`, so a block on `example.com` blocks `sub.example.com` too. On Mastodon an exact-domain block only covers that domain; subdomains require `*.example.com`. Since the import path parses Mastodon-format CSV (which uses only exact domains), every imported block will silently cover subdomains. This should either be documented in the UI ("also blocks subdomains") or aligned with Mastodon semantics. (Inline comment on `admin.ts:104-114`.)

### Code quality

3. **`AdminApiGuard` accepts only cookie sessions** — the `/api/v1/admin/domain_blocks` and `/api/v1/admin/domain_allows` routes are inaccessible to any client that presents an OAuth bearer token (e.g. external Mastodon-compatible admin tools). All other write APIs in this project use `OAuthGuard`. Consider adding an OAuth admin-scope path. (Inline comment on `AdminApiGuard.ts:33`.)

4. **`requestWithResponseSizeLimit` does not abort on a large `Content-Length` header** — the `response` event captures headers but does not check `content-length` before downloading. The size guard only fires byte-by-byte in `data` events, so the full first chunk is always downloaded before an oversized response is rejected. Checking `content-length` in the `response` handler lets you abort before any body is received. (Inline comment on `request.ts:76`.)

lib/services/federation/domainPolicy.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export const canFederateWithDomain = async (
+  database: Database,
+  value: string
+): Promise<boolean> => {
+  const domain = getDomainFromUrl(value)
+  if (!domain) return false
+  if (isLocalFederationDomain(domain)) return true
+
+  if (await isDomainBlocked(database, domain)) return false
```
llun, line 64: **`silence` and `noop` blocks have no enforcement**

`shouldSuspendDomainBlock` returns `true` only when `severity === "suspend"`. Blocks with `silence` or `noop` severity pass through here without any effect — activities, follows, and inbox deliveries are not suppressed.

Mastodon defines these severities as:
- `noop` — no action beyond storing the rule
- `silence` — hides posts from public timelines and federated search, but allows direct/follow interactions
- `suspend` — rejects all federation

If silence/noop enforcement is intentionally deferred, add a UI hint (e.g. "Note: only Suspend blocks affect federation") so admins are not misled. If it is meant to be implemented, the timeline and search queries need to filter out actors from silenced domains.

lib/database/sql/admin.ts
```diff
@@ -52,6 +66,140 @@ const toDomainActor = (row: SQLActor): Actor => {
...
+  if (domain === '*') return ['*']
+
+  const parts = domain.split('.')
+  const exactCandidates = parts.map((_, index) => parts.slice(index).join('.'))
+  const wildcardCandidates = exactCandidates
+    .slice(1)
+    .map((candidate) => `*.${candidate}`)
+
+  return [...new Set([...exactCandidates, ...wildcardCandidates, '*'])]
+}
```
llun, line 117: **Exact-domain block implicitly covers all subdomains**

`getDomainRuleCandidates("sub.example.com")` returns `["sub.example.com", "example.com", "com", "*.example.com", "*.com", "*"]`. This means a block stored as `example.com` will match a federation check for `sub.example.com`.

On Mastodon this is not the case — an exact-domain block covers only that domain; `*.example.com` is required to cover subdomains. Because the CSV import parses Mastodon-format lists (which use only exact domain entries), every imported block will silently cover subdomains here even though the source list did not intend it.

Two options:
1. Remove parent domains from `exactCandidates` (keep only the exact match and the wildcard candidates), aligning with Mastodon semantics.
2. Keep the current broad-match behaviour but display a note in the UI on each block row ("covers subdomains") and in the CSV import summary.

lib/services/guards/AdminApiGuard.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!database) {
+      return apiResponse({
+        req,
+        allowedMethods,
+        data: { error: 'Database unavailable' },
+        responseStatusCode: HTTP_STATUS.INTERNAL_SERVER_ERROR
+      })
+    }
+
+    const session = await getServerAuthSession()
```
llun, line 37: **Session-only auth blocks OAuth bearer-token clients**

`getServerAuthSession()` authenticates via a cookie session. Any client that calls this endpoint with `Authorization: Bearer <token>` (the standard for Mastodon admin API clients) will receive 403 even with a valid admin token.

All other write APIs in this project go through `OAuthGuard`. The admin routes should accept at least an admin-scoped OAuth token in addition to (or instead of) cookie sessions, otherwise the `/api/v1/admin/*` namespace is not usable by external tooling.

lib/utils/request.ts
```diff
@@ -50,5 +58,66 @@ export const request = ({
...
+  maxResponseSize: number
+): Promise<RequestResult> =>
+  new Promise((resolve, reject) => {
+    const chunks: Buffer[] = []
+    let bodyBytes = 0
+    let statusCode = 0
+    let headers: Record<string, string | string[] | undefined> = {}
+    const stream = got.stream(url, options)
+
+    stream.on('response', (response) => {
```
llun, line 76: **Early abort on `Content-Length` header not implemented**

The `response` handler captures `statusCode` and `headers` but does not inspect `content-length`. The actual size guard fires in the `data` handler, which means the connection stays open and at least one chunk is downloaded before an oversized response is rejected.

Add a check here to abort immediately when the declared size already exceeds the limit:

```ts
stream.on("response", (response) => {
  statusCode = response.statusCode
  headers = response.headers
  const declared = Number(response.headers["content-length"] ?? 0)
  if (declared > maxResponseSize) {
    stream.destroy(new Error("Response body too large"))
  }
})
```

This avoids downloading any data for responses that are obviously too large.

app/api/v1/instance/domain_blocks/route.ts
llun: Fixed in `f623c30f`: the public domain-blocks endpoint now returns only `severity: suspend` entries and its `X-Total-Count` uses the suspend-only stat.

lib/database/sql/admin.ts
llun: Fixed in `f623c30f`: `createDomainBlock` now uses `onConflict(['type', 'domain']).merge(...)` and reads back the upserted row.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in `f623c30f`: error statuses now use the destructive/red styling instead of the success styling.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in `f623c30f`: the admin form now includes `privateComment` and the server action persists it.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in `f623c30f`: the federation admin page now has pagination controls for both blocks and allows.

app/(timeline)/admin/federation/page.tsx
llun: Fixed in `f623c30f`: the admin page now reads `block.publicComment` directly instead of calling the public transformer.

lib/services/guards/AdminApiGuard.ts
llun: Fixed in `b0df5159`: `AdminApiGuard` now accepts OAuth bearer tokens, requires read scope for GET and write scope for non-GET, and still requires the token actor account to be admin.

lib/services/federation/blocklistSources.ts
llun: Fixed in `581e97c2`: known blocklist downloads now go through the got-backed request helper with a timeout, no retry, and a response-size cap.

app/api/v1/instance/domain_blocks/route.ts
llun: Fixed in `f623c30f`: the callback is typed as `NextRequest`, so the repeated casts were removed.

app/api/v1/accounts/follow/route.ts
```diff
@@ -56,6 +57,15 @@ export const POST = traceApiRoute(
     const { database, currentActor } = context
     const body = await req.json()
     const { target } = FollowRequest.parse(body)
+    if (!(await canFederateWithDomain(database, target))) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: { status: 'Forbidden' },
+        responseStatusCode: HTTP_STATUS.FORBIDDEN
+      })
```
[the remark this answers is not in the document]
llun: Fixed in `b0df5159`: the legacy DELETE follow endpoint now checks `canFederateWithDomain` before sending outbound Undo and only updates local follow state for blocked domains.

lib/database/sql/admin.ts
```diff
@@ -246,5 +394,319 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+    const normalized = normalizeDomain(domain)
+    if (!normalized) return null
+
+    const row = await database<SQLDomainFederationRule>(
+      'domain_federation_rules'
+    )
+      .where('type', 'block')
+      .whereIn('domain', getDomainRuleCandidates(normalized))
+      .orderByRaw('length(domain) DESC')
+      .orderByRaw("case when domain like '*.%' then 1 else 0 end asc")
```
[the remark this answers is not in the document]
llun: Fixed in `b0df5159`: domain-rule matching now prioritizes exact rules over wildcard rules, and the SQL ordering follows the same precedence.

lib/services/federation/domainPolicy.ts
llun: Addressed in `b0df5159` by adding UI copy that clarifies only Suspend rejects federation; Silence and Noop are stored as Mastodon-compatible metadata. Full silence timeline/search enforcement is broader behavior and intentionally outside this PR.

lib/database/sql/admin.ts
llun: Fixed in `b0df5159`: exact-domain rules now match only that domain; wildcard rules are required for subdomains, including imported Mastodon-format exact domains.

lib/services/guards/AdminApiGuard.ts
llun: Fixed in `b0df5159`: admin API routes now allow OAuth bearer-token clients through `AdminApiGuard`, while still requiring the authenticated actor account to be admin.

lib/utils/request.ts
llun: Fixed in `b0df5159`: request streaming now aborts immediately when `content-length` exceeds `maxResponseSize`, with direct test coverage.

llun:
@codex review

lib/database/sql/admin.ts
```diff
@@ -52,6 +66,144 @@ const toDomainActor = (row: SQLActor): Actor => {
...
+      const candidates = new Set(getDomainRuleCandidates(domain))
+      const row =
+        sortedRows.find((candidate) => candidates.has(candidate.domain)) ?? null
+
+      return [domain, row ? toDomainRule(row) : null]
+    })
+  )
+}
+
+const SQL_BATCH_SIZE = 50
```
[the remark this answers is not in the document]
llun: Fixed in `631a0437`: `SQL_BATCH_SIZE` is now 500, and the import batching test now inserts 510 new rows plus one existing row so it still exercises multiple chunks.

llun:
@codex review

app/api/v1/admin/domain_blocks/schema.ts
```diff
@@ -0,0 +1,39 @@
...
+    .trim()
+    .max(10_000)
+    .optional()
+    .transform((value) => value || null),
+  obfuscate: Booleanish.default(false)
+})
+
+export const DomainBlockUpdateRequest = DomainBlockRequest.omit({
+  domain: true
+}).partial()
```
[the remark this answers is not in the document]
llun: Fixed in `767bbf35`: `DomainBlockUpdateRequest` is now defined separately without create defaults, omitted fields stay `undefined`, and blank comments still clear to `null`. Added PUT route coverage for partial updates and clearing comments.

llun:
@codex review

llun (commented):
## Final review — `feat: add federation domain rules` (`767bbf3`)

All four issues raised in the previous review have been addressed. This is a clean, thorough implementation.

### ✅ All previous issues resolved

| # | Issue | Fix |
|---|-------|-----|
| 1 | `silence`/`noop` blocks silently no-op | UI now shows "Only Suspend rejects federation. Silence and Noop are stored for Mastodon-compatible metadata." |
| 2 | Exact-domain block covering subdomains | `getDomainRuleCandidates` now emits only `[domain, *.parent, *.grandparent, *]` — no parent exact candidates. `domainMatchesRule` uses strict equality for non-wildcard rules. |
| 3 | `AdminApiGuard` session-only auth | Guard now tries cookie session first; falls back to `OAuthGuard` with `read`/`write` scope + admin role check. Tests added. |
| 4 | No early abort on `Content-Length` header | `response` event handler now checks the declared length and calls `stream.destroy()` before downloading any body. |

### Additional improvements noticed

- `SQL_BATCH_SIZE` raised from 50 → 500, appropriate for the chunked import path.
- `compareDomainRuleRows` sorting fixed: wildcards correctly deprioritised before the length comparison.
- `getDomainBlockForDomain` / `getDomainAllowForDomain` SQL ordering fixed: `*` and `*.x` rows now share score 1 (exact-match rows score 0), so the most-specific rule wins.
- `DomainBlockUpdateRequest` schema rewritten as an explicit partial object (not `.omit().partial()`) so comment fields can be explicitly set to `null` via empty strings.
- New tests for `AdminApiGuard` bearer-token paths and `PUT /api/v1/admin/domain_blocks/:id` partial updates.

### One minor nit (non-blocking)

The `getRequiredOAuthScopes` helper maps GET → `read`, everything else → `write`. This works correctly (the admin role check is the real gate), but using the generic user scopes means an admin with a plain `read` token could list domain blocks, which is intentional here. If Mastodon-compatible `admin:read` / `admin:write` scopes are ever added to this project, the guard should be updated to require those instead.

lib/services/guards/AdminApiGuard.ts
```diff
@@ -0,0 +1,68 @@
...
+type AdminApiHandle<P> = (
+  request: NextRequest,
+  context: {
+    database: Database
+    params: Promise<P>
+  }
+) => Promise<Response> | Response
+
+const getRequiredOAuthScopes = (method: string): Scope[] =>
+  method === HttpMethod.enum.GET ? [Scope.enum.read] : [Scope.enum.write]
```
llun, line 22: **Nit (non-blocking): generic `read`/`write` scopes instead of admin-specific scopes**

`getRequiredOAuthScopes` maps GET → `[Scope.enum.read]` and everything else → `[Scope.enum.write]`. The admin role check at line 56 is the real gate, so this is safe. Worth revisiting if the project ever adds `admin:read` / `admin:write` OAuth scopes — at that point these routes should require the narrower admin scopes rather than the general-purpose user ones.

llun: No code change for this one: the project currently only defines generic `read`/`write` OAuth scopes, and `AdminApiGuard` also requires the authenticated actor account role to be `admin`. If `admin:read` / `admin:write` scopes are added later, this guard should be narrowed then.

lib/database/sql/admin.ts
```diff
@@ -246,5 +398,323 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+
+    const rows = await database<SQLDomainFederationRule>(
+      'domain_federation_rules'
+    )
+      .where('type', 'block')
+      .whereIn('domain', [
+        ...new Set(normalizedDomains.flatMap(getDomainRuleCandidates))
+      ])
+
+    return resolveDomainRuleMatches(normalizedDomains, rows, toDomainBlock)
```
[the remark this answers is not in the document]
llun: Fixed in b53067d90776a034379861de6c670a2c4c180141. I added a shared batched domain-rule candidate lookup helper using the existing SQL batch size so block lookups no longer send one large whereIn candidate list. Added large-input coverage in lib/database/sql/admin.test.ts.

lib/database/sql/admin.ts
```diff
@@ -246,5 +398,323 @@ export const AdminSQLDatabaseMixin = (database: Knex): AdminDatabase => ({
...
+
+    const rows = await database<SQLDomainFederationRule>(
+      'domain_federation_rules'
+    )
+      .where('type', 'allow')
+      .whereIn('domain', [
+        ...new Set(normalizedDomains.flatMap(getDomainRuleCandidates))
+      ])
+
+    return resolveDomainRuleMatches(normalizedDomains, rows, toDomainAllow)
```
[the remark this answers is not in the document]
llun: Fixed in b53067d90776a034379861de6c670a2c4c180141. Domain allow matching now uses the same batched candidate-row helper as block matching, with large-input test coverage for both paths.

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    field += char
+  }
+
+  if (field || fields.length) {
+    pushRecord()
+  }
+
+  return records
+}
```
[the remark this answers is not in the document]
llun: Fixed in b53067d90776a034379861de6c670a2c4c180141. I refined the CSV state machine so a quote only starts quoted mode at the beginning of a field, while embedded quotes in unquoted fields remain literal. Added a regression test for that case.

app/(timeline)/admin/federation/actions.ts
```diff
@@ -0,0 +1,116 @@
...
+  let skipped = 0
+  try {
+    const blocks = await downloadKnownDomainBlocklist(sourceId)
+    const result = await database.importDomainBlocks({ blocks })
+    created = result.created
+    updated = result.updated
+    skipped = result.skipped
+  } catch {
+    redirectWithStatus('import-failed')
+  }
```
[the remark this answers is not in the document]
llun: Fixed in b53067d90776a034379861de6c670a2c4c180141. The server action now logs the failed known-blocklist import with the selected source id and thrown error. The downloader also reads the got-backed response body and includes JSON message/error details, or raw text fallback, in the thrown error for logging.

lib/services/federation/domainRules.ts
```diff
@@ -0,0 +1,100 @@
...
+  const domainToParse = hasWildcard ? trimmed.slice(2) : trimmed
+  const withScheme = /^[a-z][a-z0-9+.-]*:\/\//i.test(domainToParse)
+    ? domainToParse
+    : `https://${domainToParse}`
+
+  try {
+    const url = new URL(withScheme)
+    const normalized = url.hostname.replace(/\.$/, '')
+    if (!normalized) return null
+    return hasWildcard ? `*.${normalized}` : normalized
```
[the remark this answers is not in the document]
llun: Fixed in b53067d90776a034379861de6c670a2c4c180141. normalizeDomain now rejects normalized domain values longer than 255 characters before they can reach the domain_federation_rules.domain column, with test coverage for overlong input.

llun:
@codex review

lib/services/federation/blocklistSources.ts
```diff
@@ -0,0 +1,214 @@
...
+      field += '"'
+      i++
+      continue
+    }
+
+    if (char === '"' && atFieldStart) {
+      inQuotes = true
+      atFieldStart = false
+      continue
+    }
```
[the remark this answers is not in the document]
llun: Addressed in cde7b54e407caa2a7bf0419476fb6ba38bf0059b. The manual CSV parser was removed and the blocklist import now uses csv-parse/sync with relax_quotes, relax_column_count, BOM handling, and blank-record filtering retained in our code.

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -20,6 +21,10 @@ export const ActivityPubVerifySenderGuard =
     const signatureParts = await parse(requestSignature)
     if (!signatureParts.keyId) return apiErrorResponse(400)
 
+    if (!(await canFederateWithDomain(database, signatureParts.keyId))) {
```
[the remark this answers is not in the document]
llun: Fixed in c96a008b90acf3ded6347fd4c997134f2c694846. The shared inbox now checks the parsed activity.actor domain with canFederateWithDomain before deriving/enqueuing a job, so a blocked actor cannot bypass enforcement through a different signature key host.

lib/actions/utils.ts
```diff
@@ -10,6 +11,10 @@ export const recordActorIfNeeded = async ({
   actorId,
   database
 }: RecordActorIfNeededParams): Promise<Actor | undefined> => {
+  if (!(await canFederateWithDomain(database, actorId))) {
+    return undefined
```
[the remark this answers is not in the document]
llun: Fixed in c96a008b90acf3ded6347fd4c997134f2c694846. I added an explicit blocked-domain assertion before content writes in createNoteJob, createPollJob, createAnnounceJob, and createPollVoteJob; blocked actors now throw before DB writes, while missing actor-profile fetches preserve the existing behavior. Added route/action/job tests for these cases.

PR #605, opened 2026-04-27
feat: support signed authorized fetches

llun:
## Summary
- Sign outbound ActivityPub GET requests for note, actor, reply collection, boost, and follow-resolution fetches using the current actor or a local fallback signer.
- Include query strings in HTTP Signature `(request-target)` values so authorized fetch signatures verify correctly.
- Add database support for selecting a local federation signing actor and tests for signed GET coverage.

## Validation
- yarn run prettier --write .
- yarn lint (passes with existing warnings in auth/fitness files)
- yarn build
- yarn test

## Notes
- No migration was needed; this reuses existing actor private keys.

### DIFF
--- a/lib/actions/utils.ts
+++ b/lib/actions/utils.ts
@@ -6,6 +6,7 @@ import { Actor } from '@/lib/types/domain/actor'
 interface RecordActorIfNeededParams {
   actorId: string
   database: Database
+  signingActor?: Actor
 }
 
 export class BlockedFederationDomainError extends Error {
@@ -26,7 +27,8 @@ export const assertActorCanFederate = async ({
 
 export const recordActorIfNeeded = async ({
   actorId,
-  database
+  database,
+  signingActor
 }: RecordActorIfNeededParams): Promise<Actor | undefined> => {
   await assertActorCanFederate({ actorId, database })
 
@@ -38,7 +40,7 @@ export const recordActorIfNeeded = async ({
     return existingActor
   }
   if (!existingActor) {
-    const person = await getActorPerson({ actorId })
+    const person = await getActorPerson({ actorId, signingActor })
     if (!person) return
     const actor = await database.createActor({
       actorId,
@@ -57,7 +59,7 @@ export const recordActorIfNeeded = async ({
   const currentTime = Date.now()
   // Update actor if it's older than 3 day
   if (currentTime - existingActor.updatedAt > 3 * 86_400_000) {
-    const person = await getActorPerson({ actorId })
+    const person = await getActorPerson({ actorId, signingActor })
     if (!person) return undefined
     const actor = await database.updateActor({
       actorId,
--- a/lib/actions/utils.ts
+++ b/lib/actions/utils.ts
@@ -43,8 +43,10 @@ export const recordActorIfNeeded = async ({
   }
 
   const getResolvedSigningActor = async () => {
-    const resolvedSigningActor =
-      signingActor ?? (await getFederationSigningActor(database))
+    const resolvedSigningActor = await getFederationSigningActor(
+      database,
+      signingActor
+    )
     if (!resolvedSigningActor) {
       logger.warn({
         message: 'Fetching remote actor without a federation signing actor',
--- a/lib/actions/utils.ts
+++ b/lib/actions/utils.ts
@@ -1,7 +1,9 @@
 import { getActorPerson } from '@/lib/activities/getActorPerson'
 import { Database } from '@/lib/database/types'
 import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
+import { getFederationSigningActor } from '@/lib/services/federation/getFederationSigningActor'
 import { Actor } from '@/lib/types/domain/actor'
+import { logger } from '@/lib/utils/logger'
 
 interface RecordActorIfNeededParams {
   actorId: string
@@ -39,8 +41,24 @@ export const recordActorIfNeeded = async ({
   if (existingActor?.privateKey) {
     return existingActor
   }
+
+  const getResolvedSigningActor = async () => {
+    const resolvedSigningActor =
+      signingActor ?? (await getFederationSigningActor(database))
+    if (!resolvedSigningActor) {
+      logger.warn({
+        message: 'Fetching remote actor without a federation signing actor',
+        actorId
+      })
+    }
+    return resolvedSigningActor
+  }
+
   if (!existingActor) {
-    const person = await getActorPerson({ actorId, signingActor })
+    const person = await getActorPerson({
+      actorId,
+      signingActor: await getResolvedSigningActor()
+    })
     if (!person) return
     const actor = await database.createActor({
       actorId,
@@ -59,7 +77,10 @@ export const recordActorIfNeeded = async ({
   const currentTime = Date.now()
   // Update actor if it's older than 3 day
   if (currentTime - existingActor.updatedAt > 3 * 86_400_000) {
-    const person = await getActorPerson({ actorId, signingActor })
+    const person = await getActorPerson({
+      actorId,
+      signingActor: await getResolvedSigningActor()
+    })
     if (!person) return undefined
     const actor = await database.updateActor({
       actorId,
--- a/lib/activities/getActorPosts.ts
+++ b/lib/activities/getActorPosts.ts
@@ -56,7 +56,10 @@ export const getActorPosts: GetActorPostsFunction = async ({
             })
             if (localStatus) return localStatus
 
-            const note = await getNote({ statusId: item.object })
+            const note = await getNote({
+              statusId: item.object,
+              signingActor
+            })
             if (!note) return null
             const originalStatus = fromNote(note)
             if (actor) originalStatus.actor = actor
--- a/lib/activities/index.ts
+++ b/lib/activities/index.ts
@@ -42,9 +42,11 @@ import { getTracer } from '@/lib/utils/trace'
 
 interface GetNoteParams {
   statusId: string
+  signingActor?: Actor
 }
 export const getNote = async ({
-  statusId
+  statusId,
+  signingActor
 }: GetNoteParams): Promise<Note | null> =>
   getTracer().startActiveSpan(
     'activities.getNote',
@@ -53,7 +55,12 @@ export const getNote = async ({
       try {
         const { statusCode, body } = await request({
           url: statusId,
-          headers: { Accept: DEFAULT_ACCEPT }
+          headers: {
+            Accept: DEFAULT_ACCEPT,
+            ...(signingActor
+              ? signedHeaders(signingActor, 'get', statusId)
+              : {})
+          }
         })
         if (statusCode !== 200) return null
         return JSON.parse(body)
@@ -409,7 +416,10 @@ export const follow = async (
         actor: currentActor.id,
         object: targetActorId
       }
-      const person = await getActorPerson({ actorId: targetActorId })
+      const person = await getActorPerson({
+        actorId: targetActorId,
+        signingActor: currentActor
+      })
       const targetInbox = person?.inbox
       if (!targetInbox) {
         span.end()
@@ -467,7 +477,10 @@ export const unfollow = async (currentActor: Actor, follow: Follow) =>
         }
       }
 
-      const person = await getActorPerson({ actorId: follow.targetActorId })
+      const person = await getActorPerson({
+        actorId: follow.targetActorId,
+        signingActor: currentActor
+      })
       const targetInbox = person?.inbox ?? `${follow.targetActorId}/inbox`
 
       const method = 'POST'
--- a/lib/activities/index.ts
+++ b/lib/activities/index.ts
@@ -58,7 +58,7 @@ export const getNote = async ({
           headers: {
             Accept: DEFAULT_ACCEPT,
             ...(signingActor
-              ? signedHeaders(signingActor, 'get', statusId)
+              ? signedHeaders(signingActor, 'GET', statusId)
               : {})
           }
         })
@@ -111,12 +111,7 @@ export const sendNote = async ({ currentActor, inbox, note }: SendNoteParams) =>
           url: inbox,
           method,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              inbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, inbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           body: JSON.stringify(activity)
@@ -177,12 +172,7 @@ export const sendUpdateNote = async ({
           url: inbox,
           method,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              inbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, inbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           body: JSON.stringify(activity)
@@ -241,12 +231,7 @@ export const sendAnnounce = async ({
         await request({
           url: inbox,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              inbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, inbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           method,
@@ -302,12 +287,7 @@ export const deleteStatus = async ({
         await request({
           url: inbox,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              inbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, inbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           method,
@@ -369,12 +349,7 @@ export const undoAnnounce = async ({
           url: inbox,
           method,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              inbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, inbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           body: JSON.stringify(activity)
@@ -432,12 +407,7 @@ export const follow = async (
           url: targetInbox,
           method,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              targetInbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, targetInbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           body: JSON.stringify(activity)
@@ -488,12 +458,7 @@ export const unfollow = async (currentActor: Actor, follow: Follow) =>
         const { statusCode } = await request({
           url: targetInbox,
           headers: {
-            ...signedHeaders(
-              currentActor,
-              method.toLowerCase(),
-              targetInbox,
-              activity
-            ),
+            ...signedHeaders(currentActor, method, targetInbox, activity),
             Accept: DEFAULT_ACCEPT
           },
           method,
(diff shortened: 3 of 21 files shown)


llun:
@codex revew

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -28,7 +30,8 @@ export async function getSenderPublicKey(database: Database, actorId: string) {
         if (nodeError.response.statusCode === 410) {
           const url = new URL(actorId)
           const sender = await getActorPerson({
-            actorId: `${url.protocol}//${url.host}/actor#main-key`
+            actorId: `${url.protocol}//${url.host}/actor#main-key`,
+            signingActor: await getFederationSigningActor(database)
           })
```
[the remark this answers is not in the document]
llun: Addressed in efc26ed by resolving the federation signing actor once before the try/catch and reusing it for the 410 fallback path.

llun:
@codex revew

lib/activities/getActorPosts.ts
```diff
@@ -56,7 +56,10 @@ export const getActorPosts: GetActorPostsFunction = async ({
             })
             if (localStatus) return localStatus
 
-            const note = await getNote({ statusId: item.object })
+            const note = await getNote({
+              statusId: item.object,
+              signingActor
```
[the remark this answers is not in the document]
llun: This is already defined in the getActorPosts params type and destructured in the function signature (`signingActor?: DomainActor` and `async ({ database, person, signingActor }) =>`). TypeScript and `yarn build` both passed on this commit, so there is no runtime ReferenceError here.

llun:
@codex revew

lib/database/sql/actor.ts
```diff
@@ -264,6 +264,19 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
...
+    const persistedActor = await database<SQLActor>('actors')
+      .whereNotNull('accountId')
+      .whereNotNull('privateKey')
+      .where('privateKey', '<>', '')
+      .whereNull('deletionStatus')
+      .orderBy('createdAt', 'asc')
+      .first<{ id: string }>('id')
+    if (!persistedActor) return null
+
+    return this.getActorFromId({ id: persistedActor.id })
```
llun, line 285: **Signer-selection design concerns** for `getFederationSigningActor`:

1. **No `domain` filter.** The query only requires `accountId` and `privateKey` to be set. In a multi-tenant or misconfigured deployment that has historical actors from a different `domain`, this can silently pick a signer whose `id` host doesn't match the running instance's host — remote servers will then fail to verify because they can't resolve the `keyId` URL. Consider filtering by the configured local domain.

2. **Stability / unstable signer.** Picking "oldest local actor with a private key" means deleting/rotating that account changes the signer used by every authorized fetch from this instance. There's no `fediverse_actor` / instance actor concept, so the choice is whichever real human user happens to be oldest. That ties federation health to a specific user account.

3. **Privacy.** Every authorized fetch this server makes is signed by — and therefore reveals the URL of — that one specific local user account to every remote server. Mastodon mitigates this with a dedicated headless "instance actor" (e.g. `https://example.com/actor`) used only for signing fetches. Worth either following that pattern or, at minimum, documenting that the chosen actor's identity is exposed.

Items 1 and 2 are fixable in this PR with a `where('domain', config.host)` filter and a reproducible deterministic selection. Item 3 is a follow-up but worth tracking.

---

lib/jobs/fetchRemoteStatusJob.ts
```diff
@@ -99,20 +104,24 @@ export const fetchRemoteStatusJob = createJobHandle(
...
     const { statusId } = z.object({ statusId: z.string() }).parse(message.data)
+    const signingActor = await getFederationSigningActor(database)
 
-    const status = await fetchRemoteStatus(database, statusId)
+    const status = await fetchRemoteStatus(database, statusId, 0, signingActor)
     if (!status) return
 
     // 8. Fetch replies (up to 100)
-    const note = await getNote({ statusId })
+    const note = await getNote({ statusId, signingActor })
```
llun: **Redundant signed fetch.** `fetchRemoteStatus` (line 32) already calls `getNote({ statusId, signingActor })` for the same `statusId`, then this line refetches it just to read `replies`. With unsigned GETs that was already wasteful; now that every call performs an HTTP-signature signing operation and adds a remote-server-side verify step, it's a noticeable extra round-trip per remote status.

Consider returning the (sanitized) `Note` from `fetchRemoteStatus` alongside the `Status` so this second `getNote` can be dropped.

---

lib/services/federation/getFederationSigningActor.ts
```diff
@@ -0,0 +1,11 @@
+import { Database } from '@/lib/database/types'
+import { Actor } from '@/lib/types/domain/actor'
+
+export const getFederationSigningActor = async (
+  database: Database,
+  preferredActor?: Actor
+): Promise<Actor | undefined> => {
+  if (preferredActor?.privateKey) return preferredActor
+
+  return (await database.getFederationSigningActor()) ?? undefined
```
llun, line 10: **Repeated DB lookups per job/request.** Each invocation of `recordActorIfNeeded`, `fetchRemoteStatusJob`, `createAnnounceJob`, `getSenderPublicKey`, etc. resolves the federation signing actor independently. Inside a single job run with multiple fetches (replies, parent, actor, etc.) this amounts to several identical SQL queries.

Two suggestions:

1. Cache the resolved signing actor for the lifetime of a job invocation (the caller already has a `database` handle and could memoize once at the top of the job).
2. Consider making `recordActorIfNeeded` resolve `signingActor` itself when not provided, so call sites don't have to remember to pass it. Today it's silently optional, and a missed callsite degrades to unsigned fetches without any warning.

---

lib/utils/signature.ts
```diff
@@ -75,6 +75,7 @@ export function signedHeaders(
   const url = new URL(targetUrl)
   const host = url.host
   const date = new Date().toUTCString()
+  const requestTargetPath = `${url.pathname}${url.search}`
```
llun, line 80: **`url.search` includes the leading `?`** — that's correct for the `(request-target)` line. Two adjacent suggestions while you're touching this:

- Add a unit test for an empty query string (`https://target.com/outbox`) so the `${url.pathname}${url.search}` concatenation doesn't accidentally regress for the no-query case (currently only the with-query path is exercised).
- Consider tightening the `method` parameter type from `string` to `'GET' | 'POST'` (or normalizing once at the top of the function), because callers in this PR pass a mix of `'get'`, `'GET'`, and `'post'`. The `(request-target)` value is normalized via `toLowerCase()` here, but the same `method` string is also used as the HTTP method elsewhere (see `lib/activities/index.ts`), and the inconsistency in case is easy to misread.

---

lib/actions/utils.ts
```diff
@@ -57,7 +59,7 @@ export const recordActorIfNeeded = async ({
   const currentTime = Date.now()
   // Update actor if it's older than 3 day
   if (currentTime - existingActor.updatedAt > 3 * 86_400_000) {
-    const person = await getActorPerson({ actorId })
+    const person = await getActorPerson({ actorId, signingActor })
```
llun: **Silent fallback to unsigned fetch.** Both branches of `recordActorIfNeeded` (create at line 43, refresh-after-3-days here at line 62) call `getActorPerson({ actorId, signingActor })`. If the caller forgets to pass `signingActor` — easy to do, since it's optional — `getActorPerson` will issue an *unsigned* GET, which servers configured for authorized fetch will reject with 401/403, leading to the actor never being recorded/refreshed. There's no log or trace event distinguishing this case from a genuinely unreachable actor.

At minimum, log a warn-level event when `signingActor` is undefined here. Better: resolve via `getFederationSigningActor(database)` inside this function so it's not a per-callsite responsibility.

---

lib/activities/index.ts
```diff
@@ -53,7 +55,12 @@ export const getNote = async ({
       try {
         const { statusCode, body } = await request({
           url: statusId,
-          headers: { Accept: DEFAULT_ACCEPT }
+          headers: {
+            Accept: DEFAULT_ACCEPT,
+            ...(signingActor
+              ? signedHeaders(signingActor, 'get', statusId)
+              : {})
```
llun, line 62: When `signingActor` is undefined, `getNote` falls back to an unsigned GET. For servers running with mandatory authorized fetch (Mastodon's `AUTHORIZED_FETCH=true`, etc.) this returns 401/403 and `getNote` logs nothing distinguishing — it just returns `null`. Recommend either (a) resolving a signing actor via `getFederationSigningActor` here when not provided, mirroring `fetchRemoteStatusJob`, or (b) at least emitting a span attribute / warn-log so the failure mode is observable.

Same pattern applies to `getActorPerson`, `getActorCollections`, and the inline `client.fetch` in `fetchRemoteStatusJob`.

---

lib/database/sql/actor.test.ts
```diff
@@ -110,6 +110,19 @@ describe('ActorDatabase', () => {
...
+    describe('#getFederationSigningActor', () => {
+      it('returns a local actor with a private key', async () => {
+        const actor = await database.getFederationSigningActor()
+
+        expect(actor).toMatchObject({
+          domain: TEST_DOMAIN,
+          privateKey: expect.toBeString(),
+          publicKey: expect.toBeString()
+        })
+        expect(actor?.account).toBeDefined()
```
llun, line 159: Test coverage for `getFederationSigningActor` is thin: only the happy path is asserted. Worth adding cases that drive the design choices in the implementation:

- Returns `null` when there are no actors with a private key (e.g. only remote actors in the table).
- Skips actors with `deletionStatus` set.
- Skips actors with `privateKey = ''` (the `<>` filter).
- Deterministically returns the same actor across calls when multiple eligible actors exist (locks in the `orderBy('createdAt', 'asc')` contract).

---

lib/jobs/fetchRemoteStatusJob.ts
```diff
@@ -99,20 +104,24 @@ export const fetchRemoteStatusJob = createJobHandle(
...
 
     const client = {
       fetch: async (url: string) => {
         if (!(await canFederateWithDomain(database, url))) return null
         const { body, statusCode } = await request({
           url,
-          headers: { Accept: 'application/activity+json' }
+          headers: {
+            Accept: 'application/activity+json',
+            ...(signingActor ? signedHeaders(signingActor, 'get', url) : {})
```
llun: Minor: the inline `client.fetch` here re-implements the "Accept + optional signedHeaders" header construction that `getNote`, `getActorPerson`, and `getActorCollections` all also do (variations of the same three-line spread). With this PR there are now four copies of essentially the same pattern. Worth extracting a small `signedRequest({ url, signingActor, accept })` helper so future changes (e.g. adding `User-Agent` overrides, switching algorithms) live in one place.

---

llun (commented):
Review of the signed-authorized-fetch work. Most of the comments are inline; one cross-cutting concern that doesn't fit on any line in this diff:

### Sign / verify asymmetry on `(request-target)`

`signedHeaders` now includes `url.search` in the request target, which is correct. The verifier in `lib/services/guards/ActivityPubVerifyGuard.ts:33` still constructs the target from `requestUrl.pathname` only:

```ts
`${request.method.toLowerCase()} ${requestUrl.pathname}`
```

That file isn't in this PR's diff, so I couldn't comment inline. Today the only consumer of that guard is the inbox (POST, no query string), so this is latent — but the natural counterpart to "support signed authorized fetches" is verifying *incoming* signed GETs to our own outbox / replies endpoints, which routinely carry `?page=true&min_id=0` style query strings. The first endpoint that turns the guard on for GET will fail verification silently. Suggest changing it to `${requestUrl.pathname}${requestUrl.search}` while you're already touching this signature work, so the two sides stay in lockstep.

### Summary of inline comments

- **`getFederationSigningActor` design** (no domain filter, unstable signer choice, leaks a real user's identity to every remote server) — see comment on `lib/database/sql/actor.ts`.
- **Silent fallback to unsigned fetches** when callers forget the `signingActor` param — see comments on `lib/actions/utils.ts`, `lib/activities/index.ts`. Either resolve internally or log when missing.
- **Repeated DB lookups for the signing actor** within a single job/request — see comment on `lib/services/federation/getFederationSigningActor.ts`.
- **Redundant signed `getNote` call** in `fetchRemoteStatusJob` (one in `fetchRemoteStatus`, another immediately after) — see comment on `lib/jobs/fetchRemoteStatusJob.ts:113`.
- **Header-construction duplication** — four copies of the "Accept + optional signedHeaders" spread; worth extracting a helper.
- **`signedHeaders` polish** — add a unit test for empty query strings, and tighten the `method` parameter to a typed enum to remove `'get'` vs `'GET'` vs `'post'` inconsistency across call sites.
- **Test coverage for `getFederationSigningActor`** is happy-path only — see comment on `actor.test.ts`.

No blocking issues; the verify-side query-string fix is the most worth doing in this PR.

---

lib/activities/getActorPosts.ts
```diff
@@ -56,7 +56,10 @@ export const getActorPosts: GetActorPostsFunction = async ({
...
+            })
```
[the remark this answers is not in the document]
llun: This is already defined in the `getActorPosts` params type and destructured in the function signature. `yarn build` passes on the current head, so there is no runtime `ReferenceError` here. No code change needed.

lib/utils/signature.ts
```diff
@@ -99,7 +100,7 @@ export function signedHeaders(
   const signedString = headerKeys
     .map((key) => {
       if (key === '(request-target)') {
-        return `(request-target): ${method} ${url.pathname}`
+        return `(request-target): ${method.toLowerCase()} ${requestTargetPath}`
```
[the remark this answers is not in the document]
llun: `new URL(targetUrl).pathname` always starts with `/` for the full URLs this helper accepts, and the current implementation includes `url.search` for the query string. Fixed adjacent coverage in 10ea5388 by adding no-query request-target coverage and tightening method typing; no string-concat change needed for this suggestion.

lib/database/sql/actor.ts
llun: Fixed in 10ea5388: `getFederationSigningActor` now filters to the configured local domain and uses deterministic `createdAt`, then `id` ordering. The dedicated instance-actor/privacy model is a larger follow-up outside this patch.

lib/jobs/fetchRemoteStatusJob.ts
llun: Fixed in 10ea5388: `fetchRemoteStatus` now returns the sanitized `Note` alongside the stored status, and the job reuses that note before falling back to a separate `getNote` only for existing local statuses.

lib/services/federation/getFederationSigningActor.ts
llun: Fixed in 10ea5388: the job keeps resolving the signing actor once for its fetch path, and `recordActorIfNeeded` now resolves a fallback signing actor internally when omitted and logs when none is available. A broader request-scoped cache/helper can be a follow-up.

lib/utils/signature.ts
llun: Fixed in 10ea5388: `signedHeaders` now accepts only `GET | POST`, normalizes the request-target method once, call sites pass uppercase methods, and a no-query GET request-target test was added.

lib/actions/utils.ts
llun: Fixed in 10ea5388: `recordActorIfNeeded` now resolves the federation signer itself when the caller omits `signingActor`, and emits a warn log if no signer is available before the fetch proceeds unsigned.

lib/activities/index.ts
llun: Partially addressed in 10ea5388: the critical `recordActorIfNeeded` path now resolves/logs internally, and the changed call sites continue to pass a signer. I left the low-level `getNote`/`getActorPerson`/collection helpers accepting optional signers because they do not own a database handle; a shared signed-fetch helper is better handled as a follow-up.

lib/database/sql/actor.test.ts
llun: Fixed in 10ea5388: added coverage for no usable private key, empty private key, wrong-domain actors, scheduled deletion, and deterministic selection when multiple eligible actors exist.

lib/jobs/fetchRemoteStatusJob.ts
llun: Not changed in this patch. This is useful cleanup, but it is not needed for correctness and is lower priority than the authorized-fetch behavior and tests addressed here. A small shared signed-fetch/header helper can be extracted in a follow-up.

llun:
Addressed the non-inline review summary in 10ea5388: incoming signature verification now includes `requestUrl.search` in `(request-target)`, matching the signed outbound GET behavior. Inline comments were replied to and resolved individually.

llun:
@codex review

lib/actions/utils.ts
```diff
@@ -37,8 +41,24 @@ export const recordActorIfNeeded = async ({
   if (existingActor?.privateKey) {
     return existingActor
   }
+
+  const getResolvedSigningActor = async () => {
+    const resolvedSigningActor =
+      signingActor ?? (await getFederationSigningActor(database))
```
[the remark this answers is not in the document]
llun: Fixed in 2df8e17b: `recordActorIfNeeded` now uses `getFederationSigningActor(database, signingActor)`, so the helper handles preferred-actor private-key validation and fallback selection consistently.

llun:
@codex review

llun (commented):
Re-reviewed against `2df8e17`. Every actionable item from the previous round was addressed; details below.

## Addressed

- **Sign / verify asymmetry on `(request-target)`** — `ActivityPubVerifyGuard.ts` now builds `${requestUrl.pathname}${requestUrl.search}` and a unit test in `ActivityPubVerifyGuard.test.ts` locks the behaviour in. ✅
- **`getFederationSigningActor` design** — adds a `domain = getConfiguredActorDomain()` filter, plus a deterministic `orderBy('id', 'asc')` tiebreaker on top of `createdAt`. The new `actor.test.ts` cases cover null-when-empty, `deletionStatus` skipping, and stable selection across calls. ✅
- **Silent fallback to unsigned fetches** — `recordActorIfNeeded` now resolves the signing actor internally via `getFederationSigningActor(database, signingActor)` and emits a `logger.warn` when none is available, so missed callsites are observable. ✅
- **Redundant signed `getNote`** — `fetchRemoteStatus` returns `{ status, note }` and `fetchRemoteStatusJob` reuses `result.note ?? (await getNote(...))`. Confirmed `normalizeActivityPubContent` preserves `replies` (it spreads `...data`), so the reused note is safe for the replies-collection extraction. ✅
- **`signedHeaders` polish** — `method` is now typed `'GET' | 'POST'`, lowercasing happens once via `requestTargetMethod`, and the test suite covers both with-query and empty-query GETs. All call sites in `lib/activities/index.ts` were updated to drop the redundant `.toLowerCase()`. ✅
- **Test coverage for `getFederationSigningActor`** — happy path plus null/deleted/deterministic paths added. ✅

## Minor follow-ups (non-blocking)

- The `logger.warn` in `recordActorIfNeeded` will fire on every fetch on a fresh install with no local users yet. Consider downgrading to `debug` or rate-limiting if this turns out to be noisy in practice.
- The "Accept + optional `signedHeaders`" header construction is still inlined in four spots (`getNote`, `getActorPerson`, `getActorCollections`, the inline `client.fetch` in `fetchRemoteStatusJob`). Worth extracting a small helper in a follow-up so future header changes (e.g. `User-Agent`, algorithm switch) live in one place.
- Privacy/instance-actor pattern (signing fetches with a dedicated headless instance actor à la Mastodon, instead of leaking a real user's `id` to every remote server) — out of scope for this PR but worth tracking.

LGTM on the scoped changes.

llun:
Addressing the latest review body from Apr 28:

- Keeping the `logger.warn` for the missing federation signer path for now. It was added intentionally so unsigned fallback fetches are visible; the latest note frames downgrade/rate-limiting as conditional if it becomes noisy, so I am leaving that as a follow-up rather than weakening observability in this PR.
- The shared signed-fetch/header helper is valid cleanup, but non-blocking and broader than the current signed authorized fetch fix. Existing behavior is covered and green, so I am keeping that as follow-up work.
- The dedicated headless instance actor/privacy model is agreed as the better long-term shape, but it requires a broader design/schema/route change and is out of scope for this patch.

No code change needed for these non-blocking follow-ups. All review threads are resolved, and the latest Gemini review on `2df8e17` reported no feedback.

PR #606, opened 2026-04-27
feat: implement robust ActivityPub content negotiation

llun:
## Summary
- Add robust Accept header parsing with q-value sorting, parameters, wildcards, and invalid-entry tolerance.
- Add ActivityPub negotiation/response helpers for activity+json, ActivityStreams JSON-LD, generic JSON, and wildcard requests.
- Apply the helper to actor, status, followers, following, and outbox ActivityPub GET routes while preserving HTML redirects when HTML is preferred.
- Add route and utility coverage for weighted combined headers, JSON-LD profile requests, generic JSON, wildcard requests, collection responses, and HTML fallback.

## Validation
- yarn run prettier --write .
- yarn lint (passes with existing warnings in auth/fitness files)
- yarn build
- yarn test --runTestsByPath lib/utils/acceptContentTypes.test.ts lib/utils/activityPubContentNegotiation.test.ts app/api/users/[username]/route.test.ts app/api/users/[username]/statuses/[statusId]/route.test.ts app/api/users/[username]/followers/route.test.ts app/api/users/[username]/following/route.test.ts app/api/users/[username]/outbox/route.test.ts
- yarn test --runInBand (215 suites, 1645 tests passed)

Note: the normal parallel yarn test command hit unrelated Jest worker SIGSEGV crashes in different pre-existing suites on two local runs; rerunning the named files and then the full suite serially passed.

### DIFF
--- a/app/api/users/[username]/followers/route.ts
+++ b/app/api/users/[username]/followers/route.ts
@@ -1,20 +1,24 @@
 import { OnlyLocalUserGuard } from '@/lib/services/guards/OnlyLocalUserGuard'
+import { activityPubResponse } from '@/lib/utils/activityPubContentNegotiation'
 import { ACTIVITY_STREAM_URL } from '@/lib/utils/activitystream'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 
 export const GET = traceApiRoute(
   'getActorFollowers',
-  OnlyLocalUserGuard(async (database, actor) => {
+  OnlyLocalUserGuard(async (database, actor, req) => {
     const followerId = `${actor.id}/followers`
 
     const totalItems = await database.getActorFollowersCount({
       actorId: actor.id
     })
-    return Response.json({
-      '@context': ACTIVITY_STREAM_URL,
-      id: followerId,
-      type: 'OrderedCollection',
-      totalItems
+    return activityPubResponse({
+      req,
+      data: {
+        '@context': ACTIVITY_STREAM_URL,
+        id: followerId,
+        type: 'OrderedCollection',
+        totalItems
+      }
     })
   })
 )
--- a/app/api/users/[username]/route.ts
+++ b/app/api/users/[username]/route.ts
@@ -1,5 +1,6 @@
 import { OnlyLocalUserGuard } from '@/lib/services/guards/OnlyLocalUserGuard'
 import {
+  activityPubRedirectResponse,
   activityPubResponse,
   negotiateActivityPubContentType
 } from '@/lib/utils/activityPubContentNegotiation'
@@ -20,6 +21,8 @@ export const GET = traceApiRoute(
       })
     }
 
-    return Response.redirect(`https://${actor.domain}/@${actor.username}`)
+    return activityPubRedirectResponse(
+      `https://${actor.domain}/@${actor.username}`
+    )
   })
 )
--- a/app/api/users/[username]/route.ts
+++ b/app/api/users/[username]/route.ts
@@ -1,20 +1,22 @@
 import { OnlyLocalUserGuard } from '@/lib/services/guards/OnlyLocalUserGuard'
+import {
+  activityPubResponse,
+  negotiateActivityPubContentType
+} from '@/lib/utils/activityPubContentNegotiation'
 import { getPersonFromActor } from '@/lib/utils/getPersonFromActor'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 
 export const GET = traceApiRoute(
   'getActor',
   OnlyLocalUserGuard(async (_, actor, req) => {
-    const acceptHeader = req.headers.get('accept')
-    if (acceptHeader?.startsWith('application/ld+json')) {
-      return Response.json(getPersonFromActor(actor), {
-        headers: { 'content-type': 'application/ld+json' }
-      })
-    }
-
-    if (acceptHeader?.startsWith('application/activity+json')) {
-      return Response.json(getPersonFromActor(actor), {
-        headers: { 'content-type': 'application/activity+json' }
+    const contentType = negotiateActivityPubContentType(
+      req.headers.get('accept')
+    )
+    if (contentType) {
+      return activityPubResponse({
+        req,
+        data: getPersonFromActor(actor),
+        contentType
       })
     }
 
--- a/app/api/users/[username]/statuses/[statusId]/route.ts
+++ b/app/api/users/[username]/statuses/[statusId]/route.ts
@@ -4,6 +4,10 @@ import {
 } from '@/lib/services/guards/OnlyLocalUserGuard'
 import { AppRouterParams } from '@/lib/services/guards/types'
 import { toActivityPubObject } from '@/lib/types/domain/status'
+import {
+  activityPubResponse,
+  negotiateActivityPubContentType
+} from '@/lib/utils/activityPubContentNegotiation'
 import { ACTIVITY_STREAM_URL } from '@/lib/utils/activitystream'
 import { apiErrorResponse } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
@@ -23,19 +27,15 @@ export const GET = traceApiRoute(
     const note = toActivityPubObject(status)
     if (!note) return apiErrorResponse(404)
 
-    const acceptHeader = req.headers.get('accept')
-    if (acceptHeader?.startsWith('application/ld+json')) {
-      return Response.json(
-        { '@context': ACTIVITY_STREAM_URL, ...note },
-        { headers: { 'content-type': 'application/ld+json' } }
-      )
-    }
-
-    if (acceptHeader?.startsWith('application/activity+json')) {
-      return Response.json(
-        { '@context': ACTIVITY_STREAM_URL, ...note },
-        { headers: { 'content-type': 'activity+json' } }
-      )
+    const contentType = negotiateActivityPubContentType(
+      req.headers.get('accept')
+    )
+    if (contentType) {
+      return activityPubResponse({
+        req,
+        data: { '@context': ACTIVITY_STREAM_URL, ...note },
+        contentType
+      })
     }
 
     return Response.redirect(
--- a/app/api/users/[username]/statuses/[statusId]/route.ts
+++ b/app/api/users/[username]/statuses/[statusId]/route.ts
@@ -5,6 +5,7 @@ import {
 import { AppRouterParams } from '@/lib/services/guards/types'
 import { toActivityPubObject } from '@/lib/types/domain/status'
 import {
+  activityPubRedirectResponse,
   activityPubResponse,
   negotiateActivityPubContentType
 } from '@/lib/utils/activityPubContentNegotiation'
@@ -38,7 +39,7 @@ export const GET = traceApiRoute(
       })
     }
 
-    return Response.redirect(
+    return activityPubRedirectResponse(
       `https://${status.actor?.domain}/@${actor.username}/${statusId}`
     )
   })
--- a/lib/utils/acceptContentTypes.ts
+++ b/lib/utils/acceptContentTypes.ts
@@ -84,10 +84,8 @@ export const parseAcceptContentTypes = (
           const value = parameter.slice(equalsIndex + 1)
           if (!name) return params
 
-          return {
-            ...params,
-            [name]: unquote(value)
-          }
+          params[name] = unquote(value)
+          return params
         },
         {}
       )
--- a/lib/utils/acceptContentTypes.ts
+++ b/lib/utils/acceptContentTypes.ts
@@ -44,7 +44,25 @@ const splitHeaderValue = (value: string, delimiter: ',' | ';') => {
 const unquote = (value: string) => {
   const trimmed = value.trim()
   if (trimmed.startsWith('"') && trimmed.endsWith('"')) {
-    return trimmed.slice(1, -1).replace(/\\"/g, '"')
+    let unquoted = ''
+    let escaped = false
+
+    for (const character of trimmed.slice(1, -1)) {
+      if (escaped) {
+        unquoted += character
+        escaped = false
+        continue
+      }
+
+      if (character === '\\') {
+        escaped = true
+        continue
+      }
+
+      unquoted += character
+    }
+
+    return escaped ? `${unquoted}\\` : unquoted
   }
 
   return trimmed
--- a/lib/utils/acceptContentTypes.ts
+++ b/lib/utils/acceptContentTypes.ts
@@ -1,8 +1,106 @@
-import contentType from 'content-type'
+export type AcceptedContentType = {
+  type: string
+  parameters: Record<string, string>
+  quality: number
+  specificity: number
+  index: number
+}
+
+const splitHeaderValue = (value: string, delimiter: ',' | ';') => {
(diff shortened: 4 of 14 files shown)


llun:
@codex review

lib/utils/acceptContentTypes.ts
```diff
@@ -1,8 +1,106 @@
...
+      inQuotes = !inQuotes
+    }
+
+    if (character === delimiter && !inQuotes) {
+      items.push(current)
+      current = ''
+    } else {
+      current += character
+    }
+  }
```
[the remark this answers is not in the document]
llun: Fixed by tracking escape state while splitting quoted header values. I also added regression coverage for an escaped quote followed by a comma inside a quoted parameter.

lib/utils/activityPubContentNegotiation.ts
```diff
@@ -0,0 +1,124 @@
...
+    negotiateActivityPubContentType(req.headers.get('accept')) ??
+    ACTIVITYPUB_CONTENT_TYPE
+
+  return apiResponse({
+    req,
+    allowedMethods: [HttpMethod.enum.GET],
+    data,
+    additionalHeaders: [['Content-Type', responseContentType]]
+  })
+}
```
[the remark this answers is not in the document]
llun: Fixed. `activityPubResponse` now accepts optional `allowedMethods` with `GET` as the default, and there is test coverage for overriding the CORS method list.

lib/utils/activityPubContentNegotiation.ts
```diff
@@ -0,0 +1,124 @@
...
+export const ACTIVITYSTREAM_LD_CONTENT_TYPE = `application/ld+json; profile="${ACTIVITY_STREAM_URL}"`
+export const JSON_CONTENT_TYPE = 'application/json'
+
+type ActivityPubCandidate = {
+  accepted: AcceptedContentType
+  responseContentType: string
+}
+
+const isActivityStreamsProfile = (profile: string | undefined) => {
+  if (!profile) return true
```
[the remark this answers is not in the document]
llun: Fixed. Omitted `profile` still negotiates as JSON-LD, but blank/whitespace-only profile values now return false. Added coverage for `profile=""`.

lib/utils/activityPubContentNegotiation.ts
```diff
@@ -0,0 +1,124 @@
...
+  const responseContentType =
+    contentType ??
+    negotiateActivityPubContentType(req.headers.get('accept')) ??
+    ACTIVITYPUB_CONTENT_TYPE
+
+  return apiResponse({
+    req,
+    allowedMethods: [HttpMethod.enum.GET],
+    data,
+    additionalHeaders: [['Content-Type', responseContentType]]
```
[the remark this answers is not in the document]
llun: Fixed. `activityPubResponse` now emits `Vary: Accept`, with test coverage.

lib/utils/acceptContentTypes.ts
```diff
@@ -1,8 +1,106 @@
...
+
+  items.push(current)
+  return items
+}
```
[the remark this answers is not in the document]
llun: Already addressed in the current branch: `splitHeaderValue` tracks escape state while inside quoted strings, and the parser has a regression test for an escaped quote followed by a comma inside a quoted parameter.

lib/utils/acceptContentTypes.ts
```diff
@@ -1,8 +1,106 @@
...
+          if (equalsIndex === -1) return params
+
+          const name = parameter.slice(0, equalsIndex).trim().toLowerCase()
+          const value = parameter.slice(equalsIndex + 1)
+          if (!name) return params
+
+          return {
+            ...params,
+            [name]: unquote(value)
+          }
```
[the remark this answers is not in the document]
llun: Fixed by mutating the local reduce accumulator directly instead of spreading a new object on every parameter.

lib/utils/activityPubContentNegotiation.ts
```diff
@@ -0,0 +1,124 @@
...
+  responseContentType: string
+}
+
+const isActivityStreamsProfile = (profile: string | undefined) => {
+  if (!profile) return true
+
+  return profile
+    .split(/\s+/)
+    .map((item) => item.trim())
+    .includes(ACTIVITY_STREAM_URL)
```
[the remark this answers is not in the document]
llun: Fixed. The profile string is now trimmed before splitting into whitespace-separated tokens.

app/api/users/[username]/route.ts
```diff
@@ -1,20 +1,22 @@
...
+    const contentType = negotiateActivityPubContentType(
+      req.headers.get('accept')
+    )
+    if (contentType) {
+      return activityPubResponse({
+        req,
+        data: getPersonFromActor(actor),
+        contentType
       })
     }
```
Comment (llun) [inline app/api/users/[username]/route.ts:22]: The negotiated JSON path now correctly emits `Vary: Accept` (via `activityPubResponse`), but the HTML fallback right below — `return Response.redirect(...)` on line 23 — does not. That asymmetry breaks the property `Vary` is meant to protect: a shared cache or CDN can store the 302 from a browser request and then serve that same redirect to a subsequent ActivityPub crawler with `Accept: application/activity+json` (or vice versa).

Please add `Vary: Accept` to the redirect response too — e.g. build it via `new Response(null, { status: 302, headers: { Location: ..., Vary: 'Accept' } })` (or extract a small helper). The same fix is needed in `app/api/users/[username]/statuses/[statusId]/route.ts:41`.

---

app/api/users/[username]/statuses/[statusId]/route.ts
```diff
@@ -23,19 +27,15 @@ export const GET = traceApiRoute(
...
+    const contentType = negotiateActivityPubContentType(
+      req.headers.get('accept')
+    )
+    if (contentType) {
+      return activityPubResponse({
+        req,
+        data: { '@context': ACTIVITY_STREAM_URL, ...note },
+        contentType
+      })
     }
```
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/route.ts:40]: Same `Vary: Accept` gap as in the actor route: the negotiated branch above sets `Vary: Accept`, but the `Response.redirect(...)` HTML fallback on line 41-43 does not. A shared cache that sees the redirect for a browser request can replay it to an ActivityPub fetch (or vice versa). Please make the redirect response also include `Vary: Accept`.

---

lib/utils/acceptContentTypes.ts
```diff
@@ -1,8 +1,117 @@
...
+}
+
+const unquote = (value: string) => {
+  const trimmed = value.trim()
+  if (trimmed.startsWith('"') && trimmed.endsWith('"')) {
+    return trimmed.slice(1, -1).replace(/\\"/g, '"')
+  }
+
+  return trimmed
+}
```
llun, line 69: `unquote` only collapses `\"` back to `"`, but RFC 7230's `quoted-pair` rule says any `\X` inside a quoted-string represents `X` (so `\\` should become `\`, `\t` should become `t`, etc.). Combined with `splitHeaderValue`, which already preserves the literal backslash byte for every escaped character, parameters such as `name="a\\b"` come out as `a\\b` instead of `a\b`.

In practice ActivityPub clients almost never send those values, but since this parser is now the project's source of truth for Accept parsing, consider replacing the targeted `replace(/\\"/g, '"')` with a generic unescape (e.g. iterate the slice and drop the backslash from every `\X` pair) so the function is RFC-correct.

---

lib/utils/activityPubContentNegotiation.ts
```diff
@@ -0,0 +1,127 @@
...
+  accepted: AcceptedContentType
+  responseContentType: string
+}
+
+const isActivityStreamsProfile = (profile: string | undefined) => {
+  if (profile === undefined) return true
+  if (!profile.trim()) return false
+
+  return profile.trim().split(/\s+/).includes(ACTIVITY_STREAM_URL)
+}
```
llun, line 27: Minor: `profile.trim()` is computed twice on the happy path. Trivially refactor to `const tokens = profile.trim().split(/\s+/); if (!tokens[0]) return false; return tokens.includes(ACTIVITY_STREAM_URL)`, or extract `const trimmed = profile.trim()` once and reuse it. Not blocking.

---

app/api/users/[username]/followers/route.ts
```diff
@@ -1,20 +1,24 @@
...
-      totalItems
+    return activityPubResponse({
+      req,
+      data: {
+        '@context': ACTIVITY_STREAM_URL,
+        id: followerId,
+        type: 'OrderedCollection',
+        totalItems
+      }
     })
```
Comment (llun) [inline app/api/users/[username]/followers/route.ts:22]: A consistency question (not necessarily a bug): the actor and status routes return a 302 redirect when the client prefers HTML, but `followers`, `following`, and `outbox` always fall through to the `activityPubResponse` default (`ACTIVITYPUB_CONTENT_TYPE`) for HTML-only `Accept` headers. That matches pre-PR behavior — these endpoints have no HTML view — but it does mean a browser hitting `/api/users/x/followers` directly gets `application/activity+json` with `Vary: Accept` instead of a redirect. If that's intentional, fine; otherwise consider redirecting to a profile page (or returning 406) for `Accept: text/html` here too. Same comment applies to `following/route.ts` and `outbox/route.ts`.

---

lib/utils/activityPubContentNegotiation.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      }) as never,
+      data: { ok: true },
+      allowedMethods: ['GET', 'POST']
+    })
+
+    expect(response.headers.get('access-control-allow-methods')).toBe(
+      'GET,POST'
+    )
+  })
+})
```
llun, line 99: Two coverage gaps worth filling now that order-based tie-breaking is part of the contract:

1. The "browser-style" Accept (`'text/html, application/xhtml+xml, */*;q=0.8'`) returns null — covered. But the symmetric case where AP and HTML share the same q-value (e.g. `'application/activity+json, text/html'` and the reverse `'text/html, application/activity+json'`) isn't tested. The current code uses original `index` as the final tie-breaker, so the order in the header matters; lock that in with a test so future refactors don't silently flip behavior.
2. `application/ld+json; profile="https://www.w3.org/ns/activitystreams https://example.com/other"` (multi-value profile token) — the helper handles it via the whitespace-split path; a test would document the intent.

---

llun (commented):
Reviewed the latest commits on this branch (`feat: implement robust ActivityPub content negotiation` + the two follow-up fixups). The negotiation helper and parser look solid — earlier reviewer feedback (escape handling, mutable accumulator, `Vary: Accept` on negotiated responses, optional `allowedMethods`, blank-profile rejection, whitespace-tokenized profile match) has all been addressed and there is matching test coverage.

Things I'd like to see fixed before merging:

1. **`Vary: Accept` is missing on the HTML redirect responses** in `app/api/users/[username]/route.ts:23` and `app/api/users/[username]/statuses/[statusId]/route.ts:41`. The negotiated JSON branches now correctly emit `Vary: Accept`, but the redirect branch does not — so a shared cache that stores a redirect for an HTML request can replay it to an ActivityPub fetcher (or vice versa), which is exactly the scenario `Vary` exists to prevent. This is the only real correctness issue I found.

2. **`unquote` only handles `\"`**, not the full RFC 7230 quoted-pair (`\\` → `\`, `\t` → `t`, …). Low practical impact for ActivityPub, but worth tightening since this parser is now the project's source of truth for Accept-header parsing.

3. **Minor**: `isActivityStreamsProfile` calls `profile.trim()` twice on the happy path — easy cleanup.

4. **Coverage**: would like tests pinning down (a) order-based tie-breaking when AP and HTML share `q=1`, and (b) multi-token `profile` parameter on `application/ld+json`.

5. **Question, not a bug**: the `followers`/`following`/`outbox` routes never redirect to HTML for HTML-only Accept headers — they fall through to the default `ACTIVITYPUB_CONTENT_TYPE`. That preserves pre-PR behavior, but is asymmetric with the actor/status routes; flag if that's intentional.

Items 2–5 are non-blocking; item 1 is the one I'd want resolved before this lands.

---

app/api/users/[username]/route.ts
Comment (llun) [reply app/api/users/[username]/route.ts:22]: Fixed. The actor HTML fallback now uses `activityPubRedirectResponse`, which builds the 302 with `Location` and `Vary: Accept`; the route test now asserts the header.

app/api/users/[username]/statuses/[statusId]/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/route.ts:40]: Fixed. The status HTML fallback now uses the same redirect helper and the redirect test asserts `Vary: Accept`.

lib/utils/acceptContentTypes.ts
llun: Fixed. `unquote` now drops the backslash from any quoted-pair, not only escaped quotes, and preserves malformed trailing backslashes. Added coverage for `\\`, `\t`, and `\"` style pairs.

lib/utils/activityPubContentNegotiation.ts
llun: Fixed. `isActivityStreamsProfile` now trims once and reuses the trimmed value before splitting.

app/api/users/[username]/followers/route.ts
Comment (llun) [reply app/api/users/[username]/followers/route.ts:22]: Intentional. Followers/following/outbox do not have HTML views, and this preserves the pre-PR behavior for those collection endpoints. They still emit `Vary: Accept` on the ActivityPub response.

lib/utils/activityPubContentNegotiation.test.ts
llun: Fixed. Added tests for same-q AP/HTML tie-breaking in both header orders and for a multi-token ActivityStreams JSON-LD profile.

llun (commented):
Final review on commit `97259301f7492d84b40526960254cfe5c1e02dd6` — all prior feedback has been addressed:

1. **`Vary: Accept` on HTML redirects** — Resolved. New `activityPubRedirectResponse` helper builds a 302 with both `Location` and `Vary: Accept`; the actor and status routes both use it; route tests assert `vary === 'Accept'` on the 302.
2. **`unquote` quoted-pair handling** — Resolved. The function now does a generic character-by-character unescape (any `\X` → `X`), with a trailing-backslash safeguard. Test `'unescapes generic quoted-pair parameter values'` covers `\\`, `\t`, `\"` together.
3. **Double `profile.trim()`** — Resolved via local `trimmed` variable.
4. **Tie-breaking coverage** — New test asserts both `'application/activity+json, text/html'` → AP and `'text/html, application/activity+json'` → null, locking in the index-based tie-breaker.
5. **Multi-token profile coverage** — New test for `profile="...activitystreams ...other"`.

Verified by tracing the new `unquote` against the test input `String.raw` `a\\b\t\"quoted` — produces `a\bt"quoted` as expected, and the dangling-backslash branch correctly preserves a final lone `\`.

One non-blocking nit if you want a follow-up: `activityPubRedirectResponse` constructs the `Response` directly, so it skips the `Server` header that `apiResponse` injects — purely a cosmetic inconsistency with the JSON path. Not worth holding the PR for.

LGTM from my side. Nothing else to flag.

---

llun:
Addressed the non-blocking redirect-header consistency nit from the latest review. `activityPubRedirectResponse` now emits the same `Server` header as `apiResponse`, and the actor/status redirect tests assert it alongside `Vary: Accept`.

PR #611, opened 2026-04-28
feat: share ActivityPub signed request headers

llun:
## Summary
- Add a shared ActivityPub request-header helper for default Accept headers plus optional HTTP Signature headers.
- Use the helper for signed GET fetches and signed POST ActivityPub sends.
- Add helper tests covering unsigned defaults, signed GET query-string targets, and signed POST digest headers.

## Validation
- yarn run prettier --write .
- yarn lint (passes with existing warnings in auth and fitness parsing files)
- yarn build
- yarn test

## Notes
- No migrations or config changes required.

### DIFF
--- a/lib/activities/activityPubHeaders.test.ts
+++ b/lib/activities/activityPubHeaders.test.ts
@@ -1,0 +1,80 @@
+import { MOCK_SECRET_PHASES } from '@/lib/stub/actor'
+import { Actor } from '@/lib/types/domain/actor'
+import { generateKeyPair, verify } from '@/lib/utils/signature'
+
+import { activityPubRequestHeaders } from './activityPubHeaders'
+
+describe('#activityPubRequestHeaders', () => {
+  let signingActor: Actor
+
+  beforeAll(async () => {
+    const keyPair = await generateKeyPair(MOCK_SECRET_PHASES)
+    signingActor = {
+      id: 'https://local.test/users/signer',
+      privateKey: keyPair.privateKey,
+      publicKey: keyPair.publicKey
+    } as Actor
+  }, 15000)
+
+  it('returns default ActivityPub Accept headers without a signing actor', () => {
+    const headers = activityPubRequestHeaders({
+      url: 'https://remote.test/users/alice'
+    })
+
+    expect(headers).toEqual({
+      Accept: 'application/activity+json, application/ld+json'
+    })
+  })
+
+  it('adds signed GET headers when a signing actor is provided', async () => {
+    const headers = activityPubRequestHeaders({
+      url: 'https://remote.test/users/alice/outbox?page=true',
+      signingActor
+    })
+
+    expect(headers).toMatchObject({
+      Accept: 'application/activity+json, application/ld+json',
+      host: 'remote.test',
+      signature: expect.stringContaining('headers="(request-target) host date"')
+    })
+
+    const verified = await verify(
+      'get /users/alice/outbox?page=true',
+      {
+        ...headers,
+        signature: headers.signature
+      },
+      signingActor.publicKey
+    )
+    expect(verified).toBeTruthy()
+  }, 15000)
+
+  it('adds digest headers for signed POST content', async () => {
+    const content = { type: 'Follow', object: 'https://remote.test/users/bob' }
+    const headers = activityPubRequestHeaders({
+      url: 'https://remote.test/inbox',
+      method: 'POST',
+      signingActor,
+      content
+    })
+
+    expect(headers).toMatchObject({
+      Accept: 'application/activity+json, application/ld+json',
+      'content-type': 'application/activity+json',
+      digest: expect.stringMatching(/^SHA-256=/),
+      signature: expect.stringContaining(
+        'headers="(request-target) host date digest content-type"'
+      )
+    })
+
+    const verified = await verify(
+      'post /inbox',
+      {
+        ...headers,
+        signature: headers.signature
+      },
+      signingActor.publicKey
+    )
+    expect(verified).toBeTruthy()
+  }, 15000)
+})
--- a/lib/activities/activityPubHeaders.test.ts
+++ b/lib/activities/activityPubHeaders.test.ts
@@ -22,7 +22,18 @@ describe('#activityPubRequestHeaders', () => {
     })
 
     expect(headers).toEqual({
-      Accept: 'application/activity+json, application/ld+json'
+      accept: 'application/activity+json, application/ld+json'
+    })
+  })
+
+  it('uses a custom ActivityPub Accept header when provided', () => {
+    const headers = activityPubRequestHeaders({
+      url: 'https://remote.test/users/alice',
+      accept: 'application/activity+json'
+    })
+
+    expect(headers).toEqual({
+      accept: 'application/activity+json'
     })
   })
 
@@ -33,7 +44,7 @@ describe('#activityPubRequestHeaders', () => {
     })
 
     expect(headers).toMatchObject({
-      Accept: 'application/activity+json, application/ld+json',
+      accept: 'application/activity+json, application/ld+json',
       host: 'remote.test',
       signature: expect.stringContaining('headers="(request-target) host date"')
     })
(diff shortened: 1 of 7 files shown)


llun:
@codex revew

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
+import { DEFAULT_ACCEPT } from '@/lib/activities/constants'
+import { Actor } from '@/lib/types/domain/actor'
+import { signedHeaders } from '@/lib/utils/signature'
+
+type ActivityPubMethod = 'GET' | 'POST'
```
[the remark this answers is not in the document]
llun: Addressed in bd873bf2 by deriving ActivityPubMethod from Parameters<typeof signedHeaders>[1], so the helper reuses the signedHeaders method contract instead of duplicating the union.

llun:
@codex revew

lib/activities/activityPubHeaders.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      host: 'remote.test',
+      signature: expect.stringContaining('headers="(request-target) host date"')
+    })
+
+    const verified = await verify(
+      'get /users/alice/outbox?page=true',
+      {
+        ...headers,
+        signature: headers.signature
+      },
```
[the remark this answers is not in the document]
llun: Addressed in 2037f73e by passing headers directly to verify for the signed GET test.

llun:
@codex revew

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
+import { DEFAULT_ACCEPT } from '@/lib/activities/constants'
+import { Actor } from '@/lib/types/domain/actor'
+import { signedHeaders } from '@/lib/utils/signature'
+
+type ActivityPubMethod = Parameters<typeof signedHeaders>[1]
```
llun: The `ActivityPubMethod` type is derived indirectly from `signedHeaders`'s parameter position. If that function's signature ever changes, this type silently changes too with no error here. Consider defining it explicitly:

```ts
type ActivityPubMethod = 'GET' | 'POST'
```

This is clearer about intent and decouples the helper's type from the signature utility's internal parameter shape.

---

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface ActivityPubRequestHeadersParams {
+  url: string
+  method?: ActivityPubMethod
+  signingActor?: Actor | null
+  content?: unknown
```
llun: `content?: unknown` loses all type safety at call sites — callers can pass anything and TypeScript won't complain. Every caller in this PR passes an ActivityPub activity object. Consider narrowing this to `Record<string, unknown> | undefined` or whatever the existing `Activity` / `ActivityPub` union type is in the codebase, so misuse is caught at compile time.

---

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const activityPubRequestHeaders = ({
+  url,
+  method = 'GET',
+  signingActor,
+  content,
+  accept = DEFAULT_ACCEPT
+}: ActivityPubRequestHeadersParams): Record<string, string> => ({
+  ...(signingActor ? signedHeaders(signingActor, method, url, content) : {}),
```
llun, line 22: The `Accept` key uses title-case while `signedHeaders` returns all-lowercase keys (`host`, `date`, `digest`, `content-type`, `signature`). HTTP headers are case-insensitive so this works in practice, but since this is the new canonical helper it's a good opportunity to standardize on one casing — lowercase is more idiomatic for HTTP/2 and `node-fetch`-style libraries. Consider changing to `accept` to be consistent with the rest of the headers returned by `signedHeaders`.

---

lib/activities/activityPubHeaders.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    })
+
+    const verified = await verify(
+      'post /inbox',
+      headers,
+      signingActor.publicKey
+    )
+    expect(verified).toBeTruthy()
+  }, 15000)
+})
```
llun, line 85: The test suite is missing a case for the `accept` override parameter. `fetchRemoteStatusJob.ts` passes `accept: 'application/activity+json'` (without `application/ld+json`) in production, but there is no test verifying this override is respected. Please add a test like:

```ts
it('uses a custom accept header when provided', () => {
  const headers = activityPubRequestHeaders({
    url: 'https://remote.test/users/alice',
    accept: 'application/activity+json'
  })

  expect(headers).toEqual({
    Accept: 'application/activity+json'
  })
})
```

---

llun (commented):
## Review: feat: share ActivityPub signed request headers

The refactoring is clean and the new `activityPubRequestHeaders` helper successfully eliminates a lot of duplication. The test coverage for the core signing logic is solid.

**Four items to fix before merging:**

1. **`ActivityPubMethod` type** — derived indirectly from `signedHeaders`'s parameter position; prefer an explicit `'GET' | 'POST'` union.
2. **`content?: unknown`** — too loose; every caller passes an ActivityPub object so the type should be narrowed to prevent silent misuse.
3. **`Accept` header casing** — the helper uses title-case `Accept` while `signedHeaders` returns all-lowercase keys (`host`, `date`, `content-type`, etc.). Since this is the new canonical helper, standardize on lowercase `accept` for consistency.
4. **Missing test for `accept` override** — `fetchRemoteStatusJob.ts` passes `accept: 'application/activity+json'` in production but there is no test verifying the override works correctly. This is the most important item to add.

---

llun:
@codex review

lib/activities/activityPubHeaders.ts
llun: Addressed in eae2f295 by replacing the parameter-position-derived type with the exported explicit SignedHttpMethod union from lib/utils/signature.ts.

lib/activities/activityPubHeaders.ts
llun: Addressed in eae2f295 by narrowing content to ActivityPubRequestContent = object, which rejects primitive payloads while keeping the existing typed ActivityPub activity objects accepted.

lib/activities/activityPubHeaders.ts
llun: Addressed in eae2f295 by emitting lowercase accept from the helper and updating the header expectations in tests.

lib/activities/activityPubHeaders.test.ts
llun: Addressed in eae2f295 by adding coverage for a custom accept override matching the fetchRemoteStatusJob usage.

lib/activities/activityPubHeaders.ts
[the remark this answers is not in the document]
llun: Addressed in eae2f295 by exporting SignedHttpMethod from lib/utils/signature.ts and using it in activityPubRequestHeaders.

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
...
+  Accept: accept
+})
```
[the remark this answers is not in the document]
llun: Partially addressed in eae2f295 by standardizing the helper on lowercase accept. I did not apply the suggested method.toLowerCase() call because signedHeaders accepts the uppercase SignedHttpMethod contract and already lowercases internally for (request-target); URL pathname and query-string handling also already lives in signedHeaders.

llun:
@codex review

lib/activities/activityPubHeaders.ts
```diff
@@ -0,0 +1,24 @@
+import { DEFAULT_ACCEPT } from '@/lib/activities/constants'
+import { Actor } from '@/lib/types/domain/actor'
+import { type SignedHttpMethod, signedHeaders } from '@/lib/utils/signature'
+
+type ActivityPubRequestContent = object
```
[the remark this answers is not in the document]
llun: Not changing this one: activityPubRequestHeaders is not parsing inbound request bodies, and changing this back to unknown would restore the loose call-site behavior that the previous review correctly asked us to avoid. The object constraint is intentional here because callers pass already-constructed ActivityPub payload objects, and it rejects primitive content values while preserving the existing typed ActivityPub interfaces.

PR #621, opened 2026-04-29
fix: Resolve boosted status routes and remote actor rendering

llun:
## Summary

This PR fixes boosted status detail routes and timeline rendering where a boost row could be treated as if the booster were the original post owner.

The route lookup now lives in `resolveStatusFromPath`, keeps the scoped hash lookup first, and falls back to validated unscoped lookup only when the resolved status owner matches the actor in the path. Announce rows are unwrapped to their original status for owner checks and public detail links.

It also improves remote and opaque actor handling:
- fetches public remote statuses when a local row is missing
- normalizes ActivityPub actor/content helpers in shared utilities
- validates remote Note/Announce payloads before mapping to domain statuses
- avoids displaying UUID/DID actor IDs as usernames
- derives human handles from Mastodon-style and anchored Bluesky/Bridgy URLs
- avoids linking opaque actor fallbacks to raw ActivityPub JSON actor URLs
- keeps malformed remote data observable through logging instead of silent drops

## Root Cause

Boosted status links are generated from the actual/original status, but the persisted row for a boost belongs to the actor that boosted it. The old detail route scoped hash lookup to the actor parsed from the URL, so original-owner boost links could miss the stored booster row and return 404.

Timeline rendering had the same ownership confusion in remote outbox mapping: the original status actor could be overwritten with the booster actor, producing empty boost text and the wrong displayed owner.

## Production Investigation

The originally reported hash `c137613b93abd541592aa6ad69e94e7d315962c38a4cd324d63b87cf649f8ad6` was not found in the production statuses table by `urlHash`, status `id`, or `url`, so that exact URL still depends on the underlying production row/data existing.

Known production-data examples for `@cheeaun@mastodon.social`, Bridgy Bluesky, Hackers.pub, and Hollo-style remote IDs were used to verify the corrected route/link shapes locally without mutating production data.

## Notes

`getProfileData` now uses the shared `getActorImageUrl` helper. This does not change image-clearing behavior in the current SQL adapter: `database.updateActor` only writes icon/header settings when the value is truthy, so both the previous empty string and the new `undefined` leave any persisted value unchanged.

## Validation

- `yarn run prettier --write`
- `yarn test --runTestsByPath lib/utils/activitypubActor.test.ts lib/components/posts/post.test.tsx lib/activities/getActorPosts.test.ts`
- `git diff --check`
- `yarn lint` (0 errors; existing 12 warnings remain)
- `yarn build`
- Earlier branch validation also included the resolver/path tests and full `yarn test`

### DIFF
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -20,7 +20,7 @@ import { getActorFromSession } from '@/lib/utils/getActorFromSession'
 import { Header } from './Header'
 import { RemoteStatusLoading } from './RemoteStatusLoading'
 import { StatusBox } from './StatusBox'
-import { resolveStatusFromPath } from './resolveStatusFromPath'
+import { decodePathParam, resolveStatusFromPath } from './resolveStatusFromPath'
 
 interface Props {
   params: Promise<{ actor: string; status: string }>
@@ -31,7 +31,7 @@ export const generateMetadata = async ({
 }: Props): Promise<Metadata> => {
   const { actor } = await params
   return {
-    title: `Activities.next: ${decodeURIComponent(actor)} status`
+    title: `Activities.next: ${decodePathParam(actor)} status`
   }
 }
 
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -2,6 +2,7 @@ import { Metadata } from 'next'
 import { notFound } from 'next/navigation'
 import { FC } from 'react'
 
+import { getRemoteStatus } from '@/lib/activities/getRemoteStatus'
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { FETCH_REMOTE_STATUS_JOB_NAME } from '@/lib/jobs/names'
@@ -56,7 +57,16 @@ const Page: FC<Props> = async ({ params }) => {
   })
   if (!resolvedStatus) return notFound()
 
-  const { fullStatusId, isStatusHash, status, statusId } = resolvedStatus
+  const { fullStatusId, isStatusHash } = resolvedStatus
+  let { status, statusId } = resolvedStatus
+
+  if (!status && !isStatusHash && fullStatusId) {
+    status = await getRemoteStatus({
+      statusId: fullStatusId,
+      signingActor: currentActor ?? undefined
+    })
+    statusId = status?.id ?? ''
+  }
 
   // Try to fetch remote status if not found and user is logged in
   if (!status && session && !isStatusHash) {
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -20,6 +20,7 @@ import { getActorFromSession } from '@/lib/utils/getActorFromSession'
 import { Header } from './Header'
 import { RemoteStatusLoading } from './RemoteStatusLoading'
 import { StatusBox } from './StatusBox'
+import { resolveStatusFromPath } from './resolveStatusFromPath'
 
 interface Props {
   params: Promise<{ actor: string; status: string }>
@@ -48,60 +49,14 @@ const Page: FC<Props> = async ({ params }) => {
 
   const { actor, status: statusParam } = await params
   const currentTime = Date.now()
-  const decodedActor = decodeURIComponent(actor)
-  const decodedStatusParam = (() => {
-    try {
-      return decodeURIComponent(statusParam)
-    } catch {
-      return statusParam
-    }
-  })()
-
-  const parts = decodedActor.split('@').slice(1)
-  if (parts.length !== 2) {
-    return notFound()
-  }
-
-  const actorFromPath = await database.getActorFromUsername({
-    username: parts[0],
-    domain: parts[1]
+  const resolvedStatus = await resolveStatusFromPath({
+    database,
+    actorParam: actor,
+    statusParam
   })
-  const actorIdFromPath = actorFromPath?.id
-  const isStatusHash = /^[a-f0-9]{64}$/i.test(decodedStatusParam)
+  if (!resolvedStatus) return notFound()
 
-  const protocol = parts[1].startsWith('localhost') ? 'http' : 'https'
-  const isFullStatusUrl = /^https?:\/\//.test(decodedStatusParam)
-  const fullStatusId = isFullStatusUrl
-    ? decodedStatusParam
-    : `${protocol}://${parts[1]}/users/${parts[0]}/statuses/${decodedStatusParam}`
-
-  let status: Status | null = null
-  let statusId = ''
-
-  if (isStatusHash) {
-    status = await database.getStatusFromUrlHash({
-      urlHash: decodedStatusParam,
-      actorId: actorIdFromPath
-    })
-    statusId = status?.id ?? ''
-  }
-
-  // Try full URL format first (ActivityPub standard), then fallback to raw id (for legacy/mock data)
-  if (!status) {
-    status = await database.getStatus({
-      statusId: fullStatusId,
-      withReplies: false
-    })
-    statusId = fullStatusId
-  }
-
-  if (!status && !isFullStatusUrl) {
-    status = await database.getStatus({
-      statusId: decodedStatusParam,
-      withReplies: false
-    })
-    statusId = decodedStatusParam
-  }
+  const { fullStatusId, isStatusHash, status, statusId } = resolvedStatus
 
   // Try to fetch remote status if not found and user is logged in
   if (!status && session && !isStatusHash) {
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
@@ -6,6 +6,7 @@ import { resolveStatusFromPath } from './resolveStatusFromPath'
 describe('#resolveStatusFromPath', () => {
   const originalActorId = 'https://remote.example/users/original'
   const boosterActorId = 'https://boost.example/users/booster'
+  const secondBoosterActorId = 'https://other.example/users/booster'
   const originalUrl = 'https://remote.example/@original/123'
   const statusHash = getHashFromString(originalUrl)
 
@@ -34,34 +35,33 @@ describe('#resolveStatusFromPath', () => {
     tags: []
   } as Status
 
-  const boostedStatus = {
-    id: 'https://boost.example/users/booster/statuses/456/activity',
-    actorId: boosterActorId,
-    actor: {
-      username: 'booster',
-      domain: 'boost.example'
-    },
-    type: StatusType.enum.Announce,
-    to: [],
-    cc: [],
-    edits: [],
-    originalStatus,
-    isLocalActor: false,
-    createdAt: Date.now(),
-    updatedAt: Date.now()
-  } as Status
+  const createAnnounce = (actorId = boosterActorId) =>
+    ({
+      id: `${actorId}/statuses/456/activity`,
+      actorId,
+      actor: {
+        username: 'booster',
+        domain: new URL(actorId).host
+      },
+      type: StatusType.enum.Announce,
+      to: [],
+      cc: [],
+      edits: [],
+      originalStatus,
+      isLocalActor: false,
+      createdAt: Date.now(),
+      updatedAt: Date.now()
+    }) as Status
 
-  it('resolves a hash route when the stored row actor is the booster and the path actor owns the original status', async () => {
-    const database = {
-      getActorFromUsername: jest
-        .fn()
-        .mockResolvedValue({ id: originalActorId }),
-      getStatusFromUrlHash: jest
-        .fn()
-        .mockResolvedValueOnce(null)
-        .mockResolvedValueOnce(boostedStatus),
-      getStatus: jest.fn().mockResolvedValue(null)
-    }
+  const createDatabase = () => ({
+    getActorFromUsername: jest.fn().mockResolvedValue({ id: originalActorId }),
+    getStatusFromUrlHash: jest.fn().mockResolvedValue(null),
+    getStatus: jest.fn().mockResolvedValue(null)
+  })
+
+  it('resolves a hash route from the scoped actor lookup', async () => {
+    const database = createDatabase()
+    database.getStatusFromUrlHash.mockResolvedValueOnce(originalStatus)
 
     await expect(
       resolveStatusFromPath({
@@ -70,11 +70,38 @@ describe('#resolveStatusFromPath', () => {
         statusParam: statusHash
       })
     ).resolves.toEqual({
-      fullStatusId:
-        'https://remote.example/users/original/statuses/' + statusHash,
+      fullStatusId: '',
       isStatusHash: true,
-      status: boostedStatus,
-      statusId: boostedStatus.id
+      status: originalStatus,
+      statusId: originalStatus.id
+    })
+
+    expect(database.getStatusFromUrlHash).toHaveBeenCalledTimes(1)
+    expect(database.getStatusFromUrlHash).toHaveBeenCalledWith({
+      urlHash: statusHash,
+      actorId: originalActorId
+    })
+    expect(database.getStatus).not.toHaveBeenCalled()
+  })
+
+  it('resolves a boosted hash route to the original status when the path actor owns it', async () => {
+    const database = createDatabase()
+    const boostedStatus = createAnnounce()
+    database.getStatusFromUrlHash
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce(boostedStatus)
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: statusHash
+      })
+    ).resolves.toEqual({
+      fullStatusId: '',
+      isStatusHash: true,
+      status: originalStatus,
+      statusId: originalStatus.id
     })
 
     expect(database.getStatusFromUrlHash).toHaveBeenNthCalledWith(1, {
@@ -84,22 +111,15 @@ describe('#resolveStatusFromPath', () => {
     expect(database.getStatusFromUrlHash).toHaveBeenNthCalledWith(2, {
       urlHash: statusHash
     })
+    expect(database.getStatus).not.toHaveBeenCalled()
   })
 
-  it('does not resolve an unscoped hash when neither the status nor original status belongs to the path actor', async () => {
-    const database = {
-      getActorFromUsername: jest
-        .fn()
-        .mockResolvedValue({ id: originalActorId }),
-      getStatusFromUrlHash: jest
-        .fn()
-        .mockResolvedValueOnce(null)
-        .mockResolvedValueOnce({
-          ...originalStatus,
-          actorId: 'https://other.example/users/someone'
-        }),
-      getStatus: jest.fn().mockResolvedValue(null)
-    }
+  it('returns the original status when an unscoped hash lookup finds another boost of the same status', async () => {
+    const database = createDatabase()
+    const secondBoostedStatus = createAnnounce(secondBoosterActorId)
+    database.getStatusFromUrlHash
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce(secondBoostedStatus)
 
     await expect(
       resolveStatusFromPath({
@@ -108,11 +128,146 @@ describe('#resolveStatusFromPath', () => {
         statusParam: statusHash
       })
     ).resolves.toEqual({
-      fullStatusId:
-        'https://remote.example/users/original/statuses/' + statusHash,
+      fullStatusId: '',
+      isStatusHash: true,
+      status: originalStatus,
+      statusId: originalStatus.id
+    })
+  })
+
+  it('does not resolve an unscoped hash when a note belongs to a different actor', async () => {
+    const database = createDatabase()
+    database.getStatusFromUrlHash
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce({
+        ...originalStatus,
+        actorId: 'https://other.example/users/someone'
+      })
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: statusHash
+      })
+    ).resolves.toEqual({
+      fullStatusId: '',
       isStatusHash: true,
       status: null,
       statusId: ''
+    })
+  })
+
+  it('does not resolve an unscoped hash when an announce original belongs to a different actor', async () => {
+    const database = createDatabase()
+    database.getStatusFromUrlHash
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce({
+        ...createAnnounce(),
+        originalStatus: {
+          ...originalStatus,
+          actorId: 'https://other.example/users/someone'
+        }
+      })
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: statusHash
+      })
+    ).resolves.toEqual({
+      fullStatusId: '',
+      isStatusHash: true,
+      status: null,
+      statusId: ''
+    })
+  })
+
+  it('returns null when the actor route is malformed', async () => {
+    const database = createDatabase()
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: 'plain-username',
+        statusParam: statusHash
+      })
+    ).resolves.toBeNull()
+
+    expect(database.getActorFromUsername).not.toHaveBeenCalled()
+    expect(database.getStatusFromUrlHash).not.toHaveBeenCalled()
+    expect(database.getStatus).not.toHaveBeenCalled()
+  })
+
+  it('returns null when the actor route has malformed URI escapes', async () => {
+    const database = createDatabase()
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '%E0%A4%A',
+        statusParam: statusHash
+      })
+    ).resolves.toBeNull()
+
+    expect(database.getActorFromUsername).not.toHaveBeenCalled()
+    expect(database.getStatusFromUrlHash).not.toHaveBeenCalled()
+    expect(database.getStatus).not.toHaveBeenCalled()
+  })
+
+  it('falls back from full status id to raw id for non-hash status params', async () => {
+    const database = createDatabase()
+    database.getStatus
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce(originalStatus)
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: '123'
+      })
+    ).resolves.toEqual({
+      fullStatusId: 'https://remote.example/users/original/statuses/123',
+      isStatusHash: false,
+      status: originalStatus,
+      statusId: originalStatus.id
+    })
+
+    expect(database.getStatusFromUrlHash).not.toHaveBeenCalled()
+    expect(database.getStatus).toHaveBeenNthCalledWith(1, {
+      statusId: 'https://remote.example/users/original/statuses/123',
+      withReplies: false
+    })
+    expect(database.getStatus).toHaveBeenNthCalledWith(2, {
+      statusId: '123',
+      withReplies: false
+    })
+  })
+
+  it('resolves full URL status params without raw id fallback', async () => {
+    const database = createDatabase()
+    database.getStatus.mockResolvedValueOnce(originalStatus)
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: originalStatus.id
+      })
+    ).resolves.toEqual({
+      fullStatusId: originalStatus.id,
+      isStatusHash: false,
+      status: originalStatus,
+      statusId: originalStatus.id
+    })
+
+    expect(database.getStatusFromUrlHash).not.toHaveBeenCalled()
+    expect(database.getStatus).toHaveBeenCalledTimes(1)
+    expect(database.getStatus).toHaveBeenCalledWith({
+      statusId: originalStatus.id,
+      withReplies: false
     })
   })
 })
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
@@ -1,0 +1,118 @@
+import { Status, StatusType } from '@/lib/types/domain/status'
+import { getHashFromString } from '@/lib/utils/getHashFromString'
+
+import { resolveStatusFromPath } from './resolveStatusFromPath'
+
+describe('#resolveStatusFromPath', () => {
+  const originalActorId = 'https://remote.example/users/original'
+  const boosterActorId = 'https://boost.example/users/booster'
+  const originalUrl = 'https://remote.example/@original/123'
+  const statusHash = getHashFromString(originalUrl)
+
+  const originalStatus = {
+    id: 'https://remote.example/users/original/statuses/123',
+    url: originalUrl,
+    actorId: originalActorId,
+    actor: {
+      username: 'original',
+      domain: 'remote.example'
+    },
+    type: StatusType.enum.Note,
+    to: [],
+    cc: [],
+    edits: [],
+    isLocalActor: false,
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+    text: 'Original status',
+    reply: '',
+    replies: [],
+    actorAnnounceStatusId: null,
+    isActorLiked: false,
+    totalLikes: 0,
+    attachments: [],
+    tags: []
+  } as Status
+
+  const boostedStatus = {
+    id: 'https://boost.example/users/booster/statuses/456/activity',
+    actorId: boosterActorId,
+    actor: {
+      username: 'booster',
+      domain: 'boost.example'
+    },
+    type: StatusType.enum.Announce,
+    to: [],
+    cc: [],
+    edits: [],
+    originalStatus,
+    isLocalActor: false,
+    createdAt: Date.now(),
+    updatedAt: Date.now()
+  } as Status
+
+  it('resolves a hash route when the stored row actor is the booster and the path actor owns the original status', async () => {
+    const database = {
+      getActorFromUsername: jest
+        .fn()
+        .mockResolvedValue({ id: originalActorId }),
+      getStatusFromUrlHash: jest
+        .fn()
+        .mockResolvedValueOnce(null)
+        .mockResolvedValueOnce(boostedStatus),
+      getStatus: jest.fn().mockResolvedValue(null)
+    }
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: statusHash
+      })
+    ).resolves.toEqual({
+      fullStatusId:
+        'https://remote.example/users/original/statuses/' + statusHash,
+      isStatusHash: true,
+      status: boostedStatus,
+      statusId: boostedStatus.id
+    })
+
+    expect(database.getStatusFromUrlHash).toHaveBeenNthCalledWith(1, {
+      urlHash: statusHash,
+      actorId: originalActorId
+    })
+    expect(database.getStatusFromUrlHash).toHaveBeenNthCalledWith(2, {
+      urlHash: statusHash
+    })
+  })
+
+  it('does not resolve an unscoped hash when neither the status nor original status belongs to the path actor', async () => {
+    const database = {
+      getActorFromUsername: jest
+        .fn()
+        .mockResolvedValue({ id: originalActorId }),
+      getStatusFromUrlHash: jest
+        .fn()
+        .mockResolvedValueOnce(null)
+        .mockResolvedValueOnce({
+          ...originalStatus,
+          actorId: 'https://other.example/users/someone'
+        }),
+      getStatus: jest.fn().mockResolvedValue(null)
+    }
+
+    await expect(
+      resolveStatusFromPath({
+        database,
+        actorParam: '@original@remote.example',
+        statusParam: statusHash
+      })
+    ).resolves.toEqual({
+      fullStatusId:
+        'https://remote.example/users/original/statuses/' + statusHash,
+      isStatusHash: true,
+      status: null,
+      statusId: ''
+    })
+  })
+})
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
@@ -1,0 +1,107 @@
+import { Database } from '@/lib/database/types'
+import { Status, StatusType } from '@/lib/types/domain/status'
+
+interface ResolveStatusFromPathParams {
+  database: Pick<
+    Database,
+    'getActorFromUsername' | 'getStatus' | 'getStatusFromUrlHash'
+  >
+  actorParam: string
+  statusParam: string
+}
+
+interface ResolveStatusFromPathResult {
+  status: Status | null
+  statusId: string
+  fullStatusId: string
+  isStatusHash: boolean
+}
+
+const decodeParam = (param: string) => {
+  try {
+    return decodeURIComponent(param)
+  } catch {
+    return param
+  }
+}
+
+const getStatusOwnerActorId = (status: Status) =>
+  status.type === StatusType.enum.Announce
+    ? status.originalStatus.actorId
+    : status.actorId
+
+export const resolveStatusFromPath = async ({
+  database,
+  actorParam,
+  statusParam
+}: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
+  const decodedActor = decodeURIComponent(actorParam)
+  const decodedStatusParam = decodeParam(statusParam)
+
+  const parts = decodedActor.split('@').slice(1)
+  if (parts.length !== 2) {
+    return null
+  }
+
+  const [username, domain] = parts
+  const actorFromPath = await database.getActorFromUsername({
+    username,
+    domain
+  })
+  const actorIdFromPath = actorFromPath?.id
+  const isStatusHash = /^[a-f0-9]{64}$/i.test(decodedStatusParam)
+
+  const protocol = domain.startsWith('localhost') ? 'http' : 'https'
+  const isFullStatusUrl = /^https?:\/\//.test(decodedStatusParam)
+  const fullStatusId = isFullStatusUrl
+    ? decodedStatusParam
+    : `${protocol}://${domain}/users/${username}/statuses/${decodedStatusParam}`
+
+  let status: Status | null = null
+  let statusId = ''
+
+  if (isStatusHash) {
+    status = await database.getStatusFromUrlHash({
+      urlHash: decodedStatusParam,
+      actorId: actorIdFromPath
+    })
+
+    if (!status && actorIdFromPath) {
+      const unscopedStatus = await database.getStatusFromUrlHash({
+        urlHash: decodedStatusParam
+      })
+
+      if (
+        unscopedStatus &&
+        getStatusOwnerActorId(unscopedStatus) === actorIdFromPath
+      ) {
+        status = unscopedStatus
+      }
+    }
+
+    statusId = status?.id ?? ''
+  }
+
+  if (!status) {
+    status = await database.getStatus({
+      statusId: fullStatusId,
+      withReplies: false
+    })
+    statusId = status?.id ?? ''
+  }
+
+  if (!status && !isFullStatusUrl) {
+    status = await database.getStatus({
+      statusId: decodedStatusParam,
+      withReplies: false
+    })
+    statusId = status?.id ?? ''
+  }
+
+  return {
+    status,
+    statusId,
+    fullStatusId,
+    isStatusHash
+  }
+}
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
@@ -25,17 +25,27 @@ const decodeParam = (param: string) => {
   }
 }
 
-const getStatusOwnerActorId = (status: Status) =>
-  status.type === StatusType.enum.Announce
-    ? status.originalStatus.actorId
-    : status.actorId
+const getStatusForPathActor = (status: Status, actorId: string) => {
+  if (status.actorId === actorId) return status
 
+  if (
+    status.type === StatusType.enum.Announce &&
+    status.originalStatus.actorId === actorId
+  ) {
+    return status.originalStatus
+  }
+
+  return null
+}
+
+// Returns null only when the actor route cannot be parsed. Lookup misses are
+// returned as { status: null } so callers can still queue remote fetches.
 export const resolveStatusFromPath = async ({
   database,
   actorParam,
   statusParam
 }: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
-  const decodedActor = decodeURIComponent(actorParam)
+  const decodedActor = decodeParam(actorParam)
   const decodedStatusParam = decodeParam(statusParam)
 
   const parts = decodedActor.split('@').slice(1)
@@ -53,12 +63,13 @@ export const resolveStatusFromPath = async ({
 
   const protocol = domain.startsWith('localhost') ? 'http' : 'https'
   const isFullStatusUrl = /^https?:\/\//.test(decodedStatusParam)
-  const fullStatusId = isFullStatusUrl
-    ? decodedStatusParam
-    : `${protocol}://${domain}/users/${username}/statuses/${decodedStatusParam}`
+  const fullStatusId = isStatusHash
+    ? ''
+    : isFullStatusUrl
+      ? decodedStatusParam
+      : `${protocol}://${domain}/users/${username}/statuses/${decodedStatusParam}`
 
   let status: Status | null = null
-  let statusId = ''
 
   if (isStatusHash) {
     status = await database.getStatusFromUrlHash({
@@ -71,36 +82,29 @@ export const resolveStatusFromPath = async ({
         urlHash: decodedStatusParam
       })
 
-      if (
-        unscopedStatus &&
-        getStatusOwnerActorId(unscopedStatus) === actorIdFromPath
-      ) {
-        status = unscopedStatus
+      if (unscopedStatus) {
+        status = getStatusForPathActor(unscopedStatus, actorIdFromPath)
       }
     }
-
-    statusId = status?.id ?? ''
   }
 
-  if (!status) {
+  if (!status && !isStatusHash) {
     status = await database.getStatus({
       statusId: fullStatusId,
       withReplies: false
     })
-    statusId = status?.id ?? ''
   }
 
-  if (!status && !isFullStatusUrl) {
+  if (!status && !isStatusHash && !isFullStatusUrl) {
     status = await database.getStatus({
       statusId: decodedStatusParam,
       withReplies: false
     })
-    statusId = status?.id ?? ''
   }
 
   return {
     status,
-    statusId,
+    statusId: status?.id ?? '',
     fullStatusId,
     isStatusHash
   }
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
@@ -17,7 +17,7 @@ interface ResolveStatusFromPathResult {
   isStatusHash: boolean
 }
 
-const decodeParam = (param: string) => {
+export const decodePathParam = (param: string) => {
   try {
     return decodeURIComponent(param)
   } catch {
@@ -45,8 +45,8 @@ export const resolveStatusFromPath = async ({
   actorParam,
   statusParam
 }: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
-  const decodedActor = decodeParam(actorParam)
-  const decodedStatusParam = decodeParam(statusParam)
+  const decodedActor = decodePathParam(actorParam)
+  const decodedStatusParam = decodePathParam(statusParam)
 
   const parts = decodedActor.split('@').slice(1)
   if (parts.length !== 2) {
--- a/app/(timeline)/[actor]/getProfileData.ts
+++ b/app/(timeline)/[actor]/getProfileData.ts
@@ -8,6 +8,7 @@ import { Actor } from '@/lib/types/activitypub'
 import { Actor as DomainActor } from '@/lib/types/domain/actor'
 import { Attachment } from '@/lib/types/domain/attachment'
 import { Status } from '@/lib/types/domain/status'
+import { getActorImageUrl } from '@/lib/utils/activitypubActor'
 import { getPersonFromActor } from '@/lib/utils/getPersonFromActor'
 
 type ProfileData = {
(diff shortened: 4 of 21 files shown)


app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+  status.type === StatusType.enum.Announce
+    ? status.originalStatus.actorId
+    : status.actorId
+
+export const resolveStatusFromPath = async ({
+  database,
+  actorParam,
+  statusParam
+}: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
+  const decodedActor = decodeURIComponent(actorParam)
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Inconsistent decode handling: `statusParam` is decoded through `decodeParam` (try/catch fallback) but `actorParam` is decoded with bare `decodeURIComponent`. A malformed percent-sequence in `actorParam` will throw `URIError` and surface as a 500 instead of the expected `notFound()`. Either route both through `decodeParam`, or remove the helper and accept that a malformed param 500s consistently — but the asymmetry as written is a latent bug.

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+        urlHash: decodedStatusParam
+      })
+
+      if (
+        unscopedStatus &&
+        getStatusOwnerActorId(unscopedStatus) === actorIdFromPath
+      ) {
+        status = unscopedStatus
+      }
+    }
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:88]: Correctness risk with multiple boost rows sharing the same `urlHash`. `urlHash` is derived from the original status URL, so every actor that boosts the same status produces a row with the same `urlHash` but a different `actorId`. The unscoped `getStatusFromUrlHash({ urlHash })` then returns whichever row `.first()` happens to pick. The owner-equality check `getStatusOwnerActorId(unscopedStatus) === actorIdFromPath` passes for *all* of those boost rows (because each has the same `originalStatus.actorId`), so this can return an arbitrary booster's row instead of the path actor's row.

Two consequences:
1. The page renders a status whose stored `actorId` (the booster) is not the path actor, leaving the URL/owner mismatch you set out to fix.
2. The downstream visibility check at `page.tsx:116` (`currentActor.id !== status.actorId`) operates on the wrong `actorId` for non-public statuses.

Consider explicitly preferring the original (Note) row over Announce rows when the path actor is the original author — e.g. query by `(urlHash, actorId=originalAuthorId, type=Note)` first, and only fall back to an Announce when no Note row exists.

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+    })
+    statusId = status?.id ?? ''
+  }
+
+  if (!status && !isFullStatusUrl) {
+    status = await database.getStatus({
+      statusId: decodedStatusParam,
+      withReplies: false
+    })
+    statusId = status?.id ?? ''
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Seconding the gemini-code-assist note: `statusId = status?.id ?? ''` is repeated in three branches and the value is only consumed in the return object. Compute it once at the bottom (`statusId: status?.id ?? ''`) and drop the local. Same applies to the dedicated `let statusId = ''` at line 61.

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+    domain
+  })
+  const actorIdFromPath = actorFromPath?.id
+  const isStatusHash = /^[a-f0-9]{64}$/i.test(decodedStatusParam)
+
+  const protocol = domain.startsWith('localhost') ? 'http' : 'https'
+  const isFullStatusUrl = /^https?:\/\//.test(decodedStatusParam)
+  const fullStatusId = isFullStatusUrl
+    ? decodedStatusParam
+    : `${protocol}://${domain}/users/${username}/statuses/${decodedStatusParam}`
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: When `isStatusHash` is true, `fullStatusId` is fabricated as `https://${domain}/users/${username}/statuses/${hash}` — a URL that mixes the path actor's user/domain with what is really an opaque hash, and that does not correspond to any real ActivityPub object. It is then both used as a DB lookup key on line 86 and returned to the caller. The fallback `getStatus({ statusId: fullStatusId })` at line 86 will essentially always miss for hash routes, so it's a wasted DB round-trip. Consider gating the full-URL fallback on `!isStatusHash` (and only computing/returning `fullStatusId` for the non-hash path).

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+    return decodeURIComponent(param)
+  } catch {
+    return param
+  }
+}
+
+const getStatusOwnerActorId = (status: Status) =>
+  status.type === StatusType.enum.Announce
+    ? status.originalStatus.actorId
+    : status.actorId
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: `getStatusOwnerActorId` assumes `status.originalStatus` is populated whenever `status.type === Announce`. The discriminated union on `Status` should make that safe at the type level, but in practice `getStatusFromUrlHash` returns rows hydrated by `getStatusWithAttachmentsFromData` — worth confirming that path always populates `originalStatus` for Announce rows, otherwise this can throw on a malformed/orphan boost row.

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      })
+    ).resolves.toEqual({
+      fullStatusId:
+        'https://remote.example/users/original/statuses/' + statusHash,
+      isStatusHash: true,
+      status: null,
+      statusId: ''
+    })
+  })
+})
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts:273]: Test coverage gaps. The two cases here only exercise the unscoped-fallback branch. Please add at least:

- The happy-path scoped lookup (first `getStatusFromUrlHash` returns a row) — making sure the function does **not** issue an unscoped follow-up call.
- An invalid `actorParam` (e.g. `'plain-username'` with no `@user@domain` shape) returning `null`.
- A non-hash `statusParam` falling through to `getStatus` with `fullStatusId`, then to the raw-id fallback.
- A full-URL `statusParam` (e.g. `https://remote.example/users/original/statuses/123`) so the `isFullStatusUrl` branch and the gated raw-id fallback are covered.
- A multi-boost scenario for the urlHash collision concern raised on `resolveStatusFromPath.ts` (two rows share the same `urlHash`; the path actor is the original author; ensure the function returns the correct row, not just *some* row whose owner matches).

Also — the second test's name says "neither the status nor original status belongs to the path actor", but the fixture is a Note (not an Announce), so only the `actorId` arm of `getStatusOwnerActorId` is exercised. Either rename the test or add a sibling test using an Announce whose `originalStatus.actorId` also doesn't match.

---

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -48,60 +49,14 @@ const Page: FC<Props> = async ({ params }) => {
...
-  let statusId = ''
-
-  if (isStatusHash) {
-    status = await database.getStatusFromUrlHash({
-      urlHash: decodedStatusParam,
-      actorId: actorIdFromPath
-    })
-    statusId = status?.id ?? ''
-  }
+  if (!resolvedStatus) return notFound()
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:58]: Behavior change worth double-checking: `resolveStatusFromPath` returns `null` only for a malformed `actorParam` (no `@user@domain` shape). Previously, that case returned `notFound()` directly; now it does too via `if (!resolvedStatus) return notFound()`, which is fine. But note that an **unknown** actor (well-formed `@user@domain` but no DB row) used to leave `actorIdFromPath = undefined` and continue down the lookup chain — the new code preserves that, but the union of `null`-vs-non-`null` return is now slightly load-bearing. A short comment on `resolveStatusFromPath` clarifying "returns `null` only for malformed paths; missing data is encoded as `status: null`" would help future readers.

---

llun (commented):
Review of the `resolveStatusFromPath` extraction and the boosted-hash fallback.

The refactor is a clean lift, and the scoped-then-unscoped fallback is a reasonable shape for the bug. A few items to address before merging:

**Correctness**
- Multi-boost `urlHash` collision: when several actors have boosted the same status, the unscoped lookup can return any of their rows and the owner-equality check passes for all of them. This can leave the URL/owner mismatch unfixed and feeds the wrong `actorId` into the visibility checks in `page.tsx`. See inline comment for a suggested fix (prefer the original Note row over Announce rows).
- `decodeURIComponent(actorParam)` is not wrapped in try/catch even though `statusParam` is — malformed actor params throw a `URIError` instead of 404.
- `fullStatusId` is fabricated for hash routes and then used as a DB key in the next fallback, producing a guaranteed-miss `getStatus` call.

**Tests**
- The new test file only exercises the unscoped-fallback branch. Please add coverage for the scoped-hit happy path, the invalid-actor path, the full-URL/raw-id branches, and a multi-boost collision case. Details inline.

**Style / minor**
- Repeated `statusId = status?.id ?? ''` assignments — collapse into the return (also raised by gemini-code-assist).
- A short comment on the `null` vs `{ status: null }` contract of `resolveStatusFromPath` would help readers.
- Confirm `originalStatus` is always populated for Announce rows returned by `getStatusFromUrlHash`, since `getStatusOwnerActorId` will throw otherwise.

Also note the PR description: the production hash `c137...8ad6` was not found by `urlHash`, `id`, or `url`. That suggests the originally reported 404 may have a separate cause (missing row entirely) that this fix wouldn't address — worth calling out so the original report isn't assumed resolved.

---

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,107 @@
...
+    })
+    statusId = status?.id ?? ''
+  }
+
+  return {
+    status,
+    statusId,
+    fullStatusId,
+    isStatusHash
+  }
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:110]: Fixed in 499026cb. `statusId` is now computed once in the return object, and the intermediate branch assignments were removed.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Fixed in 499026cb. `actorParam` now uses the same guarded decode helper as `statusParam`, and the new malformed-URI test covers this path.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:88]: Fixed in 499026cb. The unscoped hash fallback now returns the original status when the matching row is an `Announce`, so multiple boosts of the same original converge on the original owner row instead of an arbitrary booster row.

Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:88]: Fixed in 499026cb. The unscoped hash fallback now returns the original status when the matching row is an `Announce`, so multiple boosts of the same original converge on the original owner row instead of an arbitrary booster row.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Fixed in 499026cb. The local `statusId` variable and repeated assignments are gone; the resolver derives `statusId` once from the final status.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Fixed in 499026cb. Hash routes now use an empty `fullStatusId` and skip both `getStatus(fullStatusId)` and raw-id fallbacks; non-hash full URL/raw-id behavior is covered by tests.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts]: Confirmed; no code change needed. `getStatusFromUrlHash` hydrates through `getStatusWithAttachmentsFromData`; for `Announce` rows that helper loads `data.content` via `getStatus`, returns `null` when the original cannot be loaded, and then `StatusAnnounce.parse` requires `originalStatus`. So a non-null `Announce` from this path has `originalStatus` populated.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.test.ts:273]: Fixed in 499026cb. Added coverage for scoped hash lookup, boosted hash fallback, multi-boost collision, mismatched Note and Announce ownership, malformed actor params including bad URI escapes, raw-id fallback, and full-URL lookup.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:58]: Fixed in 499026cb. Added a short resolver contract comment: malformed actor routes return `null`; lookup misses return `{ status: null }` so callers can still handle non-hash remote fetches.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -0,0 +1,111 @@
...
+// returned as { status: null } so callers can still queue remote fetches.
+export const resolveStatusFromPath = async ({
+  database,
+  actorParam,
+  statusParam
+}: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
+  const decodedActor = decodeParam(actorParam)
+  const decodedStatusParam = decodeParam(statusParam)
+
+  const parts = decodedActor.split('@').slice(1)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:51]: No code change for this one. The canonical actor route in this app is `@user@domain`; links are generated that way via `getMention(..., true)`. Accepting bare `user@domain` or trimming decoded whitespace would widen the route contract and could turn malformed URL segments into valid actors. For this resolver, malformed actor routes intentionally return `null` and become 404.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -48,60 +49,14 @@ const Page: FC<Props> = async ({ params }) => {
...
-  }
-
-  const actorFromPath = await database.getActorFromUsername({
-    username: parts[0],
-    domain: parts[1]
+  const resolvedStatus = await resolveStatusFromPath({
+    database,
+    actorParam: actor,
+    statusParam
   })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:57]: Fixed in 79397d81. `generateMetadata` now uses the same guarded path decoder as `resolveStatusFromPath`, so malformed URI escapes no longer throw before the page can handle the request. Added `page.test.ts` coverage for the malformed actor metadata path.

llun:
@codex review

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,95 @@
...
+  } catch {
+    return actorId
+  }
+}
+
+const getImageUrl = (image: Actor['icon']) => {
+  if (!image) return undefined
+  if (Array.isArray(image)) return image.find((item) => item.url)?.url
+  return image.url
+}
```
llun: `getImageUrl`, `getActorDomain`, and `getActorProfileFromPerson` are now duplicated almost verbatim between `lib/activities/getRemoteStatus.ts` (lines 35–67) and `lib/activities/getActorPosts.ts` (lines 26–72). `getImageUrl` also exists in `lib/actions/utils.ts` (line 22) and `app/(timeline)/[actor]/getProfileData.ts` (line 24) — the latter returns `''` instead of `undefined`, so they will drift the moment one site changes its empty-value contract.

Please extract these into a shared module (e.g. `lib/utils/activitypub.ts` or `lib/types/activitypub/actor.ts`) so all four call sites use the same implementation. Right now any future tweak (e.g. handling `image` arrays differently, or stripping query strings) has to be remembered in four places.

---

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export const getRemoteStatus = async ({
+  statusId,
+  signingActor
+}: GetRemoteStatusParams): Promise<StatusNote | null> => {
+  try {
+    const remoteNote = await getNote({ statusId, signingActor })
+    if (!remoteNote) return null
+
+    const note = normalizeActivityPubContent(remoteNote) as Note
```
llun: `normalizeActivityPubContent(remoteNote) as Note` skips schema validation entirely. `getNote` returns a loosely-typed payload, and a remote server can return a Note with missing/wrong-shaped fields (`attributedTo` as object, `to`/`cc` missing, `published` invalid, etc.). With the bare cast, malformed values flow into `fromNote` and produce a broken `StatusNote` that will fail downstream rendering rather than being rejected here.

Run the value through `Note.parse(...)` (or a `safeParse` that returns `null` on failure) before passing it to `fromNote`. The existing `getNote` flow validates with Zod; this new public-status path should match.

---

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,95 @@
...
+  'as:Public'
+]
+
+const hasPublicAudience = (value: Note['to'] | Note['cc']) => {
+  const items = Array.isArray(value) ? value : [value]
+  return items.some(
+    (item): item is string =>
+      typeof item === 'string' && publicStreams.includes(item)
+  )
+}
```
llun, line 30: `hasPublicAudience` only matches when the audience entry is a plain string. ActivityPub allows `to`/`cc` entries to be link objects too (e.g. `{ id: 'https://www.w3.org/ns/activitystreams#Public', type: 'Collection' }`), and some implementations emit them. Because the `Note` schema declares `to`/`cc` as `z.string().array()`, this check today will reject a public-but-object-shaped audience as private and drop the status.

Either widen the schema and unwrap `{ id }` here, or document explicitly that this function only accepts string audiences and rely on a normalizer upstream. Right now it's an undocumented compatibility gap.

---

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,95 @@
...
+      signingActor
+    })
+    status.actor = actorPerson ? getActorProfileFromPerson(actorPerson) : null
+
+    return status
+  } catch (error) {
+    const nodeError = error as NodeJS.ErrnoException
+    logger.error(`[getRemoteStatus] ${nodeError.message}`)
+    return null
+  }
```
llun, line 65: Two issues with this catch:

1. `error as NodeJS.ErrnoException` is unsafe. `error` may be a string, `undefined`, a `ZodError`, or any non-Error value thrown by `getNote`/`getActorPerson`. Accessing `.message` on a non-object will throw, turning a logged warning into a 500. Use `error instanceof Error ? error.message : String(error)` (gemini already flagged this).
2. The catch is too coarse. `getActorPerson` is called *after* the public-status check, so a transient network error fetching the actor causes the entire successfully-fetched public note to be dropped. Consider narrowing the catch to `getNote` only, or letting `getActorPerson` failures fall through with `status.actor = null` (which the code already supports on line 87).

---

lib/activities/getActorPosts.ts
```diff
@@ -22,6 +23,54 @@ type GetActorPostsFunction = (params: {
...
+    .trim()
+
+const isOpaqueActorUsername = (actorId: string, username: string) => {
+  const actorIdUsername = getActorIdUsername(actorId)
+  return (
+    username === actorIdUsername &&
+    (username.startsWith('did:') ||
+      /^[0-9a-f]{8}-[0-9a-f-]{27,}$/i.test(username))
+  )
+}
```
llun: The opaque-actor regex `/^[0-9a-f]{8}-[0-9a-f-]{27,}$/i` is broader than it reads. It matches anything starting with 8 hex characters and a dash, then ≥27 hex-or-dash characters with no length cap and no required dash positions — so `aaaaaaaa-` followed by 27 hex characters (no dashes) matches as well, and so does an arbitrarily long suffix. If you intend to detect UUIDs (v4/v7), use a strict 8-4-4-4-12 form (`/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i`). For DIDs the `username.startsWith('did:')` check above is already correct; keep these two cases separate so the intent is obvious.

---

lib/activities/getActorPosts.ts
```diff
@@ -62,11 +141,13 @@ export const getActorPosts: GetActorPostsFunction = async ({
             })
             if (!note) return null
             const originalStatus = fromNote(note)
-            if (actor) originalStatus.actor = actor
-            return fromAnnoucne(
+            originalStatus.actor = await getActorProfile(originalStatus.actorId)
+            const announceStatus = fromAnnoucne(
               item as unknown as AnnounceStatus,
```
llun: `item as unknown as AnnounceStatus` and `obj as unknown as Note` (line 160) bypass validation entirely. The remote outbox could include announce items missing required fields (no `actor`, no `object`, malformed `published`), and `fromAnnoucne` / `fromNote` will then produce broken domain objects rather than skipping the row.

Validate with the existing Zod schemas (`AnnounceAction`/`Note`) and skip the item when parsing fails — that already aligns with how `item.type === AnnounceAction` and `obj.type !== 'Note'` early-returns are gating the same shape today.

---

app/(timeline)/[actor]/getProfileData.ts
```diff
@@ -21,6 +21,12 @@ type ProfileData = {
   hasFitnessData: boolean
 }
 
+const getImageUrl = (image: Actor['icon']) => {
+  if (!image) return ''
+  if (Array.isArray(image)) return image.find((item) => item.url)?.url || ''
+  return image.url
+}
```
Comment (llun) [inline app/(timeline)/[actor]/getProfileData.ts]: This local `getImageUrl` returns `''` on missing input, while the new helpers in `lib/activities/getRemoteStatus.ts` and `lib/activities/getActorPosts.ts` return `undefined`. Both are then passed into `database.updateActor({ iconUrl, headerImageUrl })` / `ActorProfile.parse({ iconUrl })` respectively — having the empty-value contract differ across call sites means one site may write empty strings to the DB while the other writes nothing, and downstream consumers (template fallbacks, optional-prop checks) behave differently. Please reconcile these as part of the dedup suggested in `getRemoteStatus.ts`.

---

lib/activities/getRemoteStatus.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        to: [ACTOR_ID],
+        cc: [],
+        published: new Date('2026-04-30T12:00:00.000Z').toISOString()
+      })
+    )
+
+    await expect(getRemoteStatus({ statusId: STATUS_ID })).resolves.toBeNull()
+    expect(fetchMock).toHaveBeenCalledTimes(1)
+  })
+})
```
llun, line 158: Coverage gaps for this new module:

- The `catch` branch (line 90–94) is never exercised — no test where `getNote` rejects or `Note.parse`/`getActorPerson` throws. Both the unsafe error cast and the wide catch boundary called out above are invisible to the suite right now.
- No test for the `getActorPerson` returning `null` path (line 87) — `status.actor` should still be `null` while the rest of the status is preserved.
- No test for an `Announce` payload reaching this function — `fromNote` is called unconditionally, so an attempt to fetch an `Announce` URL here would silently corrupt.
- The "non-public" test only covers `to: [ACTOR_ID]`. Add a case where `to` is a public-stream object reference (`{ id: '…#Public', type: 'Collection' }`) to lock the audience-shape behavior in (see comment on `hasPublicAudience`).

---

lib/utils/getStatusDetailPath.ts
```diff
@@ -7,5 +7,9 @@ export const getStatusDetailPath = (status: Status) => {
   const actualStatus = getActualStatus(status)
   if (!actualStatus.actor) return null
 
+  if (actualStatus.isLocalActor === false) {
+    return `/${getMention(actualStatus.actor, true)}/${encodeURIComponent(actualStatus.id)}`
```
llun, line 11: For remote actors, the path is now `/${mention}/${encodeURIComponent(actualStatus.id)}`, where `actualStatus.id` is typically a full URL like `https://remote.example/users/alice/statuses/123`. After URI-encoding that becomes `https%3A%2F%2Fremote.example%2Fusers%2Falice%2Fstatuses%2F123` — a long, opaque path segment that has to round-trip through `decodePathParam` in `resolveStatusFromPath`. That works, but it's worth confirming:

1. None of the routes/middleware along the way reject path segments above their length budget (Next.js itself is fine, but any reverse proxy or CDN may have stricter limits than you expect).
2. `getStatusDetailPathClient.ts` mirrors this exactly — both files were touched but please double-check that the client path uses the same encoding so client-side navigation and server-rendered links don't disagree for the same status.

Also: `actualStatus = getActualStatus(status)` returns `originalStatus` for `Announce`. So a boost of a remote status now produces a link to the original-author route, but `actualStatus.isLocalActor` is checked against the *original*. If the original is missing `isLocalActor` (treated as truthy/falsy), the path branch can flip — please verify the announce shape used by callers always populates `isLocalActor` on the wrapped original, or default to `false` when undefined.

---

lib/components/posts/actor.tsx
```diff
@@ -23,15 +22,50 @@ const getInitials = (value: string) =>
...
+    if (!handle) return null
+    return `@${getDisplayUsername(handle)}`
+  } catch {
+    return null
+  }
+}
+
+const getActorIdHandle = (actorId: string, statusUrl?: string | null) =>
+  getStatusUrlHandle(statusUrl) ||
+  `@${getDisplayUsername(actorId.split('/').filter(Boolean).pop() || actorId)}`
```
llun: `getActorIdHandle` falls back to `actorId.split('/').filter(Boolean).pop()` when `getStatusUrlHandle` returns null. For a Bridgy-style actor id like `https://hackers.pub/ap/actors/019382d3-…` the popped segment is the opaque UUID; the test on line 228 of `post.test.tsx` covers the case where `statusUrl` rescues us, but if `statusUrl` is also missing/opaque (e.g. `https://hackers.pub/ap/notes/<uuid>`) the displayed handle is `@<uuid>`, which is exactly the regression you're trying to avoid.

Consider adding an explicit guard: when the popped segment matches `isOpaqueActorUsername` (UUID/DID), prefer the actor profile's `preferredUsername` if present, or fall back to the actor's `domain` host as a last resort. At minimum, add a test for the "opaque actor id with no statusUrl" case so the fallback chain is documented.

---

llun (commented):
Reviewed the latest commits on this branch (52c0233..f112e8a) covering the opaque-actor handling, the new `getRemoteStatus` synchronous fetch, and the changes to `getStatusDetailPath`. The earlier-resolved threads cover the resolver itself; my comments below focus on the new code that landed after that and a couple of carry-over concerns.

Highlights of what should be fixed before merging:

1. **Helper duplication** — `getImageUrl`, `getActorDomain`, and `getActorProfileFromPerson` are now copy-pasted across `getRemoteStatus.ts`, `getActorPosts.ts`, `actions/utils.ts`, and `getProfileData.ts`, with one site already drifting on the empty-value contract (`''` vs `undefined`). Extract to a shared module.
2. **Missing schema validation in `getRemoteStatus`** — the remote payload is cast to `Note` instead of being parsed. Combined with the broad `try`/`catch` and the unsafe `error as NodeJS.ErrnoException` cast, malformed remote responses produce silent corruption and `getActorPerson` failures discard already-fetched public notes.
3. **`hasPublicAudience` only matches string entries** — public collections delivered as link objects are misclassified as private.
4. **`isOpaqueActorUsername` regex is too permissive** — tighten to a real UUID shape; keep the DID branch separate.
5. **Unsafe `as unknown as` casts in `getActorPosts`** — validate announce/note items with the existing Zod schemas before handing them to `fromAnnoucne`/`fromNote`.
6. **Page-level duplicate fetch** — `page.tsx` now both synchronously fetches `getRemoteStatus` and queues a `fetch-remote-status-…` job for the same `fullStatusId` on the failure path; short-circuit one of them.
7. **`getStatusDetailPath` change** — verify `actualStatus.isLocalActor` is reliably populated for the `Announce` → `originalStatus` unwrap, and confirm the client variant encodes identically.
8. **Test gaps** — the new `getRemoteStatus` suite never exercises the catch branch, the `null` actor branch, or non-string audience shapes; the opaque-actor handling lacks a "no statusUrl rescue" case.

Inline comments have the specifics. None of these are blockers individually, but together they cover the rough edges of the new code paths added in the most recent commits.

---

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,95 @@
...
+    const actorPerson = await getActorPerson({
+      actorId: status.actorId,
+      signingActor
+    })
+    status.actor = actorPerson ? getActorProfileFromPerson(actorPerson) : null
+
+    return status
+  } catch (error) {
+    const nodeError = error as NodeJS.ErrnoException
+    logger.error(`[getRemoteStatus] ${nodeError.message}`)
```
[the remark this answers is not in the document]
llun: Fixed in 9778fa41. `getRemoteStatus` now logs unknown errors through a small `getErrorMessage(error)` helper using `error instanceof Error ? error.message : String(error)`, and Zod parse failures are handled explicitly.

lib/activities/getRemoteStatus.ts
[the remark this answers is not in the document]
llun: Fixed in 9778fa41. I extracted the duplicated ActivityPub actor helpers into `lib/utils/activitypubActor.ts` and updated the status fetch, actor posts, action utils, and profile update paths to share the same implementation.

llun: Fixed in 9778fa41. `getActorImageUrl`, `getActorDomain`, `getActorProfileFromPerson`, and opaque username detection now live in `lib/utils/activitypubActor.ts`, and the empty image contract is consistently `undefined`.

lib/activities/getRemoteStatus.ts
llun: Fixed in 9778fa41. The remote status path now runs `Note.safeParse(normalizeActivityPubContent(remoteNote))` before `fromNote`, returning `null` for invalid remote payloads instead of casting them through.

lib/activities/getRemoteStatus.ts
llun: Fixed in 9778fa41. This path now normalizes recipient objects before validation, so `hasPublicAudience` sees string IDs. I added coverage for `{ id: "https://www.w3.org/ns/activitystreams#Public", type: "Collection" }` resolving as public.

lib/activities/getRemoteStatus.ts
llun: Fixed in 9778fa41. `getRemoteStatus` now avoids `NodeJS.ErrnoException` casts and isolates actor-profile fetch failures so a valid public note is still returned with `status.actor = null` when the actor lookup fails.

lib/activities/getActorPosts.ts
llun: Fixed in 9778fa41. The opaque actor detection now uses a strict UUID `8-4-4-4-12` regex and a separate DID check, with unit coverage for UUID, over-broad non-UUID, DID, and normal username cases.

lib/activities/getActorPosts.ts
llun: Fixed in 9778fa41. Remote outbox Announce and Create/Note items are now normalized and validated with `Announce.safeParse` / `Note.safeParse`, and invalid rows are skipped. I also guard `fromNote` failures and added malformed outbox coverage.

app/(timeline)/[actor]/getProfileData.ts
Comment (llun) [reply app/(timeline)/[actor]/getProfileData.ts]: Fixed in 9778fa41. The profile update path now uses the shared `getActorImageUrl` helper instead of a local helper returning empty strings, so missing images consistently remain `undefined`.

lib/activities/getRemoteStatus.test.ts
llun: Fixed in 9778fa41. Added coverage for object-shaped public audience plus actor lookup returning `null`, non-Note/Announce payloads returning `null`, malformed notes returning `null` through the `fromNote` catch, and malformed remote outbox activities being skipped.

lib/utils/getStatusDetailPath.ts
llun: No code change needed. Server and client helpers already have matching tests for remote Announce links using `encodeURIComponent(actualStatus.id)`, and `resolveStatusFromPath` covers decoding full remote IDs. The affected Announce/original status paths populate `isLocalActor: false` through `fromAnnoucne` / `fromNote`, so the remote branch is stable. There is no app-level route or middleware limit below what Next.js accepts; any lower reverse-proxy/CDN path budget would be deployment configuration rather than this branch.

lib/components/posts/actor.tsx
llun: Fixed in 9778fa41. Opaque actor IDs now prefer a status URL handle, support Bridgy Bluesky `profile/<handle>` URLs, and otherwise fall back to the actor domain/external actor href instead of displaying UUID/DID segments. Added tests for opaque actor IDs with no usable status handle and Bridgy profile URLs.

lib/components/posts/actor.tsx
```diff
@@ -23,15 +24,103 @@ const getInitials = (value: string) =>
...
+  const username = getDisplayUsername(getActorIdUsername(actorId))
+  if (isOpaqueActorUsernameValue(username)) {
+    const domain = getActorIdDomain(actorId)
+    return {
+      handle: domain || `@${username}`,
+      domain: '',
+      href: actorId
+    }
+  }
+
```
llun, line 139: The opaque-actor branch sets `href: actorId`, which is the raw ActivityPub URL (e.g. `https://hackers.pub/ap/actors/<uuid>`). Both the `ActorAvatar` and `ActorInfo` components then render a Next.js `<Link>` pointing at that URL, so clicking the avatar/handle on a post by an opaque actor takes the user to the server-to-server JSON document rather than a human-readable profile.

A few angles to consider:

1. **UX**: most users don't want to land on a JSON page. The Bluesky `bsky.app/bsky.brid.gy` handling above already extracts a non-opaque handle when one exists in the status URL — when it doesn't, a sensible fallback is probably "no link, just text" or "link to the actor's `url` field if present" (Hackers.pub, Bridgy, etc. expose human URLs alongside the opaque AP id).
2. **Security/UX 2**: if you do keep an external link, it should have `target="_blank" rel="noopener noreferrer"` and probably not be a Next.js `<Link>` (Link is meant for in-app navigation). Today this navigates same-tab and through Next's prefetch path.
3. **Tests**: the new test on `post.test.tsx:281` asserts `href === actorId`. That locks in the JSON-page behavior — please double-check that's intended and not just "what the code does today".

---

lib/components/posts/actor.tsx
```diff
@@ -23,15 +24,103 @@ const getInitials = (value: string) =>
...
+      parts.includes('bsky.app')
+    if (
+      isBlueskyProfileUrl &&
+      profileIndex >= 0 &&
+      profileHandle &&
+      !isOpaqueActorUsernameValue(profileHandle)
+    ) {
+      return `@${getDisplayUsername(profileHandle)}`
+    }
+
```
llun: The Bluesky special case is a useful escape hatch, but it's hardcoded and a little fragile:

1. `parts.includes('bsky.app')` matches *any* path segment equal to `bsky.app`. For a Bridgy URL like `https://bsky.brid.gy/r/https://bsky.app/profile/foo/post/…` this is what makes the test pass (the embedded `bsky.app` sneaks into `parts` after `decodeURIComponent`). But it would also fire on an unrelated host that happens to put `bsky.app` in its path, leading to a confidently-wrong handle.
2. The same shape exists for other Bridgy/proxy services (`fed.brid.gy`, `lemmy.world` `/c/<community>`, Misskey `/@<user>`, etc.). Hardcoding only Bluesky means we'll need to revisit this every time a new platform's URL pattern shows up.

Two options to make this more robust:

- Keep the heuristic but anchor it: only treat the URL as Bluesky when `url.hostname` is one of the known Bridgy/Bluesky hosts. Drop the `parts.includes('bsky.app')` fallback and instead, for `bsky.brid.gy`, parse the embedded URL directly (it's a known shape: `/r/<full-url>`).
- Generalize: extract the "find a non-opaque handle in a URL" logic into a small registry keyed by hostname pattern, with the Mastodon `@user` case as the default. New platforms become a one-line addition.

Either way, please add a test that asserts an unrelated URL with `bsky.app` somewhere in its path doesn't accidentally inherit a Bluesky handle — that would be a quiet regression today.

---

app/(timeline)/[actor]/getProfileData.ts
```diff
@@ -78,8 +79,8 @@ export const getProfileData = async (
       actorId: person.id,
       name: person.name,
       summary: person.summary || '',
-      iconUrl: person.icon?.url || '',
-      headerImageUrl: person.image?.url || '',
+      iconUrl: getActorImageUrl(person.icon),
+      headerImageUrl: getActorImageUrl(person.image),
```
Comment (llun) [inline app/(timeline)/[actor]/getProfileData.ts:83]: Subtle behavior change worth confirming. The previous local helper here returned `''` for missing images; the shared `getActorImageUrl` returns `undefined`. So `database.updateActor({ iconUrl: getActorImageUrl(person.icon), headerImageUrl: getActorImageUrl(person.image), … })` now passes `undefined` instead of `''` when the remote actor has no icon/header.

That matters because most ORM/SQL builders treat `undefined` as "skip this column" but `''` as "set to empty string". If the existing actor row already has a stale `iconUrl`, this change flips the semantics from "clear it" to "leave whatever was there", which is the opposite of what `getProfileData` previously did when the remote actor dropped their avatar.

Please verify against `database.updateActor`'s implementation. If the goal is "always reflect the current remote state", you probably want the empty-string contract back here even if the shared helper returns `undefined` — e.g. `iconUrl: getActorImageUrl(person.icon) ?? ''`. The other call sites (`recordActorIfNeeded`) already guard with `...(iconUrl ? { iconUrl } : {})`, so they're consistent with their own intent — but `getProfileData` doesn't, so it inherits a different behavior now.

---

lib/activities/getActorPosts.ts
```diff
@@ -22,6 +31,14 @@ type GetActorPostsFunction = (params: {
   statuses: Status[]
 }>
 
+const getStatusFromNote = (note: Note) => {
+  try {
+    return fromNote(note)
+  } catch {
+    return null
+  }
```
llun, line 44: Tiny but real: `fromNote` already runs after `Note.safeParse` succeeded, so the `try`/`catch` here is hiding genuinely unexpected runtime errors (e.g. a regression in `fromNote` itself) by logging nothing and returning `null`. If `Note.safeParse` accepts a payload but `fromNote` then throws, you want to know about it loudly, not have the timeline silently drop the row.

Consider either:

- Letting `fromNote` throw so the surrounding `Promise.all` rejects and the actual stack trace surfaces in your logs (the `safeParse` already gates malformed input, so a throw here means a real bug).
- Or keep the catch but `logger.error` the captured error like `getRemoteStatus` now does, so it's at least observable.

The current silent-null path is the worst of both worlds.

---

lib/utils/activitypubActor.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    expect(isOpaqueActorUsername(actorId, 'hongminhee')).toBe(false)
+  })
+
+  it('detects DID actor usernames only when they match the actor id', () => {
+    const actorId = 'https://bsky.brid.gy/ap/did:plc:alice'
+
+    expect(isOpaqueActorUsername(actorId, 'did:plc:alice')).toBe(true)
+    expect(isOpaqueActorUsername(actorId, 'alice.example')).toBe(false)
+  })
+})
```
llun, line 65: Strong overall. Two minor suggestions:

1. The DID branch test only covers `did:plc:alice`. Add a case where `username === 'alice'` but `actorId` is `…/did:plc:alice` to ensure the function returns `false` (mismatch on equality even though the username is plausibly DID-like). That guards the `username === actorIdUsername` half of the predicate, which currently only has implicit coverage.
2. Consider co-locating a test for `isOpaqueActorUsernameValue` directly — it's exported and used standalone in `actor.tsx`'s `getStatusUrlHandle`, so a regression there (e.g. someone loosening the regex) would only surface through UI tests today.

---

lib/types/domain/status.ts
```diff
@@ -211,7 +212,7 @@ export const fromNote = (note: Note): StatusNote => {
 }
 
 export const fromAnnoucne = (
-  announce: AnnounceStatus,
+  announce: AnnounceStatus | ActivityPubAnnounce,
```
llun, line 215: Widening `fromAnnoucne` to accept either `AnnounceStatus` or `ActivityPubAnnounce` is the right move for the `getActorPosts` refactor, but it does mean callers can now pass two structurally different shapes through the same function. Worth a quick audit that the implementation reads only the fields shared by both types — e.g. `id`, `actor`, `published`, `to`, `cc`, `object` — and not anything that exists only on `AnnounceStatus`. If there's a divergence (`type`-only field, `_meta`-style additions, etc.), wrapping callers in a small adapter would be safer than a union signature.

Also, please rename to `fromAnnounce` while you're touching the signature — the existing `fromAnnoucne` typo has been there for a while, but every new call site this PR adds entrenches it further.

---

llun (commented):
Re-reviewed after 9778fa4. The bulk of the previous feedback landed cleanly:

- `getImageUrl`/`getActorDomain`/`getActorProfileFromPerson` are now centralized in `lib/utils/activitypubActor.ts`, including a tightened UUID regex (`/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i`).
- `getRemoteStatus` validates the remote payload through `Note.safeParse`, narrows the catch boundary so a failed actor lookup no longer drops a successfully-fetched note, and replaces the unsafe `error as NodeJS.ErrnoException` cast with `getErrorMessage`.
- `getActorPosts` now `safeParse`s outbox items as `Announce`/`Note` instead of using `as unknown as` casts, and skips malformed rows with a dedicated test.
- `hasPublicAudience` is exercised with object-shaped public collections in the new test, and the broad-catch / null-actor / non-Note paths all have direct coverage now.

Remaining items, in order of priority:

1. **Page-level duplicate fetch** — `page.tsx` still publishes a `fetch-remote-status-${fullStatusId}` job after `getRemoteStatus` synchronously failed for the same id. Either gate the queue path on a transient/recoverable error, or drop it now that the synchronous fetch covers the success path.
2. **Opaque-actor link target** — the new opaque branch in `actor.tsx` sets `href: actorId`, which is a raw ActivityPub JSON URL. Worth confirming that's the intended UX (and adding `target="_blank" rel="noopener noreferrer"` if so, since this isn't an in-app navigation anymore).
3. **Bluesky URL heuristic** — `parts.includes('bsky.app')` will misfire for any URL that happens to embed `bsky.app` as a path segment. Anchor on `url.hostname` plus an explicit Bridgy `/r/<embedded-url>` parser, or generalize to a per-host strategy.
4. **`getProfileData` empty-value contract** — switching from the local `getImageUrl` (returned `''`) to the shared `getActorImageUrl` (returns `undefined`) silently changes what `database.updateActor` sees for missing icon/header. Please verify this matches the prior "clear stale value" intent.
5. **`getActorPosts` silent catch** — `getStatusFromNote` swallows `fromNote` exceptions without logging. Once schema validation has passed, an exception there is a real bug; either let it bubble or at least `logger.error` it.
6. **Status-domain function rename** — while `fromAnnoucne` is being widened, please rename to `fromAnnounce` rather than entrenching the typo across new call sites.

Inline comments have specifics. Items (1) and (2) are user-visible; the rest are correctness-and-clarity polish.

---

lib/components/posts/actor.tsx
llun: Fixed in ed57d169. The opaque actor fallback no longer links to the raw ActivityPub actor ID. When no human-readable handle can be derived, the avatar/handle render as non-links, and the test now asserts we do not link @hackers.pub to the AP JSON URL.

lib/components/posts/actor.tsx
llun: Fixed in ed57d169. The Bluesky parser is now anchored to direct bsky.app URLs or bsky.brid.gy /r/<embedded bsky.app URL> only, and I added a negative test for an unrelated URL that contains bsky.app in its path.

app/(timeline)/[actor]/getProfileData.ts
Comment (llun) [reply app/(timeline)/[actor]/getProfileData.ts:83]: No code change needed after checking database.updateActor. It already treats both empty strings and undefined as "do not update" because icon/header fields are included only when truthy, so the prior getProfileData empty-string helper did not clear stale image URLs either. The shared helper does not change that behavior.

lib/activities/getActorPosts.ts
llun: Fixed in ed57d169. I kept the skip behavior for malformed remote rows but added error logging in getStatusFromNote so fromNote failures are observable instead of silent.

lib/utils/activitypubActor.test.ts
llun: Fixed in ed57d169. Added the DID equality mismatch case and direct coverage for isOpaqueActorUsernameValue, including DID, strict UUID, over-broad UUID-like, and normal username inputs.

lib/types/domain/status.ts
llun: Fixed in ed57d169. The canonical export is now fromAnnounce, getActorPosts uses that spelling, and fromAnnoucne remains only as a compatibility alias. I also audited the implementation; it reads only fields shared by AnnounceStatus and ActivityPubAnnounce: id, actor, to, cc, and published.

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+  ACTIVITY_STREAM_PUBLIC,
+  ACTIVITY_STREAM_PUBLIC_COMPACT,
+  'Public',
+  'as:Public'
+]
+
+const hasPublicAudience = (value: Note['to'] | Note['cc']) => {
+  const items = Array.isArray(value) ? value : [value]
+  return items.some((item) => publicStreams.includes(item))
+}
```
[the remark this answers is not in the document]
llun: No code change needed. getRemoteStatus calls normalizeActivityPubContent(remoteNote) before Note.safeParse, and that normalizer unwraps recipient objects through id/href/url. hasPublicAudience intentionally receives the normalized Note shape, whose to/cc schema is string or string array, so the object-shaped Public audience regression test covers this path and passes.

lib/activities/getRemoteStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+const publicStreams = [
+  ACTIVITY_STREAM_PUBLIC,
+  ACTIVITY_STREAM_PUBLIC_COMPACT,
+  'Public',
+  'as:Public'
+]
+
+const hasPublicAudience = (value: Note['to'] | Note['cc']) => {
+  const items = Array.isArray(value) ? value : [value]
+  return items.some((item) => publicStreams.includes(item))
```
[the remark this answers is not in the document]
llun: No code change needed for the same normalized-audience reason. Raw ActivityPub recipient objects are normalized before Note.safeParse, and Note.to/Note.cc are typed as string or string array by the time hasPublicAudience runs. The existing object-shaped Public audience test exercises this case.

app/(timeline)/[actor]/getProfileData.ts
Comment (llun) [inline app/(timeline)/[actor]/getProfileData.ts:83]: Also unchanged at the current tip — `getProfileData.ts` still calls `database.updateActor({ iconUrl: getActorImageUrl(person.icon), headerImageUrl: getActorImageUrl(person.image), … })`. The shared `getActorImageUrl` returns `undefined` on missing input, while the previous local helper returned `''`. For a remote actor that drops their icon/header between fetches, `''` would clear the column on most ORM/SQL builders, while `undefined` typically becomes "skip this column" and leaves the stale URL in place.

This is a silent semantic flip from the prior behavior. Either:

- `iconUrl: getActorImageUrl(person.icon) ?? ''` (preserve the "always reflect remote state" intent), or
- confirm against `database.updateActor`'s implementation that `undefined` here genuinely means "set to empty / null", not "leave existing value".

Right now the PR description doesn't mention this contract change, so it's worth a one-line confirmation either in code or in the PR.

---

lib/components/posts/actor.tsx
```diff
@@ -23,41 +30,163 @@ const getInitials = (value: string) =>
...
+    const embeddedUrl = new URL(
+      decodeURIComponent(url.pathname.slice('/r/'.length))
+    )
+    if (embeddedUrl.hostname !== 'bsky.app') return null
+    return getProfileHandleFromParts(
+      embeddedUrl.pathname.split('/').filter(Boolean).map(decodeURIComponent)
+    )
+  } catch {
+    return null
+  }
```
llun, line 86: Nice — the Bridgy `/r/<embedded-url>` shape is now parsed explicitly, so the brittle `parts.includes('bsky.app')` fallback is gone, and the new test on `post.test.tsx:315` ("does not infer bsky profile handles from unrelated status url paths") locks the fix in. Combined with `bsky.app` being matched only via `url.hostname`, the heuristic is much harder to misfire now.

One nit: the `decodeURIComponent` on `url.pathname.slice('/r/'.length)` will throw `URIError` on a malformed escape sequence (e.g. `/r/%E0%A4%A`). The surrounding `try`/`catch` handles it, but consider doing the same `decodePathParam`-style guarded decode you used in `resolveStatusFromPath` so a malformed Bridgy URL is treated as "no Bluesky handle" rather than going through the catch path that swallows everything.

---

lib/types/domain/status.ts
```diff
@@ -240,6 +241,8 @@ export const fromAnnoucne = (
   })
 }
 
+export const fromAnnoucne = fromAnnounce
```
llun: The backward-compat alias is fine, but consider removing it in a follow-up. With the rename landed and `getActorPosts` already migrated, this only protects against external imports of `fromAnnoucne` — which there shouldn't be any of inside this repo. Keeping the typo alive forever is a slow form of debt; a quick `grep -r fromAnnoucne` should confirm it's safe to drop.

---

llun (commented):
Re-reviewed at ed57d16. Four of the six prior items are landed:

- ✅ **Bluesky heuristic** — `parts.includes('bsky.app')` is gone; matching now anchors on `url.hostname` with an explicit Bridgy `/r/<embedded-url>` parser, and a regression test on `post.test.tsx:315` covers the unrelated-path case.
- ✅ **Opaque-actor link target** — `getActorIdParts` no longer returns `href: actorId` for opaque actors. `ActorAvatar` renders a plain `<div>` and `ActorInfo` renders a `<span>` instead of a Next.js `<Link>` to a raw AP JSON URL. Test updated accordingly.
- ✅ **`getActorPosts` silent catch** — `getStatusFromNote` now `logger.error`s captured exceptions via a shared `getErrorMessage` helper.
- ✅ **`fromAnnoucne` rename** — renamed to `fromAnnounce`, callers migrated, with a backward-compat `fromAnnoucne = fromAnnounce` alias retained.

Two items still open at the current tip (file unchanged in this revision):

1. **`page.tsx` duplicate fetch** — synchronous `getRemoteStatus` + queue publish for the same id still both fire on the failure path.
2. **`getProfileData` empty-value contract** — `getActorImageUrl` returns `undefined` while the prior local helper returned `''`; semantics of the `database.updateActor` call have silently changed.

Plus one nit on the new code: `decodeURIComponent` on the Bridgy embedded path can throw on malformed escapes; the surrounding catch hides that, but a guarded decode would be clearer. And a recommendation to drop the `fromAnnoucne` alias in a follow-up.

Inline comments have specifics. None of the open items are blocking, but (1) and (2) are real semantic changes that I don't see called out in the PR description.

---

lib/utils/activitypubActor.ts
```diff
@@ -0,0 +1,54 @@
...
+    headerImageUrl: getActorImageUrl(person.image),
+    manuallyApprovesFollowers: person.manuallyApprovesFollowers,
+    followersUrl: person.followers || `${person.id}/followers`,
+    inboxUrl: person.inbox,
+    sharedInboxUrl: person.endpoints?.sharedInbox || '',
+    followingCount: 0,
+    followersCount: 0,
+    statusCount: 0,
+    lastStatusAt: null,
+    createdAt: person.published ? new Date(person.published).getTime() : 0
```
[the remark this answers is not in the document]
llun: Fixed in b7f787ee. getActorProfileFromPerson now routes published through a small Date.parse guard and falls back to 0 when the value is missing or invalid. Added coverage for malformed actor published dates.

app/(timeline)/[actor]/getProfileData.ts
Comment (llun) [reply app/(timeline)/[actor]/getProfileData.ts:83]: No code change needed after checking database.updateActor. The SQL adapter only writes iconUrl/headerImageUrl when the value is truthy, so the prior empty-string helper also did not clear stale image settings. I added this behavior note to the PR description so the shared helper contract is explicit.

lib/components/posts/actor.tsx
llun: Fixed in b7f787ee. Bridgy embedded URL decoding now goes through a guarded decode helper and returns no Bluesky handle on malformed escapes. Added a post rendering test for a malformed /r/%E0%A4%A URL.

lib/types/domain/status.ts
llun: Fixed in b7f787ee. grep confirmed there are no remaining repo imports of fromAnnoucne, so I removed the compatibility alias and left only the corrected fromAnnounce export.

lib/components/posts/actor.tsx
```diff
@@ -23,41 +30,178 @@ const getInitials = (value: string) =>
...
+}
+
+const decodePathParam = (value: string) => {
+  try {
+    return decodeURIComponent(value)
+  } catch {
+    return null
+  }
+}
```
llun, line 66: Nice cleanup. `getDecodedPathParts` and `decodePathParam` consolidate the guarded-decode pattern and remove the implicit reliance on the outer `try`/`catch`. The malformed-Bridgy-URL test (`https://bsky.brid.gy/r/%E0%A4%A`) locks the behavior in. No further concerns here.

---

lib/utils/activitypubActor.ts
```diff
@@ -0,0 +1,60 @@
...
+export const getActorImageUrl = (image: ActivityPubActor['icon']) => {
+  if (!image) return undefined
+  if (Array.isArray(image)) return image.find((item) => item.url)?.url
+  return image.url
+}
+
+const getActorCreatedAt = (published: string | null | undefined) => {
+  if (!published) return 0
+  const createdAt = Date.parse(published)
+  return Number.isFinite(createdAt) ? createdAt : 0
```
llun, line 37: Good fix. `Date.parse('not-a-date')` returns `NaN`, which `new Date(NaN).getTime()` would have propagated as `NaN` into `ActorProfile`'s `createdAt: number`. Falling back to `0` keeps the value finite and the unit test pins the contract. Tiny suggestion for a follow-up: also coerce non-finite values to `0` in `getActorCreatedAt` if `published` is a number masquerading as a string (e.g. `'Infinity'`) — `Date.parse` handles those inconsistently across engines.

---

llun (commented):
Final pass at b7f787e. The latest commit closes out the remaining inline items from the previous round:

- ✅ **Bridgy/Bluesky decode safety** — guarded helpers `getDecodedPathParts` and `decodePathParam` replace the bare `decodeURIComponent` calls; the surrounding `try`/`catch` is no longer load-bearing. New test `'ignores malformed bridgy embedded status urls'` (`post.test.tsx:315`) covers the `%E0%A4%A` case.
- ✅ **`fromAnnoucne` alias removed** — the rename is now clean, no more typo trail.
- ✅ **Invalid `published` dates** — bonus catch via the new `getActorCreatedAt` helper; `Date.parse` returning `NaN` is now coerced to `0` instead of leaking into the `ActorProfile.createdAt: number` contract. Unit test on `activitypubActor.test.ts:46` pins it.

**Net summary across all rounds:** the boosted-status route fix itself (resolver + Announce handling + actor profile hydration) is solid, well-tested, and addresses the original 404 cleanly. Helper duplication, schema validation, error handling, opaque-actor link UX, and Bluesky URL handling all landed.

**Two carry-over items remain at file level**, both flagged in earlier rounds and unchanged in this commit:

1. `page.tsx` — synchronous `getRemoteStatus` + queue publish for the same `fullStatusId` on the failure path. The deterministic queue id avoids in-queue duplicates but not the duplicate work between the synchronous attempt and the queued one.
2. `getProfileData.ts` — `database.updateActor({ iconUrl: getActorImageUrl(...), … })` switched from `'' | string` to `undefined | string`, which is a semantic change for any builder that interprets `undefined` as "skip column".

Both are tangential to the PR's stated goal ("Resolve boosted status hash owner routes") and reasonable to handle as follow-ups if you'd prefer to keep this PR focused — happy to defer to your judgment on scope. No new blocking concerns from this round; LGTM modulo those two follow-ups.

---

PR #626, opened 2026-05-02
minor: improve ActivityPub and Mastodon compatibility

llun:
## Summary
- Preserve opaque remote ActivityPub actor IDs instead of deriving identity from URL path segments.
- Serialize outbox `Create` objects as ActivityPub Notes/Questions and add collection endpoints for featured posts/tags plus status replies, likes, and shares.
- Federate polls as ActivityPub `Question` objects and add Mastodon-compatible poll read/vote endpoints.
- Safely accept additional inbox activities seen across Mastodon and other fediverse servers (`Block`, `Flag`, `Move`, `Add`, `Remove`, `QuoteRequest`, actor `Update`, and reference-only `Undo`).
- Centralize local actor/status ID construction behind stable helpers while keeping existing public ActivityPub actor URLs compatible.
- Align touched API routes with `apiResponse` conventions and remove non-script/client `console.*` calls from touched paths.

## Compatibility Notes
- Actor IDs remain stable on the wire; the new helpers make local ID construction consistent without changing local ActivityPub actor URLs.
- Remote actor parsing now tolerates DID, Bridgy-style, and other opaque IDs that contain colon-bearing path segments by using an encoded fallback.
- ActivityPub collection typing accepts compact collection summaries (`totalItems` without embedded `items`), which is common in Mastodon payloads.
- Poll support now maps local polls to ActivityPub `Question` choices, vote totals, close times, and closed state, and exposes Mastodon REST poll GET/vote behavior.

## Migration / Backward Compatibility
- No database migrations are included.
- No manual `package.json` version change is included; the PR title uses `minor:` for the repository version-bump workflow.
- Existing colon-style REST/status IDs continue to be accepted while safer encoded IDs are centralized through helpers.

## Validation
- Node: `v24.13.0`
- `yarn run prettier --write .` passed
- `yarn lint` passed with existing `no-explicit-any` warnings only
- `yarn build` passed
- `yarn test` passed: 232 suites, 1,782 tests

## Follow-up Scope
- This PR improves the ActivityPub and Mastodon compatibility foundation, especially federation payloads, actor IDs, collections, and polls. Full Mastodon feature parity for bookmarks, blocks, mutes, pins, lists, followed tags, search, and richer instance metadata should continue in smaller endpoint-focused PRs.


llun:
@codex review

lib/types/domain/status.ts
```diff
@@ -267,9 +278,26 @@ export const toActivityPubObject = (status: Status): Note | Question => {
...
+        type: 'Collection',
+        totalItems: 0
+      },
 
       published: getISOTimeUTC(status.createdAt),
       endTime: getISOTimeUTC(status.endAt),
+      votersCount: status.choices.reduce(
+        (totalVotes, choice) => totalVotes + choice.totalVotes,
+        0
+      ),
```
llun: `votersCount` is wrong for `pollType === 'anyOf'`. Summing `choice.totalVotes` overcounts unique voters, since each voter can pick multiple options. Mastodon's `voters_count` is the distinct voter count. Track voters separately (e.g. distinct `actorId` count from poll answers) or only emit `votersCount` for `oneOf`.

---

lib/types/domain/status.ts
```diff
@@ -267,9 +278,26 @@ export const toActivityPubObject = (status: Status): Note | Question => {
...
+      likes: {
+        id: `${status.id}/likes`,
+        type: 'Collection',
+        totalItems: status.totalLikes
+      },
+      shares: {
+        id: `${status.id}/shares`,
+        type: 'Collection',
+        totalItems: 0
+      },
```
llun, line 291: `shares.totalItems` is hardcoded to `0` for both Poll and Note paths (also in `app/api/users/[username]/statuses/[statusId]/shares/route.ts`). Boost (Announce) counts will silently be zero on every federated payload, which misleads remote servers. Compute it from the announce/share data or at minimum add a TODO that this is unimplemented.

---

app/api/users/[username]/statuses/[statusId]/likes/route.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return activityPubResponse({
+      req,
+      data: {
+        '@context': ACTIVITY_STREAM_URL,
+        id: `${status.id}/likes`,
+        type: 'Collection',
+        totalItems,
+        items: likedBy.map((likedByActor) => likedByActor.id)
+      }
```
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/likes/route.ts:41]: Privacy concern: this exposes the actor URL of every user who liked the status. Mastodon deliberately returns only `totalItems` here and keeps the liker list private. Consider dropping `items` (and the `getFavouritedBy` call) and returning just the count.

---

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.POST]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+const VotePollRequest = z.object({
+  choices: z.number().array().min(1)
+})
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:34]: `choices: z.number().array().min(1)` accepts any number, including negatives, non-integers, or values >= `status.choices.length`. Those values are then written straight into `createPollAnswer`/`incrementPollChoiceVotes`. Tighten with `z.number().int().nonnegative()` and bounds-check against `status.choices.length` after fetching the status, and reject duplicates.

---

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+      (status.pollType === 'oneOf' && hasVoted) ||
+      (status.pollType === 'oneOf' && choices.length > 1)
+    ) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_422,
+        responseStatusCode: 422
+      })
+    }
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:143]: `hasVoted` is only enforced for `oneOf`. For `anyOf` polls, the same actor can keep voting on every request and inflate tallies indefinitely. Either gate `anyOf` on `hasVoted` (per choice) or have `createPollAnswer` enforce uniqueness on `(statusId, actorId, choice)`.

---

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+      )
+    )
+    await sendPollVotes({ currentActor, status, choices })
+
+    const updatedStatus = await database.getStatus({
+      statusId,
+      withReplies: false
+    })
+    if (!updatedStatus || updatedStatus.type !== StatusType.enum.Poll) {
+      return apiResponse({
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:167]: The two sequential `Promise.all` calls (`createPollAnswer`, then `incrementPollChoiceVotes`) are not atomic. If the second batch fails after the first succeeds, the answers and totals diverge. Wrap in a transaction or compensate on failure.

---

app/api/users/[username]/statuses/[statusId]/replies/route.ts
```diff
@@ -0,0 +1,63 @@
...
+        isStatusPubliclyReadable(reply)
+    )
+
+    return activityPubResponse({
+      req,
+      data: {
+        '@context': ACTIVITY_STREAM_URL,
+        id: `${status.id}/replies`,
+        type: 'Collection',
+        totalItems: replies.length,
```
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/replies/route.ts]: `totalItems: replies.length` reports the post-truncation count (capped at 100 by `ACTIVITYPUB_REPLIES_LIMIT`), not the true number of replies. For accurate federation, use a separate count query — or implement `first`/`next` paging like Mastodon does.

---

lib/types/activitypub/collections.ts
```diff
@@ -21,8 +21,16 @@ export const CollectionWithItems = z.object({
 })
 export type CollectionWithItems = z.infer<typeof CollectionWithItems>
 
+export const CollectionSummary = z.object({
+  id: z.string(),
+  type: z.literal('Collection'),
```
llun: `type: z.literal('Collection')` only accepts `Collection`, but `CollectionWithFirstPage` accepts `z.enum(['Collection', 'OrderedCollection'])`. Many implementations (Mastodon included) emit `OrderedCollection` summaries with just `totalItems`. Widen this to the same enum so summary parsing actually catches what we see in the wild.

---

lib/jobs/createPollVoteJob.ts
```diff
@@ -87,7 +88,7 @@ export const createPollVoteJob = createJobHandle(
         choiceIndex
       })
     } catch (error) {
-      console.error('Vote creation failed:', error)
+      logger.error({ error }, 'Vote creation failed')
```
llun, line 79: Switching to the structured logger is good, but the error is still swallowed and the job ends "successful" on failure, so the queue won't retry. Re-throw after logging (or signal failure to the queue) so the dead-letter/retry path can run.

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -28,7 +28,37 @@ import {
...
+  .passthrough()
+const Activity = z.union([
+  Accept,
+  Reject,
+  Follow,
+  Like,
+  Undo,
+  ReferenceUndo,
+  GracefullyAcceptedActivity
+])
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts:61]: `Block`, `Flag`, `Move`, `Add`, `Remove`, `QuoteRequest`, and string-form `Undo` all return 202 with no log. At minimum log at info/warn so operators can see these arrived but were not processed — `Move` in particular is meaningful (account migration), and silently dropping it will surprise users.

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -131,7 +161,17 @@ export const POST = traceApiRoute(
               }
               case 'Undo': {
                 const undoRequest = activity as UndoFollow | UndoLike
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts]: The cast `activity as UndoFollow | UndoLike` is no longer safe once `ReferenceUndo` joined the union — a `ReferenceUndo` whose `object` is `{ type: 'Follow' }` (no inner `object`/`actor`) will reach `undoFollowRequest` after the `typeof === 'string'` short-circuit and likely throw. Add a structural check on `undoObject` before each branch instead of relying on the type narrowing.

---

lib/utils/urlToId.ts
```diff
@@ -1,5 +1,33 @@
...
+
+const encodeBase64Url = (value: string) => {
+  if (typeof Buffer !== 'undefined') {
+    return Buffer.from(value, 'utf8').toString('base64url')
+  }
+
+  return btoa(unescape(encodeURIComponent(value)))
+    .replaceAll('+', '-')
+    .replaceAll('/', '_')
+    .replaceAll('=', '')
```
llun, line 13: `unescape`/`escape` are deprecated and removed/blocked in some runtimes (notably edge runtimes). The Buffer path always wins on Node, but if this code is ever evaluated on the edge runtime (or in modern strict-mode browsers) the fallback will throw. Prefer `TextEncoder`/`TextDecoder` here.

---

lib/utils/urlToId.ts
```diff
@@ -23,6 +56,14 @@ export const urlToId = (idInURLFormat: string) => {
 export const idToUrl = (id: string) => {
   if (!id) return ''
 
+  if (id.startsWith(OPAQUE_URL_ID_PREFIX)) {
+    try {
+      return decodeBase64Url(id.slice(OPAQUE_URL_ID_PREFIX.length))
+    } catch {
+      return id
```
llun: Silent fallback: on decode failure `idToUrl` returns the prefixed id verbatim, which then gets handed to downstream callers as if it were a URL. At minimum log the failure; ideally surface the error so the caller can 404.

---

lib/components/posts/poll.tsx
```diff
@@ -68,8 +68,8 @@ export const Poll: FC<Props> = ({ status, currentTime, currentActorId }) => {
       await votePoll({ statusId: status.id, choices: selectedChoices })
       setVotedChoices(selectedChoices)
       setSelectedChoices([])
-    } catch (error) {
-      console.error('Failed to vote:', error)
+    } catch {
+      setSelectedChoices([])
```
llun: Voting failures are now silently swallowed: the user sees `selectedChoices` cleared with no error message and no log. Either surface a visible error to the user or route through a client logger — losing the `console.error` without replacement is a UX regression.

---

lib/components/follow-request-card/FollowRequestCard.tsx
```diff
@@ -40,8 +40,8 @@ export const FollowRequestCard: FC<Props> = ({
     try {
       await onReject(account.url)
       setStatus('rejected')
-    } catch (error) {
-      console.error('Failed to reject follow request:', error)
+    } catch {
+      setStatus('pending')
```
llun, line 48: Same pattern in both `handleAccept` and `handleReject`: the catch silently flips status back to `pending` with no error message and no log. Users will see the button do nothing. Surface an error toast or at least call a client logger.

---

app/(timeline)/notifications/NotificationsList.tsx
```diff
@@ -55,8 +55,8 @@ export const NotificationsList = ({
         })
         // Refresh the layout to update the notification badge count
         router.refresh()
-      } catch (error) {
-        console.error('Failed to mark notifications as read:', error)
+      } catch {
+        // Keep unread state; the next refresh or retry can mark them later.
```
llun: The catch is now a no-op with a comment. If the mark-as-read request fails repeatedly (auth issues, network, etc.) there's no signal at all in production. Route through a client-side logger instead of dropping the error entirely.

---

app/api/v1/polls/[id]/route.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      choiceIndex: 0
+    })
+    expect(mockSendPollVotes).toHaveBeenCalledWith({
+      currentActor: mockCurrentActor,
+      status: pollStatus,
+      choices: [0]
+    })
+    expect(await response.json()).toEqual(mastodonPoll)
+  })
+})
```
Comment (llun) [inline app/api/v1/polls/[id]/route.test.ts:369]: Coverage is happy-path only. Please add tests for: status not found (404), non-poll status, expired (`Date.now() > endAt`), `hasVoted` rejection on `oneOf`, `oneOf` with `choices.length > 1`, and `getMastodonStatus` returning no `poll` (500). The vote endpoint has the most logic in the PR and currently only one branch is exercised.

---

llun (commented):
Code review covering the 9 commits on this branch. Highlights:

**Correctness**
- `votersCount` overcounts for `anyOf` polls (sum of per-choice totals ≠ unique voters).
- `/likes` exposes liker actor IDs; Mastodon keeps that list private.
- `/replies` `totalItems` reports the truncated count, not the real total.
- `shares.totalItems` is hardcoded to `0` everywhere.
- Poll vote endpoint doesn't bounds-check choice indexes and doesn't gate `anyOf` re-voting.
- The `Undo` cast in the inbox is no longer safe now that `ReferenceUndo` joined the union.
- `CollectionSummary` only accepts `Collection`, not `OrderedCollection`.

**Observability / UX**
- Several client components (poll vote, follow accept/reject, notifications mark-read, add-actor dialog) drop `console.error` without surfacing or logging the failure.
- Inbox silently 202s on `Block`/`Flag`/`Move`/`Add`/`Remove`/`QuoteRequest`/string-form `Undo` with no log.
- `createPollVoteJob` logs the error but still completes successfully, so the queue can't retry.

**Safety**
- `urlToId` browser fallback uses deprecated `escape`/`unescape`; `idToUrl` swallows decode errors silently.
- Poll vote write path is two non-atomic `Promise.all` batches.

**Tests** — happy-path only for polls/replies/featured; no tests for `/likes` or `/shares` at all. Specific gaps called out inline.

Inline comments left on each item — happy to dig deeper on any.

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -147,17 +187,17 @@ export const POST = traceApiRoute(
...
                       allowedMethods: CORS_HEADERS,
-                      data: { target: undoRequest.object.object },
+                      data: { target: undoObject.object },
                       responseStatusCode: 202
                     })
                   }
                   case 'Like': {
                     await database.deleteLike({
-                      actorId: undoRequest.object.actor,
+                      actorId: undoObject.actor,
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Fixed in bcad40c3. `Undo` now parses a full `Like` before deleting and uses the verified outer `activity.actor` as the `deleteLike.actorId`; partial/reference-only Undo objects are accepted as no-ops.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+    const { choices } = parsed.data
+    const hasVoted = await database.hasActorVoted({
+      statusId,
+      actorId: currentActor.id
+    })
+
+    if (
+      (status.pollType === 'oneOf' && hasVoted) ||
+      (status.pollType === 'oneOf' && choices.length > 1)
+    ) {
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:136]: Fixed in bcad40c3. The `hasVoted` check now rejects repeat votes for all poll types; multiple-choice polls still accept multiple indices in the first request.

app/api/users/[username]/statuses/[statusId]/replies/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/replies/route.ts]: Fixed in bcad40c3. The replies collection now uses `getStatusRepliesCount` for `totalItems` instead of the limited page length.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+
+interface Params {
+  id: string
+}
+
+export const POST = traceApiRoute(
+  'voteMastodonPoll',
+  OAuthGuard<Params>([Scope.enum.write], async (req, context) => {
+    const { database, currentActor, params } = context
+    const parsed = VotePollRequest.safeParse(await req.json())
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. The vote route now catches malformed or empty JSON and returns a 400 response instead of surfacing a 500.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (Date.now() > status.endAt) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_422,
+        responseStatusCode: 422
+      })
+    }
+
+    const { choices } = parsed.data
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. Submitted choices are deduplicated with `Set` before validation and persistence.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+    const hasVoted = await database.hasActorVoted({
+      statusId,
+      actorId: currentActor.id
+    })
+
+    if (
+      (status.pollType === 'oneOf' && hasVoted) ||
+      (status.pollType === 'oneOf' && choices.length > 1)
+    ) {
+      return apiResponse({
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:137]: Fixed in bcad40c3. The vote route now requires non-negative integer choices and validates each index against `status.choices.length` before recording votes.

lib/types/domain/status.ts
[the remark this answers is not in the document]
llun: Fixed in bcad40c3. `votersCount` is now emitted only for `oneOf` polls, where summed option totals match unique voters; it is omitted for `anyOf` polls.

app/api/v1/polls/[id]/votes/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:143]: Fixed in bcad40c3. The vote path now deduplicates choices, validates bounds, rejects repeat votes for every poll type, and only applies the single-choice limit to `oneOf` polls.

app/api/users/[username]/statuses/[statusId]/replies/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/replies/route.ts]: Fixed in bcad40c3. `totalItems` for the ActivityPub replies collection now comes from a count query rather than the capped page result.

app/api/v1/polls/[id]/votes/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. JSON parsing is wrapped and invalid JSON returns 400.

lib/types/domain/status.ts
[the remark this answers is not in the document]
llun: Fixed in bcad40c3. `votersCount` is omitted for multiple-choice polls because the database does not expose unique voter count separately from per-option totals.

app/api/v1/polls/[id]/route.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const GET = traceApiRoute(
+  'getMastodonPoll',
+  OAuthGuard<Params>([Scope.enum.read], async (req, context) => {
+    const { database, currentActor, params } = context
+    const encodedPollId = (await params).id
+    const statusId = idToUrl(encodedPollId)
+    const status = await database.getStatus({
+      statusId,
+      withReplies: false
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/route.ts]: Fixed in bcad40c3. Poll reads now use `OptionalOAuthGuard` plus `canActorReadStatus`, so public polls can be read without auth while non-public poll metadata remains protected.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_422,
+        responseStatusCode: 422
+      })
+    }
+
+    await Promise.all(
+      choices.map((choice) =>
+        database.createPollAnswer({
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. The vote route deduplicates submitted choices before any persistence and added coverage for duplicate choice payloads.

lib/types/domain/status.ts
```diff
@@ -267,9 +278,26 @@ export const toActivityPubObject = (status: Status): Note | Question => {
...
 
       published: getISOTimeUTC(status.createdAt),
       endTime: getISOTimeUTC(status.endAt),
+      votersCount: status.choices.reduce(
+        (totalVotes, choice) => totalVotes + choice.totalVotes,
+        0
```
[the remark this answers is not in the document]
llun: Fixed in bcad40c3. Multiple-choice `Question` serialization no longer emits an inflated `votersCount`.

lib/types/domain/status.ts
```diff
@@ -267,9 +278,26 @@ export const toActivityPubObject = (status: Status): Note | Question => {
...
+      ),
+      ...(status.endAt <= Date.now()
+        ? { closed: getISOTimeUTC(status.endAt) }
+        : null),
```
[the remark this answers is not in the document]
llun: Fixed in bcad40c3. The conditional spreads now use `{}` fallbacks for open polls and missing `updatedAt` values.

app/api/users/[username]/inbox/route.ts
```diff
@@ -147,17 +187,17 @@ export const POST = traceApiRoute(
...
                       statusId:
-                        typeof undoRequest.object.object === 'string'
-                          ? undoRequest.object.object
-                          : undoRequest.object.object.id
+                        typeof undoObject.object === 'string'
+                          ? undoObject.object
+                          : undoObject.object.id
                     })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Fixed in bcad40c3. Reference-only `Undo` objects are handled as no-ops unless they match the full `Follow` or `Like` schemas; full Undo Like handling now uses the verified outer actor.

app/api/v1/polls/[id]/votes/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:34]: Fixed in bcad40c3. `choices` are constrained to non-negative integers, deduplicated, and bounds-checked before the transactional vote write.

app/api/v1/polls/[id]/votes/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. Repeat voting is rejected regardless of poll type; `anyOf` support is limited to submitting multiple choices in the first vote request.

app/api/v1/polls/[id]/votes/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. `req.json()` is now wrapped and parse failures return 400.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,133 @@
...
+          actorId: currentActor.id,
+          choice
+        })
+      )
+    )
+    await Promise.all(
+      choices.map((choiceIndex) =>
+        database.incrementPollChoiceVotes({ statusId, choiceIndex })
+      )
+    )
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in bcad40c3. Vote answer insertion and option-total increments now go through `recordPollVotes`, which performs the existing-vote check, answer inserts, and count increments in one database transaction.

llun:
@codex review

lib/types/domain/status.ts
llun: Fixed in bcad40c3: ActivityPub anyOf polls now omit votersCount because distinct voter count cannot be derived from per-choice totals; oneOf still emits the summed total.

lib/types/domain/status.ts
llun: Fixed in deaa6fc5 and 26274bb6: the /shares collection route now returns the reblog counter, and serialized Note/Question objects use database-backed totalShares instead of hardcoded zero.

app/api/users/[username]/statuses/[statusId]/likes/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/likes/route.ts:41]: Fixed in deaa6fc5: the ActivityPub likes collection now returns only totalItems and no liker actor list, matching Mastodon privacy behavior.

app/api/v1/polls/[id]/votes/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:34]: Fixed in bcad40c3: poll vote choices are validated as integer, non-negative, unique indexes and then bounded against the poll choice count before writes.

app/api/v1/polls/[id]/votes/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:143]: Fixed in 14e4dba0: poll votes now claim a unique poll_voters marker for (statusId, actorId) inside the transaction, so anyOf duplicate votes are rejected atomically.

app/api/v1/polls/[id]/votes/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:167]: Fixed in bcad40c3 and hardened in 14e4dba0: vote answer insertion and choice counter updates run in a transaction, with a unique poll_voters marker preventing duplicate-vote races before counters are incremented.

app/api/users/[username]/statuses/[statusId]/replies/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/replies/route.ts]: Fixed in bcad40c3: replies totalItems now uses a separate reply count instead of the capped page length.

lib/types/activitypub/collections.ts
llun: Fixed in deaa6fc5: collection schemas now accept both Collection and OrderedCollection, including summary collections with totalItems only.

lib/jobs/createPollVoteJob.ts
llun: Fixed in deaa6fc5: the poll vote job logs the failure and rethrows so queue retry/dead-letter behavior can run.

app/api/users/[username]/inbox/route.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.ts:61]: Fixed in deaa6fc5: accepted-but-unimplemented inbox activity types now emit structured info logs before returning 202.

app/api/users/[username]/inbox/route.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Fixed in bcad40c3: Undo handling no longer casts reference-only Undo as follow/like undo data; it safely accepts reference-only forms without dereferencing missing inner fields.

lib/utils/urlToId.ts
llun: Fixed in deaa6fc5: the browser/runtime fallback now uses TextEncoder/TextDecoder instead of deprecated escape/unescape.

lib/utils/urlToId.ts
llun: Fixed in deaa6fc5: invalid opaque REST ids no longer fall through as prefixed ids; decode failure returns an empty id so callers treat it as not found.

lib/components/posts/poll.tsx
llun: Fixed in deaa6fc5: poll vote failures now preserve selection state and show a visible error message to the user.

lib/components/follow-request-card/FollowRequestCard.tsx
llun: Fixed in deaa6fc5: follow request accept/reject failures now restore pending state and surface a visible error message.

app/(timeline)/notifications/NotificationsList.tsx
llun: Fixed in deaa6fc5: notification mark-as-read failures now set a visible error state instead of silently dropping the failure.

app/api/v1/polls/[id]/route.test.ts
Comment (llun) [reply app/api/v1/polls/[id]/route.test.ts:369]: Fixed in deaa6fc5: vote route tests now cover missing status, non-poll status, expired poll, repeated oneOf vote, invalid multi-choice oneOf vote, invalid choice bounds, and missing Mastodon poll response.

lib/database/sql/status.ts
```diff
@@ -1264,9 +1265,86 @@ export const StatusSQLDatabaseMixin = (
...
+    choices
+  }: RecordPollVotesParams): Promise<boolean> {
+    if (choices.length === 0) return false
+
+    const currentTime = new Date()
+    return database.transaction(async (trx) => {
+      const existingVote = await trx('poll_answers')
+        .where({ statusId, actorId })
+        .first()
+      if (existingVote) return false
```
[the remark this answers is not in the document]
llun: Fixed in 14e4dba0: recordPollVotes uses a poll_voters table with a unique (statusId, actorId) marker claimed inside the transaction before inserting answers or incrementing totals.

lib/types/activitypub/objects.ts
```diff
@@ -114,6 +114,8 @@ export const BaseContent = z.object({
     ])
     .nullish(),
   replies: Collection.nullish(),
+  likes: Collection.nullish(),
+  shares: Collection.nullish(),
```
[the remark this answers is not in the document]
llun: Fixed in deaa6fc5: BaseContent likes/shares validation accepts OrderedCollection as well as Collection and includes coverage for OrderedCollection summaries.

app/api/users/[username]/statuses/[statusId]/shares/route.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (status.type === StatusType.enum.Announce) return apiErrorResponse(404)
+
+    return activityPubResponse({
+      req,
+      data: {
+        '@context': ACTIVITY_STREAM_URL,
+        id: `${status.id}/shares`,
+        type: 'Collection',
+        totalItems: 0,
+        items: []
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/shares/route.ts]: Fixed in deaa6fc5 and 26274bb6: the shares route returns real reblog totals, and ActivityPub Note/Question serialization now carries database-backed share totals.

llun:
@codex review

lib/jobs/createPollVoteJob.ts
```diff
@@ -87,7 +88,8 @@ export const createPollVoteJob = createJobHandle(
         choiceIndex
       })
```
llun, line 76: The local Mastodon vote route is now atomic via `recordPollVotes`, but the federated vote path here still uses `createPollAnswer` + a separate `incrementPollChoiceVotes` — exactly the non-atomic two-step the previous review flagged. It also doesn't insert into the new `poll_voters` table, so the unique constraint that backstops one-vote-per-actor doesn't apply to incoming federation votes. Please switch this path to `recordPollVotes` so federated and local votes share the same invariant.

---

migrations/20260502120000_add_poll_voters_table.js
```diff
@@ -0,0 +1,40 @@
...
+  const now = new Date()
+  await knex('poll_voters').insert(
+    existingVotes.map((vote) => ({
+      statusId: vote.statusId,
+      actorId: vote.actorId,
+      createdAt: now,
+      updatedAt: now
+    }))
+  )
+}
```
llun, line 60: `existingVotes` loads every distinct `(statusId, actorId)` pair into JS memory and inserts in one batch. On larger installs this can OOM and on SQLite it can hit `SQLITE_MAX_COMPOUND_SELECT` (~32k by default) or driver parameter limits. Either chunk the insert (e.g. 500 rows at a time) or do it server-side with `INSERT INTO poll_voters (statusId, actorId, createdAt, updatedAt) SELECT DISTINCT statusId, actorId, ?, ? FROM poll_answers WHERE statusId IS NOT NULL`.

---

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,98 @@ export const StatusSQLDatabaseMixin = (
...
+            createdAt: currentTime,
+            updatedAt: currentTime
+          }))
+        )
+
+        await Promise.all(
+          selectedChoices.map((choice) =>
+            trx('poll_choices')
+              .where({ statusId, choiceId: choice.choiceId })
+              .increment('totalVotes', 1)
```
llun: The new uniqueness invariant lives on `poll_voters`, but the early-existence check reads `poll_answers`. That's inconsistent: a concurrent insert that already wrote `poll_voters` (but not yet `poll_answers`) will slip past this read and fall into the unique-constraint catch — which works, but the read itself is dead weight (and the wrong table). Either query `poll_voters` here or drop this read entirely and rely on the constraint.

---

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,170 @@
...
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_422,
+        responseStatusCode: 422
+      })
+    }
+
+    const choices = [...new Set(parsed.data.choices)]
+    const hasVoted = await database.hasActorVoted({
+      statusId,
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts]: With `recordPollVotes` enforcing `(statusId, actorId)` uniqueness via the `poll_voters` constraint, this `hasActorVoted` round trip is redundant and adds a TOCTOU window. Drop it (or push the dedupe/bounds check entirely into `recordPollVotes`) so there's a single source of truth for vote-eligibility.

---

app/api/users/[username]/statuses/[statusId]/replies/route.ts
```diff
@@ -0,0 +1,70 @@
...
+      database.getStatusRepliesCount({
+        statusId: status.id,
+        url: status.url,
+        publicOnly: true
+      })
+    ])
+
+    const publicReplies = replies.filter(
+      (reply): reply is StatusNote | StatusPoll =>
+        reply.type !== StatusType.enum.Announce &&
```
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/replies/route.ts:57]: `getStatusReplies` was already called with `publicOnly: true` (DB-side filter), then this re-filters with `isStatusPubliclyReadable`. If the DB layer is correct the second pass is dead code; if you want it as defense-in-depth, please add a comment explaining why — otherwise a future reader will assume `publicOnly: true` is broken and "fix" one of the two.

---

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,98 @@ export const StatusSQLDatabaseMixin = (
...
+      )
+    }
+
+    const result = await query
+      .whereNot('type', StatusType.enum.Announce)
+      .count<{ count: string }>('* as count')
+      .first()
+
+    return parseInt(String(result?.count ?? '0'), 10)
+  }
```
llun, line 1320: When `publicOnly` is set this now does a full `COUNT(*)` joined against `recipients` on every `/replies` GET, instead of reading the counter. Correctness wins, but for popular statuses this can be hot. Worth either an index on `recipients(actorId, statusId)` (if not already covered) and `statuses(reply, type)`, or maintaining a public-only reply counter alongside the existing total.

---

lib/types/domain/status.ts
```diff
@@ -267,12 +279,31 @@ export const toActivityPubObject = (status: Status): Note | Question => {
...
       },
+      likes: {
+        id: `${status.id}/likes`,
+        type: 'Collection',
+        totalItems: status.totalLikes
+      },
+      shares: {
+        id: `${status.id}/shares`,
+        type: 'Collection',
+        totalItems: status.totalShares ?? 0
```
llun: `totalShares` is omitted when 0 in the SQL mixin, then defaulted to `0` here. That works, but it conflates "0 known shares" with "shares unknown" for any future consumer of the domain `Status` type. Suggest emitting `totalShares` unconditionally from the mixin (parallel to `totalLikes`), and dropping `?? 0` here.

---

app/api/v1/polls/[id]/route.ts
```diff
@@ -0,0 +1,79 @@
...
+      database,
+      status,
+      currentActor
+    })
+    if (!hasAccess) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_404,
+        responseStatusCode: 404
```
Comment (llun) [inline app/api/v1/polls/[id]/route.ts:56]: `getStatus` already takes `currentActorId` and presumably filters by visibility, then `canActorReadStatus` re-checks. Either layer is fine on its own, but having both is confusing — please add a one-liner clarifying which is authoritative, so a future change doesn't accidentally remove the wrong one.

---

lib/utils/urlToId.ts
```diff
@@ -23,6 +58,17 @@ export const urlToId = (idInURLFormat: string) => {
...
   if (!id) return ''
 
+  if (id.startsWith(OPAQUE_URL_ID_PREFIX)) {
+    try {
+      const decoded = decodeBase64Url(id.slice(OPAQUE_URL_ID_PREFIX.length))
+      const decodedUrl = new URL(decoded)
+      if (!['http:', 'https:'].includes(decodedUrl.protocol)) return ''
+      return decoded
+    } catch {
+      return ''
```
llun, line 68: Returning `''` on decode failure is better than the previous silent prefix passthrough, but downstream callers will then see "no such status" with no signal that an id was actually corrupt. Consider `logger.warn` here so operators can spot bad ids in dashboards.

---

llun (commented):
Re-review of the 4 follow-up commits — most prior feedback is addressed (privacy on `/likes`, real `totalItems` on `/replies` and `/shares`, `votersCount` only for `oneOf`, atomic `recordPollVotes` with `poll_voters` table + unique constraint, JSON parse guard, choice dedupe + bounds check, `OptionalOAuthGuard` + `canActorReadStatus` for read access, structural Undo parsing with verified outer actor, inbox logging for accepted-no-side-effects activities, `OrderedCollection` summaries, `TextEncoder`/`TextDecoder` in `urlToId`, `throw error` in `createPollVoteJob`, error UI in `Poll`/`FollowRequestCard`/`NotificationsList`, expanded test coverage). Nice cleanup.

Remaining items, mostly tightening:

**Bigger** — the federated vote path in `lib/jobs/createPollVoteJob.ts` still uses the old non-atomic `createPollAnswer` + `incrementPollChoiceVotes` and only gates `hasVoted` for `oneOf`. The local route uses `recordPollVotes` and gates both poll types — please bring the federated path onto the same primitive so the `poll_voters` unique-constraint actually backstops federation too.

**Smaller** — migration backfill is unbatched, `recordPollVotes` checks `poll_answers` instead of `poll_voters`, the route's `hasActorVoted` pre-check is now redundant with the unique constraint, replies route double-filters public visibility, `getStatusRepliesCount` does a full count on every public `/replies` request, `totalShares: undefined` vs `0` semantics, GET poll route runs both `getStatus(currentActorId)` and `canActorReadStatus` without saying which is authoritative, and `idToUrl` now silently returns `''` on decode failure with no log.

Inline comments on each.

---

llun:
Follow-up to my review (one inline comment failed because the line wasn't in the diff):

**`lib/jobs/createPollVoteJob.ts:75-77`** — the federated vote path still gates only `pollType === 'oneOf'` for `hasVoted`. The local route was tightened to gate both poll types; please mirror that here (or, better, route this path through `recordPollVotes` so the `poll_voters` unique constraint is the single source of truth for both local and federated votes).

---

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,98 @@ export const StatusSQLDatabaseMixin = (
...
+
+  async function recordPollVotes({
+    statusId,
+    actorId,
+    choices
+  }: RecordPollVotesParams): Promise<boolean> {
```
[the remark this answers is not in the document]
llun: Fixed in 78d3a833: recordPollVotes now deduplicates choices with a Set before validation, answer inserts, and counter increments; SQL coverage now submits [0, 0, 1] and verifies each choice is counted once.

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,98 @@ export const StatusSQLDatabaseMixin = (
...
+    if (choices.length === 0) return false
+
+    const currentTime = new Date()
+    try {
+      return await database.transaction(async (trx) => {
+        const existingVote = await trx('poll_answers')
+          .where({ statusId, actorId })
+          .first()
```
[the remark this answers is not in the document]
llun: Fixed in 78d3a833: the stale poll_answers pre-check was removed entirely, so recordPollVotes now relies on the poll_voters unique constraint as the single duplicate-vote gate inside the transaction.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,170 @@
...
+  id: string
+}
+
+export const POST = traceApiRoute(
+  'voteMastodonPoll',
+  OAuthGuard<Params>([Scope.enum.write], async (req, context) => {
+    const { database, currentActor, params } = context
+    let body: unknown
+    try {
+      body = await req.json()
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in 78d3a833: POST /api/v1/polls/:id/votes now accepts JSON, application/x-www-form-urlencoded, and multipart form bodies with Mastodon-style choices[] values; route coverage includes form-encoded votes.

lib/jobs/createPollVoteJob.ts
llun: Fixed in 78d3a833: createPollVoteJob now routes federated poll votes through recordPollVotes, so incoming federation votes use the same poll_voters uniqueness and atomic answer/counter transaction as local Mastodon votes.

migrations/20260502120000_add_poll_voters_table.js
llun: Fixed in 78d3a833: the poll_voters backfill now reads distinct voter pairs in 500-row batches, filters null actor ids, and inserts each batch separately to avoid one large in-memory insert.

lib/database/sql/status.ts
llun: Fixed in 78d3a833: the early poll_answers lookup was dropped; recordPollVotes validates choices, claims poll_voters, then writes answers/counters in one transaction, returning false on the unique-constraint duplicate path.

app/api/v1/polls/[id]/votes/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in 78d3a833: the Mastodon vote route no longer calls hasActorVoted before writing; it delegates duplicate detection to recordPollVotes. I also moved the older local account vote endpoint onto recordPollVotes so UI votes use the same invariant.

app/api/users/[username]/statuses/[statusId]/replies/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/replies/route.ts:57]: Fixed in 78d3a833: added a one-line comment explaining that the route keeps the object-level readability check as defense-in-depth after the database publicOnly filter.

lib/database/sql/status.ts
llun: Fixed in 78d3a833: added a follow-up migration with recipients(actorId, statusId) and statuses(reply, type) indexes for the public reply count/read path.

lib/types/domain/status.ts
llun: Fixed in 78d3a833: SQL status hydration now emits totalShares unconditionally, ActivityPub serialization uses totalShares directly, and tests/fixtures include the explicit zero value.

app/api/v1/polls/[id]/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/route.ts:56]: Fixed in 78d3a833: added a route comment clarifying that getStatus loads by id and canActorReadStatus is the authoritative visibility gate.

lib/utils/urlToId.ts
llun: No code change for this one: idToUrl is imported by client code through lib/client.ts and React components, so importing the server logger here would violate the project boundary against logger usage in client-side code and risks bundling pino into client paths. The empty-string return intentionally makes corrupt opaque ids resolve as not found; server-route telemetry for invalid ids should be added at route boundaries if needed.

llun:
Addressed the top-level follow-up on lib/jobs/createPollVoteJob.ts:75-77 in 78d3a833. The federated vote job now calls recordPollVotes directly, so it no longer has the oneOf-only hasVoted gate or the old createPollAnswer/incrementPollChoiceVotes split.

llun:
@codex review

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,189 @@
...
+
+const parseFormChoices = (formData: Pick<FormData, 'getAll'>) => {
+  const choices = [
+    ...formData.getAll('choices[]'),
+    ...formData.getAll('choices')
+  ]
+  return {
+    choices: choices.map((choice) =>
+      typeof choice === 'string' && choice.trim() ? Number(choice) : Number.NaN
+    )
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:43]: `parseFormChoices` concatenates `formData.getAll('choices[]')` and `formData.getAll('choices')`. If a Mastodon client ever sends both keys (some libraries do, depending on encoder), the arrays merge into one. The dedupe further down hides the symptom but it can also silently coerce intent (e.g. `{choices: '0', 'choices[]': ['1','2']}` becomes `[0,1,2]`). Suggest preferring `choices[]` if present, otherwise fall back to `choices`, rather than merging.

---

app/api/v1/accounts/vote/route.ts
```diff
@@ -20,8 +21,30 @@ export const POST = traceApiRoute(
   'voteToAccount',
   AuthenticatedGuard(async (req, context) => {
     const { database, currentActor } = context
-    const body = await req.json()
-    const { statusId, choices } = VotePollRequest.parse(body)
+    let body: unknown
+    try {
+      body = await req.json()
```
llun, line 28: The public `/api/v1/polls/[id]/votes` endpoint now accepts `application/x-www-form-urlencoded` and `multipart/form-data`, but this internal `/api/v1/accounts/vote` route still only handles JSON. If the intent is Mastodon-style parity at both entry points, extract `parseVotePollRequestBody` into a shared helper and use it here too — otherwise leave a comment that this endpoint is JSON-only by design.

---

migrations/20260502120000_add_poll_voters_table.js
```diff
@@ -0,0 +1,52 @@
...
+    await knex('poll_voters').insert(
+      existingVotes.map((vote) => ({
+        statusId: vote.statusId,
+        actorId: vote.actorId,
+        createdAt: now,
+        updatedAt: now
+      }))
+    )
+
+    offset += existingVotes.length
```
llun: `LIMIT/OFFSET` over a `DISTINCT` query rescans `O(offset)` rows per batch, which can get expensive on a large `poll_answers` table. Acceptable for a one-shot migration, but if you want it cheaper consider a keyset cursor on `(statusId, actorId)` — or just do `INSERT INTO poll_voters (statusId, actorId, createdAt, updatedAt) SELECT DISTINCT statusId, actorId, ?, ? FROM poll_answers WHERE statusId IS NOT NULL AND actorId IS NOT NULL` and let the DB do it server-side.

---

migrations/20260502123000_add_public_reply_lookup_indexes.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async (knex) => {
+  await knex.schema.alterTable('recipients', function (table) {
+    table.index(['actorId', 'statusId'], 'recipients_actorId_statusId_idx')
```
llun, line 7: Worth confirming `recipients` doesn't already carry an index leading with `actorId` — if it does, this composite duplicates the leading column for very little gain. A quick `\d recipients` (or whatever the equivalent is for the project's primary DB) before merging will save you index bloat.

---

llun (commented):
Round 3 — all the meaningful follow-ups are addressed:

- Federated vote path now goes through `recordPollVotes` (atomic + `poll_voters` constraint) and the `oneOf`-only `hasVoted` gate is gone.
- Local route's redundant `hasActorVoted` check removed; constraint is the single source of truth.
- `recordPollVotes` no longer reads `poll_answers` for existence, and dedupes choices internally.
- `totalShares` is now `z.number().default(0)`; no more `?? 0` ambiguity.
- Migration backfill is batched (500-row chunks) with deterministic ordering and null filtering.
- New `recipients(actorId, statusId)` and `statuses(reply, type)` indexes added for `getStatusRepliesCount`.
- Comments added explaining the visibility check duplication in `/replies` and `/polls/[id]`.
- Bonus: `/polls/[id]/votes` now accepts form-encoded and multipart bodies, plus a parity refactor of `/accounts/vote`.
- Tests updated accordingly (form-encoded poll vote, dedupe at DB layer, `recordPollVotes` rejects duplicate voters in `anyOf`, queue rethrow via `recordPollVotes`).

Remaining nits (non-blocking) inline:
- `parseFormChoices` merges both `choices[]` and `choices` keys instead of preferring one.
- `/api/v1/accounts/vote` is still JSON-only while `/polls/[id]/votes` accepts form bodies — consider sharing the parser or commenting on the asymmetry.
- Migration backfill uses LIMIT/OFFSET on a DISTINCT query (rescans on each batch); a server-side `INSERT … SELECT DISTINCT` would be cheaper.
- New composite index on `recipients(actorId, statusId)` may overlap with an existing leading-`actorId` index — worth checking before merge.

Looks good overall.

---

app/api/v1/polls/[id]/votes/route.ts
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts:43]: Fixed in 119bfcf4: parseFormChoices now prefers choices[] when present and falls back to choices only when choices[] is absent, so mixed encoders cannot merge both key styles.

app/api/v1/accounts/vote/route.ts
llun: No behavior change needed; 119bfcf4 adds a clarifying comment. /api/v1/accounts/vote is the internal JSON-only UI endpoint; Mastodon-compatible form submissions are handled by /api/v1/polls/:id/votes.

migrations/20260502120000_add_poll_voters_table.js
llun: Fixed in 119bfcf4: the poll_voters backfill now uses a keyset cursor over (statusId, actorId) instead of LIMIT/OFFSET, while still inserting in 500-row batches.

migrations/20260502123000_add_public_reply_lookup_indexes.js
llun: Verified/no code change: existing recipients indexes lead with statusId or type, not actorId. The current public reply query filters by actorId without a type predicate, so recipients(actorId, statusId) is not duplicating an actorId-leading index.

app/api/v1/accounts/vote/route.ts
```diff
@@ -20,8 +21,30 @@ export const POST = traceApiRoute(
...
+        data: ERROR_422,
+        responseStatusCode: 422
+      })
+    }
+
+    const { statusId } = parsed.data
+    const choices = [...new Set(parsed.data.choices)]
 
     const status = await database.getStatus({ statusId, withReplies: false })
     if (!status || status.type !== StatusType.enum.Poll) {
```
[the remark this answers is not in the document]
llun: No code change: this route already rejects expired polls before choice validation/writes via Date.now() > status.endAt; see the existing check immediately after the poll status is loaded.

app/api/v1/accounts/vote/types.ts
```diff
@@ -1,6 +1,6 @@
 import { z } from 'zod'
 
 export const VotePollRequest = z.object({
-  statusId: z.string(),
-  choices: z.number().array().min(1)
+  statusId: z.string().min(1),
+  choices: z.number().int().nonnegative().array().min(1)
```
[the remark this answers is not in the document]
llun: Fixed in 119bfcf4: the internal vote request schema now caps choices at 20.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,189 @@
...
+} from '@/lib/utils/response'
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.POST]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+const VotePollRequest = z.object({
+  choices: z.number().int().nonnegative().array().min(1)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/polls/[id]/votes/route.ts]: Fixed in 119bfcf4: the Mastodon poll vote request schema now caps choices at 20.

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,94 @@ export const StatusSQLDatabaseMixin = (
...
+  }: RecordPollVotesParams): Promise<boolean> {
+    const uniqueChoices = [...new Set(choices)]
+    if (uniqueChoices.length === 0) return false
+
+    const currentTime = new Date()
+    try {
+      return await database.transaction(async (trx) => {
+        const pollChoices = await trx('poll_choices')
+          .where({ statusId })
+          .orderBy('choiceId', 'asc')
```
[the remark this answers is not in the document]
llun: No code change: choiceId is a global autoincrement primary key on poll_choices, so setting it to the per-poll option index would collide across polls. createPoll relies on DB-assigned monotonic ids for insertion order; changing that column would need a separate schema migration to make ordering per-status.

lib/jobs/createPollVoteJob.ts
```diff
@@ -66,28 +67,16 @@ export const createPollVoteJob = createJobHandle(
...
         statusId: pollStatus.id,
         actorId: note.attributedTo,
-        choice: choiceIndex
-      })
-
-      await database.incrementPollChoiceVotes({
-        statusId: pollStatus.id,
-        choiceIndex
+        choices: [choiceIndex]
       })
```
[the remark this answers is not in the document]
llun: Fixed in 119bfcf4: federated anyOf poll votes now call recordPollVotes with allowAdditionalChoices, so separate Create activities from the same actor can append distinct choices while duplicate choice activities are ignored without inflating totals.

llun:
@codex review

app/api/v1/accounts/vote/route.ts
```diff
@@ -42,12 +67,13 @@ export const POST = traceApiRoute(
       })
     }
```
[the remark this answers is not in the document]
llun: No code change: the internal route already has the expiration guard at app/api/v1/accounts/vote/route.ts:61-68, returning 422 when Date.now() > status.endAt before any vote write or federation send.

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,119 @@ export const StatusSQLDatabaseMixin = (
...
+          }))
+        )
+
+        await Promise.all(
+          newSelectedChoices.map((choice) =>
+            trx('poll_choices')
+              .where({ statusId, choiceId: choice.choiceId })
+              .increment('totalVotes', 1)
+          )
+        )
```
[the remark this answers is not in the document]
llun: Fixed in 2cc252bd: poll choice counters now update in a single whereIn increment instead of issuing one update per selected choice.

llun:
@codex review

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,119 @@ export const StatusSQLDatabaseMixin = (
...
+          (choice) => !existingChoices.has(choice.choiceIndex)
+        )
+        if (newSelectedChoices.length === 0) return false
+
+        if (!existingVote) {
+          await trx('poll_voters').insert({
+            statusId,
+            actorId,
+            createdAt: currentTime,
+            updatedAt: currentTime
```
llun, line 1372: Possible race for federated `anyOf` votes from the same actor: two jobs land concurrently with `existingVote === undefined`, both reach the `poll_voters` insert, the second hits the unique constraint and the whole transaction rolls back — losing those choices. With queue serialization per poll/actor this is unlikely, but worth either using an `INSERT … ON CONFLICT DO NOTHING` for the `poll_voters` row (and re-reading) or adding a comment that the queue guarantees serialization.

---

lib/database/sql/status.ts
```diff
@@ -1264,9 +1284,119 @@ export const StatusSQLDatabaseMixin = (
...
+      .count<{ count: string }>('* as count')
+      .first()
+
+    return parseInt(String(result?.count ?? '0'), 10)
+  }
+
+  async function recordPollVotes({
+    statusId,
+    actorId,
+    choices,
```
llun, line 1325: `recordPollVotes` honors `allowAdditionalChoices` purely based on the caller's word — if some future caller passes `allowAdditionalChoices: true` against a `oneOf` poll it'll happily append choices. Cheap defense: load the poll type inside the transaction and reject the flag for non-`anyOf` polls. Or document it explicitly: "callers MUST only set this for anyOf polls."

---

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -0,0 +1,195 @@
...
+  ERROR_422,
+  ERROR_500,
+  apiResponse,
+  defaultOptions
+} from '@/lib/utils/response'
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.POST]
+const MAX_POLL_CHOICES_PER_VOTE = 20
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:22]: `MAX_POLL_CHOICES_PER_VOTE = 20` is now duplicated in `app/api/v1/accounts/vote/types.ts` and `app/api/v1/polls/[id]/votes/route.ts`. Worth lifting to a shared module so the two endpoints can't drift.

---

llun (commented):
Round 4 — the two new commits address the previous nits and add a real fix for federated multi-choice voting. Looks ready to me.

Resolved:
- `parseFormChoices` now prefers `choices[]` and falls back to `choices` instead of merging both — covered by the updated form-encoded test.
- `/api/v1/accounts/vote` carries a comment explaining its JSON-only stance.
- Migration backfill switched from LIMIT/OFFSET to keyset pagination on `(statusId, actorId)` — no more `O(offset)` rescans.
- `recordPollVotes` now reads `poll_voters` for the existence check (the right table for the new constraint).
- New `MAX_POLL_CHOICES_PER_VOTE = 20` cap prevents abusive choice-array sizes; tests cover the 422 path.
- Counter increments are now a single batch `UPDATE … WHERE choiceId IN (…)` instead of N concurrent statements — nice.
- New `allowAdditionalChoices` flag on `recordPollVotes` makes federated `anyOf` votes append correctly: each Create activity carries one choice, and remote senders may emit several. Tests cover both the additive case and the duplicate-as-no-op case.

Remaining nits (non-blocking) inline:
- Concurrent `anyOf` Create jobs from the same actor could race on `poll_voters` insert; either rely on per-poll job serialization (and comment that), or use `ON CONFLICT DO NOTHING`.
- `recordPollVotes` trusts the `allowAdditionalChoices` flag without verifying `pollType === 'anyOf'` inside the transaction; either validate or document the contract.
- `MAX_POLL_CHOICES_PER_VOTE` is now duplicated in `accounts/vote/types.ts` and `polls/[id]/votes/route.ts` — worth extracting.

Otherwise LGTM. Glad to have helped iterate on this.

---

PR #633, opened 2026-05-03
feat: add actor route heatmap recreation script

llun:
## Summary
- Add a reusable route heatmap recreation helper that enumerates every discoverable all-time, yearly, and monthly variant for an actor.
- Mirror known cached region variants, soft-delete existing route heatmap rows, and publish fresh unique generation jobs so failed prior attempts do not block recreation.
- Add `scripts/recreateFitnessRouteHeatmaps.ts` with `--actor-id` and `--dry-run` support.

## Validation
- `node -v && yarn run prettier --write .`
- `node -v && yarn lint` (0 errors; existing warnings remain)
- `node -v && yarn build`
- `node -v && yarn test`

### DIFF
--- a/lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
+++ b/lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
@@ -1,0 +1,167 @@
+import { GENERATE_FITNESS_ROUTE_HEATMAP_JOB_NAME } from '@/lib/jobs/names'
+import { recreateFitnessRouteHeatmapJobs } from '@/lib/jobs/recreateFitnessRouteHeatmapJobs'
+import { FitnessFile } from '@/lib/types/database/fitnessFile'
+
+const mockPublish = jest.fn()
+
+jest.mock('@/lib/services/queue', () => ({
+  getQueue: jest.fn(() => ({
+    publish: mockPublish
+  }))
+}))
+
+const makeFitnessFile = (
+  overrides: Partial<FitnessFile> = {}
+): FitnessFile => ({
+  id: overrides.id ?? 'fitness-file-1',
+  actorId: overrides.actorId ?? 'actor-1',
+  path: overrides.path ?? 'fitness-file.fit',
+  fileName: overrides.fileName ?? 'fitness-file.fit',
+  fileType: overrides.fileType ?? 'fit',
+  mimeType: overrides.mimeType ?? 'application/octet-stream',
+  bytes: overrides.bytes ?? 1024,
+  processingStatus: overrides.processingStatus ?? 'completed',
+  isPrimary: overrides.isPrimary ?? true,
+  activityType: overrides.activityType,
+  activityStartTime: overrides.activityStartTime,
+  createdAt: overrides.createdAt ?? Date.UTC(2026, 0, 1),
+  updatedAt: overrides.updatedAt ?? Date.UTC(2026, 0, 1)
+})
+
+describe('recreateFitnessRouteHeatmapJobs', () => {
+  beforeEach(() => {
+    mockPublish.mockClear()
+  })
+
+  it('deletes existing heatmaps and queues every discoverable actor variant', async () => {
+    const getFitnessFilesByActor = jest
+      .fn()
+      .mockResolvedValueOnce([
+        makeFitnessFile({
+          id: 'run-2026-01',
+          activityType: 'run',
+          activityStartTime: Date.UTC(2026, 0, 15, 8)
+        }),
+        makeFitnessFile({
+          id: 'cycle-2026-02',
+          activityType: 'cycle',
+          activityStartTime: Date.UTC(2026, 1, 3, 10)
+        })
+      ])
+      .mockResolvedValueOnce([])
+    const database = {
+      getDistinctRouteHeatmapRegionsForActor: jest
+        .fn()
+        .mockResolvedValue(['encoded-region']),
+      getFitnessFilesByActor,
+      deleteFitnessRouteHeatmapsForActor: jest.fn().mockResolvedValue(3)
+    }
+
+    const result = await recreateFitnessRouteHeatmapJobs({
+      database,
+      actorId: 'actor-1',
+      runId: 'test-run'
+    })
+
+    expect(
+      database.getDistinctRouteHeatmapRegionsForActor
+    ).toHaveBeenCalledWith({
+      actorId: 'actor-1'
+    })
+    expect(getFitnessFilesByActor).toHaveBeenNthCalledWith(1, {
+      actorId: 'actor-1',
+      processingStatus: 'completed',
+      isPrimary: true,
+      limit: 500,
+      offset: 0
+    })
+    expect(database.deleteFitnessRouteHeatmapsForActor).toHaveBeenCalledWith({
+      actorId: 'actor-1'
+    })
+
+    expect(result.deletedCount).toBe(3)
+    expect(result.queuedCount).toBe(20)
+    expect(result.failedCount).toBe(0)
+    expect(result.variants).toEqual(
+      expect.arrayContaining([
+        {
+          activityType: null,
+          periodType: 'all_time',
+          periodKey: 'all'
+        },
+        {
+          activityType: null,
+          periodType: 'yearly',
+          periodKey: '2026'
+        },
+        {
+          activityType: null,
+          periodType: 'monthly',
+          periodKey: '2026-01'
+        },
+        {
+          activityType: 'run',
+          periodType: 'monthly',
+          periodKey: '2026-01'
+        },
+        {
+          activityType: 'cycle',
+          periodType: 'monthly',
+          periodKey: '2026-02',
+          region: 'encoded-region'
+        }
+      ])
+    )
+    expect(mockPublish).toHaveBeenCalledTimes(20)
+    expect(mockPublish).toHaveBeenCalledWith(
+      expect.objectContaining({
+        name: GENERATE_FITNESS_ROUTE_HEATMAP_JOB_NAME,
+        data: {
+          actorId: 'actor-1',
+          activityType: 'run',
+          periodType: 'monthly',
+          periodKey: '2026-01'
+        }
+      })
+    )
+    expect(mockPublish).toHaveBeenCalledWith(
+      expect.objectContaining({
+        name: GENERATE_FITNESS_ROUTE_HEATMAP_JOB_NAME,
+        data: {
+          actorId: 'actor-1',
+          activityType: 'cycle',
+          periodType: 'monthly',
+          periodKey: '2026-02',
+          region: 'encoded-region'
+        }
+      })
+    )
+  })
+
+  it('skips deletion and publishing in dry-run mode', async () => {
+    const database = {
+      getDistinctRouteHeatmapRegionsForActor: jest.fn().mockResolvedValue([]),
+      getFitnessFilesByActor: jest
+        .fn()
+        .mockResolvedValueOnce([
+          makeFitnessFile({
+            activityStartTime: Date.UTC(2026, 4, 1)
+          })
+        ])
+        .mockResolvedValueOnce([]),
+      deleteFitnessRouteHeatmapsForActor: jest.fn()
+    }
+
+    const result = await recreateFitnessRouteHeatmapJobs({
+      database,
+      actorId: 'actor-1',
+      dryRun: true
+    })
+
+    expect(result.deletedCount).toBe(0)
+    expect(result.queuedCount).toBe(0)
+    expect(result.variants).toHaveLength(3)
+    expect(database.deleteFitnessRouteHeatmapsForActor).not.toHaveBeenCalled()
+    expect(mockPublish).not.toHaveBeenCalled()
+  })
+})
--- a/lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
+++ b/lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
@@ -1,13 +1,23 @@
 import { GENERATE_FITNESS_ROUTE_HEATMAP_JOB_NAME } from '@/lib/jobs/names'
-import { recreateFitnessRouteHeatmapJobs } from '@/lib/jobs/recreateFitnessRouteHeatmapJobs'
+import {
+  buildRecreateFitnessRouteHeatmapVariants,
+  recreateFitnessRouteHeatmapJobs
+} from '@/lib/jobs/recreateFitnessRouteHeatmapJobs'
 import { FitnessFile } from '@/lib/types/database/fitnessFile'
+import { logger } from '@/lib/utils/logger'
 
 const mockPublish = jest.fn()
+const mockWarn = logger.warn as jest.Mock
 
 jest.mock('@/lib/services/queue', () => ({
   getQueue: jest.fn(() => ({
     publish: mockPublish
   }))
+}))
+jest.mock('@/lib/utils/logger', () => ({
+  logger: {
+    warn: jest.fn()
+  }
 }))
 
 const makeFitnessFile = (
@@ -30,7 +40,9 @@ const makeFitnessFile = (
 
 describe('recreateFitnessRouteHeatmapJobs', () => {
   beforeEach(() => {
-    mockPublish.mockClear()
+    mockPublish.mockReset()
+    mockPublish.mockResolvedValue(undefined)
+    mockWarn.mockReset()
   })
 
   it('deletes existing heatmaps and queues every discoverable actor variant', async () => {
@@ -66,7 +78,8 @@ describe('recreateFitnessRouteHeatmapJobs', () => {
     expect(
       database.getDistinctRouteHeatmapRegionsForActor
     ).toHaveBeenCalledWith({
-      actorId: 'actor-1'
+      actorId: 'actor-1',
+      includeDeleted: true
     })
     expect(getFitnessFilesByActor).toHaveBeenNthCalledWith(1, {
       actorId: 'actor-1',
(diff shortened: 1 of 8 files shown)


llun:
@codex review

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+export const recreateFitnessRouteHeatmapJobs = async ({
+  database,
+  actorId,
+  dryRun = false,
+  runId = crypto.randomUUID()
+}: RecreateFitnessRouteHeatmapJobsParams): Promise<RecreateFitnessRouteHeatmapJobsResult> => {
+  const [regions, fitnessFiles] = await Promise.all([
+    database.getDistinctRouteHeatmapRegionsForActor({ actorId }),
+    getCompletedPrimaryFitnessFilesForActor(database, actorId)
+  ])
```
llun, line 313: **P1 — Regions are lost on a retry of a partial failure.**

`getDistinctRouteHeatmapRegionsForActor` filters with `whereNull('deletedAt')` (see `lib/database/sql/fitnessRouteHeatmap.ts:391`), and `deleteFitnessRouteHeatmapsForActor` soft-deletes by setting `deletedAt`. The flow here reads regions and immediately soft-deletes every row.

If the first run soft-deletes everything but then crashes (e.g. queue outage) before the region-scoped variants are published, a subsequent rerun will:

1. Call `getDistinctRouteHeatmapRegionsForActor` → returns `[]` (every row is now soft-deleted).
2. Call `deleteFitnessRouteHeatmapsForActor` → returns `0`.
3. Build variants without any `region` mirrors.

Region caches are then permanently absent — exactly the failure mode the script is meant to recover from. Either query the regions ignoring `deletedAt` for the recreation path, or capture/persist the region set before performing the soft-delete and pass it through.

---

scripts/recreateFitnessRouteHeatmaps.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (result.failedCount > 0) {
+    console.error(
+      `Failed to queue ${result.failedCount} route heatmap generation job(s):`
+    )
+    for (const { variant, error } of result.errors) {
+      console.error(`  - ${formatVariant(variant)}: ${error}`)
+    }
+  }
+
+  return result.failedCount > 0 ? 1 : 0
```
llun: **Missing `database.destroy()` — DB pool is never released.**

Every other long-running script in `scripts/` wraps work in a `try { … } finally { await database.destroy() }` block (see `scripts/cleanupLegacyFitnessHeatmaps.ts:23-72`, `scripts/cleanupMediaStorage.ts:52`, `scripts/fixAttachmentUrls.ts:156-158`, `scripts/repairStravaActivityFiles.ts:332`). This script never closes the connection pool; it relies on `process.exit()` to terminate the process while the pool is still open, which can drop in-flight queries and prevents this script from being safely composable.

Wrap from `const database = getDatabase()` down through the success/failure returns in a `try { … } finally { await database.destroy() }` block.

---

scripts/recreateFitnessRouteHeatmaps.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  for (let index = 0; index < args.length; index += 1) {
+    const argument = args[index]
+    if (!argument.startsWith('--')) {
+      throw new Error(`Unexpected argument: ${argument}`)
+    }
+
+    const [rawKey, inlineValue] = argument.slice(2).split('=', 2)
+    if (rawKey === 'dry-run') {
+      parsedArgs[rawKey] = inlineValue === undefined ? true : inlineValue
```
llun: `--dry-run=false` (or any explicit value) silently breaks.

When `inlineValue` is the string `'false'`, `parsedArgs['dry-run']` becomes the string `'false'`, which then fails `z.boolean()` validation in `CliArgs.parse(...)` and throws — but the message will simply say `Expected boolean, received string`, which is confusing for a CLI flag.

Either reject `--dry-run=...` (since it's documented purely as a flag) or coerce the string with `z.coerce.boolean()`/explicit `'true'`/`'false'` handling. Today, only the bare `--dry-run` form works, and that's not enforced.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+    addVariant(variants, {
+      activityType,
+      periodType: 'all_time',
+      periodKey: 'all'
+    })
+  }
+
+  if (!activityDate) {
+    return
+  }
```
llun, line 113: **Behavioural divergence from `enqueueFitnessRouteHeatmapJobs`.**

The on-write path (`lib/jobs/enqueueFitnessRouteHeatmapJobs.ts:21-31`) falls back to `new Date()` when `activityStartTime` is missing and emits yearly/monthly variants for the current period. Here, files with no `activityStartTime` only produce the all-time variant and skip yearly/monthly entirely.

That means a recreation run can leave gaps: any cached yearly/monthly heatmap that was originally enqueued under the "today" fallback won't be regenerated by this script. Either match the existing fallback, or add a comment justifying the divergence so it isn't silently inherited.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+  getDistinctRouteHeatmapRegionsForActor(params: {
+    actorId: string
+  }): Promise<string[]>
+  getFitnessFilesByActor(params: {
+    actorId: string
+    processingStatus: 'completed'
+    isPrimary: true
+    limit: number
+    offset: number
+  }): Promise<FitnessFile[]>
```
llun: The interface pins `processingStatus: 'completed'` and `isPrimary: true` as **required literals**, even though the underlying `Database.getFitnessFilesByActor` signature (`lib/database/sql/fitnessFile.ts:50-59`) declares both as optional and wider. This forces every consumer/mocker to mirror this exact narrow shape, and makes the test mock object (`recreateFitnessRouteHeatmapJobs.test.ts:52-58`) fragile against unrelated changes to `GetFitnessFilesByActorParams`. Prefer typing this argument as `Pick<Database, 'getDistinctRouteHeatmapRegionsForActor' | 'getFitnessFilesByActor' | 'deleteFitnessRouteHeatmapsForActor'>` (or a structural subtype with the wider param shape) so the helper accepts the real DB object directly without a custom interface.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  const sortedBaseVariants = Array.from(baseVariants.values()).sort(
+    sortVariants
+  )
+  const variants = new Map<string, RecreateFitnessRouteHeatmapVariant>()
+
+  for (const variant of sortedBaseVariants) {
+    addVariant(variants, variant)
+  }
```
llun: The first loop (lines 169-171) just copies `sortedBaseVariants` into a fresh `Map`. Since `baseVariants` already deduplicates them and `sortedBaseVariants` preserves the same identities, this is equivalent to `const variants = new Map(baseVariants)` and the intermediate `sortedBaseVariants.sort(...)` on line 164 is also redundant given the final `Array.from(variants.values()).sort(sortVariants)` on line 179.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+        }
+      })
+      queuedCount += 1
+    } catch (error) {
+      errors.push({
+        variant,
+        error: (error as Error).message
+      })
+    }
+  }
```
llun: Publishing is fully sequential (`await` inside the `for` loop). For an actor with hundreds of activity-types × periods × regions this can easily exceed 1000 sequential round-trips to QStash. `enqueueFitnessRouteHeatmapJobs` already uses `Promise.allSettled` for the same publish loop (`lib/jobs/enqueueFitnessRouteHeatmapJobs.ts:104-130`); the recreation path should match.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      dryRun: true
+    })
+
+    expect(result.deletedCount).toBe(0)
+    expect(result.queuedCount).toBe(0)
+    expect(result.variants).toHaveLength(3)
+    expect(database.deleteFitnessRouteHeatmapsForActor).not.toHaveBeenCalled()
+    expect(mockPublish).not.toHaveBeenCalled()
+  })
+})
```
llun, line 392: Coverage gaps worth filling for this helper:

- **Region normalization** — pass `['  region-a  ', '', 'region-a', 'region-b']` and assert the variants only contain trimmed, deduped, sorted regions.
- **File without `activityStartTime`** — assert that yearly/monthly variants are *not* produced (or, if matching `enqueueFitnessRouteHeatmapJobs`, that today's bucket is). Right now this branch in `getActivityDate` is untested.
- **`activityType` whitespace handling** — `addFileVariants` does `file.activityType?.trim() || null`; verify a `'  '`/`''` activityType collapses into the null bucket.
- **Pagination** — the test only ever returns a single page < `FITNESS_FILE_PAGE_SIZE`, so the paging loop in `getCompletedPrimaryFitnessFilesForActor` is uncovered. Add a case where `mockResolvedValueOnce` returns 500 items so the second page fetch is exercised.
- **Per-variant publish failure** — make `mockPublish` reject for one variant and assert `failedCount`/`errors` are populated correctly.

---

llun (commented):
Review of PR #633.

The functional split between `buildRecreateFitnessRouteHeatmapVariants` (pure) and `recreateFitnessRouteHeatmapJobs` (effects) is clean, the dry-run path is sensible, and the `runId`-scoped job IDs correctly side-step queue dedup against prior failed attempts. A few things to address before merge:

**Must fix**
- `lib/jobs/recreateFitnessRouteHeatmapJobs.ts` reads regions from non-deleted heatmaps and then immediately soft-deletes them. A retry after a partial failure will see zero regions and never recreate any region-scoped variants — the very recovery scenario this script targets. Either bypass `deletedAt` for the recreation read, or capture regions before the delete.
- `scripts/recreateFitnessRouteHeatmaps.ts` never calls `database.destroy()`. Other scripts in this repo all wrap their bodies in `try { … } finally { await database.destroy() }`; please match that pattern.

**Should fix**
- `--dry-run=<value>` is silently broken: an inline string flows to `z.boolean()` and throws with a confusing message. Either reject the `=value` form or coerce.
- Files with no `activityStartTime` only emit the all-time variant here, while `enqueueFitnessRouteHeatmapJobs` falls back to "today" and emits yearly/monthly variants. Decide which one is correct and align (or leave a comment justifying the divergence).
- `RecreateFitnessRouteHeatmapJobsDatabase` pins `processingStatus: 'completed'` and `isPrimary: true` as required literals — narrower than the actual `Database` shape and more brittle than necessary.

**Nits**
- The intermediate `sortedBaseVariants` sort + re-add to `variants` is redundant; the final sort is sufficient.
- The publish loop is fully sequential; `Promise.allSettled` (as in `enqueueFitnessRouteHeatmapJobs`) would be much faster for actors with many variants.
- Tests don't cover region normalization, the no-`activityStartTime` branch, the pagination loop, or per-variant publish failure.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
+import crypto from 'crypto'
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: switched to `import { randomUUID } from 'node:crypto'`.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+      variant.periodKey,
+      variant.region ?? ''
+    ].join(':')
+  )
+
+export const recreateFitnessRouteHeatmapJobs = async ({
+  database,
+  actorId,
+  dryRun = false,
+  runId = crypto.randomUUID()
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: the default run ID now calls the destructured `randomUUID()` import.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+}): RecreateFitnessRouteHeatmapVariant[] => {
+  const baseVariants = new Map<string, RecreateFitnessRouteHeatmapVariant>()
+
+  for (const file of fitnessFiles) {
+    addFileVariants(baseVariants, file)
+  }
+
+  const sortedBaseVariants = Array.from(baseVariants.values()).sort(
+    sortVariants
+  )
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: removed the redundant intermediate sort and now only sorts once when returning variants.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: publishing now uses `Promise.allSettled`, keeps per-variant failure records, and logs rejected publishes with the server logger. Variants are already deduplicated by key while they are built, so no extra JSON-stringify dedupe step is needed.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export const recreateFitnessRouteHeatmapJobs = async ({
+  database,
+  actorId,
+  dryRun = false,
+  runId = crypto.randomUUID()
+}: RecreateFitnessRouteHeatmapJobsParams): Promise<RecreateFitnessRouteHeatmapJobsResult> => {
+  const [regions, fitnessFiles] = await Promise.all([
+    database.getDistinctRouteHeatmapRegionsForActor({ actorId }),
+    getCompletedPrimaryFitnessFilesForActor(database, actorId)
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: `getDistinctRouteHeatmapRegionsForActor` now supports `includeDeleted`, and the recreation path reads regions with `includeDeleted: true` before rebuilding variants.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
llun: Fixed in 6a29d0f8: the recreation path now reads region keys with `includeDeleted: true`, so a rerun after a partial soft-delete still recreates region-scoped variants.

scripts/recreateFitnessRouteHeatmaps.ts
llun: Fixed in 6a29d0f8: the CLI now wraps database work in `try/finally` and always calls `database.destroy()`.

scripts/recreateFitnessRouteHeatmaps.ts
llun: Fixed in 6a29d0f8: `--dry-run`, `--dry-run=true`, and `--dry-run=false` are handled explicitly, and invalid inline values now get a targeted error message.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
llun: Left the behavior unchanged and documented it in 6a29d0f8. The route heatmap generator filters yearly/monthly jobs by `activityStartTime`, so files without a timestamp cannot contribute to scoped caches. Matching the enqueue fallback here would only recreate arbitrary current-period empty jobs, not useful data.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
llun: Fixed in 6a29d0f8: the helper database dependency is now typed as a `Pick<Database, ...>` using the real method signatures.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
llun: Fixed in 6a29d0f8: removed the redundant copy/sort path and now mirrors region variants from the existing deduped base map.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
llun: Fixed in 6a29d0f8: the publish loop now uses `Promise.allSettled` like the normal route heatmap enqueue path.

lib/jobs/recreateFitnessRouteHeatmapJobs.test.ts
llun: Fixed in 6a29d0f8: added coverage for region normalization, blank activity types, missing `activityStartTime`, pagination past the 500-file page size, and per-variant publish failure reporting.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+    actorId
+  })
+  const queue = getQueue()
+  const errors: RecreateFitnessRouteHeatmapJobError[] = []
+  let queuedCount = 0
+
+  for (const variant of variants) {
+    try {
+      await queue.publish({
+        id: buildJobId({ runId, actorId, variant }),
```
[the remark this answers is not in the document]
llun: Not changing the delete-before-publish ordering. This is an explicit cache recreation script, so clearing stale rows before fresh generation is intentional. Publishing first would be unsafe for the local `NoQueue` backend because `publish()` executes the generation job synchronously, then a later delete would hide the freshly generated rows. In 6a29d0f8 the script still returns non-zero on publish failure and now records per-variant publish errors.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+  while (true) {
+    const page = await database.getFitnessFilesByActor({
+      actorId,
+      processingStatus: 'completed',
+      isPrimary: true,
+      limit: FITNESS_FILE_PAGE_SIZE,
+      offset
+    })
+
+    fitnessFiles.push(...page)
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: the helper now derives unique variants while paging through fitness files instead of materializing every completed file in memory first.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,292 @@
...
+
+    if (page.length < FITNESS_FILE_PAGE_SIZE) {
+      break
+    }
+
+    offset += page.length
```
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: added a pagination test with a full 500-item first page and a second page fetch at offset 500.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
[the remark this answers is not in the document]
llun: Fixed in 6a29d0f8: region discovery now uses `includeDeleted: true`, so reruns after partial recreation failures keep region-scoped variants.

llun:
@codex review

llun (commented):
Re-review of `6a29d0f` — every previous concern has been properly addressed.

**Verified fixes**
- **Region recovery (P1):** `getDistinctRouteHeatmapRegionsForActor` now accepts `includeDeleted`, the recreation path passes `true`, and there's a DB-level test that proves regions survive a soft-delete. The fix is minimal and the only other call site (`enqueueFitnessRouteHeatmapJobs.ts:90`) keeps the safe default.
- **`database.destroy()` leak:** CLI body is wrapped in `try { … } finally { await database.destroy() }`.
- **`--dry-run=value` parsing:** `parseDryRunValue` handles `true` / `false` / no-value explicitly and produces a clear error otherwise — much better than the prior zod fall-through.
- **Files without `activityStartTime`:** The intentional divergence from `enqueueFitnessRouteHeatmapJobs` is now documented inline with a clear rationale (yearly/monthly generation filters by `activityStartTime`).
- **Database type:** `RecreateFitnessRouteHeatmapJobsDatabase` is now `Pick<Database, …>`, so the helper takes the real DB type and tests no longer have to mirror narrowed literals.
- **Variant building:** Replaced the redundant double-sort with `new Map(baseVariants)` and only sorts once at return.
- **Publish concurrency:** `Promise.allSettled` with per-variant error capture and `logger.warn` matches the on-write path.
- **Memory:** Variants are now derived during pagination via `getCompletedPrimaryFitnessFileVariantsForActor`, no longer materializing every fitness file (good catch from Copilot).
- **Test coverage:** New cases for region normalization + blank activity types, full pagination across the 500-row boundary, and per-variant publish failure all look well-targeted.

CI is green. No further blockers from my side — LGTM.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,313 @@
...
+    actorId
+  })
+  const queue = getQueue()
+  const errors: RecreateFitnessRouteHeatmapJobError[] = []
+  let queuedCount = 0
+
+  const publishResults = await Promise.allSettled(
+    variants.map((variant) =>
+      queue.publish({
+        id: buildJobId({ runId, actorId, variant }),
```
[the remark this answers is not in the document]
llun: Fixed in 5f52bf3a9c24a66b819ccf685119577dac194982: route heatmap recreation publishes now run through a strict concurrency-limited worker pool, so the NoQueue backend cannot start every generation job inline at once while per-variant success/failure accounting is preserved.

llun:
@codex review

llun (commented):
Re-review of `5f52bf3` — focused, surgical follow-up that replaces the unbounded `Promise.allSettled` publish with a 4-worker pool. Looks good.

**What I verified**
- Worker pool semantics are correct: `nextIndex` is shared between workers (safe in single-threaded JS), each worker drains until exhausted, and per-variant `try/catch` guarantees one failure can't kill the worker. Errors are written to `results[index]` so positional alignment with `variants` is preserved for the downstream `forEach`.
- Edge cases hold up: `variants.length === 0` → `workerCount = 0` → `Promise.all([])` → empty result array. `variants.length < CONCURRENCY` → `Math.min(...)` correctly clamps the worker count.
- The new test exercises the actual concurrency bound by holding publishes open with manual `releasePublish` callbacks and asserts `maxActivePublishCount === 4` while still draining all 26 variants. That's the right way to test it (rather than just the call count).
- Refactoring `publishRouteHeatmapVariant` out as its own function keeps the worker body readable; the typing on `Queue` is correctly imported from `@/lib/services/queue/type`.

**Nit (non-blocking)**
- The concurrency value is a hard-coded `4`. If QStash rate-limit tuning ever becomes a concern, lifting it to config/env would be straightforward — but for a recovery script it's fine as-is.

CI is green, all prior threads remain resolved. LGTM.

---

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,373 @@
...
+      deletedCount: 0,
+      queuedCount: 0,
+      failedCount: 0,
+      errors: []
+    }
+  }
+
+  const deletedCount = await database.deleteFitnessRouteHeatmapsForActor({
+    actorId
+  })
```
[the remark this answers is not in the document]
llun: Leaving the delete-before-publish ordering unchanged. This script is explicitly for cache recreation, so hiding stale rows before fresh generation is intentional. Publishing first would be unsafe with the local `NoQueue` backend because `publish()` executes generation inline; deleting afterward could hide newly generated rows. The current PR now preserves rerun region discovery, reports publish failures, and caps publish concurrency.

scripts/recreateFitnessRouteHeatmaps.ts
```diff
@@ -0,0 +1,178 @@
...
+  for (let index = 0; index < args.length; index += 1) {
+    const argument = args[index]
+    if (!argument.startsWith('--')) {
+      throw new Error(`Unexpected argument: ${argument}`)
+    }
+
+    const [rawKey, inlineValue] = argument.slice(2).split('=', 2)
+    if (rawKey === 'dry-run') {
+      parsedArgs[rawKey] = parseDryRunValue(inlineValue)
+      continue
```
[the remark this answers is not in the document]
llun: Fixed in 1d955d27ace4b0a11c4bce81974aa2df477a3a6f: the CLI now accepts `--dry-run false` and `--dry-run true` by consuming a following boolean token, while keeping bare `--dry-run` and inline `--dry-run=false` support. Added parser coverage in `scripts/recreateFitnessRouteHeatmaps.test.ts`.

lib/jobs/recreateFitnessRouteHeatmapJobs.ts
```diff
@@ -0,0 +1,373 @@
...
+  let offset = 0
+
+  while (true) {
+    const page = await database.getFitnessFilesByActor({
+      actorId,
+      processingStatus: 'completed',
+      isPrimary: true,
+      limit: FITNESS_FILE_PAGE_SIZE,
+      offset
+    })
```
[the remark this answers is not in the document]
llun: Fixed in dca6a7cd163e50b64bfaeffaffad27a440c5c250: `getFitnessFilesByActor` now orders by `createdAt desc, id desc`, giving offset-based pagination a deterministic tiebreaker when files share the same timestamp. Added database coverage for same-`createdAt` files to lock the stable ordering.

PR #645, opened 2026-05-09
fix: make status navigation timestamp-only

llun:
## Summary

- Move status-detail navigation from the whole timeline/comment container to the timestamp control.
- Keep poll/body clicks local to the post so poll interaction no longer opens the status page.
- Add regression coverage for timeline posts and status-detail comment boxes.
- Add a client SHA-256 fallback for non-secure local hosts where `crypto.subtle` is unavailable, which keeps timestamp status links working on `activities.local` over HTTP.

## Root Cause

Timeline posts and comment status boxes attached `router.push` to the outer post container. Poll clicks bubbled to that container, so clicking poll content opened the status detail page.

## Validation

- `yarn build`
- `yarn test` (247 suites, 1843 tests)
- `yarn migrate` against a local SQLite dev DB
- Browser verified on `http://activities.local:3000`: clicking poll content stayed on the profile page, clicking the timestamp opened the status detail page.

### DIFF
--- a/app/(timeline)/[actor]/[status]/StatusBox.test.tsx
+++ b/app/(timeline)/[actor]/[status]/StatusBox.test.tsx
@@ -6,10 +6,15 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import { ReactNode } from 'react'
 
 import { votePoll } from '@/lib/client'
-import { StatusPoll, StatusType } from '@/lib/types/domain/status'
+import {
+  pollStatusCurrentTime,
+  pollStatusFixture
+} from '@/lib/components/posts/__fixtures__/poll-status'
 import { getStatusDetailPathClient } from '@/lib/utils/getStatusDetailPathClient'
 
 import { StatusBox } from './StatusBox'
+
+const mockPush = jest.fn()
 
 jest.mock('@/lib/components/posts/collapsible-content', () => ({
   CollapsibleContent: ({ children }: { children: ReactNode }) => (
@@ -40,65 +45,8 @@ jest.mock('@/lib/utils/getStatusDetailPathClient', () => ({
   getStatusDetailPathClient: jest.fn()
 }))
 
-const mockPush = jest.fn()
 const mockVotePoll = jest.mocked(votePoll)
 const mockGetStatusDetailPathClient = jest.mocked(getStatusDetailPathClient)
-
-const currentTime = new Date('2026-04-26T10:00:00.000Z').getTime()
-
-const pollStatus: StatusPoll = {
-  id: 'https://activities.local/users/llun/statuses/poll-1',
-  actorId: 'https://activities.local/users/llun',
-  actor: {
-    id: 'https://activities.local/users/llun',
-    username: 'llun',
-    domain: 'activities.local',
-    name: 'Llun',
-    followersUrl: 'https://activities.local/users/llun/followers',
-    inboxUrl: 'https://activities.local/users/llun/inbox',
-    sharedInboxUrl: 'https://activities.local/inbox',
-    followingCount: 0,
-    followersCount: 0,
-    statusCount: 0,
-    lastStatusAt: null,
-    createdAt: currentTime
-  },
-  to: [],
-  cc: [],
-  edits: [],
-  isLocalActor: true,
-  createdAt: currentTime,
-  updatedAt: currentTime,
-  type: StatusType.enum.Poll,
-  url: 'https://activities.local/@llun/poll-1',
-  text: 'Question',
-  summary: null,
-  reply: '',
-  replies: [],
-  actorAnnounceStatusId: null,
-  isActorLiked: false,
-  totalLikes: 0,
-  totalShares: 0,
-  tags: [],
-  endAt: new Date('2026-12-31T10:00:00.000Z').getTime(),
-  choices: [
-    {
-      title: 'Option A',
-      totalVotes: 2,
-      statusId: 'https://activities.local/users/llun/statuses/poll-1',
-      createdAt: currentTime,
-      updatedAt: currentTime
-    },
-    {
-      title: 'Option B',
-      totalVotes: 1,
-      statusId: 'https://activities.local/users/llun/statuses/poll-1',
-      createdAt: currentTime,
-      updatedAt: currentTime
-    }
-  ],
-  pollType: 'oneOf'
-}
 
 describe('StatusBox', () => {
   beforeEach(() => {
@@ -111,14 +59,18 @@ describe('StatusBox', () => {
     render(
       <StatusBox
         host="activities.local"
-        currentActor={pollStatus.actor}
-        currentTime={currentTime}
-        status={pollStatus}
+        currentActor={pollStatusFixture.actor}
+        currentTime={pollStatusCurrentTime}
+        status={pollStatusFixture}
         variant="comment"
       />
     )
 
     const option = screen.getByLabelText('Option A')
+
+    fireEvent.click(screen.getByText('Question'))
+    expect(mockGetStatusDetailPathClient).not.toHaveBeenCalled()
+    expect(mockPush).not.toHaveBeenCalled()
 
     fireEvent.click(option)
 
--- a/app/(timeline)/[actor]/[status]/StatusBox.test.tsx
+++ b/app/(timeline)/[actor]/[status]/StatusBox.test.tsx
@@ -1,0 +1,135 @@
+/**
+ * @jest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { ReactNode } from 'react'
+
+import { votePoll } from '@/lib/client'
+import { StatusPoll, StatusType } from '@/lib/types/domain/status'
+import { getStatusDetailPathClient } from '@/lib/utils/getStatusDetailPathClient'
+
+import { StatusBox } from './StatusBox'
+
+jest.mock('@/lib/components/posts/collapsible-content', () => ({
+  CollapsibleContent: ({ children }: { children: ReactNode }) => (
+    <div>{children}</div>
+  )
+}))
+
+jest.mock('./FitnessStatusDetail', () => ({
+  FitnessStatusDetail: () => null
+}))
+
+jest.mock('./StatusLikes', () => ({
+  StatusLikes: () => null
+}))
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({
+    push: mockPush,
+    refresh: jest.fn()
+  })
+}))
+
+jest.mock('@/lib/client', () => ({
+  votePoll: jest.fn()
+}))
+
+jest.mock('@/lib/utils/getStatusDetailPathClient', () => ({
+  getStatusDetailPathClient: jest.fn()
+}))
+
+const mockPush = jest.fn()
+const mockVotePoll = jest.mocked(votePoll)
+const mockGetStatusDetailPathClient = jest.mocked(getStatusDetailPathClient)
+
+const currentTime = new Date('2026-04-26T10:00:00.000Z').getTime()
+
+const pollStatus: StatusPoll = {
+  id: 'https://activities.local/users/llun/statuses/poll-1',
+  actorId: 'https://activities.local/users/llun',
+  actor: {
+    id: 'https://activities.local/users/llun',
+    username: 'llun',
+    domain: 'activities.local',
+    name: 'Llun',
+    followersUrl: 'https://activities.local/users/llun/followers',
+    inboxUrl: 'https://activities.local/users/llun/inbox',
+    sharedInboxUrl: 'https://activities.local/inbox',
+    followingCount: 0,
+    followersCount: 0,
+    statusCount: 0,
+    lastStatusAt: null,
+    createdAt: currentTime
+  },
+  to: [],
+  cc: [],
+  edits: [],
+  isLocalActor: true,
+  createdAt: currentTime,
+  updatedAt: currentTime,
+  type: StatusType.enum.Poll,
+  url: 'https://activities.local/@llun/poll-1',
+  text: 'Question',
+  summary: null,
+  reply: '',
+  replies: [],
+  actorAnnounceStatusId: null,
+  isActorLiked: false,
+  totalLikes: 0,
+  totalShares: 0,
+  tags: [],
+  endAt: new Date('2026-12-31T10:00:00.000Z').getTime(),
+  choices: [
+    {
+      title: 'Option A',
+      totalVotes: 2,
+      statusId: 'https://activities.local/users/llun/statuses/poll-1',
+      createdAt: currentTime,
+      updatedAt: currentTime
+    },
+    {
+      title: 'Option B',
+      totalVotes: 1,
+      statusId: 'https://activities.local/users/llun/statuses/poll-1',
+      createdAt: currentTime,
+      updatedAt: currentTime
+    }
+  ],
+  pollType: 'oneOf'
+}
+
+describe('StatusBox', () => {
+  beforeEach(() => {
+    mockPush.mockClear()
+    mockVotePoll.mockReset()
+    mockGetStatusDetailPathClient.mockResolvedValue('/@llun/poll-1')
+  })
+
+  it('only opens comment status detail pages from the timestamp', async () => {
+    render(
+      <StatusBox
+        host="activities.local"
+        currentActor={pollStatus.actor}
+        currentTime={currentTime}
+        status={pollStatus}
+        variant="comment"
+      />
+    )
+
+    const option = screen.getByLabelText('Option A')
+
+    fireEvent.click(option)
+
+    expect(option).toBeChecked()
+    expect(mockGetStatusDetailPathClient).not.toHaveBeenCalled()
+    expect(mockPush).not.toHaveBeenCalled()
+
+    fireEvent.click(screen.getByRole('button', { name: /Open status by Llun/ }))
+
+    await waitFor(() => {
+      expect(mockPush).toHaveBeenCalledWith('/@llun/poll-1')
+    })
+  })
+})
--- a/app/(timeline)/[actor]/[status]/StatusBox.tsx
+++ b/app/(timeline)/[actor]/[status]/StatusBox.tsx
@@ -67,9 +67,9 @@ export const StatusBox: FC<Props> = ({
     )
   }
 
-  const openStatus = () => {
+  const openStatus = (statusToOpen: Status) => {
     void (async () => {
-      const detailPath = await getStatusDetailPathClient(status)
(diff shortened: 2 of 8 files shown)


lib/components/posts/post.tsx
```diff
@@ -231,9 +238,23 @@ export const Post: FC<PostProps> = (props) => {
...
-            </span>
+            {props.onOpenStatus ? (
+              <button
+                type="button"
+                className="text-muted-foreground text-xs whitespace-nowrap hover:underline"
+                aria-label={openStatusLabel}
+                onClick={(event) => {
+                  event.stopPropagation()
+                  props.onOpenStatus?.()
+                }}
```
llun, line 252: `event.stopPropagation()` is now a no-op: the parent `<article>` in `StatusBox.tsx` (line 79) and the wrapper in `posts.tsx` no longer have any `onClick`. The two pre-existing `stopPropagation` calls below (the external-link `<a>` around line 263 and the actions `<div>` around line 279) are dead for the same reason. Worth removing all three so future readers don't think there's a parent handler being defended against.

---

lib/components/posts/post.tsx
```diff
@@ -68,6 +69,12 @@ export const Post: FC<PostProps> = (props) => {
   const externalStatusUrl = actualStatus.url || actualStatus.id
   const showExternalLink =
     !actualStatus.isLocalActor && Boolean(externalStatusUrl)
+  const relativeCreatedAt = formatDistanceToNow(actualStatus.createdAt)
```
llun: Pre-existing, but you're touching this exact line: `formatDistanceToNow(actualStatus.createdAt)` uses `Date.now()` internally, which causes hydration mismatches between SSR and client. `Post` already accepts `currentTime: number` as a prop (line 34) — please pass it explicitly. Other components in the codebase already do this, e.g. `poll.tsx:156` (`formatDistance(status.endAt, now)`) and `SessionsList.tsx:55`. CLAUDE.md explicitly calls this pattern out.

---

lib/components/posts/post.tsx
```diff
@@ -68,6 +69,12 @@ export const Post: FC<PostProps> = (props) => {
...
+  const actorName =
+    actualStatus.actor?.name ||
+    actualStatus.actor?.username ||
+    getActorIdMention(actualStatus.actorId)
```
llun: When `actualStatus.actor` is undefined, this falls back to `getActorIdMention(actualStatus.actorId)`, which can return the bare actor URL. The resulting aria-label "Open status by https://… from 5 minutes ago" reads oddly to screen readers. Consider dropping the "by {actorName}" portion when no actor is available, or substituting a friendlier fallback.

---

lib/components/posts/post.tsx
```diff
@@ -231,9 +238,23 @@ export const Post: FC<PostProps> = (props) => {
...
+                  props.onOpenStatus?.()
+                }}
+              >
+                {relativeCreatedAt}
+              </button>
+            ) : (
+              <span className="text-muted-foreground text-xs whitespace-nowrap">
+                {relativeCreatedAt}
+              </span>
+            )}
```
llun, line 258: The button and span branches duplicate the `text-muted-foreground text-xs whitespace-nowrap` classes. Extracting them into a constant (or rendering a single element with conditional handlers) would prevent style drift between the two paths.

---

lib/components/posts/posts.tsx
```diff
@@ -73,39 +73,37 @@ export const Posts: FC<Props> = ({
...
+            currentActor={currentActor}
+            status={status}
+            showActions={showActions}
+            editable={currentActor?.id === status.actorId}
+            collapsible
+            postLineLimit={postLineLimit}
+            onReply={handleReply}
+            onEdit={onEdit}
+            onPostDeleted={onPostDeleted}
+            onOpenStatus={() => openStatus(status)}
```
llun: `onOpenStatus={() => openStatus(status)}` allocates a new closure per status on every render, which would defeat any future memoization on `Post`. Minor — only worth doing for long timelines — but if it matters, hoist `openStatus` to take the status as part of the callback shape (e.g. `onOpenStatus={openStatus.bind(null, status)}` or have `Post` invoke `onOpenStatus(actualStatus)` and pass `openStatus` directly).

---

app/(timeline)/[actor]/[status]/StatusBox.tsx
```diff
@@ -68,21 +67,16 @@ export const StatusBox: FC<Props> = ({
...
-        )}
-        onClick={() => {
-          if (variant === 'detail') return
-          void (async () => {
-            const detailPath = await getStatusDetailPathClient(status)
-            if (detailPath) router.push(detailPath)
-          })()
-        }}
-      >
+      <article className="p-4 transition-colors">
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/StatusBox.tsx:79]: With the `onClick` and conditional `cursor-pointer hover:bg-muted/40` removed, this `<article>` no longer has any visual affordance that the post is interactive (the previous version highlighted on hover). The clickable area shrinks to just the timestamp string ("5m"), which is a hard tap target on touch devices. Was that the intended UX? If yes, please confirm you verified it in the browser per CLAUDE.md ("use the browser to verify any UI change"). If you want to keep some affordance, consider expanding the clickable region to the meta row (avatar + actor + timestamp) rather than only the timestamp text.

---

app/(timeline)/[actor]/[status]/StatusBox.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    fireEvent.click(option)
+
+    expect(option).toBeChecked()
+    expect(mockGetStatusDetailPathClient).not.toHaveBeenCalled()
+    expect(mockPush).not.toHaveBeenCalled()
+
+    fireEvent.click(screen.getByRole('button', { name: /Open status by Llun/ }))
+
+    await waitFor(() => {
+      expect(mockPush).toHaveBeenCalledWith('/@llun/poll-1')
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/StatusBox.test.tsx:84]: This test asserts that clicking a poll radio option doesn't navigate, but radio clicks never navigated under the old behavior either (the `<input type="radio">` swallowed the event). To actually guard the "timestamp-only" invariant against a re-introduced article-level `onClick`, also click the article container or the post text and assert `mockPush` was not called.

---

lib/utils/getHashFromStringClient.ts
```diff
@@ -1,7 +1,110 @@
...
+    hash[2] = (hash[2] + c) >>> 0
+    hash[3] = (hash[3] + d) >>> 0
+    hash[4] = (hash[4] + e) >>> 0
+    hash[5] = (hash[5] + f) >>> 0
+    hash[6] = (hash[6] + g) >>> 0
+    hash[7] = (hash[7] + h) >>> 0
+  }
+
+  return hash.map(toHex).join('')
+}
```
llun, line 98: The PR description says the fallback is for "non-secure local hosts where `crypto.subtle` is unavailable, which keeps timestamp status links working on `activities.local` over HTTP." A few concerns about the trade-off:

1. `~100` lines of hand-rolled crypto is a meaningful maintenance burden and widens the surface area (even though the hash here is non-security-critical — just URL paths).
2. CLAUDE.md says local dev runs at `https://activities.local`, where `crypto.subtle` is available. If there is a real path that hits HTTP, please document it in a comment near the fallback (which dev workflow / browser / port?). Otherwise this is dead code in practice.
3. If the goal is just "URL hash on HTTP," a much smaller alternative is to import a hashed function from a vetted lib (e.g. `js-sha256` is ~4 KB and battle-tested), or skip hashing entirely on insecure contexts and use `encodeURIComponent(actualStatus.url)` like the non-local-actor branch in `getStatusDetailPathClient.ts:11` already does. Either keeps the call-site simple without re-implementing SHA-256.

If you keep the implementation, please at least add a one-line comment explaining the runtime that motivated it so it isn't deleted in a future cleanup.

---

lib/utils/getHashFromStringClient.test.ts
```diff
@@ -24,4 +24,24 @@ describe('getHashFromStringClient', () => {
...
+      await expect(getHashFromStringClient('test string')).resolves.toBe(
+        'd5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b'
+      )
+    } finally {
+      Object.defineProperty(globalThis, 'crypto', {
+        configurable: true,
+        value: originalCrypto
+      })
+    }
+  })
```
llun, line 55: Two suggestions:

1. The expected hex `d5579c46…930b` is SHA-256(`'test string'`). Adding a one-line comment saying so would help readers verify the constant.
2. Replacing the entire `globalThis.crypto` is more aggressive than needed and makes the test fragile if any module caches the original ref. Consider mocking only `subtle`:
   ```ts
   const restore = Object.getOwnPropertyDescriptor(globalThis.crypto, 'subtle')
   Object.defineProperty(globalThis.crypto, 'subtle', { configurable: true, value: undefined })
   try { /* ... */ } finally { Object.defineProperty(globalThis.crypto, 'subtle', restore!) }
   ```
   This exercises the real fallback branch (`!globalThis.crypto?.subtle`) without disturbing other crypto consumers.

---

app/(timeline)/[actor]/[status]/StatusBox.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+jest.mock('@/lib/client', () => ({
+  votePoll: jest.fn()
+}))
+
+jest.mock('@/lib/utils/getStatusDetailPathClient', () => ({
+  getStatusDetailPathClient: jest.fn()
+}))
+
+const mockPush = jest.fn()
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/StatusBox.test.tsx]: `mockPush` is referenced inside the `jest.mock('next/navigation', ...)` factory but declared a few lines below. This works because the factory captures `mockPush` lazily (Jest allows variables prefixed with `mock`), but the ordering reads as if `mockPush` is being used before declaration. Moving `const mockPush = jest.fn()` above the `jest.mock` calls makes this less surprising.

Also: the `pollStatus` fixture and the `next/navigation` / `@/lib/client` / `getStatusDetailPathClient` mock setup in `lib/components/posts/posts.test.tsx` are nearly identical to this file. Consider lifting them into a shared test fixture under `lib/components/posts/__fixtures__/` (or similar).

---

llun (commented):
Code review of `cc77b3f` ("fix: make status navigation timestamp-only"). Inline comments below; this is a summary of what should be addressed before merge.

## Top items to fix

1. **Confirm the UX intent and verify in the browser** — `StatusBox.tsx:79`. Click target shrank from "entire post body" to "the timestamp string". On touch devices that's hard to hit, and the previous `cursor-pointer hover:bg-muted/40` affordance is gone. PR description says it was browser-verified; please confirm specifically on a touch viewport.
2. **Drop the dead `event.stopPropagation()` calls** — `post.tsx:249` (new button), and the two pre-existing ones at `post.tsx:263` (external link) and `post.tsx:279` (actions wrapper). With both parent handlers removed, none of the three are doing anything.
3. **Re-evaluate the hand-rolled SHA-256 fallback** — `getHashFromStringClient.ts`. Either delete it (CLAUDE.md says dev runs over HTTPS, where `crypto.subtle` exists), document the specific HTTP path that motivated it, or replace with a vetted dependency. ~100 lines of hand-rolled crypto for a non-security-critical URL hash is a high maintenance cost.
4. **Strengthen the regression test** — `StatusBox.test.tsx`. The current "doesn't navigate when poll option clicked" assertion wouldn't have caught the original bug (radio inputs swallow clicks). Click the article body / post text and assert no navigation to actually guard the "timestamp-only" invariant.

## Smaller fixes

5. **Pre-existing hydration risk on the timestamp** — `post.tsx:72`. `formatDistanceToNow(actualStatus.createdAt)` uses `Date.now()` internally; pass the existing `currentTime` prop instead, the way `poll.tsx:156` and `SessionsList.tsx:55` do. CLAUDE.md explicitly calls this pattern out.
6. **Aria-label fallback** — `post.tsx:76`. When `actor` is undefined the label can read "Open status by https://…", which is awkward for screen readers.
7. **Class duplication between the button and span** — `post.tsx:241-257`.
8. **Inline closure per row in `Posts`** — `posts.tsx:102`. Minor perf / memoization concern.
9. **Test fixture duplication** — `StatusBox.test.tsx` and `posts.test.tsx` share ~60 lines of `pollStatus` and mock setup; lift to a shared fixture.
10. **Mock scope and constants** — `getHashFromStringClient.test.ts`. Mock just `crypto.subtle` rather than replacing the whole `crypto` global, and add a comment noting the expected hex is SHA-256 of `'test string'`.

## What looks good

- Switching to a real `<button>` with `aria-label` is a meaningful accessibility win over "click anywhere on the article" with no role / label.
- `onOpenStatus` is correctly gated by `variant === 'comment'` in `StatusBox.tsx:92`, preserving prior behavior for the detail variant.
- `cn` import correctly removed once the conditional class went away.

Pre-commit gate (per CLAUDE.md: `prettier --write .` → `lint` → `build` → `test`) is reportedly green per the PR description; I have not re-run it locally.

---

lib/components/posts/posts.tsx
```diff
@@ -73,39 +73,37 @@ export const Posts: FC<Props> = ({
...
+      const detailPath = await getStatusDetailPathClient(status)
+      if (detailPath) router.push(detailPath)
+    })()
+  }
+
   return (
     <section className={cn('w-full grid grid-cols-1', className)}>
       {statuses.map((status, index) => (
         <article
           key={`${index}-${status.id}`}
```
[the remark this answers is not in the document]
llun: Fixed in c7c92480: changed the timeline row key to status.id and removed the index from the map callback so prepends/reorders keep stable row identity.

lib/components/posts/post.tsx
```diff
@@ -68,6 +69,12 @@ export const Post: FC<PostProps> = (props) => {
   const externalStatusUrl = actualStatus.url || actualStatus.id
   const showExternalLink =
     !actualStatus.isLocalActor && Boolean(externalStatusUrl)
+  const relativeCreatedAt = formatDistanceToNow(actualStatus.createdAt)
+  const actorName =
+    actualStatus.actor?.name ||
+    actualStatus.actor?.username ||
+    getActorIdMention(actualStatus.actorId)
+  const openStatusLabel = `Open status by ${actorName} from ${relativeCreatedAt}`
```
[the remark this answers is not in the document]
llun: Fixed in c7c92480: the visible timestamp stays the same, and the aria-label now reads 'Open status by X, posted Y ago'.

lib/utils/getHashFromStringClient.test.ts
```diff
@@ -24,4 +24,24 @@ describe('getHashFromStringClient', () => {
...
+
+    try {
+      await expect(getHashFromStringClient('test string')).resolves.toBe(
+        'd5579c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e930b'
+      )
+    } finally {
+      Object.defineProperty(globalThis, 'crypto', {
+        configurable: true,
+        value: originalCrypto
+      })
```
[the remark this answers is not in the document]
llun: Fixed in c7c92480: the fallback test now mocks only crypto.subtle and restores the original descriptor in finally instead of replacing globalThis.crypto.

lib/utils/getHashFromStringClient.ts
```diff
@@ -1,7 +1,110 @@
 const encoder = new TextEncoder()
 
+const SHA_256_INITIAL_HASH = [
+  0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a, 0x510e527f, 0x9b05688c,
+  0x1f83d9ab, 0x5be0cd19
+]
```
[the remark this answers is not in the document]
llun: Reply-only: I kept the fallback inline for now. It is a small deterministic fallback for the non-secure activities.local HTTP browser-verification path where crypto.subtle is unavailable; dynamic import would add module-splitting complexity without a measured bundle-size problem. I documented that runtime in c7c92480 so the tradeoff is explicit.

lib/components/posts/post.tsx
llun: Fixed in c7c92480: removed the now-dead propagation guards from the timestamp button, fitness-file link, external status link, and Actions wrapper.

lib/components/posts/post.tsx
llun: Fixed in c7c92480: replaced formatDistanceToNow with formatDistance(actualStatus.createdAt, props.currentTime) so the timestamp is derived from the existing serialized currentTime prop.

lib/components/posts/post.tsx
llun: Fixed in c7c92480: when actor data is missing, the aria-label now omits the 'by ...' phrase instead of falling back to the actor URL.

lib/components/posts/post.tsx
llun: Fixed in c7c92480: extracted the shared timestamp class string so the button and span branches stay in sync.

lib/components/posts/posts.tsx
llun: Fixed in c7c92480: Post now calls onOpenStatus with the status, so Posts can pass openStatus directly instead of allocating an onOpenStatus closure per row.

app/(timeline)/[actor]/[status]/StatusBox.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/StatusBox.tsx:79]: Addressed in c7c92480: kept timestamp-only navigation to preserve the requested UX, but expanded the timestamp button hit target with padding/min-height and focus styling. I also verified on activities.local in a 390px-wide browser viewport: clicking the poll body and poll option did not navigate, while clicking the timestamp opened the status detail page.

app/(timeline)/[actor]/[status]/StatusBox.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/StatusBox.test.tsx:84]: Fixed in c7c92480: the StatusBox regression test now clicks the post text and asserts no status-detail navigation before checking the poll option and timestamp behavior.

lib/utils/getHashFromStringClient.ts
llun: Addressed in c7c92480: added the runtime comment for activities.local over HTTP, where Web Crypto can be unavailable during browser verification. I kept the fallback instead of switching to encodeURIComponent because local status detail URLs need to match the hashed server-side route, and I did not add a new dependency for this non-security fallback.

lib/utils/getHashFromStringClient.test.ts
llun: Fixed in c7c92480: added the SHA-256('test string') comment and changed the test to mock only crypto.subtle, restoring the original descriptor afterward.

app/(timeline)/[actor]/[status]/StatusBox.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/StatusBox.test.tsx]: Fixed in c7c92480: moved mockPush above the jest.mock factory and lifted the shared poll fixture into lib/components/posts/__fixtures__/poll-status.ts.

llun:
@codex review

PR #653, opened 2026-05-10
fix: Blank notification rows

llun:
## Summary

- Render `activity_import` and `reblog` notifications in the timeline notifications UI instead of falling through to empty cards.
- Render status-backed notifications with missing status data as explicit unavailable rows so they remain visible and can be marked read.
- Avoid nested anchors in notification previews because processed status text can contain links.
- Address review follow-up by using generic fitness import copy, shared notification render types, nullable notification render fields, a notification-local status path helper, and consistent grouped notification fallbacks.

## Root cause

Valid notification types existed in the database and API layer, but the web `NotificationItem` switch did not handle `activity_import` or `reblog`. The component still rendered the outer notification card, so unsupported types produced blank rows. Stale status-backed notifications could also be filtered out or return `null`, leaving unread rows invisible.

## Verification

- `yarn run prettier --write app/(timeline)/notifications/NotificationItem.tsx app/(timeline)/notifications/NotificationItem.test.tsx app/(timeline)/notifications/components/ActivityImportNotification.tsx app/(timeline)/notifications/components/LikeNotification.tsx app/(timeline)/notifications/components/MentionNotification.tsx app/(timeline)/notifications/components/ReblogNotification.tsx app/(timeline)/notifications/components/ReplyNotification.tsx app/(timeline)/notifications/getNotificationStatusPath.ts app/(timeline)/notifications/types.ts`
- `git diff --check origin/codex/fix-blank-notifications..HEAD`
- `yarn test --runTestsByPath 'app/(timeline)/notifications/NotificationItem.test.tsx' --runInBand` - 16 tests passed.
- `yarn lint` - 0 errors, 24 existing `no-explicit-any` warnings in auth/fitness code and duplicated under `.claude/worktrees`.
- `yarn build` - passed.
- `yarn test --runInBand --modulePathIgnorePatterns='<rootDir>/.claude' --testPathIgnorePatterns='<rootDir>/.claude'` - 260 suites / 1,953 tests passed.
- Browser verification at `https://activities.local/notifications` as `@ride@llun.social`: grouped import row renders generic fitness import copy, stale import rows render explicit unavailable text, and the fresh route console has no errors or warnings.

## Notes

Notification preview snippets are intentionally not full-card links. Each notification keeps an explicit `post`, `View activity`, or `View latest activity` link; making the whole preview clickable would reintroduce nested or competing interactive elements because processed status text can render links.

The local stack is running `activitynext-dev` from the Node 24 dev container. `docker compose logs activitynext-dev` showed `/notifications` served from that container; the same tail also included an older timeline hydration warning from `/`, unrelated to this notifications route check.

## Review Follow-up

- Removed unused `groupedAccounts` hydration from the notifications UI path and made follow grouped copy rely on `groupedCount` like the other grouped notification rows.
- Removed the self-avatar from `activity_import` notifications.
- Collapsed repeated status-required fallback guards and made type-only client imports explicit.
- Latest verification: focused notification tests passed with 18 tests, `yarn lint` passed with the same 24 existing warnings, `yarn build` passed, full Jest passed with 260 suites / 1,955 tests, and browser verification at `https://activities.local/notifications` showed the import row with no images and no fresh route console warnings/errors.

## Latest Review Follow-up

- Removed unused `groupedAccounts` hydration from the notifications UI request path, avoiding up to 3 extra actor account lookups per grouped notification while keeping grouped copy based on `groupedCount`.
- Kept `groupedActors` in the grouping service because the Mastodon API notification path still uses it to populate `grouped_accounts`.
- Fixed `activity_import` row alignment in cff9c25cfe5213298147863682cde9b63641ec2f by using the shared notification row shell with a non-account fitness glyph gutter.
- Added an exhaustiveness guard to the notification type switch in cff9c25cfe5213298147863682cde9b63641ec2f while preserving the runtime fallback for unknown or legacy rows.
- Latest verification: focused notification tests passed with 18 tests, `yarn lint` passed with the same 24 existing warnings, `yarn build` passed, full Jest passed with 260 suites / 1,955 tests, and browser verification at `https://activities.local/notifications` showed the import row aligned with visible copy and unavailable fallback rows.

### DIFF
--- a/app/(timeline)/notifications/NotificationItem.test.tsx
+++ b/app/(timeline)/notifications/NotificationItem.test.tsx
@@ -47,6 +47,12 @@ const account: Mastodon.Account = {
   following_count: 0
 }
 
+const accountWithAvatar: Mastodon.Account = {
+  ...account,
+  avatar: 'https://llun.social/avatar.png',
+  avatar_static: 'https://llun.social/avatar.png'
+}
+
 const status: StatusNote = {
   id: 'https://llun.social/users/ride/statuses/activity-1',
   actorId: 'https://llun.social/users/ride',
@@ -118,7 +124,7 @@ describe('NotificationItem', () => {
       isRead: true,
       createdAt: currentTime,
       updatedAt: currentTime,
-      account,
+      account: accountWithAvatar,
       status
     })
 
@@ -137,6 +143,7 @@ describe('NotificationItem', () => {
     container.querySelectorAll('a').forEach((link) => {
       expect(link.querySelector('a')).toBeNull()
     })
+    expect(container.querySelector('img')).toBeNull()
   })
 
   it('renders grouped activity import notifications as multiple imports', () => {
@@ -215,6 +222,24 @@ describe('NotificationItem', () => {
     }
   )
 
+  it('renders grouped follow notifications without grouped account data', () => {
+    renderNotificationItem({
+      id: 'notification-follow',
+      actorId: 'https://llun.social/users/llun',
+      type: 'follow',
+      sourceActorId: account.id,
+      isRead: true,
+      createdAt: currentTime,
+      updatedAt: currentTime,
+      account,
+      groupedCount: 3
+    })
+
+    expect(
+      screen.getByText(/and 2 others started following you/)
+    ).toBeInTheDocument()
+  })
+
   it('renders grouped activity import notifications with generic import copy', () => {
     renderNotificationItem({
       id: 'notification-activity-import',
@@ -235,24 +260,31 @@ describe('NotificationItem', () => {
     ).toBeInTheDocument()
   })
 
-  it('renders unavailable copy for stale activity import notifications', () => {
-    renderNotificationItem({
-      id: 'notification-3',
-      actorId: account.id,
-      type: 'activity_import',
-      sourceActorId: account.id,
-      statusId: status.id,
-      isRead: true,
-      createdAt: currentTime,
-      updatedAt: currentTime,
-      account,
-      status: null
-    })
+  it.each([
+    ['activity_import', 'This imported activity is no longer available.'],
+    ['like', 'This post is no longer available.'],
+    ['reply', 'This post is no longer available.'],
+    ['mention', 'This post is no longer available.'],
+    ['reblog', 'This post is no longer available.']
+  ] as const)(
+    'renders unavailable status text for stale %s notifications',
+    (type, expectedText) => {
+      renderNotificationItem({
+        id: `notification-stale-${type}`,
+        actorId: account.id,
+        type,
+        sourceActorId: account.id,
+        statusId: status.id,
+        isRead: true,
+        createdAt: currentTime,
+        updatedAt: currentTime,
+        account,
+        status: null
+      })
 
-    expect(
-      screen.getByText('This imported activity is no longer available.')
-    ).toBeInTheDocument()
-  })
+      expect(screen.getByText(expectedText)).toBeInTheDocument()
+    }
+  )
 
   it('observes unread stale status-backed notifications so they can be marked read', () => {
     const observeElement = jest.fn()
@@ -307,47 +339,6 @@ describe('NotificationItem', () => {
       ).toBeInTheDocument()
     }
   )
-
-  it.each(['like', 'reply'] as const)(
-    'renders unavailable post text for %s notifications with missing statuses',
-    (type) => {
-      renderNotificationItem({
-        id: `notification-${type}`,
-        actorId: account.id,
-        type,
-        sourceActorId: account.id,
-        statusId: status.id,
-        isRead: true,
-        createdAt: currentTime,
-        updatedAt: currentTime,
-        account,
-        status: null
-      })
-
-      expect(
-        screen.getByText('This post is no longer available.')
-      ).toBeInTheDocument()
-    }
-  )
-
-  it('renders unavailable post text for reblog notifications with missing statuses', () => {
-    renderNotificationItem({
-      id: 'notification-reblog',
-      actorId: account.id,
-      type: 'reblog',
-      sourceActorId: account.id,
-      statusId: status.id,
-      isRead: true,
-      createdAt: currentTime,
-      updatedAt: currentTime,
-      account,
-      status: null
-    })
-
-    expect(
-      screen.getByText('This post is no longer available.')
-    ).toBeInTheDocument()
-  })
 
   it('renders unavailable notification text for unknown notification types', () => {
     renderNotificationItem({
--- a/app/(timeline)/notifications/NotificationItem.test.tsx
+++ b/app/(timeline)/notifications/NotificationItem.test.tsx
@@ -1,0 +1,215 @@
+/**
+ * @jest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { render, screen } from '@testing-library/react'
+
+import { Mastodon } from '@/lib/types/activitypub'
+import { StatusNote, StatusType } from '@/lib/types/domain/status'
+
+import { NotificationItem } from './NotificationItem'
+
+const currentTime = new Date('2026-05-10T09:19:26.175Z').getTime()
+
+const account: Mastodon.Account = {
+  id: 'https://llun.social/users/ride',
+  username: 'ride',
+  acct: 'ride@llun.social',
+  url: 'https://llun.social/@ride',
+  display_name: 'Ride',
+  note: '',
+  avatar: '',
+  avatar_static: '',
+  header: '',
+  header_static: '',
+  locked: false,
+  source: {
+    note: '',
+    fields: [],
+    privacy: 'public',
+    sensitive: false,
+    language: 'en',
+    follow_requests_count: 0
+  },
+  fields: [],
+  emojis: [],
+  bot: false,
+  group: false,
+  discoverable: true,
+  created_at: '2026-01-01T00:00:00.000Z',
+  last_status_at: '2026-05-10',
+  statuses_count: 1,
+  followers_count: 0,
+  following_count: 0
+}
+
+const status: StatusNote = {
+  id: 'https://llun.social/users/ride/statuses/activity-1',
+  actorId: 'https://llun.social/users/ride',
+  actor: {
+    id: 'https://llun.social/users/ride',
+    username: 'ride',
+    domain: 'llun.social',
+    name: 'Ride',
+    followersUrl: 'https://llun.social/users/ride/followers',
+    inboxUrl: 'https://llun.social/users/ride/inbox',
+    sharedInboxUrl: 'https://llun.social/inbox',
+    followingCount: 0,
+    followersCount: 0,
+    statusCount: 1,
+    lastStatusAt: null,
+    createdAt: currentTime
+  },
+  to: [],
+  cc: [],
+  edits: [],
+  isLocalActor: true,
+  createdAt: currentTime,
+  updatedAt: currentTime,
+  type: StatusType.enum.Note,
+  url: 'https://llun.social/@ride/activity-1',
+  text: 'Morning Ride\n20.6 km in 53:09 https://www.strava.com/activities/123',
+  summary: null,
+  reply: '',
+  replies: [],
+  actorAnnounceStatusId: null,
+  isActorLiked: false,
+  totalLikes: 0,
+  totalShares: 0,
+  attachments: [],
+  tags: []
+}
+
+describe('NotificationItem', () => {
+  it('renders activity import notifications with linked activity content', () => {
+    const { container } = render(
+      <NotificationItem
+        notification={{
+          id: 'notification-1',
+          actorId: account.id,
+          type: 'activity_import',
+          sourceActorId: account.id,
+          statusId: status.id,
+          isRead: true,
+          createdAt: currentTime,
+          updatedAt: currentTime,
+          account,
+          status
+        }}
+        currentActorId={account.id}
+        host="llun.social"
+        isRead={true}
+        observeElement={jest.fn()}
+      />
+    )
+
+    expect(
+      screen.getByText(/Your Strava fitness activity was imported/)
+    ).toBeInTheDocument()
+    expect(screen.getByRole('link', { name: 'View activity' })).toHaveAttribute(
+      'href',
+      expect.stringContaining('/@ride@llun.social/')
+    )
+    expect(screen.getByText(/Morning Ride/)).toBeInTheDocument()
+    expect(screen.getByRole('link', { name: /strava.com/ })).toHaveAttribute(
+      'href',
+      'https://www.strava.com/activities/123'
+    )
+    container.querySelectorAll('a').forEach((link) => {
+      expect(link.querySelector('a')).toBeNull()
+    })
+  })
+
+  it('renders grouped activity import notifications as multiple imports', () => {
+    render(
+      <NotificationItem
+        notification={{
+          id: 'notification-1',
+          actorId: account.id,
+          type: 'activity_import',
+          sourceActorId: account.id,
+          statusId: status.id,
+          isRead: true,
+          createdAt: currentTime,
+          updatedAt: currentTime,
+          account,
+          status,
+          groupedCount: 2,
+          groupedIds: ['notification-1', 'notification-2']
+        }}
+        currentActorId={account.id}
+        host="llun.social"
+        isRead={true}
+        observeElement={jest.fn()}
+      />
+    )
+
+    expect(
+      screen.getByText(/Your Strava fitness activities were imported/)
+    ).toBeInTheDocument()
+    expect(
+      screen.getByRole('link', { name: 'View latest activity' })
+    ).toHaveAttribute('href', expect.stringContaining('/@ride@llun.social/'))
+  })
+
+  it('renders reblog notifications instead of an empty card', () => {
+    render(
+      <NotificationItem
+        notification={{
+          id: 'notification-2',
+          actorId: 'https://llun.social/users/llun',
+          type: 'reblog',
+          sourceActorId: account.id,
+          statusId: status.id,
+          isRead: true,
+          createdAt: currentTime,
+          updatedAt: currentTime,
+          account,
+          status
+        }}
+        currentActorId="https://llun.social/users/llun"
+        host="llun.social"
+        isRead={true}
+        observeElement={jest.fn()}
+      />
+    )
+
+    expect(screen.getByText(/reblogged your/)).toBeInTheDocument()
+    expect(screen.getByRole('link', { name: 'Ride' })).toHaveAttribute(
+      'href',
+      '/@ride@llun.social'
+    )
+    expect(screen.getByRole('link', { name: 'post' })).toHaveAttribute(
+      'href',
+      expect.stringContaining('/@ride@llun.social/')
+    )
+    expect(screen.getByText(/Morning Ride/)).toBeInTheDocument()
+  })
+
+  it('renders stale status-backed notifications so they can be marked read', () => {
+    render(
+      <NotificationItem
+        notification={{
+          id: 'notification-3',
+          actorId: account.id,
+          type: 'activity_import',
+          sourceActorId: account.id,
+          statusId: status.id,
+          isRead: true,
+          createdAt: currentTime,
+          updatedAt: currentTime,
+          account,
+          status: null
+        }}
+        currentActorId={account.id}
+        host="llun.social"
+        isRead={true}
+        observeElement={jest.fn()}
+      />
+    )
+
+    expect(
+      screen.getByText('This imported activity is no longer available.')
+    ).toBeInTheDocument()
+  })
+})
--- a/app/(timeline)/notifications/NotificationItem.test.tsx
+++ b/app/(timeline)/notifications/NotificationItem.test.tsx
@@ -85,13 +85,24 @@ const status: StatusNote = {
 }
 
 const renderNotificationItem = (notification: NotificationItemNotification) => {
+  return renderNotificationItemWithOptions(notification)
+}
+
+const renderNotificationItemWithOptions = (
+  notification: NotificationItemNotification,
+  options: {
+    currentActorId?: string
+    isRead?: boolean
+    observeElement?: (element: HTMLElement | null) => void
+  } = {}
+) => {
   return render(
     <NotificationItem
       notification={notification}
-      currentActorId={notification.actorId}
+      currentActorId={options.currentActorId ?? notification.actorId}
       host="llun.social"
-      isRead={true}
-      observeElement={jest.fn()}
+      isRead={options.isRead ?? true}
+      observeElement={options.observeElement ?? jest.fn()}
     />
   )
 }
@@ -112,7 +123,7 @@ describe('NotificationItem', () => {
     })
 
     expect(
-      screen.getByText(/Your Strava fitness activity was imported/)
+      screen.getByText(/Your fitness activity was imported/)
     ).toBeInTheDocument()
     expect(screen.getByRole('link', { name: 'View activity' })).toHaveAttribute(
       'href',
@@ -145,7 +156,7 @@ describe('NotificationItem', () => {
     })
 
     expect(
-      screen.getByText(/Your Strava fitness activities were imported/)
+      screen.getByText(/Your fitness activities were imported/)
     ).toBeInTheDocument()
     expect(
       screen.getByRole('link', { name: 'View latest activity' })
@@ -178,11 +189,37 @@ describe('NotificationItem', () => {
     expect(screen.getByText(/Morning Ride/)).toBeInTheDocument()
   })
 
-  it('renders grouped reblog notifications without grouped account data', () => {
+  it.each([
+    ['like', /and 2 others liked your/],
+    ['reply', /and 2 others replied to your/],
+    ['mention', /and 2 others mentioned you in a/],
+    ['reblog', /and 2 others reblogged your/]
+  ] as const)(
+    'renders grouped %s notifications without grouped account data',
+    (type, expectedText) => {
+      renderNotificationItem({
+        id: `notification-${type}`,
+        actorId: 'https://llun.social/users/llun',
+        type,
+        sourceActorId: account.id,
+        statusId: status.id,
+        isRead: true,
+        createdAt: currentTime,
+        updatedAt: currentTime,
+        account,
+        status,
+        groupedCount: 3
+      })
+
+      expect(screen.getByText(expectedText)).toBeInTheDocument()
+    }
+  )
+
+  it('renders grouped activity import notifications with generic import copy', () => {
     renderNotificationItem({
-      id: 'notification-2',
-      actorId: 'https://llun.social/users/llun',
-      type: 'reblog',
+      id: 'notification-activity-import',
+      actorId: account.id,
+      type: 'activity_import',
       sourceActorId: account.id,
       statusId: status.id,
       isRead: true,
@@ -193,10 +230,12 @@ describe('NotificationItem', () => {
       groupedCount: 3
     })
 
-    expect(screen.getByText(/and 2 others reblogged your/)).toBeInTheDocument()
+    expect(
+      screen.getByText(/Your fitness activities were imported/)
+    ).toBeInTheDocument()
   })
 
-  it('renders stale status-backed notifications so they can be marked read', () => {
+  it('renders unavailable copy for stale activity import notifications', () => {
     renderNotificationItem({
       id: 'notification-3',
       actorId: account.id,
@@ -213,6 +252,38 @@ describe('NotificationItem', () => {
     expect(
       screen.getByText('This imported activity is no longer available.')
     ).toBeInTheDocument()
+  })
+
+  it('observes unread stale status-backed notifications so they can be marked read', () => {
+    const observeElement = jest.fn()
+
+    renderNotificationItemWithOptions(
+      {
+        id: 'notification-unread-stale',
+        actorId: account.id,
+        type: 'activity_import',
+        sourceActorId: account.id,
+        statusId: status.id,
+        isRead: false,
+        createdAt: currentTime,
+        updatedAt: currentTime,
+        account,
+        status: null
+      },
+      {
+        isRead: false,
+        observeElement
+      }
+    )
+
+    expect(
+      screen.getByText('This imported activity is no longer available.')
+    ).toBeInTheDocument()
+    expect(observeElement).toHaveBeenCalledTimes(1)
+    expect(observeElement.mock.calls[0][0]).toHaveAttribute(
+      'data-notification-id',
+      'notification-unread-stale'
+    )
   })
 
   it.each(['like', 'reply'] as const)(
--- a/app/(timeline)/notifications/NotificationItem.test.tsx
+++ b/app/(timeline)/notifications/NotificationItem.test.tsx
@@ -3,6 +3,7 @@
  */
 import '@testing-library/jest-dom'
 import { render, screen } from '@testing-library/react'
+import { type ComponentProps } from 'react'
 
 import { Mastodon } from '@/lib/types/activitypub'
 import { StatusNote, StatusType } from '@/lib/types/domain/status'
@@ -10,6 +11,9 @@ import { StatusNote, StatusType } from '@/lib/types/domain/status'
 import { NotificationItem } from './NotificationItem'
 
 const currentTime = new Date('2026-05-10T09:19:26.175Z').getTime()
+type NotificationItemNotification = ComponentProps<
+  typeof NotificationItem
+>['notification']
 
 const account: Mastodon.Account = {
   id: 'https://llun.social/users/ride',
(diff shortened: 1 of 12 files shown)


llun (commented):
Overall direction is good — moving the unavailable-notification fallback into the render layer rather than the filter layer is the right call, and the nested-anchor fix is correct (both `LikeNotification` and `ReplyNotification` already have an inline "post" link, so removing the `<Link>` wrapper from the preview box is safe).

A few issues to fix before merging:

---

### Bug — `ReblogNotification`: grouped-text condition is wrong

```tsx
// current
{hasMultiple && groupedAccounts ? (
  <p>… X and Y others reblogged …</p>
) : (
  <p>… reblogged your post</p>
)}
```

`groupedAccounts` is optional (`(Mastodon.Account | null)[] | null | undefined`), so when `groupedCount > 1` but `groupedAccounts` is absent the UI silently falls through to the singular form. Every other component (`LikeNotification`, `ReplyNotification`) uses only `hasMultiple` for the text branch. Change to:

```tsx
{hasMultiple ? (
  ...
) : (
  ...
)}
```

---

### Dead code — `if (!status.actor) return null` in both new components

`ActivityImportNotification` and `ReblogNotification` both guard with:

```tsx
if (!status.actor) return null
```

But their prop type is `status: Status` (non-nullable), and the caller (`NotificationItem`) only reaches these components via `notificationWithStatus`, which is already gated on `status?.actor` being truthy:

```tsx
const notificationWithStatus = notificationWithAccount.status?.actor
  ? (notificationWithAccount as ...)
  : null
```

This guard can never fire. Remove it from both components.

---

### Dead code — `if (!content)` in `NotificationItem`

```tsx
const content = renderNotification()
if (!content) {
  return null
}
```

After the refactor every branch of `renderNotification` returns JSX — the `default` case and all early-exit paths call `renderUnavailableNotification(...)`. The null check is unreachable. Remove it.

---

### Code quality — `renderUnavailableNotification` should be module-level

```tsx
// inside the component body
const renderUnavailableNotification = (message: string) => (
  <div className="text-sm text-muted-foreground">{message}</div>
)
```

This helper captures no props or state, so defining it inside the component recreates a new function reference on every render for no reason. Hoist it above the component as a plain function or small component.

---

### Code quality — Strava hardcoded in a generic notification type

The notification type is named `activity_import` (generic), but `ActivityImportNotification` unconditionally renders "Your Strava fitness activity was imported." If the import pipeline ever supports other sources this string will be wrong. Either rename the type to `strava_import` to make the scope explicit, or accept a `service` field for the label. At minimum add a comment explaining this is intentionally Strava-only.

---

### Tests — missing coverage for new fallback paths

The tests cover `activity_import` with `status: null` (stale-status path), but the following new branches in `NotificationItem` are untested:

- `like` or `reply` with a null account → "This notification is no longer available."
- `like` or `reply` with a null status → "This post is no longer available."
- `reblog` with a null status → same
- The `default` branch (unknown notification type) → "This notification is no longer available."

Without these, regressions in the fallback logic won't be caught.

---

app/(timeline)/notifications/components/LikeNotification.tsx
```diff
@@ -72,14 +72,11 @@ export const LikeNotification: FC<Props> = ({ host, notification }) => {
             </Link>
           </p>
         )}
-        <Link
-          href={statusUrl}
-          className="mt-2 block rounded-md bg-muted/50 p-2 text-xs text-muted-foreground hover:bg-muted"
-        >
+        <div className="mt-2 block rounded-md bg-muted/50 p-2 text-xs text-muted-foreground">
```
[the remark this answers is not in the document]
llun: No code change needed for this thread. `LikeNotification` already keeps an inline `post` link to `statusUrl` in both text branches, while the preview is intentionally a non-link container to avoid nested anchors. Verified unchanged after review commit `e8f0980f`.

app/(timeline)/notifications/components/ReplyNotification.tsx
```diff
@@ -72,14 +72,11 @@ export const ReplyNotification: FC<Props> = ({ host, notification }) => {
             </Link>
           </p>
         )}
-        <Link
-          href={statusUrl}
-          className="mt-2 block rounded-md bg-muted/50 p-2 text-xs text-muted-foreground hover:bg-muted"
-        >
+        <div className="mt-2 block rounded-md bg-muted/50 p-2 text-xs text-muted-foreground">
```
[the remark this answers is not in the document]
llun: No code change needed for this thread. `ReplyNotification` already keeps an inline `post` link to `statusUrl` in both text branches, while the preview is intentionally a non-link container to avoid nested anchors. Verified unchanged after review commit `e8f0980f`.

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -68,44 +91,64 @@ export const NotificationItem = ({
...
+        return renderUnavailableNotification(
+          'This notification is no longer available.'
+        )
     }
   }
 
+  const content = renderNotification()
+  if (!content) {
+    return null
+  }
```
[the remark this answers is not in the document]
llun: Fixed in `e8f0980f`: removed the unreachable `if (!content) return null` guard. The same commit also hoists the unavailable-notification helper and expands fallback coverage.

llun:
Addressed the non-inline review feedback in `e8f0980f`.

Changes made:
- `ReblogNotification` now uses `hasMultiple` directly, so grouped reblogs render plural text even without `groupedAccounts`.
- Removed the dead `status.actor` guards by narrowing status to a `StatusWithActor` before rendering the new notification components.
- Removed the unreachable `if (!content) return null` branch.
- Hoisted the unavailable-notification helper out of the component body.
- Added fallback tests for null account, null status, reblog missing status, and unknown notification type. The notification test now has 11 passing cases.

For the Strava label: I kept the `activity_import` UI text Strava-specific and added a comment because this notification is currently emitted only by the Strava importer. A future generic importer should add source metadata for the label rather than renaming the existing notification type in this PR.

Verification after the review fix:
- `yarn test --runTestsByPath 'app/(timeline)/notifications/NotificationItem.test.tsx' --runInBand` - 11 passed.
- `yarn lint` - 0 errors, existing 24 warnings.
- `yarn build` - passed.
- `yarn test --runInBand --modulePathIgnorePatterns='<rootDir>/.claude' --testPathIgnorePatterns='<rootDir>/.claude'` - 260 suites / 1,948 tests passed.
- Browser checked `https://activities.local/notifications` as `@ride@llun.social`; route rendered expected rows and had no fresh console warnings/errors.

llun:
@codex review

app/(timeline)/notifications/components/ActivityImportNotification.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export const ActivityImportNotification: FC<Props> = ({
+  host,
+  notification
+}) => {
+  const { account, status } = notification
+
+  const hasMultiple = notification.groupedCount && notification.groupedCount > 1
+  // activity_import is currently emitted only by the Strava importer.
+  const serviceLabel = 'Strava'
```
llun: `serviceLabel` is hard-coded to `'Strava'` with only a code comment as the contract ("activity_import is currently emitted only by the Strava importer"). The moment another importer emits an `activity_import` notification, every row will silently mislabel itself. Please derive the label from data on the notification/status (e.g. a `service`/`provider` field on the import, or the status' attribution/tag) instead of hard-coding it — even an explicit fallback like "fitness activity" when the service can't be determined would be safer than always saying "Strava".

---

app/(timeline)/notifications/components/ReblogNotification.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Image
+            src={account.avatar}
+            alt={account.display_name || account.username}
+            fill
+            className="rounded-full object-cover"
+          />
+        )}
+      </div>
+      <div className="flex-1 min-w-0">
+        {hasMultiple ? (
```
llun, line 34: Inconsistent with `LikeNotification` and `MentionNotification`, which gate the grouped-rendering branch on `hasMultiple && groupedAccounts` and fall back to the single-actor copy when `groupedAccounts` isn't populated. Here only `hasMultiple` is checked, so a grouped reblog without `groupedAccounts` will render "X and N others reblogged" while the like/mention paths would render the singular form for the same input. Please either align this with the existing components or, if the new behavior is the intended one, update Like/Mention to match so all three render consistently. The interface even declares `groupedAccounts?` but it's never used here.

---

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -45,18 +67,23 @@ export const NotificationItem = ({
...
-    ...notification,
-    account: notification.account,
-    status: notification.status ?? undefined
-  }
+    const notificationWithAccount: NotificationWithData = {
+      ...notification,
+      account: notification.account,
+      status: notification.status ?? undefined
+    }
+
```
llun: The `as NotificationWithStatus` cast leans on the runtime check above, but TypeScript can't see that the cast is sound — `notificationWithAccount.status` is still typed as `Status | undefined` after the check. Consider extracting a small type guard so the narrowing is checked rather than asserted, e.g.:

```ts
const hasStatusActor = (
  n: NotificationWithData
): n is NotificationWithStatus => Boolean(n.status?.actor)

const notificationWithStatus = hasStatusActor(notificationWithAccount)
  ? notificationWithAccount
  : null
```

Removes the `as` and makes the invariant explicit at the only place it's established.

---

app/(timeline)/notifications/components/LikeNotification.tsx
llun, line 62: Replacing the wrapping `<Link>` with a `<div>` fixes the nested-anchor warning, but it also drops two pieces of behavior the previous UI had:

1. The whole preview block used to be clickable (it linked to `statusUrl`). Now only the inline "post" / "View activity" link is — the preview itself is inert. That is a UX regression worth confirming.
2. The `hover:bg-muted` class was removed along with the `<Link>`, so the block no longer has any hover affordance.

If the goal was just to avoid nested anchors, an alternative is to keep the preview clickable using a non-anchor element (e.g. wrap in a `<button>`/clickable card with a programmatic navigation, or render the inner text without internal `<a>` tags). If the intent is genuinely "preview is no longer clickable", please call that out in the PR description, and the `hover:` styling on the inner block should probably go from the other components too. Same comment applies to the identical changes in `MentionNotification.tsx` and `ReplyNotification.tsx`.

---

app/(timeline)/notifications/NotificationItem.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+      statusId: status.id,
+      isRead: true,
+      createdAt: currentTime,
+      updatedAt: currentTime,
+      account,
+      status,
+      groupedCount: 3
+    })
+
+    expect(screen.getByText(/and 2 others reblogged your/)).toBeInTheDocument()
```
llun: The test name says "so they can be marked read" but it passes `isRead={true}` and only asserts on the rendered text. The "marked read" path (the `useEffect` that calls `observeElement` when `!isRead`) isn't actually exercised here, so the test doesn't prove what its name claims.

Either rename the test to match what it verifies (e.g. "renders unavailable copy for stale activity_import notifications"), or extend it to assert that `observeElement` is called when `isRead={false}` for an unavailable notification — that's the actual behavior change motivating this PR ("so unread rows remain visible and can be marked read"). A regression where stale unread rows stop being observed wouldn't be caught by the current assertions.

---

app/(timeline)/notifications/page.tsx
```diff
@@ -90,31 +90,20 @@ const Page = async ({ searchParams }: Props) => {
...
-  })
-
   return (
     <div className="space-y-6">
       <div>
         <h1 className="text-2xl font-semibold">Notifications</h1>
       </div>
 
-      {filteredNotifications.length === 0 ? (
+      {notificationsWithData.length === 0 ? (
```
llun, line 85: Now that `notificationsWithData` is no longer filtered, the "No notifications yet" empty state will not be shown when every notification in the list is unavailable — instead the user gets a list of "This notification is no longer available." rows. That's probably acceptable (and consistent with the PR's "keep them visible so they can be marked read" goal), but worth a moment's thought: a user who has never received a real notification but has stale rows in the DB will see a wall of "no longer available" cards instead of the friendly empty state. If that's not a real scenario in practice, ignore.

---

llun (commented):
Review of the blank-notifications fix. The core direction is right — handling `activity_import` and `reblog` in the switch and rendering an explicit "unavailable" row instead of silently dropping or returning `null` is exactly what the bug needed, and removing the nested anchors is a correct fix for the React warning. A few things worth addressing before merge:

1. **Hard-coded `serviceLabel = 'Strava'`** in `ActivityImportNotification.tsx` will silently mislabel any non-Strava importer added later. Derive it from the notification/status instead, or use a generic fallback.
2. **`ReblogNotification` grouped branch is inconsistent** with `LikeNotification`/`MentionNotification` — the others gate on `hasMultiple && groupedAccounts`, this one only gates on `hasMultiple`. Pick one and apply it everywhere.
3. **Preview block is no longer clickable** in Like/Reply/Mention (and the `hover:bg-muted` style is gone). If that's intentional, please mention it in the PR; otherwise keep the clickability with a non-anchor wrapper.
4. **`as NotificationWithStatus` cast** in `NotificationItem.tsx` is OK at runtime but unchecked at the type level. A small `is`-style type guard removes the assertion and keeps the invariant explicit.
5. **`renders stale status-backed notifications so they can be marked read`** test passes `isRead={true}`, so it doesn't actually verify the "can be marked read" behavior — either rename or extend it to assert `observeElement` is invoked for unread unavailable rows.
6. **Empty-state copy** in `page.tsx`: dropping the filter means a list made up entirely of "no longer available" rows won't show the "No notifications yet" card. Probably fine, just flagging.

No blockers, mostly polish/consistency. Nice tests and clear PR description.

---

app/(timeline)/notifications/components/ActivityImportNotification.tsx
llun: Fixed in `4c3fc1b3`: the activity-import copy is now generic fitness-activity copy instead of hard-coding Strava. The notification/status model does not currently carry provider metadata, so future provider-specific labels should come from explicit metadata rather than an importer assumption.

app/(timeline)/notifications/components/ReblogNotification.tsx
llun: Fixed in `4c3fc1b3`: `LikeNotification`, `MentionNotification`, `ReplyNotification`, and `ReblogNotification` now use `groupedCount`/`hasMultiple` consistently even when `groupedAccounts` is absent. I also added grouped fallback coverage for all four types.

app/(timeline)/notifications/NotificationItem.tsx
llun: Fixed in `4c3fc1b3`: added a shared `hasStatusActor` type guard plus notification render types in `app/(timeline)/notifications/types.ts`, and removed the `as NotificationWithStatus` cast from `NotificationItem`.

app/(timeline)/notifications/components/LikeNotification.tsx
llun: No code change here. The preview block is intentionally inert now; the explicit inline `post`, `View activity`, or `View latest activity` link remains. Restoring full-preview clickability would reintroduce nested or competing interactive elements because `processStatusText` can render anchors. I updated the PR description to call out that behavior explicitly.

app/(timeline)/notifications/NotificationItem.test.tsx
llun: Fixed in `4c3fc1b3`: renamed the stale-copy test so it only claims the rendered unavailable copy, and added an unread stale status-backed notification test that asserts `observeElement` is called with the notification row.

app/(timeline)/notifications/page.tsx
llun: No code change here. The empty state means there are zero notification records, not zero hydratable status/account rows. Keeping unavailable rows visible is intentional because unread stale notifications still need to be observable and markable as read; hiding them would bring back invisible unread notifications.

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -18,6 +20,12 @@ interface NotificationWithData extends GroupedNotification {
   groupedAccounts?: (Mastodon.Account | null)[] | null
 }
 
+type StatusWithActor = Status & { actor: NonNullable<Status['actor']> }
+
+type NotificationWithStatus = NotificationWithData & {
+  status: StatusWithActor
+}
```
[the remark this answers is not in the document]
llun: Fixed in `4c3fc1b3`: centralized `NotificationWithAccount`, `NotificationWithStatus`, `StatusWithActor`, and `hasStatusActor` in `app/(timeline)/notifications/types.ts` and updated the notification components to import those shared types.

app/(timeline)/notifications/components/ActivityImportNotification.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const statusUrl =
+    getStatusDetailPath(status) ??
+    `/${getMention(status.actor, true)}/${encodeURIComponent(status.id)}`
```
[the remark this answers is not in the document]
llun: Fixed locally in `4c3fc1b3`: added notification-local `getNotificationStatusPath` and reused it across the notification components. I intentionally did not broaden `lib/utils/getStatusDetailPath` because its `null` return is the general status-routing contract; this fallback depends on the notification render path having a `StatusWithActor`.

llun:
@codex review

app/(timeline)/notifications/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { GroupedNotification } from '@/lib/services/notifications/groupNotifications'
+import type { Mastodon } from '@/lib/types/activitypub'
+import type { Status } from '@/lib/types/domain/status'
+
+export interface NotificationWithAccount extends GroupedNotification {
+  account: Mastodon.Account
+  status?: Status
+  groupedAccounts?: (Mastodon.Account | null)[] | null
```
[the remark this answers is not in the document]
llun: Fixed in `9994a649`: `NotificationWithAccount` now preserves `status: Status | null` and `groupedAccounts: (Mastodon.Account | null)[] | null` instead of making those nullable domain fields optional.

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -45,18 +59,23 @@ export const NotificationItem = ({
...
-  // After null check, we know account is non-null
-  const notificationWithAccount: NotificationWithData = {
-    ...notification,
-    account: notification.account,
-    status: notification.status ?? undefined
-  }
+    const notificationWithAccount: NotificationWithAccount = {
+      ...notification,
+      account: notification.account,
+      status: notification.status ?? undefined
```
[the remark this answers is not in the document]
llun: Fixed in `9994a649`: `NotificationItem` now keeps the transformed status as `notification.status ?? null` and normalizes `groupedAccounts` to `null`, preserving the nullable-field convention at the render boundary.

llun:
@codex review

app/(timeline)/notifications/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { GroupedNotification } from '@/lib/services/notifications/groupNotifications'
+import type { Mastodon } from '@/lib/types/activitypub'
+import type { Status } from '@/lib/types/domain/status'
+
+export interface NotificationWithAccount extends GroupedNotification {
+  account: Mastodon.Account
+  status: Status | null
+  groupedAccounts: (Mastodon.Account | null)[] | null
```
llun: Now that none of the notification components actually read `groupedAccounts` (Like/Reply/Mention/Reblog all destructure only `{ account, status, groupedCount }` and the gate on `groupedAccounts` was removed), this required field is dead state — it's threaded through `NotificationItem` only to immediately be ignored. Either drop it from `NotificationWithAccount`/`NotificationWithStatus` and from the spread in `NotificationItem.tsx`, or wire it back into the components if you intend to render extra avatars in the grouped headers later. Right now keeping it just adds a prop to maintain.

---

app/(timeline)/notifications/components/ActivityImportNotification.tsx
```diff
@@ -0,0 +1,53 @@
...
+  notification
+}) => {
+  const { account, status } = notification
+
+  const hasMultiple = notification.groupedCount && notification.groupedCount > 1
+  const statusUrl = getNotificationStatusPath(status)
+
+  return (
+    <div className="flex items-start gap-4">
+      <div className="relative size-12 shrink-0">
```
llun: Minor UX note (not blocking): for `activity_import` the `account` is the user themselves, so this renders the viewer's own avatar next to "Your fitness activity was imported." The avatar adds little context and can read as if someone else acted on you. Consider either omitting the avatar slot for self-actions or substituting a service/import icon. Not required for this PR — flagging for follow-up.

---

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -68,44 +88,61 @@ export const NotificationItem = ({
...
+          <ReblogNotification
+            host={host}
+            notification={notificationWithStatus}
+          />
+        )
+      case 'activity_import':
+        if (!notificationWithStatus)
+          return renderUnavailableStatusNotification(notification.type)
+
+        return (
```
llun: Optional cleanup: this `if (!notificationWithStatus) return renderUnavailableStatusNotification(notification.type)` block is repeated identically in five cases (`like`, `reply`, `mention`, `reblog`, `activity_import`). You could collapse the status-required branches by checking once before the switch, e.g.:

```ts
const statusRequiredTypes = new Set([
  'like', 'reply', 'mention', 'reblog', 'activity_import'
])
if (statusRequiredTypes.has(notificationWithAccount.type) && !notificationWithStatus) {
  return renderUnavailableStatusNotification(notificationWithAccount.type)
}
```

then drop the per-case guards. Also worth considering an `assertNever`-style default so the next notification type added forces the switch to be updated rather than silently rendering "no longer available." Both are nice-to-haves; current code is correct.

---

app/(timeline)/notifications/components/LikeNotification.tsx
```diff
@@ -44,7 +31,7 @@ export const LikeNotification: FC<Props> = ({ host, notification }) => {
         )}
       </div>
       <div className="flex-1 min-w-0">
-        {hasMultiple && groupedAccounts ? (
+        {hasMultiple ? (
```
llun, line 34: Behavior change worth confirming: this used to gate on `hasMultiple && groupedAccounts`, so when a grouped notification arrived without resolved actor data the UI fell back to the singular "X liked your post" copy. With the new `hasMultiple` gate alone, the same input now renders "X and N others liked your post" using only `groupedCount`. That matches the PR's stated "consistent grouped notification fallbacks" goal and `groupNotifications.ts` only increments `groupedCount` per actual grouped row, so the count should be trustworthy. Just calling out the user-visible change in case it surprises anyone reviewing later. Same applies to `MentionNotification.tsx` and `ReplyNotification.tsx`.

---

llun (commented):
Round 2 — all prior feedback was addressed cleanly:

- ✅ Generic "fitness activity" / "fitness activities" copy replaces the hard-coded `Strava` label.
- ✅ `hasStatusActor` is a real `is`-style type guard in the new `types.ts`; the `as` cast is gone.
- ✅ Like/Reply/Mention/Reblog now use a consistent `hasMultiple` gate, with new parameterized tests covering the grouped-without-`groupedAccounts` case for all four.
- ✅ The "marked read" test is renamed and now actually asserts `observeElement` is invoked for unread stale rows.
- ✅ Shared `NotificationWithStatus` type and `getNotificationStatusPath` helper remove the per-component `getStatusDetailPath` / `getMention` boilerplate.
- ✅ The intentional "preview is not full-card link" decision is documented in the PR notes.

Remaining items are all non-blocking polish:

1. **`groupedAccounts` is dead state** — declared as required on `NotificationWithAccount`, threaded through `NotificationItem`, but no consumer reads it. Drop it from the type (and the spread) or wire it back into the headers if you mean to render extra avatars later.
2. **Repeated guard in the switch** — the same `if (!notificationWithStatus) return renderUnavailableStatusNotification(...)` appears five times. Could collapse to a single pre-switch check; an `assertNever` default would also catch new types being added without UI support.
3. **`ActivityImportNotification` avatar** — rendering the user's own avatar for a self-action ("Your fitness activity was imported.") reads slightly oddly. A service/import icon (or no avatar) would be clearer. Polish, not required here.
4. **Behavior nudge to call out** — Like/Reply/Mention previously fell back to singular copy when `groupedAccounts` was missing; they now show "and N others" purely from `groupedCount`. Matches the PR intent and `groupNotifications.ts` increments `groupedCount` correctly, just flagging the user-visible shift.

Looks good to me overall.

---

app/(timeline)/notifications/types.ts
llun: Fixed in `a475dc1a`: removed `groupedAccounts` from the notifications UI data path and stopped fetching grouped actor account records on the page. Follow notifications now use `groupedCount` for grouped copy, matching the other grouped notification components.

app/(timeline)/notifications/components/ActivityImportNotification.tsx
llun: Fixed in `a475dc1a`: `ActivityImportNotification` no longer renders the actor avatar/self-avatar slot. I added coverage using an account with an avatar to assert the activity import row renders without an image.

app/(timeline)/notifications/NotificationItem.tsx
llun: Fixed in `a475dc1a`: collapsed the repeated status-required guard into a local `renderStatusNotification` helper and expanded stale status-backed notification coverage across `activity_import`, `like`, `reply`, `mention`, and `reblog`.

app/(timeline)/notifications/components/LikeNotification.tsx
llun: No code change for the status-backed grouped copy. Confirmed intentional: `groupedCount` is the authoritative count from grouped notification rows, while grouped actor account hydration was optional display data. Falling back to singular copy when that optional lookup is absent undercounts the notification. The grouped fallback behavior remains covered for like/reply/mention/reblog, and `a475dc1a` aligns follow notifications to the same `groupedCount` behavior.

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -2,22 +2,22 @@
 
 import { useEffect, useRef } from 'react'
 
+import {
+  type NotificationWithAccount,
+  hasStatusActor
+} from '@/app/(timeline)/notifications/types'
 import { GroupedNotification } from '@/lib/services/notifications/groupNotifications'
 import { Mastodon } from '@/lib/types/activitypub'
 import { Status } from '@/lib/types/domain/status'
```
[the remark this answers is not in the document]
llun: Fixed in `a475dc1a`: `NotificationItem` now imports `GroupedNotification`, `Mastodon`, and `Status` with `import type`, keeping only the real runtime imports in the client component.

llun:
@codex review

app/(timeline)/notifications/page.tsx
```diff
@@ -68,53 +68,28 @@ const Page = async ({ searchParams }: Props) => {
...
-            notification.groupedActors
-              .slice(0, 3)
-              .map((actorId) =>
-                database.getMastodonActorFromId({ id: actorId })
-              )
-          )
-        ).filter(Boolean)
-      }
-
       return {
```
llun, line 71: Nice cleanup — dropping the per-notification `Promise.all` over `groupedActors` removes a real chunk of work from the request path (up to 3 actor lookups × N notifications). Worth noting for the PR description that this also has a perf benefit, not just a code-cleanliness one.

One follow-up: `GroupedNotification.groupedActors` and the related accumulation logic in `lib/services/notifications/groupNotifications.ts` are now unused by any UI consumer (page no longer fetches them, components no longer read them). If nothing on the server side reads them either, that field can be retired in a follow-up; otherwise, fine to leave for now.

---

app/(timeline)/notifications/components/ActivityImportNotification.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const ActivityImportNotification: FC<Props> = ({
+  host,
+  notification
+}) => {
+  const { status } = notification
+
+  const hasMultiple = notification.groupedCount && notification.groupedCount > 1
+  const statusUrl = getNotificationStatusPath(status)
+
+  return (
```
llun, line 24: Heads-up that this row now has a different shape from every other notification: Like/Reply/Mention/Reblog all wrap in `<div className="flex items-start gap-4">` with a 48px avatar gutter, but activity_import is a single-column `<div className="min-w-0">` flush to the card edge. In a mixed list the activity-import rows will visually misalign with the rest. If that's the intended look (de-emphasizing self-actions), no change needed; otherwise consider keeping the `flex items-start gap-4` shell with an empty 48px spacer, or substituting a small fitness/import glyph in the avatar slot to keep the column alignment.

---

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -45,18 +64,35 @@ export const NotificationItem = ({
...
+    const renderStatusNotification = (
+      StatusNotification: StatusNotificationComponent
+    ) => {
+      if (!notificationWithStatus) {
+        return renderUnavailableStatusNotification(notificationWithAccount.type)
+      }
+
+      return (
+        <StatusNotification host={host} notification={notificationWithStatus} />
+      )
```
llun, line 95: Nice — this collapses the five repeated guard blocks into one helper and the switch becomes purely a routing table. Two small nudges if you do another pass:

- The `default` branch silently renders "no longer available" for any new notification type. Adding an `assertNever`-style call here (or an `// @ts-expect-error if exhaustive` shape) would force the next contributor adding a new `Notification.type` value to update this switch instead of shipping a generic fallback.
- `renderStatusNotification` closes over `notificationWithStatus`, `notificationWithAccount`, and `host`, so it has to be recreated on every render. That's inconsequential here (one render per notification), but if you ever lift it out of `renderNotification`, pass those as args explicitly.

Neither blocking.

---

llun (commented):
Round 3 — all of round 2's items are in:

- ✅ `groupedAccounts` removed end-to-end (types, `NotificationItem` props, `NotificationsList`, `FollowNotification`, and the per-row `Promise.all` over `groupedActors` in `page.tsx`).
- ✅ Five repeated `if (!notificationWithStatus) return …` blocks collapsed into a single `renderStatusNotification` helper backed by a `StatusNotificationComponent` type — switch is now a clean routing table.
- ✅ `ActivityImportNotification` no longer renders the user's own avatar; new test asserts no `` even when the account has one.
- ✅ Type-only imports made explicit across the touched files.
- ✅ Test coverage expanded: parameterized stale-status table for all five status-required types, new `follow` grouped-fallback case, and the avatar-suppression check.

Remaining notes are all non-blocking polish:

1. **Visual alignment** — without the avatar, activity_import rows lose the 48px gutter the other notification rows have, so they'll not line up in a mixed list. Decide whether to keep the flex shell with a spacer/glyph or accept the asymmetry.
2. **Dead `groupedActors`** — now that the page no longer hydrates accounts from `groupedActors`, the field on `GroupedNotification` (and the accumulation in `groupNotifications.ts`) has no consumer. Could be retired in a follow-up if nothing server-side reads it.
3. **Exhaustiveness** — the `default` branch silently renders "no longer available" for unknown notification types. An `assertNever` would force future type additions through the switch instead of falling through.

Looks good — ready to merge from my side once those (or a decision to not act on them) are sorted.

---

app/(timeline)/notifications/page.tsx
llun: Updated the PR description to call out the request-path perf benefit explicitly: the UI no longer hydrates `groupedAccounts`, avoiding up to 3 extra actor account lookups per grouped notification. I left `GroupedNotification.groupedActors` in place because `lib/services/notifications/getMastodonNotification.ts` still uses it to populate Mastodon API `grouped_accounts`; removing it would be a follow-up API-contract cleanup, not part of this UI fix.

app/(timeline)/notifications/components/ActivityImportNotification.tsx
llun: Fixed in cff9c25cfe5213298147863682cde9b63641ec2f: `activity_import` now uses the same `flex items-start gap-4` row shell with a 48px non-account fitness glyph gutter, so mixed notification rows keep the same left column alignment without reintroducing the viewer self-avatar. The notification item test now asserts the row/gutter remains and that no `img` is rendered.

app/(timeline)/notifications/NotificationItem.tsx
llun: Partially fixed in cff9c25cfe5213298147863682cde9b63641ec2f: added a no-op `assertNever(notificationWithAccount.type)` in the default branch, so adding a new typed `Notification.type` fails compile until the switch is updated while runtime unknown or legacy rows still render the unavailable fallback. I did not lift `renderStatusNotification` because the closure is one small per-row helper over normalized render-local data; threading those values into a top-level helper would add parameter noise with no behavior benefit here.

llun:
@codex review

llun:
## Post-merge code review

Overall the fix is solid — the refactor to shared types in `types.ts`, the `hasStatusActor` guard, and removing the client-side filter that was hiding stale rows are all the right moves. A few things worth cleaning up in a follow-up:

---

### 1. `assertNever` is a no-op at runtime (`NotificationItem.tsx:52`)

```ts
const assertNever = (_value: never) => {}
```

This only provides TypeScript exhaustiveness checking — it does nothing at runtime. The naming implies it throws, and the conventional implementation is:

```ts
const assertNever = (value: never): never => {
  throw new Error(`Unhandled notification type: ${JSON.stringify(value)}`)
}
```

Right now, if an unknown type reaches the `default` branch at runtime (e.g., from a future API change), the function silently no-ops and then falls through to `renderUnavailableNotification`, which is a graceful degradation. That's not terrible, but development builds should throw to surface the gap early.

---

### 2. `renderUnavailableStatusNotification` takes `string` instead of the typed union (`NotificationItem.tsx:42`)

```ts
const renderUnavailableStatusNotification = (notificationType: string) => {
  if (notificationType === 'activity_import') { ...
```

The parameter should be `NotificationWithAccount['type']` (or `GroupedNotification['type']`) rather than `string`. As written, the compiler won't catch a typo in the string literal, and if a new status-backed type is added that also needs a distinct unavailability message, the untyped comparison will silently skip it.

---

### 3. New component files import `FC` as a value, not a type (`ActivityImportNotification.tsx:3`, `ReblogNotification.tsx:3`)

Both new files use:

```ts
import { FC } from 'react'
```

`NotificationItem.tsx` uses `import { type ComponentType, ... }` with the `type` modifier. For consistency and to signal intent clearly, the new files should use:

```ts
import type { FC } from 'react'
```

The pre-existing notification components (`LikeNotification`, `MentionNotification`, `ReplyNotification`) have the same issue, so this can be fixed in one pass across all notification components.

---

### 4. N+1 database queries in `page.tsx` (lines 57–77)

The `Promise.all` loop issues up to two DB queries per notification (one for `account`, one for `status`), meaning up to 50 queries at the default page size of 25. The PR removed the `groupedAccounts` lookup (good), but that was a chance to also batch the `getMastodonActorFromId` calls since all `sourceActorId` values are known upfront. Not introduced here but worth a follow-up optimisation — a batch actor fetch would cut query count significantly for busy notification feeds.

---

PR #656, opened 2026-05-11
minor: render large fitness heatmaps reliably (+ clear route cache)

llun:
## Summary

- Render large route heatmap caches with the SVG route fallback instead of sending oversized geometry through Mapbox GL.
- Set Mapbox access token/resize defensively for smaller interactive route maps.
- Stop polling stale in-progress route heatmap summaries so old staging rows do not hammer the history endpoint.

## Root Cause

The staging `@ride@llun.social` all-time heatmap has 80,000 cached route points. With a Mapbox token configured, the view chose the Mapbox branch and produced a blank map area for that large payload. The route cache list also treated six-day-old `generating` summaries as live work and kept polling `/fitness-route-heatmaps`.

## Verification

- `yarn test --runTestsByPath 'app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx' --runInBand --modulePathIgnorePatterns='.claude/worktrees'`
- `yarn lint` (passes with existing warnings)
- `yarn build`
- `yarn test` fails only because local `.claude/worktrees` tests/mocks are discovered
- `yarn test --modulePathIgnorePatterns='.claude/worktrees'`
- Browser-verified `https://activities.local/@ride@llun.social/fitness/heatmap` renders route geometry and no longer repeats heatmap summary polling after reload.

### DIFF
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -102,6 +102,50 @@ const pendingSummary: FitnessRouteHeatmapSummaryData = {
 }
 
 const TEST_NOW = 1_700_000_060_000
+const IN_FLIGHT_HISTORY_POLL_WINDOW_MS = 15 * 60_000
+
+const pendingSummaryAtAge = (
+  ageMs: number
+): FitnessRouteHeatmapSummaryData => ({
+  ...pendingSummary,
+  updatedAt: TEST_NOW - ageMs
+})
+
+const buildLargeHeatmap = (): FitnessRouteHeatmapData => {
+  const segmentCount = 80
+  const pointsPerSegment = 1000
+  const segments: FitnessRouteHeatmapData['segments'] = Array.from(
+    { length: segmentCount },
+    (_, segmentIndex) => ({
+      points: Array.from({ length: pointsPerSegment }, (_point, pointIndex) => {
+        const progress =
+          (segmentIndex * pointsPerSegment + pointIndex) /
+          (segmentCount * pointsPerSegment - 1)
+        const branchOffset = (segmentIndex % 8) * 0.0008
+
+        return {
+          lat: 52.36 + progress * 0.03 + Math.sin(pointIndex / 30) * 0.0002,
+          lng: 4.88 + branchOffset + progress * 0.03
+        }
+      })
+    })
+  )
+
+  return {
+    ...completedHeatmap,
+    id: 'route-heatmap-large',
+    bounds: {
+      minLat: 52.36,
+      maxLat: 52.39,
+      minLng: 4.88,
+      maxLng: 4.916
+    },
+    segments,
+    activityCount: segmentCount,
+    pointCount: segmentCount * pointsPerSegment,
+    updatedAt: 3
+  }
+}
 
 describe('FitnessHeatmapView', () => {
   beforeEach(() => {
@@ -121,10 +165,7 @@ describe('FitnessHeatmapView', () => {
     jest.useFakeTimers()
     jest.setSystemTime(TEST_NOW)
     mockGetFitnessRouteHeatmaps.mockResolvedValue([
-      {
-        ...pendingSummary,
-        updatedAt: TEST_NOW
-      }
+      pendingSummaryAtAge(IN_FLIGHT_HISTORY_POLL_WINDOW_MS)
     ])
 
     render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
@@ -159,10 +200,7 @@ describe('FitnessHeatmapView', () => {
     jest.useFakeTimers()
     jest.setSystemTime(TEST_NOW)
     mockGetFitnessRouteHeatmaps.mockResolvedValue([
-      {
-        ...pendingSummary,
-        updatedAt: TEST_NOW - 10 * 60_000
-      }
+      pendingSummaryAtAge(IN_FLIGHT_HISTORY_POLL_WINDOW_MS + 1)
     ])
 
     render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
@@ -291,11 +329,57 @@ describe('RouteHeatmapMap', () => {
     )
   })
 
+  it('falls back with a diagnostic reason when Mapbox setup fails', async () => {
+    const mapConstructor = jest.fn().mockImplementation(() => ({
+      on: (_event: string, callback: () => void) => callback(),
+      remove: jest.fn(),
+      resize: jest.fn(),
+      addSource: jest.fn(() => {
+        throw new Error('source unavailable')
+      }),
+      addLayer: jest.fn(),
+      getSource: jest.fn(),
+      fitBounds: jest.fn()
+    }))
+    mockLoadMapboxModule.mockResolvedValue({
+      Map: mapConstructor
+    })
+
+    const { container } = render(
+      <RouteHeatmapMap
+        heatmap={completedHeatmap}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    await waitFor(() => expect(screen.getByText('Routes')).toBeInTheDocument())
+    expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
+    expect(
+      container.querySelector('[data-mapbox-fallback-reason="render-failed"]')
+    ).toBeInTheDocument()
+  })
+
+  it('falls back with a diagnostic reason when Mapbox fails to load', async () => {
+    mockLoadMapboxModule.mockRejectedValue(new Error('module unavailable'))
+
+    const { container } = render(
+      <RouteHeatmapMap
+        heatmap={completedHeatmap}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    await waitFor(() => expect(screen.getByText('Routes')).toBeInTheDocument())
+    expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
+    expect(
+      container.querySelector(
+        '[data-mapbox-fallback-reason="module-load-failed"]'
+      )
+    ).toBeInTheDocument()
+  })
+
   it('uses the SVG route fallback for large route caches', () => {
-    const largeHeatmap = {
-      ...completedHeatmap,
-      pointCount: 80_000
-    }
+    const largeHeatmap = buildLargeHeatmap()
 
     const { container } = render(
       <RouteHeatmapMap
@@ -306,7 +390,9 @@ describe('RouteHeatmapMap', () => {
 
     expect(screen.getByText('Routes')).toBeInTheDocument()
     expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
-    expect(container.querySelector('polyline')).toBeInTheDocument()
+    expect(container.querySelectorAll('polyline')).toHaveLength(
+      largeHeatmap.segments.length
+    )
     expect(mockLoadMapboxModule).not.toHaveBeenCalled()
   })
 })
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -107,6 +107,13 @@ const pendingSummary: FitnessRouteHeatmapSummaryData = {
   updatedAt: 2
 }
 
+const completedSummary: FitnessRouteHeatmapSummaryData = {
+  ...pendingSummary,
+  id: 'route-heatmap-completed',
+  status: 'completed',
+  cursorOffset: 0
+}
+
 const TEST_NOW = 1_700_000_060_000
 const IN_FLIGHT_HISTORY_POLL_WINDOW_MS = 15 * 60_000
 
@@ -270,6 +277,32 @@ describe('FitnessHeatmapView', () => {
           periodType: 'yearly',
           periodKey: '2026',
           retry: false
+        })
+      )
+    })
+  })
+
+  it('allows manual generation retry for a missing selection when other route caches exist', async () => {
+    mockGetFitnessRouteHeatmap.mockResolvedValue(null)
+    mockGetFitnessRouteHeatmaps.mockResolvedValue([completedSummary])
+    mockTriggerFitnessRouteHeatmap
+      .mockRejectedValueOnce(new Error('queue unavailable'))
+      .mockResolvedValueOnce(true)
+
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    expect(await screen.findByText('queue unavailable')).toBeInTheDocument()
+
+    const generateButton = screen.getByRole('button', { name: /Generate/i })
+    fireEvent.click(generateButton)
+
+    await waitFor(() => {
+      expect(mockTriggerFitnessRouteHeatmap).toHaveBeenCalledTimes(2)
+      expect(mockTriggerFitnessRouteHeatmap).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          actorId: 'https://llun.test/users/llun',
+          periodType: 'all_time',
+          periodKey: 'all'
         })
       )
     })
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -101,6 +101,8 @@ const pendingSummary: FitnessRouteHeatmapSummaryData = {
   updatedAt: 2
 }
 
+const TEST_NOW = 1_700_000_060_000
+
 describe('FitnessHeatmapView', () => {
   beforeEach(() => {
     jest.clearAllMocks()
@@ -115,8 +117,53 @@ describe('FitnessHeatmapView', () => {
     jest.useRealTimers()
   })
 
-  it('keeps polling history entries without stalling a completed selection', async () => {
+  it('keeps polling fresh history entries without stalling a completed selection', async () => {
     jest.useFakeTimers()
+    jest.setSystemTime(TEST_NOW)
+    mockGetFitnessRouteHeatmaps.mockResolvedValue([
+      {
+        ...pendingSummary,
+        updatedAt: TEST_NOW
+      }
+    ])
+
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    await waitFor(() => {
+      expect(mockGetFitnessRouteHeatmap).toHaveBeenCalledTimes(1)
+      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(1)
+      expect(screen.getByText('Generating…')).toBeInTheDocument()
+    })
+
+    mockGetFitnessRouteHeatmap.mockClear()
+
+    const callsAfterInitialLoad = mockGetFitnessRouteHeatmaps.mock.calls.length
+
+    for (let i = 0; i < 2; i++) {
+      await act(async () => {
+        jest.advanceTimersByTime(5000)
+        await Promise.resolve()
+      })
+    }
+
+    expect(mockGetFitnessRouteHeatmaps.mock.calls.length).toBeGreaterThan(
+      callsAfterInitialLoad
+    )
+    expect(mockGetFitnessRouteHeatmap).not.toHaveBeenCalled()
+    expect(
+      screen.queryByText('Route cache is taking longer than expected')
+    ).not.toBeInTheDocument()
+  })
+
+  it('does not keep polling stale in-progress history entries', async () => {
+    jest.useFakeTimers()
+    jest.setSystemTime(TEST_NOW)
+    mockGetFitnessRouteHeatmaps.mockResolvedValue([
+      {
+        ...pendingSummary,
+        updatedAt: TEST_NOW - 10 * 60_000
+      }
+    ])
 
     render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
 
@@ -125,31 +172,16 @@ describe('FitnessHeatmapView', () => {
       expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(1)
     })
 
-    mockGetFitnessRouteHeatmap.mockClear()
-
     const callsAfterInitialLoad = mockGetFitnessRouteHeatmaps.mock.calls.length
 
-    for (let i = 0; i < 35; i++) {
-      await act(async () => {
-        jest.advanceTimersByTime(5000)
-        await Promise.resolve()
-      })
-    }
-    const callsAfterStallWindow = mockGetFitnessRouteHeatmaps.mock.calls.length
-
     await act(async () => {
-      jest.advanceTimersByTime(5000)
+      jest.advanceTimersByTime(15_000)
       await Promise.resolve()
     })
 
-    expect(callsAfterStallWindow).toBeGreaterThan(callsAfterInitialLoad)
-    expect(mockGetFitnessRouteHeatmaps.mock.calls.length).toBeGreaterThan(
-      callsAfterStallWindow
+    expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(
+      callsAfterInitialLoad
     )
-    expect(mockGetFitnessRouteHeatmap).not.toHaveBeenCalled()
-    expect(
-      screen.queryByText('Route cache is taking longer than expected')
-    ).not.toBeInTheDocument()
   })
 
   it('ignores stale selection responses before mutating state or enqueueing', async () => {
@@ -257,5 +289,24 @@ describe('RouteHeatmapMap', () => {
         attributionControl: true
       })
     )
+  })
+
+  it('uses the SVG route fallback for large route caches', () => {
+    const largeHeatmap = {
+      ...completedHeatmap,
+      pointCount: 80_000
+    }
+
+    const { container } = render(
+      <RouteHeatmapMap
+        heatmap={largeHeatmap}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    expect(screen.getByText('Routes')).toBeInTheDocument()
+    expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
+    expect(container.querySelector('polyline')).toBeInTheDocument()
+    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
   })
 })
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -375,6 +375,33 @@ describe('FitnessHeatmapView', () => {
     expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
   })
 
+  it('keeps page errors when the clear route cache dialog is opened and cancelled', async () => {
+    mockTriggerFitnessRouteHeatmap.mockRejectedValueOnce(
+      new Error('refresh broken')
+    )
+
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
+    })
+
+    fireEvent.click(screen.getByRole('button', { name: /Refresh/i }))
+
+    expect(await screen.findByText('refresh broken')).toBeInTheDocument()
+
+    fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+    const dialog = screen.getByRole('dialog')
+
+    expect(screen.getByText('refresh broken')).toBeInTheDocument()
+    expect(within(dialog).queryByRole('alert')).not.toBeInTheDocument()
+
+    fireEvent.click(within(dialog).getByRole('button', { name: /Cancel/i }))
+
+    expect(screen.getByText('refresh broken')).toBeInTheDocument()
+    expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
+  })
+
   it('surfaces an error when clearing route caches fails', async () => {
     mockClearFitnessRouteHeatmaps.mockRejectedValue(new Error('boom'))
 
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -419,6 +419,11 @@ describe('RouteHeatmapMap', () => {
     expect(
       container.querySelector('[data-mapbox-fallback-reason="render-failed"]')
     ).toBeInTheDocument()
+    expect(
+      container.querySelector(
+        '[data-mapbox-fallback-error="source unavailable"]'
+      )
+    ).toBeInTheDocument()
   })
 
   it('falls back with a diagnostic reason when Mapbox fails to load', async () => {
@@ -438,6 +443,70 @@ describe('RouteHeatmapMap', () => {
         '[data-mapbox-fallback-reason="module-load-failed"]'
       )
     ).toBeInTheDocument()
+    expect(
+      container.querySelector(
+        '[data-mapbox-fallback-error="module unavailable"]'
+      )
+    ).toBeInTheDocument()
+  })
+
+  it('retries Mapbox when the same route cache is regenerated', async () => {
+    const failingMapConstructor = jest.fn().mockImplementation(() => ({
+      on: (_event: string, callback: () => void) => callback(),
+      remove: jest.fn(),
+      resize: jest.fn(),
+      addSource: jest.fn(() => {
+        throw new Error('source unavailable')
+      }),
+      addLayer: jest.fn(),
+      getSource: jest.fn(),
+      fitBounds: jest.fn()
+    }))
+    mockLoadMapboxModule.mockResolvedValue({
+      Map: failingMapConstructor
+    })
+
+    const { container, rerender } = render(
+      <RouteHeatmapMap
+        heatmap={completedHeatmap}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    await waitFor(() =>
+      expect(
+        container.querySelector('[data-mapbox-fallback-reason="render-failed"]')
+      ).toBeInTheDocument()
+    )
+
+    const workingMapConstructor = jest.fn().mockImplementation(() => ({
+      on: (_event: string, callback: () => void) => callback(),
+      remove: jest.fn(),
+      resize: jest.fn(),
+      addSource: jest.fn(),
+      addLayer: jest.fn(),
+      getSource: jest.fn(),
+      fitBounds: jest.fn()
+    }))
+    mockLoadMapboxModule.mockResolvedValue({
+      Map: workingMapConstructor
+    })
+
+    rerender(
+      <RouteHeatmapMap
+        heatmap={{
+          ...completedHeatmap,
+          updatedAt: completedHeatmap.updatedAt + 1
+        }}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    await waitFor(() => expect(screen.getByText('Mapbox')).toBeInTheDocument())
+    await waitFor(() => expect(workingMapConstructor).toHaveBeenCalled())
+    expect(
+      container.querySelector('[data-mapbox-fallback-reason]')
+    ).not.toBeInTheDocument()
   })
 
   it('uses the SVG route fallback for large route caches', () => {
@@ -451,10 +520,27 @@ describe('RouteHeatmapMap', () => {
     )
 
     expect(screen.getByText('Routes')).toBeInTheDocument()
+    expect(
+      screen.getByText('Interactive map unavailable. Showing routes.')
+    ).toBeInTheDocument()
     expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
-    expect(container.querySelectorAll('polyline')).toHaveLength(
-      largeHeatmap.segments.length
-    )
+    const polylines = Array.from(container.querySelectorAll('polyline'))
+    expect(polylines).toHaveLength(largeHeatmap.segments.length)
+    expect(
+      polylines.map(
+        (polyline) =>
+          polyline.getAttribute('points')?.trim().split(/\s+/).filter(Boolean)
+            .length ?? 0
+      )
+    ).toEqual(largeHeatmap.segments.map((segment) => segment.points.length))
+    expect(
+      container.querySelector(
+        '[data-mapbox-fallback-reason="route-cache-too-large"]'
+      )
+    ).toBeInTheDocument()
+    expect(
+      container.querySelector('[data-mapbox-fallback-error]')
+    ).not.toBeInTheDocument()
     expect(mockLoadMapboxModule).not.toHaveBeenCalled()
   })
 })
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -275,10 +275,7 @@ describe('FitnessHeatmapView', () => {
     })
   })
 
-  it('clears all route caches and reloads the current selection', async () => {
-    const confirmSpy = jest
-      .spyOn(window, 'confirm')
-      .mockImplementation(() => true)
+  it('clears all route caches without immediately requeueing the current selection', async () => {
     mockClearFitnessRouteHeatmaps.mockResolvedValue(2)
     mockGetFitnessRouteHeatmap
       .mockResolvedValueOnce(completedHeatmap)
@@ -294,6 +291,8 @@ describe('FitnessHeatmapView', () => {
     })
 
     fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+    fireEvent.click(screen.getByRole('button', { name: /Clear route caches/i }))
 
     await waitFor(() => {
       expect(mockClearFitnessRouteHeatmaps).toHaveBeenCalledWith({
@@ -301,6 +300,15 @@ describe('FitnessHeatmapView', () => {
       })
       expect(mockGetFitnessRouteHeatmap).toHaveBeenCalledTimes(2)
       expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(2)
+      expect(mockTriggerFitnessRouteHeatmap).not.toHaveBeenCalled()
+    })
+
+    const generateButton = await screen.findByRole('button', {
+      name: /Generate/i
+    })
+    fireEvent.click(generateButton)
+
+    await waitFor(() => {
       expect(mockTriggerFitnessRouteHeatmap).toHaveBeenCalledWith(
         expect.objectContaining({
           actorId: 'https://llun.test/users/llun',
@@ -309,14 +317,23 @@ describe('FitnessHeatmapView', () => {
         })
       )
     })
-
-    confirmSpy.mockRestore()
   })
 
   it('does not clear route caches when confirmation is cancelled', async () => {
-    const confirmSpy = jest
-      .spyOn(window, 'confirm')
-      .mockImplementation(() => false)
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
+    })
+
+    fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+    fireEvent.click(screen.getByRole('button', { name: /Cancel/i }))
+
+    expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
+  })
+
+  it('surfaces an error when clearing route caches fails', async () => {
+    mockClearFitnessRouteHeatmaps.mockRejectedValue(new Error('boom'))
 
     render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
 
@@ -325,9 +342,13 @@ describe('FitnessHeatmapView', () => {
     })
 
     fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+    fireEvent.click(screen.getByRole('button', { name: /Clear route caches/i }))
 
-    expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
-    confirmSpy.mockRestore()
+    expect(await screen.findByText('boom')).toBeInTheDocument()
+    expect(mockClearFitnessRouteHeatmaps).toHaveBeenCalledWith({
+      actorId: 'https://llun.test/users/llun'
+    })
+    expect(mockTriggerFitnessRouteHeatmap).not.toHaveBeenCalled()
   })
 })
 
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -2,7 +2,14 @@
  * @jest-environment jsdom
  */
 import '@testing-library/jest-dom'
-import { act, fireEvent, render, screen, waitFor } from '@testing-library/react'
+import {
+  act,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+  within
+} from '@testing-library/react'
 
 import {
   clearFitnessRouteHeatmaps,
@@ -306,6 +313,9 @@ describe('FitnessHeatmapView', () => {
         })
       )
     })
+    await waitFor(() => {
+      expect(screen.queryByText('queue unavailable')).not.toBeInTheDocument()
+    })
   })
 
   it('clears all route caches without immediately requeueing the current selection', async () => {
@@ -375,13 +385,19 @@ describe('FitnessHeatmapView', () => {
     })
 
     fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
-    fireEvent.click(screen.getByRole('button', { name: /Clear route caches/i }))
+    const dialog = screen.getByRole('dialog')
+    fireEvent.click(
+      within(dialog).getByRole('button', { name: /Clear route caches/i })
+    )
 
-    expect(await screen.findByText('boom')).toBeInTheDocument()
+    expect(await within(dialog).findByRole('alert')).toHaveTextContent('boom')
     expect(mockClearFitnessRouteHeatmaps).toHaveBeenCalledWith({
       actorId: 'https://llun.test/users/llun'
     })
     expect(mockTriggerFitnessRouteHeatmap).not.toHaveBeenCalled()
+
+    fireEvent.click(within(dialog).getByRole('button', { name: /Cancel/i }))
+    expect(screen.queryByText('boom')).not.toBeInTheDocument()
   })
 })
 
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -5,6 +5,7 @@ import '@testing-library/jest-dom'
 import { act, fireEvent, render, screen, waitFor } from '@testing-library/react'
 
 import {
+  clearFitnessRouteHeatmaps,
   getDistinctFitnessActivityTypes,
   getFitnessCalendarData,
   getFitnessRouteHeatmap,
@@ -24,6 +25,7 @@ jest.mock('@/lib/utils/mapbox', () => ({
 }))
 
 jest.mock('@/lib/client', () => ({
+  clearFitnessRouteHeatmaps: jest.fn(),
   getDistinctFitnessActivityTypes: jest.fn(),
   getFitnessCalendarData: jest.fn(),
   getFitnessRouteHeatmap: jest.fn(),
@@ -34,6 +36,10 @@ jest.mock('@/lib/client', () => ({
 const mockLoadMapboxModule = loadMapboxModule as jest.MockedFunction<
   typeof loadMapboxModule
 >
+const mockClearFitnessRouteHeatmaps =
+  clearFitnessRouteHeatmaps as jest.MockedFunction<
+    typeof clearFitnessRouteHeatmaps
+  >
 const mockGetDistinctFitnessActivityTypes =
   getDistinctFitnessActivityTypes as jest.MockedFunction<
     typeof getDistinctFitnessActivityTypes
@@ -154,6 +160,7 @@ describe('FitnessHeatmapView', () => {
     mockGetFitnessCalendarData.mockResolvedValue([])
     mockGetFitnessRouteHeatmap.mockResolvedValue(completedHeatmap)
     mockGetFitnessRouteHeatmaps.mockResolvedValue([pendingSummary])
+    mockClearFitnessRouteHeatmaps.mockResolvedValue(0)
     mockTriggerFitnessRouteHeatmap.mockResolvedValue(true)
   })
 
@@ -266,6 +273,61 @@ describe('FitnessHeatmapView', () => {
         })
       )
     })
+  })
+
+  it('clears all route caches and reloads the current selection', async () => {
+    const confirmSpy = jest
+      .spyOn(window, 'confirm')
+      .mockImplementation(() => true)
+    mockClearFitnessRouteHeatmaps.mockResolvedValue(2)
+    mockGetFitnessRouteHeatmap
+      .mockResolvedValueOnce(completedHeatmap)
+      .mockResolvedValueOnce(null)
+    mockGetFitnessRouteHeatmaps
+      .mockResolvedValueOnce([pendingSummary])
+      .mockResolvedValueOnce([])
+
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
+    })
+
+    fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+
+    await waitFor(() => {
+      expect(mockClearFitnessRouteHeatmaps).toHaveBeenCalledWith({
+        actorId: 'https://llun.test/users/llun'
+      })
+      expect(mockGetFitnessRouteHeatmap).toHaveBeenCalledTimes(2)
+      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(2)
+      expect(mockTriggerFitnessRouteHeatmap).toHaveBeenCalledWith(
+        expect.objectContaining({
+          actorId: 'https://llun.test/users/llun',
+          periodType: 'all_time',
+          periodKey: 'all'
+        })
+      )
+    })
+
+    confirmSpy.mockRestore()
+  })
+
+  it('does not clear route caches when confirmation is cancelled', async () => {
+    const confirmSpy = jest
+      .spyOn(window, 'confirm')
+      .mockImplementation(() => false)
+
+    render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
+    })
+
+    fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+
+    expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
+    confirmSpy.mockRestore()
   })
 })
 
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
@@ -29,6 +29,15 @@ import {
 } from '@/lib/components/fitness/FitnessCalendarHeatmap'
 import { FitnessHeatmapList } from '@/lib/components/fitness/FitnessHeatmapList'
 import { RegionSelector } from '@/lib/components/fitness/RegionSelector'
+import { Button } from '@/lib/components/ui/button'
+import {
+  Dialog,
+  DialogContent,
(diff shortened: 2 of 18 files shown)


app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -56,14 +56,19 @@ type MapboxMap = {
...
 }
 
 const MAP_WIDTH = 960
 const MAP_HEIGHT = 560
 const MAP_PADDING = 52
 const currentYear = new Date().getUTCFullYear()
+const ROUTE_HEATMAP_POLLING_INTERVAL_MS = 5000
 const STALLED_POLLING_LIMIT = 30
+const STALE_IN_FLIGHT_HEATMAP_MS =
+  STALLED_POLLING_LIMIT * ROUTE_HEATMAP_POLLING_INTERVAL_MS
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Deriving `STALE_IN_FLIGHT_HEATMAP_MS` as `STALLED_POLLING_LIMIT * ROUTE_HEATMAP_POLLING_INTERVAL_MS` couples two unrelated concepts: the focused-selection stall detection (number of identical poll responses tolerated) and list-summary data staleness (how long since the server last touched `updatedAt`). They happen to coincide numerically here, but if anyone later bumps `STALLED_POLLING_LIMIT` to tune stall detection, they'll silently change the staleness window too.

Also worth noting: the value works out to 150 s (2.5 minutes). The PR description mentions "six-day-old `generating` summaries" as the symptom, so a 2.5-minute threshold is *way* more aggressive than needed. A genuinely slow generation that goes >2.5 min between `updatedAt` advances will get its list entry dropped from polling, and we'll only pick it up again on the next remount/refresh.

Suggest declaring this as an independent constant motivated by expected generation latency (e.g., `15 * 60_000` or similar) and adding a brief comment explaining the rationale.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -56,14 +56,19 @@ type MapboxMap = {
...
+const MAPBOX_MAX_ROUTE_POINTS = 20_000
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:93]: `MAPBOX_MAX_ROUTE_POINTS = 20_000` is a magic number with no in-code explanation. The PR description says 80k points produced a blank Mapbox map — but why 20k specifically? Was that determined empirically (e.g., "tested OK at 18k, broken at 25k") or just a 4× safety margin? Future maintainers tuning this for new datasets will have no clue.

A one-line comment with the reasoning (or a link to where the threshold was measured) would save the next person a lot of guesswork. Also consider: if the real failure mode is GeoJSON parse size rather than point count, a budget based on `segments.reduce((n, s) => n + s.points.length, 0)` may match what's actually happening — but `pointCount` may already aggregate that, in which case this is fine.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -597,8 +624,11 @@ export const FitnessHeatmapView: FC<Props> = ({
   }, [fetchData])
 
   const hasAnyListInFlight = useMemo(
-    () => heatmaps.some(isRouteHeatmapInFlight),
-    [heatmaps]
+    () =>
+      heatmaps.some((heatmap) =>
+        shouldPollRouteHeatmapSummary(heatmap, currentTime)
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:713]: `currentTime` is only refreshed every 60 s (see the `setInterval` on the `setCurrentTime` line near the top of `FitnessHeatmapView`). With `STALE_IN_FLIGHT_HEATMAP_MS = 150_000`, the effective staleness window will drift up to +60 s in practice (so 150–210 s, not a precise 2.5 min). At a 2.5-minute threshold that's a 40 % imprecision in how long we keep polling.

If the threshold stays this small, either bump the `setCurrentTime` tick to e.g. 10 s so the window means what it says, or replace the polling tick with a `setTimeout` aligned to the actual staleness boundary. If the threshold is raised to something like 15 min (see the comment on `STALE_IN_FLIGHT_HEATMAP_MS`), the 60 s drift becomes negligible and this can be ignored.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -298,25 +317,33 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
+              },
+              paint: MAPBOX_ROUTE_LINE_PAINT
+            })
+            map.fitBounds(mapBounds, { padding: 56, duration: 0 })
+            setIsMapLoaded(true)
+          } catch {
+            if (!cancelled) {
+              setMapboxFailed(true)
+            }
+          }
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Empty `catch {}` swallows the error completely. If `addSource`/`addLayer`/`fitBounds` ever throws (the actual failure mode this PR is chasing), there's no signal in the console — we just silently flip to the SVG fallback. Debugging the next variant of this bug will be painful.

At minimum, log the error in development (`if (process.env.NODE_ENV !== 'production') console.warn('Mapbox setup failed', error)`), or capture it via the same telemetry path other client errors use. Same comment applies to the `.catch(() => { ... })` on the outer promise just below — silent module-load failures are equally hard to diagnose.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -298,25 +317,33 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
-            layout: {
-              'line-cap': 'round',
-              'line-join': 'round'
-            },
-            paint: MAPBOX_ROUTE_LINE_PAINT
-          })
-          map.fitBounds(mapBounds, { padding: 56, duration: 0 })
-          setIsMapLoaded(true)
+          try {
+            map.resize()
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: `map.resize()` is called twice — once immediately after `new mapboxgl.Map(...)` and again inside the `load` handler. The first one runs synchronously after construction, before the map has rendered anything, so it's essentially a no-op against a fresh container that already has the right dimensions (thanks to the `h-[420px]` container change on line 390). The second call in the `load` handler is the one that actually matters when the container's `clientWidth/clientHeight` weren't measurable at construction time.

Suggest keeping just the resize inside `load` and dropping the synchronous one. If the doubled call is meant to defend against a specific scenario, please add a one-line comment so future readers know not to delete it.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -289,6 +307,7 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
       .then((mapboxgl) => {
         if (cancelled || !containerRef.current) return
 
+        mapboxgl.accessToken = mapboxAccessToken
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Setting `mapboxgl.accessToken` mutates module-level state on the cached `mapboxgl` object (since `loadMapboxModule` memoizes the global). Two concerns:

1. The `Map` constructor already receives `accessToken: mapboxAccessToken` two lines below, so this assignment is redundant for that call.
2. If a future feature ever uses a different token on the same page (e.g., a per-tenant token swap), the last-write-wins behavior here will silently affect other components.

If this is a workaround for a specific Mapbox version where the constructor option is ignored, please pin the rationale in a comment. Otherwise, dropping the assignment keeps the side effect out of shared state.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -257,8 +269,14 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
+  const isWithinMapboxBudget =
+    heatmap === null || heatmap.pointCount <= MAPBOX_MAX_ROUTE_POINTS
   const shouldUseMapbox = Boolean(
-    mapboxAccessToken && hasRoutes && bounds && !mapboxFailed
+    mapboxAccessToken &&
+    hasRoutes &&
+    bounds &&
+    isWithinMapboxBudget &&
+    !mapboxFailed
   )
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Nit: the continuation operands are indented 4 spaces here rather than the 6 spaces Prettier normally emits for `&&` chains inside `Boolean(...)`. The PR description says `yarn run prettier --write .` was run, so either the repo's Prettier config tolerates both, or this file was missed. Worth a `yarn run prettier --check app/...` to confirm CI will pass.

Also, a minor readability point: `heatmap === null || heatmap.pointCount <= MAPBOX_MAX_ROUTE_POINTS` makes Mapbox the default for the `heatmap === null` case, but `hasRoutes` already short-circuits that branch (`hasRoutes` requires `status === 'completed'`). So the `heatmap === null` clause is dead code via the larger `shouldUseMapbox` expression. Consider simplifying to `heatmap !== null && heatmap.pointCount <= MAPBOX_MAX_ROUTE_POINTS` — or just inline the comparison — to remove the misleading branch.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -360,7 +387,7 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
 
   if (shouldUseMapbox) {
     return (
-      <div className="relative min-h-[420px] overflow-hidden bg-muted">
+      <div className="relative h-[420px] overflow-hidden bg-muted">
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Changing `min-h-[420px]` → `h-[420px]` for the Mapbox container is reasonable (Mapbox needs a measurable height to render), but note that the SVG branch below (line 408) still uses `min-h-[420px]`. This now-inconsistent sizing means the Mapbox map is a fixed 420 px while the SVG fallback can grow taller — users toggling between the two for the same selection will see the map area jump in height. If that's not intentional, align the two containers to the same sizing rule.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -258,4 +290,23 @@ describe('RouteHeatmapMap', () => {
...
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    expect(screen.getByText('Routes')).toBeInTheDocument()
+    expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
+    expect(container.querySelector('polyline')).toBeInTheDocument()
+    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
+  })
 })
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:643]: This test verifies the SVG fallback is *selected* for an 80k-point cache, but it asserts on a `largeHeatmap` whose `segments` still only contain a single 2-point segment (it's `{...completedHeatmap, pointCount: 80_000}`). So we're not actually exercising the slow path — we're checking that a flag flips.

The original bug was a blank map because Mapbox choked on real 80k-point geometry. The corresponding worry for the SVG branch is whether 80k points serialized into `<polyline>` strings render fast enough in a browser. A test with realistic segment volume (even synthetic — generate ~80k synthetic points across N segments) would catch the case where the SVG fallback itself becomes the new bottleneck. As-is, this test passes regardless of whether the SVG renderer can actually handle that payload.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -115,43 +117,73 @@ describe('FitnessHeatmapView', () => {
...
+
+    const callsAfterInitialLoad = mockGetFitnessRouteHeatmaps.mock.calls.length
+
+    await act(async () => {
+      jest.advanceTimersByTime(15_000)
+      await Promise.resolve()
+    })
+
+    expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(
+      callsAfterInitialLoad
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:242]: This test only advances time by 15 s and asserts the call count is unchanged. That's a *weak* assertion for "does not keep polling stale in-progress history entries" because the in-flight selection is also `completed` here (`mockGetFitnessRouteHeatmap` returns `completedHeatmap` via the default in `beforeEach`), so there's nothing driving polling regardless of the staleness logic. A passing assertion here doesn't prove staleness is what stopped the polling.

To make this a real regression test, also assert that the same setup with a *fresh* `updatedAt` (`TEST_NOW`) *does* trigger another `getFitnessRouteHeatmaps` call within the same 15 s window. The new "keeps polling fresh history entries" test above proves the positive case, but it uses a different fixture, so the two tests can both pass for unrelated reasons. Either share a helper or add a paired assertion to make the contrast tight.

---

llun (commented):
Review of the fixes for large fitness heatmap rendering and stale polling. The overall direction looks right — both the Mapbox-budget guard and the staleness check address real bugs — but a few things stand out:

**Sizing / threshold concerns**
- `STALE_IN_FLIGHT_HEATMAP_MS` is derived from the focused-poll stall limit, which conflates two unrelated knobs. The resulting 2.5-minute value is also much shorter than the symptom motivating the PR (six-day-old summaries) and may drop list polling for legitimately slow generations.
- `currentTime` ticks every 60 s, so the staleness check has up to a minute of drift on top of that threshold.
- `MAPBOX_MAX_ROUTE_POINTS = 20_000` is a magic number with no in-code rationale — the PR description doesn't explain why 20k vs. some other fraction of the failing 80k.

**Mapbox setup**
- The `mapboxgl.accessToken = ...` mutation duplicates the constructor option and writes to shared module state.
- `map.resize()` is called twice (once post-construction, once in `load`); the first is likely a no-op.
- The new `catch {}` silently swallows errors, which will make the next variant of the blank-map bug very hard to diagnose.

**Tests**
- The new "large route caches" test uses a heatmap with `pointCount: 80_000` but only one 2-point segment, so it verifies the flag flips but not that the SVG fallback can actually render the payload.
- The "stale polling" test would pass even if the staleness check did nothing, because the focused selection is `completed` and wouldn't drive polling anyway. A paired positive/negative assertion (or a shared fixture) would tighten it.

**Style**
- The `Boolean(...)` chain in `shouldUseMapbox` uses 4-space rather than the usual 6-space Prettier continuation. Worth a `yarn run prettier --check` on this file before merging.
- The Mapbox container is now `h-[420px]` while the SVG fallback is still `min-h-[420px]`; not a bug, but the heights can diverge.

Inline comments have the details. Nothing here is a hard blocker — most of these are tuning or robustness concerns — but the staleness threshold and the silent `catch` are worth addressing before this lands.

---

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b: decoupled the stale in-flight history window from the stall polling counter, raised it to 15 minutes, and added the rationale comment.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:93]: Fixed in a3c50d1b: added an in-code comment explaining why large route caches bypass Mapbox and use the SVG fallback. pointCount already aggregates the route geometry points, so it is the intended budget input here.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:713]: Addressed in a3c50d1b by raising the staleness threshold to 15 minutes. With that window, the existing 60 second currentTime tick drift is small enough that I left the timer unchanged.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b: both Mapbox failure paths now set a typed fallback reason, expose it on the SVG fallback DOM, and have tests for render setup failure and module load failure. I avoided console/logger calls because this is a client component and the repo rules forbid console.* in committed code.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b: removed the synchronous resize and kept the resize inside the Mapbox load handler.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b: removed the module-level mapboxgl.accessToken mutation and kept the constructor accessToken option only.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b: simplified the Mapbox budget guard to heatmap !== null && heatmap.pointCount <= MAPBOX_MAX_ROUTE_POINTS and verified formatting with Prettier.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in a3c50d1b and completed in b33833f4: the SVG fallback uses the same fixed height as Mapbox, and the later commit extracted the shared height class for all route-map states.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:643]: Fixed in a3c50d1b: the large-cache test now generates 80k synthetic points across 80 segments and asserts the SVG fallback emits the expected number of route polylines.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:242]: Fixed in a3c50d1b: the stale polling tests now share the same age helper and contrast the 15 minute fresh boundary against the stale boundary with the same fixture/window.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in b33833f4: extracted ROUTE_HEATMAP_MAP_HEIGHT_CLASS and applied it consistently to the Mapbox, SVG fallback, and empty route-map states.

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -301,27 +323,34 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
+              },
+              paint: MAPBOX_ROUTE_LINE_PAINT
+            })
+            map.fitBounds(mapBounds, { padding: 56, duration: 0 })
+            setIsMapLoaded(true)
+          } catch {
+            if (!cancelled) {
+              setMapboxFallbackReason('render-failed')
+            }
+          }
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: The `catch {}` still discards the underlying error completely. The new `mapboxFallbackReason` state captures *that* something failed and the data attribute lets tests assert on the category, but the actual `Error` (with message + stack) is gone before anyone can see it.

This bites when triaging real failures: a user reports "the map shows the SVG fallback", the DOM says `data-mapbox-fallback-reason="render-failed"`, but there's no way to know whether it was an out-of-memory error, an invalid GeoJSON feature, a Mapbox style fetch issue, etc.

Minimal change:

```ts
} catch (error) {
  if (!cancelled) {
    if (process.env.NODE_ENV !== 'production') {
      console.warn('Mapbox render failed', error)
    }
    setMapboxFallbackReason('render-failed')
  }
}
```

Same idea for the outer `.catch(() => { ... })` on module load. (Or capture the error into state alongside the reason, if you'd prefer to keep the console clean in development.)

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -59,11 +59,19 @@ type MapboxGL = {
...
 const MAP_WIDTH = 960
 const MAP_HEIGHT = 560
 const MAP_PADDING = 52
 const currentYear = new Date().getUTCFullYear()
+const ROUTE_HEATMAP_POLLING_INTERVAL_MS = 5000
 const STALLED_POLLING_LIMIT = 30
+// Keep recent background jobs live while ignoring restored/stuck rows that are days old.
+const STALE_IN_FLIGHT_HEATMAP_MS = 15 * 60_000
+// Mapbox GL can render smaller route sets interactively; large caches use SVG to avoid blank canvases.
+const MAPBOX_MAX_ROUTE_POINTS = 20_000
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Thanks for adding the comment. It explains *what* the constant does, but the next maintainer still has no way to know whether 20_000 is empirically validated, a guess, or a 4× safety margin off the failing 80k. If you tested it (e.g., "manual smoke at 18k OK, 25k janky"), please record that in the comment. If it was picked by intuition, a "Conservative; revisit when benchmarks are available" note is enough to keep someone from cargo-culting the number.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -405,6 +455,9 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
       <div className="absolute left-3 top-3 rounded bg-background/90 px-2 py-1 text-xs text-muted-foreground shadow-sm">
         Routes
       </div>
+      {mapboxFallbackReason ? (
+        <p className="sr-only">Interactive map unavailable. Showing routes.</p>
+      ) : null}
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:510]: The `sr-only` "Interactive map unavailable" message is only rendered when `mapboxFallbackReason` is set — i.e., when Mapbox actually tried and failed (load/render). The other path to the SVG branch is `isWithinMapboxBudget === false` (line 277), which kicks in *before* Mapbox is even attempted. For a screen-reader user, both cases look the same in the DOM (SVG with "Routes" label), but only one announces the difference.

Either:
- Add a third `MapboxFallbackReason`, e.g. `'route-cache-too-large'`, set whenever the budget cap is hit, so the sr-only branch and `data-mapbox-fallback-reason` cover all three cases uniformly; or
- Move the sr-only paragraph so it renders whenever the SVG branch is reached *and* a Mapbox token exists (i.e., the user could have seen the interactive map but didn't).

The current behavior isn't wrong, just inconsistent across two functionally-equivalent fallback paths.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -250,15 +265,22 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
+  const isWithinMapboxBudget =
+    heatmap !== null && heatmap.pointCount <= MAPBOX_MAX_ROUTE_POINTS
   const shouldUseMapbox = Boolean(
-    mapboxAccessToken && hasRoutes && bounds && !mapboxFailed
+    mapboxAccessToken &&
+    hasRoutes &&
+    bounds &&
+    isWithinMapboxBudget &&
+    !mapboxFallbackReason
   )
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Nit: still 4-space continuation rather than the 6-space hanging indent Prettier normally emits for `&&` chains inside a function call. The PR claims `prettier --write .` was run, so this either:

- passes Prettier as-is due to a config nuance I'm missing, or
- the formatter wasn't actually applied to this specific block.

A quick `yarn run prettier --check 'app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx'` will resolve it. If Prettier *does* want the 6-space form, CI will fail on this file later.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -258,4 +328,71 @@ describe('RouteHeatmapMap', () => {
...
+        heatmap={largeHeatmap}
+        mapboxAccessToken="mapbox-token"
+      />
+    )
+
+    expect(screen.getByText('Routes')).toBeInTheDocument()
+    expect(screen.queryByText('Mapbox')).not.toBeInTheDocument()
+    expect(container.querySelectorAll('polyline')).toHaveLength(
+      largeHeatmap.segments.length
+    )
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx]: Nice fixture — 80 segments × 1000 points hits the exact failure mode from the staging report. One small follow-up: this asserts `polyline` count equals `segments.length`, which proves segments-to-elements mapping but not that the points actually made it into each `polyline`. If `buildSvgMap` ever regresses (e.g., truncates point lists or filters out segments under a too-aggressive minimum length), this test still passes.

Consider also asserting on the first polyline's `points` attribute length, e.g. `expect(container.querySelector('polyline')!.getAttribute('points')!.split(' ').length).toBe(pointsPerSegment)`. Optional — not blocking.

---

llun (commented):
Second pass — most of the round-1 feedback is well addressed:

- `STALE_IN_FLIGHT_HEATMAP_MS` is now an independent constant at 15 min with a rationale comment ✅
- `mapboxgl.accessToken` mutation removed; redundant pre-`load` `resize()` removed ✅
- `isWithinMapboxBudget` dead-code branch fixed (`heatmap !== null && …`) ✅
- `mapboxFallbackReason` replacing the boolean is a real upgrade — gives both tests and the DOM a typed reason ✅
- `ROUTE_HEATMAP_MAP_HEIGHT_CLASS` removes the height duplication and keeps the four branches aligned ✅
- `buildLargeHeatmap` actually exercises the 80k-point payload that motivated the PR ✅
- The stale-polling test now does a proper boundary check (`WINDOW + 1` vs. `WINDOW`) ✅
- New tests cover the two diagnostic fallback reasons explicitly ✅

A few remaining items, mostly polish:

1. **`catch {}` still drops the error itself.** The `mapboxFallbackReason` tells you a class of failure happened, but the `Error` (message + stack) is gone. Adding `console.warn` in dev (or capturing into state) would close the debuggability gap for the same bug shape this PR is fixing. Same for the outer `.catch` on module load.
2. **`MAPBOX_MAX_ROUTE_POINTS = 20_000`** — the new comment explains *what*, not *why this number*. Even a one-liner ("conservative; not yet benchmarked" or "smoke-tested OK at 18k") would prevent future cargo-culting.
3. **`sr-only` fallback message is asymmetric.** It only renders when Mapbox actually attempted+failed, not when the budget cap (`isWithinMapboxBudget === false`) preempts Mapbox. From a screen-reader user's perspective both look the same.
4. **`Boolean(…)` continuation indentation** in `shouldUseMapbox` is still 4-space rather than Prettier's usual 6-space hanging form. Worth a `yarn run prettier --check` on this file to confirm CI is happy.
5. **Optional nit on the large-cache test** — counts `<polyline>` elements but doesn't verify point payload was preserved. A regression that drops points per segment would still pass.

None of these are blockers; (1) is the most worth doing before merge. The rest are polish.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:510]: Fixed in 4042f0be: the point-budget SVG fallback now sets route-cache-too-large as the Mapbox fallback reason, so the sr-only interactive-map-unavailable message is rendered when a Mapbox token exists but the route cache is too large.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 4042f0be: the render and module-load catch paths now preserve the underlying error in fallback state and expose the message through data-mapbox-fallback-error outside production. I avoided console.warn because this client path follows the repo no-console rule.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 4042f0be: the constant comment now records the staging 80k-point blank-canvas observation and states the 20k cap is a conservative 4x safety margin pending browser/device benchmarks.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:510]: Fixed in 4042f0be: same accessibility path as the Gemini thread; large route caches now get route-cache-too-large fallback reason and render the sr-only interactive-map-unavailable message.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Verified in 4042f0be: prettier accepts the current formatting, so no code change was needed for this thread. Command run: yarn run prettier --check app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx lib/jobs/generateFitnessRouteHeatmapJob.ts lib/jobs/generateFitnessRouteHeatmapJob.test.ts lib/database/sql/fitnessRouteHeatmap.ts lib/database/sql/fitnessRouteHeatmap.test.ts.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx]: Fixed in 4042f0be: the large-cache test now verifies each generated polyline preserves the corresponding segment point count, not only the number of polylines.

llun:
@codex review

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -59,11 +61,28 @@ type MapboxGL = {
...
+type MapboxFallbackReason =
+  | 'module-load-failed'
+  | 'render-failed'
+  | 'route-cache-too-large'
+
+interface MapboxFallbackError {
+  message: string
+  stack?: string
+}
+
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:82]: PR-level concern (not specific to this line): the PR now includes commit `6baa3d0 feat: clear fitness route heatmap cache` — a genuinely new user-facing feature — but the PR title still starts with `fix:`. Per `CLAUDE.md`:

> **For `minor` or `major` version bumps, the PR title MUST start with `minor:` or `major:`.** PRs are squash-merged, so the PR title becomes the commit subject on `main` and drives the version-bump workflow.

The PR is squash-merged, so the `feat:` commit prefix in the body is only a fallback. Please rename the PR to `minor: render large fitness heatmaps reliably (+ clear cache)` (or similar) to make sure the auto-bumper produces a `minor` release.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -794,9 +892,43 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+    setError(null)
+    try {
+      await clearFitnessRouteHeatmaps({ actorId })
+      generationKeyRef.current = null
+      pollingProgressRef.current = null
+      setHeatmapData(null)
+      setHeatmaps([])
+      setGenerationPending(false)
+      setPollingStalled(false)
+      await fetchData()
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: After `clearFitnessRouteHeatmaps` succeeds, `fetchData()` is awaited — which in turn enqueues a fresh generation via `queueCurrentRouteHeatmap` when `heatmap === null` (the expected post-clear state). So clicking **"Clear cache"** doesn't *just* clear; it also immediately kicks off a new background job for the currently-selected period.

That may be intentional UX (the user clears because they want fresh data), but it isn't what the button label or the confirm dialog ("Clear all route heatmap cache for this account…") promises. Two consequences worth thinking about:

1. A user who clicks Clear because the generator is misbehaving will trigger another generation with the same parameters, which may rehit the same bug.
2. There's no way to truly clear without retriggering — short of immediately closing the page.

Options:
- Update the button copy to "Clear and regenerate" (and update the confirm dialog) to match behavior; or
- Skip the auto-trigger here and let the user kick a new generation via the existing **Refresh** button on the heatmap.

The test on lines 278–314 currently *asserts* `triggerFitnessRouteHeatmap` is called after clear, so this is wired in as expected behavior. Either way, the behavior should match the affordance.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -794,9 +892,43 @@ export const FitnessHeatmapView: FC<Props> = ({
...
 
+  const clearRouteCache = useCallback(async () => {
+    if (isClearingCache) return
+    if (
+      !window.confirm(
+        'Clear all route heatmap cache for this account? This includes queued, generating, and failed route caches.'
+      )
+    ) {
+      return
+    }
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: `window.confirm()` blocks the JS thread, is unstyled (looks foreign next to the rest of the UI), and is inconsistent with how the rest of the app handles destructive actions. It's also tricky for screen-reader users (Chrome/Firefox dialogs are read inconsistently) and impossible to internationalize cleanly.

This dialog is the only safeguard between a click and a destructive bulk delete. Recommend swapping it for a proper modal/confirmation component if one already exists in the design system — or, if there's nothing existing, a small inline "Are you sure? [Yes / No]" toggle on the button cluster would already be a step up.

Not a blocker, but worth filing as a follow-up if you'd prefer to land the feature first.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -794,9 +892,43 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+    } catch (err) {
+      setError(
+        err instanceof Error
+          ? err.message
+          : 'Failed to clear route heatmap cache.'
+      )
+    } finally {
+      setIsClearingCache(false)
+    }
+  }, [actorId, fetchData, isClearingCache])
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: `isClearingCache` in the deps means a new `clearRouteCache` function identity is created on every toggle of the flag (i.e., twice per invocation). The early-return on line 896 already guards against re-entrancy, so this dep doesn't gain anything — it just churns the memo and any consumers that depend on the callback identity (none here, but still wasteful).

```ts
}, [actorId, fetchData])
```

is the right dep set. The early-return reads `isClearingCache` via the closure of *that specific invocation*, which is fine: the flag is set synchronously, so subsequent clicks before the state update propagates will short-circuit just the same.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -197,6 +274,61 @@ describe('FitnessHeatmapView', () => {
...
+
+    await waitFor(() => {
+      expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
+    })
+
+    fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))
+
+    expect(mockClearFitnessRouteHeatmaps).not.toHaveBeenCalled()
+    confirmSpy.mockRestore()
+  })
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:428]: No test asserts what happens when `clearFitnessRouteHeatmaps` rejects. The catch on lines 916–921 sets the `error` state from the rejection — that branch isn't exercised. Easy to add:

```ts
it('surfaces an error when clear fails', async () => {
  jest.spyOn(window, 'confirm').mockReturnValue(true)
  mockClearFitnessRouteHeatmaps.mockRejectedValue(new Error('boom'))

  render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
  await waitFor(() =>
    expect(screen.getByRole('button', { name: /Clear cache/i })).toBeEnabled()
  )
  fireEvent.click(screen.getByRole('button', { name: /Clear cache/i }))

  await waitFor(() => expect(screen.getByText('boom')).toBeInTheDocument())
  expect(mockTriggerFitnessRouteHeatmap).not.toHaveBeenCalled()
})
```

Also covers the "do not re-enqueue after a failed clear" expectation implicitly, which protects against a regression where someone moves `await fetchData()` outside the try.

---

lib/database/sql/fitnessRouteHeatmap.ts
```diff
@@ -365,6 +367,12 @@ export const FitnessRouteHeatmapSQLDatabaseMixin = (
     const query = database('fitness_route_heatmaps').where('id', id)
     if (!clearDeleted) {
       query.whereNull('deletedAt')
+    } else if (clearDeletedBefore !== undefined) {
+      query.where((builder) => {
+        builder
+          .whereNull('deletedAt')
+          .orWhere('deletedAt', '<=', new Date(clearDeletedBefore))
+      })
```
llun, line 376: Nit: with `clearDeleted=true` and `clearDeletedBefore=undefined`, the `where` chain ends up with **no `deletedAt` filter at all** — the update can touch a tombstoned row regardless of when it was deleted. That's not what any current caller does (`generateFitnessRouteHeatmapJob.ts` always passes `clearDeletedBefore: requestedAt ?? 0`), but the API allows it.

Two ways to make this less of a foot-gun:

1. Require `clearDeletedBefore` when `clearDeleted` is true (e.g., make them a discriminated union in the params type), so the unsafe combo is unreachable.
2. Or default `clearDeletedBefore` to `0` inside the function when `clearDeleted` is true, matching the job's existing fallback.

Optional — flagging because right now this is only safe by convention, not by the type system.

---

app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts
```diff
@@ -181,6 +181,7 @@ export const GET = traceApiRoute(
 export const POST = traceApiRoute(
   'triggerFitnessRouteHeatmap',
   async (req: NextRequest, params: AppRouterParams<Params>) => {
+    const requestedAt = Date.now()
```
Comment (llun) [inline app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts:184]: Two of the three POST `apiResponse` calls return inside the `catch` after `req.json()` parsing failure — those run *after* `requestedAt = Date.now()` (line 184). That's harmless, but it means `requestedAt` is captured at request entry and used for the queue payload below. If the request gets delayed by auth/parsing/etc. before queueing, the gap between `requestedAt` and `getQueue().publish(...)` can be tens of ms.

That's still strictly less than any subsequent clear's `deletedAt` (which is set *after* this whole handler runs), so the race-guard in `generateFitnessRouteHeatmapJob` still works correctly. Just noting: `requestedAt` here means "request received", not "queued at". If the semantic ever changes (e.g. backpressure delays publish by seconds), tighten this to right-before the publish.

---

llun (commented):
Round 3. The Mapbox-fallback feedback from round 2 is well addressed:

- Underlying error is now captured into state (`mapboxFallbackError`) and surfaced as `data-mapbox-fallback-error` in dev — fixes the "silent swallow" debuggability gap ✅
- `MapboxFallbackReason` gained the `'route-cache-too-large'` variant so the sr-only message and `data-mapbox-fallback-reason` are consistent across all three fallback paths ✅
- `MAPBOX_MAX_ROUTE_POINTS` comment now explains the empirical reasoning (80k blank canvases, 4× safety margin) ✅
- The new `runtimeMapboxFallbackReason` / `budgetMapboxFallbackReason` split keeps `shouldUseMapbox` honest while letting the budget cap participate in the reason surface ✅
- `Boolean(…)` chain replaced with a plain `&&` chain — sidesteps the Prettier indentation nit ✅

The PR has also grown into something bigger: a new **Clear cache** feature with a `DELETE` endpoint, the `requestedAt` race-protection field on jobs, and matching DB-layer changes. Comments below — most are about that new code:

**Most important**

1. **PR title prefix.** The PR now contains a `feat:` commit but the title is still `fix:`. Since this repo squash-merges, the title drives the version bump. Please rename to `minor: …` per `CLAUDE.md`. Comment inline on line 73 of `FitnessHeatmapView.tsx`.
2. **Clear cache auto-triggers a new generation.** `clearRouteCache → fetchData → queueCurrentRouteHeatmap` immediately re-enqueues a job for the current selection. The button label and confirm dialog only promise "clear". Either retitle the button or skip the auto-trigger. The existing test on lines 278–314 codifies the current behavior, so this is a deliberate design call worth confirming.

**Worth doing**

3. **`window.confirm()`** for a destructive bulk action is inconsistent with the rest of the UI and weak on a11y/i18n. Consider a custom modal or inline confirm.
4. **No error-path test for clear cache.** Easy to add and would lock in the "do not retrigger after a failed clear" expectation.
5. **`clearDeleted=true` + `clearDeletedBefore=undefined`** in `updateFitnessRouteHeatmapStatus` is safe-by-convention only. A discriminated-union param or an internal default to `0` would make it safe-by-construction.

**Nits**

6. **`isClearingCache` in the `useCallback` deps** is unnecessary — early-return already prevents re-entrancy, and the dep just churns the memo.
7. **`requestedAt = Date.now()`** in the POST handler is captured at request entry. Currently fine because the clear's `deletedAt` is always later, but if any future change pushes publish latency up, capture this right before `queue.publish(…)` instead.

The job-side race protection (skip-if-`existing.deletedAt > requestedAt`) and the `clearDeletedBefore` filter in `updateFitnessRouteHeatmapStatus` look correct — those plug the obvious window between enqueue and clear cleanly. Tests in `generateFitnessRouteHeatmapJob.test.ts` and the route tests cover the auth and basic happy paths.

Nothing here is blocking — the heatmap-render fix is solid. The clear-cache scope is functional but has the UX-mismatch (item 2) and the title-prefix issue (item 1) that I'd resolve before merge.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -59,11 +61,28 @@ type MapboxGL = {
...
+}
+
 const MAP_WIDTH = 960
 const MAP_HEIGHT = 560
 const MAP_PADDING = 52
 const currentYear = new Date().getUTCFullYear()
+const ROUTE_HEATMAP_POLLING_INTERVAL_MS = 5000
 const STALLED_POLLING_LIMIT = 30
+// Keep recent background jobs live while ignoring restored/stuck rows that are days old.
+const STALE_IN_FLIGHT_HEATMAP_MS = 15 * 60_000
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:90]: No code change: pending and generating are already both covered by isRouteHeatmapInFlight(), and the 15-minute stale cutoff only applies to background list polling. The focused selection uses the separate progress-based stalled polling guard, and route-generation continuations checkpoint updatedAt while work is progressing.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 782c52e6: Clear cache now uses the app Dialog/Button components instead of window.confirm, keeps the async clear action disabled while running, and keeps the existing try/catch error path covered by tests.

app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts
```diff
@@ -306,6 +307,7 @@ export const POST = traceApiRoute(
           periodType,
           periodKey,
           region,
+          requestedAt,
           ...(shouldResume
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts:316]: Fixed in 782c52e6: post-clear rebuilds now publish a restore-versioned job id when the route cache row is soft-deleted, so QStash dedupe cannot suppress the fresh rebuild behind an older pre-clear message. I covered the POST route and the fitness-file enqueue path.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:82]: Done: renamed the PR to "minor: render large fitness heatmaps reliably (+ clear route cache)" so the squash merge subject drives the minor version bump.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 782c52e6: Clear cache now clears without immediately requeueing the current selection. After the clear completes, the route cache controls expose a manual Generate action for the current selection.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 782c52e6: replaced the native confirm with the shared Dialog/Button pattern, including Cancel and destructive confirm actions, and the dialog cannot be dismissed while the clear request is in flight.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 782c52e6: clearRouteCache now uses a ref for the in-flight guard and the callback deps are reduced to actorId and fetchData.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:428]: Fixed in 782c52e6: added coverage for clearFitnessRouteHeatmaps rejecting and asserting the surfaced error without triggering route regeneration.

lib/database/sql/fitnessRouteHeatmap.ts
llun: Fixed in 782c52e6: clearDeleted without an explicit clearDeletedBefore now defaults to an epoch cutoff instead of dropping the deletedAt filter, with a regression test proving it does not revive a recently deleted row.

app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts:184]: No code change: the requestedAt timestamp is intentionally captured at request entry. If a clear happens after this request starts, that clear should invalidate this older queued request; moving requestedAt later could let a pre-clear request revive a cache after the user cleared it. The QStash dedupe issue is fixed separately by versioning post-clear job IDs.

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -794,9 +907,41 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+    }
+  }, [actorId, fetchData])
+
   const routeCount = heatmapData?.segments.length ?? 0
   const hasCompletedRoutes =
     heatmapData?.status === 'completed' && heatmapData.pointCount > 0
+  const hasRouteCache =
+    Boolean(heatmapData) || heatmaps.length > 0 || generationPending
+  const canRefreshRouteCache =
+    !isLoading && (Boolean(heatmapData) || !hasRouteCache)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in bad48a0c: the Generate/Refresh button guard now keys off current-selection loading and in-flight state instead of whether any route cache exists in history, and the regression test covers a missing current selection with other cached history after the automatic enqueue fails.

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -794,9 +907,41 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+      setGenerationPending(false)
+      setPollingStalled(false)
+      await fetchData({ queueMissing: false })
+      setIsClearCacheDialogOpen(false)
+    } catch (err) {
+      setError(
+        err instanceof Error
+          ? err.message
+          : 'Failed to clear route heatmap cache.'
+      )
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:940]: When `clearFitnessRouteHeatmaps` (or `fetchData`) rejects, `setError(...)` fires but `setIsClearCacheDialogOpen(false)` doesn't run (it's only on the success path at line 925). The error banner is rendered at the top of the page (around line 989 in `FitnessHeatmapView`), but `DialogContent` renders through a portal with the Radix overlay at `z-50` — so the banner sits *behind* the modal while the dialog is still open.

Net effect: the user clicks "Clear route caches", the API fails, the dialog stays open with nothing visibly indicating failure, the destructive button is enabled again, and only after the user manually dismisses the dialog (Escape / click outside / Cancel) do they discover what went wrong.

Two reasonable options:

1. **Close the dialog on error** as well — the page-level banner will then be visible.
   ```ts
   } catch (err) {
     setError(...)
     setIsClearCacheDialogOpen(false)
   }
   ```
2. **Render the error inside the dialog** — add a small `error` paragraph between `DialogDescription` and `DialogFooter` so the user sees the failure without leaving the modal. This is the friendlier option if you might want a follow-up action ("Retry").

The new `'surfaces an error when clearing route caches fails'` test happens to pass either way because it queries `screen.findByText('boom')`, which traverses the portaled dialog DOM too. So the bug is silent to the test suite.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1017,18 +1162,35 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-                    <RefreshCw
-                      className={cn('size-3', isRetrying && 'animate-spin')}
-                    />
-                    Refresh
-                  </button>
-                )}
+                <div className="flex items-center gap-1.5">
+                  {hasRouteCache && (
+                    <button
+                      onClick={() => setIsClearCacheDialogOpen(true)}
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Nit: the `error` state isn't cleared when the dialog is opened or canceled. If a previous Clear attempt failed (setting `error` to e.g. "boom"), then the user opens the Clear dialog again, the banner still says "boom" from the prior attempt. Same if they Cancel — the stale error persists.

A `setError(null)` on the dialog-open trigger (line 1168 `onClick`) and/or on Cancel (line 1225) would make the banner reflect only the current attempt. Optional polish.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1040,6 +1202,47 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+              onClick={clearRouteCache}
+              disabled={isClearingCache}
+            >
+              {isClearingCache ? (
+                <Trash2 className="animate-pulse" />
+              ) : (
+                <Trash2 />
+              )}
+              Clear route caches
+            </Button>
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:1255]: Nits:

- The conditional rendering of `<Trash2 />` can collapse:
  ```tsx
  <Trash2 className={isClearingCache ? 'animate-pulse' : undefined} />
  ```
- While the button is disabled and the spinner is pulsing, the label still says "Clear route caches". Adding `aria-busy={isClearingCache}` (or swapping the label to "Clearing…" while busy) would let assistive tech announce the in-progress state — useful since the pulsing icon is invisible to screen readers.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -197,6 +281,108 @@ describe('FitnessHeatmapView', () => {
...
+
+    await waitFor(() => {
+      expect(mockTriggerFitnessRouteHeatmap).toHaveBeenCalledTimes(2)
+      expect(mockTriggerFitnessRouteHeatmap).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          actorId: 'https://llun.test/users/llun',
+          periodType: 'all_time',
+          periodKey: 'all'
+        })
+      )
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:314]: Nice — the new test exercises the realistic case where an initial `triggerFitnessRouteHeatmap` from `fetchData → queueCurrentRouteHeatmap` fails, then the user manually retries via the "Generate" button and the second attempt succeeds. This locks in two behaviors that were brittle in earlier iterations:

1. A failed initial enqueue resets `generationKeyRef` so a manual retry can re-enter the enqueue path.
2. The "Generate" label appears on the refresh button when `heatmapData === null`.

Suggestion (optional): also assert the error banner clears after the successful retry — that protects against a regression where `setError(null)` in `retryCurrent`/`fetchData` accidentally gets removed.

```ts
await waitFor(() => {
  expect(screen.queryByText('queue unavailable')).not.toBeInTheDocument()
})
```

---

lib/jobs/enqueueFitnessRouteHeatmapJobs.ts
```diff
@@ -102,17 +103,29 @@ export const enqueueFitnessRouteHeatmapJobs = async ({
...
+        activityType: variant.activityType,
+        periodType: variant.periodType,
+        periodKey: variant.periodKey,
+        region: variant.region ?? '',
+        includeDeleted: true
+      })
+      const jobId = getHashFromString(
+        existing?.deletedAt
+          ? `${baseJobId}:restore:${existing.id}:${existing.deletedAt}`
+          : baseJobId
```
llun, line 128: The `:restore:${id}:${deletedAt}` suffix is a clean way to keep the queue from deduping a fresh enqueue against a still-pending tombstoned job. Two follow-up thoughts worth considering (no change required):

1. **Multiple rapid clears**: if the user clears, then clears again before the first DB write settles, both clears get distinct `deletedAt` values but `getFitnessRouteHeatmapByKey({ includeDeleted: true })` always returns the latest (highest `deletedAt`). So a subsequent POST embeds the *latest* `deletedAt` in the job ID — anything else that was already in flight is unaffected, which is what you want.
2. **`getFitnessRouteHeatmapByKey` is now called twice** in the enqueue path here (once in this function, once by the job's own existence check). On a high-volume enqueue (e.g., `recreateFitnessRouteHeatmapJobs` re-firing every variant for an actor), that doubles the query load on `fitness_route_heatmaps`. Probably fine for current scale; flagging in case it shows up in DB metrics later.

---

llun (commented):
Round 4. All round-3 items are resolved:

- PR title now `minor: render large fitness heatmaps reliably (+ clear route cache)` — auto-bumper happy ✅
- Clear cache no longer auto-regenerates; dialog explicitly says "This does not immediately start a new route cache job"; new "Generate" button surfaces when the current selection has no cache ✅
- `window.confirm` replaced with a Radix-backed `Dialog` — proper a11y, styling, focus management ✅
- New error-path test for clear-cache rejection (`'surfaces an error when clearing route caches fails'`) ✅
- `clearDeletedBefore` defaults to `0` inside `updateFitnessRouteHeatmapStatus` — safe-by-construction ✅
- `isClearingCache` dropped from `useCallback` deps; re-entrancy guarded via a ref ✅

Bonus improvements I didn't ask for but appreciate:

- `:restore:${id}:${deletedAt}` job-ID suffix in both `POST /fitness-route-heatmap` and `enqueueFitnessRouteHeatmapJobs` so the queue can't dedupe a fresh enqueue against a tombstoned predecessor.
- `bad48a0` fixed the case where Generate disappeared if other route caches existed for the actor — the new `canRefreshRouteCache = !isLoading && !generationPending && !isRouteHeatmapInFlight(heatmapData)` is much clearer.

Remaining items, all minor:

1. **Errors during Clear are hidden behind the dialog.** The page-level error banner sits behind the Radix overlay, so a failed clear shows nothing visible until the user dismisses the modal. Either close the dialog on error or render an inline error inside it. Comment on line 931.
2. **Stale `error` state across dialog opens / cancels.** Calling `setError(null)` when opening the dialog (line 1168) and on Cancel (line 1225) would keep the banner aligned to the current attempt.
3. **Nits in the dialog button**: the `<Trash2>` ternary collapses to a `className={isClearingCache ? 'animate-pulse' : undefined}`, and `aria-busy={isClearingCache}` (or a "Clearing…" label) would let screen-reader users perceive the in-progress state.
4. **Optional test polish**: the new generate-retry test could also assert the error banner clears on the successful second attempt, locking in the `setError(null)` reset path.
5. **`getFitnessRouteHeatmapByKey` is queried twice on enqueue** now (in `enqueueFitnessRouteHeatmapJobs` for the restore suffix, and again in the job itself). Probably fine for current scale; flagging only if `recreateFitnessRouteHeatmapJobs` ever fans out a lot.

None of these block. The PR is in good shape to ship — item 1 is the only one I'd consider before merging.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:940]: Fixed in 75e0808c: clear failures now render an inline alert inside the open dialog, so the user sees the API error without needing to dismiss the modal first.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in 75e0808c: opening, closing, or canceling the Clear route cache dialog now clears stale error state so old failures do not remain visible on later attempts.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:1255]: Fixed in 75e0808c: the destructive action now exposes aria-busy while clearing, switches the label to "Clearing...", and keeps the pulsing icon conditional simple.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx:314]: Fixed in 75e0808c: the manual Generate retry test now asserts the previous enqueue error clears after the successful retry.

lib/jobs/enqueueFitnessRouteHeatmapJobs.ts
llun: No code change: agreed this is worth watching in DB metrics, but it is not a current blocker. The extra lookup is limited to enqueue-time job ID versioning so post-clear QStash dedupe cannot suppress fresh rebuilds; the job-side lookup still needs to remain as the authoritative existence/race check.

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1017,18 +1162,38 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-                      className={cn('size-3', isRetrying && 'animate-spin')}
-                    />
-                    Refresh
-                  </button>
-                )}
+                <div className="flex items-center gap-1.5">
+                  {hasRouteCache && (
+                    <button
+                      onClick={() => {
+                        setError(null)
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: One subtle side effect of using the page-level `error` state for the dialog alert: if an unrelated operation (e.g., `fetchData` after a region change) has already set `error` and the user then opens the Clear dialog, the `setError(null)` here wipes that unrelated error from the page banner — even though the user wasn't trying to dismiss it.

Same on Cancel (line 1238) and on `onOpenChange` (line 1212): each clears whatever error was on screen, regardless of source.

Practical impact is small (the user can re-trigger the failing op), but a dedicated `clearCacheError` state in the dialog would keep the two concerns isolated. Optional — flagging because the coupling isn't immediately obvious from reading the local code.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1040,6 +1205,58 @@ export const FitnessHeatmapView: FC<Props> = ({
           </div>
         </aside>
       </div>
+      <Dialog
+        open={isClearCacheDialogOpen}
+        onOpenChange={(open) => {
+          if (isClearingCache) return
+          setError(null)
+          setIsClearCacheDialogOpen(open)
+        }}
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Both the button's `onClick` (line 1168) and `onOpenChange` (line 1210) call `setError(null)` when transitioning to open. Radix's controlled `<Dialog open>` will fire `onOpenChange(true)` as a result of `setIsClearCacheDialogOpen(true)`, so the second clear is redundant — small nit.

Picking one or the other:

- **Just `onOpenChange`**: single source of truth for "dialog opened or closed" cleanup. Lets you remove the inline `setError(null)` from the button and from the Cancel handler.
- **Just inline onClicks**: more explicit per call site.

Either is fine; the current "both" is just slightly noisy.

---

llun (commented):
Round 5. All round-4 items are landed:

- Errors during Clear now render inside the dialog via `role="alert"` (line 1226-1232) — no more banner-hidden-behind-overlay ✅
- `setError(null)` runs on Clear-button click, on Cancel, and via `onOpenChange` — stale errors cleared on every dialog transition ✅
- `<Trash2>` collapsed to a single element with a conditional className ✅
- `aria-busy={isClearingCache}` + "Clearing..." label on the destructive button ✅
- The generate-retry test now also asserts the error banner clears after the successful second attempt ✅
- The clear-error test was rewritten to use `within(dialog)` and to verify Cancel clears the inline alert ✅

The PR is ready to ship.

Two very minor follow-ups inline; both are optional:

1. **Shared `error` state for Clear vs. page-level operations**: opening or canceling the dialog clears any pre-existing page-level error, even if it was set by an unrelated op. A separate `clearCacheError` would isolate the concerns.
2. **Redundant `setError(null)` calls**: the controlled `onOpenChange` fires when `setIsClearCacheDialogOpen(true)` lands, so the inline `setError(null)` on the Clear button and on Cancel duplicates the cleanup in `onOpenChange`. Pick one site.

Nothing else from prior rounds is outstanding. Approving on logic and tests.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in e72e0148: Clear cache now uses dedicated clearCacheError state, so opening/canceling the dialog no longer clears unrelated page-level errors. Added a regression test for preserving a refresh error through open/cancel.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Fixed in e72e0148: replaced the duplicated setError(null) call sites with a single setClearCacheDialogOpen helper that clears only clearCacheError while opening/closing the dialog.

PR #657, opened 2026-05-11
fix: reject absolute media file paths in public files route

llun:
### Motivation

- Prevent unauthenticated local-file disclosure through public media reads by rejecting absolute route input and enforcing media-root containment in local file storage.

### Description

- Reject absolute POSIX and Windows drive-prefixed paths in `app/api/v1/files/[...pathname]/route.ts` before media lookup.
- Add storage-layer containment in `LocalFileStorage.getFile` so any caller of `getMedia()` cannot read outside the configured media root via absolute or traversal paths.
- Add regression coverage for POSIX absolute paths, Windows drive-prefixed paths, mixed-slash route traversal normalization, and local storage escape attempts.

### Testing

- `yarn lint` (passes with existing no-explicit-any warnings in auth and fitness files)
- `yarn build`
- `yarn test --runInBand --runTestsByPath app/api/v1/files/[...pathname]/route.test.ts lib/services/medias/localFile.test.ts`
- `yarn test`

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a00fcd5285483289163984d55763529)

### DIFF
--- a/app/api/v1/files/[...pathname]/route.ts
+++ b/app/api/v1/files/[...pathname]/route.ts
@@ -15,9 +15,15 @@ export const GET = traceApiRoute(
   'getFile',
   async (req: NextRequest, context: { params: Promise<Params> }) => {
     const { pathname } = await context.params
-    const userPath = path
-      .normalize(Array.isArray(pathname) ? pathname.join('/') : pathname)
-      .replace(/^(\.\.(\/|\\|$))+/, '')
+    const normalizedPath = path.normalize(
+      Array.isArray(pathname) ? pathname.join('/') : pathname
+    )
+
+    if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
+      return apiErrorResponse(404)
+    }
+
+    const userPath = normalizedPath.replace(/^(\.\.(\/|\\|$))+/, '')
     const database = getDatabase()
     if (!database) {
       return apiErrorResponse(500)
--- a/app/api/v1/files/[...pathname]/route.ts
+++ b/app/api/v1/files/[...pathname]/route.ts
@@ -19,6 +19,7 @@ export const GET = traceApiRoute(
       Array.isArray(pathname) ? pathname.join('/') : pathname
     )
 
+    // POSIX hosts do not treat Windows drive paths as absolute.
     if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
       return apiErrorResponse(404)
     }
--- a/app/api/v1/files/[...pathname]/route.test.ts
+++ b/app/api/v1/files/[...pathname]/route.test.ts
@@ -1,0 +1,57 @@
+import { NextRequest } from 'next/server'
+
+import { getMedia } from '@/lib/services/medias'
+
+import { GET } from './route'
+
+const mockDatabase = { id: 'test-database' }
+
+jest.mock('@/lib/database', () => ({
+  getDatabase: jest.fn(() => mockDatabase)
+}))
+
+jest.mock('@/lib/services/medias', () => ({
+  getMedia: jest.fn()
+}))
+
+describe('GET /api/v1/files/[...pathname]', () => {
+  const mockGetMedia = getMedia as jest.MockedFunction<typeof getMedia>
+
+  beforeEach(() => {
+    mockGetMedia.mockReset()
+  })
+
+  const getFile = (pathname: string[]) =>
+    GET(new NextRequest('https://llun.test/api/v1/files/test.png'), {
+      params: Promise.resolve({ pathname })
+    })
+
+  it('rejects absolute POSIX paths before media lookup', async () => {
+    const response = await getFile(['/etc/passwd'])
+
+    expect(response.status).toBe(404)
+    expect(mockGetMedia).not.toHaveBeenCalled()
+  })
+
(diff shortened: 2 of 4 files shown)


app/api/v1/files/[...pathname]/route.ts
```diff
@@ -15,9 +15,15 @@ export const GET = traceApiRoute(
...
-    const userPath = path
-      .normalize(Array.isArray(pathname) ? pathname.join('/') : pathname)
-      .replace(/^(\.\.(\/|\\|$))+/, '')
+    const normalizedPath = path.normalize(
+      Array.isArray(pathname) ? pathname.join('/') : pathname
+    )
+
+    if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
+      return apiErrorResponse(404)
+    }
```
Comment (llun) [inline app/api/v1/files/[...pathname]/route.ts:25]: The `isAbsolute` guard is a useful addition, but it addresses only the narrow case where an attacker directly supplies an absolute path (e.g. `/etc/passwd`). The primary path-traversal protection is still the `replace(/^(\.\.(\/|\\|$))+/, '')` regex on line 26, and the real vulnerability surface is actually one layer deeper.

`LocalFileStorage.getFile` in `lib/services/medias/localFile.ts` does:
```ts
const fullPath = path.resolve(this._config.path, filePath)
```
`path.resolve` will happily escape the storage root if `filePath` contains traversal sequences that survive the regex. The correct, defence-in-depth fix belongs **in the storage layer**, not only in the route handler:

```ts
async getFile(filePath: string) {
  const root = path.resolve(this._config.path)
  const fullPath = path.resolve(root, filePath)
  if (!fullPath.startsWith(root + path.sep)) {
    return null   // path escaped the media root
  }
  // ... rest of method
}
```

Keeping the validation only in the route handler means any future caller of `getMedia()` that skips equivalent sanitisation will reintroduce the vulnerability.

---

app/api/v1/files/[...pathname]/route.ts
```diff
@@ -15,9 +15,15 @@ export const GET = traceApiRoute(
...
   async (req: NextRequest, context: { params: Promise<Params> }) => {
     const { pathname } = await context.params
-    const userPath = path
-      .normalize(Array.isArray(pathname) ? pathname.join('/') : pathname)
-      .replace(/^(\.\.(\/|\\|$))+/, '')
+    const normalizedPath = path.normalize(
+      Array.isArray(pathname) ? pathname.join('/') : pathname
+    )
+
+    if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
```
Comment (llun) [inline app/api/v1/files/[...pathname]/route.ts:23]: The Windows drive-letter regex `/^[a-zA-Z]:/` is a reasonable safety net when the server runs on Linux (where `path.isAbsolute` won't flag `C:\...`), but note that `path.normalize` on Linux does **not** alter Windows-style paths — `path.normalize('C:\\Windows\\System32')` returns `C:\Windows\System32` unchanged. The regex correctly catches this, so the logic is fine, but it is worth a comment explaining *why* the extra regex is needed alongside `isAbsolute`.

---

app/api/v1/files/[...pathname]/route.ts
```diff
@@ -15,9 +15,15 @@ export const GET = traceApiRoute(
...
+      return apiErrorResponse(404)
+    }
+
+    const userPath = normalizedPath.replace(/^(\.\.(\/|\\|$))+/, '')
```
Comment (llun) [inline app/api/v1/files/[...pathname]/route.ts:27]: This `replace` regex is still the primary traversal guard and was already present before this PR. Worth noting that it only strips leading `../` sequences. A path like `foo/../../etc/passwd` after `path.normalize` becomes `../etc/passwd`, and the regex strips the leading `../` to give `etc/passwd` — so it works correctly. However, this subtle chain of reasoning is exactly why the containment check in the storage layer (see comment above) is the more robust fix: it makes correctness self-evident regardless of what preconditions are met upstream.

---

llun (commented):
## Review summary

The intent is correct — preventing unauthenticated local-file disclosure via the public media route — but the fix has structural and test-coverage gaps that need addressing before merging.

### 1. Fix is at the wrong layer (blocking)

The route handler is a reasonable place for *input validation*, but path-containment enforcement must live in **`LocalFileStorage.getFile`** (`lib/services/medias/localFile.ts:54`). Today that method does:

```ts
const fullPath = path.resolve(this._config.path, filePath)
```

…with no check that `fullPath` actually starts with the configured media root. Any future caller of `getMedia()` that omits equivalent sanitisation will silently reintroduce the vulnerability. See the inline comment on line 24 for the recommended containment check.

### 2. Tests were not run (blocking)

`CLAUDE.md` explicitly requires:
> Run `yarn test` before committing

The PR description acknowledges the test suite could not be validated due to a native-module ABI mismatch. This must be resolved before merging — either fix the environment or confirm tests pass in CI.

### 3. No new tests for the added behaviour

There are no tests covering:
- Requests with an absolute POSIX path (e.g. `/etc/passwd` as the pathname segment)
- Requests with a Windows drive-prefixed path (e.g. `C:/Windows/system32/hosts`)
- Requests with mixed-slash traversal sequences

These are exactly the cases this PR is meant to block; they should have regression tests.

### 4. Minor: misleading PR description

The description says "introducing `normalizedPath = path.normalize(...)`" as if normalization is new, but `path.normalize` was already called in the original code. The actual new logic is the `isAbsolute` + Windows-drive guard. Please update the description to avoid confusion for future readers.

---

PR #661, opened 2026-05-12
fix: bind shared inbox actor to HTTP signature key

llun:
### Motivation
- The shared inbox route accepted requests signed with a valid HTTP Signature but did not ensure the signing `keyId` corresponds to `activity.actor`, which permits forged Delete/Undo operations by an unrelated signer.

### Description
- Added an explicit signer→actor binding in `app/api/inbox/route.ts` by importing `parse` and adding `normalizeActorId` to strip fragment identifiers from actor/key URLs.
- The route now reads the `signature` header, parses the `keyId`, and rejects the request with `403` when the normalized `keyId` does not match the normalized `activity.actor`.
- Preserved existing inbox flow, CORS headers, queue publishing, and job dispatch behavior for valid requests.

### Testing
- Ran `yarn run prettier --write .` and it completed successfully.
- Ran `yarn lint` and it completed successfully (existing unrelated warnings remain).
- Ran `yarn build` and it completed successfully.
- Ran `yarn test` which failed in this environment because Node is `v20` while the repo requires Node `24`; observed failure `Array.fromAsync is not a function` in `scripts/productionArchive.test.ts`, so full test suite could not be validated here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a010b9a32a483289b729a9d30c56eb1)

### DIFF
--- a/app/api/inbox/route.ts
+++ b/app/api/inbox/route.ts
@@ -2,7 +2,7 @@ import { StatusActivity } from '@/lib/activities/statusAction'
 import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
 import { ActivityPubVerifySenderGuard } from '@/lib/services/guards/ActivityPubVerifyGuard'
 import { getQueue } from '@/lib/services/queue'
-import { normalizeActorId } from '@/lib/utils/activitypub'
+import { extractActivityPubId, normalizeActorId } from '@/lib/utils/activitypub'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import {
   DEFAULT_202,
@@ -21,17 +21,32 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.POST]
 
 export const OPTIONS = defaultOptions(CORS_HEADERS)
 
+const getActivityBody = async (
+  request: Request,
+  activityBody: unknown | null
+) => {
+  if (activityBody !== null) return activityBody
+
+  try {
+    return (await request.json()) as unknown
+  } catch {
+    return null
+  }
+}
+
 export const POST = traceApiRoute(
   'sharedInbox',
-  ActivityPubVerifySenderGuard(async (request, { database }) => {
-    const body = await request.json()
+  ActivityPubVerifySenderGuard(async (request, { activityBody, database }) => {
+    const body = await getActivityBody(request, activityBody)
+    const actor = isRecord(body) ? extractActivityPubId(body.actor) : undefined
+
     // The guard enforces signed POST actors; keep route validation before casting unknown JSON.
     if (
       !isRecord(body) ||
       typeof body.id !== 'string' ||
       typeof body.type !== 'string' ||
-      typeof body.actor !== 'string' ||
-      !normalizeActorId(body.actor)
+      !actor ||
+      !normalizeActorId(actor)
     ) {
       return apiResponse({
         req: request,
@@ -40,7 +55,7 @@ export const POST = traceApiRoute(
         responseStatusCode: 400
       })
     }
-    const activity = body as unknown as StatusActivity
+    const activity = { ...body, actor } as unknown as StatusActivity
     if (!(await canFederateWithDomain(database, activity.actor))) {
       return apiResponse({
         req: request,
--- a/app/api/inbox/route.ts
+++ b/app/api/inbox/route.ts
@@ -11,7 +11,6 @@ import {
   apiResponse,
   defaultOptions
 } from '@/lib/utils/response'
-import { parse } from '@/lib/utils/signature'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 
 import { getJobMessage } from './getJobMessage'
@@ -23,8 +22,6 @@ export const OPTIONS = defaultOptions(CORS_HEADERS)
 const isRecord = (value: unknown): value is Record<string, unknown> =>
   typeof value === 'object' && value !== null && !Array.isArray(value)
 
-const normalizeActorId = (actorId: string) => actorId.split('#')[0]
-
 export const POST = traceApiRoute(
   'sharedInbox',
   ActivityPubVerifySenderGuard(async (request, { database }) => {
@@ -32,7 +29,8 @@ export const POST = traceApiRoute(
     if (
       !isRecord(body) ||
       typeof body.id !== 'string' ||
-      typeof body.type !== 'string'
+      typeof body.type !== 'string' ||
+      typeof body.actor !== 'string'
     ) {
       return apiResponse({
         req: request,
@@ -42,28 +40,6 @@ export const POST = traceApiRoute(
       })
     }
     const activity = body as unknown as StatusActivity
-    const signatureHeader = request.headers.get('signature')
-    if (!signatureHeader) {
-      return apiResponse({
-        req: request,
-        allowedMethods: CORS_HEADERS,
-        data: ERROR_400,
-        responseStatusCode: 400
-      })
-    }
-    const signatureParts = await parse(signatureHeader)
-    if (
-      !signatureParts.keyId ||
-      normalizeActorId(signatureParts.keyId) !==
-        normalizeActorId(activity.actor)
-    ) {
-      return apiResponse({
-        req: request,
-        allowedMethods: CORS_HEADERS,
-        data: ERROR_403,
-        responseStatusCode: 403
-      })
-    }
     if (!(await canFederateWithDomain(database, activity.actor))) {
       return apiResponse({
         req: request,
--- a/app/api/inbox/route.ts
+++ b/app/api/inbox/route.ts
@@ -11,6 +11,7 @@ import {
   apiResponse,
   defaultOptions
 } from '@/lib/utils/response'
+import { parse } from '@/lib/utils/signature'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 
 import { getJobMessage } from './getJobMessage'
@@ -21,6 +22,8 @@ export const OPTIONS = defaultOptions(CORS_HEADERS)
 
 const isRecord = (value: unknown): value is Record<string, unknown> =>
   typeof value === 'object' && value !== null && !Array.isArray(value)
+
+const normalizeActorId = (actorId: string) => actorId.split('#')[0]
 
 export const POST = traceApiRoute(
   'sharedInbox',
@@ -39,6 +42,28 @@ export const POST = traceApiRoute(
       })
     }
     const activity = body as unknown as StatusActivity
+    const signatureHeader = request.headers.get('signature')
+    if (!signatureHeader) {
+      return apiResponse({
+        req: request,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+    const signatureParts = await parse(signatureHeader)
+    if (
+      !signatureParts.keyId ||
+      normalizeActorId(signatureParts.keyId) !==
+        normalizeActorId(activity.actor)
+    ) {
+      return apiResponse({
+        req: request,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_403,
+        responseStatusCode: 403
+      })
+    }
     if (!(await canFederateWithDomain(database, activity.actor))) {
       return apiResponse({
         req: request,
--- a/app/api/inbox/route.ts
+++ b/app/api/inbox/route.ts
@@ -2,6 +2,7 @@ import { StatusActivity } from '@/lib/activities/statusAction'
 import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
 import { ActivityPubVerifySenderGuard } from '@/lib/services/guards/ActivityPubVerifyGuard'
 import { getQueue } from '@/lib/services/queue'
+import { normalizeActorId } from '@/lib/utils/activitypub'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import {
   DEFAULT_202,
@@ -12,6 +13,7 @@ import {
   defaultOptions
 } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { isRecord } from '@/lib/utils/typeGuards'
 
 import { getJobMessage } from './getJobMessage'
 
@@ -19,18 +21,17 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.POST]
 
 export const OPTIONS = defaultOptions(CORS_HEADERS)
 
-const isRecord = (value: unknown): value is Record<string, unknown> =>
-  typeof value === 'object' && value !== null && !Array.isArray(value)
-
 export const POST = traceApiRoute(
   'sharedInbox',
   ActivityPubVerifySenderGuard(async (request, { database }) => {
     const body = await request.json()
+    // The guard enforces signed POST actors; keep route validation before casting unknown JSON.
     if (
       !isRecord(body) ||
       typeof body.id !== 'string' ||
       typeof body.type !== 'string' ||
-      typeof body.actor !== 'string'
+      typeof body.actor !== 'string' ||
+      !normalizeActorId(body.actor)
     ) {
       return apiResponse({
         req: request,
--- a/app/api/users/[username]/inbox/route.ts
+++ b/app/api/users/[username]/inbox/route.ts
@@ -104,7 +104,9 @@ export const POST = traceApiRoute(
               })
             }
 
-            const parsed = Activity.safeParse(await req.json())
+            const parsed = Activity.safeParse(
+              context.activityBody ?? (await req.json())
+            )
             if (!parsed.success) {
               return apiResponse({
                 req,
--- a/lib/services/guards/ActivityPubVerifyGuard.ts
+++ b/lib/services/guards/ActivityPubVerifyGuard.ts
@@ -143,6 +143,11 @@ export const ActivityPubVerifySenderGuard =
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
+    const activityActor = await getPostActivityActor(request)
+    if (!activityActor.valid) {
+      return guardErrorResponse(request, 400, allowedMethods)
+    }
+
     if (!(await canFederateWithDomain(database, signatureParts.keyId))) {
       return guardErrorResponse(request, 403, allowedMethods)
     }
@@ -157,11 +162,6 @@ export const ActivityPubVerifySenderGuard =
     if (
       !(await verify(requestTarget, request.headers, senderPublicKey.publicKey))
     ) {
-      return guardErrorResponse(request, 400, allowedMethods)
-    }
-
-    const activityActor = await getPostActivityActor(request)
-    if (!activityActor.valid) {
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
--- a/lib/services/guards/ActivityPubVerifyGuard.ts
+++ b/lib/services/guards/ActivityPubVerifyGuard.ts
@@ -4,7 +4,7 @@ import crypto from 'node:crypto'
 import { getDatabase } from '@/lib/database'
 import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
 import { getHeadersValue } from '@/lib/services/guards/getHeaderValue'
-import { normalizeActorId } from '@/lib/utils/activitypub'
+import { extractActivityPubId, normalizeActorId } from '@/lib/utils/activitypub'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import {
   StatusCode,
@@ -55,24 +55,33 @@ const getExpectedSha256Digest = (digestHeader: string) =>
     })
     .find((part) => part?.algorithm === 'sha-256')?.value
 
-const getPostActivityActor = async (request: NextRequest) => {
-  if (request.method.toUpperCase() !== 'POST') {
-    return { actor: null, valid: true }
+const getPostActivity = ({
+  bodyText,
+  method
+}: {
+  bodyText: string | null
+  method: string
+}) => {
+  if (method.toUpperCase() !== 'POST') {
+    return { actor: null, body: null, valid: true }
   }
 
   try {
-    const body = (await request.clone().json()) as unknown
-    if (
-      !isRecord(body) ||
-      typeof body.actor !== 'string' ||
-      !normalizeActorId(body.actor)
-    ) {
-      return { actor: null, valid: false }
+    if (bodyText === null) return { actor: null, body: null, valid: false }
+
+    const body = JSON.parse(bodyText) as unknown
+    if (!isRecord(body)) {
+      return { actor: null, body: null, valid: false }
     }
 
-    return { actor: body.actor, valid: true }
+    const actor = extractActivityPubId(body.actor)
+    if (!actor || !normalizeActorId(actor)) {
+      return { actor: null, body: null, valid: false }
+    }
+
+    return { actor, body: { ...body, actor }, valid: true }
   } catch {
-    return { actor: null, valid: false }
+    return { actor: null, body: null, valid: false }
   }
 }
 
@@ -95,12 +104,15 @@ const isDateHeaderFresh = (
 const digestMatches = async (request: NextRequest, signedHeaders: string[]) => {
   const digestHeader = getHeadersValue(request.headers, 'digest')
   if (!digestHeader)
-    return ['GET', 'HEAD'].includes(request.method.toUpperCase())
-  if (Array.isArray(digestHeader)) return false
-  if (!signedHeaders.includes('digest')) return false
+    return {
+      bodyText: null,
+      valid: ['GET', 'HEAD'].includes(request.method.toUpperCase())
+    }
+  if (Array.isArray(digestHeader)) return { bodyText: null, valid: false }
+  if (!signedHeaders.includes('digest')) return { bodyText: null, valid: false }
 
   const expectedDigest = getExpectedSha256Digest(digestHeader)
-  if (!expectedDigest) return false
+  if (!expectedDigest) return { bodyText: null, valid: false }
 
   const bodyBuffer = Buffer.from(await request.clone().arrayBuffer())
   const actualDigest = crypto
@@ -111,9 +123,14 @@ const digestMatches = async (request: NextRequest, signedHeaders: string[]) => {
   const actualDigestBuffer = Buffer.from(actualDigest, 'base64')
   const expectedDigestBuffer = Buffer.from(expectedDigest, 'base64')
 
-  if (actualDigestBuffer.length !== expectedDigestBuffer.length) return false
+  if (actualDigestBuffer.length !== expectedDigestBuffer.length) {
+    return { bodyText: null, valid: false }
+  }
 
-  return crypto.timingSafeEqual(actualDigestBuffer, expectedDigestBuffer)
+  return {
+    bodyText: bodyBuffer.toString('utf8'),
+    valid: crypto.timingSafeEqual(actualDigestBuffer, expectedDigestBuffer)
+  }
 }
 
 export const ActivityPubVerifySenderGuard =
@@ -139,12 +156,16 @@ export const ActivityPubVerifySenderGuard =
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
-    if (!(await digestMatches(request, signedHeaders))) {
+    const digestResult = await digestMatches(request, signedHeaders)
+    if (!digestResult.valid) {
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
-    const activityActor = await getPostActivityActor(request)
-    if (!activityActor.valid) {
+    const activity = getPostActivity({
+      bodyText: digestResult.bodyText,
+      method: request.method
+    })
+    if (!activity.valid) {
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
@@ -165,14 +186,18 @@ export const ActivityPubVerifySenderGuard =
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
-    if (activityActor.actor) {
+    if (activity.actor) {
       const normalizedOwner = normalizeActorId(senderPublicKey.owner)
-      const normalizedActor = normalizeActorId(activityActor.actor)
+      const normalizedActor = normalizeActorId(activity.actor)
 
       if (!normalizedOwner || normalizedOwner !== normalizedActor) {
         return guardErrorResponse(request, 403, allowedMethods)
       }
     }
 
-    return handle(request, { database, params: context.params })
+    return handle(request, {
+      activityBody: activity.body,
+      database,
+      params: context.params
+    })
   }
--- a/lib/services/guards/ActivityPubVerifyGuard.ts
+++ b/lib/services/guards/ActivityPubVerifyGuard.ts
@@ -13,7 +13,7 @@ import {
 } from '@/lib/utils/response'
 import { parse, verify } from '@/lib/utils/signature'
 
-import { getSenderPublicKey } from './getSenderPublicKey'
+import { getSenderPublicKeyDetails } from './getSenderPublicKey'
 import { headerHost } from './headerHost'
 import { ActivityPubVerifiedSenderHandle, AppRouterParams } from './types'
 
@@ -52,6 +52,26 @@ const getExpectedSha256Digest = (digestHeader: string) =>
       }
     })
     .find((part) => part?.algorithm === 'sha-256')?.value
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null && !Array.isArray(value)
+
+const getPostActivityActor = async (request: NextRequest) => {
+  if (request.method.toUpperCase() !== 'POST') {
+    return { actor: null, valid: true }
+  }
+
+  try {
+    const body = (await request.clone().json()) as unknown
+    if (!isRecord(body) || typeof body.actor !== 'string') {
+      return { actor: null, valid: false }
+    }
+
+    return { actor: body.actor, valid: true }
+  } catch {
+    return { actor: null, valid: false }
+  }
+}
 
 const isDateHeaderFresh = (
   headers: Headers,
@@ -127,9 +147,23 @@ export const ActivityPubVerifySenderGuard =
     const host = headerHost(request.headers)
     const requestUrl = new URL(request.url, `http://${host}`)
     const requestTarget = `${request.method.toLowerCase()} ${requestUrl.pathname}${requestUrl.search}`
-    const publicKey = await getSenderPublicKey(database, signatureParts.keyId)
-    if (!(await verify(requestTarget, request.headers, publicKey))) {
+    const senderPublicKey = await getSenderPublicKeyDetails(
+      database,
+      signatureParts.keyId
+    )
+    if (
+      !(await verify(requestTarget, request.headers, senderPublicKey.publicKey))
+    ) {
       return guardErrorResponse(request, 400, allowedMethods)
+    }
+
+    const activityActor = await getPostActivityActor(request)
+    if (!activityActor.valid) {
+      return guardErrorResponse(request, 400, allowedMethods)
+    }
+
+    if (activityActor.actor && senderPublicKey.owner !== activityActor.actor) {
+      return guardErrorResponse(request, 403, allowedMethods)
     }
 
     return handle(request, { database, params: context.params })
--- a/lib/services/guards/ActivityPubVerifyGuard.ts
+++ b/lib/services/guards/ActivityPubVerifyGuard.ts
@@ -4,6 +4,7 @@ import crypto from 'node:crypto'
 import { getDatabase } from '@/lib/database'
 import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
 import { getHeadersValue } from '@/lib/services/guards/getHeaderValue'
+import { normalizeActorId } from '@/lib/utils/activitypub'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import {
   StatusCode,
@@ -12,6 +13,7 @@ import {
   codeMap
 } from '@/lib/utils/response'
 import { parse, verify } from '@/lib/utils/signature'
+import { isRecord } from '@/lib/utils/typeGuards'
 
 import { getSenderPublicKeyDetails } from './getSenderPublicKey'
 import { headerHost } from './headerHost'
@@ -53,9 +55,6 @@ const getExpectedSha256Digest = (digestHeader: string) =>
     })
     .find((part) => part?.algorithm === 'sha-256')?.value
 
-const isRecord = (value: unknown): value is Record<string, unknown> =>
-  typeof value === 'object' && value !== null && !Array.isArray(value)
-
 const getPostActivityActor = async (request: NextRequest) => {
   if (request.method.toUpperCase() !== 'POST') {
     return { actor: null, valid: true }
@@ -63,7 +62,11 @@ const getPostActivityActor = async (request: NextRequest) => {
 
   try {
     const body = (await request.clone().json()) as unknown
-    if (!isRecord(body) || typeof body.actor !== 'string') {
+    if (
+      !isRecord(body) ||
+      typeof body.actor !== 'string' ||
+      !normalizeActorId(body.actor)
+    ) {
       return { actor: null, valid: false }
     }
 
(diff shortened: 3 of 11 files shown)


llun (commented):
The intention here is correct — enforcing that the HTTP Signature `keyId` matches `activity.actor` is a real gap. But the implementation has several issues that need to be addressed before merging.

---

**Issue 1 — Redundant signature-header null check (dead code)**

`ActivityPubVerifySenderGuard` already rejects requests with no `signature` header (returns 400) before the handler ever runs (`ActivityPubVerifyGuard.ts` lines 105–107). The new `if (!signatureHeader)` block added in this PR is unreachable dead code and should be removed.

---

**Issue 2 — Redundant `parse()` call**

`ActivityPubVerifySenderGuard` already calls `parse(requestSignature)` internally and validates `keyId` exists (`ActivityPubVerifyGuard.ts` lines 109–112). Calling `parse(signatureHeader)` again inside the handler parses the same header a second time on every request. The clean fix is to extend the guard's context object to forward the already-validated `keyId` to the handler:

```ts
// ActivityPubVerifyGuard.ts — pass keyId through to handler
return handle(request, { database, keyId: signatureParts.keyId, params: context.params })
```

Then consume it from context in the handler instead of re-parsing. This also keeps the actor-binding concern inside the guard where it belongs.

---

**Issue 3 — Missing type guard on `activity.actor` (potential runtime crash)**

The body validation block checks `body.id` and `body.type` are strings but never checks `body.actor`. ActivityPub activities don't always carry an `actor` field, and when they do the spec allows it to be an object or array. Calling `normalizeActorId(activity.actor)` with `undefined` or an object will throw:

```
TypeError: actorId.split is not a function
```

The fix is to extend the existing body validation:

```ts
if (
  !isRecord(body) ||
  typeof body.id !== 'string' ||
  typeof body.type !== 'string' ||
  typeof body.actor !== 'string'   // add this
) {
  return apiResponse({ ..., data: ERROR_400, responseStatusCode: 400 })
}
```

---

**Issue 4 — `normalizeActorId` breaks federation with servers that use path-based key IDs**

`normalizeActorId` only strips fragment identifiers (`#...`). This works for the common Mastodon pattern (`https://example.com/users/alice#main-key` → `https://example.com/users/alice`), but other ActivityPub implementations use path-based key documents:

- `https://example.com/users/alice/keys/main`
- `https://example.com/actors/alice/publickey`

After fragment-stripping, these remain unchanged and will never equal `https://example.com/users/alice`. Those servers will receive 403 for every request — a federation-breaking regression for any non-Mastodon peer.

The correct approach is to use the actor URL from the already-fetched public key document (which `getSenderPublicKey` retrieves inside the guard), rather than string-manipulating the `keyId` URL. The guard has everything it needs to do this check authoritatively; it should perform the actor-binding validation itself and either pass the resolved actor URL to the handler or reject early.

---

llun (commented):
## Review: fix: bind shared inbox actor to HTTP signature key

The intent is correct — verifying that the HTTP signature's `keyId` matches `activity.actor` is a meaningful security check. However, the implementation has several issues that need to be addressed.

---

### 1. Unreachable code — the guard already rejects missing signatures and missing keyId

`ActivityPubVerifySenderGuard` (in `ActivityPubVerifyGuard.ts`) already returns 400 when the `signature` header is absent and when `keyId` is missing from the parsed signature. The new blocks inside the handler that re-check for a missing signature header and a missing `keyId` can never be reached. This is dead code.

```ts
// ActivityPubVerifyGuard.ts – already handles both cases
const requestSignature = request.headers.get('signature')
if (!requestSignature)
  return guardErrorResponse(request, 400, allowedMethods)

const signatureParts = await parse(requestSignature)
if (!signatureParts.keyId) {
  return guardErrorResponse(request, 400, allowedMethods)
}
```

---

### 2. `parse` is called twice for every request

The guard already calls `parse(requestSignature)` and extracts `keyId`. The handler now calls `parse` a second time on the same header string. The `keyId` should be extracted once inside the guard and forwarded to the handler (e.g. via the context object or a wrapper type), rather than re-parsed.

---

### 3. The check belongs in the guard, not the handler

The keyId–actor binding is part of request authentication, which is the guard's responsibility. Adding it in the handler leaks authentication logic into business code. The user inbox route (`/api/users/[username]/inbox/route.ts`) uses the same guard and has the same vulnerability — placing the fix in the guard would fix both routes at once.

---

### 4. May break federation with servers using non-standard key URLs

The comparison `normalizeActorId(keyId) !== normalizeActorId(activity.actor)` assumes that stripping the fragment from the key URL yields exactly the actor URL. This holds for Mastodon-style `https://example.com/users/alice#main-key` → `https://example.com/users/alice`, but some ActivityPub implementations use a completely separate key endpoint (e.g. `https://example.com/keys/alice`). Those requests will be rejected with 403 even though the signature is cryptographically valid, breaking federation with those servers.

---

### 5. `activity.actor` is not validated as a string before calling `normalizeActorId`

After the `isRecord` + `id`/`type` checks, `body` is cast to `StatusActivity`. But `actor` is not validated to be a string at that point. If a malformed payload sends `actor: null` or `actor: 42`, calling `.split('#')[0]` on it will throw a runtime error, producing an unhandled 500 instead of a 400.

Add a type guard for `actor` alongside the existing `id` and `type` checks:

```ts
if (
  !isRecord(body) ||
  typeof body.id !== 'string' ||
  typeof body.type !== 'string' ||
  typeof body.actor !== 'string'
) { ... }
```

---

llun (commented):
Three issues to fix before merging:

**1. Dead code — unreachable signature-missing check (`app/api/inbox/route.ts`, lines 45–53)**

`ActivityPubVerifySenderGuard` already rejects any request without a `signature` header (returns 400 at `lib/services/guards/ActivityPubVerifyGuard.ts:106`). The new block that re-reads `request.headers.get('signature')` and returns a 400 can never execute; remove it.

**2. Redundant `parse()` call (`app/api/inbox/route.ts`, line 54)**

The guard already calls `parse(requestSignature)` at `ActivityPubVerifyGuard.ts:109` and validates that `keyId` is present before the handler is invoked. Calling `parse()` a second time in the route duplicates work on every request. A better fix is to thread `keyId` through the guard context: extend `ActivityPubVerifiedSenderHandle`'s context type to include `keyId: string`, set it in the guard after the existing parse call, and read it directly in the handler without re-parsing.

**3. Overly strict `normalizeActorId` normalization (`app/api/inbox/route.ts`, line 26)**

`normalizeActorId` only strips a URL fragment (`#main-key`). This works for the common Mastodon convention (`https://domain/users/alice#main-key`) but will produce false positives for implementations that use a path-based key URL without a fragment — e.g., `https://domain/users/alice/publickey` or `https://domain/users/alice/key`. After stripping the fragment those strings still don't match `activity.actor`, so legitimate federation from those servers will be rejected with 403. The correct approach is to resolve the key owner: `getSenderPublicKey` (called by the guard) already fetches the actor document; expose the resolved actor URL through the guard context and compare that instead of doing a raw string comparison.

---

app/api/inbox/route.ts
```diff
@@ -39,6 +42,28 @@ export const POST = traceApiRoute(
...
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+    const signatureParts = await parse(signatureHeader)
+    if (
+      !signatureParts.keyId ||
+      normalizeActorId(signatureParts.keyId) !==
+        normalizeActorId(activity.actor)
+    ) {
```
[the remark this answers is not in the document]
llun: Fixed in 8681237b. The guard now rejects signed POST bodies without a string `actor` before the handler runs, and the shared inbox route also validates `body.actor` before casting so malformed actors return 400 instead of throwing.

app/api/inbox/route.ts
```diff
@@ -39,6 +42,28 @@ export const POST = traceApiRoute(
...
+    if (!signatureHeader) {
+      return apiResponse({
+        req: request,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+    const signatureParts = await parse(signatureHeader)
+    if (
```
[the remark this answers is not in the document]
llun: Fixed in 8681237b. The shared inbox no longer reads or parses the `signature` header; signer-to-actor binding now lives in `ActivityPubVerifySenderGuard` using the already parsed `keyId` and resolved key owner.

app/api/inbox/route.ts
```diff
@@ -39,6 +42,28 @@ export const POST = traceApiRoute(
...
+      normalizeActorId(signatureParts.keyId) !==
+        normalizeActorId(activity.actor)
+    ) {
+      return apiResponse({
+        req: request,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_403,
+        responseStatusCode: 403
+      })
+    }
```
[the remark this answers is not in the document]
llun: Fixed in 8681237b. Added guard tests for missing/non-string actor, owner mismatch 403, and path-based key id accepted when owned by the activity actor, plus key helper tests for actor/key documents. The missing-signature 400 case was already covered by the existing guard CORS error test, so I kept that coverage rather than duplicating it in the route suite.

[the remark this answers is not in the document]
llun: Fixed in 8681237b. I removed the redundant shared-inbox signature/keyId checks rather than applying the non-null parse suggestion, because the guard already parses the signature and now owns the binding check. That also covers `/api/users/[username]/inbox`.

app/api/inbox/route.ts
```diff
@@ -39,6 +42,28 @@ export const POST = traceApiRoute(
...
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+    const signatureParts = await parse(signatureHeader)
+    if (
+      !signatureParts.keyId ||
+      normalizeActorId(signatureParts.keyId) !==
+        normalizeActorId(activity.actor)
```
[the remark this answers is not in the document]
llun: Fixed in 8681237b and hardened in 1ded7f67. Key lookup now resolves `publicKey.owner` from actor/key documents, so path-based key ids are accepted when owned by `activity.actor`; the follow-up commit rejects fetched key documents whose `id` does not match the requested `keyId`.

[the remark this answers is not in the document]
llun: Fixed in 8681237b and hardened in 1ded7f67. The raw `keyId` string equality check is gone; the guard resolves the key owner/controller from the fetched actor or key document and compares that owner to `activity.actor`, while the follow-up hardening rejects a mismatched returned key document id.

llun:
@codex review

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,41 +1,157 @@
...
+}
+
+const parseSenderPublicKeyDetails = ({
+  body,
+  keyId
+}: {
+  body: string
+  keyId: string
+}): SenderPublicKeyDetails | null => {
+  const json = JSON.parse(body)
```
llun: `JSON.parse(body)` throws synchronously on malformed JSON, but `parseSenderPublicKeyDetails` has no try/catch around it. The call site `fetchSenderPublicKeyDetails` also doesn't catch it, so the exception bubbles up to the top-level catch in `getSenderPublicKeyDetails` and silently becomes `EMPTY_PUBLIC_KEY_DETAILS`. This is safe from a security standpoint, but it makes it very hard to diagnose malformed responses from remote servers. Wrap this in a try/catch and return `null` explicitly, or move the `JSON.parse` into `fetchSenderPublicKeyDetails` where the HTTP error handling already lives.

---

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,41 +1,157 @@
...
+    if (response.statusCode === 410) {
+      const url = new URL(actorId)
+      const fallbackResponse = await fetchSenderPublicKeyDetails(
+        `${url.protocol}//${url.host}/actor#main-key`,
+        signingActor
+      )
+      return fallbackResponse.details ?? EMPTY_PUBLIC_KEY_DETAILS
+    }
+  } catch {
+    return EMPTY_PUBLIC_KEY_DETAILS
```
llun: This bare `catch` swallows **all** exceptions — network timeouts, DNS failures, programming errors — and converts them into the same silent `EMPTY_PUBLIC_KEY_DETAILS` return. The old `getSenderPublicKey` preserved the error and recorded it on the trace span. Now those errors are completely invisible.

At a minimum, log or rethrow non-transient errors (e.g., those that are not `HTTPError` / network errors), or record the caught error somewhere so diagnostics remain possible. Compare with the behaviour that was removed from the old `getSenderPublicKey`.

---

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,41 +1,157 @@
...
-        return localActor.publicKey
-      }
-
-      const signingActor = await getFederationSigningActor(database)
       try {
-        const sender = await getActorPerson({ actorId, signingActor })
-        return sender?.publicKey.publicKeyPem ?? ''
+        const sender = await getSenderPublicKeyDetails(database, actorId)
+        return sender.publicKey
       } catch (error) {
```
llun, line 308: This `catch` block is dead code. `getSenderPublicKeyDetails` catches all exceptions internally and never throws, so `span.recordException(nodeError)` can never execute. The tracing that existed in the old implementation for key-lookup failures is now completely lost. Either:
1. Have `getSenderPublicKeyDetails` propagate errors so this span can catch and record them, or
2. Move the `tracer.startActiveSpan` call inside `getSenderPublicKeyDetails` so key-lookup errors are still observable.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +147,24 @@ export const ActivityPubVerifySenderGuard =
...
+    if (!activityActor.valid) {
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
+    if (activityActor.actor && senderPublicKey.owner !== activityActor.actor) {
+      return guardErrorResponse(request, 403, allowedMethods)
+    }
+
     return handle(request, { database, params: context.params })
   }
```
llun, line 203: The owner/actor comparison is a raw string equality check with no URL normalization. Two issues:

1. **Fragment bypass**: if `body.actor` contains a fragment (e.g. `https://remote.test/users/alice#inbox`), the comparison fails against the normalized `senderPublicKey.owner` (`https://remote.test/users/alice`), producing a false 403 for a legitimate request. The `removeFragment` helper already exists in `getSenderPublicKey.ts` — apply it to both sides before comparing.

2. **Mismatch with PR description**: the description says `normalizeActorId` is used to strip fragments, but no such normalization is present in the diff. The `removeFragment` helper defined in `getSenderPublicKey.ts` is only used for local actor lookup, not here.

Suggested fix:
```ts
const normalizedOwner = senderPublicKey.owner?.split('#')[0] ?? null
const normalizedActor = activityActor.actor?.split('#')[0] ?? null
if (normalizedActor && normalizedOwner !== normalizedActor) {
  return guardErrorResponse(request, 403, allowedMethods)
}
```

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -53,6 +53,26 @@ const getExpectedSha256Digest = (digestHeader: string) =>
     })
     .find((part) => part?.algorithm === 'sha-256')?.value
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
llun: `isRecord` is already defined identically in `app/api/inbox/route.ts`. Extract it to a shared utility (e.g. `lib/utils/typeGuards.ts`) and import it in both files.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +147,24 @@ export const ActivityPubVerifySenderGuard =
     const host = headerHost(request.headers)
     const requestUrl = new URL(request.url, `http://${host}`)
     const requestTarget = `${request.method.toLowerCase()} ${requestUrl.pathname}${requestUrl.search}`
```
llun, line 178: The guard now calls `getSenderPublicKeyDetails` directly, bypassing the `tracer.startActiveSpan('guard.getSenderPublicKey', ...)` wrapper. Key-lookup latency and errors are no longer traced. Use `getSenderPublicKey` (which still has the span) and retrieve the owner separately, or move the tracing into `getSenderPublicKeyDetails` itself.

---

app/api/inbox/route.ts
```diff
@@ -29,7 +29,8 @@ export const POST = traceApiRoute(
     if (
       !isRecord(body) ||
       typeof body.id !== 'string' ||
-      typeof body.type !== 'string'
+      typeof body.type !== 'string' ||
```
llun, line 47: The `typeof body.actor !== 'string'` check added here is now redundant. `ActivityPubVerifyGuard` (which wraps this handler) already rejects POST requests where `body.actor` is not a string before the handler is ever called. The check is harmless, but it creates a false impression that this handler might receive a non-string actor, and it will drift out of sync with the guard over time. Consider removing it and adding a comment that the guard enforces this invariant, or leave both and document the intentional defense-in-depth.

---

llun (commented):
The security intent (binding the HTTP-signature key owner to the activity actor) is correct and the test coverage is solid. However there are several issues that need to be fixed before merging:

**Must fix**
1. **Missing URL normalization in the owner/actor comparison** — raw string equality on `senderPublicKey.owner !== activityActor.actor` can false-403 legitimate requests if either side carries a `#fragment`. The `removeFragment` helper already exists in this file; use it on both values.
2. **Dead `catch` block / lost tracing in `getSenderPublicKey`** — because `getSenderPublicKeyDetails` never throws, `span.recordException` is unreachable and key-lookup errors are now invisible in traces.
3. **Bare `catch` swallows all errors in `getSenderPublicKeyDetails`** — network timeouts and programming bugs are silently converted to `EMPTY_PUBLIC_KEY_DETAILS`. At minimum, log or re-throw non-transient errors.

**Should fix**
4. **`JSON.parse` without try/catch in `parseSenderPublicKeyDetails`** — malformed remote responses throw synchronously and are silently absorbed by the outer catch, making them very hard to diagnose.
5. **Tracing regression** — the guard calls `getSenderPublicKeyDetails` directly, losing the `guard.getSenderPublicKey` span. Move tracing into `getSenderPublicKeyDetails` or use the existing `getSenderPublicKey` wrapper.

**Minor**
6. **Duplicate `isRecord`** — now defined in both `ActivityPubVerifyGuard.ts` and `app/api/inbox/route.ts`; extract to a shared utility.
7. **Redundant actor check in `route.ts`** — the guard already enforces this; clarify whether it's intentional defense-in-depth or an oversight.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +147,24 @@ export const ActivityPubVerifySenderGuard =
...
+    ) {
+      return guardErrorResponse(request, 400, allowedMethods)
+    }
+
+    const activityActor = await getPostActivityActor(request)
+    if (!activityActor.valid) {
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
+    if (activityActor.actor && senderPublicKey.owner !== activityActor.actor) {
```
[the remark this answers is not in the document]
llun: Addressed in 19b8eec9. The guard now rejects POST bodies without a non-empty actor identity before owner comparison, then compares `normalizeActorId(senderPublicKey.owner)` to `normalizeActorId(activityActor.actor)`. If the owner is missing or the normalized values differ, it fails closed with 403. Added coverage for fragment-only owner/actor differences and empty actor rejection.

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,41 +1,157 @@
...
+    }
+  }
+
+  const publicKeyDocument = PublicKeyDocument.safeParse(json)
+  if (!publicKeyDocument.success) return null
+  if (publicKeyDocument.data.id !== keyId) return null
+
+  return {
+    owner: publicKeyDocument.data.owner,
+    publicKey: publicKeyDocument.data.publicKeyPem
```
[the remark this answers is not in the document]
llun: Addressed in 19b8eec9. Raw key documents are no longer trusted solely because they claim an `owner`; the lookup now fetches the owner actor and accepts the key only when that actor publishes the same key id and public key. Added negative coverage for mismatched key id, mismatched public key, wrong actor id, and delegated owner.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 19b8eec9. JSON parsing now happens through `parseJsonBody`; malformed remote key responses return `null` explicitly and emit `logger.warn` with the actor id and error, without logging the response body.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 19b8eec9. Lookup failures still fail closed to empty key details, but they are no longer silent: `getSenderPublicKeyDetails` records exceptions on the span and logs a warning with the actor id and error.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 19b8eec9. The `guard.getSenderPublicKey` span now lives inside `getSenderPublicKeyDetails`, which is the path the guard calls. Lookup exceptions are recorded there before returning empty details, so the old wrapper catch is no longer needed.

lib/services/guards/ActivityPubVerifyGuard.ts
llun: Addressed in 19b8eec9. Owner and actor comparison now uses shared `normalizeActorId` on both sides, with tests covering fragment-only differences. Missing owner or mismatch returns 403 for POST activities.

lib/services/guards/ActivityPubVerifyGuard.ts
llun: Addressed in 19b8eec9. `isRecord` now lives in `lib/utils/typeGuards.ts` and is imported by both the guard and the shared inbox route.

lib/services/guards/ActivityPubVerifyGuard.ts
llun: Addressed in 19b8eec9. The guard still calls `getSenderPublicKeyDetails` directly, but that function now owns the `guard.getSenderPublicKey` tracing span, so key lookup latency and errors are traced on the direct path.

app/api/inbox/route.ts
llun: Left this intentionally as defense-in-depth. The guard enforces the signed POST actor invariant before the handler, and the route now has a comment documenting that while still validating a non-empty normalized actor before casting unknown JSON to `StatusActivity`.

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,41 +1,157 @@
...
+  const json = JSON.parse(body)
+  const actor = Actor.safeParse(json)
+  if (actor.success) {
+    if (actor.data.id !== keyId && actor.data.publicKey.id !== keyId) {
+      return null
+    }
+
+    return {
+      owner: actor.data.publicKey.owner,
+      publicKey: actor.data.publicKey.publicKeyPem
```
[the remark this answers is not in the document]
llun: Addressed in 19b8eec9. Actor documents fetched from key URLs now require the same owner confirmation as raw key documents: the claimed owner actor must publish the same key id and public key before we trust the owner. Added an attacker/victim regression test for this path.

llun:
@codex review

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,45 +1,285 @@
...
+}
+
+const resolveFetchedPublicKey = async (
+  document: ParsedSenderPublicKey | null,
+  signingActor: Awaited<ReturnType<typeof getFederationSigningActor>>
+) => {
+  if (!document) return null
+  if (document.type === 'actor') {
+    if (!document.requiresOwnerValidation) return document.details
+    return validateOwnerActorKey(
```
llun, line 231: **Performance regression for the typical Mastodon-style federation case.**

For the standard ActivityPub pattern where `keyId = '<actor>#main-key'`:
1. We fetch `keyId` → HTTP strips the fragment → request hits `<actor>`, returns the canonical actor document with `id: '<actor>'`.
2. `parseSenderPublicKey` sets `requiresOwnerValidation: true` (because `actor.data.id !== keyId`).
3. `validateOwnerActorKey` then fetches `owner = document.actorId = '<actor>'` — **the exact same URL we just hit**, producing a redundant HTTP round-trip on every remote inbox POST.

The owner re-fetch is genuinely needed when the keyId URL is *different* from the actor URL (e.g., path-based `…/keys/main`, or a bare PublicKeyDocument), but it adds no security when the keyId-without-fragment already equals `actor.data.id` — the document we have is already the canonical self-attestation.

Suggested short-circuit:
```ts
if (document.type === 'actor') {
  if (!document.requiresOwnerValidation) return document.details
  if (normalizeActorId(document.keyId) === document.actorId) {
    // keyId only differs from actorId by fragment, so we already fetched
    // the canonical actor URL — no need to re-fetch.
    return document.details
  }
  return validateOwnerActorKey(...)
}
```

This halves the outgoing HTTP requests per remote inbox delivery for the most common federation pattern. The test `'returns public key owner details from remote actor'` (which currently makes only one mocked fetch) implicitly assumes this optimization — but `'accepts actor documents from key URLs only when the owner actor publishes the same key'` and the Mastodon-style path now make two.

---

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,45 +1,285 @@
...
+  signingActor: Awaited<ReturnType<typeof getFederationSigningActor>>
+) => {
+  if (!owner) return null
+
+  const ownerResponse = await fetchSenderPublicKey(owner, signingActor)
+  if (ownerResponse.statusCode !== 200) return null
+  if (ownerResponse.document?.type !== 'actor') return null
+
+  const ownerDocument = ownerResponse.document
+  if (ownerDocument.actorId !== owner) return null
```
llun: `owner` here can come from a bare `PublicKeyDocument`'s `owner` field, which the Zod schema only validates as `z.string()` — it could carry a `#fragment`. The HTTP request strips the fragment before hitting the server, but then `ownerDocument.actorId !== owner` will compare the canonical actor id (no fragment) against the fragment-bearing `owner` and reject. Consider normalizing first:

```ts
const normalizedOwner = normalizeActorId(owner)
if (!normalizedOwner) return null
const ownerResponse = await fetchSenderPublicKey(normalizedOwner, signingActor)
...
if (ownerDocument.actorId !== normalizedOwner) return null
```

Minor edge case — most servers won't put fragments in `publicKey.owner` — but currently a legitimate document with such a fragment would fail this check.

---

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,45 +1,285 @@
...
+}): ParsedSenderPublicKey | null => {
+  const json = parseJsonBody({ body, actorId: keyId })
+  if (!json) return null
+
+  const actor = Actor.safeParse(json)
+  if (actor.success) {
+    if (actor.data.id !== keyId && actor.data.publicKey.id !== keyId) {
+      return null
+    }
+    if (actor.data.publicKey.owner !== actor.data.id) {
```
llun: Strict equality on `actor.data.publicKey.owner !== actor.data.id` is good security hardening, but neither field is normalized. If a server (legitimately) publishes an actor with `id: 'https://remote.test/users/alice'` and `publicKey.owner: 'https://remote.test/users/alice#main-key'` (unusual but spec-permitted), this rejects them. Consider `normalizeActorId(...)` on both sides if interop reports come in — for now this is probably acceptable.

---

llun (commented):
Second-round review. The previous round's findings have all been addressed cleanly:

- ✅ `normalizeActorId` now strips fragments on both sides of the owner/actor comparison (plus a dedicated test that fragment-only differences are accepted).
- ✅ `isRecord` extracted to `lib/utils/typeGuards.ts` and reused by both call sites.
- ✅ Tracing moved into `getSenderPublicKeyDetails` so the `guard.getSenderPublicKey` span and `span.recordException` are restored.
- ✅ Malformed JSON is now caught in `parseJsonBody` and surfaced via `logger.warn`, and the outer catch in `getSenderPublicKeyDetails` also logs (with the test verifying the body is not leaked into the log).
- ✅ Route handler now imports the shared `isRecord` and adds a `!normalizeActorId(body.actor)` check that also rejects empty / `#`-only actor strings, with explicit test coverage.
- ✅ The new owner-validation step (`validateOwnerActorKey`) plus the `actor.data.publicKey.owner !== actor.data.id` invariant materially strengthen the bind — the test matrix covering forged owner, mismatched key id, mismatched PEM, key delegation, and key URLs that the actor doesn't confirm is excellent.

Remaining items (none block correctness; one is worth considering before merge):

1. **Redundant HTTP fetch in the Mastodon path** — for `keyId = '<actor>#main-key'`, `requiresOwnerValidation` is always true, so we re-fetch the owner URL which is the same URL we just hit (the fragment is stripped by HTTP). This doubles outbound requests for the most common federation case. A small short-circuit when `normalizeActorId(document.keyId) === document.actorId` keeps the security guarantee and removes the second round-trip.

2. **No fragment normalization on `owner`** in `validateOwnerActorKey` — minor, only affects servers that put a fragment in `publicKey.owner`, but those documents would currently be rejected.

3. **Strict equality on `publicKey.owner !== actor.data.id`** — same theme; fine as-is unless real-world interop issues show up.

Nice security hardening overall.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +150,29 @@ export const ActivityPubVerifySenderGuard =
...
+    }
+
+    if (activityActor.actor) {
+      const normalizedOwner = normalizeActorId(senderPublicKey.owner)
+      const normalizedActor = normalizeActorId(activityActor.actor)
+
+      if (!normalizedOwner || normalizedOwner !== normalizedActor) {
+        return guardErrorResponse(request, 403, allowedMethods)
+      }
+    }
```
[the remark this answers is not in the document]
llun: Addressed in 8295e541. POST activity actor extraction/validation now runs immediately after digest verification and before federation policy, public-key lookup, and signature verification, so malformed or missing actor bodies fail before the expensive work. I also added guard tests asserting invalid actor bodies do not call those downstream steps.

lib/services/guards/getSenderPublicKey.ts
```diff
@@ -1,45 +1,285 @@
...
+  body: string
+  keyId: string
+}): ParsedSenderPublicKey | null => {
+  const json = parseJsonBody({ body, actorId: keyId })
+  if (!json) return null
+
+  const actor = Actor.safeParse(json)
+  if (actor.success) {
+    if (actor.data.id !== keyId && actor.data.publicKey.id !== keyId) {
+      return null
```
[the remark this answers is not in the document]
llun: Addressed in 8295e541. `keyId`, actor IDs, and public-key IDs are now normalized for URI comparison, including scheme/host casing, before equality checks. Added regression coverage for differently-cased key IDs.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 8295e541. Actor documents fetched through a fragment key ID now treat `normalizeActorId(keyId) === actor.id` as the already-validated canonical actor fetch, so the common `<actor>#main-key` path returns after one request. Path-based key URLs and bare PublicKeyDocument responses still go through owner actor validation.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 8295e541. Owner actor validation now normalizes the owner first, fetches the normalized owner URI, and compares the returned actor/key IDs with normalized URI helpers. Added coverage for a bare PublicKeyDocument whose owner carries a fragment.

lib/services/guards/getSenderPublicKey.ts
llun: Addressed in 8295e541 as part of the same normalization pass. Actor document validation now compares `publicKey.owner` and actor ID through `normalizeActorId`, preserving the security check while accepting owner values that only differ by fragment. Added a regression test for that case.

llun:
@codex review

llun (commented):
Third-round review. All previous concerns have been addressed cleanly:

- ✅ **Redundant Mastodon fetch eliminated** — `requiresOwnerValidation` is now `normalizedActorId !== normalizedKeyOwner`, so a fragment-only keyId no longer triggers the owner re-fetch. The new test `'accepts actor documents fetched by fragment key id without refetching the owner actor'` asserts `fetchMock` is called exactly once — exactly the short-circuit I suggested last round.
- ✅ **Owner normalized in `validateOwnerActorKey`** — `normalizeActorId(owner)` is used both for the fetch URL and the equality check against `ownerDocument.actorId`. The new test `'accepts public key documents whose owner differs from the actor id only by fragment'` covers the legitimate-fragment-in-owner case.
- ✅ **`publicKey.owner` normalization** — `parseSenderPublicKey` now normalizes both sides of `actor.data.publicKey.owner !== actor.data.id`. Covered by `'accepts actor documents whose public key owner differs from the actor id only by fragment'`.

Bonus improvements I noticed:
- The new `normalizeActivityPubUri` lowercases the scheme and hostname, which improves interop for servers that respond with mixed-case URIs (`'HTTPS://REMOTE.TEST/...'` is exercised by a dedicated test).
- The activity-actor extraction is now performed before `canFederateWithDomain`, so a malformed body fails fast with 400 without touching network/policy code. The added assertions in `'rejects POST requests when the date header is too stale'` and the new actor-validation tests pin the early-exit behavior.

LGTM — no remaining blockers from my side. Two very minor cosmetic notes you can take or leave:

1. `parseJsonBody` takes a parameter named `actorId` but the caller in `parseSenderPublicKey` always passes a `keyId`. Renaming to `keyId` (or just `context: string`) would match the log field semantics.
2. The 410 fallback path uses `${url.protocol}//${url.host}/actor#main-key`. Once you're already manipulating a `URL` object, `new URL('/actor#main-key', url).toString()` is a touch sturdier (handles non-default ports, etc.).

Nice work on the hardening and the test matrix.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -53,6 +55,27 @@ const getExpectedSha256Digest = (digestHeader: string) =>
...
+      !normalizeActorId(body.actor)
+    ) {
+      return { actor: null, valid: false }
+    }
+
+    return { actor: body.actor, valid: true }
+  } catch {
+    return { actor: null, valid: false }
+  }
+}
```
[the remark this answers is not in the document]
llun: Addressed in 7943dba2. The guard now reuses the digest-read body text for POST JSON parsing, passes the normalized activity body through `activityBody` in the guard context, and both inbox handlers consume that context instead of reparsing. Route fallback parsing is caught and returns 400; added invalid JSON/body-consumption regression tests.

llun:
@codex review

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +176,28 @@ export const ActivityPubVerifySenderGuard =
...
     const requestUrl = new URL(request.url, `http://${host}`)
     const requestTarget = `${request.method.toLowerCase()} ${requestUrl.pathname}${requestUrl.search}`
-    const publicKey = await getSenderPublicKey(database, signatureParts.keyId)
-    if (!(await verify(requestTarget, request.headers, publicKey))) {
+    const senderPublicKey = await getSenderPublicKeyDetails(
+      database,
+      signatureParts.keyId
+    )
+    if (
+      !(await verify(requestTarget, request.headers, senderPublicKey.publicKey))
```
llun, line 184: **Bug: `verify` may be called with an empty public key, causing a crash**

`getSenderPublicKeyDetails` returns `EMPTY_PUBLIC_KEY_DETAILS = { owner: null, publicKey: '' }` on any lookup failure (network error, 410 Gone fallback, etc.). Calling `verify(…, '')` will then throw a crypto error ("Invalid key" or similar), which propagates as an unhandled exception and returns a 500 instead of the intended 400/403.

Add an explicit guard before this call:

```typescript
if (!senderPublicKey.publicKey) {
  return guardErrorResponse(request, 400, allowedMethods)
}
if (!(await verify(requestTarget, request.headers, senderPublicKey.publicKey))) {
```

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -127,10 +176,28 @@ export const ActivityPubVerifySenderGuard =
...
     }
 
-    return handle(request, { database, params: context.params })
+    if (activity.actor) {
+      const normalizedOwner = normalizeActorId(senderPublicKey.owner)
+      const normalizedActor = normalizeActorId(activity.actor)
+
+      if (!normalizedOwner || normalizedOwner !== normalizedActor) {
+        return guardErrorResponse(request, 403, allowedMethods)
+      }
```
llun, line 195: **Missing null guard on `normalizedActor` makes the binding condition fragile**

`normalizeActorId(activity.actor)` can return `null` if the URL is somehow malformed at this point. The current condition only checks `!normalizedOwner`, so if `normalizedOwner` is a valid string and `normalizedActor` is `null`, the comparison `normalizedOwner !== null` is `true` and the request is correctly rejected — but this depends on an implicit side effect rather than an explicit check.

More importantly, if a future refactor changes `getPostActivity` to allow an actor that fails normalization to pass through (it currently validates at line 78), the binding check here would silently become ineffective.

Fix by guarding both:

```typescript
if (!normalizedOwner || !normalizedActor || normalizedOwner !== normalizedActor) {
  return guardErrorResponse(request, 403, allowedMethods)
}
```

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -104,7 +104,9 @@ export const POST = traceApiRoute(
               })
             }
 
-            const parsed = Activity.safeParse(await req.json())
+            const parsed = Activity.safeParse(
+              context.activityBody ?? (await req.json())
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts:108]: **Latent security risk: fallback to `req.json()` bypasses guard's actor binding**

`context.activityBody` is `null` for GET requests (guard sets `activity.body = null`). This route only handles POST, so `activityBody` is always the guard's validated, actor-bound body in practice — but the `?? (await req.json())` fallback silently reads an **unvalidated** request body if `activityBody` is ever null for a POST.

If the guard's behavior changes (e.g. passes `null` in an error path), this fallback would feed unverified JSON to the activity handler, circumventing the signer→actor binding that this PR introduces.

Since this route is POST-only, remove the fallback and require `activityBody` to be present:

```typescript
if (!context.activityBody) {
  return apiResponse({
    req,
    allowedMethods: CORS_HEADERS,
    data: ERROR_400,
    responseStatusCode: 400
  })
}
const parsed = Activity.safeParse(context.activityBody)
```

---

app/api/inbox/route.ts
```diff
@@ -12,24 +13,40 @@ import {
...
 
 export const OPTIONS = defaultOptions(CORS_HEADERS)
 
-const isRecord = (value: unknown): value is Record<string, unknown> =>
-  typeof value === 'object' && value !== null && !Array.isArray(value)
+const getActivityBody = async (
+  request: Request,
+  activityBody: unknown | null
+) => {
+  if (activityBody !== null) return activityBody
```
llun, line 28: **`!== null` check doesn't catch `undefined`, silently falling through to `request.json()`**

If `activityBody` is `undefined` (not in the declared type `unknown | null`, but possible through `as` casts or future type drift), `activityBody !== null` is `true` so `undefined` is returned — then on line 41 `isRecord(undefined)` is `false` and the route returns 400. That's safe but surprising.

The bigger concern is the symmetric issue in the user inbox (see other comment): any path where the guard provides `null` here falls back to re-reading the raw request body. For the shared inbox this route is POST-only, so replace the fallback with a hard guard matching the user inbox recommendation.

---

llun (commented):
## Code Review: `fix: bind shared inbox actor to HTTP signature key`

The signer→actor binding is the right fix for the forged Delete/Undo vulnerability, and the overall structure is sound. Four issues to address before treating this as complete:

1. **`verify` crash on empty public key** — `getSenderPublicKeyDetails` returns an empty key string on any lookup failure; calling `verify` with it throws a crypto error and surfaces as a 500. Needs an explicit guard before the `verify` call.

2. **Missing `!normalizedActor` in binding condition** — the condition only explicitly guards `normalizedOwner`; if `normalizedActor` is ever null the rejection is implicit and fragile. Add `!normalizedActor` to make the invariant explicit.

3. **Unvalidated `req.json()` fallback in user inbox** — `context.activityBody ?? (await req.json())` falls through to a raw body read if `activityBody` is null for any reason, bypassing the guard's binding. Remove the fallback; a null `activityBody` on a POST route should be a hard 400.

4. **`!== null` vs `undefined` in shared inbox `getActivityBody`** — minor type gap; same structural fix recommended as #3.

---

PR #662, opened 2026-05-12
fix: remove admin env var secret reveal path

llun:
### Motivation
- Close an information-disclosure vulnerability where the admin system page exposed raw `process.env` values via a server action callable from the browser. 
- Ensure operator-held secrets (database, auth, email, queue, etc.) are not retrievable by application admin accounts.

### Description
- Delete the server action `revealEnvVar` so there is no server-side endpoint that returns `process.env` values to the client.  
- Change the admin system page to collect and pass only `ACTIVITIES_*` environment variable names (not values) to the UI.  
- Simplify `EnvironmentVariables` to always render masked placeholders and remove all client-side reveal/toggle logic and the `revealEnvVar` prop.  
- Update wording in the UI to reflect that values are hidden rather than revealable.

### Testing
- Ran `yarn run prettier --write .` which completed successfully.  
- Ran `yarn lint` which completed successfully with pre-existing warnings and no new errors.  
- Ran `yarn build` which completed successfully.  
- Ran `yarn test` which failed in this environment due to Node version incompatibility (current environment Node `v20.20.2` lacks `Array.fromAsync` used by an existing test); repository policy requires Node `24` for running node commands.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a00d59f0d60832894a6e4abfdbf8e61)


lib/components/admin/EnvironmentVariables.tsx
```diff
@@ -1,55 +1,16 @@
 'use client'
```
llun: The `'use client'` directive should be removed. After stripping all hooks (`useState`) and event handlers, this component is now purely static — it renders a fixed list with hardcoded `••••••••` placeholders and has no browser-side interactivity whatsoever. Keeping `'use client'` here causes Next.js to serialize the `variables` array (env var key names) to JSON and ship it to the browser as part of the client bundle, which is unnecessary and leaks the names of all `ACTIVITIES_*` env vars to any authenticated admin's browser. Dropping the directive turns it into a Server Component and avoids that serialization entirely.

---

app/(timeline)/admin/system/page.tsx
```diff
@@ -41,18 +24,11 @@ const Page = async () => {
     process.env.ACTIVITIES_PUSH_VAPID_EMAIL
   )
 
-  // Collect ACTIVITIES_* environment variables; mask sensitive values server-side
+  // Collect ACTIVITIES_* environment variable names only.
```
llun: Per project convention (CLAUDE.md): "Default to writing no comments. Only add one when the WHY is non-obvious." The code reads clearly — `filter` + `map([key])` already communicates that only keys are collected. This comment should be removed.

---

app/(timeline)/admin/system/page.tsx
```diff
@@ -41,18 +24,11 @@ const Page = async () => {
...
   const envVars = Object.entries(process.env)
     .filter(([key]) => key.startsWith(ENV_VAR_PREFIX))
     .sort(([a], [b]) => a.localeCompare(b))
```
llun: Even with values removed, the **names** of all `ACTIVITIES_*` env vars are still passed as a prop to `EnvironmentVariables`. Because that component is still `'use client'`, the full key list is serialized to JSON and sent to the browser. This reveals which services are configured (e.g., `ACTIVITIES_DATABASE_URL`, `ACTIVITIES_REDIS_PASSWORD`, etc.). If the component is converted to a Server Component (by removing `'use client'`), Next.js will render it on the server and nothing will be sent to the client, which is the correct outcome for a purely display component.

---

llun (commented):
## Review

The motivation is correct — `revealEnvVar` was a real vulnerability and removing it is the right call. The direction of the fix is good, but there's one remaining issue that means env var data is still leaking to the browser, and a convention violation to clean up.

### Blocking: `EnvironmentVariables` is still a Client Component

`lib/components/admin/EnvironmentVariables.tsx` retains `'use client'` even though all hooks and event handlers have been removed. The component now renders a purely static list with hardcoded `••••••••` placeholders — there is zero browser-side logic left. Keeping `'use client'` means:

1. Next.js serializes the `variables` prop (all `ACTIVITIES_*` key names) to JSON and ships it to the browser.
2. Env var names like `ACTIVITIES_DATABASE_URL`, `ACTIVITIES_STRIPE_SECRET_KEY`, etc. are exposed to any authenticated admin's browser DevTools / network tab.

**Fix:** Remove the `'use client'` directive from `EnvironmentVariables.tsx`. The component will then be treated as a Server Component, rendered entirely on the server, and no key names will reach the client.

### Minor: unnecessary comment in `page.tsx`

The line `// Collect ACTIVITIES_* environment variable names only.` should be removed. The `filter` + `.map(([key]) => ({ key }))` chain already communicates this clearly, and per CLAUDE.md the project convention is to write no comments unless the WHY is non-obvious.

### Note: tests not verified on Node 24

The PR description acknowledges that `yarn test` was run on Node v20, where it fails due to `Array.fromAsync`. CLAUDE.md requires Node 24 for all node/yarn commands. Please re-run `yarn test` under Node 24 before merging to confirm no regressions.

---

llun (commented):
## Review: fix: remove admin env var secret reveal path

The security goal is correct — removing the `revealEnvVar` server action eliminates the path that could expose `process.env` values to admin users. The overall approach is sound.

---

### `'use client'` is no longer needed

After the changes, `EnvironmentVariables` has no React hooks, no state, and no event handlers — it is a pure render function. The `'use client'` directive should be removed to make it a Server Component. Keeping it a Client Component unnecessarily increases the JS bundle sent to the browser.

```diff
-'use client'
-
 import { FC } from 'react'
```

---

llun (commented):
One issue to fix:

**`'use client'` directive no longer needed (`lib/components/admin/EnvironmentVariables.tsx`, line 1)**

After removing the `useState`, `useCallback`, event handlers, and the `Eye`/`EyeOff` imports, the component has no client-side interactivity at all — it just maps over an array and renders static markup. The `'use client'` directive should be removed so Next.js can render it as a Server Component, keeping the component out of the client bundle entirely. Since `envVars` is now just `{ key: string }[]`, there is no client-side data that needs to be hydrated either.

---

lib/components/admin/EnvironmentVariables.tsx
llun: Fixed in 98673e64. `EnvironmentVariables` is now a Server Component and no longer imports `FC`, so the client component prop serialization and JS boundary are removed. One nuance: the `ACTIVITIES_*` names are still intentionally rendered as visible admin UI; this fix removes the client serialization path while keeping values hidden.

app/(timeline)/admin/system/page.tsx
llun: Fixed in f5ebcfcd. Removed the redundant comment from `app/(timeline)/admin/system/page.tsx`; the `Object.keys()`/filter/map chain now carries the intent directly.

app/(timeline)/admin/system/page.tsx
llun: Fixed in 98673e64. The component receiving `envVars` is no longer a Client Component, so Next.js does not create a client-boundary prop serialization path for this list. The names remain visible in the rendered admin page by design; values stay hidden.

app/(timeline)/admin/system/page.tsx
```diff
@@ -41,18 +24,11 @@ const Page = async () => {
...
-    .map(([key, value]) => {
-      const isSensitive = isSensitiveKey(key)
-      return {
-        key,
-        value: isSensitive ? null : (value ?? ''),
-        isSensitive
-      }
-    })
+    .map(([key]) => ({ key }))
```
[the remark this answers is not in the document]
llun: Fixed in f5ebcfcd. `envVars` is now built from `Object.keys(process.env)`, sorted by key, and mapped to `{ key }` without materializing discarded values.

lib/components/admin/EnvironmentVariables.tsx
```diff
@@ -1,55 +1,16 @@
...
 
-import { Eye, EyeOff } from 'lucide-react'
-import { FC, useState } from 'react'
-
-import { Button } from '@/lib/components/ui/button'
+import { FC } from 'react'
```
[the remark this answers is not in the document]
llun: Fixed in 98673e64. `EnvironmentVariables` no longer has a client boundary and no longer imports React types just to render this static list.

lib/components/admin/EnvironmentVariables.tsx
```diff
@@ -1,55 +1,16 @@
...
-    } finally {
-      setLoading((prev) => {
-        const next = new Set(prev)
-        next.delete(key)
-        return next
-      })
-    }
-  }
-
+export const EnvironmentVariables: FC<Props> = ({ variables }) => {
```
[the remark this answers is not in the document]
llun: Fixed in 98673e64. Removed `use client` and the now-unneeded React `FC` import, leaving this as a typed Server Component.

llun:
Review-body follow-up:

- Node 24 verification is now complete on this branch: `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` were run under Node v24.13.0. Lint reports 0 errors with the existing 12 `no-explicit-any` warnings; Jest reports 259 suites and 1,937 tests passing.
- Copilot’s suppressed empty-state note is fixed in 98673e64: the empty state now says `No ACTIVITIES_* environment variables found`, matching the filtered list.
- Gemini’s review summary items are fixed by 98673e64 (`EnvironmentVariables` is no longer client-rendered) and f5ebcfcd (`Object.keys(process.env)`).
- The two Claude review-body notes from 04:56 and 04:57 UTC both requested removing the `use client` directive; that is fixed in 98673e64.

llun:
@codex review

llun (commented):
## Re-review

All previous concerns are addressed in `cf2e26f` and `aead808`:

- `'use client'` directive removed — `EnvironmentVariables` is now a Server Component, so env var key names are rendered on the server and never serialized to the browser.
- `FC` import and all client-side state/handlers cleaned up; `Object.entries` simplified to `Object.keys` since values are no longer needed.
- The `// Collect ACTIVITIES_* environment variable names only.` comment was removed.
- Empty-state copy improved to "No ACTIVITIES_* environment variables found", which is more accurate.

### Nit (non-blocking)

`EnvVar` is now a wrapper around a single `key: string`. You could simplify the prop to `variables: string[]` and render `envVar` directly, dropping the `EnvVar` interface entirely. Keep or drop — purely stylistic.

### Reminder

CLAUDE.md requires `yarn test` to be run on Node 24 before merging. The PR description still notes tests failed on Node v20 — please confirm a clean test run on Node 24 prior to merge.

Otherwise this looks good to me.

---

lib/components/admin/EnvironmentVariables.tsx
```diff
@@ -1,60 +1,17 @@
...
-  }
-
+export const EnvironmentVariables = ({ variables }: Props) => {
   return (
     <div className="space-y-2">
       {variables.length === 0 ? (
         <p className="text-sm text-muted-foreground">
-          No environment variables found
+          No ACTIVITIES_* environment variables found
         </p>
```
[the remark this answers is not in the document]
llun: Addressed by removing the environment-variable listing surface instead of preserving the empty state. `EnvironmentVariables` and `adminConstants` are deleted, and `/admin/system` no longer enumerates or renders `ACTIVITIES_*` variable names, so there is no hardcoded prefix or empty-state branch left to maintain.

llun:
Review follow-up after `b725ac36`:

- Gemini's empty-state/`ENV_VAR_PREFIX` suggestion is superseded by removing the environment-variable listing entirely: `EnvironmentVariables` and `adminConstants` are deleted, and `/admin/system` no longer enumerates or renders `ACTIVITIES_*` variable names.
- Claude's re-review nit about simplifying the `EnvVar` wrapper is no longer applicable because that component/type was removed.
- Node 24 verification is complete on the pushed branch: `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all passed under Node v24.11.1. Lint still reports 0 errors with the existing 12 warnings; Jest reports 264 suites and 1,983 tests passing.

All inline review threads are resolved.

PR #663, opened 2026-05-12
fix: avoid persisting PAT credentials in sync-main-to-oltp workflow

llun:
### Motivation
- Prevent a long-lived `secrets.PAT_TOKEN` from being persisted by `actions/checkout` where repository-controlled code (merge content, dependencies, or AI actions) could read the credential and exfiltrate or misuse it.

### Description
- Add `persist-credentials: false` to the `actions/checkout@v6` step to avoid persisting checkout credentials into local git config.
- Scope PAT usage to the final push step by injecting `PAT_TOKEN` only into that step (`env: PAT_TOKEN: ${{ secrets.PAT_TOKEN }}`) and setting an authenticated remote with `git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"` immediately before `git push`.

### Testing
- Ran `yarn run prettier --write .` which completed successfully.
- Ran `yarn lint` and `yarn build` which completed successfully in this environment.
- Ran `yarn test` which failed due to a native module mismatch (`better-sqlite3` compiled for a different Node version), unrelated to the workflow YAML change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a00d5afbb008328b82e85c687d27f91)

### DIFF
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -43,6 +43,7 @@ jobs:
           ref: oltp
           fetch-depth: 0
           token: ${{ secrets.PAT_TOKEN }}
+          persist-credentials: false
 
       - name: Enable corepack
         shell: bash
@@ -177,4 +178,8 @@ jobs:
       - name: Push oltp
         if: steps.synced.outputs.skip == 'false'
         shell: bash
-        run: git push origin HEAD:oltp
+        env:
+          PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
+        run: |
+          git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+          git push origin HEAD:oltp
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -42,7 +42,6 @@ jobs:
         with:
           ref: oltp
           fetch-depth: 0
-          token: ${{ secrets.PAT_TOKEN }}
           persist-credentials: false
 
       - name: Enable corepack
@@ -180,6 +179,4 @@ jobs:
         shell: bash
         env:
           PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
-        run: |
-          git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
-          git push origin HEAD:oltp
+        run: git push "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:oltp
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -179,4 +179,6 @@ jobs:
         shell: bash
         env:
           PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
-        run: git push "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:oltp
+        run: |
+          git -c credential.helper='!f() { echo "username=x-access-token"; echo "password=${PAT_TOKEN}"; }; f' \
+            push origin HEAD:oltp


.github/workflows/sync-main-to-oltp.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           ref: oltp
           fetch-depth: 0
           token: ${{ secrets.PAT_TOKEN }}
+          persist-credentials: false
```
llun, line 45: The `token: ${{ secrets.PAT_TOKEN }}` is still passed here alongside `persist-credentials: false`. This means PAT is still used at checkout, which contradicts the PR description's claim to "Scope PAT usage to the final push step only."

`persist-credentials: false` prevents the token from being written to the git credential store after checkout, but the token is still consumed at this point and will be available to the `actions/checkout` action itself.

If the `oltp` branch is accessible with the default `GITHUB_TOKEN` (i.e., no special PAT access is required to read it), removing `token` here entirely would be the cleanest fix — then PAT would truly only appear at the push step. If PAT is genuinely required for checkout, that should be noted in the PR description.

---

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -177,4 +178,8 @@ jobs:
       - name: Push oltp
         if: steps.synced.outputs.skip == 'false'
         shell: bash
-        run: git push origin HEAD:oltp
+        env:
+          PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
+        run: |
+          git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
llun: `git remote set-url` writes the token-embedded URL into `.git/config` in plain text. This is actually **worse** than `persist-credentials: true` from `actions/checkout`, because:

1. The credential is now stored as a URL in `.git/config`, readable by anything that reads that file.
2. It persists in the config for the remainder of the job (and any steps added after this one in future edits).

A safer approach is to pass the authenticated URL directly to `git push` without modifying the stored remote config at all:

```bash
git push "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:oltp
```

This avoids writing the credential anywhere in the git config. If you must use `set-url`, at minimum reset it back to a clean form after the push:

```bash
git remote set-url origin "https://github.com/${GITHUB_REPOSITORY}.git"
```

---

llun (commented):
The intent here is good — scoping PAT usage is the right direction — but the implementation has two issues worth fixing before merging.

**Issue 1: PAT is still used at checkout.**
`persist-credentials: false` prevents the token from being stored post-checkout, but `token: ${{ secrets.PAT_TOKEN }}` is still passed to the checkout action. If the `oltp` branch is readable with the default `GITHUB_TOKEN`, remove `token` from the checkout step entirely so PAT truly only appears in the push step.

**Issue 2: `git remote set-url` embeds the credential in `.git/config`.**
This is at least as risky as `persist-credentials: true` — the token now sits in the git config in plain text for the duration of the job. Use `git push <url> HEAD:oltp` directly (passing the authenticated URL as an argument) to avoid writing the credential anywhere in the config file:

```bash
git push "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:oltp
```

See inline comments for details and suggested fixes.

---

llun (commented):
## Review: fix: avoid persisting PAT credentials in sync-main-to-oltp workflow

The approach of using `persist-credentials: false` combined with injecting the token only at the push step is the right pattern. However, it introduces a breakage in steps that still need remote git access.

---

### The "Fetch main and oltp" step will fail

`persist-credentials: false` prevents the checkout action from writing credentials into `.git/config`. After that, any subsequent `git remote` operation has no stored credentials.

The "Fetch main and oltp" step runs after checkout:

```yaml
- name: Fetch main and oltp
  shell: bash
  run: |
    git fetch --no-tags origin main oltp
    git rev-parse "${MAIN_SHA}^{commit}"
```

With no credentials stored and no credential helper configured, this `git fetch` will fail with an authentication error on a private repository.

**Suggested fix**: Set the authenticated remote URL once in a dedicated step placed right after checkout, and remove the `git remote set-url` from the push step:

```yaml
- name: Configure authenticated remote
  shell: bash
  env:
    PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
  run: git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```

This keeps the token out of the checkout step's stored credentials while making it available for all subsequent git operations in the job.

---

### `claude-code-action` may also need remote git access

The `claude-code-action@v1` step runs Claude with `--dangerously-skip-permissions` and may execute `git fetch` internally during conflict resolution. Those operations will also fail without stored credentials. The authenticated-remote approach above fixes this transparently since Claude inherits the already-configured remote URL.

---

llun (commented):
One issue to fix:

**Token embedded in the remote URL may appear in git error output (`.github/workflows/sync-main-to-oltp.yml`, line 184)**

```yaml
git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```

When the `PAT_TOKEN` is injected via `env:` and expanded into the URL string, git may include that full URL (with the token) in error messages or verbose output. GitHub Actions automatically redacts values referenced directly as `${{ secrets.PAT_TOKEN }}`, but a token that has been interpolated into a longer string is not guaranteed to be redacted in all output paths.

A safer approach that avoids embedding the token in the URL is to configure git's credential helper inline:

```yaml
run: |
  git -c credential.helper='!f() { echo "username=x-access-token"; echo "password=${PAT_TOKEN}"; }; f' \
    push origin HEAD:oltp
```

This keeps the token out of the remote URL entirely while still authenticating the push.

---

.github/workflows/sync-main-to-oltp.yml
llun: Fixed in 724da86f135d6063825be960a17bf11ca383ca31. The checkout step no longer passes `secrets.PAT_TOKEN`; it keeps `persist-credentials: false`, so the PAT is not exposed to `actions/checkout`.

.github/workflows/sync-main-to-oltp.yml
llun: Fixed in 724da86f135d6063825be960a17bf11ca383ca31. The push step no longer writes the PAT into `origin` or `.git/config`; follow-up a67e349ca6370a0c63df100ec4ba2421b3d12213 keeps that property while moving auth to an inline credential helper.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           ref: oltp
           fetch-depth: 0
           token: ${{ secrets.PAT_TOKEN }}
```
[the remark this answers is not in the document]
llun: Fixed in 724da86f135d6063825be960a17bf11ca383ca31. Checkout now relies on the default token path and no longer receives the long-lived PAT.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -177,4 +178,8 @@ jobs:
       - name: Push oltp
         if: steps.synced.outputs.skip == 'false'
         shell: bash
-        run: git push origin HEAD:oltp
+        env:
+          PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
+        run: |
+          git remote set-url origin "https://x-access-token:${PAT_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+          git push origin HEAD:oltp
```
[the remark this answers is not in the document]
llun: Fixed in 724da86f135d6063825be960a17bf11ca383ca31. The push step no longer persists the PAT in git remote config; follow-up a67e349ca6370a0c63df100ec4ba2421b3d12213 now supplies the PAT through an inline credential helper instead of a token URL.

llun:
Responding to the later top-level review about configuring an authenticated remote after checkout: I am not implementing that suggestion for this repository. `llun/activities.next` is public, `oltp` is in the same repo, and the branch is not protected, so checkout/fetch do not need PAT-backed credentials. Setting the authenticated remote immediately after checkout would write the PAT into `.git/config` for the rest of the job, including the Claude action and later repo-controlled commands, which is the exposure this PR is trying to avoid.

Fixed the actionable inline checkout and remote-config items in 724da86f135d6063825be960a17bf11ca383ca31. Follow-up a67e349ca6370a0c63df100ec4ba2421b3d12213 keeps the PAT out of checkout and remote config, and also avoids embedding it in the push URL by using an inline credential helper.

llun:
Addressed the token-in-push-URL review in a67e349ca6370a0c63df100ec4ba2421b3d12213. The final push now uses an inline `git -c credential.helper=...` helper with the helper body single-quoted, so `${PAT_TOKEN}` is not expanded into the shell command, Git argv, remote URL, or `.git/config`; it is expanded only inside the helper subprocess when Git requests credentials.

llun:
@codex review

PR #667, opened 2026-05-12
fix: authenticate OLTP sync push with PAT header

llun:
## Summary
- keep checkout credentials non-persistent for the OLTP sync workflow
- authenticate only the final `git push origin HEAD:oltp` using a temporary PAT-backed HTTP authorization header
- clear the temporary git header with a trap and disable credential helpers for the push

## Root Cause
The workflow disabled persisted checkout credentials, then relied on an inline Git credential helper for the final push. On the runner, another credential helper could still satisfy Git before the inline PAT helper, causing GitHub to reject the push with `Invalid username or token`.

## Validation
- `yarn lint`
- `yarn build`
- `yarn test`

Base branch: `main`

### DIFF
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -185,20 +185,15 @@ jobs:
             exit 1
           fi
 
-          if printf '%s' "${PAT_TOKEN}" | grep -q $'[\r\n]'; then
+          if [[ "${PAT_TOKEN}" == *$'\r'* || "${PAT_TOKEN}" == *$'\n'* ]]; then
             echo "secrets.PAT_TOKEN must not contain line breaks" >&2
             exit 1
           fi
 
-          auth_header="$(printf 'x-access-token:%s' "${PAT_TOKEN}" | base64 | tr -d '\n')"
+          auth_header="$(printf 'x-access-token:%s' "${PAT_TOKEN}" | base64 -w 0)"
           echo "::add-mask::${auth_header}"
 
-          cleanup() {
-            git config --local --unset-all http.https://github.com/.extraheader || true
-          }
-
-          trap cleanup EXIT
-          cleanup
-
-          git config --local http.https://github.com/.extraheader "AUTHORIZATION: basic ${auth_header}"
-          git -c credential.helper= push origin HEAD:oltp
+          git \
+            -c http.https://github.com/.extraheader="AUTHORIZATION: basic ${auth_header}" \
+            -c credential.helper= \
+            push origin HEAD:oltp
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -180,5 +180,25 @@ jobs:
         env:
           PAT_TOKEN: ${{ secrets.PAT_TOKEN }}
         run: |
-          git -c credential.helper='!f() { echo "username=x-access-token"; echo "password=${PAT_TOKEN}"; }; f' \
-            push origin HEAD:oltp
+          if [ -z "${PAT_TOKEN}" ]; then
+            echo "Missing secrets.PAT_TOKEN" >&2
+            exit 1
+          fi
+
+          if printf '%s' "${PAT_TOKEN}" | grep -q $'[\r\n]'; then
+            echo "secrets.PAT_TOKEN must not contain line breaks" >&2
+            exit 1
+          fi
+
+          auth_header="$(printf 'x-access-token:%s' "${PAT_TOKEN}" | base64 | tr -d '\n')"
+          echo "::add-mask::${auth_header}"
+
+          cleanup() {
+            git config --local --unset-all http.https://github.com/.extraheader || true
+          }
+
+          trap cleanup EXIT
+          cleanup
+
+          git config --local http.https://github.com/.extraheader "AUTHORIZATION: basic ${auth_header}"
+          git -c credential.helper= push origin HEAD:oltp


.github/workflows/sync-main-to-oltp.yml
```diff
@@ -180,5 +180,25 @@ jobs:
...
+
+          cleanup() {
+            git config --local --unset-all http.https://github.com/.extraheader || true
+          }
+
+          trap cleanup EXIT
+          cleanup
+
+          git config --local http.https://github.com/.extraheader "AUTHORIZATION: basic ${auth_header}"
+          git -c credential.helper= push origin HEAD:oltp
```
llun: Consider passing the auth header inline via `git -c` instead of writing it to `.git/config` and cleaning it up with a trap. The runner is ephemeral and we only do one push, so the local-config + trap dance adds complexity without buying anything over:

```bash
git -c http.https://github.com/.extraheader="AUTHORIZATION: basic ${auth_header}" \
    -c credential.helper= \
    push origin HEAD:oltp
```

This removes the `cleanup`/`trap`/pre-cleanup lines entirely, and there is no window where the credential is persisted in `.git/config` — which is the safer default and also eliminates the small risk that an unrelated failure between `git config --local ... extraheader` and the `trap` firing leaves the header in place.

---

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -180,5 +180,25 @@ jobs:
...
-          git -c credential.helper='!f() { echo "username=x-access-token"; echo "password=${PAT_TOKEN}"; }; f' \
-            push origin HEAD:oltp
+          if [ -z "${PAT_TOKEN}" ]; then
+            echo "Missing secrets.PAT_TOKEN" >&2
+            exit 1
+          fi
+
+          if printf '%s' "${PAT_TOKEN}" | grep -q $'[\r\n]'; then
+            echo "secrets.PAT_TOKEN must not contain line breaks" >&2
+            exit 1
```
llun, line 190: Nit: `printf '%s' "${PAT_TOKEN}" | grep -q $'[\r\n]'` is a useful guard, but `$'...'` is bash-specific ANSI-C quoting. The step already declares `shell: bash`, so this works — just worth noting it will break if anyone later changes the shell to `sh`. A more portable alternative is `case "${PAT_TOKEN}" in *[$'\r\n']*) ... ;; esac` (still bash-specific) or stripping with `tr -d '\r\n'` upfront so any stray whitespace is normalized rather than fatal.

Also: `grep -q` on the token value is fine here because `grep` is exec'd (PAT value is on stdin, not argv), but the encoder one line below passes the PAT to `printf` as an argv argument. Since `printf` is a bash builtin no separate process is forked, so this is safe — just confirm no one swaps `printf` for `/usr/bin/printf` later.

---

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -180,5 +180,25 @@ jobs:
...
+          fi
+
+          auth_header="$(printf 'x-access-token:%s' "${PAT_TOKEN}" | base64 | tr -d '\n')"
```
llun: Minor: `base64 | tr -d '\n'` works, but `base64 -w 0` is the idiomatic GNU coreutils way to produce a single line and avoids relying on the default wrap-at-76 behavior (which differs on BSD/macOS `base64`). Since the runner is pinned to `ubuntu-latest` it's fine, but `-w 0` makes the intent clearer.

---

llun (commented):
The diagnosis (a system/global credential helper winning over the inline `-c credential.helper=...` helper) and the fix direction (switch to `http.<url>.extraheader`, which bypasses helpers entirely) both look correct. The new step also adds nice defensive checks: PAT presence, CR/LF guard, log masking of the encoded header, and `-c credential.helper=` to suppress helpers on the push itself. `persist-credentials: false` on checkout pairs well with this.

A few suggestions inline — the main one is that the `extraheader` can be passed inline via `git -c` rather than written to `.git/config` with a trap, which removes a moving part. The other two are nits about portability of the base64/CR-LF helpers.

Nothing here blocks merging; treat as polish.

---

.github/workflows/sync-main-to-oltp.yml
llun: Updated in 328178d8: the CR/LF guard now checks `PAT_TOKEN` directly with Bash pattern matching instead of piping the value through `grep`. The encoder still uses shell builtin `printf`.

.github/workflows/sync-main-to-oltp.yml
llun: Implemented in 328178d8: the push now passes the PAT auth header inline with `git -c http.https://github.com/.extraheader=...`, clears credential helpers for that command, and no longer writes the header to local git config.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -180,5 +180,25 @@ jobs:
...
-            push origin HEAD:oltp
+          if [ -z "${PAT_TOKEN}" ]; then
+            echo "Missing secrets.PAT_TOKEN" >&2
+            exit 1
+          fi
+
+          if printf '%s' "${PAT_TOKEN}" | grep -q $'[\r\n]'; then
+            echo "secrets.PAT_TOKEN must not contain line breaks" >&2
+            exit 1
+          fi
```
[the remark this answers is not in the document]
llun: Fixed in 328178d8: replaced the `grep` validation with a direct Bash string check for `$'\r'` and `$'\n'`, so line breaks in `PAT_TOKEN` are detected before auth header generation.

.github/workflows/sync-main-to-oltp.yml
llun: Updated in 328178d8: switched the encoder to `base64 -w 0`, which matches the Ubuntu runner and makes the single-line output requirement explicit.

PR #668, opened 2026-05-12
fix: bind verified ActivityPub sender to activity actor

llun:
### Motivation
- Prevent untrusted `activity.actor` values from being used to perform cross-account state changes (Block/Undo) when HTTP signatures are verified for a different actor. 
- Ensure the inbox route only applies actor-scoped activities when the signing key is bound to the same actor claimed in the activity JSON. 

### Description
- Derive a normalized `verifiedSenderActorId` from the verified HTTP-signature `keyId` and pass it through the guard context from `ActivityPubVerifySenderGuard` by adding `verifiedSenderActorId` to the handler context. 
- Update the `ActivityPubVerifiedSenderHandle` type to include `verifiedSenderActorId` so callers must handle the verified sender identity. 
- Add a check in the per-user inbox handler to reject with `403` when `activity.actor` does not equal `context.verifiedSenderActorId`, preventing forged actor assertions from taking effect. 
- Update mocks and tests for the inbox route to provide the `verifiedSenderActorId` and add a regression test that asserts a mismatched actor is rejected with no side effects. 

### Testing
- Ran formatting with `yarn run prettier --write .` which completed successfully. 
- Ran lint with `yarn lint` which completed successfully (only unrelated warnings). 
- Built the project with `yarn build` which completed successfully. 
- Ran the full test suite with `yarn test` which failed in this environment due to an unrelated Node `Array.fromAsync` incompatibility in an existing test, not caused by this change. 
- Ran the targeted tests `lib/services/guards/ActivityPubVerifyGuard.test.ts` and `app/api/users/[username]/inbox/route.test.ts` with `yarn test lib/services/guards/ActivityPubVerifyGuard.test.ts app/api/users/[username]/inbox/route.test.ts`, and both test suites passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a00cf096094832898f6564fbf7e1412)


llun (commented):
The direction here is correct — binding the verified HTTP-signature key to the activity `actor` is the right mitigation for forged-actor attacks. However there are a few issues that need to be fixed before this can land.

---

## Critical: shared inbox (`/api/inbox`) is not protected

`app/api/inbox/route.ts` also uses `ActivityPubVerifySenderGuard`, and `verifiedSenderActorId` is now available in its handler context, but the actor check was not added there.

```ts
// app/api/inbox/route.ts – line 27
ActivityPubVerifySenderGuard(async (request, { database }) => {
  // ...
  const activity = body as unknown as StatusActivity
  // ⚠️  activity.actor is never compared to verifiedSenderActorId
  if (!(await canFederateWithDomain(database, activity.actor))) { ... }
```

The shared inbox processes `Create`, `Delete`, `Update`, `Announce`, and other activities. An attacker can sign with actor A's key and claim `actor: B` in the JSON body — the same class of vulnerability the PR is trying to fix. The actor check added to the per-user inbox should be mirrored here.

---

## Medium: `keyId.split('#')[0]` is fragile for non-Mastodon key formats

```ts
// lib/services/guards/ActivityPubVerifyGuard.ts – line 135
const verifiedSenderActorId = signatureParts.keyId.split('#')[0]
```

This works for the dominant convention (`https://example.com/users/alice#main-key`) but will produce a wrong result for servers that use a standalone key URL with no `#` fragment, e.g. `https://example.com/users/alice/publicKey`. In that case `split('#')[0]` returns the full key URL, which will never equal the `actor` field in the activity JSON, causing every request from such a server to receive a spurious 403.

A more robust approach is to resolve the actor from the key object. `getSenderPublicKey` already fetches the actor document via `getActorPerson`; the actor's `id` (from the fetched JSON) is the canonical actor URL and would be safe to use regardless of key ID format. That requires threading the actor ID back through `getSenderPublicKey` or doing a small refactor, but it eliminates the assumption about key URL structure.

---

## Minor: `ActivityPubVerifyGuard.test.ts` should assert the `verifiedSenderActorId` value

The existing guard tests verify that the handler is called, but none of them assert the value passed as `verifiedSenderActorId`. A test like the following would catch regressions:

```ts
it('passes verifiedSenderActorId derived from keyId to the handler', async () => {
  const handler = jest.fn().mockResolvedValue(Response.json({ ok: true }))
  const guard = ActivityPubVerifySenderGuard(handler)
  // ... set up a valid signed request with keyId="https://remote.test/users/alice#main-key" ...
  await guard(req, { params: Promise.resolve({}) })
  expect(handler).toHaveBeenCalledWith(
    expect.anything(),
    expect.objectContaining({ verifiedSenderActorId: 'https://remote.test/users/alice' })
  )
})
```

---

## Minor: shared inbox test mock is out of sync with the updated type

`app/api/inbox/route.test.ts` still defines the mock context without `verifiedSenderActorId`:

```ts
// lines 19-23
handle: (
  req: NextRequest,
  context: { database: typeof mockDatabase; params: Promise<{}> }  // missing verifiedSenderActorId
) => ...
handle(req, { database: mockDatabase, params: context.params })   // missing verifiedSenderActorId
```

Because `jest.mock` factories return `any`, TypeScript won't catch this today, but it should be updated to match `ActivityPubVerifiedSenderHandle` so the mock stays consistent with the real type. If the actor check is added to the shared inbox route (see first issue), a mismatch regression test analogous to the one added in `route.test.ts` should be added here too.

---

## Note: the PR has a merge conflict

`mergeable_state` is `dirty`. This will need to be resolved against `main` before merge regardless of the above.


---

app/api/users/[username]/inbox/route.ts
```diff
@@ -115,6 +115,14 @@ export const POST = traceApiRoute(
             }
 
             const activity = parsed.data
+            if (activity.actor !== context.verifiedSenderActorId) {
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts]: **The actor-binding check needs to be applied to the shared inbox too.**

`app/api/inbox/route.ts` is the *other* consumer of `ActivityPubVerifySenderGuard`, and it processes the same class of actor-scoped activities (Create / Update / Delete / Announce / Undo Announce — see `getJobMessage`). With this PR a peer whose key is verified for actor X is still able to POST `{ "type": "Delete", "actor": "https://victim/users/Y", "object": ... }` to `/api/inbox` and have it queued as a `DELETE_OBJECT_JOB_NAME` job against actor Y's objects — exactly the forged-actor pattern this PR is supposed to close.

Please add the equivalent `activity.actor !== context.verifiedSenderActorId → 403` guard in `app/api/inbox/route.ts` (right after the `body` is parsed and `activity.actor` is known), plus a regression test in `app/api/inbox/route.test.ts`. Otherwise the per-user inbox is hardened but the shared inbox still admits the same forgery.

---

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -132,5 +132,11 @@ export const ActivityPubVerifySenderGuard =
       return guardErrorResponse(request, 400, allowedMethods)
     }
 
-    return handle(request, { database, params: context.params })
+    const verifiedSenderActorId = signatureParts.keyId.split('#')[0]
```
llun: **`keyId.split('#')[0]` is not a safe derivation of the actor URL.**

The HTTP-signatures spec doesn't require `keyId` to be `<actor-id>#<fragment>`. Several real implementations publish a key URL that is *not* the actor URL:

- Mastodon uses `https://host/users/alice#main-key` → base equals actor ✓
- Some implementations (Pleroma forks, GoToSocial historically, custom servers) advertise `https://host/users/alice/main-key` or `https://host/keys/123` — no fragment, base ≠ actor.
- Some include query strings (`?key=1`); `split('#')` keeps those.
- Trailing slash / case / percent-encoding differences between the `keyId` URL and what the remote server publishes as `activity.actor` will also fail strict `!==`.

After this PR, any of those peers will be uniformly rejected with `403`, breaking federation rather than fixing a bug.

The correct binding is "the verified key belongs to the actor named in `activity.actor`". Practically that means: resolve the actor document of `activity.actor`, then assert that its advertised `publicKey.id` (or one of `publicKey[].id`) equals the verified `signatureParts.keyId`. The actor's own `id` from that document is what should be compared against `activity.actor` (with URL normalization). `getSenderPublicKey` is already doing the lookup — extend it (or add a sibling) to return both the key and the owning actor id, and pass *that* through as `verifiedSenderActorId`.

At minimum, please document the assumption "keyId is `<actorId>#fragment`" loudly and add a fallback path for keyIds without `#`.

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -115,6 +115,14 @@ export const POST = traceApiRoute(
...
+              return apiResponse({
+                req,
+                allowedMethods: CORS_HEADERS,
+                data: ERROR_403,
+                responseStatusCode: 403
+              })
+            }
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts]: **Top-level actor binding only — nested `object.actor` is still trusted.**

This check binds `activity.actor` to the verified signer, but several branches below still trust the *inner* actor:

- `Undo` of `Follow` (line 246-268): `undoFollowRequest` receives `request.object.actor` (the inner Follow's actor) and uses it for matching. A signer who is bound to outer `activity.actor = X` can still embed `object: { type: 'Follow', actor: 'Y', ... }` and undo Y's follow.
- `Undo` of `Like` (line 270-286): `database.deleteLike({ actorId: activity.actor, ... })` — uses outer actor, so OK after this PR.
- `Undo` of `Block` (line 288-309): `applyRemoteUnblock({ actorId: activity.actor, ... })` — outer actor, OK.
- `Accept` / `Reject` (line 136-167): consume `activity.actor` and an inner `object` (the original Follow). Worth confirming the inner Follow's `actor` still refers to a *local* user the responder can legitimately Accept/Reject for.

Recommend either asserting `object.actor === activity.actor` for nested actor-bearing objects, or pushing the binding check down into `undoFollowRequest` / `acceptFollowRequest` / `rejectFollowRequest`.

---

app/api/users/[username]/inbox/route.test.ts
```diff
@@ -184,6 +186,25 @@ describe('POST /api/users/[username]/inbox', () => {
...
+          actor: 'https://remote.test/users/bob',
+          object: 'https://activities.local/users/llun'
+        })
+      }),
+      { params: Promise.resolve({ username: 'llun' }) }
+    )
+
+    expect(response.status).toBe(403)
+    expect(mockCreateFollower).not.toHaveBeenCalled()
+  })
```
Comment (llun) [inline app/api/users/[username]/inbox/route.test.ts:226]: **Regression coverage is too narrow given the PR motivation.**

The PR description specifically calls out `Block` and `Undo` as the activities the forged-actor bug could be used against, but the only new test covers a `Follow` mismatch. Please add explicit cases for:

- A `Block` with `actor` ≠ verified sender → 403, and `applyRemoteBlock` not called.
- An `Undo` of a `Block` with outer `actor` ≠ verified sender → 403, and `applyRemoteUnblock` not called.
- A negative case for the inner-object actor too (e.g., `Undo` whose outer `actor` matches the verified sender but whose inner `object.actor` is a different remote actor) — see my comment on `route.ts` about nested-actor trust.

Also, the mock's hard-coded `verifiedSenderActorId: 'https://remote.test/users/alice'` is fine for these unit tests but means the test suite isn't exercising the actual `keyId.split('#')[0]` derivation in `ActivityPubVerifySenderGuard`. A guard-level test for keyIds with/without fragments, trailing slashes, and query strings would catch the federation-breakage risk flagged on the guard file.

---

lib/services/guards/types.ts
```diff
@@ -27,5 +27,9 @@ export type OptionalAuthenticatedApiHandle<P> = (
 
 export type ActivityPubVerifiedSenderHandle<P> = (
   request: NextRequest,
-  context: { database: Database; params: Promise<P> }
+  context: {
+    database: Database
+    params: Promise<P>
+    verifiedSenderActorId: string
+  }
```
llun, line 34: Minor: making `verifiedSenderActorId` a required field on every `ActivityPubVerifiedSenderHandle` context is a breaking change for the shared inbox handler too (`app/api/inbox/route.ts`). That's actually a good thing — it forced the call sites to acknowledge the new field — but it makes the omission in `app/api/inbox/route.ts` (see my comment there) more conspicuous: that handler now has the field in scope and just isn't using it. Worth wiring it up in the same PR rather than leaving "we exposed it but didn't enforce it" as a foot-gun for the next change.

---

llun (commented):
Good direction — binding the verified signer to `activity.actor` closes a real forgery vector. Before merging, a few issues need attention:

1. **Shared inbox isn't covered.** `app/api/inbox/route.ts` uses the same guard but doesn't apply the new check, so `Delete` / `Update` / `Announce` / `Undo Announce` forgery is still possible there. Details in the inline comment on `route.ts`.

2. **`keyId.split('#')[0]` is not a sound way to derive the actor URL.** It assumes the Mastodon-style `<actor>#main-key` convention; peers that publish a separate key URL, omit the fragment, or normalize URLs differently will be uniformly rejected with `403`, breaking federation. Recommend deriving `verifiedSenderActorId` from the actor document referenced by the resolved public key. Inline comment on `ActivityPubVerifyGuard.ts`.

3. **Nested object actors are still trusted** for `Undo Follow`, `Accept`, and `Reject`. Inline comment on `route.ts`.

4. **Test coverage is too narrow** — the motivation mentions Block and Undo but only a Follow mismatch is tested. Inline comment on `route.test.ts`.

5. **PR description note** — `yarn test` was reported as failing on an "unrelated" issue. Please confirm in a clean Node 24 env that the full suite (not just the two targeted files) passes before merging.

---

lib/services/guards/ActivityPubVerifyGuard.ts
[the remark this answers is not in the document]
llun: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. The guard no longer derives the sender actor with `keyId.split('#')[0]`; it uses `getSenderPublicKeyDetails`, validates the fetched key owner/actor document, and compares normalized owner/activity actor IDs with `normalizeActorId`. Requests fail closed when the owner or actor cannot be normalized.

lib/services/guards/ActivityPubVerifyGuard.ts
```diff
@@ -132,5 +132,11 @@ export const ActivityPubVerifySenderGuard =
...
+
+    return handle(request, {
+      database,
+      params: context.params,
+      verifiedSenderActorId
+    })
```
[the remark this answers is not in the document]
llun: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. Instead of clearing the hash from `keyId` in the route, the guard now binds the signed key to `senderPublicKey.owner` from validated key/actor documents and compares it with `normalizeActorId(activity.actor)`, which handles fragments and URL normalization without trusting the key URL shape.

app/api/users/[username]/inbox/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. The raw route-level `activity.actor !== verifiedSenderActorId` check is gone; guard-level binding now normalizes both the sender key owner and parsed activity actor before the inbox handlers run.

app/api/users/[username]/inbox/route.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662 by bringing the shared inbox onto the same guard-level actor binding. `/api/inbox` consumes the guard-provided `activityBody`, so mismatched signed actors are rejected before `getJobMessage` or queue publication. I did not duplicate a route-level equality check because the centralized guard enforcement covers both shared and per-user inboxes.

lib/services/guards/ActivityPubVerifyGuard.ts
llun: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. `getSenderPublicKeyDetails` now resolves and validates actor/public-key documents, supports path-based key IDs, verifies that the owner actor publishes the same key, and returns the verified owner. The old `keyId.split('#')[0]` path is gone; missing or invalid ownership returns empty details and the guard rejects.

app/api/users/[username]/inbox/route.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.ts]: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. Added normalized nested `object.actor` checks for full `Undo Follow` and full `Undo Block`; mismatches return 403 before `undoFollowRequest` or `applyRemoteUnblock`. I left `Accept`/`Reject` unchanged because their inner Follow is the original follow object being accepted/rejected, and those handlers resolve by follow id rather than authorizing a remote actor side effect from `object.actor`.

app/api/users/[username]/inbox/route.test.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.test.ts:226]: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. The merged guard tests cover top-level signing-key/activity-actor mismatch, path-based keys, and fragment normalization; the per-user route tests cover Block and Undo Block dispatch. This commit adds forged nested `Undo Follow` and `Undo Block` regression cases. I did not add per-route top-level Block/Undo mismatch tests because top-level actor binding is enforced centrally in `ActivityPubVerifySenderGuard`, not in the route mock.

lib/services/guards/types.ts
llun: Addressed in c88b57467f0e45e51ecdb3a4377b4c6e15f9c662. The required `verifiedSenderActorId` context was removed in favor of guard-provided `activityBody`; the shared inbox now uses the same guard-level binding before queueing. That removes the exposed-but-unused context foot-gun.

llun:
@codex review

app/api/users/[username]/inbox/route.ts
```diff
@@ -281,6 +301,15 @@ export const POST = traceApiRoute(
...
                 const undoBlock = Block.safeParse(undoObject)
                 if (undoBlock.success) {
+                  if (!actorIdsMatch(activity.actor, undoBlock.data.actor)) {
+                    return apiResponse({
+                      req,
+                      allowedMethods: CORS_HEADERS,
+                      data: ERROR_403,
+                      responseStatusCode: 403
+                    })
+                  }
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/inbox/route.ts:309]: Addressed in 19ca959e368ddbb70d10532e62f24b3b4c736ff2. `applyRemoteUnblock` now uses `normalizeActorId` for actor and target actor comparisons on both reference and full `Undo Block` objects, and it keeps `deleteBlockByUri` deletion keyed to the stored block actor id to avoid mismatches with persisted rows. Added direct unit coverage for normalized full-object unblocks, normalized URI unblocks, and mismatched actors.

lib/actions/applyRemoteUnblock.ts
```diff
@@ -17,37 +28,48 @@ export const applyRemoteUnblock = async ({
...
+      actorId: block.actorId,
       uri: object.id
     })
   }
 
   return database.deleteBlock({
-    actorId,
-    targetActorId
+    actorId: normalizedActorId,
+    targetActorId: normalizedTargetActorId
```
llun: **Normalized fallback can fail to delete blocks stored with un-normalized actor IDs.** (Piling on the codex-connector comment on the line above — adding a concrete repro.)

When `getBlockByUri` returns `null` (a legitimate situation if the remote never sent the original Block via this server, or the row was stored under a slightly different URI), this fallback runs with `normalizedActorId` / `normalizedTargetActorId`. But existing block rows were inserted before this normalization existed — they may persist host casing, trailing fragments, or path differences. The DB lookup is exact, so:

- Row stored with `actorId: "https://Remote.Test/users/alice"` (legacy upper-case host, or actor came in via a redirect that preserved case).
- Undo arrives, `normalizedActorId` becomes `"https://remote.test/users/alice"`.
- `deleteBlock({ actorId, targetActorId })` matches nothing → returns falsy → route returns 404, block stays.

Before this PR, `deleteBlock({ actorId, targetActorId })` was called with the raw `actorId` (which matched whatever was stored at write time). The new normalization in the read path needs to be matched by either (a) normalizing on write, (b) backfilling existing rows, or (c) widening the lookup (e.g., `findBlock` that compares normalized form server-side).

At minimum, please confirm that all existing block rows are guaranteed to already be normalized (e.g., normalization at insert in `applyRemoteBlock` or a migration). If not, this introduces a "stuck block" regression.

---

app/api/users/[username]/inbox/route.ts
```diff
@@ -70,6 +71,16 @@ const Activity = z.union([
...
 
+const actorIdsMatch = (firstActorId: string, secondActorId: string) => {
+  const normalizedFirstActorId = normalizeActorId(firstActorId)
+  const normalizedSecondActorId = normalizeActorId(secondActorId)
+
+  return (
+    Boolean(normalizedFirstActorId) &&
+    normalizedFirstActorId === normalizedSecondActorId
+  )
+}
```
Comment (llun) [inline app/api/users/[username]/inbox/route.ts:82]: **`?? (await req.json())` fallback in the handler is dead code and a future foot-gun.**

After this PR the guard always populates `context.activityBody` for valid POSTs (and rejects with 400 if JSON parsing fails or the body has no actor — see `getPostActivity` in `ActivityPubVerifyGuard.ts`). So `context.activityBody ?? (await req.json())` in the handler is unreachable in production.

Worse, the `request.clone().arrayBuffer()` in `digestMatches` only leaves the body re-readable as a side effect of `clone()` — any future change that drops the clone would make `req.json()` here throw on a consumed stream and silently hit the catch-all `catch {}` returning a misleading 400.

Two cleaner alternatives:
1. Tighten `ActivityPubVerifiedSenderHandle.context.activityBody` from `unknown | null` to `unknown` and drop the fallback in both this route and `app/api/inbox/route.ts`.
2. Or, since the guard already validated the JSON shape, pass the parsed activity (rather than the raw body) through context and skip a second `Activity.safeParse` in the handler too.

Same applies to `app/api/inbox/route.ts:24-35` (`getActivityBody`).

---

llun (commented):
Significant improvements since round 1 — the centralized guard binding via verified key owner is the right architecture, the shared inbox is now covered, the inner-object actor check is in place for `Undo Follow` / `Undo Block`, and the new normalization helper + `applyRemoteUnblock` tests are good additions. Two remaining items:

1. **Possible regression in `applyRemoteUnblock` fallback** when stored block rows weren't written with normalized actor IDs — see inline comment. Worth confirming `applyRemoteBlock` (or a migration) guarantees stored rows are already normalized before relying on `deleteBlock({ actorId: normalizedActorId, ... })` for the fallback path.

2. **Dead-code fallback in the inbox handlers** (not anchorable to the diff — the line is unchanged):
   - `app/api/users/[username]/inbox/route.ts:119` — `Activity.safeParse(context.activityBody ?? (await req.json()))`.
   - `app/api/inbox/route.ts:40` via `getActivityBody` — same `activityBody ?? request.json()` pattern.

   The guard now always populates `activityBody` for valid POSTs (rejecting with 400 otherwise via `getPostActivity`), so the `??` branch is unreachable in production. It's a foot-gun if a future change drops the `request.clone()` in `digestMatches` — `req.json()` would throw on the consumed stream and the catch-all `try/catch` in the route handler would return a misleading 400. Suggest tightening `ActivityPubVerifiedSenderHandle.context.activityBody` from `unknown | null` to `unknown` and removing the fallbacks; or going further and passing the already-validated activity object through context so handlers can skip the second `Activity.safeParse`.

Nice work consolidating the binding in the guard — that's a much cleaner fix than the original route-level check.

---

lib/actions/applyRemoteUnblock.ts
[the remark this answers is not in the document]
llun: Fixed in 9c70effa. The fallback now preserves raw legacy IDs by trying the raw Undo Block object actor/target pair first, then the raw request actor/target pair, and only then the normalized pair. URI deletes still use the stored block actor ID. Added regression coverage for raw legacy rows and normalized fallback.

llun: Fixed in 9c70effa. I did not add a migration because the unblock fallback now widens the exact-match attempts: raw object actor/target, raw request actor/target, then normalized actor/target. That covers legacy un-normalized rows while keeping the normalized path for new/canonical IDs. Added tests for both the raw legacy case and the normalized fallback case.

app/api/users/[username]/inbox/route.ts
Comment (llun) [reply app/api/users/[username]/inbox/route.ts:82]: Fixed in 9c70effa. Tightened ActivityPubVerifiedSenderHandle.context.activityBody to unknown and removed the req.json() fallbacks in both inbox handlers, including the shared inbox getActivityBody helper. The handlers now only use the guard-parsed body.

llun (commented):
Round 3 — both follow-ups from the previous round are addressed cleanly:

1. **`applyRemoteUnblock` fallback regression** — now tries raw `object.actor`/object-target, then raw request `actorId`/`targetActorId`, then the normalized pair via `deleteBlockByActorPairs` (with dedup via `uniqueActorPairs`). This handles legacy un-normalized rows without needing a migration. Test coverage at `applyRemoteUnblock.test.ts` exercises both the raw-legacy hit and the normalized fallback path. The 3-pair search is safe because the `actorIdsMatch` precondition on line 78-83 guarantees all three pairs refer to the same logical entity, so picking whichever matches the stored row first can't delete an unintended block.

2. **Dead-code `?? req.json()` fallback** — `ActivityPubVerifiedSenderHandle.context.activityBody` tightened from `unknown | null` to `unknown`, the fallback removed in both `app/api/users/[username]/inbox/route.ts:118` and `app/api/inbox/route.ts:27`, and `getActivityBody` deleted. Handlers now consume only the guard-parsed body, matching the guard's contract.

Nothing further required from me on the security side. Optional nit: `uniqueActorPairs` / `deleteBlockByActorPairs` are only used in one place — could be inlined as a single loop with a `Set` literal — but the extraction is fine for readability.

LGTM. Up to you whether to merge or wait on `yarn test` confirmation in clean Node 24 (the PR description still notes the original `yarn test` failed on an "unrelated" `Array.fromAsync` issue).

---

llun:
## Code Review

Overall the direction is correct — adding inner-object actor binding for `Undo` activities is the right fix. A few issues need to be addressed:

---

### 1. `Undo Like` is missing the inner-actor binding check (security gap)

`Undo Follow` and `Undo Block` both got the `actorIdsMatch` guard, but `Undo Like` did not. The `Like` type has an `actor` field (`actor: z.string()`), so the same attack vector applies: an attacker signed as `alice` can send:

```json
{ "type": "Undo", "actor": "alice", "object": { "type": "Like", "actor": "bob", "object": "..." } }
```

The guard passes (alice's HTTP signature is valid), and the like is deleted using `activity.actor` (`alice`), but the `undoLike.data.actor` (`bob`) is never checked. The fix in `route.ts` around the `undoLike` branch (lines 282–298) should add the same check before calling `database.deleteLike`:

```ts
if (!actorIdsMatch(activity.actor, undoLike.data.actor)) {
  return apiResponse({ req, allowedMethods: CORS_HEADERS, data: ERROR_403, responseStatusCode: 403 })
}
```

---

### 2. `actorIdsMatch` is duplicated across two files

The helper is defined identically in both:
- `app/api/users/[username]/inbox/route.ts` (lines 74–82)
- `lib/actions/applyRemoteUnblock.ts`

Both files already import `normalizeActorId` from `@/lib/utils/activitypub`. `actorIdsMatch` should be extracted and exported from that same module to avoid divergence.

---

### 3. No test for `Undo Like` with mismatched inner actor

Tests were added for `Undo Follow` and `Undo Block` with a mismatched inner actor, but there is no equivalent test case for `Undo Like`. Once the fix in point 1 is applied, a corresponding regression test should be added alongside the other two new tests in `app/api/users/[username]/inbox/route.test.ts`.

---

### 4. PR description describes `verifiedSenderActorId` that was not implemented

The description says:
> Derive a normalized `verifiedSenderActorId` from the verified HTTP-signature `keyId` and pass it through the guard context … Update the `ActivityPubVerifiedSenderHandle` type to include `verifiedSenderActorId`

None of that is in the actual diff. The type change in `lib/services/guards/types.ts` is only `unknown | null → unknown` (removing a redundant `| null`). The guard already enforces the outer `keyId owner === activity.actor` binding (lines 189–196 of `ActivityPubVerifyGuard.ts`) so no new field was needed, but the description is misleading and should be corrected to reflect what was actually done.

---

PR #672, opened 2026-05-12
fix: enforce ActivityPub sender and visibility checks

llun:
## Summary
- Carry the ActivityPub inbox guard's verified sender actor IRI into queued jobs and enforce it for Create Note, Create Question, and poll-vote handling.
- Reject spoofed Create payloads whose attributedTo or inner object actor does not match the verified sender.
- Apply timeline-style visibility checks to status-adjacent routes and public ActivityPub outbox pages.

## Tests
- yarn run prettier --write .
- yarn lint
- yarn build
- yarn test

## Notes
- No migrations or config changes.

### DIFF
--- a/app/api/inbox/getJobMessage.test.ts
+++ b/app/api/inbox/getJobMessage.test.ts
@@ -1,8 +1,26 @@
+import { CREATE_NOTE_JOB_NAME } from '@/lib/jobs/names'
+
 import { getJobMessage } from './getJobMessage'
 
 const verifiedSenderActorId = 'https://remote.test/users/alice'
 
 describe('getJobMessage', () => {
+  it('rejects Create Note activities when the verified sender actor id is missing', () => {
+    const result = getJobMessage({
+      id: 'https://remote.test/activities/create-unverified',
+      type: 'Create',
+      actor: verifiedSenderActorId,
+      object: {
+        id: 'https://remote.test/users/alice/statuses/1',
+        type: 'Note',
+        attributedTo: verifiedSenderActorId,
+        content: 'Unverified sender'
+      }
+    } as never)
+
+    expect(result).toBeNull()
+  })
+
   it('rejects Create Note activities without object actor attribution when the sender is verified', () => {
     const result = getJobMessage(
       {
@@ -19,6 +37,54 @@ describe('getJobMessage', () => {
     )
 
     expect(result).toBeNull()
+  })
+
+  it('rejects Create Note activities when any nested attribution differs from the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-mixed-attribution',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          attributedTo: [
+            { id: `${verifiedSenderActorId}#main-key` },
+            [{ id: 'https://remote.test/users/mallory' }]
+          ],
+          content: 'Mixed attribution'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it('accepts Create Note activities only when every object actor id matches the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-matching-attribution',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          attributedTo: [
+            verifiedSenderActorId,
+            { id: `${verifiedSenderActorId}#main-key` }
+          ],
+          actor: { id: verifiedSenderActorId },
+          content: 'Matching attribution'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toMatchObject({
+      name: CREATE_NOTE_JOB_NAME,
+      verifiedSenderActorId
+    })
   })
 
   it.each([
--- a/app/api/inbox/getJobMessage.test.ts
+++ b/app/api/inbox/getJobMessage.test.ts
@@ -115,6 +115,31 @@ describe('getJobMessage', () => {
     })
   })
 
+  it('accepts Create Note activities when the object actor is a link object', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-link-actor',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          attributedTo: {
+            type: 'Link',
+            href: verifiedSenderActorId
+          },
+          content: 'Actor link object'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toMatchObject({
+      name: CREATE_NOTE_JOB_NAME,
+      verifiedSenderActorId
+    })
+  })
+
   it.each([
     ['Update', 'Update', 'https://remote.test/users/alice#main-key'],
     ['Announce', 'Announce', 'https://remote.test/users/alice#main-key'],
@@ -190,6 +215,39 @@ describe('getJobMessage', () => {
         object: 'https://remote.test/users/alice/statuses/1'
       } as never,
       ''
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it('rejects Delete activities when the activity actor differs from the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/delete-spoofed',
+        type: 'Delete',
+        actor: 'https://remote.test/users/mallory',
+        object: 'https://remote.test/users/alice/statuses/1'
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it('rejects Undo Announce activities when the activity actor differs from the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/undo-spoofed',
+        type: 'Undo',
+        actor: 'https://remote.test/users/mallory',
+        object: {
+          id: 'https://remote.test/users/alice/statuses/boost-1',
+          type: 'Announce',
+          actor: verifiedSenderActorId,
+          object: 'https://remote.test/users/alice/statuses/1'
+        }
+      } as never,
+      verifiedSenderActorId
     )
 
     expect(result).toBeNull()
--- a/app/api/inbox/getJobMessage.test.ts
+++ b/app/api/inbox/getJobMessage.test.ts
@@ -5,18 +5,21 @@ import { getJobMessage } from './getJobMessage'
 const verifiedSenderActorId = 'https://remote.test/users/alice'
 
 describe('getJobMessage', () => {
-  it('rejects Create Note activities when the verified sender actor id is missing', () => {
-    const result = getJobMessage({
-      id: 'https://remote.test/activities/create-unverified',
-      type: 'Create',
-      actor: verifiedSenderActorId,
-      object: {
-        id: 'https://remote.test/users/alice/statuses/1',
-        type: 'Note',
-        attributedTo: verifiedSenderActorId,
-        content: 'Unverified sender'
-      }
-    } as never)
+  it('rejects Create Note activities when the verified sender actor id is invalid', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-unverified',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          attributedTo: verifiedSenderActorId,
+          content: 'Unverified sender'
+        }
+      } as never,
+      ''
+    )
 
     expect(result).toBeNull()
   })
@@ -76,6 +79,31 @@ describe('getJobMessage', () => {
           ],
           actor: { id: verifiedSenderActorId },
           content: 'Matching attribution'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toMatchObject({
+      name: CREATE_NOTE_JOB_NAME,
+      verifiedSenderActorId
+    })
+  })
+
+  it('accepts Create Note activities when the inline actor object id matches the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-inline-actor',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          attributedTo: {
+            id: verifiedSenderActorId,
+            url: 'https://remote.test/@alice'
+          },
+          content: 'Inline actor object'
         }
       } as never,
       verifiedSenderActorId
@@ -148,6 +176,20 @@ describe('getJobMessage', () => {
         object: 'https://remote.test/users/alice/statuses/1'
       } as never,
       verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it('rejects Announce activities when the verified sender actor id is invalid', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/announce-unverified',
+        type: 'Announce',
+        actor: verifiedSenderActorId,
+        object: 'https://remote.test/users/alice/statuses/1'
+      } as never,
+      ''
     )
 
     expect(result).toBeNull()
--- a/app/api/inbox/getJobMessage.test.ts
+++ b/app/api/inbox/getJobMessage.test.ts
@@ -1,0 +1,89 @@
+import { getJobMessage } from './getJobMessage'
+
+const verifiedSenderActorId = 'https://remote.test/users/alice'
+
+describe('getJobMessage', () => {
+  it('rejects Create Note activities without object actor attribution when the sender is verified', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/create-no-attribution',
+        type: 'Create',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/alice/statuses/1',
+          type: 'Note',
+          content: 'Missing attribution'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it.each([
+    ['Update', 'Update', 'https://remote.test/users/alice#main-key'],
+    ['Announce', 'Announce', 'https://remote.test/users/alice#main-key'],
+    ['Delete', 'Delete', 'https://remote.test/users/alice#main-key']
+  ])(
+    'attaches the normalized verified sender actor id to %s job messages',
+    (_label, type, senderActorId) => {
+      const object =
+        type === 'Update'
+          ? {
+              id: 'https://remote.test/users/alice/statuses/1',
+              type: 'Note',
+              attributedTo: verifiedSenderActorId,
+              content: 'Updated content'
+            }
+          : 'https://remote.test/users/alice/statuses/1'
+
+      const result = getJobMessage(
+        {
+          id: `https://remote.test/activities/${type.toLowerCase()}-1`,
+          type,
+          actor: verifiedSenderActorId,
+          object
+        } as never,
+        senderActorId
+      )
+
+      expect(result).toMatchObject({
+        verifiedSenderActorId: verifiedSenderActorId.toLowerCase()
+      })
+    }
+  )
+
+  it('rejects Update Note activities when object attribution differs from the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/update-spoofed',
+        type: 'Update',
+        actor: verifiedSenderActorId,
+        object: {
+          id: 'https://remote.test/users/mallory/statuses/1',
+          type: 'Note',
+          attributedTo: 'https://remote.test/users/mallory',
+          content: 'Spoofed content'
+        }
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+
+  it('rejects Announce activities when the activity actor differs from the verified sender', () => {
+    const result = getJobMessage(
+      {
+        id: 'https://remote.test/activities/announce-spoofed',
+        type: 'Announce',
+        actor: 'https://remote.test/users/mallory',
+        object: 'https://remote.test/users/alice/statuses/1'
+      } as never,
+      verifiedSenderActorId
+    )
+
+    expect(result).toBeNull()
+  })
+})
--- a/app/api/inbox/getJobMessage.ts
+++ b/app/api/inbox/getJobMessage.ts
@@ -79,7 +79,10 @@ const extractActivityPubIds = (value: unknown): string[] => {
   if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
   if (!isRecord(value)) return []
 
-  return typeof value.id === 'string' ? [value.id] : []
+  if (typeof value.id === 'string') return [value.id]
+  if (typeof value.href === 'string') return [value.href]
+  if (typeof value.url === 'string') return [value.url]
+  return []
 }
 
 const createObjectActorMismatch = (
--- a/app/api/inbox/getJobMessage.ts
+++ b/app/api/inbox/getJobMessage.ts
@@ -10,6 +10,7 @@ import {
   UPDATE_NOTE_JOB_NAME,
   UPDATE_POLL_JOB_NAME
 } from '@/lib/jobs/names'
+import type { JobMessage } from '@/lib/services/queue/type'
 import { ENTITY_TYPE_NOTE, ENTITY_TYPE_QUESTION } from '@/lib/types/activitypub'
 import {
   AnnounceAction,
@@ -18,7 +19,9 @@ import {
   UndoAction,
   UpdateAction
 } from '@/lib/types/activitypub/activities'
+import { extractActivityPubId, normalizeActorId } from '@/lib/utils/activitypub'
 import { getHashFromString } from '@/lib/utils/getHashFromString'
+import { isRecord } from '@/lib/utils/typeGuards'
 
 const ENTITY_TYPE_IMAGE = 'Image'
 const ENTITY_TYPE_PAGE = 'Page'
@@ -33,7 +36,43 @@ const NOTE_TYPES = [
   ENTITY_TYPE_VIDEO
 ]
 
-export const getJobMessage = (activity: StatusActivity) => {
+const createJobMessage = ({
+  data,
+  id,
+  name,
+  verifiedSenderActorId
+}: JobMessage) => ({
+  id,
+  name,
+  data,
+  ...(verifiedSenderActorId ? { verifiedSenderActorId } : {})
+})
+
+const createObjectActorMismatch = (
+  object: unknown,
+  verifiedSenderActorId?: string
+) => {
+  if (!verifiedSenderActorId || !isRecord(object)) return false
+
+  const normalizedVerifiedSenderActorId = normalizeActorId(
+    verifiedSenderActorId
+  )
+  if (!normalizedVerifiedSenderActorId) return true
+
+  const objectActorIds = [
+    extractActivityPubId(object.attributedTo),
+    extractActivityPubId(object.actor)
+  ].filter((actorId): actorId is string => Boolean(actorId))
+
+  return objectActorIds.some(
+    (actorId) => normalizeActorId(actorId) !== normalizedVerifiedSenderActorId
+  )
+}
+
+export const getJobMessage = (
+  activity: StatusActivity,
+  verifiedSenderActorId?: string
+) => {
   const deduplicationId = getHashFromString(activity.id)
 
   if (activity.type === CreateAction) {
@@ -42,6 +81,10 @@ export const getJobMessage = (activity: StatusActivity) => {
       activity.object !== null &&
       NOTE_TYPES.includes(activity.object.type)
     ) {
+      if (createObjectActorMismatch(activity.object, verifiedSenderActorId)) {
+        return null
+      }
+
       if (
         activity.object.type === ENTITY_TYPE_NOTE &&
         activity.object.inReplyTo &&
@@ -49,18 +92,20 @@ export const getJobMessage = (activity: StatusActivity) => {
         activity.object.name &&
         !activity.object.content
       ) {
-        return {
+        return createJobMessage({
           id: deduplicationId,
           name: CREATE_POLL_VOTE_JOB_NAME,
-          data: activity.object
-        }
+          data: activity.object,
+          verifiedSenderActorId
+        })
       }
 
-      return {
+      return createJobMessage({
         id: deduplicationId,
         name: CREATE_NOTE_JOB_NAME,
-        data: activity.object
-      }
+        data: activity.object,
+        verifiedSenderActorId
+      })
     }
 
     if (
@@ -68,11 +113,16 @@ export const getJobMessage = (activity: StatusActivity) => {
       activity.object !== null &&
       activity.object.type === ENTITY_TYPE_QUESTION
     ) {
-      return {
+      if (createObjectActorMismatch(activity.object, verifiedSenderActorId)) {
+        return null
+      }
+
+      return createJobMessage({
         id: deduplicationId,
         name: CREATE_POLL_JOB_NAME,
-        data: activity.object
-      }
+        data: activity.object,
+        verifiedSenderActorId
+      })
     }
   }
 
--- a/app/api/inbox/getJobMessage.ts
+++ b/app/api/inbox/getJobMessage.ts
@@ -58,10 +58,8 @@ const createJobMessage = ({
 
 const activityActorMismatch = (
   activity: StatusActivity,
-  verifiedSenderActorId?: string
+  verifiedSenderActorId: string
 ) => {
-  if (!verifiedSenderActorId) return false
-
   const normalizedVerifiedSenderActorId = normalizeActorId(
     verifiedSenderActorId
   )
@@ -81,12 +79,12 @@ const extractActivityPubIds = (value: unknown): string[] => {
   if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
   if (!isRecord(value)) return []
 
-  return [value.id, value.href, value.url].flatMap(extractActivityPubIds)
+  return typeof value.id === 'string' ? [value.id] : []
 }
 
 const createObjectActorMismatch = (
   object: unknown,
-  verifiedSenderActorId?: string
+  verifiedSenderActorId: string
 ) => {
   if (!isRecord(object)) return false
 
@@ -109,7 +107,7 @@ const createObjectActorMismatch = (
 
 export const getJobMessage = (
   activity: StatusActivity,
-  verifiedSenderActorId?: string
+  verifiedSenderActorId: string
 ) => {
   const deduplicationId = getHashFromString(activity.id)
 
--- a/app/api/inbox/getJobMessage.ts
+++ b/app/api/inbox/getJobMessage.ts
@@ -76,11 +76,19 @@ const activityActorMismatch = (
   )
 }
 
+const extractActivityPubIds = (value: unknown): string[] => {
+  if (typeof value === 'string') return [value]
+  if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
+  if (!isRecord(value)) return []
+
+  return [value.id, value.href, value.url].flatMap(extractActivityPubIds)
+}
+
 const createObjectActorMismatch = (
   object: unknown,
   verifiedSenderActorId?: string
 ) => {
-  if (!verifiedSenderActorId || !isRecord(object)) return false
+  if (!isRecord(object)) return false
 
   const normalizedVerifiedSenderActorId = normalizeActorId(
     verifiedSenderActorId
@@ -88,14 +96,14 @@ const createObjectActorMismatch = (
   if (!normalizedVerifiedSenderActorId) return true
 
   const objectActorIds = [
-    extractActivityPubId(object.attributedTo),
-    extractActivityPubId(object.actor)
-  ].filter((actorId): actorId is string => Boolean(actorId))
+    ...extractActivityPubIds(object.attributedTo),
+    ...extractActivityPubIds(object.actor)
+  ]
 
   if (objectActorIds.length === 0) return true
 
-  return objectActorIds.some(
-    (actorId) => normalizeActorId(actorId) !== normalizedVerifiedSenderActorId
+  return !objectActorIds.every(
+    (actorId) => normalizeActorId(actorId) === normalizedVerifiedSenderActorId
   )
 }
 
--- a/app/api/inbox/getJobMessage.ts
+++ b/app/api/inbox/getJobMessage.ts
@@ -41,12 +41,40 @@ const createJobMessage = ({
   id,
   name,
   verifiedSenderActorId
-}: JobMessage) => ({
-  id,
-  name,
-  data,
-  ...(verifiedSenderActorId ? { verifiedSenderActorId } : {})
-})
+}: JobMessage) => {
+  const normalizedVerifiedSenderActorId = normalizeActorId(
+    verifiedSenderActorId
+  )
+
+  return {
+    id,
+    name,
+    data,
+    ...(normalizedVerifiedSenderActorId
+      ? { verifiedSenderActorId: normalizedVerifiedSenderActorId }
+      : {})
+  }
+}
+
+const activityActorMismatch = (
+  activity: StatusActivity,
+  verifiedSenderActorId?: string
+) => {
+  if (!verifiedSenderActorId) return false
+
+  const normalizedVerifiedSenderActorId = normalizeActorId(
+    verifiedSenderActorId
+  )
+  const normalizedActivityActorId = normalizeActorId(
+    extractActivityPubId(activity.actor)
+  )
+
+  return (
+    !normalizedVerifiedSenderActorId ||
+    !normalizedActivityActorId ||
+    normalizedActivityActorId !== normalizedVerifiedSenderActorId
+  )
+}
 
 const createObjectActorMismatch = (
   object: unknown,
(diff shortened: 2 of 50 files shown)


llun:
@codex review

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,43 @@ const NOTE_TYPES = [
...
+    verifiedSenderActorId
+  )
+  if (!normalizedVerifiedSenderActorId) return true
+
+  const objectActorIds = [
+    extractActivityPubId(object.attributedTo),
+    extractActivityPubId(object.actor)
+  ].filter((actorId): actorId is string => Boolean(actorId))
+
+  return objectActorIds.some(
```
llun: **Bug: empty `objectActorIds` silently allows activities with no attribution.**

When an object contains neither `attributedTo` nor `actor`, `objectActorIds` will be an empty array and `Array.prototype.some()` returns `false`, meaning "no mismatch" — so the activity is allowed through without any sender verification.

For Create Note and Create Question objects, `attributedTo` is required by the ActivityPub spec; a missing attribution should be treated as a mismatch, not a pass. Consider flipping the guard:

```ts
if (objectActorIds.length === 0) return true // no actors to verify against → treat as mismatch

return objectActorIds.some(
  (actorId) => normalizeActorId(actorId) !== normalizedVerifiedSenderActorId
)
```

Note that `actorMatchesVerifiedSender` in `lib/jobs/verifiedSender.ts` handles the analogous case correctly (returns `false` when `normalizedActorId` is falsy), so this early guard is inconsistent with the job-level check.

---

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -49,6 +57,21 @@ jest.mock('@/lib/config', () => ({
...
 
+type StatusRouteHandler = (
+  req: NextRequest,
+  context: { params: Promise<{ id: string }> }
+) => Promise<Response> | Response
+
+const inaccessibleStatusRouteCases: Array<
+  [string, 'GET' | 'POST', StatusRouteHandler]
+> = [
+  ['source', 'GET', getStatusSource],
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:81]: **Missing test coverage for `unbookmark` and `unmute`.**

Both routes were updated to use `getReadableStatus` in this PR, but they are absent from `inaccessibleStatusRouteCases` and have no dedicated tests. Please add them:

```ts
const inaccessibleStatusRouteCases: Array<[string, 'GET' | 'POST', StatusRouteHandler]> = [
  ['source',        'GET',  getStatusSource],
  ['bookmark',      'POST', bookmarkStatus],
  ['unbookmark',    'POST', unbookmarkStatus],   // missing
  ['mute',          'POST', muteStatus],
  ['unmute',        'POST', unmuteStatus],        // missing
  ['favourited_by', 'GET',  getStatusFavouritedBy],
  ['reblogged_by',  'GET',  getStatusRebloggedBy]
]
```

---

app/api/users/[username]/outbox/route.ts
```diff
@@ -31,7 +32,12 @@ export const GET = traceApiRoute(
         })
       }
```
Comment (llun) [inline app/api/users/[username]/outbox/route.ts:70]: **Double-filtering — clarify intent or consolidate.**

`publicOnly: true` already filters at the SQL level to statuses whose recipients include the ActivityStreams public IRI. The subsequent `.filter(isStatusPubliclyReadable)` is therefore redundant for plain Notes/Questions.

The only case where the second filter adds value is Announce statuses: `publicOnly: true` checks whether the *announce* has public recipients, but `isStatusPubliclyReadable` additionally recurses into the boosted status to confirm it is also public. If that edge-case defence is intentional, please add a short comment explaining it so the redundancy is not removed by a future refactor. If it is not needed, remove one of the two filters.

---

llun (commented):
Overall the direction is solid — threading `verifiedSenderActorId` through the queue and applying visibility guards to status-adjacent routes are both good improvements. A few things need to be fixed before merging:

**High — logic bug**
- `createObjectActorMismatch` returns `false` (no mismatch) when the object has neither `attributedTo` nor `actor` (empty array → `some()` is always false). An attributionless Create Note/Question should be rejected, not allowed through. See inline comment on `app/api/inbox/getJobMessage.ts`.

**High — incomplete coverage**
- `favourite/route.ts` and `unfavourite/route.ts` still call `database.getStatus()` directly. A user can favourite a followers-only post they cannot read. These two routes need the same `getReadableStatus` treatment applied to the rest of the adjacent routes.

**Medium — missing tests**
- `unbookmark` and `unmute` were updated to use `getReadableStatus` but are absent from `inaccessibleStatusRouteCases` and have no dedicated visibility tests. See inline comment on `app/api/v1/statuses/[id]/route.test.ts`.

**Low — outbox double-filtering**
- `publicOnly: true` (SQL) + `.filter(isStatusPubliclyReadable)` (in-memory) are redundant for most statuses. If the second filter is intentionally kept as defence for Announce recursion, add a comment; otherwise consolidate. See inline comment on `app/api/users/[username]/outbox/route.ts`.

**Nit — 404 for spoofed activities**
- When `createObjectActorMismatch` fires, the inbox route returns 404. This conflates "unrecognised activity type" with "security rejection". A 400 or 403 would be more semantically appropriate and avoid misleading legitimate clients.

---

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,43 @@ const NOTE_TYPES = [
...
+    (actorId) => normalizeActorId(actorId) !== normalizedVerifiedSenderActorId
+  )
+}
+
+export const getJobMessage = (
+  activity: StatusActivity,
+  verifiedSenderActorId?: string
+) => {
```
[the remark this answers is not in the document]
llun: Fixed. Update, Announce, and Delete now build jobs through createJobMessage and carry a normalized verifiedSenderActorId; Delete processing also filters status deletion by actorId.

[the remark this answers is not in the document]
llun: Fixed. verifiedSenderActorId is normalized and propagated beyond Create, Update object attribution is checked before enqueueing, Announce/Delete activity actors are checked, and Delete status removal is scoped by actorId.

app/api/inbox/getJobMessage.ts
llun: Fixed. When a verified sender is present, Create/Update objects with neither attributedTo nor actor are now treated as a mismatch and rejected before enqueueing.

app/api/v1/statuses/[id]/route.test.ts
Comment (llun) [reply app/api/v1/statuses/[id]/route.test.ts:81]: Fixed. The inaccessible status route matrix now includes unbookmark and unmute.

app/api/users/[username]/outbox/route.ts
Comment (llun) [reply app/api/users/[username]/outbox/route.ts:70]: Fixed. Added a comment documenting that the second outbox filter is intentional because Announce needs the boosted original status to be publicly readable too.

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,43 @@ const NOTE_TYPES = [
...
+  const normalizedVerifiedSenderActorId = normalizeActorId(
+    verifiedSenderActorId
+  )
+  if (!normalizedVerifiedSenderActorId) return true
+
+  const objectActorIds = [
+    extractActivityPubId(object.attributedTo),
+    extractActivityPubId(object.actor)
+  ].filter((actorId): actorId is string => Boolean(actorId))
+
```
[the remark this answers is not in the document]
llun: Fixed. Missing object actor attribution now fails closed as a mismatch when verifiedSenderActorId is present.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -49,6 +57,21 @@ jest.mock('@/lib/config', () => ({
...
+  ['bookmark', 'POST', bookmarkStatus],
+  ['mute', 'POST', muteStatus],
+  ['favourited_by', 'GET', getStatusFavouritedBy],
+  ['reblogged_by', 'GET', getStatusRebloggedBy]
+]
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/route.test.ts:94]: Fixed. favourite/unfavourite/reblog/unreblog/pin/unpin now use getReadableStatus before acting, and the inaccessible status route matrix covers them.

app/api/users/[username]/outbox/route.ts
```diff
@@ -31,7 +32,12 @@ export const GET = traceApiRoute(
...
 
-      const statuses = await database.getActorStatuses({ actorId: actor.id })
+      const statuses = (
+        await database.getActorStatuses({
+          actorId: actor.id,
+          publicOnly: true
+        })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/outbox/route.ts]: Fixed. The outbox collection totalItems is now derived from the same public-readable status set used for page items.

llun:
@codex review

app/api/users/[username]/outbox/route.ts
```diff
@@ -18,20 +33,30 @@ export const GET = traceApiRoute(
       const pageParam = url.searchParams.get('page')
       if (!pageParam) {
         const outboxId = getLocalActorOutboxId(actor.id)
+        const statuses = await getPubliclyReadableActorStatuses(
+          database,
+          actor.id,
+          actor.statusCount
+        )
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/outbox/route.ts]: Fixed in 7e6532dc. The collection metadata path now uses getActorStatusesCount({ publicOnly: true }) so totalItems is computed with a SQL count instead of hydrating all statuses.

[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/outbox/route.ts]: Fixed in 7e6532dc. The outbox root no longer calls getActorStatuses just to derive totalItems; it uses the public-only count query directly.

lib/services/statusRouteAccess.ts
```diff
@@ -0,0 +1,52 @@
...
+}) =>
+  (
+    await Promise.all(
+      statuses.map(async (status) =>
+        (await canActorReadStatus({ database, status, currentActor }))
+          ? status
+          : null
+      )
+    )
+  ).filter((status): status is Status => status !== null)
```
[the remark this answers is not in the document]
llun: Fixed in 7e6532dc. Status reply queries can now take visibleToActorId to prefilter to statuses potentially visible to that actor in SQL, and filterReadableStatuses skips async access checks for already-public statuses.

llun:
@codex review

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,73 @@ const NOTE_TYPES = [
...
+  if (objectActorIds.length === 0) return true
+
+  return objectActorIds.some(
+    (actorId) => normalizeActorId(actorId) !== normalizedVerifiedSenderActorId
+  )
+}
```
llun, line 109: Agreeing with Gemini's open thread here. `extractActivityPubId` is a "first-match" helper — when `attributedTo` or `actor` is an array it returns only the first element, so `[legitimate_actor, spoofed_actor]` produces `objectActorIds = [legitimate_actor]` and the spoofed entry is never examined. The immediate practical risk is low because (a) the job-level schemas resolve `attributedTo` to the first element when storing to the DB, and (b) ActivityPub payloads rarely use array attribution in practice. But this should still be fixed for defence-in-depth.

Gemini's suggested `getActorIds` helper (extract all IDs recursively) is the right approach. The check should confirm that **every** actor ID present in `attributedTo`/`actor` matches the verified sender, not just the first one. Alternatively, the function could flat-map all IDs and require `objectActorIds.every(id => normalizeActorId(id) === normalizedVerifiedSenderActorId)` once the empty-array guard is in place.

---

app/api/v1/statuses/[id]/unreblog/route.ts
```diff
@@ -35,6 +36,19 @@ export const POST = traceApiRoute(
       })
 
     const statusId = idToUrl(encodedStatusId)
+    const status = await getReadableStatus({
+      database,
+      statusId,
+      currentActor,
+      withReplies: false
+    })
+    if (!status)
```
Comment (llun) [inline app/api/v1/statuses/[id]/unreblog/route.ts]: Agreeing with ChatGPT Codex's open thread. The `statusId` passed here is the **original post's ID** (Mastodon API convention for unreblog). `getReadableStatus` calls `canActorReadStatus`, which checks the actor's current ability to read that post. This creates two breakage scenarios:

1. **Visibility downgrade**: the original author changes the post from public to followers-only after the boost. The current actor (not a follower) can no longer unboost it.
2. **Remote account/status deleted**: the original post is purged from the local DB. `getReadableStatus` returns null, and the actor is permanently stuck with a dangling boost they can't remove.

For `unreblog`, the authorization question is "does this actor own an active boost of this status?" not "can this actor currently read the original?" `userUndoAnnounce` already handles the case where no boost exists (returns null → 422). The `getReadableStatus` guard should be replaced with a plain `database.getStatus` call (or removed entirely and delegated to `userUndoAnnounce`):

```ts
// Check only that the status exists; ownership of the boost is enforced by userUndoAnnounce
const statusExists = await database.getStatus({ statusId, withReplies: false })
if (!statusExists) { /* 404 */ }
```

The same logic applies to `unfavourite`: if a status becomes inaccessible after the user has already liked it, the user should still be able to undo the like. Consider applying the same fix there too.

---

lib/jobs/deleteObjectJob.ts
```diff
@@ -24,7 +37,8 @@ export const deleteObjectJob = createJobHandle(
         const tombStone = tombStoneResult.data
         span.setAttribute('statusId', tombStone.id)
         await database.deleteStatus({
-          statusId: tombStone.id
+          statusId: tombStone.id,
+          actorId: getVerifiedSenderActorId(message.verifiedSenderActorId)
         })
         span.end()
```
llun, line 44: **New: `deleteStatus` without a `verifiedSenderActorId` is unscoped.**

`getVerifiedSenderActorId` returns `undefined` when `message.verifiedSenderActorId` is unset (e.g., messages queued before this PR or from a codepath that didn't carry the verified sender). The SQL implementation correctly adds `WHERE actorId = ?` only when `actorId` is truthy — so when it is `undefined`, the delete is scoped only by `statusId`, which is the pre-existing behaviour.

This is an acceptable backward-compatibility trade-off, but it means the `actorId` scope silently vanishes for older queued messages. Worth adding a short comment to `deleteStatus` clarifying that `actorId: undefined` intentionally falls back to an unscoped delete, so a future maintainer doesn't assume the filter is always applied.

---

llun (commented):
Good progress — all issues from the first round were addressed. Two open threads from Gemini and ChatGPT Codex still need resolution, and I found one additional item.

**Still unresolved (open threads)**

1. **`extractActivityPubId` only checks first array element** (Gemini's thread on `getJobMessage.ts` line 100) — I've added a comment supporting this. An `attributedTo: [legitimate, spoofed]` payload would pass the current check undetected. Fix: extract all IDs from the array and require every one to match the verified sender.

2. **`unreblog` blocks legitimate unboosting** (ChatGPT thread on `unreblog/route.ts` line 45) — I've added a comment supporting this. If the original post's visibility changes or the post is deleted, the actor is permanently stuck with a boost they can't remove. The authorization for unreblog should be "does this actor own an active boost?" not "can this actor currently read the original?". The same concern applies to `unfavourite`. See my inline comment for a concrete fix.

**New finding**

3. **`deleteStatus` unscoped when `verifiedSenderActorId` is absent** (`deleteObjectJob.ts` line 43) — low severity, backward-compatible edge case documented in the inline comment. No code change required, but a clarifying comment in `deleteStatus` would help future readers understand the intent.

---

app/api/inbox/getJobMessage.ts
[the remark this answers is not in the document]
llun: Fixed in 9dd3abfe. Create/Update object actor checks now recursively collect all IDs from object.attributedTo and object.actor, normalize them against the verified sender, fail closed without a valid verified sender, and require every collected actor ID to match.

app/api/v1/statuses/[id]/unreblog/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/unreblog/route.ts]: Fixed in 9dd3abfe. Unreblog now authorizes from the current actor owned Announce, including the Mastodon-style original status id path, so loss of readability on the original no longer blocks removing an owned boost. I also applied the same ownership fallback to unfavourite, gated by an existing Like.

app/api/inbox/getJobMessage.ts
llun: Fixed in 9dd3abfe. The first-match actor extraction was replaced with recursive ID collection for attributedTo and actor, and every collected ID must normalize to the verified sender.

app/api/v1/statuses/[id]/unreblog/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/unreblog/route.ts]: Fixed in 9dd3abfe. The unreblog route now looks up the current actor owned Announce by the original status id when the original is no longer readable, and unfavourite has the same ownership fallback for existing Likes.

lib/jobs/deleteObjectJob.ts
llun: Fixed in 9dd3abfe. Added a compatibility comment clarifying that an undefined verified sender intentionally preserves the legacy unscoped delete fallback for older queued messages.

llun:
@codex review

llun:
@codex review

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,81 @@ const NOTE_TYPES = [
...
+    !normalizedVerifiedSenderActorId ||
+    !normalizedActivityActorId ||
+    normalizedActivityActorId !== normalizedVerifiedSenderActorId
+  )
+}
+
+const extractActivityPubIds = (value: unknown): string[] => {
+  if (typeof value === 'string') return [value]
+  if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
+  if (!isRecord(value)) return []
```
llun, line 80: **Potential false positive: `extractActivityPubIds` extracts `href` and `url` from objects.**

`extractActivityPubIds` recurses into `value.id`, `value.href`, and `value.url` when the input is a record. This means if `attributedTo` is a full Actor object (not just a string), all three properties are extracted and every one must match the verified sender. In practice, Mastodon actors have `id: "https://mastodon.social/users/alice"` and `url: "https://mastodon.social/@alice"` — two different strings. If any remote server inlines the full actor object in `attributedTo`, the check extracts both `id` and `url`, they normalize to different IRIs, and the mismatch check rejects the activity even though it is genuine.

In ActivityPub, `id` is the canonical identifier. `href` and `url` are presentation-layer alternatives and should not participate in identity comparisons. The helper should only extract `id` from objects:

```ts
const extractActivityPubIds = (value: unknown): string[] => {
  if (typeof value === 'string') return [value]
  if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
  if (!isRecord(value)) return []
  if (typeof value.id === 'string') return [value.id]
  return []
}
```

This is consistent with how `extractActivityPubId` (the existing single-value variant) treats object inputs: it checks `id`, then falls back to `href`, then `url` — but each only as a fallback, not simultaneously. The new `extractActivityPubIds` should not combine all three.

---

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,81 @@ const NOTE_TYPES = [
...
+
+  return [value.id, value.href, value.url].flatMap(extractActivityPubIds)
+}
+
+const createObjectActorMismatch = (
+  object: unknown,
```
llun, line 89: **Behavior change: `createObjectActorMismatch` now fails closed when `verifiedSenderActorId` is absent.**

Previously, when `verifiedSenderActorId` was `undefined`, the function returned `false` (no mismatch — activity allowed through). Now `normalizeActorId(undefined)` returns `null`, the `!normalizedVerifiedSenderActorId` guard fires, and the function returns `true` (mismatch — activity rejected). Any caller that passes `undefined` as the second argument will have all Create/Update activities silently dropped.

In production this is safe because the inbox routes always supply a non-empty string from the guard. But the function signature is `getJobMessage(activity, verifiedSenderActorId?: string)` — `undefined` is a valid call. This is also inconsistent with `activityActorMismatch`, which still explicitly allows through (`return false`) when no sender is provided.

Consider one of:
- Making the parameter required: `verifiedSenderActorId: string` (breaking but honest)
- Restoring the "no sender = allow" path: `if (!normalizedVerifiedSenderActorId) return false`
- At minimum, adding a comment to `createObjectActorMismatch` clarifying that passing `undefined` intentionally fails closed, unlike the Announce/Delete path

---

app/api/v1/statuses/[id]/unreblog/route.ts
```diff
@@ -35,10 +37,45 @@ export const POST = traceApiRoute(
...
+    })
+    if (
+      readableStatus?.type !== StatusType.enum.Announce ||
+      readableStatus.actorId !== currentActor.id
+    ) {
+      const ownedAnnounce = await database.getStatus({
+        statusId,
+        withReplies: false,
+        currentActorId: currentActor.id
+      })
```
Comment (llun) [inline app/api/v1/statuses/[id]/unreblog/route.ts:54]: **Nit: intermediate `database.getStatus` call is a no-op in the fallback path.**

In the normal Mastodon API flow, the client sends the original post's ID as `:id`. `statusId` here is that original post ID. The `database.getStatus({ statusId, ... })` call will return the original Note/Poll — never an Announce — so `ownedAnnounce?.type === StatusType.enum.Announce` is always false, and execution always falls through to `database.getActorAnnounceStatus`. This adds a redundant round-trip to the DB.

The `ownedAnnounce` variable and its check can be removed without changing behaviour:

```ts
const actorAnnounce = await database.getActorAnnounceStatus({
  actorId: currentActor.id,
  statusId
})
```

---

llun (commented):
All previous round issues are fixed. Three new findings from this round:

**Medium — potential false positive in `extractActivityPubIds`**

The helper extracts `id`, `href`, and `url` from objects simultaneously and requires every extracted value to match the verified sender. An Actor object with `id: "https://example.com/users/alice"` and `url: "https://example.com/@alice"` (Mastodon's standard representation) would produce two non-equal strings, causing a mismatch and silently dropping a valid Create/Update. Only `id` should be used for identity comparison; `href` and `url` are presentation-layer alternatives. See inline comment at `getJobMessage.ts` line 82.

**Low — `createObjectActorMismatch` behavior inconsistency when sender is absent**

Passing `verifiedSenderActorId: undefined` now fails closed (returns `true`) for Create/Update objects, while `activityActorMismatch` still allows through (returns `false`) for Announce/Delete when the sender is absent. In production this does not matter because the inbox always provides a sender, but the inconsistency is a latent footgun. See inline comment at `getJobMessage.ts` line 88.

**Nit — redundant DB call in `unreblog` fallback**

`database.getStatus(statusId)` in the fallback path always fetches the original Note and never returns an Announce; the check is always false and the call can be removed. See inline comment at `unreblog/route.ts` line 55.

---

app/api/inbox/getJobMessage.ts
llun: Fixed in 594e3e64. getJobMessage, createObjectActorMismatch, and activityActorMismatch now require a verified sender string, and Announce handling fails closed when the sender value is invalid. Added coverage for the invalid-sender Announce path.

app/api/inbox/getJobMessage.ts
llun: Fixed in 594e3e64. extractActivityPubIds now treats inline actor objects by canonical id only, while arrays still recurse, so presentation url values no longer cause false mismatches. Added coverage for an inline actor object with a different url.

app/api/v1/statuses/[id]/unreblog/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/unreblog/route.ts:54]: Fixed in 594e3e64. Removed the fallback database.getStatus round trip; the fallback path now directly asks getActorAnnounceStatus for the current actor and original status.

app/api/inbox/getJobMessage.test.ts
```diff
@@ -0,0 +1,197 @@
...
+        id: 'https://remote.test/activities/announce-unverified',
+        type: 'Announce',
+        actor: verifiedSenderActorId,
+        object: 'https://remote.test/users/alice/statuses/1'
+      } as never,
+      ''
+    )
+
+    expect(result).toBeNull()
+  })
```
llun, line 221: **Missing rejection tests for `Delete` and `Undo+Announce` activity types.**

The new tests added in this commit cover Announce rejection (spoofed actor, invalid sender). The `it.each` table includes `Delete` in the *acceptance* case (normalized sender attached to job message), but there is no rejection test for Delete with a spoofed actor — and `Undo+Announce` has no test at all.

Both paths go through `activityActorMismatch`, which is the same security guard as Announce. Consider adding:

```ts
it('rejects Delete activities when the activity actor differs from the verified sender', () => {
  const result = getJobMessage(
    {
      id: 'https://remote.test/activities/delete-spoofed',
      type: 'Delete',
      actor: 'https://remote.test/users/mallory',
      object: 'https://remote.test/users/alice/statuses/1'
    } as never,
    verifiedSenderActorId
  )
  expect(result).toBeNull()
})

it('rejects Undo+Announce activities when the activity actor differs from the verified sender', () => {
  const result = getJobMessage(
    {
      id: 'https://remote.test/activities/undo-spoofed',
      type: 'Undo',
      actor: 'https://remote.test/users/mallory',
      object: {
        type: 'Announce',
        id: 'https://remote.test/users/alice/statuses/boost-1',
        actor: verifiedSenderActorId,
        object: 'https://remote.test/users/alice/statuses/1'
      }
    } as never,
    verifiedSenderActorId
  )
  expect(result).toBeNull()
})
```

---

lib/database/sql/status.ts
```diff
@@ -84,6 +85,67 @@ export const StatusSQLDatabaseMixin = (
...
+    query,
+    visibleToActorId
+  }: {
+    query: Knex.QueryBuilder
+    visibleToActorId: string
+  }) => {
+    const clientName = String(database.client.config.client)
+    const followersAudienceExpression = clientName.includes('mysql')
+      ? "followers_recipients.actorId = CONCAT(statuses.actorId, '/followers')"
+      : "followers_recipients.actorId = statuses.actorId || '/followers'"
```
llun, line 211: Agreeing with the open ChatGPT Codex thread here. A concrete fix path:

`getActorStatuses` already accepts a `followersAudience` parameter (the caller passes the actor's actual `followersUrl`) and passes it to the recipient filter correctly. `getStatusReplies` → `applyPotentiallyReadableStatusFilter` lacks that parameter and falls back to the synthetic `actorId + '/followers'` URL.

The fix is to add `followersUrl?: string` to `GetStatusRepliesParams`, thread it through `applyPotentiallyReadableStatusFilter`, and prefer it over the hard-coded suffix:

```typescript
// GetStatusRepliesParams
export type GetStatusRepliesParams = BaseStatusParams & {
  url?: string
  limit?: number
  publicOnly?: boolean
  visibleToActorId?: string | null
  followersUrl?: string          // add this
}

// applyPotentiallyReadableStatusFilter
const applyPotentiallyReadableStatusFilter = ({
  query,
  visibleToActorId,
  followersUrl            // add this
}: {
  query: Knex.QueryBuilder
  visibleToActorId: string
  followersUrl?: string   // add this
}) => {
  const clientName = String(database.client.config.client)
  // Use the actor's actual followers collection URL when available
  const defaultFollowersExpression = clientName.includes('mysql')
    ? "followers_recipients.actorId = CONCAT(statuses.actorId, '/followers')"
    : "followers_recipients.actorId = statuses.actorId || '/followers'"

  return query.where((qb) => {
    qb.whereIn('statuses.id', ...)
      ...
      .orWhereExists(function () {
        this.select(database.raw('1'))
          .from('recipients as followers_recipients')
          .whereRaw('followers_recipients.statusId = statuses.id')
          .where((inner) => {
            if (followersUrl) {
              inner.where('followers_recipients.actorId', followersUrl)
            } else {
              inner.whereRaw(defaultFollowersExpression)
            }
          })
          ...
      })
  })
}
```

The caller (`getStatusReplies`) would receive `followersUrl` from the status-context route, which already has the current actor object and can supply `currentActor.followersUrl`.

---

lib/database/sql/status.ts
```diff
@@ -851,17 +928,22 @@ export const StatusSQLDatabaseMixin = (
...
-        await deleteStatus({ statusId, trx })
+        await deleteStatus({ actorId, statusId, trx })
       })
       return
     }
 
-    const status = await trx('statuses').where('id', statusId).first()
+    const statusQuery = trx('statuses').where('id', statusId)
+    if (actorId) {
+      statusQuery.where('actorId', actorId)
```
llun: Agreeing with the open ChatGPT Codex thread on this line. The root cause is in ingestion: `createNoteJob.ts` stores `actorId: note.attributedTo` without normalization. If any remote implementation sends `attributedTo: "https://remote.test/users/alice#main-key"` in the Note body (unusual but valid per ActivityPub), the stored row has the fragment suffix, while the verified sender is normalized to `"https://remote.test/users/alice"` — the scoped delete silently no-ops.

The simplest fix is to normalize at write time in `createNoteJob.ts`:

```typescript
// Before:
actorId: note.attributedTo

// After:
actorId: normalizeActorId(note.attributedTo) ?? note.attributedTo
```

This ensures stored `actorId` values are always fragment-free, making the equality comparison in `deleteStatus` safe for any normalization variant the sender uses.

---

llun (commented):
Round 5 review on commit `e7c7f1c`. The `getJobMessage.ts` and `unreblog/route.ts` changes look correct. Three items remain:

**New finding (latest commit)**
- `getJobMessage.test.ts`: The test suite now covers Announce rejection but has no equivalent rejection tests for `Delete` and `Undo+Announce`. Both go through `activityActorMismatch` — same logic, same risk surface — and an inline comment above shows the two missing cases.

**Three open ChatGPT Codex threads (still unresolved)**
1. **`lib/database/sql/status.ts` line 120** — `applyPotentiallyReadableStatusFilter` hard-codes `actorId + '/followers'` as the followers audience expression. `getActorStatuses` already carries the actual `followersUrl` via its `followersAudience` parameter; `getStatusReplies` does not. Inline comment above has a concrete fix path.
2. **`lib/database/sql/status.ts` line 106** — `applyPublicReadableStatusFilter` applies a one-hop check (Announce → content must be public), while `isStatusPubliclyReadable` recurses. This can cause `totalItems` > actual `orderedItems` for nested Announces pointing to private posts. Very rare in practice (announce-of-announce is non-standard), but the inconsistency is real.
3. **`lib/database/sql/status.ts` line 944** — `deleteStatus` compares a normalized `actorId` against a potentially non-normalized stored value. Root cause: `createNoteJob.ts` stores `actorId: note.attributedTo` raw. Inline comment above proposes normalizing at ingestion time.

---

lib/database/sql/status.ts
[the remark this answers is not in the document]
llun: Fixed in ceb36deb. The visibleToActorId SQL prefilter now matches followers-only replies against the reply author stored followersUrl, with the synthetic actorId/followers URL kept as a fallback. Added regression coverage for a custom followersUrl reply visible to an accepted follower.

app/api/inbox/getJobMessage.test.ts
llun: Fixed in ceb36deb. Added rejection tests for spoofed Delete activities and spoofed Undo Announce activities, covering the activityActorMismatch guard for both paths.

app/api/inbox/getJobMessage.ts
```diff
@@ -33,7 +36,79 @@ const NOTE_TYPES = [
...
+    normalizedActivityActorId !== normalizedVerifiedSenderActorId
+  )
+}
+
+const extractActivityPubIds = (value: unknown): string[] => {
+  if (typeof value === 'string') return [value]
+  if (Array.isArray(value)) return value.flatMap(extractActivityPubIds)
+  if (!isRecord(value)) return []
+
+  return typeof value.id === 'string' ? [value.id] : []
```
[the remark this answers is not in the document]
llun: Fixed in ceb36deb. extractActivityPubIds now keeps canonical id precedence for actor objects and falls back to href/url only when id is absent, so Link-form actor references are accepted without reintroducing the id+url false mismatch. Added coverage for a Create Note with a Link attributedTo object.

lib/database/sql/status.ts
```diff
@@ -84,6 +85,67 @@ export const StatusSQLDatabaseMixin = (
...
+
+  const applyPublicReadableStatusFilter = (query: Knex.QueryBuilder) =>
+    query.whereIn('statuses.id', publicRecipientStatusIds()).andWhere((qb) => {
+      qb.whereNot('statuses.type', StatusType.enum.Announce).orWhereExists(
+        function () {
+          this.select(database.raw('1'))
+            .from('statuses as original_statuses')
+            .whereRaw('original_statuses.id = statuses.content')
+            .whereIn('original_statuses.id', publicRecipientStatusIds())
+        }
```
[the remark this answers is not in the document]
llun: Fixed in ceb36deb. The publicOnly SQL filter now uses a recursive readable-status CTE, so announce-of-announce chains are only counted when the terminal boosted status is publicly readable. Added coverage for a nested public announce over a private root.

lib/database/sql/status.ts
```diff
@@ -851,17 +928,22 @@ export const StatusSQLDatabaseMixin = (
...
+        await deleteStatus({ actorId, statusId, trx })
       })
       return
     }
 
-    const status = await trx('statuses').where('id', statusId).first()
+    const statusQuery = trx('statuses').where('id', statusId)
+    if (actorId) {
+      statusQuery.where('actorId', actorId)
+    }
```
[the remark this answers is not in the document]
llun: Fixed in ceb36deb. Scoped status deletion now compares normalized stored and expected actor ids before deleting, and createNoteJob stores normalized actor ids for new remote notes. Added regression coverage for fragment/casing variants.

lib/database/sql/status.ts
llun: Fixed in ceb36deb. createNoteJob now normalizes note.attributedTo before storing actorId, and deleteStatus also normalizes both sides of scoped delete comparisons so existing legacy rows with raw actor ids can still be removed. Added tests for both paths.

app/api/users/[username]/outbox/route.ts
```diff
@@ -10,6 +12,19 @@ import { ACTIVITY_STREAM_URL } from '@/lib/utils/activitystream'
...
+  actorId: string,
+  limit?: number
+) =>
+  (
+    await database.getActorStatuses({
+      actorId,
+      ...(limit === undefined ? {} : { limit }),
+      publicOnly: true
+    })
+  ).filter(isStatusPubliclyReadable)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/outbox/route.ts]: Fixed in ceb36deb. Outbox page generation now backfills after post-filtering by fetching additional publicOnly batches with maxStatusId until the page is filled or there are no more candidates. Added route coverage for unreadable announces occupying a full first batch.

lib/database/sql/status.ts
llun: Fixed in ceb36deb. I did not pass currentActor.followersUrl into getStatusReplies because reply visibility depends on each reply author followers collection, not the viewer collection. The SQL prefilter now joins the status actor and compares recipients against that actor stored followersUrl, with the actorId/followers fallback preserved.

lib/database/sql/status.ts
```diff
@@ -84,6 +86,105 @@ export const StatusSQLDatabaseMixin = (
...
+  ]
+
+  const publicRecipientStatusIds = () =>
+    database('recipients')
+      .select('statusId')
+      .whereIn('recipients.actorId', publicActivityRecipients)
+
+  const publiclyReadableStatusIds = () =>
+    database
+      .withRecursive('publicly_readable_statuses', ['id'], (cte) => {
```
llun: **`WITH RECURSIVE` requires MySQL 8.0+.**

`knex.withRecursive` emits `WITH RECURSIVE` SQL. SQLite has supported this since 3.8.3 (2013) and PostgreSQL has supported it for years, but MySQL only added recursive CTE support in 8.0 (released 2018). Deployments running MySQL 5.7 or MariaDB < 10.2 will get a runtime error whenever `getActorStatusesCount({ publicOnly: true })`, `getActorStatuses({ publicOnly: true })`, or `getStatusReplies({ publicOnly: true })` is called.

If MySQL 8.0+ is an explicit requirement for this project, this is fine — just worth documenting. If MySQL 5.7 support is still needed, the Announce recursion would need to be handled differently (e.g., the original one-hop check plus a JS post-filter, as was done before).

---

lib/database/sql/status.ts
```diff
@@ -84,6 +86,105 @@ export const StatusSQLDatabaseMixin = (
...
+              .whereIn('announces.id', publicRecipientStatusIds())
+          })
+      })
+      .select('id')
+      .from('publicly_readable_statuses')
+
+  const applyPublicReadableStatusFilter = (query: Knex.QueryBuilder) =>
+    query.whereIn('statuses.id', publiclyReadableStatusIds())
+
+  const actorSettingsTextExpression = (actorAlias: string, key: string) => {
```
llun: **Nit: raw string interpolation in SQL helper — risky pattern for future callers.**

`actorAlias` and `key` are interpolated directly into the returned SQL string with no escaping:

```typescript
return `json_extract(${actorAlias}.settings, '$.${key}')`
```

The current callers (`'status_actors'`, `'followersUrl'`) are both hardcoded constants, so there is no injection risk today. But the function signature accepts arbitrary `string` parameters, and a future caller passing a user-supplied or externally-derived value would open a SQL injection path.

Consider restricting the API to avoid the risk pattern, for example by accepting a `key` parameter typed as a string literal union of the known valid keys, or by renaming the function to signal it only handles internal constants:

```typescript
// Option A: type-restrict to known keys
const actorSettingsTextExpression = (
  actorAlias: string,
  key: 'followersUrl'  // extend the union as needed
) => { ... }

// Option B: make the purpose clear
const followersUrlExpression = (actorAlias: string) =>
  actorSettingsTextExpression(actorAlias, 'followersUrl')
```

---

llun (commented):
Round 6 review on commit `ceb36de`. The three previously open ChatGPT Codex threads are now addressed — the recursive CTE replaces the one-hop SQL check, the `applyPotentiallyReadableStatusFilter` joins with the actors table to use the stored `followersUrl`, and `deleteStatus` normalizes both sides of the actorId comparison (with `createNoteJob.ts` also normalizing at ingestion time). The missing Delete/Undo+Announce rejection tests and the backfill pagination test have also been added. Overall the implementation is solid.

Two items to consider:

1. **`WITH RECURSIVE` MySQL 8.0+ requirement** (inline comment on line 101) — if MySQL 5.7 support is still needed, this will break at runtime. Worth confirming or documenting the minimum MySQL version.

2. **`actorSettingsTextExpression` raw string interpolation** (inline comment on line 125) — low risk today (hardcoded args), but a potentially dangerous pattern if the function is called with dynamic input in the future. Restricting the `key` parameter to a literal union or inlining the expression would eliminate the risk.

---

app/api/users/[username]/outbox/route.ts
```diff
@@ -1,15 +1,48 @@
...
+      limit,
+      ...(maxStatusId ? { maxStatusId } : {}),
+      publicOnly: true
+    })
+    if (batch.length === 0) break
+
+    statuses.push(...batch.filter(isStatusPubliclyReadable))
+    if (batch.length < limit) break
+
+    maxStatusId = batch[batch.length - 1].id
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/users/[username]/outbox/route.ts:40]: Fixed in 3b88884e. The outbox route still pages with maxStatusId, but getActorStatuses now orders and cursors by (createdAt, id), so backfill no longer skips readable statuses that share a timestamp with filtered announces. Added regression coverage for tied createdAt pagination.

lib/database/sql/status.ts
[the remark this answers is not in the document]
llun: Fixed in 3b88884e. Public readability filtering now seeds the recursive CTE from caller-provided target status IDs; actor status counts/pages and reply counts/pages pass actor/reply-scoped target sets instead of recursively scanning global statuses. Added SQL-shape coverage that the actor target filter is present before the recursive union.

lib/jobs/createNoteJob.ts
```diff
@@ -61,21 +69,22 @@ export const createNoteJob = createJobHandle(
 
     const text = getContent(note)
     const summary = getSummary(note)
+    const actorId = normalizeActorId(note.attributedTo) ?? note.attributedTo
```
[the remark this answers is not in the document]
llun: Fixed in 3b88884e. createNoteJob now writes attachments with the same normalized canonical actorId used for the status and actor records. Added regression coverage that actor-scoped attachment lookup finds fragment-attributed media under the normalized actor id and not the raw attributedTo value.

lib/database/sql/status.ts
llun: Addressed in 3b88884e. For MySQL-compatible clients, the public-readability query no longer emits WITH RECURSIVE; it uses a conservative non-recursive fallback for public notes and direct public announces of public non-announces, avoiding runtime failure on MySQL 5.7-era deployments. Added SQL-shape coverage that MySQL output does not include WITH RECURSIVE. Nested announce chains remain conservatively hidden on that fallback rather than leaking private chains or failing at runtime.

lib/database/sql/status.ts
llun: Fixed in 3b88884e. Replaced the generic actorSettingsTextExpression(actorAlias, key) helper with a followersUrl-only expression using hardcoded alias/key SQL per dialect, so future callers cannot pass arbitrary dynamic keys or aliases into the raw expression.

llun (commented):
Round 7 review on commit `3b88884`. Both issues from round 6 are cleanly addressed.

**MySQL compatibility** — `buildPubliclyReadableStatusIdsQuery` now branches on `clientName.includes('mysql')` and falls back to the original one-hop check when `WITH RECURSIVE` is not available. The MySQL path also adds `.whereNot('original_statuses.type', StatusType.enum.Announce)` to the EXISTS subquery, which means Announce-of-Announce chains are excluded on MySQL even when the root is public — a narrower set than the recursive CTE would produce on SQLite/PostgreSQL. This is an intentional and documented trade-off, not a bug.

**SQL injection risk** — `actorSettingsTextExpression(actorAlias, key)` is replaced with the inlined `statusActorFollowersUrlExpression()` that has no parameters at all. The three per-DB strings are hardcoded. ✓

**Everything else** looks correct:
- The `targetStatusIds` scoping in the CTE is well-structured — the base case scopes to the caller's set, the recursive step is free to follow content chains into the rest of the `statuses` table, and `targetId` threading ensures the final result contains the correct IDs.
- The `createdAt + id` compound tiebreaker for `minStatusId`/`maxStatusId` pagination is logically consistent with the added `orderBy('id', 'desc')` secondary sort.
- The `createAttachment` bug (using raw `note.attributedTo` instead of normalized `actorId`) is fixed.
- The structural SQL tests for `buildPubliclyReadableStatusIdsQuery` (checking `with recursive` presence and `actorid` position relative to `union`) are a reasonable unit-level sanity check for the branching logic.

No new issues found.

---

PR #675, opened 2026-05-12
fix: harden ActivityPub outbound fetches

llun:
## Summary
- Add a centralized safe remote-fetch helper for ActivityPub outbound requests.
- Enforce HTTPS-only remote fetches, DNS pre-resolution with private/local range rejection, redirect rechecks, redirect cap, body-size cap, timeouts, and cross-host credential/header stripping.
- Wire lib/utils/request.ts through the helper and add configurable max response size.

## Tests
- yarn run prettier --write .
- yarn lint
- yarn build
- yarn test


llun:
@codex review

lib/utils/request.ts
```diff
@@ -1,21 +1,40 @@
...
+  'TRACE'
+])
+const RETRYABLE_ERROR_CODES = new Set([
+  'EADDRINUSE',
+  'EAI_AGAIN',
+  'ECONNRESET',
+  'ENETUNREACH',
+  'ENOTFOUND',
+  'ETIMEDOUT'
+])
```
llun, line 32: `EADDRINUSE` means a **local** port is already in use — it is a server-side socket-binding error, not something that can happen during an outbound HTTP request. Including it in `RETRYABLE_ERROR_CODES` will never help: the remote server is not the problem, and a retry will hit the same local error. Please remove it from the set.

---

lib/utils/request.ts
```diff
@@ -28,105 +47,92 @@ export interface RequestResult {
...
+    } catch (error) {
+      if (
+        attempt >= options.numberOfRetry ||
+        !isRetryableRequestError(error, method)
+      ) {
+        throw error
+      }
+      attempt += 1
+    }
   }
```
llun, line 250: `retryNoise` is still a first-class field in `RequestConfig` and is settable via `ACTIVITIES_REQUEST_RETRY_NOISE`, but the manual retry loop here never applies it — there is no sleep/delay between attempts at all. This is a silent regression: callers who configured jitter to spread out thundering-herd retries no longer get it. Either remove `retryNoise` from `RequestConfig` (and the env-var docs), or add a delay that respects the configured noise value between loop iterations.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+          url: currentUrl.toString()
+        }
+      }
+
+      if (redirectCount >= effectiveMaxRedirects) {
+        throw new SafeRemoteFetchError(
+          'Too many redirects',
+          'ERR_TOO_MANY_REDIRECTS'
+        )
+      }
```
llun, line 660: The body of every response — including redirects — is fully buffered (up to `maxBodyBytes`, default 2 MB) before checking whether a `Location` header is present. Redirect bodies are usually empty or a few bytes of HTML, so this wastes bandwidth and lets a slow server hold the connection open for longer than necessary.

Suggestion: call `getRedirectLocation` first; if it returns a URL, destroy the body stream immediately instead of reading it, then proceed with the redirect.

```ts
const redirectUrl = getRedirectLocation(response, currentUrl)
if (redirectUrl) {
  response.body.destroy()
  // ... redirect handling
} else {
  const responseBody = await readResponseBody(response, maxBodyBytes)
  return { body: responseBody, ... }
}
```

---

lib/utils/request.ts
```diff
@@ -28,105 +47,92 @@ export interface RequestResult {
...
   const retryLimit = config.request?.numberOfRetry ?? MAX_RETRY_LIMIT
-  const retryNoise = config.request?.retryNoise
   const defaultResponseTimeout =
     responseTimeout ||
     config.request?.timeoutInMilliseconds ||
     DEFAULT_RESPONSE_TIMEOUT
+  const maxBodyBytes =
+    maxResponseSize ||
+    config.request?.maxResponseSizeInBytes ||
+    DEFAULT_SAFE_REMOTE_FETCH_MAX_BODY_BYTES
```
llun: Using `||` here means `maxResponseSize: 0` silently falls through to the config/default value instead of enforcing a zero-byte limit. Prefer a nullish/type check:

```ts
const maxBodyBytes =
  typeof maxResponseSize === 'number'
    ? maxResponseSize
    : config.request?.maxResponseSizeInBytes ??
      DEFAULT_SAFE_REMOTE_FETCH_MAX_BODY_BYTES
```

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+])
+const BODY_HEADERS = new Set([
+  'content-encoding',
+  'content-language',
+  'content-length',
+  'content-location',
+  'content-type',
+  'digest',
+  'signature'
+])
```
llun, line 26: `'signature'` appears in **both** `SENSITIVE_REDIRECT_HEADERS` and `BODY_HEADERS`. In `buildHeaders` the two checks are independent (`isCrossHostRedirect` vs `methodChanged`), so the duplication is harmless in practice — but it obscures intent. If the goal is "strip the Signature header on any redirect that either crosses a host boundary or changes the method", a brief comment explaining the overlap would prevent future confusion (or the overlap could be modelled explicitly).

---

lib/utils/safeRemoteFetch.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      })
+    })
+
+    await expect(
+      safeRemoteFetch({
+        url: 'https://safe.example/large',
+        maxBodyBytes: 8
+      })
+    ).rejects.toThrow('Response body too large')
+  })
```
llun, line 75: This test covers the streaming size limit (chunks accumulate past `maxBodyBytes`), but there is no test for the **`content-length` header fast-path** in `readResponseBody` (lines 441–448 of `safeRemoteFetch.ts`). That path immediately destroys the stream and throws when `content-length > maxBodyBytes` without reading any bytes. A dedicated test case here would make the coverage explicit and guard against regressions:

```ts
it('rejects immediately when content-length exceeds the byte limit', async () => {
  const safeRemoteFetch = createSafeRemoteFetch({
    resolveHost: async () => [{ address: '203.0.113.10', family: 4 }],
    transport: async () => ({
      statusCode: 200,
      headers: { 'content-length': '100' },
      body: streamFrom(['small'])
    })
  })

  await expect(
    safeRemoteFetch({ url: 'https://safe.example/large', maxBodyBytes: 8 })
  ).rejects.toThrow('Response body too large')
})
```

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
+import got, { Headers, Method, OptionsInit } from 'got'
+import { lookup } from 'node:dns/promises'
+import net from 'node:net'
+import { Readable } from 'node:stream'
```
llun: `Readable` is only used as a type annotation (`body: Readable` in `SafeRemoteFetchTransportResponse`). Use `import type` to make this explicit and let TypeScript strip it at compile time:

```ts
import type { Readable } from 'node:stream'
```

---

lib/utils/safeRemoteFetch.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      transport: async () => okResponse()
+    })
+
+    await expect(
+      safeRemoteFetch({ url: 'https://[::1]/actor' })
+    ).rejects.toThrow('Unsafe remote address')
+    await expect(
+      safeRemoteFetch({ url: 'https://[fc00::1]/actor' })
+    ).rejects.toThrow('Unsafe remote address')
+  })
```
llun, line 246: The IPv6 literal tests cover `::1` (loopback) and `fc00::1` (unique-local), but there is no test for **IPv4-mapped IPv6** literals such as `https://[::ffff:10.0.0.1]/actor`. The code handles this via `isIpv4MappedIpv6` → `isUnsafeIpv4`, but without a regression test the path can silently break. Consider adding:

```ts
await expect(
  safeRemoteFetch({ url: 'https://[::ffff:10.0.0.1]/actor' })
).rejects.toThrow('Unsafe remote address')
```

---

llun (commented):
The overall approach is solid — DNS pre-resolution with a fixed `dnsLookup` override on `got`, redirect re-validation, body-size caps, credential stripping, and HTTPS enforcement are all the right mechanisms for hardening outbound ActivityPub fetches.

There are a few issues to fix before merging:

**Bugs / regressions**
1. `EADDRINUSE` in `RETRYABLE_ERROR_CODES` — local bind error, never a retryable outbound condition. See inline comment.
2. `retryNoise` is silently dropped — the config field and env var still exist but the new retry loop never sleeps/jitters between attempts. See inline comment.
3. Redirect bodies are fully buffered before the redirect is detected — check `getRedirectLocation` first and destroy the stream on a redirect rather than reading it. See inline comment.
4. `maxResponseSize || ...` coerces `0` to the default — use a `typeof` or `??` check. See inline comment.

**Test coverage gaps**
5. No test for the `content-length` header fast-path in `readResponseBody`. See inline comment.
6. No test for IPv4-mapped IPv6 literals (`::ffff:10.0.0.1`). See inline comment.

**Minor**
7. `'signature'` is duplicated across `SENSITIVE_REDIRECT_HEADERS` and `BODY_HEADERS` — harmless but worth a comment explaining why. See inline comment.
8. `import { Readable }` should be `import type { Readable }` since it is only used as a type. See inline comment.

---

lib/utils/request.ts
llun: Fixed. Removed EADDRINUSE from RETRYABLE_ERROR_CODES and added regression coverage that local port binding errors are not retried.

lib/utils/request.ts
llun: Fixed. Request retries now wait between attempts using exponential backoff plus configured retryNoise jitter, with regression coverage.

lib/utils/safeRemoteFetch.ts
llun: Fixed. Redirect responses are checked for Location before body buffering, and redirect bodies are destroyed immediately.

lib/utils/request.ts
llun: Fixed. maxResponseSize now uses an explicit number check so maxResponseSize: 0 is honored, with regression coverage.

lib/utils/safeRemoteFetch.ts
llun: Fixed. Added a comment explaining that Signature covers request metadata and the signed body digest, so it is stripped for cross-host and method-changing redirects.

lib/utils/safeRemoteFetch.test.ts
llun: Fixed. Added content-length fast-path coverage that verifies the stream is destroyed without reading bytes when declared length exceeds the limit.

lib/utils/safeRemoteFetch.ts
llun: Fixed. Readable is now imported with import type.

lib/utils/safeRemoteFetch.test.ts
llun: Fixed. Added IPv4-mapped IPv6 regression coverage for ::ffff:10.0.0.1.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+  if (unsafeAddress) {
+    throw createUnsafeUrlError(
+      `Unsafe remote address: ${unsafeAddress.address}`
+    )
+  }
+
+  const firstAddress = addresses[0]
+  if (!firstAddress) throw createUnsafeUrlError('Unable to resolve remote host')
+
+  return firstAddress
```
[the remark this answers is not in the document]
llun: Fixed. safeRemoteFetch now carries all validated DNS answers into the transport, and dnsLookup returns the full set for Node family autoselection instead of pinning only the first address.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+  const isUnspecified = bytes.every((byte) => byte === 0)
+  const isLoopback =
+    bytes.slice(0, 15).every((byte) => byte === 0) && bytes[15] === 1
+  const isUniqueLocal = (bytes[0] & 0xfe) === 0xfc
+  const isLinkLocal = bytes[0] === 0xfe && (bytes[1] & 0xc0) === 0x80
+  const isMulticast = bytes[0] === 0xff
+
+  return (
+    isUnspecified || isLoopback || isUniqueLocal || isLinkLocal || isMulticast
+  )
```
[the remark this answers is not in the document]
llun: Fixed. IPv4-compatible embedded IPv6 tails now go through the IPv4 unsafe-range checks, with regression coverage for ::192.168.0.1.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+      }
+    }
+    const stream = got.stream(url.toString(), options)
+    const rejectBeforeResponse = (error: Error) => {
+      reject(error)
+    }
+
+    stream.once('error', rejectBeforeResponse)
+    stream.once('response', (response) => {
+      stream.off('error', rejectBeforeResponse)
```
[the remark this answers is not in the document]
llun: Fixed. gotTransport now keeps a no-op error observer for the stream lifetime so errors emitted after response headers are still observed.

lib/utils/request.ts
```diff
@@ -28,105 +47,92 @@ export interface RequestResult {
...
-    method: method as Method,
-    body
+    maxBodyBytes,
+    method,
+    numberOfRetry:
+      typeof numberOfRetry === 'number' ? numberOfRetry : retryLimit,
+    readTimeoutInMilliseconds: defaultResponseTimeout,
+    timeoutInMilliseconds: defaultResponseTimeout
   }
 }
```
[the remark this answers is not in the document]
llun: Fixed. The manual retry loop now uses backoff plus retryNoise jitter instead of retrying immediately.

lib/utils/request.test.ts
```diff
@@ -114,6 +114,16 @@ const closeServer = async (server: Server) => {
   })
 }
 
+const withDevelopmentNodeEnv = async (callback: () => Promise<void>) => {
+  const originalNodeEnv = process.env.NODE_ENV
+  process.env.NODE_ENV = 'development'
+  try {
+    await callback()
+  } finally {
+    process.env.NODE_ENV = originalNodeEnv
```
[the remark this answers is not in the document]
llun: Fixed. withDevelopmentNodeEnv now deletes NODE_ENV when it was originally unset.

lib/utils/safeRemoteFetch.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      const safeRemoteFetch = createSafeRemoteFetch({
+        resolveHost: async () => [{ address: '10.0.0.5', family: 4 }],
+        transport: async () => okResponse()
+      })
+
+      await expect(
+        safeRemoteFetch({ url: 'http://localhost/actor' })
+      ).rejects.toThrow('Unsafe remote address')
+    } finally {
+      process.env.NODE_ENV = originalNodeEnv
```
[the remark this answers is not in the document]
llun: Fixed. The safeRemoteFetch test restores NODE_ENV by deleting it when it was originally unset.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+    first === 0 ||
+    first === 10 ||
+    first === 127 ||
+    (first === 169 && second === 254) ||
+    (first === 172 && second >= 16 && second <= 31) ||
+    (first === 192 && second === 168) ||
+    (first >= 224 && first <= 239) ||
+    first >= 240 ||
+    (first === 255 && second === 255 && third === 255 && fourth === 255)
+  )
```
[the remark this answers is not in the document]
llun: Fixed. Expanded IPv4 unsafe ranges to include shared address space, IETF protocol assignments, TEST-NET ranges, 6to4 relay, benchmarking, and first >= 224.

lib/utils/safeRemoteFetch.ts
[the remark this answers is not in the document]
llun: Fixed. Expanded IPv6 unsafe ranges to include discard-only, ORCHIDv2, and documentation prefixes with regression coverage.

lib/utils/safeRemoteFetch.ts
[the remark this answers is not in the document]
llun: Fixed. Added 100.64.0.0/10 and simplified multicast/reserved handling with first >= 224.

lib/utils/safeRemoteFetch.ts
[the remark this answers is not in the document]
llun: Fixed. Added the 64:ff9b::/96 IPv4-IPv6 translation prefix to the unsafe IPv6 checks with regression coverage.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+        throw new SafeRemoteFetchError(
+          'Too many redirects',
+          'ERR_TOO_MANY_REDIRECTS'
+        )
+      }
+
+      previousUrl = currentUrl
+      currentUrl = redirectUrl
+      redirectCount += 1
+      methodChanged = response.statusCode === 303
```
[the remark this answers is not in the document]
llun: Fixed. Redirects now rewrite non-GET methods to lowercase get for 301, 302, and 303. I kept the status list internal because safeRemoteFetch has one fixed redirect safety policy and no callers need alternate redirect semantics.

lib/utils/request.ts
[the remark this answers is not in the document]
llun: Fixed. The manual retry loop now waits with exponential backoff plus configured jitter; each attempt continues to use safeRemoteFetch connect/request/response/socket timeouts.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+    first === 10 ||
+    first === 127 ||
+    (first === 169 && second === 254) ||
+    (first === 172 && second >= 16 && second <= 31) ||
+    (first === 192 && second === 168) ||
+    (first >= 224 && first <= 239) ||
+    first >= 240 ||
+    (first === 255 && second === 255 && third === 255 && fourth === 255)
+  )
+}
```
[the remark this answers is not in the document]
llun: Fixed. Expanded isUnsafeIpv4 for the listed RFC 6890/reserved ranges and added reserved-range tests.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,573 @@
...
+  const isLoopback =
+    bytes.slice(0, 15).every((byte) => byte === 0) && bytes[15] === 1
+  const isUniqueLocal = (bytes[0] & 0xfe) === 0xfc
+  const isLinkLocal = bytes[0] === 0xfe && (bytes[1] & 0xc0) === 0x80
+  const isMulticast = bytes[0] === 0xff
+
+  return (
+    isUnspecified || isLoopback || isUniqueLocal || isLinkLocal || isMulticast
+  )
+}
```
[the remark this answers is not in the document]
llun: Fixed. Expanded isUnsafeIpv6 for discard-only, documentation, and ORCHIDv2 ranges with regression coverage.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,628 @@
...
+      resolvedAddresses[0]
+    if (!resolvedAddress) return
+    ;(
+      callback as (
+        error: NodeJS.ErrnoException | null,
+        address: string,
+        family: 4 | 6
+      ) => void
+    )(null, resolvedAddress.address, resolvedAddress.family)
+  }) as OptionsInit['dnsLookup']
```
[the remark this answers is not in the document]
llun: Fixed in 3ba0fe80. The fixed dnsLookup implementation now handles callback-only, numeric family, all:true, and all:true with family filtering overloads while returning only prevalidated addresses.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,628 @@
...
+    bytes[0] === 0x00 &&
+    bytes[1] === 0x64 &&
+    bytes[2] === 0xff &&
+    bytes[3] === 0x9b &&
+    bytes.slice(4, 12).every((byte) => byte === 0)
+  const isOrchidV2 =
+    bytes[0] === 0x20 &&
+    bytes[1] === 0x01 &&
+    bytes[2] === 0x00 &&
+    (bytes[3] & 0xf0) === 0x10
```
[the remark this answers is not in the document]
llun: Fixed in 3ba0fe80. The unsafe IPv6 checks now cover both deprecated ORCHID 2001:10::/28 and ORCHIDv2 2001:20::/28, with tests.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,628 @@
...
+    isUnspecified ||
+    isLoopback ||
+    isDiscard ||
+    isIpv4Ipv6Translation ||
+    isOrchidV2 ||
+    isDocumentation ||
+    isUniqueLocal ||
+    isLinkLocal ||
+    isMulticast
+  )
```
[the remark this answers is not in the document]
llun: Fixed in 3ba0fe80. Teredo 2001::/32 and 6to4 2002::/16 are now rejected by the IPv6 safety checks.

lib/utils/request.ts
```diff
@@ -28,105 +49,112 @@ export interface RequestResult {
...
+  error: unknown,
+  method: SafeRemoteFetchMethod
+) => {
+  const code = (error as NodeJS.ErrnoException).code
+  return (
+    RETRYABLE_METHODS.has(method.toUpperCase()) &&
+    typeof code === 'string' &&
+    RETRYABLE_ERROR_CODES.has(code)
+  )
+}
```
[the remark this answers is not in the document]
llun: Fixed in 3ba0fe80. SafeRemoteFetch validation errors, including ERR_UNSAFE_REMOTE_URL, are explicitly treated as non-retryable and covered by a request test.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,628 @@
...
+  const isDiscard =
+    bytes[0] === 0x01 &&
+    bytes[1] === 0x00 &&
+    bytes.slice(2, 8).every((byte) => byte === 0)
+  const isIpv4Ipv6Translation =
+    bytes[0] === 0x00 &&
+    bytes[1] === 0x64 &&
+    bytes[2] === 0xff &&
+    bytes[3] === 0x9b &&
+    bytes.slice(4, 12).every((byte) => byte === 0)
```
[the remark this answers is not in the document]
llun: Fixed in 3ba0fe80. NAT64 64:ff9b::/96 addresses now decode the embedded IPv4 and apply the same IPv4 safety validation instead of rejecting the whole prefix.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,652 @@
...
+  }
+
+  const isUnspecified = bytes.every((byte) => byte === 0)
+  const isLoopback =
+    bytes.slice(0, 15).every((byte) => byte === 0) && bytes[15] === 1
+  const isDiscard =
+    bytes[0] === 0x01 &&
+    bytes[1] === 0x00 &&
+    bytes.slice(2, 8).every((byte) => byte === 0)
+  const isTeredo =
```
llun, line 242: The `64:ff9b:1::/48` prefix (RFC 8215 Well-Known Prefix 2 for NAT64) is not covered here. `isNat64Ipv6` only checks `bytes[0-3] == 00 64 ff 9b` and `bytes[4-11]` all-zero, which matches `64:ff9b::/96` but not `64:ff9b:1::/48`.

An address like `[64:ff9b:1::10.0.0.1]` has `bytes[5] = 0x01`, so `bytes.slice(4, 12).every(b => b === 0)` is `false` and `isNat64Ipv6` returns `false`. None of the other checks (`isUniqueLocal`, `isLinkLocal`, etc.) catch it either, so the embedded `10.0.0.1` is not detected and the address is allowed through.

A minimal fix is to add a second branch:
```ts
const isNat64WKP2Ipv6 = (bytes: number[]) =>
  bytes[0] === 0x00 &&
  bytes[1] === 0x64 &&
  bytes[2] === 0xff &&
  bytes[3] === 0x9b &&
  bytes[4] === 0x00 &&
  bytes[5] === 0x01 &&
  bytes.slice(6, 12).every((byte) => byte === 0)
```
and include it in the `isIpv4MappedIpv6 || hasIpv4CompatibleIpv6Tail || isNat64Ipv6` guard, plus add a corresponding test case (e.g. `https://[64:ff9b:1::10.0.0.1]/actor`).

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,652 @@
...
+        )
+      }
+
+      previousUrl = currentUrl
+      currentUrl = redirectUrl
+      redirectCount += 1
+      methodChanged =
+        [301, 302, 303].includes(response.statusCode) &&
+        currentMethod.toUpperCase() !== 'GET'
+      if (methodChanged) {
```
llun: RFC 7230 §3.1.1 says HTTP method tokens are case-sensitive and must be uppercase. Setting `currentMethod = 'get'` sends a lowercase method to subsequent requests after a method-changing redirect. `got` happens to normalize it internally, but it is still technically incorrect and the test name ("to lowercase GET") reads like a bug documented as a feature rather than intentional behaviour. Please use `'GET'` here.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,652 @@
...
+      method,
+      retry: RETRY_DISABLED,
+      throwHttpErrors: false,
+      timeout: {
+        connect: connectTimeoutInMilliseconds,
+        request: connectTimeoutInMilliseconds + readTimeoutInMilliseconds,
+        response: readTimeoutInMilliseconds,
+        socket: readTimeoutInMilliseconds
+      }
+    }
```
llun, line 448: The empty `observeStreamError` listener is added permanently and never removed. Its only purpose is to prevent Node.js from throwing an unhandled `'error'` event, but because it is never torn down it silently swallows any stream errors that occur after `rejectBeforeResponse` is unregistered (i.e. between the `'response'` event and when `readResponseBody` begins consuming the stream). Errors in that window are eaten without being surfaced to the caller.

A cleaner alternative: keep `rejectBeforeResponse` registered until the stream is fully consumed. For example, keep it as a long-lived listener and remove it only after `readResponseBody` returns (or throws):

```ts
stream.once('error', rejectBeforeResponse)
stream.once('response', (response) => {
  resolve({
    body: stream,
    headers: response.headers,
    statusCode: response.statusCode
  })
})
```

Then in `readResponseBody`, wrap the `for await` so that any stream error is surfaced:
```ts
try {
  for await (const chunk of response.body) { ... }
} finally {
  response.body.off('error', /* the reject handler */)
}
```

Or at minimum rename `observeStreamError` to something explicit like `suppressUnhandledStreamError` so the intent is obvious.

---

lib/utils/request.ts
```diff
@@ -1,24 +1,50 @@
...
 
 const USER_AGENT = `activities.next/${packageJson.version}`
 const DEFAULT_RESPONSE_TIMEOUT = 10000
 const MAX_RETRY_LIMIT = 1
+const DEFAULT_RETRY_NOISE = 100
+const RETRY_BACKOFF_LIMIT = Number.POSITIVE_INFINITY
+const RETRYABLE_METHODS = new Set([
+  'DELETE',
+  'GET',
+  'HEAD',
```
llun, line 18: `RETRY_BACKOFF_LIMIT` is `Number.POSITIVE_INFINITY`, meaning the exponential backoff has no ceiling. With the default `numberOfRetry = 1` this is harmless, but any caller that sets a higher retry count (or the env-var `ACTIVITIES_REQUEST_RETRY`) could end up waiting minutes between attempts (`2^10 * 1000 = ~17 minutes` for 11 retries). A sensible cap like `30_000` ms would bound the worst case without changing typical behaviour.

---

llun (commented):
Good progress — all eight issues from the first round are addressed. Three new issues remain, one of which is a security concern:

**Security**
1. `isNat64Ipv6` doesn't cover `64:ff9b:1::/48` (RFC 8215 WKP2) — an address like `[64:ff9b:1::10.0.0.1]` with an embedded private IPv4 passes all checks undetected. See inline comment.

**Bugs**
2. `currentMethod = 'get'` (lowercase) after method-changing redirects — technically incorrect per RFC 7230 and the test name documents this as "lowercase GET". Should be `'GET'`. See inline comment.
3. `observeStreamError` empty listener is never removed — silently swallows stream errors that occur between the `'response'` event and when `readResponseBody` begins consuming the stream. See inline comment.

**Non-critical**
4. `RETRY_BACKOFF_LIMIT = Number.POSITIVE_INFINITY` gives unbounded retry delays for high retry counts. See inline comment.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,652 @@
...
+
+      previousUrl = currentUrl
+      currentUrl = redirectUrl
+      redirectCount += 1
+      methodChanged =
+        [301, 302, 303].includes(response.statusCode) &&
+        currentMethod.toUpperCase() !== 'GET'
+      if (methodChanged) {
+        currentMethod = 'get'
+        currentBody = undefined
```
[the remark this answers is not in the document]
llun: Fixed in 25f76e88. safeRemoteFetch now preserves method, body, and body headers across redirects instead of rewriting non-GET deliveries to GET, while still stripping sensitive headers on cross-host redirects.

lib/utils/request.ts
```diff
@@ -28,105 +54,119 @@ export interface RequestResult {
...
-  numberOfRetry
-}: Omit<RequestOptions, 'url' | 'maxResponseSize'>): OptionsInit => {
+  numberOfRetry,
+  retryNoise,
+  maxResponseSize
+}: Omit<RequestOptions, 'url'>) => {
   const config = getConfig()
   const retryLimit = config.request?.numberOfRetry ?? MAX_RETRY_LIMIT
-  const retryNoise = config.request?.retryNoise
+  const configuredRetryNoise = config.request?.retryNoise ?? DEFAULT_RETRY_NOISE
```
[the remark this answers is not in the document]
llun: Fixed in 25f76e88. Request option resolution now preserves an explicit retryNoise: null instead of replacing it with the default jitter.

lib/utils/safeRemoteFetch.ts
llun: Fixed in 25f76e88. Added RFC 8215 64:ff9b:1::/48 detection and tests so embedded private IPv4 addresses in that NAT64 prefix are rejected.

lib/utils/safeRemoteFetch.ts
llun: Resolved in 25f76e88. The method-changing redirect branch was removed because non-GET redirects are now preserved, so we no longer generate a lowercase GET on redirected requests.

lib/utils/safeRemoteFetch.ts
llun: Fixed in 25f76e88. got stream errors after the response event are now forwarded through a PassThrough body and listeners are removed when the response body closes, avoiding the permanent empty error listener.

lib/utils/request.ts
llun: Fixed in 25f76e88. Retry backoff is now capped at 30 seconds including configured jitter.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,651 @@
...
+        throw new SafeRemoteFetchError(
+          'Too many redirects',
+          'ERR_TOO_MANY_REDIRECTS'
+        )
+      }
+
+      previousUrl = currentUrl
+      currentUrl = redirectUrl
+      redirectCount += 1
+      currentHeaders = requestHeaders
```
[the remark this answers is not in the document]
llun: Fixed in c76fac3f. 303 redirects now switch the follow-up request to GET, drop the request body, and remove body/signature headers before following.

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,651 @@
...
+    if (bodyBytes > maxBodyBytes) {
+      const error = createResponseTooLargeError()
+      response.body.destroy()
+      throw error
+    }
+
+    chunks.push(buffer)
+  }
+
+  return Buffer.concat(chunks).toString('utf8')
```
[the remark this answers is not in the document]
llun: Fixed in c76fac3f. safeRemoteFetch now honors an explicit response charset and avoids UTF-8 replacement corruption for undeclared non-UTF8 text bodies, with regression coverage.

llun:
@codex review

lib/utils/request.ts
```diff
@@ -28,105 +54,123 @@ export interface RequestResult {
...
-      const declaredLength = Number(declaredLengthValue ?? 0)
-      if (Number.isFinite(declaredLength) && declaredLength > maxResponseSize) {
-        stream.destroy(new Error('Response body too large'))
-      }
-    })
-    stream.on('data', (chunk) => {
-      const buffer = Buffer.from(chunk)
-      bodyBytes += buffer.byteLength
+const getRetryDelay = (attempt: number, retryNoise: number | null) => {
+  const noise = retryNoise ? Math.random() * retryNoise : 0
```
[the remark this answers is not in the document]
llun: Fixed in d97c39af. Negative retryNoise now uses its absolute value, while null and 0 preserve no-noise behavior.

lib/utils/request.ts
```diff
@@ -28,105 +54,123 @@ export interface RequestResult {
...
+        method: options.method,
+        readTimeoutInMilliseconds: options.readTimeoutInMilliseconds,
+        timeoutInMilliseconds: options.timeoutInMilliseconds,
+        url
+      })
+    } catch (error) {
+      if (
+        attempt >= options.numberOfRetry ||
+        !isRetryableRequestError(error, method)
+      ) {
```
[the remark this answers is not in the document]
llun: Fixed in d97c39af. Returned transient HTTP statuses are retried for retryable methods after the response body is read.

lib/utils/request.ts
```diff
@@ -1,24 +1,50 @@
...
+  'PUT',
+  'TRACE'
+])
+const RETRYABLE_ERROR_CODES = new Set([
+  'EAI_AGAIN',
+  'ECONNRESET',
+  'ENETUNREACH',
+  'ENOTFOUND',
+  'ETIMEDOUT'
+])
```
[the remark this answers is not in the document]
llun: Fixed in d97c39af. Restored the full default transient socket retry code set.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,721 @@
...
+      resolvedAddresses.find(filterByRequestedFamily) ?? resolvedAddresses[0]
+    if (!resolvedAddress) return
+    ;(
+      callback as (
+        error: NodeJS.ErrnoException | null,
+        address: string,
+        family: 4 | 6
+      ) => void
+    )(null, resolvedAddress.address, resolvedAddress.family)
+  }) as OptionsInit['dnsLookup']
```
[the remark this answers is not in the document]
llun: No code change required here; this review note confirms the fixed-DNS lookup is pinning requests to pre-validated IPs as intended.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,721 @@
...
+  'content-length',
+  'content-type',
+  'digest',
+  'signature'
+])
+const RETRY_DISABLED = { limit: 0 }
+
+export type SafeRemoteFetchMethod = Method
+export type SafeRemoteFetchHeaders = Headers
+
```
llun, line 31: `content-encoding` is absent from `BODY_REDIRECT_HEADERS`. If an outgoing POST carries `Content-Encoding: gzip` (or any other encoding) and receives a 303, `stripBodyHeaders` is called but `content-encoding` survives into the follow-up GET request — despite the GET having no body. Strictly speaking the header is meaningless on a bodyless request, but it is incorrect and some servers may reject it.

The same applies to `content-language` and `content-location`, which were in the previous `BODY_HEADERS` set but didn't make it into `BODY_REDIRECT_HEADERS`.

Suggested additions:
```ts
const BODY_REDIRECT_HEADERS = new Set([
  'content-digest',
  'content-encoding',   // add
  'content-language',   // add
  'content-length',
  'content-location',   // add
  'content-type',
  'digest',
  'signature'
])
```
The 303 test should also gain a `content-encoding` header in the initial POST to verify it is stripped on the redirect.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -0,0 +1,721 @@
...
+) => {
+  const charset = getResponseCharset(headers)
+  if (charset) {
+    try {
+      return new TextDecoder(charset).decode(body)
+    } catch {
+      return body.toString('utf8')
+    }
+  }
+
```
llun: Charset detection and the Latin-1 fallback are new behaviour unrelated to the security hardening goal of this PR. ActivityPub content is always JSON-LD; RFC 8259 mandates UTF-8, so any non-UTF-8 response would fail to parse as JSON regardless. This adds untested complexity:

1. The `charset` value in the `Content-Type` header comes from the remote server (potentially attacker-controlled). Although `TextDecoder` throws on unknown charsets and the code falls back to UTF-8, the error path is untested.
2. The Latin-1 fallback (`body.toString('latin1')`) silently changes the decoded string for any response that isn't valid UTF-8, which differs from the previous behaviour and could hide bugs.

Suggestion: keep the original `Buffer.concat(chunks).toString('utf8')` here and move charset detection to a follow-up PR if it is genuinely needed.

---

llun (commented):
All four issues from the previous round are fixed. Only two minor issues remain:

1. **`BODY_REDIRECT_HEADERS` is missing `content-encoding`, `content-language`, and `content-location`** — on a 303 POST→GET redirect, these headers from the original POST survive into the GET request that has no body. See inline comment.

2. **`decodeResponseBody` charset detection is scope creep** — ActivityPub is always JSON-LD/UTF-8 (RFC 8259). The new charset-aware decoding and Latin-1 fallback add untested complexity and change existing behaviour. This would be better as a follow-up PR. See inline comment.

The core security model (DNS pre-resolution pinning, NAT64 WKP2 coverage, redirect body destruction, stream error forwarding, retry backoff cap, correct 303-only method change) all look good.

---

lib/utils/safeRemoteFetch.ts
llun: Fixed in fb0149a6. Added content-encoding, content-language, and content-location to BODY_REDIRECT_HEADERS and extended the 303 redirect test to verify only host remains on the follow-up GET.

lib/utils/safeRemoteFetch.ts
llun: Fixed in fb0149a6. Removed charset parsing, TextDecoder usage, and the Latin-1 fallback; readResponseBody now decodes the buffered response with UTF-8 again.

lib/utils/request.ts
```diff
@@ -1,24 +1,56 @@
...
+  'EAI_AGAIN',
+  'EADDRINUSE',
+  'ECONNREFUSED',
+  'ECONNRESET',
+  'EPIPE',
+  'ENETUNREACH',
+  'ENOTFOUND',
+  'ETIMEDOUT'
+])
+const RETRYABLE_STATUS_CODES = new Set([
```
llun, line 33: 413 (Content Too Large) is a permanent client error — the server is rejecting the payload because it exceeds the configured limit. Retrying the exact same request with the same body will never succeed; it only wastes the retry budget and delays the eventual failure returned to the caller.

The other codes in the set (408, 429, 5xx, Cloudflare 521/522/524) are all transient — the server may recover or the client may succeed on a later attempt. 413 does not share that property.

Please remove 413 from `RETRYABLE_STATUS_CODES`.

---

lib/utils/request.test.ts
```diff
@@ -201,57 +226,322 @@ describe('request utility', () => {
...
+        })
+      ).rejects.toMatchObject({
+        code: 'ERR_UNSAFE_REMOTE_URL',
+        name: 'SafeRemoteFetchError'
+      })
+      expect(setTimeoutSpy).not.toHaveBeenCalled()
+      expect(fetchMock).not.toHaveBeenCalled()
+    })
+
+    it('retries transient HTTP status responses after reading the body', async () => {
```
llun: The status-code retry path is only tested for 500. The other codes — 408, 413, 429, 502, 503, 504, 521, 522, 524 — have no coverage. An `it.each` parallel to the socket-error test would close this gap:

```ts
it.each([408, 429, 500, 502, 503, 504, 521, 522, 524])(
  'retries transient HTTP status %s',
  async (status) => {
    jest.useFakeTimers()
    fetchMock.mockResponseOnce('transient', { status })
    fetchMock.mockResponseOnce('ok', { status: 200 })

    const responsePromise = request({
      url: 'https://example.com/api/test',
      numberOfRetry: 1,
      retryNoise: null
    })
    await jest.advanceTimersByTimeAsync(1000)

    await expect(responsePromise).resolves.toMatchObject({ statusCode: 200 })
    expect(fetchMock).toHaveBeenCalledTimes(2)
  }
)
```

This also makes it straightforward to verify that 413 (once removed) is no longer retried.

---

llun (commented):
Both issues from round 3 are fixed — `BODY_REDIRECT_HEADERS` now includes `content-encoding`, `content-language`, and `content-location` (with a test verifying all three are stripped on 303), and the charset-detection code is gone.

The new status-code retry logic in `request.ts` looks good overall, but two small items remain:

1. **413 should not be in `RETRYABLE_STATUS_CODES`** — it is a permanent client error; retrying the same payload will always fail. See inline comment.

2. **Missing `it.each` test for all retryable status codes** — only 500 is currently exercised. See inline comment.

Everything else — the IP-range checks, stream handling, redirect logic, DNS pinning, retry backoff cap, and noise handling — looks correct and well tested.

---

lib/utils/request.ts
llun: Fixed in 38d1f9fb. Removed 413 from RETRYABLE_STATUS_CODES so Content Too Large responses return immediately without spending a retry attempt.

lib/utils/request.test.ts
llun: Fixed in 38d1f9fb. Replaced the single 500 retry test with table coverage for 408, 429, 500, 502, 503, 504, 521, 522, and 524, and added an assertion that 413 is not retried.

lib/utils/request.ts
```diff
@@ -28,105 +60,140 @@ export interface RequestResult {
...
+      if (
+        attempt >= options.numberOfRetry ||
+        !isRetryableStatusCode(response.statusCode, method)
+      ) {
+        return response
+      }
 
-  return got(url, options)
+      attempt += 1
+      await wait(getRetryDelay(attempt - 1, options.retryNoise))
```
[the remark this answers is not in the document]
llun: Fixed in 1b8524ba. Status retries now honor Retry-After headers before retrying, including timeout-bounded behavior matching the previous Got retry semantics, with regression coverage.

lib/config/request.ts
```diff
@@ -17,26 +18,32 @@ export const RequestConfig = z.object({
...
+    .positive()
+    .safe()
+    .default(DEFAULT_MAX_RESPONSE_SIZE_IN_BYTES)
 })
 export type RequestConfig = z.infer<typeof RequestConfig>
 
+const getOptionalInteger = (key: string) => {
+  const value = process.env[key]
+  if (!value) return undefined
+  return parseInt(value, 10)
```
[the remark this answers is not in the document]
llun: Fixed in 1b8524ba. getOptionalInteger now explicitly treats only undefined and empty strings as missing, and config coverage verifies ACTIVITIES_REQUEST_RETRY_NOISE="0" parses as 0.

lib/utils/request.test.ts
```diff
@@ -201,57 +226,389 @@ describe('request utility', () => {
...
+      await jest.advanceTimersByTimeAsync(1000)
+
+      await expect(responsePromise).resolves.toMatchObject({
+        body: 'ok',
+        statusCode: 200
+      })
+      expect(fetchMock).toHaveBeenCalledTimes(2)
+    })
+
+    it('does not retry when retry-after exceeds the request timeout', async () => {
```
llun, line 416: `getRetryAfterDelay` supports both the delta-seconds form and the HTTP-date form (`Date.parse`), but only the seconds form is exercised in tests. A malformed date or a timezone edge case in the `Date.parse` path could silently regress. One test case is enough to pin the behaviour:

```ts
it('honors retry-after with an HTTP-date header', async () => {
  jest.useFakeTimers()
  const retryAt = new Date(Date.now() + 2000).toUTCString()
  fetchMock.mockResponseOnce('rate limited', {
    headers: { 'retry-after': retryAt },
    status: 429
  })
  fetchMock.mockResponseOnce('ok', { status: 200 })

  const responsePromise = request({
    url: 'https://example.com/api/test',
    numberOfRetry: 1,
    retryNoise: null
  })
  await jest.advanceTimersByTimeAsync(2000)

  await expect(responsePromise).resolves.toMatchObject({ statusCode: 200 })
  expect(fetchMock).toHaveBeenCalledTimes(2)
})
```

---

llun (commented):
Both issues from round 4 are fixed — 413 is removed from `RETRYABLE_STATUS_CODES`, there is an explicit "does not retry 413" test, and an `it.each` now covers every remaining retryable status code.

The new `Retry-After` support is clean: delta-seconds and HTTP-date are both handled, the timeout guard (`retryAfterDelay > timeoutInMilliseconds → don't retry`) is correct, and the past-date edge case (`Math.max(1, …)`) is sound. One minor gap: only the delta-seconds path is tested; the `Date.parse` branch has no test case. See inline comment — it is non-blocking but worth adding before merge.

The overall security hardening is solid and the review history shows systematic resolution of every issue raised. Ready to merge once the HTTP-date test is added (or accepted as-is).

---

lib/utils/request.test.ts
llun: Fixed in df7c65cd. Added coverage for Retry-After HTTP-date headers with a fixed fake timer clock so the Date.parse path is pinned alongside the delta-seconds case.

PR #686, opened 2026-05-14
Bump version to v2.0.0

llun:
Automated version bump to v2.0.0.


llun:
This shouldn't happened.

PR #690, opened 2026-05-14
fix: make version bump workflow resilient

llun:
## Summary
- Debounce `main` push handling for five minutes and skip stale workflow runs when `origin/main` has moved.
- Calculate the next version from the latest reachable tag and the highest unreleased bump prefix across the combined commit batch.
- Use one stable `version-bump/main` release PR branch, close stale `version-bump/v*` PRs, and re-check `main` before merging while keeping `tag-version.yml` unchanged.

## Test Plan
- Parsed `.github/workflows/version-bump.yml` as YAML and ran `bash -n` against every embedded shell block.
- Simulated patch, minor, major, stale SHA, `.github/`-only, and `none:` release-bump cases with the extracted workflow script.
- `yarn lint` (0 errors, existing warnings only)
- `yarn build`
- `yarn test` (rerun passed after one parallel Jest worker SIGSEGV on the first full run; the failed suite passed in-band before the full rerun).

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -5,19 +5,65 @@ on:
     branches:
       - main
 
-concurrency:
-  group: version-bump
-  cancel-in-progress: false
-
 permissions:
   actions: write
   contents: write
   pull-requests: write
 
+env:
+  DEBOUNCE_SECONDS: 300
+  VERSION_BUMP_BRANCH: version-bump/main
+
 jobs:
+  coalesce:
+    name: Wait for merge batch
+    runs-on: ubuntu-latest
+    outputs:
+      latest_sha: ${{ steps.latest-main.outputs.latest_sha }}
+      should_run: ${{ steps.latest-main.outputs.should_run }}
+    steps:
+      - name: Wait for nearby main merges
+        run: sleep "$DEBOUNCE_SECONDS"
+
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check latest main
+        id: latest-main
+        run: |
+          set -euo pipefail
+
+          git fetch --force origin main '+refs/tags/*:refs/tags/*'
+
+          LATEST_SHA=$(git rev-parse origin/main)
+          echo "latest_sha=$LATEST_SHA" >> "$GITHUB_OUTPUT"
+
+          if [ "$GITHUB_SHA" != "$LATEST_SHA" ]; then
+            echo "Skipping: $GITHUB_SHA is no longer the latest main commit ($LATEST_SHA)."
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          HEAD_MSG=$(git log -1 --pretty=format:"%s" "$LATEST_SHA")
+          if echo "$HEAD_MSG" | grep -qE '^Bump version to v[0-9]+\.[0-9]+\.[0-9]+'; then
+            echo "Skipping: latest main commit is already a version bump."
+            echo "should_run=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          echo "should_run=true" >> "$GITHUB_OUTPUT"
+
   version-bump:
     name: Bump Version
+    needs: coalesce
+    if: needs.coalesce.outputs.should_run == 'true'
     runs-on: ubuntu-latest
+    concurrency:
+      group: version-bump-main
+      cancel-in-progress: false
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/version-bump.yml
```diff
@@ -175,54 +231,119 @@ jobs:
...
+            exit 0
           fi
 
-          PR_URL=$(gh pr create \
-            --title "Bump version to v${NEW_VERSION}" \
-            --body "Automated version bump to v${NEW_VERSION}." \
-            --base main \
-            --head "$BRANCH")
+          git commit -m "Bump version to v${NEW_VERSION}"
+          git push --force-with-lease origin "$VERSION_BUMP_BRANCH"
```
llun: **P1 — `--force-with-lease` will fail on every reuse of the stable branch.**

This job only ever fetches `main` and tags (`git fetch --force origin main '+refs/tags/*:refs/tags/*'`); it never fetches `version-bump/main`. The branch is then created fresh each run with `git switch -C`, so there is no `refs/remotes/origin/version-bump/main` remote-tracking ref. Bare `--force-with-lease` derives its expected value from that missing ref, i.e. it expects to *create* the branch. On every run after the first — exactly the "reuse one stable branch" case this PR is built around — the remote branch already exists and the push is rejected with stale-info, failing the job under `set -euo pipefail`.

Fix: `git fetch origin "$VERSION_BUMP_BRANCH"` before the push (and base the local branch / lease on it), or use a plain `--force` here since the branch is intentionally disposable.

---

.github/workflows/version-bump.yml
```diff
@@ -175,54 +231,119 @@ jobs:
...
+
+          if [ "$LATEST_MAIN_SHA" != "$BASE_SHA" ]; then
+            echo "Leaving $PR_URL open: main moved from $BASE_SHA to $LATEST_MAIN_SHA while checks were running."
+            exit 0
+          fi
 
-          echo "Created PR: $PR_URL"
-          gh pr merge --auto --squash "$PR_URL"
-          echo "Enabled auto-merge for $PR_URL"
+          gh pr merge --squash --delete-branch "$PR_URL"
```
llun: **P2 — `gh pr merge --squash` without `--auto` regresses the auto-merge behaviour.**

The previous workflow used `gh pr merge --auto`, which tolerates a PR that is not yet immediately mergeable. This immediate merge runs under `set -euo pipefail`, so if branch protection requires a review approval, or required status contexts haven't yet registered as successful at the exact moment of this call, `gh pr merge` exits non-zero and fails the whole job.

Consider keeping `--auto` (let GitHub merge once requirements are met) or explicitly tolerating a non-zero exit here.

---

.github/workflows/version-bump.yml
```diff
@@ -5,19 +5,65 @@ on:
...
 jobs:
+  coalesce:
+    name: Wait for merge batch
+    runs-on: ubuntu-latest
+    outputs:
+      latest_sha: ${{ steps.latest-main.outputs.latest_sha }}
+      should_run: ${{ steps.latest-main.outputs.should_run }}
+    steps:
+      - name: Wait for nearby main merges
+        run: sleep "$DEBOUNCE_SECONDS"
```
llun, line 28: **P2 — `coalesce` has no concurrency control and unconditionally burns ~5 runner-minutes per push.**

Every push to `main` spawns a `coalesce` job that `sleep`s for `DEBOUNCE_SECONDS` (300s) with no `concurrency:` group, so N pushes within a window hold N runners idle for 5 minutes each. The `version-bump` job got a `version-bump-main` concurrency group, but `coalesce` did not, so overlapping debounce windows don't collapse.

Consider adding a `concurrency` group to `coalesce` (so only the latest debounce window survives) or otherwise bounding the idle-runner cost.

---

.github/workflows/version-bump.yml
```diff
@@ -5,19 +5,65 @@ on:
...
+env:
+  DEBOUNCE_SECONDS: 300
+  VERSION_BUMP_BRANCH: version-bump/main
+
 jobs:
+  coalesce:
+    name: Wait for merge batch
+    runs-on: ubuntu-latest
+    outputs:
+      latest_sha: ${{ steps.latest-main.outputs.latest_sha }}
```
llun: **Nit — `latest_sha` output is unused.**

`coalesce` exports `latest_sha`, but `version-bump` recomputes `BASE_SHA` from `origin/main` itself and never reads `needs.coalesce.outputs.latest_sha`. Either drop the unused output or have `version-bump` consume it for consistency.

---

llun (commented):
Reviewed the `version-bump.yml` rewrite. The debounce + stable-branch + batch-bump direction is sound, but a few things need fixing before this is reliable:

- **P1:** `git push --force-with-lease` will fail on every run after the first because the stable branch is never fetched (no remote-tracking ref to lease against). See inline comment.
- **P2:** `gh pr merge --squash` dropped `--auto`, so the job hard-fails if the PR isn't immediately mergeable.
- **P2:** the `coalesce` job has no concurrency group and sleeps 5 min per push, holding a runner idle each time.
- **Nit:** `coalesce` output `latest_sha` is unused.

I also agree with the two existing bot comments: gate `gh pr checks --watch` on `--required`, and don't fail on the missing `PAT_TOKEN` before the tag-only fallback path that doesn't need it.

---

.github/workflows/version-bump.yml
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. The `coalesce` job now has job-level concurrency keyed by `github.ref` with `cancel-in-progress: true`, so superseded debounce sleeps are cancelled instead of burning runner minutes.

.github/workflows/version-bump.yml
```diff
@@ -175,54 +231,119 @@ jobs:
...
+          else
+            PR_URL=$(gh pr create \
+              --title "Bump version to v${NEW_VERSION}" \
+              --body-file "$BODY_FILE" \
+              --base main \
+              --head "$VERSION_BUMP_BRANCH")
+            echo "Created PR: $PR_URL"
+          fi
+
+          if ! gh pr checks "$PR_URL" --watch --interval 10; then
```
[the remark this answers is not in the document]
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. The workflow now waits on `gh pr checks --required --watch`, so optional checks no longer block automated version bump PRs.

.github/workflows/version-bump.yml
```diff
@@ -175,54 +231,119 @@ jobs:
...
-          # If an open PR already exists for this version, ensure auto-merge
-          # is enabled (covers reruns where PR was created but auto-merge failed).
-          EXISTING_PR=$(gh pr list --state open --head "$BRANCH" --json number --jq '.[0].number')
-          if [ -n "$EXISTING_PR" ]; then
-            echo "PR #${EXISTING_PR} already exists; ensuring auto-merge is enabled."
-            gh pr merge --auto --squash "#${EXISTING_PR}"
+          if [ -z "${GH_TOKEN:-}" ]; then
+            echo "::error::PAT_TOKEN is required to push the bump branch and manage the PR."
+            exit 1
+          fi
```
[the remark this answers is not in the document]
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. The `PAT_TOKEN` guard now runs after the no-change direct-tag fallback, so bootstrap/direct-tag cases can still use the checkout token path without requiring `gh`.

.github/workflows/version-bump.yml
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. The workflow now fetches `refs/heads/${VERSION_BUMP_BRANCH}` when it exists and pushes with an explicit `--force-with-lease` expected SHA for deterministic stable-branch reuse.

.github/workflows/version-bump.yml
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. The merge step now uses `gh pr merge --auto --squash --delete-branch`, restoring the previous auto-merge behavior after the final freshness check.

.github/workflows/version-bump.yml
llun: Fixed in 5880cd48aeca444e1d4b17cfcc934c8fdc103e03. Removed the unused `latest_sha` job output and the corresponding `GITHUB_OUTPUT` write; `version-bump` continues to recompute `origin/main` for freshness.

llun:
@codex review

llun (commented):
Second review round on `5880cd4` — all six previously raised threads are correctly addressed:

- **force-with-lease**: now `ls-remote`-checks the branch, fetches `refs/heads/version-bump/main` when present, and pushes with an explicit `--force-with-lease=<ref>:<expected-sha>`. Deterministic and safe for stable-branch reuse.
- **merge step**: restored to `gh pr merge --auto --squash --delete-branch` after the final freshness check.
- **coalesce concurrency**: job-level `concurrency` keyed by `github.ref` with `cancel-in-progress: true` — superseded debounce sleeps are now cancelled instead of holding runners.
- **`--required` checks**: `gh pr checks --required --watch` no longer blocks on optional/informational checks.
- **PAT guard**: moved below the no-change direct-tag fallback, so the bootstrap/tag-only path still works with just the checkout token.
- **unused output**: `latest_sha` removed.

No new issues spotted in the updated workflow. LGTM.

---

PR #691, opened 2026-05-15
fix: sign presigned media upload headers

llun:
## Summary

Fixes image post creation through S3-backed presigned media uploads by signing the checksum headers that the browser sends during the PUT request.

## Root Cause

The media API returned a presigned S3 URL plus `x-amz-checksum-sha1` and `x-amz-meta-checksumsha1` headers for the browser upload. The URL did not keep those headers in the signed header set, so S3 rejected the PUT with `AccessDenied` and `HeadersNotSigned`, causing the post composer to fail before the post could be created.

## Changes

- Mark checksum headers as `unhoistableHeaders` when generating presigned upload URLs.
- Add a regression test that verifies the S3 presigner receives those required header names.

## Verification

- `yarn lint` (passes with existing warnings in auth/fitness files)
- `yarn build`
- `yarn test`

### DIFF
--- a/lib/services/medias/S3StorageFile.test.ts
+++ b/lib/services/medias/S3StorageFile.test.ts
@@ -1,8 +1,10 @@
 import {
   DeleteObjectCommand,
   HeadObjectCommand,
+  PutObjectCommand,
   S3Client
 } from '@aws-sdk/client-s3'
+import { getSignedUrl } from '@aws-sdk/s3-request-presigner'
 
 import { MediaStorageType } from '@/lib/config/mediaStorage'
 import { Database } from '@/lib/database/types'
@@ -25,6 +27,10 @@ jest.mock('@aws-sdk/client-s3', () => {
   }
 })
 
+jest.mock('@aws-sdk/s3-request-presigner', () => ({
+  getSignedUrl: jest.fn().mockResolvedValue('https://storage.example/upload')
+}))
+
 describe('S3FileStorage presigned upload completion', () => {
   const send = jest.fn()
   const actor = {
@@ -35,7 +41,11 @@ describe('S3FileStorage presigned upload completion', () => {
   const checksumBase64 = Buffer.from(checksumHex, 'hex').toString('base64')
 
   const database = {
+    createMedia: jest.fn(),
+    getActorFromId: jest.fn(),
+    getFitnessStorageUsageForAccount: jest.fn(),
     getMediaByIdForAccount: jest.fn(),
+    getStorageUsageForAccount: jest.fn(),
     markMediaUploadVerified: jest.fn(),
     deleteMedia: jest.fn()
   } as unknown as jest.Mocked<Database>
@@ -45,6 +55,23 @@ describe('S3FileStorage presigned upload completion', () => {
     ;(S3Client as jest.MockedClass<typeof S3Client>).mockImplementation(
       () => ({ send }) as unknown as S3Client
     )
+    database.createMedia.mockResolvedValue({
+      id: 'media-1',
+      actorId: 'actor-1',
+      original: {
+        path: 'medias/2026-01-01/upload.png',
+        bytes: 1024,
+        mimeType: 'image/png',
+        metaData: {
+          width: 10,
+          height: 10
+        },
+        fileName: 'upload.png'
+      }
+    } as never)
+    database.getActorFromId.mockResolvedValue(actor)
+    database.getStorageUsageForAccount.mockResolvedValue(0)
+    database.getFitnessStorageUsageForAccount.mockResolvedValue(0)
     database.getMediaByIdForAccount.mockResolvedValue({
       id: 'media-1',
       actorId: 'actor-1',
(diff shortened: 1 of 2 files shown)


lib/services/medias/S3StorageFile.ts
```diff
@@ -77,6 +77,11 @@ const getObjectMetadataChecksumSha1 = (
   metadata?.['checksum-sha1'] ??
   null
 
+const PRESIGNED_UPLOAD_SIGNED_AMZ_HEADERS = new Set([
+  'x-amz-checksum-sha1',
+  'x-amz-meta-checksumsha1'
```
llun, line 82: The constant name `PRESIGNED_UPLOAD_SIGNED_AMZ_HEADERS` is misleading. These headers are passed to `unhoistableHeaders`, meaning they are prevented from being hoisted out of the request headers into the URL query string — they are not simply "signed AMZ headers" in a general sense.

Suggest renaming to `PRESIGNED_UPLOAD_UNHOISTABLE_HEADERS` so the name reflects the role it plays in the presigner options:

```ts
const PRESIGNED_UPLOAD_UNHOISTABLE_HEADERS = new Set([
  'x-amz-checksum-sha1',
  'x-amz-meta-checksumsha1'
])
```

---

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -91,6 +118,47 @@ describe('S3FileStorage presigned upload completion', () => {
...
+      size: 1024
+    })
+
+    expect(PutObjectCommand).toHaveBeenCalledWith(
+      expect.objectContaining({
+        ChecksumSHA1: checksumBase64,
+        Metadata: {
+          checksumSha1: checksumHex
+        }
+      })
```
llun, line 148: The assertion order here is fragile. `mock.calls[0][2]` is accessed on line 147 **before** asserting the call count on line 148. If `getSignedUrl` was never called, `mock.calls[0]` is `undefined` and accessing `[2]` throws a confusing `TypeError` rather than a clear jest assertion failure.

Swap the two lines so the count is asserted first:

```ts
expect(getSignedUrl).toHaveBeenCalledTimes(1)
const presignOptions = (getSignedUrl as jest.Mock).mock.calls[0][2]
```

---

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -91,6 +118,47 @@ describe('S3FileStorage presigned upload completion', () => {
...
+    )
+    const presignOptions = (getSignedUrl as jest.Mock).mock.calls[0][2]
```
llun: This assertion is overly prescriptive. The fix being tested is about `unhoistableHeaders`; asserting that `signableHeaders` is `undefined` couples the test to an unrelated implementation detail. If `signableHeaders` were added later for a different reason, this test would fail even though the actual fix is still correct.

Remove this line — the test already validates the `unhoistableHeaders` are set correctly on the lines below.

---

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -35,7 +41,11 @@ describe('S3FileStorage presigned upload completion', () => {
   const checksumBase64 = Buffer.from(checksumHex, 'hex').toString('base64')
 
   const database = {
+    createMedia: jest.fn(),
```
llun, line 44: `getActorFromId` and `getFitnessStorageUsageForAccount` don't appear to be called by `getPresigedForSaveFileUrl` (which only calls `createMedia` after the presign step). If these aren't exercised by any test in the describe block, they're dead setup that inflates the mock surface and makes it harder to see what each test actually depends on.

Please verify whether these are needed. If not, remove them from the `database` mock object and from the `beforeEach` setup to keep the test focused.

---

llun (commented):
## Review: fix: sign presigned media upload headers

The root cause diagnosis and fix are correct — marking `x-amz-checksum-sha1` and `x-amz-meta-checksumsha1` as `unhoistableHeaders` is the right approach to stop the AWS presigner from moving them to the query string, which was causing S3 to reject the browser PUT with `AccessDenied / HeadersNotSigned`. The change itself is minimal and well-targeted.

Four things to address before merging:

1. **Misleading constant name** (`S3StorageFile.ts`) — `PRESIGNED_UPLOAD_SIGNED_AMZ_HEADERS` should be `PRESIGNED_UPLOAD_UNHOISTABLE_HEADERS` to match how it is actually used.
2. **Fragile test assertion order** (`S3StorageFile.test.ts`) — `mock.calls[0][2]` is dereferenced before the call-count assertion, so a missing call produces a `TypeError` rather than a useful jest failure.
3. **Overly prescriptive assertion** (`S3StorageFile.test.ts`) — asserting `signableHeaders` is `undefined` couples the test to an unrelated option and will break false-positively if that option is ever set for a different reason.
4. **Unnecessary mock methods** (`S3StorageFile.test.ts`) — `getActorFromId` and `getFitnessStorageUsageForAccount` don't appear to be called by `getPresigedForSaveFileUrl`; if confirmed, remove them to keep the mock surface minimal.

See inline comments for specifics. Also note the branch is currently behind `main` and will need a rebase before it can be merged.

---

lib/services/medias/S3StorageFile.ts
llun: Fixed in ccec0693. Renamed the constant to PRESIGNED_UPLOAD_UNHOISTABLE_HEADERS and kept it wired to getSignedUrl unhoistableHeaders.

lib/services/medias/S3StorageFile.test.ts
llun: Fixed in ccec0693. Removed the signableHeaders assertion so this test only constrains the unhoistable checksum headers relevant to the fix.

lib/services/medias/S3StorageFile.test.ts
llun: Fixed in ccec0693. The test now asserts getSignedUrl was called before reading mock.calls[0][2].

lib/services/medias/S3StorageFile.test.ts
llun: I checked this path and kept these mocks. getPresigedForSaveFileUrl calls checkQuotaAvailable before createMedia, and checkQuotaAvailable calls getActorFromId, getStorageUsageForAccount, and getFitnessStorageUsageForAccount. Removing them would make this describe block fail before the presign and create-media assertions run.

llun:
@codex review

llun (commented):
## Round 2 review

All four issues from round 1 are resolved:

- ✅ Constant renamed to `PRESIGNED_UPLOAD_UNHOISTABLE_HEADERS`
- ✅ Assertion order fixed (`toHaveBeenCalledTimes(1)` before `mock.calls[0][2]`)
- ✅ Overly prescriptive `signableHeaders` assertion removed
- ✅ Extra database mocks kept — the explanation that `checkQuotaAvailable` calls `getActorFromId` / `getStorageUsageForAccount` / `getFitnessStorageUsageForAccount` before `createMedia` is correct; those mocks are necessary

The new `result.toMatchObject({ url, headers: {...} })` assertion is a good addition that validates the end-to-end contract (confirmed that `PresignedUrlOutput` does include a `headers` field at source lines 261–264).

Two non-blocking items before merging:

1. **Branch is still behind `main`** — rebase or merge main in before merging.
2. **PR is still draft** — mark it ready when you're happy with it.

The fix itself is correct and the tests cover what matters. LGTM once the branch is up to date.

---

llun:
Addressed the round 2 non-blocking items. Merged origin/main in 69097586 so the branch includes ec14e16c from main, and marked the PR ready for review. No additional code changes were needed; the merge only brought in .github/workflows/version-bump.yml from main and did not alter the presigned upload fix files.

llun (commented):
## Round 3 review

No code changes since round 2 — this is a clean rebase + ready-for-review promotion.

Both remaining items from round 2 are now resolved:

- ✅ Branch is up to date with main (`mergeable_state: clean`)
- ✅ PR is no longer a draft

Code is correct, test coverage is solid, and all prior review issues are addressed. Good to merge.

---

PR #692, opened 2026-05-15
fix: sign redirected ActivityPub key fetches

llun:
## Summary

This fixes an inbox verification regression introduced after PR #675 hardened outbound ActivityPub fetches.

Inbound inbox POSTs verify the sender signature by fetching the sender public key. After #675, `safeRemoteFetch` follows redirects internally while reusing the original request headers. For signed ActivityPub key fetches, that meant a redirected fetch could carry a `Signature` header whose `(request-target)` was still the original key URL instead of the redirected URL. Remote servers that validate signed fetches can reject that redirected key lookup, leaving the inbox guard without a usable sender public key and causing 400 responses. If key ownership resolution falls back to or resolves a different owner than the activity actor, the guard can also reject with 403.

## Fix

- Allow `safeRemoteFetch`/`request` callers to provide a header builder function that is evaluated for each effective URL.
- Use that dynamic header builder for `getSenderPublicKey`, so ActivityPub public-key fetches are signed for the actual redirected URL.
- Preserve the existing cross-host redirect protections by still stripping sensitive dynamic headers when the redirect changes hosts.
- Add regression coverage for redirected signed public-key fetches and dynamic-header cross-host stripping.

## Validation

- `yarn test lib/utils/safeRemoteFetch.test.ts lib/utils/request.test.ts lib/services/guards/getSenderPublicKey.test.ts --runInBand`
- `yarn lint` (passes with existing warnings in unrelated files)
- `yarn build`
- `yarn test`

### DIFF
--- a/lib/services/guards/getSenderPublicKey.test.ts
+++ b/lib/services/guards/getSenderPublicKey.test.ts
@@ -1,4 +1,5 @@
 import fetchMock, { enableFetchMocks } from 'jest-fetch-mock'
+import crypto from 'node:crypto'
 
 import { getTestSQLDatabase } from '@/lib/database/testUtils'
 import {
@@ -10,6 +11,7 @@ import { TEST_DOMAIN } from '@/lib/stub/const'
 import { seedDatabase } from '@/lib/stub/database'
 import { seedActor1 } from '@/lib/stub/seed/actor1'
 import { logger } from '@/lib/utils/logger'
+import { parse } from '@/lib/utils/signature'
 
 enableFetchMocks()
 
@@ -35,6 +37,7 @@ jest.mock('@/lib/utils/logger', () => ({
 }))
 
 jest.mock('@/lib/utils/trace', () => ({
+  getSpan: () => mockSpan,
   getTracer: () => ({
     startActiveSpan: mockStartActiveSpan
   })
@@ -62,6 +65,31 @@ const createActorDocument = ({
     publicKeyPem
   }
 })
+
+const verifySignedRequestTarget = async ({
+  headers,
+  publicKey,
+  requestTarget
+}: {
+  headers: Record<string, string>
+  publicKey: string
+  requestTarget: string
+}) => {
+  const parsedSignature = await parse(headers.signature)
+  const signedString = (parsedSignature.headers ?? '')
+    .split(' ')
+    .map((header) => {
+      if (header === '(request-target)') {
+        return `(request-target): ${requestTarget}`
+      }
+
+      return `${header}: ${headers[header]}`
+    })
+    .join('\n')
+  const verifier = crypto.createVerify(parsedSignature.algorithm)
+  verifier.update(signedString)
+  return verifier.verify(publicKey, parsedSignature.signature, 'base64')
+}
 
 describe('getSenderPublicKey', () => {
   const database = getTestSQLDatabase()
@@ -611,6 +639,60 @@ describe('getSenderPublicKey', () => {
       host: 'remote.test',
       signature: expect.stringContaining('headers="(request-target) host date"')
     })
+  })
+
+  it('signs redirected remote public key fetches for the redirect target', async () => {
+    const actorId = 'https://remote.test/users/redirected-key'
+    const keyId = `${actorId}#main-key`
+    const redirectTarget = 'https://remote.test/@redirected-key'
+    fetchMock.resetMocks()
+    fetchMock.mockResponse(async (request) => {
+      const url = new URL(request.url)
+
+      if (url.pathname === '/users/redirected-key') {
+        return {
+          headers: { location: redirectTarget },
+          status: 302
+        }
+      }
+
+      if (url.pathname === '/@redirected-key') {
+        return {
+          body: JSON.stringify(
+            createActorDocument({
+              id: actorId,
+              publicKeyId: keyId,
+              publicKeyPem: 'redirected-public-key'
+            })
+          ),
+          status: 200
+        }
+      }
+
+      return { status: 404 }
+    })
+
+    const publicKey = await getSenderPublicKeyDetails(database, keyId)
+    const signingActor = await database.getFederationSigningActor()
+    if (!signingActor) fail('Federation signing actor is required')
+
+    const redirectedRequest = fetchMock.mock.calls.at(1)?.[1]
+    const redirectedHeaders = redirectedRequest?.headers as Record<
+      string,
+      string
+    >
+
+    expect(publicKey).toEqual({
+      owner: actorId,
+      publicKey: 'redirected-public-key'
+    })
+    expect(
+      await verifySignedRequestTarget({
+        headers: redirectedHeaders,
+        publicKey: signingActor.publicKey,
+        requestTarget: 'get /@redirected-key'
+      })
+    ).toBe(true)
   })
 
   it('returns empty string when remote actor not found', async () => {
--- a/lib/services/guards/getSenderPublicKey.test.ts
+++ b/lib/services/guards/getSenderPublicKey.test.ts
@@ -676,7 +676,11 @@ describe('getSenderPublicKey', () => {
     const signingActor = await database.getFederationSigningActor()
     if (!signingActor) fail('Federation signing actor is required')
 
-    const redirectedRequest = fetchMock.mock.calls.at(1)?.[1]
+    const redirectedCall = fetchMock.mock.calls.find(
+      ([url]) => url === redirectTarget
+    )
+    expect(redirectedCall).toBeDefined()
+    const redirectedRequest = redirectedCall?.[1]
     const redirectedHeaders = redirectedRequest?.headers as Record<
       string,
       string
--- a/lib/services/guards/getSenderPublicKey.test.ts
+++ b/lib/services/guards/getSenderPublicKey.test.ts
@@ -37,7 +37,6 @@ jest.mock('@/lib/utils/logger', () => ({
 }))
 
 jest.mock('@/lib/utils/trace', () => ({
-  getSpan: () => mockSpan,
   getTracer: () => ({
     startActiveSpan: mockStartActiveSpan
   })
--- a/lib/utils/safeRemoteFetch.test.ts
+++ b/lib/utils/safeRemoteFetch.test.ts
@@ -350,6 +350,77 @@ describe('safeRemoteFetch', () => {
     })
   })
 
+  it('keeps dynamic credentials stripped after cross-host redirect chains', async () => {
+    const seenRequests: Array<{
+      url: string
+      headers: Record<string, string | string[] | undefined>
+    }> = []
+    const safeRemoteFetch = createSafeRemoteFetch({
+      resolveHost: async () => [SAFE_ADDRESS],
+      transport: async ({ headers, url }) => {
+        seenRequests.push({ headers, url: url.toString() })
+        if (url.hostname === 'safe.example') {
+          return {
+            statusCode: 302,
+            headers: {
+              location: 'https://other.example/redirected'
+            },
+            body: streamFrom([])
+          }
+        }
+
+        if (url.pathname === '/redirected') {
+          return {
+            statusCode: 302,
+            headers: {
+              location: 'https://other.example/final'
+            },
+            body: streamFrom([])
+          }
+        }
+
+        return okResponse()
+      }
+    })
+
+    await safeRemoteFetch({
+      url: 'https://safe.example/actor',
+      headers: ({ url }) => ({
+        authorization: 'Bearer secret',
+        cookie: 'session=secret',
+        signature: `keyId="${url.toString()}",signature="secret"`
+      })
+    })
+
+    expect(seenRequests).toEqual([
+      {
+        url: 'https://safe.example/actor',
+        headers: expect.objectContaining({
+          authorization: 'Bearer secret',
+          cookie: 'session=secret',
+          host: 'safe.example',
+          signature: 'keyId="https://safe.example/actor",signature="secret"'
+        })
+      },
+      {
+        url: 'https://other.example/redirected',
+        headers: expect.not.objectContaining({
+          authorization: expect.any(String),
+          cookie: expect.any(String),
+          signature: expect.any(String)
+        })
+      },
+      {
+        url: 'https://other.example/final',
+        headers: expect.not.objectContaining({
+          authorization: expect.any(String),
+          cookie: expect.any(String),
+          signature: expect.any(String)
+        })
+      }
+    ])
+  })
+
   it('destroys redirect response bodies without buffering them', async () => {
     const redirectBody = streamFrom(['redirect body'])
     const destroy = jest.spyOn(redirectBody, 'destroy')
--- a/lib/utils/safeRemoteFetch.test.ts
+++ b/lib/utils/safeRemoteFetch.test.ts
@@ -308,6 +308,48 @@ describe('safeRemoteFetch', () => {
     })
   })
 
+  it('strips dynamic credentials and auth headers on cross-host redirects', async () => {
+    const seenRequests: Array<{
+      url: string
+      headers: Record<string, string | string[] | undefined>
+    }> = []
+    const safeRemoteFetch = createSafeRemoteFetch({
+      resolveHost: async () => [SAFE_ADDRESS],
+      transport: async ({ headers, url }) => {
+        seenRequests.push({ headers, url: url.toString() })
+        if (url.hostname === 'safe.example') {
+          return {
+            statusCode: 302,
+            headers: {
+              location: 'https://other.example/private-key'
+            },
+            body: streamFrom([])
+          }
+        }
+
+        return okResponse()
+      }
+    })
+
+    await safeRemoteFetch({
+      url: 'https://safe.example/actor',
+      headers: ({ url }) => ({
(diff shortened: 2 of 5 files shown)


lib/utils/safeRemoteFetch.ts
```diff
@@ -665,9 +703,13 @@ export const createSafeRemoteFetch = ({
...
-        currentHeaders = stripBodyHeaders(requestHeaders)
+        currentHeaders =
+          typeof currentHeaders === 'function'
+            ? currentHeaders
+            : stripBodyHeaders(requestHeaders)
       } else {
-        currentHeaders = requestHeaders
+        currentHeaders =
+          typeof currentHeaders === 'function' ? currentHeaders : requestHeaders
       }
```
llun, line 756: When `currentHeaders` is a function and a 303 response is received, `stripBodyHeaders` is never applied. For static headers the framework defensively strips body-specific headers (`content-type`, `content-encoding`, `content-length`, `digest`, etc.) because they no longer describe any body after the POST→GET conversion. The function path skips that stripping entirely.

The builder does receive `method: 'GET'` and `body: undefined` on the next hop, so a well-written builder won't emit body headers, but that relies on caller discipline rather than enforcement. The static-headers path enforces it unconditionally.

Note that `signature` is in `BODY_REDIRECT_HEADERS` too, but for the function case a fresh signature is intentionally generated by the builder for the new URL — stripping it here would defeat the purpose of this PR. So the fix isn't simply `stripBodyHeaders(headers(...))`. One approach: wrap the function to strip only the non-signature body headers on 303:

```ts
const BODY_ONLY_REDIRECT_HEADERS = new Set([
  'content-digest', 'content-encoding', 'content-language',
  'content-length', 'content-location', 'content-type', 'digest'
])
// then on 303:
currentHeaders = (req: SafeRemoteFetchHeaderBuilderRequest) => {
  const built = (currentHeaders as SafeRemoteFetchHeaderBuilder)(req)
  return stripHeaders(built, BODY_ONLY_REDIRECT_HEADERS)
}
```

Or document explicitly that the caller's builder is responsible for not emitting body headers on non-POST methods, since it already receives `method` in the request context.

---

lib/utils/safeRemoteFetch.test.ts
llun, line 1: There's no test for a 303 redirect with a function header builder. The existing 303 test (`'converts POST redirects to GET for 303 responses'`) only exercises the static-headers path. A complementary case should verify:

1. The function is re-invoked on the follow-up hop with `method: 'GET'` and `body: undefined` (so the builder can conditionally omit body headers).
2. Body-specific headers that a builder may emit unconditionally (`content-type`, `digest`) are **not** forwarded to the GET request — or at minimum, the test documents what the current contract is.

This gap is especially worth closing given that the 303 path has a meaningful behavioral divergence from the static-headers path (see comment on `safeRemoteFetch.ts`).

---

lib/services/guards/getSenderPublicKey.test.ts
```diff
@@ -613,6 +641,60 @@ describe('getSenderPublicKey', () => {
...
+          ),
+          status: 200
+        }
+      }
+
+      return { status: 404 }
+    })
+
+    const publicKey = await getSenderPublicKeyDetails(database, keyId)
+    const signingActor = await database.getFederationSigningActor()
```
llun, line 675: `fetchMock.mock.calls.at(1)` is fragile — it assumes the redirected request is always the second entry in the global calls array. If call ordering shifts (e.g., due to tracing internals, retry logic, or test setup changes) the assertion silently operates on the wrong request.

The existing test `'signs remote public key fetches with a local actor'` uses URL-based lookup, which is much more robust:

```ts
const redirectedCall = fetchMock.mock.calls.find(([url]) => url === redirectTarget)
const redirectedHeaders = redirectedCall?.[1]?.headers as Record<string, string>
```

If the redirected call is not found at all, the assertion on `redirectedHeaders.signature` will throw an obvious error rather than quietly verifying the wrong request's headers.

---

lib/services/guards/getSenderPublicKey.test.ts
```diff
@@ -35,6 +37,7 @@ jest.mock('@/lib/utils/logger', () => ({
 }))
 
 jest.mock('@/lib/utils/trace', () => ({
```
llun, line 39: `getSpan` is not imported or called anywhere in `getSenderPublicKey.ts` — only `getTracer` is used. If no code path reachable from this test suite calls `getSpan`, this mock entry is dead code and should be removed. If something in the call chain (e.g., `activityPubRequestHeaders` or a tracing helper) does call it, please add a brief comment explaining why it's needed here.

---

llun (commented):
The core idea is correct and the same-host signed redirect case is well-handled. Four things to address before merging:

**Must fix**
1. **`safeRemoteFetch.ts` – 303 redirect skips `stripBodyHeaders` for function header builders.** The static-headers path strips body-specific headers (`content-type`, `content-encoding`, `digest`, etc.) on 303 POST→GET conversions. The function path bypasses this entirely. The builder does receive `method: 'GET'` so a careful caller is safe, but the framework should enforce the invariant rather than rely on caller discipline — or the divergence should be explicitly documented and tested.

**Should fix**
2. **`safeRemoteFetch.test.ts` – No test for 303 + function header builder.** The 303-redirect test only covers static headers. A counterpart covering the function path is needed to pin and document the current contract (especially given the body-header stripping gap above).

3. **`getSenderPublicKey.test.ts` – Index-based `fetchMock.mock.calls.at(1)` is fragile.** Use URL-based lookup (`find(([url]) => url === redirectTarget)`) as the existing signing test already does.

4. **`getSenderPublicKey.test.ts` – `getSpan: () => mockSpan` in the trace mock appears unused.** `getSenderPublicKey.ts` only calls `getTracer`. Remove or explain.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -510,6 +518,34 @@ const buildHeaders = ({
...
+      previousUrl,
+      url
+    })
+  }
+
+  return buildHeaders({
+    headers,
+    previousUrl,
+    url
+  })
```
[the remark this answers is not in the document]
llun: Thanks. I left this unchanged because the suggested `effectiveHeaders` refactor is behaviorally equivalent and only changes style. Keeping the function/static branches explicit makes the dynamic header builder path easier to audit around redirects.

lib/utils/safeRemoteFetch.ts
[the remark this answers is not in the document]
llun: I kept `GET` uppercase here. `got` accepts uppercase methods, the rest of this helper already uses uppercase method values, and the ActivityPub signature helper lowercases only when constructing `(request-target)`. The real 303 function-header gap is fixed in 29c878be by stripping body-only headers while still allowing a regenerated dynamic signature.

llun: Fixed in 29c878be. Dynamic header builders are still re-invoked with the redirected URL, `method: "GET"`, and `body: undefined`, but the 303 path now strips body-only headers such as `content-type` and `digest` from the rebuilt request. `signature` is intentionally preserved for dynamic builders so same-host signed redirects can regenerate a valid signature for the target URL.

lib/utils/safeRemoteFetch.test.ts
llun: Fixed in 29c878be. Added a 303 dynamic-header test that verifies the builder is called again with `method: "GET"` and `body: undefined`, and that body-only headers from an unconditional builder are not sent on the follow-up GET request.

lib/services/guards/getSenderPublicKey.test.ts
llun: Fixed in d29f3f86. The redirected fetch assertion now finds the request by `redirectTarget` URL before reading headers, so the test no longer depends on global mock call ordering.

lib/services/guards/getSenderPublicKey.test.ts
llun: Fixed in 32f72b20. `getSpan` was dead mock setup for this test file, so I removed that mock entry instead of documenting unused setup.

llun:
@codex review

lib/utils/safeRemoteFetch.ts
```diff
@@ -24,10 +24,30 @@ const BODY_REDIRECT_HEADERS = new Set([
...
 ])
+const DYNAMIC_BODY_REDIRECT_HEADERS = new Set([
+  'content-digest',
+  'content-encoding',
+  'content-language',
+  'content-length',
+  'content-location',
+  'content-type',
+  'digest'
+])
```
llun: `DYNAMIC_BODY_REDIRECT_HEADERS` is exactly `BODY_REDIRECT_HEADERS` minus `signature`. Having both as independent literals means a future maintainer who adds a new body-related header to one set can silently miss the other.

Encode the relationship explicitly so they can never drift:

```ts
const BODY_REDIRECT_HEADERS = new Set([
  'content-digest',
  'content-encoding',
  'content-language',
  'content-length',
  'content-location',
  'content-type',
  'digest',
  'signature'
])
const DYNAMIC_BODY_REDIRECT_HEADERS = new Set(
  [...BODY_REDIRECT_HEADERS].filter((h) => h !== 'signature')
)
```

Or define `DYNAMIC_BODY_REDIRECT_HEADERS` first and derive `BODY_REDIRECT_HEADERS` from it — either direction works, as long as one is built from the other.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -665,9 +728,17 @@ export const createSafeRemoteFetch = ({
       if (response.statusCode === 303) {
         currentBody = undefined
         currentMethod = 'GET'
-        currentHeaders = stripBodyHeaders(requestHeaders)
+      }
+
+      if (typeof currentHeaders === 'function') {
+        if (response.statusCode === 303) {
+          currentHeaders = stripDynamicBodyHeaders(currentHeaders)
+        }
```
llun, line 750: Every 303 in a redirect chain re-wraps `currentHeaders` with another `stripDynamicBodyHeaders` call. A 303→302→303 sequence (within the 3-redirect cap) produces `stripDynamic(stripDynamic(original))`. It's functionally harmless — the stripping is idempotent — but the redundant wrapping is worth guarding against.

Simplest fix: only wrap if `currentHeaders` is the plain builder (not already wrapped). One lightweight approach is to check with a marker property:

```ts
const stripDynamicBodyHeaders = (
  headerBuilder: SafeRemoteFetchHeaderBuilder
): SafeRemoteFetchHeaderBuilder => {
  if ((headerBuilder as { __dynamicStripped?: boolean }).__dynamicStripped) {
    return headerBuilder
  }
  const wrapped = (request: SafeRemoteFetchHeaderBuilderRequest) =>
    stripHeaders(headerBuilder(request), DYNAMIC_BODY_REDIRECT_HEADERS)
  ;(wrapped as { __dynamicStripped?: boolean }).__dynamicStripped = true
  return wrapped
}
```

Alternatively, store a separate `isBodyStripped` boolean alongside `currentHeaders`. Not a blocker given the max-3-redirect cap, but a small cleanup worth doing.

---

llun (commented):
All four issues from the first round are addressed — the 303 body-stripping is fixed, the 303+function test is added, the index-based lookup is replaced with URL-based, and the dead `getSpan` mock is removed. Two small remaining items in `safeRemoteFetch.ts`:

1. **`DYNAMIC_BODY_REDIRECT_HEADERS` duplicates `BODY_REDIRECT_HEADERS` minus `signature`** — should be derived from it to prevent future drift.
2. **`stripDynamicBodyHeaders` accumulates wrappers on repeated 303s** — harmless given the 3-redirect cap, but a small guard would keep it clean.

---

lib/utils/safeRemoteFetch.ts
```diff
@@ -665,9 +728,17 @@ export const createSafeRemoteFetch = ({
...
       } else {
```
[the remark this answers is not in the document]
llun: Fixed in 69a60886. Dynamic builders now carry redirect strip state after a cross-host hop, so `authorization`, `cookie`, and `signature` stay stripped for the rest of that redirected chain. I also added a regression test for `safe.example -> other.example -> other.example` to cover the leak path.

lib/utils/safeRemoteFetch.ts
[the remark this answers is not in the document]
llun: Partially fixed in 69a60886. The recursive wrapper path is gone; dynamic stripping is now applied through explicit redirect strip state during header resolution. I intentionally kept `GET` uppercase because `safeRemoteFetch` uses uppercase method values in its public behavior and tests, `got` accepts uppercase methods, and ActivityPub signing lowercases only when constructing `(request-target)`.

lib/utils/safeRemoteFetch.ts
llun: Fixed in 69a60886. `DYNAMIC_BODY_REDIRECT_HEADERS` is now derived from `BODY_REDIRECT_HEADERS` by excluding `signature`, so the two sets cannot drift when body-header stripping changes later.

lib/utils/safeRemoteFetch.ts
llun: Fixed in 69a60886. The nested wrapper approach has been removed; dynamic header stripping now uses explicit booleans for body-header and sensitive-header stripping, applied once per request while resolving headers.

lib/utils/safeRemoteFetch.ts
```diff
@@ -570,15 +627,20 @@ const getRedirectLocation = (
...
+  stripBodyHeaders,
+  stripSensitiveHeaders
+}: {
+  stripBodyHeaders: boolean
+  stripSensitiveHeaders: boolean
+}) => {
+  return new Set([
+    ...(stripBodyHeaders ? DYNAMIC_BODY_REDIRECT_HEADERS : []),
+    ...(stripSensitiveHeaders ? SENSITIVE_REDIRECT_HEADERS : [])
+  ])
```
llun, line 643: The parameter `stripBodyHeaders: boolean` shadows the module-level `const stripBodyHeaders` function defined a few lines above. Inside this function body the two are unambiguous to the compiler, but to a reader scanning the file, seeing `stripBodyHeaders` used as a boolean condition looks like calling the function as a truthy check. It will also trip any `no-shadow` lint rule.

Rename the parameters to match the outer state variables (`shouldStripBodyHeaders`, `shouldStripSensitiveHeaders`) — or at least give them names that don't collide with existing identifiers:

```ts
const getDynamicHeadersToStrip = ({
  shouldStripBodyHeaders,
  shouldStripSensitiveHeaders
}: {
  shouldStripBodyHeaders: boolean
  shouldStripSensitiveHeaders: boolean
}) => {
  return new Set([
    ...(shouldStripBodyHeaders ? DYNAMIC_BODY_REDIRECT_HEADERS : []),
    ...(shouldStripSensitiveHeaders ? SENSITIVE_REDIRECT_HEADERS : [])
  ])
}
```

---

llun (commented):
Both round-2 issues resolved cleanly:
- `DYNAMIC_BODY_REDIRECT_HEADERS` derived from `BODY_REDIRECT_HEADERS` ✅
- Flag-based approach (`shouldStripDynamicBodyHeaders` / `shouldStripDynamicSensitiveHeaders`) replaces wrapper accumulation ✅
- New chain-redirect test covers the same-host-after-cross-host case ✅

One remaining minor issue: the `stripBodyHeaders` / `stripSensitiveHeaders` parameter names in `getDynamicHeadersToStrip` shadow the module-level `stripBodyHeaders` function, which is confusing to read and will trigger `no-shadow` linting.

---

lib/utils/safeRemoteFetch.ts
llun: Fixed in dc7879fe. The `getDynamicHeadersToStrip` parameters are now named `shouldStripBodyHeaders` and `shouldStripSensitiveHeaders`, so they no longer shadow the nearby `stripBodyHeaders` helper.

lib/utils/safeRemoteFetch.test.ts
```diff
@@ -308,6 +308,119 @@ describe('safeRemoteFetch', () => {
...
+      })
+    })
+
+    expect(seenRequests[1]).toEqual({
+      url: 'https://other.example/private-key',
+      headers: expect.not.objectContaining({
+        authorization: expect.any(String),
+        cookie: expect.any(String),
+        signature: expect.any(String)
+      })
```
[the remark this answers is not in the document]
llun: Fixed in 2e552bbd. The sensitive header checks now assert `authorization`, `cookie`, and `signature` are each absent individually, so a single leaked header will fail the test.

lib/utils/safeRemoteFetch.test.ts
```diff
@@ -308,6 +308,119 @@ describe('safeRemoteFetch', () => {
...
+      },
+      {
+        url: 'https://other.example/final',
+        headers: expect.not.objectContaining({
+          authorization: expect.any(String),
+          cookie: expect.any(String),
+          signature: expect.any(String)
+        })
+      }
+    ])
```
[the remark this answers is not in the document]
llun: Fixed in 2e552bbd. I replaced the chain test negative `objectContaining` checks with per-header absence assertions on each redirected request, so any individual sensitive header leak fails.

PR #695, opened 2026-05-15
fix: Fitness heatmap staging map

llun:
## Summary

- Move the fitness route map into its own full-width section above the calendar/cache grid.
- Fix the Mapbox container sizing so the canvas fills the full route-map area on desktop and mobile.
- Allow Docker builds to receive the public Mapbox token at build time so standalone staging output includes the Mapbox CSP entries.
- Tighten small accessibility details around icon/back controls, button types, pressed state, live retry errors, and focus rings.

## Why

The staging standalone build was reading the Mapbox token only at runtime, so generated headers did not include Mapbox CSP sources. After passing the public token during build, Mapbox scripts/styles/API calls are allowed. The Mapbox GL stylesheet also overrides the container positioning class, which collapsed the map container height; using explicit full height/width keeps the canvas sized to the intended map area.

## Validation

- `yarn test --runTestsByPath 'app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.test.tsx' 'lib/components/fitness/FitnessHeatmapList.test.tsx' 'next.config.test.ts' --runInBand --modulePathIgnorePatterns='<rootDir>/.claude'` - 48 passed
- `yarn lint` - 0 errors, 24 existing warnings
- `yarn node ./node_modules/next/dist/bin/next build`
- `git diff --check`
- Staging rebuild: `../activities.local/scripts/staging_run.sh --rebuild`
- Browser verification at `https://activities.local/@ride@llun.social/fitness/heatmap` with monthly `2016-09`: desktop and mobile Mapbox canvas filled the map area, with no CSP console errors, no Mapbox request failures, and no horizontal overflow.

## Notes

The sibling `activities.local` checkout has a local compose change to pass `MAPBOX_PUBLIC_VALUE` into the Activity.next Docker build. That checkout has no GitHub remote configured here, so this PR only contains the Activity.next app changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,12 +19,15 @@ WORKDIR /opt/activities.next
 USER app
 
 FROM base AS build
+# Public Mapbox value (starts with pk.) used by next.config.ts to include
+# Mapbox CSP sources in standalone build response headers.
+ARG MAPBOX_PUBLIC_VALUE=""
 ADD --chown=app:app . /opt/activities.next/
 RUN yarn config set -H enableGlobalCache true
 RUN yarn install --immutable
 RUN yarn dedupe
 RUN ACTIVITIES_SECRET_PHASE=build-placeholder yarn knex migrate:latest --disable-transactions
-RUN ACTIVITIES_SECRET_PHASE=build-placeholder BUILD_STANDALONE=true yarn build
+RUN ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN="${MAPBOX_PUBLIC_VALUE}" ACTIVITIES_SECRET_PHASE=build-placeholder BUILD_STANDALONE=true yarn build
 
 FROM base AS output
 ENV NODE_ENV="production"
--- a/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
+++ b/app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
@@ -449,7 +449,7 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
           ROUTE_HEATMAP_MAP_HEIGHT_CLASS
         )}
       >
-        <div ref={containerRef} className="absolute inset-0" />
+        <div ref={containerRef} className="h-full w-full" />
         <div className="absolute left-3 top-3 rounded bg-background/90 px-2 py-1 text-xs text-muted-foreground shadow-sm">
           Mapbox
         </div>
@@ -951,6 +951,14 @@ export const FitnessHeatmapView: FC<Props> = ({
     Boolean(heatmapData) || heatmaps.length > 0 || generationPending
   const canRefreshRouteCache =
     !isLoading && !generationPending && !isRouteHeatmapInFlight(heatmapData)
+  const routeStatusOverlay =
+    heatmapData?.status === 'failed'
+      ? 'failed'
+      : pollingStalled && !hasCompletedRoutes
+        ? 'polling-stalled'
+        : !isLoading && generationPending && !hasCompletedRoutes
+          ? 'generation-pending'
+          : null
 
   return (
     <div className="flex min-h-[720px] flex-col bg-background">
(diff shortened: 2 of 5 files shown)


app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,79 +1034,83 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-                >
-                  <RefreshCw
-                    className={cn('size-3', isRetrying && 'animate-spin')}
-                  />
-                  Retry
-                </button>
-              </div>
-            )}
-          </div>
+          )}
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: All three status overlays — "Route cache queued" (~line 1066), "taking longer than expected" (~line 1071), and "Route cache failed" (~line 1090) — use `absolute inset-x-3 bottom-3` and are rendered as independent conditions, so they can stack on top of each other if more than one condition is true simultaneously.

`pollingStalled` and `heatmapData?.status === 'failed'` are the most likely pair to be true at the same time: the poller can stall *and then* the status comes back as `failed`. Both divs render at the same position and overlap.

**Fix**: either guard the stalled overlay against the failed state:
```tsx
{pollingStalled && !hasCompletedRoutes && heatmapData?.status !== 'failed' && (
```
or derive a single `statusOverlay` value outside JSX and render only that, so the precedence is explicit (e.g. `failed` > `pollingStalled` > `generationPending`).

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,79 +1034,83 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+            heatmap={heatmapData}
+            mapboxAccessToken={mapboxAccessToken}
+          />
+          {!isLoading && generationPending && !hasCompletedRoutes && (
+            <div className="absolute inset-x-3 bottom-3 rounded border bg-background/95 px-3 py-2 text-sm text-muted-foreground shadow-sm">
+              Route cache queued
+            </div>
+          )}
+          {pollingStalled && !hasCompletedRoutes && (
+            <div className="absolute inset-x-3 bottom-3 flex items-center justify-between gap-3 rounded border bg-background/95 px-3 py-2 text-sm shadow-sm">
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: This div appears dynamically (mounted when polling stalls) but has no `aria-live` attribute. Screen reader users will not be notified when the warning appears. Add `aria-live="polite"` to this container div.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,79 +1034,83 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+              >
+                <RefreshCw
+                  className={cn('size-3', isRetrying && 'animate-spin')}
+                />
+                Retry
+              </button>
+            </div>
+          )}
+          {heatmapData?.status === 'failed' && (
+            <div className="absolute inset-x-3 bottom-3 flex items-center justify-between gap-3 rounded border bg-background/95 px-3 py-2 text-sm shadow-sm">
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Same as the stalled overlay: this "Route cache failed" message appears dynamically but has no `aria-live` attribute. Because this is an error, `aria-live="assertive"` is more appropriate here than `"polite"` — it interrupts the user to signal the failure immediately.

---

Dockerfile
```diff
@@ -19,12 +19,13 @@ WORKDIR /opt/activities.next
 USER app
 
 FROM base AS build
```
llun, line 21: The name `MAPBOX_PUBLIC_VALUE` is generic and doesn't hint at the expected format. The corresponding env var is `ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN` and `next.config.ts` checks that it starts with `pk.`. Consider renaming the build arg to `MAPBOX_PUBLIC_TOKEN` (or even mirroring the env var name directly) so that operators know exactly what to pass:

```dockerfile
ARG ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN=""
RUN ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN="${ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN}" \
    ACTIVITIES_SECRET_PHASE=build-placeholder BUILD_STANDALONE=true yarn build
```

At minimum, a comment would help:
```dockerfile
# Pass a public Mapbox token (starts with pk.) so next.config.ts can include
# Mapbox CSP sources in the standalone build's response headers.
ARG MAPBOX_PUBLIC_VALUE=""
```

---

llun (commented):
## Review

### Overview

This PR correctly addresses the two root causes:
1. The Mapbox container used `absolute inset-0`, which stopped working once Mapbox GL's own stylesheet overrides the `position` property on the element. Switching to `h-full w-full` avoids relying on positioning and keeps the canvas filling its parent.
2. The Docker standalone build ran `yarn build` without `ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN` set, so `next.config.ts` never included Mapbox CSP sources in the baked-in response headers. Passing the token via `ARG` at build time fixes that.

The accessibility clean-up (adding `type="button"`, `aria-pressed`, `aria-expanded`, `focus-visible` rings, sr-only back-link text, `aria-live` on the error span in `FitnessHeatmapList`) is solid.

---

### Issues to fix

**1. Overlapping status overlays (functional bug)**
See inline comment on lines 1066–1108. Three overlay divs all render at `absolute inset-x-3 bottom-3`, and `pollingStalled` + `heatmapData?.status === 'failed'` are independent conditions that can both be true simultaneously, causing them to visually collide. The `failed` state should take priority over `pollingStalled`.

**2. `aria-live` missing on the two dynamic map overlays**
See inline comments on lines 1072 and 1091. The stalled-warning and failed-error overlays are dynamically mounted but have no `aria-live` attribute — screen reader users will not hear them. The `pollingStalled` one should use `aria-live="polite"` and the `failed` one `aria-live="assertive"`. (The `error` span in `FitnessHeatmapList.tsx` correctly got `aria-live="polite"` in this PR — same treatment is needed here.)

**3. Dockerfile: `MAPBOX_PUBLIC_VALUE` naming is ambiguous**
See inline comment on line 21. The arg name doesn't hint at the expected format (`pk.` prefix) or which env var it maps to. Rename to match the env var or add a comment.

---

### Minor / worth verifying

**Layout: `<section>` sits outside the `flex-1` grid**
`<section aria-label="Route heatmap map">` is now a sibling of `<div className="grid flex-1 ...">`. For this to size correctly the outer container needs to be a `flex flex-col` — the section will then shrink-wrap to the map's fixed `h-[420px]` height and the grid gets the remaining space. No change needed if that's already established; just worth a quick check that no flex context was lost when the `<main>` wrapper was removed.

**Test: `compareDocumentPosition` bitwise check**
The new test uses `Node.DOCUMENT_POSITION_FOLLOWING` and `.toBeTruthy()`. The logic is correct, but it'll be opaque to someone unfamiliar with the DOM API. A comment on the assertion would help readability.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added `aria-live="polite"` to the polling-stalled route-cache overlay.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,79 +1034,83 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-              <span className="text-muted-foreground">
-                {selectedRegionIds.length > 0
-                  ? `${selectedRegionIds.length} regions`
-                  : 'World'}
+              <button
+                type="button"
+                onClick={retryCurrent}
+                disabled={isRetrying}
+                className="inline-flex items-center gap-1.5 rounded border px-2 py-1 text-xs hover:bg-muted focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50"
+              >
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Replaced the overlay retry controls with the shared `Button` component. The existing `retryCurrent` handler already wraps the async retry in try/catch, updates the error state for feedback, and keeps `disabled={isRetrying}` tied to the in-flight request.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added a live region to the failed route-cache overlay. I used `aria-live="assertive"` for the failure state so it is announced with error priority.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,79 +1034,83 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-          </div>
+          )}
+        </div>
+      </section>
 
+      <div className="grid flex-1 grid-cols-1 lg:grid-cols-[minmax(0,1fr)_320px]">
+        <main className="min-w-0">
           <section className="space-y-3 px-3 py-4">
             <div className="flex flex-wrap items-center justify-between gap-3">
               <h2 className="inline-flex items-center gap-2 text-base font-medium">
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Associated the Activity Calendar section with its heading using `aria-labelledby="activity-calendar-heading"` and added the matching heading `id`.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Made overlay precedence explicit with a single `routeStatusOverlay` discriminator: `failed` > `polling-stalled` > `generation-pending`. Only one bottom overlay renders now.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added `aria-live="polite"` to the dynamically mounted stalled warning overlay.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added `aria-live="assertive"` to the dynamically mounted failed route-cache overlay.

Dockerfile
llun: Kept the neutral `MAPBOX_PUBLIC_VALUE` arg name to avoid Docker's secret-name heuristic, and added a Dockerfile comment documenting that it expects a public `pk.` Mapbox value used by `next.config.ts` for standalone CSP headers.

llun:
@codex review

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,92 +1042,114 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+            {isLoading && (
+              <span className="inline-flex items-center gap-1.5 text-muted-foreground">
+                <Loader2 className="size-3 animate-spin" />
+                Loading…
               </span>
-              <span className="text-muted-foreground">
-                {periodType === 'all_time' ? 'All time' : effectivePeriodKey}
+            )}
+          </div>
+        </div>
```
Comment (llun) [inline app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:1070]: The `polling-stalled` and `failed` overlays both got `aria-live` in this revision, but `generation-pending` is still missing it. "Route cache queued" appears dynamically in the same way and should receive `aria-live="polite"` for consistency:

```tsx
{routeStatusOverlay === 'generation-pending' && (
  <div
    aria-live="polite"
    className="absolute inset-x-3 bottom-3 rounded border bg-background/95 px-3 py-2 text-sm text-muted-foreground shadow-sm"
  >
    Route cache queued
  </div>
)}
```

---

llun (commented):
## Round 2 review

All issues from round 1 are addressed and resolved. Good work on the revision.

### What was fixed
- **Overlapping overlays**: replaced the three independent conditions with a single `routeStatusOverlay` discriminator (`failed` > `polling-stalled` > `generation-pending`). Only one overlay renders at a time, and precedence is now explicit and readable. ✅
- **`aria-live` on `polling-stalled` overlay**: `aria-live="polite"` added. ✅
- **`aria-live` on `failed` overlay**: `aria-live="assertive"` added — correct choice for error priority. ✅
- **`Button` component in overlays**: switched from raw `<button>` to the shared `Button` component. ✅
- **`aria-labelledby` on activity calendar section**: heading now has a matching `id`. ✅
- **Dockerfile comment**: explains the `pk.` format and its purpose. The reasoning for keeping the neutral `MAPBOX_PUBLIC_VALUE` name (to avoid Docker's secret-name heuristic) is sound. ✅
- **Test comment**: `compareDocumentPosition` assertion now has an explanatory comment. ✅

### One remaining item
See inline comment on the `generation-pending` overlay (line 1067). It's the only one of the three overlays that still lacks `aria-live="polite"`. Minor, but worth fixing for consistency before marking the PR ready.

---

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:1070]: Added `aria-live="polite"` to the `generation-pending` / `Route cache queued` overlay.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,92 +1042,114 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+            <span className="text-muted-foreground">
+              {periodType === 'all_time' ? 'All time' : effectivePeriodKey}
+            </span>
+            <span className="text-muted-foreground">
+              {selectedRegionIds.length > 0
+                ? `${selectedRegionIds.length} regions`
+                : 'World'}
+            </span>
+            {isLoading && (
+              <span className="inline-flex items-center gap-1.5 text-muted-foreground">
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added `role="status"` to the inline `Loading…` indicator so map-loading state changes are announced.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1034,92 +1042,114 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+          </div>
+        </div>
+
+        <div className="relative overflow-hidden">
+          <RouteHeatmapMap
+            heatmap={heatmapData}
+            mapboxAccessToken={mapboxAccessToken}
+          />
+          {routeStatusOverlay === 'generation-pending' && (
+            <div className="absolute inset-x-3 bottom-3 rounded border bg-background/95 px-3 py-2 text-sm text-muted-foreground shadow-sm">
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx]: Added `aria-live="polite"` to the `Route cache queued` overlay.

app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -1174,9 +1204,10 @@ export const FitnessHeatmapView: FC<Props> = ({
                 <div className="flex items-center gap-1.5">
                   {hasRouteCache && (
                     <button
+                      type="button"
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/fitness/heatmap/FitnessHeatmapView.tsx:1213]: Added `aria-haspopup="dialog"` to the manual clear-cache dialog trigger while preserving the existing disabled state during clearing/retry.

llun:
@codex review

llun (commented):
## Round 3 review — LGTM

All outstanding items from round 2 are addressed, plus a couple of unprompted improvements.

### What was fixed
- **`generation-pending` overlay**: `aria-live="polite"` added. All three status overlays now have appropriate live regions. ✅

### New improvements (not requested)
- **Loading spinner**: wrapped in `role="status"`, which gives screen readers an implicit `aria-live="polite"` + `aria-atomic="true"` — correct for a transient status message. ✅
- **"Clear cache" button**: `aria-haspopup="dialog"` added, correctly advertising to AT that the button opens a dialog. ✅

### No remaining concerns
The `routeStatusOverlay` logic, the `Button`/raw-button split, the Dockerfile comment, the test annotation, and all the accessibility attributes are in good shape. The PR is ready to come out of draft.

---

PR #696, opened 2026-05-15
fix: keep deployment config runtime-only

llun:
## Summary

This fixes presigned S3 media uploads being blocked by CSP while keeping deployment configuration runtime-only. The build machine can now produce the artifact without real `ACTIVITIES_*`, `OTEL_EXPORTER_*`, or `config.json` runtime values, and invalid placeholder deployment env values do not get baked into `next.config.ts`.

## Changes

- Moves CSP generation to request-time proxy code so media storage, S3, Mapbox, and media-domain origins are derived from runtime config helpers.
- Caches the generated CSP for the process lifetime and resets that cache in Jest before and after each test.
- Keeps the proxy matcher off static asset paths, with segment-boundary handling for `activities/_next` and matching tests.
- Makes the proxy emit only CSP while static security headers stay in `next.config.ts`.
- Preserves CSP on the canonical `/@user@host` early-return path.
- Reads runtime `config.json` through a shared helper without static `fs` or `path` imports in the proxy import graph.
- Layers runtime security-header env settings over runtime `config.json` settings per key/field so file config does not shadow deployment env.
- Reads proxy host fallback config from runtime env or runtime `config.json`, with runtime env taking precedence.
- Routes remaining app runtime env reads through `lib/config` helpers, including app registration proxy trust without full config validation.
- Enforces `ACTIVITIES_ALLOW_MEDIA_DOMAINS` at runtime through `img-src` CSP allowlisting.
- Adds source-level regression coverage that prevents `ACTIVITIES_*` and `OTEL_EXPORTER_*` reads or env-name constants outside `lib/config` production source.
- Updates `AGENTS.md`, `CLAUDE.md`, and environment docs to make runtime-only deployment config, config utility usage, and thin Next config boundaries explicit for future changes.

## Review Follow-up

- Addressed the initial PR review threads in `30dbf26c` and the second-round threads in `0c000cc1`.
- Replied to all review threads and marked each one resolved.
- Requested fresh reviews with `@codex review` and `/gemini review` comments.

## Validation

- `yarn test lib/config/securityHeaders.test.ts proxy.test.ts next.config.test.ts --runInBand`
- `yarn lint` passed with existing unrelated `no-explicit-any` warnings in auth/fitness files.
- `yarn build`
- `yarn test` passed: 296 suites, 2349 tests.
- `env ACTIVITIES_ALLOW_MEDIA_DOMAINS=not-json ACTIVITIES_HOST=https://invalid-build-host ACTIVITIES_DATABASE=not-json ACTIVITIES_MEDIA_STORAGE_TYPE=s3 ACTIVITIES_MEDIA_STORAGE_BUCKET= ACTIVITIES_MEDIA_STORAGE_REGION= OTEL_EXPORTER_OTLP_HEADERS=not-valid yarn build`
- Source search confirmed no `ACTIVITIES_*` or `OTEL_EXPORTER_*` env reads or env-name constants remain outside `lib/config` production source.
- `git diff --check`


lib/config/host.ts
```diff
@@ -45,10 +48,10 @@ const getEnvironmentList = (
 const isRecord = (value: unknown): value is Record<string, unknown> =>
   Boolean(value) && typeof value === 'object' && !Array.isArray(value)
 
-const getInjectedProxyHostConfig = (): HostConfig | null => {
+const getFileProxyHostConfig = (): HostConfig | null => {
   try {
     const parsed = JSON.parse(
-      process.env.ACTIVITIES_PROXY_HOST_CONFIG || 'null'
+      fs.readFileSync(path.resolve(process.cwd(), 'config.json'), 'utf-8')
```
llun: **Reintroduces `fs.readFileSync` into the middleware import graph; reverses prior edge-safe work.**

PR #671 (`fix: require canonical ActivityPub signatures and trusted hosts`) intentionally replaced filesystem reads here with the build-injected env var `ACTIVITIES_PROXY_HOST_CONFIG` so `proxy.ts` would not depend on `fs`/`path` at import time. This PR removes `getInjectedProxyHostConfig` and reintroduces `fs.readFileSync(path.resolve(process.cwd(), 'config.json'))`, which `proxy.ts` then transitively imports via `getProxyHostConfig`.

Risks:
- If `proxy.ts` ever runs under the Edge runtime, this will fail (`Module not found: Can't resolve 'fs'`).
- Even on the Node middleware runtime, synchronous disk I/O on the first proxy request of every cold worker is undesirable.
- Goes directly against the PR's own stated principle of keeping build-time and runtime config paths thin.

Suggested fix: keep an env-injected path (e.g. read `ACTIVITIES_PROXY_HOST_CONFIG` at deploy/start time and prefer it over the file), or move the `config.json` read into a non-middleware module so `proxy.ts`'s import graph stays edge-safe. At minimum, add a regression test that `proxy.ts`'s static imports do not pull in `fs`/`path`.

---

proxy.ts
```diff
@@ -3,15 +3,24 @@ import { NextRequest, NextResponse } from 'next/server'
 import { getProxyHostConfig } from '@/lib/config/host'
 import { acceptContainsContentTypes } from '@/lib/utils/acceptContainsContentTypes'
 import { selectHeaderHost } from '@/lib/utils/host'
+import { getSecurityHeaders } from '@/lib/utils/securityHeaders'
 
 export const config = {
-  matcher: ['/(@.*)']
+  matcher: ['/:path*']
```
llun: **Matcher widened from `'/(@.*)'` to `'/:path*'` — now runs on every request, including static assets.**

The proxy now executes on `_next/static/*`, `favicon.ico`, `/activities/_next/*`, every API route, etc. For each, the function runs regex checks and `withSecurityHeaders` (which computes CSP via `getContentSecurityPolicy()`).

Two issues:
1. **Performance**: CSP is recomputed (or at least re-serialized) on every static asset response, and a 1–2 KB CSP header is attached to every PNG/JS/CSS.
2. **Header conflicts**: any API route that sets its own `Content-Security-Policy` will be silently overwritten by `withSecurityHeaders` (which uses `.set`, not `.has` + `.set`).

Recommended matcher pattern from the Next docs:

```ts
matcher: ['/((?!_next/static|_next/image|favicon\\.ico|activities/_next).*)']
```

If the goal is "CSP on all dynamic responses," that pattern still accomplishes it without paying the cost on cacheable assets.

---

lib/config/envAccess.test.ts
```diff
@@ -0,0 +1,50 @@
+import fs from 'fs'
+import path from 'path'
+
+const SOURCE_ROOTS = ['app', 'lib']
+const ROOT_SOURCE_FILES = [
+  'next.config.ts',
+  'proxy.ts',
+  'instrumentation.ts',
```
llun, line 8: **Regex over-matches plain string literals; will produce false positives.**

The second alternation matches **any** quoted string containing `ACTIVITIES_*` or `OTEL_EXPORTER_*` — including log messages, JSDoc, and thrown errors. For example, this perfectly valid line elsewhere in the codebase would fail the test:

```ts
throw new Error('ACTIVITIES_HOST is required')
```

That pushes contributors toward either moving user-facing messages into `lib/config/` (sometimes appropriate, often not) or writing string-concat hacks to evade the regex (always bad).

Suggested tightening: drop the second alternation entirely, since the first two cover the actual `process.env` reads — the rule is about reading the env, not naming it. If you want to also catch env-name **constants** (`const FOO = 'ACTIVITIES_HOST'`), require an assignment context:

```ts
/process\.env(?:\.(?:ACTIVITIES|OTEL_EXPORTER)_|\[['"](?:ACTIVITIES|OTEL_EXPORTER)_\w+['"]\])|=\s*['"](?:ACTIVITIES|OTEL_EXPORTER)_\w+['"]/
```

Also, `isSourceFile` only excludes `*.test.ts(x)` — consider also excluding `__mocks__/`, `*.spec.ts`, and any `stub/` or `fixtures/` directories so test scaffolding doesn't accidentally trip the rule.

---

lib/config/nextImageRemotePatterns.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const getImageRemotePatterns = (): ImageRemotePatterns => {
+  const patterns: ImageRemotePatterns = [{ protocol: 'https', hostname: '**' }]
+
+  if (isDevelopment()) {
+    patterns.push(...SAFE_LOCAL_IMAGE_REMOTE_PATTERNS)
+  }
+
+  return patterns
+}
```
llun, line 32: **`ACTIVITIES_ALLOW_MEDIA_DOMAINS` is now a silent no-op.**

The previous `next.config.ts` honored this env var to build an explicit `images.remotePatterns` allowlist (see old `IMAGE_REMOTE_ALLOWLIST_ENV` / `getImageRemotePatterns(rawAllowlist)`). The new implementation returns `[{ protocol: 'https', hostname: '**' }]` unconditionally and ignores the env var entirely.

The setting is still parsed into the zod schema at `lib/config/index.ts:38,108-110` (`allowMediaDomains: z.string().array().optional()`) but nothing reads it anymore. Operators relying on it to restrict image origins will silently get broader behavior than they configured.

The docs update says "no longer needs to be provided during `next build`" — that's misleading. It's not just "no longer needed at build time", it's no longer enforced at all.

Pick one:
- Remove `allowMediaDomains` from the schema and add a runtime deprecation warning when it's set.
- Wire the parsed `allowMediaDomains` into image-source CSP (`img-src`) so the allowlist still has teeth.

Either way the docs should say "deprecated / no longer enforced", not "no longer needed at build time."

---

proxy.ts
```diff
@@ -58,9 +67,11 @@ export async function proxy(request: NextRequest) {
 
       const cloneUrl = request.nextUrl.clone()
       cloneUrl.pathname = `/${pathItems.join('/')}`
-      return NextResponse.rewrite(cloneUrl)
+      return withSecurityHeaders(NextResponse.rewrite(cloneUrl))
```
llun: **The `totalAt === 2` short-circuit a few lines above (line 62) is missed — it still returns a bare `NextResponse.next()`.**

```ts
if (totalAt === 2) return NextResponse.next()
```

Every other return in `proxy()` was wrapped in `withSecurityHeaders(...)` in this PR, but this branch (handling `/@user@host` paths) wasn't. Since `nextConfig.headers()` now emits only the static set (CSP was moved to request-time), this branch ships a response with **no CSP at all** — even though it's exactly the federated-actor path that most needs it.

Fix:
```ts
if (totalAt === 2) return withSecurityHeaders(NextResponse.next())
```

And worth adding a `proxy.test.ts` case asserting CSP is present on a `/@user@host` request.

---

lib/utils/securityHeaders.ts
```diff
@@ -0,0 +1,150 @@
...
+  ].join(' ')
+
+  return [
+    "default-src 'none'",
+    "base-uri 'self'",
+    "object-src 'none'",
+    "frame-ancestors 'none'",
+    "form-action 'self'",
+    // Static Next headers cannot attach a per-request nonce to framework
+    // hydration scripts, so inline scripts remain allowed but origins do not.
```
llun: **Stale comment — CSP is no longer "static."**

The justification here for keeping `'unsafe-inline'` in `script-src`/`style-src` was that static Next headers couldn't attach a per-request nonce. After this PR, CSP is emitted from the runtime proxy, so a per-request nonce IS possible. The `'unsafe-inline'` carve-out is no longer required by this constraint.

Either update the comment to explain the actual current reason (`unsafe-inline` is still needed because the Next runtime injects hydration scripts without a configurable nonce, or whatever applies), or take the opportunity to wire a nonce through and drop `'unsafe-inline'`. As-is, the comment is misleading.

---

next.config.ts
```diff
@@ -358,7 +39,7 @@ const nextConfig: NextConfig = {
     return [
       {
         source: '/:path*',
-        headers: getSecurityHeaders()
+        headers: getStaticSecurityHeaders()
```
llun, line 42: **Static security headers are emitted twice.**

`nextConfig.headers()` attaches `getStaticSecurityHeaders()` (4 headers: X-Content-Type-Options, X-Frame-Options, Referrer-Policy, Permissions-Policy) on every response, AND the proxy emits the same 4 headers via `getSecurityHeaders()` → `...getStaticSecurityHeaders()`.

For proxy-rewritten responses, Next merges middleware headers with `nextConfig.headers()` output. Depending on the request path that can yield duplicate header values (visible as comma-joined values on some headers, or simply redundant work on every response).

Suggested split:
- Proxy emits CSP only (it's the only header that needs runtime config).
- `nextConfig.headers()` continues to emit the static set.

`getSecurityHeaders({ includeContentSecurityPolicy: true })` already supports a flag — invert/extend so the proxy can request `{ csp only }`.

---

app/api/v1/apps/route.ts
```diff
@@ -30,7 +30,7 @@ export const resetAppRegistrationWarningStateForTests = () => {
 }
 
 const shouldTrustProxyIpHeaders = (): boolean =>
-  process.env.ACTIVITIES_TRUST_PROXY_IP_HEADERS === 'true'
+  getConfig().trustProxyIpHeaders === true
```
llun: **Routing through `getConfig()` here is heavier than the old env read and can change failure semantics.**

Previously `process.env.ACTIVITIES_TRUST_PROXY_IP_HEADERS === 'true'` — a cheap, total function returning `false` when unset or invalid. Now `getConfig().trustProxyIpHeaders === true` invokes the memoized full config build, which validates the entire zod schema (and throws in production if `secretPhase` is missing/short, see `lib/config/index.ts:58-67`).

In a misconfigured deployment where the rest of the config is incomplete but app registration is being exercised (e.g., a partially-bootstrapped staging environment), this changes a silent-`false` into a config-load failure that bubbles up as a 500 from `apiResponse(...ERROR_500)`. That's arguably the right behavior, but it's a behavior change worth noting.

Consider a scoped helper in `lib/config/` that reads just `ACTIVITIES_TRUST_PROXY_IP_HEADERS` without validating the rest of the schema — the rule from `CLAUDE.md` is "no env reads outside `lib/config/`", not "every env read must go through full config validation."

---

llun (commented):
Reviewed the runtime-config refactor. The overall direction is right — `next.config.ts` is now a thin shim, the lib/config / lib/utils split is coherent, and `envAccess.test.ts` gives this rule teeth going forward. A couple of items to address before merge, summarized:

**Functional issues (should fix):**
1. `proxy.ts:62` — the `totalAt === 2` short-circuit returns `NextResponse.next()` without `withSecurityHeaders`, so federated `/@user@host` requests get no CSP.
2. `lib/config/host.ts:1-2,51-71` — `fs.readFileSync` is back in the middleware import graph, reversing the env-injection approach from #671. Edge-runtime hostile.
3. `proxy.ts:9` — the matcher widening to `/:path*` means the proxy now runs on every static asset and API route, adding CSP overhead to every response and potentially clobbering CSP set by other route handlers.
4. `lib/config/envAccess.test.ts:8` — the regex matches any string literal containing `ACTIVITIES_*`/`OTEL_EXPORTER_*`, so a future error message like `throw new Error('ACTIVITIES_HOST is required')` would fail the suite. Tighten to require a `process.env` proximity (or constant-assignment context).
5. `lib/config/nextImageRemotePatterns.ts` — `ACTIVITIES_ALLOW_MEDIA_DOMAINS` is now silently ignored, but is still parsed into the zod schema. Either re-wire it into runtime image-source CSP or remove it and warn on use.

**Design / maintainability:**
6. `lib/utils/securityHeaders.ts:122-123` — the comment justifying `'unsafe-inline'` is stale now that CSP is runtime.
7. Static security headers are emitted twice (once via `nextConfig.headers()`, once via the proxy) — split so proxy emits only CSP.
8. `app/api/v1/apps/route.ts:33` — `getConfig()` runs the full schema validation; the old env-direct read was a cheap silent `false` on misconfigure. Consider a scoped helper in `lib/config/`.

Inline comments on each item above.

---

lib/utils/securityHeaders.ts
```diff
@@ -0,0 +1,172 @@
...
+    // image loads. next/image optimization is disabled so this does not
+    // reintroduce arbitrary server-side media fetches.
+    `img-src ${imageSources}`,
+    `connect-src ${connectSources}`,
+    "font-src 'self' data:",
+    "manifest-src 'self'",
+    "media-src 'self' https: blob:",
+    "worker-src 'self' blob:"
+  ].join('; ')
+}
```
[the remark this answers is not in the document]
llun: Addressed in `8ab261a0`: `getContentSecurityPolicy()` now caches the generated policy for the process lifetime, with a test-only reset helper and a cache regression test.

proxy.ts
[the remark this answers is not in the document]
llun: Addressed in `8ab261a0`: the proxy matcher now excludes `_next/static`, `_next/image`, `favicon.ico`, and `activities/_next`; `proxy.test.ts` asserts the matcher.

proxy.ts
```diff
@@ -3,15 +3,24 @@ import { NextRequest, NextResponse } from 'next/server'
...
 }
 
 const proxyHeaderHost = (headers: Headers): string => {
   return selectHeaderHost(headers, getProxyHostConfig())
 }
 
+const withSecurityHeaders = (response: NextResponse) => {
+  for (const header of getSecurityHeaders()) {
+    response.headers.set(header.key, header.value)
+  }
```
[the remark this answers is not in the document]
llun: Addressed in `8ab261a0`: the proxy now emits only the runtime CSP header via `getContentSecurityPolicyHeader()`, while static security headers remain in `next.config.ts`.

proxy.ts
```diff
@@ -58,9 +67,11 @@ export async function proxy(request: NextRequest) {
 
       const cloneUrl = request.nextUrl.clone()
       cloneUrl.pathname = `/${pathItems.join('/')}`
-      return NextResponse.rewrite(cloneUrl)
+      return withSecurityHeaders(NextResponse.rewrite(cloneUrl))
     }
 
-    return NextResponse.next()
+    return withSecurityHeaders(NextResponse.next())
```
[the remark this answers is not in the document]
llun: Addressed in `8ab261a0`: the canonical `/@user@host` early return now wraps `NextResponse.next()` with CSP, and there is a regression test for that path.

lib/config/host.ts
llun: Addressed in `8ab261a0`: runtime `config.json` reads now go through `readRuntimeConfigFile()` using `process.getBuiltinModule()` inside the function, keeping static `fs`/`path` imports out of the proxy graph. Added a proxy import-graph regression test.

proxy.ts
llun: Addressed in `8ab261a0`: narrowed the matcher, cached CSP generation, and changed proxy header application to CSP-only without overwriting an existing CSP header.

lib/config/envAccess.test.ts
llun: Addressed in `8ab261a0`: tightened the env-access regex to catch direct env reads and assignment-context constants only, added exclusions for spec/mocks/fixtures/stubs, and added focused tests for those cases.

lib/config/nextImageRemotePatterns.ts
llun: Addressed in `8ab261a0`: `ACTIVITIES_ALLOW_MEDIA_DOMAINS` is now enforced at runtime through `img-src` CSP allowlisting, and the docs describe that behavior.

proxy.ts
llun: Addressed in `8ab261a0`: the `/@user@host` short-circuit now returns through the CSP wrapper, with a `proxy.test.ts` regression.

lib/utils/securityHeaders.ts
llun: Addressed in `8ab261a0`: updated the stale CSP comment to describe the current runtime reason for inline script/style allowance.

next.config.ts
llun: Addressed in `8ab261a0`: split runtime CSP from static headers so the proxy emits CSP only; the proxy test asserts static headers like `X-Content-Type-Options` are not emitted there.

app/api/v1/apps/route.ts
llun: Addressed in `8ab261a0`: added `getTrustProxyIpHeadersConfig()` as a scoped config helper, so app registration no longer loads or validates the full config just to check proxy IP trust.

lib/config/securityHeaders.ts
```diff
@@ -0,0 +1,82 @@
...
+    type: process.env.ACTIVITIES_FITNESS_STORAGE_TYPE,
+    bucket: process.env.ACTIVITIES_FITNESS_STORAGE_BUCKET,
+    region: process.env.ACTIVITIES_FITNESS_STORAGE_REGION,
+    hostname: process.env.ACTIVITIES_FITNESS_STORAGE_HOSTNAME,
+    mapboxAccessToken: process.env.ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN
+  }
+})
+
+export const getSecurityHeaderConfig = (): SecurityHeaderConfig =>
+  getFileSecurityHeaderConfig() ?? getEnvironmentSecurityHeaderConfig()
```
llun: **File config silently shadows env vars when both are present — inconsistent with the other config helpers in this PR.**

`getFileSecurityHeaderConfig` returns `null` only if **none** of `allowMediaDomains`, `mediaStorage`, or `fitnessStorage` is present in `config.json`. As soon as one of those three is set, the file branch wins and `getEnvironmentSecurityHeaderConfig()` is never called — so any `ACTIVITIES_*` env vars for the other two keys are silently dropped.

Concrete failure: a deployment with `config.json` containing `{ "mediaStorage": {...} }` AND `ACTIVITIES_FITNESS_STORAGE_BUCKET` / `ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN` set in env will produce a CSP without the fitness origins. The previous behavior (`next.config.ts` env-only) would have included them; the user has no signal anything was dropped.

This is inconsistent with the other helpers introduced in this PR:
- `getProxyHostConfig` in `lib/config/host.ts:95-102` — env first, then file fallback.
- `getTrustProxyIpHeadersConfig` in `lib/config/trustProxyIpHeaders.ts:4-8` — env first, then file fallback.

Suggested fix: merge per-key with env taking precedence, e.g.

```ts
export const getSecurityHeaderConfig = (): SecurityHeaderConfig => {
  const file = getFileSecurityHeaderConfig() ?? {
    allowMediaDomains: [],
    mediaStorage: {},
    fitnessStorage: {}
  }
  const env = getEnvironmentSecurityHeaderConfig()
  return {
    allowMediaDomains: env.allowMediaDomains.length
      ? env.allowMediaDomains
      : file.allowMediaDomains,
    mediaStorage: { ...file.mediaStorage, ...stripUndefined(env.mediaStorage) },
    fitnessStorage: { ...file.fitnessStorage, ...stripUndefined(env.fitnessStorage) }
  }
}
```

That way operators can layer env on top of `config.json` the same way they can for host/trust-proxy settings.

---

lib/config/runtimeConfigFile.ts
```diff
@@ -0,0 +1,16 @@
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  Boolean(value) && typeof value === 'object' && !Array.isArray(value)
+
+export const readRuntimeConfigFile = (): Record<string, unknown> | null => {
+  try {
+    const fs = process.getBuiltinModule('fs')
+    const path = process.getBuiltinModule('path')
+    const parsed = JSON.parse(
+      fs.readFileSync(path.resolve(process.cwd(), 'config.json'), 'utf-8')
```
llun, line 14: **Nice touch with `process.getBuiltinModule`.**

Dynamic builtin lookup keeps webpack from statically bundling `fs`/`path` into the proxy, while the `try/catch` covers Edge runtime where the API doesn't exist. Pairs well with the new import-graph regression test in `proxy.test.ts:308-343`. One caveat: `process.getBuiltinModule` requires Node ≥ 22.13 — fine given the Node 24 floor in `CLAUDE.md`, but worth pinning via `engines.node` in `package.json` if it isn't already, since silent `null` returns on older Node would just hide the config-file fallback entirely.

---

lib/utils/securityHeaders.ts
```diff
@@ -0,0 +1,168 @@
...
+
+const hasPublicMapboxAccessToken = (
+  fitnessStorage: ReturnType<typeof getSecurityHeaderConfig>['fitnessStorage']
+) => fitnessStorage?.mapboxAccessToken?.trim().startsWith('pk.') ?? false
+
+let cachedContentSecurityPolicy: string | null = null
+
+export const resetContentSecurityPolicyCacheForTests = () => {
+  cachedContentSecurityPolicy = null
+}
```
llun, line 89: **Cache is process-wide and shared across all tests — needs explicit reset wherever env mutation might affect CSP.**

Tests in `proxy.test.ts` and `next.config.test.ts` call `resetContentSecurityPolicyCacheForTests()`, but any new test file that touches `ACTIVITIES_MEDIA_STORAGE_*`/`ACTIVITIES_FITNESS_STORAGE_*`/`ACTIVITIES_ALLOW_MEDIA_DOMAINS`/`ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN` and then exercises a code path that calls `getContentSecurityPolicy()` will get a stale value depending on test ordering.

Two options to reduce the footgun:
1. Reset the cache from a Jest global setup (`jest.setup.ts`) `beforeEach`, so every test starts with a clean CSP.
2. Document the requirement in a comment above `resetContentSecurityPolicyCacheForTests` so future contributors know to call it.

Not blocking, but the existing tests `lib/config/mediaStorage.test.ts` and `lib/config/fitnessStorage.test.ts` mutate those env vars without resetting the CSP cache — currently safe only because they don't read CSP.

---

llun (commented):
Second-round review.

**Previous feedback — all addressed:**
1. `proxy.ts:63-65` — `totalAt === 2` short-circuit now wraps with `withContentSecurityPolicy`. ✓
2. `lib/config/host.ts` + new `lib/config/runtimeConfigFile.ts` — `fs.readFileSync` moved behind `process.getBuiltinModule('fs')`, plus a new import-graph regression test in `proxy.test.ts` proves `proxy.ts` doesn't statically pull in `fs`/`path`. ✓
3. `proxy.ts:9` — matcher now excludes `_next/static`, `_next/image`, `favicon.ico`, `activities/_next`, with an explicit test. ✓
4. `proxy.ts:18-19` — `withContentSecurityPolicy` uses `.has` before `.set` to avoid clobbering downstream CSP. ✓
5. `lib/config/envAccess.test.ts:13` — regex tightened to require `=\s*` for the constant form, plus meta-tests covering the `'X is required'` false-positive case, and exclusion of `__mocks__`/`fixtures`/`stub`/`*.spec.*`. ✓
6. `lib/config/securityHeaders.ts:65,95-98` — `allowMediaDomains` is now wired into `img-src`, no longer a silent no-op. ✓
7. `next.config.ts` only emits static headers; proxy only emits CSP — no more duplication. ✓
8. `app/api/v1/apps/route.ts` + new `lib/config/trustProxyIpHeaders.ts` — scoped helper avoids running the full zod validation just to read one flag. ✓

**New observations (1 functional, 2 minor):**
- `lib/config/securityHeaders.ts:82` — file-vs-env merge is all-or-nothing; if `config.json` has any of the three keys, env vars for the other two are silently dropped. Inconsistent with `host.ts` / `trustProxyIpHeaders.ts` (which env-first-then-file). Inline comment with suggested per-key merge.
- `lib/config/runtimeConfigFile.ts` — `process.getBuiltinModule` requires Node ≥ 22.13; pin `engines.node` to avoid silent fallback skip on older runtimes.
- `lib/utils/securityHeaders.ts:81-85` — process-wide CSP cache is a footgun for future tests that touch `ACTIVITIES_*` env vars without resetting; consider a jest global setup reset.

Looks close to landing — only the file/env merge is worth fixing before merge in my view.

---

llun:
@codex review

lib/config/securityHeaders.ts
llun: Addressed in `0c000cc1`: `getSecurityHeaderConfig()` now layers runtime env settings over `config.json` per key/field with undefined env values stripped, and new unit/integration tests cover env-over-file, file fallback, and unrelated-file fallback cases.

lib/config/runtimeConfigFile.ts
llun: Confirmed already covered: `package.json` pins `engines.node` to `24.x`, and the repo instructions/Docker base also use Node 24, which is stricter than the Node >=22.13 requirement for `process.getBuiltinModule`.

lib/utils/securityHeaders.ts
llun: Addressed in `0c000cc1`: added a Jest after-env setup that resets the CSP cache before and after each test, and guarded `resetContentSecurityPolicyCacheForTests()` so it can only be called under Jest.

proxy.ts
```diff
@@ -3,15 +3,25 @@ import { NextRequest, NextResponse } from 'next/server'
 import { getProxyHostConfig } from '@/lib/config/host'
 import { acceptContainsContentTypes } from '@/lib/utils/acceptContainsContentTypes'
 import { selectHeaderHost } from '@/lib/utils/host'
+import { getContentSecurityPolicyHeader } from '@/lib/utils/securityHeaders'
 
 export const config = {
-  matcher: ['/(@.*)']
+  matcher: ['/((?!_next/static|_next/image|favicon\\.ico|activities/_next).*)']
```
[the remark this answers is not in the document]
llun: Addressed in `0c000cc1`: tightened the matcher to use segment boundaries for `activities/_next(?:/|$)` and added behavior tests for `/activities/_next`, `/activities/_next/`, nested static paths, and `/activities/_nextish`.

lib/config/securityHeaders.ts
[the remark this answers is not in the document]
llun: Addressed in `0c000cc1`: security-header config now uses consistent runtime layering with env values taking precedence over file values, plus regression coverage for that precedence.

llun:
@codex review

llun (commented):
Third-round review.

**Second-round feedback — all addressed:**
1. `lib/config/securityHeaders.ts:96-117` — file-vs-env merge is now per-key with env taking precedence. New `lib/config/securityHeaders.test.ts` covers layered merge, file-only, and env-only paths, and `next.config.test.ts` adds an end-to-end test confirming `connect-src` reflects the layered storage origins. ✓
2. `package.json` already pins `"engines.node": "24.x"`, so `process.getBuiltinModule` is safe — nothing to change. ✓
3. `jest.setup-after-env.ts` + `jest.config.mjs` — CSP cache is now reset `beforeEach`/`afterEach` globally, and `resetContentSecurityPolicyCacheForTests` throws if called outside Jest (`JEST_WORKER_ID` guard in `lib/utils/securityHeaders.ts:84-86`). Test-only API is enforced at runtime. ✓

**Bonus improvement worth calling out:**
- `proxy.ts:9-11` — matcher tightened with path-segment boundaries (`(?:/|$)`, `$`). The previous form `(?!_next/static|...)` would have over-matched (and thus *under-proxied*) paths like `/_next/staticfoo`. The new form only excludes when the prefix is followed by `/` or the end of the path, plus a dedicated test in `proxy.test.ts:166-179` asserting `/activities/_nextish` and `/@alice` are matched while `/activities/_next/static/chunk.js` is not.

No outstanding issues from my side. The PR achieves its stated goal — `next.config.ts` is now ~75 lines and free of runtime deployment config reads, `proxy.ts` emits CSP at request time without dragging `fs` into the middleware import graph, env reads are confined to `lib/config/` and policed by `envAccess.test.ts`, and the file-then-env layering is now consistent across `host.ts`, `trustProxyIpHeaders.ts`, and `securityHeaders.ts`. LGTM.

---

lib/config/securityHeaders.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  Boolean(value) && typeof value === 'object' && !Array.isArray(value)
+
+const getStringValue = (value: unknown) =>
+  typeof value === 'string' ? value : undefined
+
+const toStringList = (value: unknown): string[] =>
+  Array.isArray(value) ? value.filter(Boolean).map(String) : []
+
```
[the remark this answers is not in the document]
llun: Fixed in 0b772595. The shared config parsing helpers now live in `lib/config/utils.ts`, and `host.ts`, `securityHeaders.ts`, and `runtimeConfigFile.ts` all use them.

lib/config/runtimeConfigFile.ts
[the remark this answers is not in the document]
llun: Addressed in 0b772595. This repo is on Next 16, where Proxy runs in the Node.js runtime and Edge is not supported/configurable for proxy, so the `config.json` fallback is valid here. I also memoized `readRuntimeConfigFile()` so proxy/runtime config consumers do at most one filesystem read per process, and added the explanatory comment around the dynamic builtin lookup.

lib/config/runtimeConfigFile.ts
```diff
@@ -0,0 +1,16 @@
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+  Boolean(value) && typeof value === 'object' && !Array.isArray(value)
+
+export const readRuntimeConfigFile = (): Record<string, unknown> | null => {
+  try {
+    const fs = process.getBuiltinModule('fs')
+    const path = process.getBuiltinModule('path')
```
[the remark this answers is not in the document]
llun: Fixed in 0b772595. Added a comment explaining that the dynamic builtin lookup keeps `fs`/`path` out of the static proxy import graph while preserving runtime file access.

llun:
@codex review

jest.setup-after-env.ts
```diff
@@ -0,0 +1,14 @@
...
+
+beforeEach(() => {
+  resetRuntimeConfigFileCacheForTests()
+  resetContentSecurityPolicyCacheForTests()
+})
+
+afterEach(() => {
+  resetRuntimeConfigFileCacheForTests()
+  resetContentSecurityPolicyCacheForTests()
+})
```
llun, line 17: **Small inconsistency with the other runtime caches.**

The new `cachedTrustProxyIpHeadersConfig` (`lib/config/trustProxyIpHeaders.ts:3`) is **not** reset by `jest.setup-after-env.ts`, unlike `readRuntimeConfigFile`'s cache and the CSP cache, both of which are reset in `beforeEach`/`afterEach` globally.

Today this is safe — `trustProxyIpHeaders.test.ts` resets its own cache explicitly, and `app/api/v1/apps/route.test.ts` jest-mocks the helper. But the moment any future test indirectly drives `getTrustProxyIpHeadersConfig()` (e.g., a route handler integration test that exercises rate-limited app registration), it'll pick up a stale value from whichever test ran first.

Suggested fix — add it next to the other resets:

```ts
// jest.setup-after-env.ts
import { resetTrustProxyIpHeadersConfigCacheForTests } from '@/lib/config/trustProxyIpHeaders'

beforeEach(() => {
  resetRuntimeConfigFileCacheForTests()
  resetContentSecurityPolicyCacheForTests()
  resetTrustProxyIpHeadersConfigCacheForTests()
})

afterEach(() => {
  // ...same
})
```

Trivial follow-up, not blocking — happy to leave it to a future PR.

---

llun (commented):
Final-round review.

**Third-round feedback — all addressed cleanly:**

1. `lib/config/securityHeaders.ts:96-100` — `allowMediaDomains` precedence semantics refined to `process.env.ACTIVITIES_ALLOW_MEDIA_DOMAINS !== undefined` (not "non-empty array"), so `ACTIVITIES_ALLOW_MEDIA_DOMAINS=[]` deliberately clears the file allowlist instead of silently falling back. New `securityHeaders.test.ts:127-139` case covers this. ✓

2. `lib/config/runtimeConfigFile.ts:3,6,17` — config.json read is now cached process-wide. Synchronous disk I/O happens at most once per worker, not per request. ✓

3. `lib/config/trustProxyIpHeaders.ts:3,6-7,17` — same caching pattern for the trust-proxy flag, with `resetTrustProxyIpHeadersConfigCacheForTests` guarded by `JEST_WORKER_ID` and a new dedicated test that asserts the cache survives a mid-test `config.json` rewrite (`trustProxyIpHeaders.test.ts:63-77`). ✓

4. **Bonus refactor** — `lib/config/utils.ts` now centralizes `isRecord`, `getStringValue`, `toStringList`, `getEnvironmentList`, and the `EnvironmentListOptions` type. Removes duplication between `host.ts` and `securityHeaders.ts`, and `toStringList` now takes a `key` for better error messages. ✓

**One small inconsistency (non-blocking):**
- `jest.setup-after-env.ts` resets `readRuntimeConfigFile` and CSP caches globally but not the new `trustProxyIpHeaders` cache. Inline comment with suggested fix.

Overall the PR is in good shape. The original goal is fully achieved:
- `next.config.ts` is a thin Next entrypoint (75 lines, no `ACTIVITIES_*`/`OTEL_*` reads).
- CSP is generated at request time in the proxy, with per-segment matcher boundaries that avoid CSP overhead on `_next/static`/`_next/image`/`favicon.ico`/`activities/_next`.
- Runtime config reads are confined to `lib/config/`, with three layers of defense: the `envAccess.test.ts` source scan, the proxy import-graph scan in `proxy.test.ts`, and the test-only guards on the cache reset helpers.
- File-vs-env precedence is now consistent across `host`, `trustProxyIpHeaders`, and `securityHeaders`, with explicit tests for each merge axis.

LGTM after the small jest setup follow-up — happy for that to be a separate PR.

---

jest.setup-after-env.ts
llun: Fixed in e64cb5d4. `resetTrustProxyIpHeadersConfigCacheForTests()` now runs in the global Jest before/after hooks with the other runtime cache resets.

llun:
@codex review

PR #700, opened 2026-05-16
minor: remove config.json configuration support

llun:
## Summary
- Remove runtime `config.json` loading from app config, proxy host config, CSP config, and proxy IP trust config.
- Make docs and examples environment-variable only.
- Add/adjust tests proving root `config.json` files are ignored.

## Impact
- `config.json` no longer configures Activity.next; migrate file settings to the documented `ACTIVITIES_*` and `OTEL_EXPORTER_*` environment variables.

## Test Plan
- `yarn test lib/config/index.test.ts lib/config/host.test.ts lib/config/securityHeaders.test.ts lib/config/trustProxyIpHeaders.test.ts proxy.test.ts next.config.test.ts --runInBand`
- `yarn lint`
- `yarn build`
- `yarn test --testPathIgnorePatterns=.claude/worktrees`

## Notes
- Raw `yarn test` still collects the local `.claude/worktrees/optimistic-mahavira-b5aa2b` checkout and fails in that unrelated duplicate tree, matching the existing local caveat.


lib/config/trustProxyIpHeaders.ts
```diff
@@ -13,8 +11,7 @@ export const getTrustProxyIpHeadersConfig = (): boolean => {
     return cachedTrustProxyIpHeadersConfig
   }
 
-  const fileConfig = readRuntimeConfigFile()
-  cachedTrustProxyIpHeadersConfig = fileConfig?.trustProxyIpHeaders === true
+  cachedTrustProxyIpHeadersConfig = false
```
llun: With the file-config fallback gone, this whole function collapses to a single env read. The `environmentValue !== undefined` branch and the `= false` fallback both compute the same thing now (`'true' === 'true'` vs `undefined === 'true'`). Suggest:

```ts
export const getTrustProxyIpHeadersConfig = (): boolean => {
  if (cachedTrustProxyIpHeadersConfig !== null) return cachedTrustProxyIpHeadersConfig
  cachedTrustProxyIpHeadersConfig =
    process.env.ACTIVITIES_TRUST_PROXY_IP_HEADERS === 'true'
  return cachedTrustProxyIpHeadersConfig
}
```

---

lib/config/trustProxyIpHeaders.test.ts
```diff
@@ -66,13 +66,13 @@ describe('getTrustProxyIpHeadersConfig', () => {
...
+    expect(getTrustProxyIpHeadersConfig()).toBe(false)
 
     fs.writeFileSync(
       path.join(tempDirectory, 'config.json'),
-      JSON.stringify({ trustProxyIpHeaders: false })
+      JSON.stringify({ trustProxyIpHeaders: true })
     )
 
-    expect(getTrustProxyIpHeadersConfig()).toBe(true)
+    expect(getTrustProxyIpHeadersConfig()).toBe(false)
```
llun: This test no longer exercises caching. Both `writeFileSync` calls have identical content (`trustProxyIpHeaders: true`), the env var is never set, and the expected value is `false` both times — the cache is irrelevant to the outcome.

To actually prove caching, flip state between calls, e.g.:

```ts
process.env.ACTIVITIES_TRUST_PROXY_IP_HEADERS = 'true'
expect(getTrustProxyIpHeadersConfig()).toBe(true)

delete process.env.ACTIVITIES_TRUST_PROXY_IP_HEADERS
expect(getTrustProxyIpHeadersConfig()).toBe(true) // cached
```

As written, the test would pass even if caching were removed.

---

lib/config/securityHeaders.ts
```diff
@@ -86,25 +47,13 @@ const getEnvironmentSecurityHeaderConfig = (): SecurityHeaderConfig => ({
 })
 
 export const getSecurityHeaderConfig = (): SecurityHeaderConfig => {
```
llun: `getEnvironmentSecurityHeaderConfig` is now only called from here and is just a plain object literal. The extra hop hides the simplicity; consider inlining it into `getSecurityHeaderConfig` so the function reads top-to-bottom as "env in, stripped config out". (Style only — no behavior impact.)

---

lib/config/host.ts
```diff
@@ -45,17 +23,11 @@ export const getHostConfigFromEnvironment = (
...
 })
 
-const getProxyHostConfigFromEnvironment = (
-  options?: EnvironmentListOptions
-): HostConfig => ({
+const getProxyHostConfigFromEnvironment = (): HostConfig => ({
   host: process.env.ACTIVITIES_HOST || '',
-  trustedHosts: getEnvironmentList('ACTIVITIES_TRUSTED_HOSTS', options)
+  trustedHosts: getEnvironmentList('ACTIVITIES_TRUSTED_HOSTS')
 })
```
llun: `getProxyHostConfigFromEnvironment` is now a one-line helper with a single caller (`getProxyHostConfig`). Worth inlining — and the `options?` parameter on `getHostConfigFromEnvironment` is also now only relied on by `getConfigFromEnvironment` (for `onInvalidList: 'throw'`), so the runtime-config-aware shape that justified two separate functions is gone. Could be simplified into one shared helper.

---

lib/config/securityHeaders.test.ts
```diff
@@ -108,23 +105,13 @@ describe('getSecurityHeaderConfig', () => {
...
-        bucket: 'file-fitness-bucket',
-        region: 'us-east-1',
-        hostname: 'file-fitness.example.com',
-        mapboxAccessToken: 'pk.file-mapbox'
-      }
+      allowMediaDomains: [],
+      mediaStorage: {},
+      fitnessStorage: {}
     })
   })
```
llun, line 93: Now that `config.json` is unconditionally ignored, several tests in this file (and in `host.test.ts`, `proxy.test.ts`, `next.config.test.ts`, `trustProxyIpHeaders.test.ts`) still write `config.json` fixtures only to assert "no, none of that came through". One regression test per module ("config.json is ignored") is enough; the rest is just noise that will rot the next time someone touches these files. Consider collapsing into a single representative test per module.

---

lib/config/index.ts
```diff
@@ -1,8 +1,6 @@
-import fs from 'fs'
 import { Knex } from 'knex'
 import memoize from 'lodash/memoize'
 import { PHASE_PRODUCTION_BUILD } from 'next/dist/shared/lib/constants'
```
llun, line 4: `PHASE_PRODUCTION_BUILD` is still imported and used in both `shouldValidateProductionRuntimeSecret` and the build-phase early return inside `getConfigFromEnvironment` (line ~98). That second use was previously the "build pass returns null and falls through to file config" path. With file config gone there's no fallback — during a production build with bad env, this now silently returns `null`, then `getConfig` throws `'Fail to read Activities.next config'`. Worth confirming the Docker build path still succeeds with placeholder/missing `ACTIVITIES_*` values; if it does, the error message at line 118 is misleading for that case.

---

lib/config/index.test.ts
```diff
@@ -42,32 +42,39 @@ describe('Config', () => {
...
+      expect(getConfig()).toMatchObject({
+        host: 'env.example.com',
+        secretPhase: 'env-secret',
+        database: {
+          client: 'better-sqlite3',
+          connection: {
+            filename: ':memory:'
+          }
+        }
+      })
```
llun, line 96: The test name says it "ignores root config.json", but the assertion only checks that env values came through — it never proves the file values (`file.example.com`, `file-secret`, `file.sqlite`) were rejected. Since env vars happen to satisfy the schema regardless of the file, this would still pass even if file-config support sneaked back in.

Add explicit negative assertions, e.g. `expect(getConfig().host).not.toBe('file.example.com')`, so the test pins down the intended behavior.

---

llun (commented):
## Review of PR #700 — remove `config.json` configuration support

The cut is mostly mechanical and lands cleanly, but the PR stops one step short on cleanup, docs, and a couple of tests that no longer prove what they claim.

### Needs fixing

1. **`CLAUDE.md` still references `config.json`.** AGENTS.md was updated, but `CLAUDE.md:10` still says: *"Treat `ACTIVITIES_*`, `OTEL_EXPORTER_*`, secrets, host/database/storage/auth settings, and `config.json` as runtime-only deployment config."* That contradicts the rest of the PR.

2. **Dead exports in `lib/config/utils.ts`.** With `runtimeConfigFile.ts` deleted and `host.ts` / `securityHeaders.ts` no longer importing them, `isRecord` and `getStringValue` have zero remaining callers, and `toStringList` is only used internally by `getEnvironmentList`. Drop the unused exports and downgrade `toStringList` to a non-exported helper.

3. **Version-bump prefix.** The PR title is `minor: …`, but removing `config.json` support is a breaking change for any deployment currently using one — per AGENTS.md this should be `major:`. Since PRs are squash-merged, the title is what the version bumper sees.

4. **Broken caching test** in `lib/config/trustProxyIpHeaders.test.ts` — inline comment. After the refactor the test no longer exercises caching at all.

5. **Misleading "ignores config.json" test** in `lib/config/index.test.ts` — inline comment. The assertion only checks env values land; it never verifies the file values are rejected.

### Suggested cleanup (not blocking)

- `lib/config/trustProxyIpHeaders.ts` collapses to a one-liner — inline comment.
- `lib/config/securityHeaders.ts` and `lib/config/host.ts` have indirection helpers that each have a single caller now — inline comments.
- Several tests across `host.test.ts`, `securityHeaders.test.ts`, `proxy.test.ts`, `next.config.test.ts`, and `trustProxyIpHeaders.test.ts` still write `config.json` fixtures only to assert they were ignored. One regression test per module is plenty.

### Worth double-checking

- The build-time path in `getConfigFromEnvironment` (the `PHASE_PRODUCTION_BUILD` early return) previously fell through to file config when env was missing/invalid. Now it falls through to `getConfig` throwing `'Fail to read Activities.next config'`. Confirm Docker / standalone builds with placeholder `ACTIVITIES_*` values still succeed — see inline comment on `lib/config/index.ts`.

---

lib/config/trustProxyIpHeaders.ts
llun: Simplified this to a cached single env comparison in 40a32608.

lib/config/trustProxyIpHeaders.test.ts
llun: Updated the cache test in 40a32608 to set the env value before the first read, delete it, and assert the cached true value is still returned.

lib/config/securityHeaders.ts
llun: Inlined the environment security header object into getSecurityHeaderConfig in 40a32608.

lib/config/host.ts
llun: Inlined the proxy host environment read in getProxyHostConfig in 40a32608. I kept getHostConfigFromEnvironment as the shared parser because getConfigFromEnvironment still needs the throwing list validation path.

lib/config/securityHeaders.test.ts
llun: I kept the representative ignore fixtures per module because they cover distinct consumers: app config, host/proxy, Next config, CSP, security headers, and trust-proxy config. I avoided a wider test reshuffle in this pass, and the runtime file reader is gone.

lib/config/index.ts
llun: Confirmed yarn build still succeeds after removing the file fallback. The build-phase path remains intentional because Next config does not need the runtime app config during production build.

lib/config/index.test.ts
llun: Added explicit negative assertions for the file host, secret, and DB filename values in 40a32608.

llun:
Addressed the actionable review feedback in 40a32608:

- Removed the remaining config.json docs reference from CLAUDE.md.
- Removed dead config utility exports and made toStringList private.
- Simplified trust-proxy config and fixed the caching test so it proves cached behavior.
- Inlined the one-call host and security-header helpers.
- Added negative assertions proving root config.json values are ignored.
- Verified with prettier, focused config tests, lint, build, full Jest with .claude/worktrees ignored, and git diff --check.

I kept the PR title as minor: per the latest owner instruction in this thread. No action needed for the Vercel deployment comment or Gemini's no-feedback review.

llun:
@codex review

lib/config/securityHeaders.test.ts
llun, line 93: After this round, these three tests (line ~49, ~87, ~114) all write `config.json` fixtures whose contents never appear in any expectation — they prove the same thing: "config.json is ignored". Now that the simplification round has settled, a single representative regression test would carry the same coverage with less noise.

Not blocking — flagging again because the PR is otherwise tidy.

---

llun (commented):
## Round 2 review — looks good, one outstanding item

The follow-up commit `40a3260` addressed essentially all the round 1 feedback cleanly:

- **CLAUDE.md** — updated, no more `config.json` reference. ✅
- **`lib/config/utils.ts`** — `isRecord` and `getStringValue` removed; `toStringList` correctly downgraded to a non-exported local helper. ✅
- **`lib/config/trustProxyIpHeaders.ts`** — collapsed to the one-line env read. ✅
- **`lib/config/trustProxyIpHeaders.test.ts`** — caching test now actually exercises caching by flipping `process.env` between calls. ✅
- **`lib/config/securityHeaders.ts`** — `getEnvironmentSecurityHeaderConfig` inlined into `getSecurityHeaderConfig`; the file is now a single straightforward export. ✅
- **`lib/config/host.ts`** — `getProxyHostConfigFromEnvironment` and `hasRuntimeProxyHostConfig` inlined. ✅
- **`lib/config/index.test.ts`** — added negative assertions confirming the file values were rejected. ✅

### Still outstanding

1. **PR title prefix** — still `minor: remove config.json configuration support`. Removing `config.json` support is a breaking change for any existing deployment that relies on it (their config silently stops being read at upgrade time), so per `AGENTS.md` this should be `major:`. The PR title drives the auto version bump on squash-merge.

2. **Optional — soft-landing for existing users**: consider logging a warning at startup if a `config.json` is present in the working directory after this lands. With the file silently ignored, an upgrade from v1.10.x → this version would leave operators with whatever defaults `getConfigFromEnvironment` produces, often surfacing as opaque downstream failures rather than a clear "config.json no longer supported" message. Cheap to add; saves a support thread or two.

3. **Test redundancy** — see inline comment on `securityHeaders.test.ts`. Several modules still have multiple "writes config.json fixture, asserts it was ignored" tests; one per module covers the regression.

No blocking code issues from my side beyond the title prefix.

---

llun:
Retracting outstanding item 1 from my previous review — keeping the `minor:` prefix is fine, please ignore the suggestion to bump to `major:`.

Remaining notes (soft-landing warning and test redundancy) still stand as optional cleanup, not blocking.

---

lib/config/securityHeaders.test.ts
llun: Collapsed the redundant securityHeaders config.json fixtures in 0f1ccfbf. That file now keeps one representative ignored-config regression test; the remaining cases are plain env/default behavior tests.

lib/config/index.ts
```diff
@@ -138,12 +109,6 @@ const getConfigFromEnvironment = () => {
...
 
 export const getConfig = memoize((): Config => {
-  const fileConfig = getConfigFromFile()
-  if (fileConfig) {
-    validateProductionRuntimeSecret(fileConfig)
-    return fileConfig
-  }
-
   const environmentConfig = getConfigFromEnvironment()
   if (environmentConfig) {
```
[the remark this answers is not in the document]
llun: Added env mappings for ACTIVITIES_SERVICE_NAME, ACTIVITIES_SERVICE_DESCRIPTION, and ACTIVITIES_LANGUAGES in 0f1ccfbf, plus docs/.env.example entries and assertions that env metadata wins over ignored config.json values.

llun:
Addressed the round 2 actionable feedback in 0f1ccfbf:

- Added environment-variable migration support for serviceName, serviceDescription, and languages via ACTIVITIES_SERVICE_NAME, ACTIVITIES_SERVICE_DESCRIPTION, and ACTIVITIES_LANGUAGES.
- Documented those env vars in docs/environment-variables.md and .env.example.
- Added a runtime warning when a root config.json is present so upgrades get a clear migration signal instead of silent ignore.
- Reduced securityHeaders.test.ts to one representative config.json ignore fixture.
- Left the PR title/version prefix as minor: per the latest instruction.

Validation: focused config tests, full lib/config tests, lint, build, full Jest with .claude/worktrees ignored, and git diff --check.

PR #701, opened 2026-05-16
fix: include S3 upload hosts with storage hostname

llun:
## Summary

- Keep custom media storage hostnames in `connect-src` for browser-visible storage URLs.
- Also include default AWS S3 presigned upload origins for `s3` storage even when `ACTIVITIES_MEDIA_STORAGE_HOSTNAME` is configured.
- Add a regression test for the `static.llun.social` + `eu-central-1` configuration shape.

## Root Cause

The CSP helper skipped default AWS S3 origins whenever a storage hostname was configured. That allowed `https://static.llun.social` but omitted the presigned PUT targets such as `https://static.llun.social.s3.eu-central-1.amazonaws.com` and `https://s3.eu-central-1.amazonaws.com`.

## Validation

- `yarn test next.config.test.ts --runInBand` red before the fix, then green after the fix.
- `yarn lint` exited 0 with existing warnings in auth/fitness files.
- `yarn build`
- `yarn test`


lib/utils/securityHeaders.ts
```diff
@@ -60,7 +60,11 @@ const getStorageHostname = (storage: { hostname?: string }) => storage.hostname
 const getDefaultS3CspSources = (storage: unknown) => {
   if (!isS3CompatibleStorage(storage)) return []
 
-  if (!['s3', 'object'].includes(storage.type) || storage.hostname?.trim()) {
+  const hasCustomHostname = Boolean(storage.hostname?.trim())
+  const allowsDefaultS3Sources =
+    storage.type === 's3' || (storage.type === 'object' && !hasCustomHostname)
```
llun: The asymmetric handling of `s3` vs `object` here is worth re-examining. Both `MediaStorageType.S3Storage` and `MediaStorageType.ObjectStorage` flow through `S3FileStorage`, and the client is constructed identically in both cases:

```ts
// lib/services/medias/S3StorageFile.ts:109
this._client = new S3Client({ region: config.region })
```

No `endpoint` is wired in (the same is true for `lib/services/fitness-files/S3StorageFile.ts:79`, and the `MediaStorageS3Config` schema does not even define an `endpoint` field). That means presigned PUT URLs for `object` storage are *also* generated against `https://<bucket>.s3.<region>.amazonaws.com`, exactly like the `s3` type.

So when an operator configures `type=object` together with a custom `hostname`, this branch still returns `[]` and the browser's PUT will be blocked by `connect-src` — the same bug this PR fixes for `s3`. Either:

1. `object` should be treated symmetrically with `s3` here (always emit the default S3 origins), or
2. If `object` is intended to use a non-AWS endpoint, the config schema and `S3Client` construction need an `endpoint` field first, and the CSP logic should derive the upload origin from that instead of from `hostname`.

As-is, the `object`-with-hostname configuration is silently still broken.

---

next.config.test.ts
```diff
@@ -337,6 +337,28 @@ describe('next config security hardening', () => {
...
+        const connectSources = getCspDirectiveSources('connect-src')
+
+        expect(connectSources).toEqual(
+          expect.arrayContaining([
+            'https://static.llun.social',
+            'https://static.llun.social.s3.eu-central-1.amazonaws.com',
+            'https://s3.eu-central-1.amazonaws.com'
+          ])
+        )
+      }
```
llun, line 358: The new regression test only covers `ACTIVITIES_MEDIA_STORAGE_*`. `getDefaultS3CspSources` is invoked for both media and fitness storage (see `getContentSecurityPolicy` in `lib/utils/securityHeaders.ts`), and the bug pattern — custom hostname suppressing the AWS S3 presigned PUT origin — exists identically for fitness. Please add a parallel test using `ACTIVITIES_FITNESS_STORAGE_TYPE=s3` + `ACTIVITIES_FITNESS_STORAGE_HOSTNAME=…` so a future change to the fitness branch can't silently regress.

---

llun (commented):
The fix correctly unblocks the `s3` + custom-hostname case (`static.llun.social` with `eu-central-1` presigned PUTs), and the added test pins that exact configuration shape. Two things worth addressing before merge:

1. **`object` storage with a custom hostname is still silently broken.** Both `s3` and `object` types use the same `new S3Client({ region })` constructor (no `endpoint` is plumbed through), so presigned PUTs for `object` also target `*.amazonaws.com`. The asymmetric branch here keeps suppressing the default S3 origins for `object` whenever a hostname is set, reproducing the original bug for that configuration. See the inline comment on `getDefaultS3CspSources`.
2. **No regression test for fitness storage.** The same code path is reused for `fitnessStorage`, but only `mediaStorage` has a `hostname + s3` test. Adding a parallel `ACTIVITIES_FITNESS_STORAGE_*` test would prevent the fitness branch from silently regressing.

Other than these, the refactor of the conditional into the named `allowsDefaultS3Sources` is a readability win, and the `hasCustomHostname` extraction is good.

---

next.config.test.ts
llun: Fixed in 946affb18549db2f164641fab8654228e2ef6573: added the explicit ACTIVITIES_FITNESS_STORAGE_TYPE=s3 plus custom fitness hostname CSP regression test so the default AWS S3 presigned upload hosts stay in connect-src for fitness storage. I also added the symmetric fitness object-storage endpoint CSP test to cover the new endpoint-vs-public-hostname split.

lib/utils/securityHeaders.ts
llun: Fixed in bb21388c55c1b59df9f8063f811e83a39f1a7d54: added separate ACTIVITIES_MEDIA_STORAGE_ENDPOINT and ACTIVITIES_FITNESS_STORAGE_ENDPOINT config for the S3-compatible API/presigned upload endpoint while keeping *_STORAGE_HOSTNAME as the public media/CDN hostname. CSP now includes endpoint origins for uploads and falls back to AWS S3 origins when no endpoint is configured; the media/fitness S3 clients, Strava archive read path, and maintenance/archive scripts all use endpoint for S3 API access. This intentionally does not treat public hostnames like static.llun.social as S3 API endpoints because those may be CloudFront/public serving hosts only.

llun:
@codex review

lib/services/storage/s3Client.ts
```diff
@@ -0,0 +1,36 @@
...
+export const normalizeStorageEndpoint = (endpoint: string) => {
+  const trimmedEndpoint = endpoint.trim()
+  const url = new URL(
+    trimmedEndpoint.includes('://')
+      ? trimmedEndpoint
+      : `https://${trimmedEndpoint}`
+  )
+
+  return `${url.protocol}//${url.host}`
+}
```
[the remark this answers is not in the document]
llun: Addressed in 8aa4404c6e8a97a15ec7c7a891c552b6881ccc28: added regression tests for plain local endpoints (`localhost:9000`, `127.0.0.1`) and preserved endpoint path segments during normalization. Plain hostnames were already prefixed before calling `new URL`; the test now pins that behavior. I did not add Unix socket endpoint support because this storage endpoint is an S3-compatible HTTP(S)/host endpoint, not a filesystem socket connection.

lib/services/storage/s3Client.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const normalizeStorageEndpoint = (endpoint: string) => {
+  const trimmedEndpoint = endpoint.trim()
+  const url = new URL(
+    trimmedEndpoint.includes('://')
+      ? trimmedEndpoint
+      : `https://${trimmedEndpoint}`
+  )
+
+  return `${url.protocol}//${url.host}`
```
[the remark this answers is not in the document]
llun: Fixed in 8aa4404c6e8a97a15ec7c7a891c552b6881ccc28: `normalizeStorageEndpoint` now keeps the URL pathname while trimming trailing slashes, so path-based S3-compatible endpoints like `https://storage.example.com/s3` stay intact. Added direct tests for endpoints with and without a protocol.

scripts/productionArchive.ts
```diff
@@ -575,8 +578,8 @@ const getSharedS3FitnessPrefix = (
   if (mediaStorage.bucket !== fitnessStorage.bucket) return null
   if (mediaStorage.region !== fitnessStorage.region) return null
   if (
-    getStorageEndpointIdentity(mediaStorage.hostname) !==
-    getStorageEndpointIdentity(fitnessStorage.hostname)
+    getStorageEndpointIdentity(mediaStorage.endpoint) !==
+    getStorageEndpointIdentity(fitnessStorage.endpoint)
```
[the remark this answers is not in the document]
llun: Fixed in 8aa4404c6e8a97a15ec7c7a891c552b6881ccc28: the shared S3 fitness-prefix check now compares the explicit endpoint, or the legacy object-storage hostname fallback when no endpoint is configured. Added a regression test so hostname-only object storage with different hostnames no longer deduplicates `fitness/` paths incorrectly.

scripts/productionArchive.ts
```diff
@@ -1135,19 +1138,19 @@ export const getStorageEndpoint = (hostname: string) => {
...
   source: Extract<StorageSource, { kind: 's3' }>
 ) =>
   new S3Client({
     region: source.region,
-    ...(source.hostname
+    ...(source.endpoint
       ? {
-          endpoint: getStorageEndpoint(source.hostname),
+          endpoint: getStorageEndpoint(source.endpoint),
           forcePathStyle: true
```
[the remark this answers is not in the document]
llun: Fixed in 8aa4404c6e8a97a15ec7c7a891c552b6881ccc28: `productionArchive` now carries an `endpointFallback` for legacy object-storage configs that only have `hostname`, and `createS3Client` uses `endpoint ?? endpointFallback`. The fallback is intentionally object-storage-only so AWS S3/public-CDN hostnames are not treated as API endpoints again.

llun:
@codex review

lib/utils/securityHeaders.ts
```diff
@@ -56,11 +57,16 @@ const isS3CompatibleStorage = (
...
+const getStorageEndpoint = (storage: { endpoint?: string }) => storage.endpoint
 
 const getDefaultS3CspSources = (storage: unknown) => {
   if (!isS3CompatibleStorage(storage)) return []
 
-  if (!['s3', 'object'].includes(storage.type) || storage.hostname?.trim()) {
+  const hasCustomEndpoint = Boolean(storage.endpoint?.trim())
+  const allowsDefaultS3Sources =
+    ['s3', 'object'].includes(storage.type) && !hasCustomEndpoint
+
```
llun, line 68: Minor: when `type=object` is configured without a hostname *and* without an endpoint (legacy operator running an S3-compatible service who never realized the runtime always uploaded to AWS), `getDefaultS3CspSources` will now still emit `https://<bucket>.s3.<region>.amazonaws.com` even for regions like `auto` (Cloudflare R2 idiom). That produces a meaningless `https://<bucket>.s3.auto.amazonaws.com` in `connect-src`. It's harmless (no real host to talk to), but you may want to skip the defaults when `region === 'auto'` since that strongly implies an S3-compatible provider that requires an explicit endpoint anyway. Not blocking.

---

lib/services/storage/s3Client.ts
```diff
@@ -0,0 +1,37 @@
...
+  const normalizedEndpoint = endpoint?.trim()
+    ? normalizeStorageEndpoint(endpoint)
+    : undefined
+
+  return new S3Client({
+    region,
+    ...(normalizedEndpoint
+      ? {
+          endpoint: normalizedEndpoint,
+          forcePathStyle: true
```
llun, line 33: `forcePathStyle: true` is hard-coded whenever an endpoint is set. That's correct for MinIO and the common R2/DO Spaces configurations, but it permanently locks out subdomain-style virtual-hosting (some operators front S3-compatible storage with a CDN that requires virtual-hosted-style addressing). Consider plumbing `forcePathStyle` from config so operators can opt out without a code change. Not blocking — fine to ship as-is and revisit if anyone reports it.

---

scripts/productionArchive.ts
```diff
@@ -530,10 +532,22 @@ const storageSourceFromMediaConfig = (
...
+      }
+    case MediaStorageType.ObjectStorage:
+      return {
+        bucket: storage.bucket,
+        ...(storage.endpoint ? { endpoint: storage.endpoint } : null),
+        ...(!storage.endpoint && storage.hostname
+          ? { endpointFallback: storage.hostname }
+          : null),
         ...(storage.hostname ? { hostname: storage.hostname } : null),
         kind: 's3',
```
llun, line 553: The `endpointFallback` is asymmetric — it's wired only for `MediaStorageType.ObjectStorage` / `FitnessStorageType.ObjectStorage`, not for `S3Storage`. That asymmetry deserves a one-line comment here explaining the reasoning, because the code path looks inconsistent at first read.

The reasoning, as I understand it: the *runtime* `S3StorageFile` always constructed `new S3Client({ region })` without an endpoint, so a legacy `s3`-type hostname was never actually used as an upload endpoint at runtime — only the production-archive script repurposed it that way. The fallback preserves that archive-only legacy behavior for `object` storage (where it was plausibly intentional) but not for `s3` (where it was almost certainly accidental). Worth one comment so future readers don't try to "fix" the symmetry.

---

docs/environment-variables.md
```diff
@@ -127,7 +127,8 @@ Required for media uploads (images and video in posts). If no media storage is c
...
 | `ACTIVITIES_MEDIA_STORAGE_BUCKET`            | S3 bucket name (for `s3` or `object`).                                                                                                                                                   |
 | `ACTIVITIES_MEDIA_STORAGE_REGION`            | S3 region (e.g., `us-east-1`).                                                                                                                                                           |
-| `ACTIVITIES_MEDIA_STORAGE_HOSTNAME`          | Custom S3 endpoint hostname (for S3-compatible services like MinIO, DigitalOcean Spaces).                                                                                                |
+| `ACTIVITIES_MEDIA_STORAGE_HOSTNAME`          | Public media hostname/CDN used to serve stored media files.                                                                                                                              |
+| `ACTIVITIES_MEDIA_STORAGE_ENDPOINT`          | S3-compatible API endpoint used for storage operations and browser presigned uploads (for services like MinIO, DigitalOcean Spaces, Cloudflare R2).                                      |
```
llun: Existing operators with `ACTIVITIES_MEDIA_STORAGE_TYPE=object` plus `ACTIVITIES_MEDIA_STORAGE_HOSTNAME` previously had a misleading documentation: the doc here used to say "Custom S3 endpoint hostname (for S3-compatible services like MinIO, DigitalOcean Spaces)." In practice, the runtime `S3StorageFile` ignored that value as an endpoint and uploads always went to AWS — so anyone relying on the documented behavior was silently broken.

This PR changes the documentation contract without a release note. Please consider adding an upgrade note (here or in the PR description) explicitly calling out: "If you were setting `ACTIVITIES_*_STORAGE_HOSTNAME` to your MinIO/R2/DO Spaces API endpoint, move that value to `ACTIVITIES_*_STORAGE_ENDPOINT` to get the intended behavior." Since this commit is prefixed `fix:` and not `minor:`, the existing-operator-facing config change is easy to miss.

---

next.config.test.ts
```diff
@@ -379,6 +401,32 @@ describe('next config security hardening', () => {
...
+            'https://storage.example.com'
+          ])
+        )
+        expect(connectSources).not.toContain(
+          'https://media-object-bucket.s3.auto.amazonaws.com'
+        )
+        expect(connectSources).not.toContain('https://s3.auto.amazonaws.com')
+      }
+    )
+  })
```
llun, line 476: Nit: the new test only exercises `type: 'object'` + endpoint. A parallel test for `type: 's3'` + a non-AWS endpoint would help pin down that the `s3` type also respects an explicit endpoint and correctly suppresses the AWS defaults (the `getDefaultS3CspSources` logic now keys off endpoint alone regardless of type, so it's worth a regression anchor).

---

llun (commented):
The PR has grown well beyond the original CSP one-liner — it now cleanly separates `HOSTNAME` (public CDN) from `ENDPOINT` (S3-compatible API), consolidates `S3Client` construction in a shared `createStorageS3Client` helper, and adds the symmetric fitness-storage tests I asked for previously. Both of my earlier blocking concerns are addressed.

A few remaining nits, none blocking:

1. **Hard-coded `forcePathStyle: true`** in `createStorageS3Client` — fine for MinIO/R2/DO Spaces but locks out subdomain-style virtual-hosting providers. Worth making configurable later.
2. **`region: 'auto'` with no endpoint** still emits a nonsense `https://<bucket>.s3.auto.amazonaws.com` source. Harmless but ugly; skipping the AWS defaults when region is `auto` would be tidier.
3. **`endpointFallback` asymmetry** in `productionArchive.ts` (only for `Object` types) deserves a one-line comment so a future reader doesn't try to "fix" the inconsistency.
4. **Missing migration note** — the docs change re-defines what `ACTIVITIES_*_STORAGE_HOSTNAME` means for `object` storage. The old wording said "Custom S3 endpoint hostname (for S3-compatible services like MinIO, ...)", and operators who took that at face value need to be told to move their value to `ENDPOINT`. Note that the *runtime* never actually used hostname-as-endpoint (the old `new S3Client({ region })` ignored it), so the actual behavior change is small, but the documented contract is changing. A line in the PR description / release notes would help.
5. **One additional regression test** worth adding: `type: 's3'` + a non-AWS `ENDPOINT`. The current tests cover `type: 'object'` + endpoint and `type: 's3'` without endpoint, but not `s3` + endpoint. Since `getDefaultS3CspSources` now keys off endpoint regardless of type, anchoring that with a test would prevent a future regression.

Overall: looks good. Code quality, test coverage, and design are all improved over the first revision.

---

docs/environment-variables.md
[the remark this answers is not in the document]
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: documented media HOSTNAME/ENDPOINT defaults precisely. HOSTNAME is the public/CDN serving hostname and, when unset, files are served through the app from the configured storage backend; ENDPOINT is the S3 API/presigned-upload endpoint and, when unset, the AWS SDK uses the standard AWS S3 endpoint for the configured region.

docs/environment-variables.md
```diff
@@ -141,7 +142,8 @@ For fitness activity file uploads (.fit, .gpx, .tcx). Falls back to media storag
...
-| `ACTIVITIES_FITNESS_STORAGE_HOSTNAME`                       | Custom S3 endpoint hostname.                                                                                                                                                                                                                                       |
+| `ACTIVITIES_FITNESS_STORAGE_HOSTNAME`                       | Public fitness file hostname/CDN used to serve stored fitness files.                                                                                                                                                                                               |
+| `ACTIVITIES_FITNESS_STORAGE_ENDPOINT`                       | S3-compatible API endpoint used for storage operations and browser presigned uploads.                                                                                                                                                                              |
```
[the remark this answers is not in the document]
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: documented the fitness HOSTNAME/ENDPOINT fallback behavior. Fitness only inherits media storage values when fitness storage is not explicitly configured; with explicit fitness storage, an unset hostname means no separate public fitness hostname and app serving is used, while an unset endpoint means the AWS SDK uses the standard AWS regional endpoint.

scripts/cleanupMediaStorage.ts
```diff
@@ -332,7 +339,8 @@ async function cleanupMediaStorage() {
           await deleteS3File(
             config.mediaStorage.bucket,
             config.mediaStorage.region,
-            file
+            file,
+            config.mediaStorage.endpoint
           )
           break
```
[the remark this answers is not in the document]
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: cleanupMediaStorage now reuses a single S3Client for the deletion loop and destroys S3 clients after list/delete operations.

lib/utils/securityHeaders.ts
llun: Addressed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: skipped default AWS S3 CSP sources when region is `auto`, with regression coverage for object storage `region=auto` without an explicit endpoint.

lib/services/storage/s3Client.ts
llun: Not changed in this loop. Making `forcePathStyle` configurable would add a new runtime config surface plus docs/tests, and the review marked this nonblocking. The current default still matches the common S3-compatible endpoint setups this PR is fixing (MinIO, R2, DigitalOcean Spaces), so I’m leaving this as future config work rather than expanding this fix PR.

scripts/productionArchive.ts
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: added archive-only compatibility comments above both object-storage `endpointFallback` blocks so future readers can see why the fallback is intentionally asymmetric and not applied to `s3`.

docs/environment-variables.md
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: added an upgrade note telling operators to move S3-compatible API endpoint values from `*_STORAGE_HOSTNAME` to `*_STORAGE_ENDPOINT`, and clarified that HOSTNAME is for public/CDN serving rather than S3 API operations or browser presigned uploads.

next.config.test.ts
llun: Fixed in 39168aac71a97e71946a348a93f12b3bc4ccfd59: added a `type=s3` plus non-AWS endpoint CSP regression test that asserts the explicit endpoint and public hostname are allowed in `connect-src` while default AWS S3 hosts are suppressed.

llun:
@codex review

lib/utils/securityHeaders.ts
```diff
@@ -56,17 +57,22 @@ const isS3CompatibleStorage = (
...
+  if (!allowsDefaultS3Sources) {
     return []
   }
 
   const bucket = storage.bucket.trim()
   const region = storage.region.trim()
-  if (!bucket || !region) return []
+  if (!bucket || !region || region.toLowerCase() === 'auto') return []
```
llun, line 75: Nit: the bare string `'auto'` is a magic value tied to the Cloudflare R2 idiom (R2 always uses `region: 'auto'` and requires an explicit endpoint). A short comment here — `// 'auto' is the Cloudflare R2 region marker and never resolves to a real AWS S3 endpoint` — would save the next reader from having to git-blame this line. Not blocking.

---

scripts/cleanupMediaStorage.ts
```diff
@@ -316,37 +329,48 @@ async function cleanupMediaStorage() {
...
+            break
+          case MediaStorageType.S3Storage:
+          case MediaStorageType.ObjectStorage:
+            if (!deleteS3Client) {
+              throw new Error('S3 client is not defined for S3 storage')
+            }
+            await deleteS3File(deleteS3Client, config.mediaStorage.bucket, file)
+            break
+        }
+        deletedCount++
```
llun, line 360: Nice catch on lifting the S3 client out of the per-file loop — previously this constructed a new `S3Client` for every orphan deletion, which is wasteful and (depending on SDK internals) leaks short-lived HTTP agents. The single-construction + `try/finally` destroy pattern is the right shape here.

The `if (!deleteS3Client) throw` guard is defensive against a case that's logically unreachable (we only enter this `case` when type is S3/Object, which is exactly when `deleteS3Client` is constructed). TypeScript can't narrow it from the outer scope so the runtime check is the pragmatic fix. Fine as-is.

---

llun (commented):
All five round-2 nits addressed:

1. ✅ `region === 'auto'` now skips the AWS S3 defaults, with a dedicated regression test.
2. ✅ Added `type: 's3'` + non-AWS endpoint test pinning the new behavior.
3. ✅ One-line comment added at both `storageSourceFromMediaConfig` / `storageSourceFromFitnessConfig` explaining the `endpointFallback` rationale.
4. ✅ Upgrade note added to `docs/environment-variables.md` instructing operators to move misconfigured `HOSTNAME` values to `ENDPOINT`.
5. (Not addressed by design — `forcePathStyle: true` is still hard-coded, which I marked non-blocking.)

Bonus: `cleanupMediaStorage.ts` is also refactored to construct the S3 client once outside the deletion loop with proper `try/finally` cleanup — a real efficiency win unrelated to the original bug.

LGTM. Only nit left is a one-line clarifying comment on the magic `'auto'` string (inline). Feel free to ship without it.

---

PR #705, opened 2026-05-16
fix: prevent mobile post actions overflow

llun:
## Summary
- Rebalanced mobile post actions into two equal-spaced grid rows: reply, repost, like, and edit history on the first row; visibility, edit, and delete on a left-aligned second row.
- Kept desktop behavior as a single-line flex layout at `sm` and above.
- Added accessible group labels and explicit icon-button labels, including expanded/controls state for edit history.
- Added an explicit close control for the mobile edit-history dialog.

## Root cause
The previous mobile layout kept every post action in one horizontal flex row. Owner-only controls pushed the row wider than the post content, so the rightmost delete action could overflow and become hidden or unreachable.

## Mobile screenshot
Mobile post actions at 390x844

## Browser verification
- Started `yarn dev` with Node 24 and disposable local SQLite runtime env.
- Opened `http://localhost:3000` in Browser at `390x844`.
- Seeded and signed in as a local owner account.
- Verified owner post shows all seven actions: reply, repost, like, edit history, visibility, edit, and delete.
- Verified first action row contains 4 controls and second action row contains 3 left-aligned controls.
- Verified no horizontal overflow: viewport width `390`, document scroll width `375`.
- Verified both action rows stay within post content width: row width `257` <= post content width `257`.
- Verified the edit-history dialog opens, exposes a close button, and returns focus to the trigger after close.

## Tests
- `yarn run prettier --write .` - pass
- `yarn test lib/components/posts/post.test.tsx` - pass (19 tests)
- `yarn lint` - pass (existing warnings only in auth/fitness files)
- `yarn build` - pass
- `yarn test` - pass (297 suites, 2375 tests)

## Config and migrations
No config, API, database, or migration changes.

### DIFF
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -6,7 +6,6 @@ import {
   StatusType,
   getOriginalStatus
 } from '@/lib/types/domain/status'
-import { cn } from '@/lib/utils'
 
 import { DeleteButton } from './delete-button'
 import { EditButton } from './edit-button'
@@ -20,12 +19,8 @@ interface Props extends PostProps {
   onShowEdits?: (status: Status) => void
 }
 
-const statusActionGridColumnsByCount: Record<number, string> = {
-  1: 'grid-cols-1',
-  2: 'grid-cols-2',
-  3: 'grid-cols-3',
-  4: 'grid-cols-4'
-}
+const actionRowClassName =
+  'grid w-full grid-cols-4 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6'
 
 export const Actions: FC<Props> = ({
   host,
@@ -50,10 +45,19 @@ export const Actions: FC<Props> = ({
     Boolean(actualStatus.isLocalActor) &&
     currentActor.id === actualStatus.actorId
   const hasEditHistory = actualStatus.edits.length > 0
-  const statusActions: ReactNode[] = []
+  const primaryActions: ReactNode[] = [
+    <ReplyButton key="reply" status={actualStatus} onReply={onReply} />,
+    <RepostButton
+      key="repost"
+      currentActor={currentActor}
+      status={actualStatus}
+    />,
+    <LikeButton key="like" currentActor={currentActor} status={actualStatus} />
+  ]
+  const secondaryActions: ReactNode[] = []
 
   if (hasEditHistory) {
-    statusActions.push(
+    primaryActions.push(
       <EditHistoryButton
         key="edit-history"
         status={actualStatus}
@@ -64,13 +68,13 @@ export const Actions: FC<Props> = ({
   }
 
   if (isOwner) {
-    statusActions.push(
+    secondaryActions.push(
       <VisibilityButton key="visibility" status={actualStatus} />
     )
   }
 
   if (canEdit) {
-    statusActions.push(
+    secondaryActions.push(
       <EditButton key="edit" status={actualStatus} onEdit={onEdit} />,
       <DeleteButton
         key="delete"
@@ -80,32 +84,25 @@ export const Actions: FC<Props> = ({
     )
   }
 
-  const hasStatusActions = statusActions.length > 0
-  const statusActionGridColumns =
-    statusActionGridColumnsByCount[statusActions.length] ?? 'grid-cols-1'
+  const hasSecondaryActions = secondaryActions.length > 0
 
   return (
     <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
       <div
         role="group"
-        aria-label="Post social actions"
-        className="grid w-full grid-cols-3 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6"
+        aria-label="Post primary actions"
+        className={actionRowClassName}
       >
-        <ReplyButton status={actualStatus} onReply={onReply} />
-        <RepostButton currentActor={currentActor} status={actualStatus} />
-        <LikeButton currentActor={currentActor} status={actualStatus} />
+        {primaryActions}
       </div>
 
-      {hasStatusActions && (
+      {hasSecondaryActions && (
         <div
           role="group"
-          aria-label="Post status actions"
-          className={cn(
-            'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6',
-            statusActionGridColumns
-          )}
+          aria-label="Post secondary actions"
+          className={actionRowClassName}
         >
-          {statusActions}
+          {secondaryActions}
         </div>
       )}
     </div>
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -86,16 +86,14 @@ export const Actions: FC<Props> = ({
   }
 
   const hasSecondaryActions = secondaryActions.length > 0
+  const actionColumnClassName = hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'
 
   return (
     <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
       <div
         role="group"
         aria-label="Post primary actions"
-        className={cn(
-          actionRowClassName,
-          hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'
-        )}
+        className={cn(actionRowClassName, actionColumnClassName)}
       >
         {primaryActions}
       </div>
@@ -104,7 +102,7 @@ export const Actions: FC<Props> = ({
         <div
           role="group"
           aria-label="Post secondary actions"
-          className={cn(actionRowClassName, 'grid-cols-4')}
+          className={cn(actionRowClassName, actionColumnClassName)}
         >
           {secondaryActions}
         </div>
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -6,6 +6,7 @@ import {
   StatusType,
   getOriginalStatus
 } from '@/lib/types/domain/status'
+import { cn } from '@/lib/utils'
 
 import { DeleteButton } from './delete-button'
 import { EditButton } from './edit-button'
@@ -20,7 +21,7 @@ interface Props extends PostProps {
 }
 
 const actionRowClassName =
-  'grid w-full grid-cols-4 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6'
+  'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6'
 
 export const Actions: FC<Props> = ({
   host,
@@ -91,7 +92,10 @@ export const Actions: FC<Props> = ({
       <div
         role="group"
         aria-label="Post primary actions"
-        className={actionRowClassName}
+        className={cn(
+          actionRowClassName,
+          hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'
+        )}
       >
         {primaryActions}
       </div>
@@ -100,7 +104,7 @@ export const Actions: FC<Props> = ({
         <div
           role="group"
           aria-label="Post secondary actions"
-          className={actionRowClassName}
+          className={cn(actionRowClassName, 'grid-cols-4')}
         >
           {secondaryActions}
         </div>
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -88,6 +88,7 @@ export const Actions: FC<Props> = ({
   }
 
   const hasSecondaryActions = secondaryActions.length > 0
+  // Use the same mobile grid for both rows so secondary controls align under primary controls.
   const actionColumnClassName = hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'
 
   return (
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react'
+import { FC, ReactNode } from 'react'
 
 import { PostProps } from '@/lib/components/posts/post'
 import {
@@ -43,15 +43,43 @@ export const Actions: FC<Props> = ({
     Boolean(actualStatus.isLocalActor) &&
     currentActor.id === actualStatus.actorId
   const hasEditHistory = actualStatus.edits.length > 0
-  const hasStatusActions = hasEditHistory || isOwner || canEdit
-  const statusActionCount =
-    (hasEditHistory ? 1 : 0) + (isOwner ? 1 : 0) + (canEdit ? 2 : 0)
+  const statusActions: ReactNode[] = []
+
+  if (hasEditHistory) {
+    statusActions.push(
+      <EditHistoryButton
+        key="edit-history"
+        status={actualStatus}
+        host={host}
+        onShowEdits={onShowEdits}
+      />
+    )
+  }
+
+  if (isOwner) {
+    statusActions.push(
+      <VisibilityButton key="visibility" status={actualStatus} />
+    )
+  }
+
+  if (canEdit) {
+    statusActions.push(
+      <EditButton key="edit" status={actualStatus} onEdit={onEdit} />,
+      <DeleteButton
+        key="delete"
+        status={actualStatus}
+        onPostDeleted={onPostDeleted}
+      />
+    )
+  }
+
+  const hasStatusActions = statusActions.length > 0
   const statusActionGridColumns =
-    statusActionCount === 4
+    statusActions.length === 4
       ? 'grid-cols-4'
-      : statusActionCount === 3
+      : statusActions.length === 3
         ? 'grid-cols-3'
-        : statusActionCount === 2
+        : statusActions.length === 2
           ? 'grid-cols-2'
           : 'grid-cols-1'
 
@@ -76,23 +104,7 @@ export const Actions: FC<Props> = ({
             statusActionGridColumns
           )}
         >
-          {hasEditHistory && (
-            <EditHistoryButton
-              status={actualStatus}
-              host={host}
-              onShowEdits={onShowEdits}
-            />
-          )}
-          {isOwner && <VisibilityButton status={actualStatus} />}
-          {canEdit && (
-            <>
-              <EditButton status={actualStatus} onEdit={onEdit} />
-              <DeleteButton
-                status={actualStatus}
-                onPostDeleted={onPostDeleted}
-              />
-            </>
-          )}
+          {statusActions}
         </div>
       )}
     </div>
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -6,6 +6,7 @@ import {
   StatusType,
   getOriginalStatus
 } from '@/lib/types/domain/status'
+import { cn } from '@/lib/utils'
 
 import { DeleteButton } from './delete-button'
 import { EditButton } from './edit-button'
@@ -43,13 +44,23 @@ export const Actions: FC<Props> = ({
     currentActor.id === actualStatus.actorId
   const hasEditHistory = actualStatus.edits.length > 0
   const hasStatusActions = hasEditHistory || isOwner || canEdit
+  const statusActionCount =
+    (hasEditHistory ? 1 : 0) + (isOwner ? 1 : 0) + (canEdit ? 2 : 0)
+  const statusActionGridColumns =
+    statusActionCount === 4
+      ? 'grid-cols-4'
+      : statusActionCount === 3
+        ? 'grid-cols-3'
+        : statusActionCount === 2
+          ? 'grid-cols-2'
+          : 'grid-cols-1'
 
   return (
     <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
       <div
         role="group"
         aria-label="Post social actions"
-        className="flex w-full items-center justify-between gap-2 sm:w-auto sm:justify-start sm:gap-6"
+        className="grid w-full grid-cols-3 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6"
       >
         <ReplyButton status={actualStatus} onReply={onReply} />
         <RepostButton currentActor={currentActor} status={actualStatus} />
@@ -60,7 +71,10 @@ export const Actions: FC<Props> = ({
         <div
           role="group"
           aria-label="Post status actions"
-          className="flex w-full items-center justify-end gap-2 sm:w-auto sm:justify-start sm:gap-6"
+          className={cn(
+            'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6',
+            statusActionGridColumns
+          )}
         >
           <EditHistoryButton
             status={actualStatus}
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -20,6 +20,13 @@ interface Props extends PostProps {
   onShowEdits?: (status: Status) => void
 }
 
+const statusActionGridColumnsByCount: Record<number, string> = {
+  1: 'grid-cols-1',
+  2: 'grid-cols-2',
+  3: 'grid-cols-3',
+  4: 'grid-cols-4'
+}
+
 export const Actions: FC<Props> = ({
   host,
   currentActor,
@@ -75,13 +82,7 @@ export const Actions: FC<Props> = ({
 
   const hasStatusActions = statusActions.length > 0
   const statusActionGridColumns =
-    statusActions.length === 4
-      ? 'grid-cols-4'
-      : statusActions.length === 3
-        ? 'grid-cols-3'
-        : statusActions.length === 2
-          ? 'grid-cols-2'
-          : 'grid-cols-1'
+    statusActionGridColumnsByCount[statusActions.length] ?? 'grid-cols-1'
 
   return (
     <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -26,6 +26,7 @@ const actionRowClassName =
 export const Actions: FC<Props> = ({
   host,
   currentActor,
+  currentTime,
   status,
   editable = false,
   showActions = false,
@@ -63,6 +64,7 @@ export const Actions: FC<Props> = ({
         key="edit-history"
         status={actualStatus}
         host={host}
+        currentTime={currentTime}
         onShowEdits={onShowEdits}
       />
     )
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -41,27 +41,44 @@ export const Actions: FC<Props> = ({
   const isOwner =
     Boolean(actualStatus.isLocalActor) &&
     currentActor.id === actualStatus.actorId
+  const hasEditHistory = actualStatus.edits.length > 0
+  const hasStatusActions = hasEditHistory || isOwner || canEdit
 
   return (
-    <div className="mt-3 flex items-center gap-6 text-muted-foreground">
-      <ReplyButton status={actualStatus} onReply={onReply} />
-      <RepostButton currentActor={currentActor} status={actualStatus} />
-      <LikeButton currentActor={currentActor} status={actualStatus} />
-
-      <div className="flex items-center gap-6">
-        <EditHistoryButton
-          status={actualStatus}
-          host={host}
-          onShowEdits={onShowEdits}
-        />
-        {isOwner && <VisibilityButton status={actualStatus} />}
-        {canEdit && (
-          <>
-            <EditButton status={actualStatus} onEdit={onEdit} />
-            <DeleteButton status={actualStatus} onPostDeleted={onPostDeleted} />
-          </>
-        )}
+    <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
+      <div
+        role="group"
+        aria-label="Post social actions"
+        className="flex w-full items-center justify-between gap-2 sm:w-auto sm:justify-start sm:gap-6"
+      >
+        <ReplyButton status={actualStatus} onReply={onReply} />
+        <RepostButton currentActor={currentActor} status={actualStatus} />
+        <LikeButton currentActor={currentActor} status={actualStatus} />
       </div>
+
+      {hasStatusActions && (
+        <div
+          role="group"
+          aria-label="Post status actions"
+          className="flex w-full items-center justify-end gap-2 sm:w-auto sm:justify-start sm:gap-6"
+        >
+          <EditHistoryButton
+            status={actualStatus}
+            host={host}
+            onShowEdits={onShowEdits}
+          />
+          {isOwner && <VisibilityButton status={actualStatus} />}
+          {canEdit && (
+            <>
+              <EditButton status={actualStatus} onEdit={onEdit} />
+              <DeleteButton
+                status={actualStatus}
+                onPostDeleted={onPostDeleted}
+              />
+            </>
+          )}
+        </div>
+      )}
     </div>
   )
 }
--- a/lib/components/posts/actions/actions.tsx
+++ b/lib/components/posts/actions/actions.tsx
@@ -76,11 +76,13 @@ export const Actions: FC<Props> = ({
             statusActionGridColumns
           )}
         >
-          <EditHistoryButton
-            status={actualStatus}
-            host={host}
-            onShowEdits={onShowEdits}
-          />
+          {hasEditHistory && (
+            <EditHistoryButton
+              status={actualStatus}
+              host={host}
+              onShowEdits={onShowEdits}
+            />
+          )}
           {isOwner && <VisibilityButton status={actualStatus} />}
           {canEdit && (
             <>
--- a/lib/components/posts/actions/edit-history-button.tsx
+++ b/lib/components/posts/actions/edit-history-button.tsx
@@ -80,7 +80,7 @@ export const EditHistoryButton: FC<Props> = ({
             {edits.map((edit, index) => {
               return (
                 <li
-                  key={edit.createdAt + index}
+                  key={`${edit.createdAt}-${index}`}
                   className="flex flex-col items-start p-3"
                 >
                   <div className="self-end bg-primary text-primary-foreground px-3 py-1 rounded-full text-xs">
--- a/lib/components/posts/actions/edit-history-button.tsx
+++ b/lib/components/posts/actions/edit-history-button.tsx
@@ -23,19 +23,24 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
   }`
 
   return (
-    <button
-      className="relative flex items-center gap-1.5 rounded-full px-2 py-1 text-sm hover:bg-muted transition-colors"
-      onClick={(e) => {
-        e.stopPropagation()
-        onShowEdits?.(status)
-        setShowHistory((value) => !value)
-      }}
-      title={editCountLabel}
-      aria-label={`Show edit history, ${editCountLabel}`}
-    >
-      <History className="h-4 w-4" />
+    <div className="relative inline-flex">
+      <button
+        className="flex items-center gap-1.5 rounded-full px-2 py-1 text-sm hover:bg-muted transition-colors"
+        onClick={(e) => {
+          e.stopPropagation()
+          onShowEdits?.(status)
+          setShowHistory((value) => !value)
+        }}
+        title={editCountLabel}
+        aria-label={`Show edit history, ${editCountLabel}`}
+      >
+        <History className="h-4 w-4" />
+      </button>
       {showHistory && (
-        <div className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto">
+        <div
+          className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto"
+          onClick={(e) => e.stopPropagation()}
+        >
           <ul className="divide-y divide-border rounded-lg border bg-background shadow-lg">
             {status.edits.reverse().map((edit, index) => {
               return (
@@ -59,6 +64,6 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
           </ul>
         </div>
       )}
-    </button>
+    </div>
   )
 }
--- a/lib/components/posts/actions/edit-history-button.tsx
+++ b/lib/components/posts/actions/edit-history-button.tsx
@@ -23,6 +23,7 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
   const editCountLabel = `${status.edits.length} ${
     status.edits.length === 1 ? 'edit' : 'edits'
   }`
+  const edits = [...status.edits].reverse()
   const closeHistory = () => {
     setShowHistory(false)
     triggerRef.current?.focus()
@@ -70,7 +71,7 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
             </button>
           </div>
           <ul className="max-h-80 divide-y divide-border overflow-auto max-md:max-h-[calc(100vh-10rem)]">
-            {status.edits.reverse().map((edit, index) => {
+            {edits.map((edit, index) => {
               return (
                 <li
                   key={edit.createdAt + index}
--- a/lib/components/posts/actions/edit-history-button.tsx
+++ b/lib/components/posts/actions/edit-history-button.tsx
@@ -1,6 +1,6 @@
 import { formatDistance } from 'date-fns'
-import { History } from 'lucide-react'
-import { FC, useState } from 'react'
+import { History, X } from 'lucide-react'
+import { FC, useId, useState } from 'react'
 
 import { Status, StatusNote, StatusPoll } from '@/lib/types/domain/status'
 import { cleanClassName } from '@/lib/utils/text/cleanClassName'
@@ -15,6 +15,7 @@ interface Props {
 
 export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
   const [showHistory, setShowHistory] = useState<boolean>(false)
+  const editHistoryId = useId()
 
   if (status.edits.length === 0) return null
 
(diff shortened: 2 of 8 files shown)


llun:
@codex review

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +42,57 @@ export const Actions: FC<Props> = ({
...
+  const statusActionCount =
+    (hasEditHistory ? 1 : 0) + (isOwner ? 1 : 0) + (canEdit ? 2 : 0)
+  const statusActionGridColumns =
+    statusActionCount === 4
+      ? 'grid-cols-4'
+      : statusActionCount === 3
+        ? 'grid-cols-3'
+        : statusActionCount === 2
+          ? 'grid-cols-2'
+          : 'grid-cols-1'
```
llun: The `statusActionCount` manually mirrors the number of items rendered inside the status action group. This creates a hidden coupling: the count is correct only because `EditHistoryButton` internally returns `null` when `status.edits.length === 0` rather than being conditionally rendered at this call site.

If someone later wraps or moves that null-guard, or adds a new button to the group, the grid column count silently goes wrong.

A more robust alternative that avoids manual counting entirely:

```tsx
// Instead of statusActionCount + statusActionGridColumns,
// just use auto-fit so CSS handles it:
className="grid w-full auto-cols-fr grid-flow-col items-center ..."
```

Or, conditionally render `EditHistoryButton` here (guarded by `hasEditHistory`) and use `grid-cols-[subgrid]` / `auto-fit`. At minimum, add a comment explaining why the count matches the rendered children, so a future maintainer knows to keep them in sync.

---

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +42,57 @@ export const Actions: FC<Props> = ({
...
+          className={cn(
+            'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6',
+            statusActionGridColumns
+          )}
+        >
+          <EditHistoryButton
+            status={actualStatus}
+            host={host}
+            onShowEdits={onShowEdits}
+          />
```
llun: `EditHistoryButton` is rendered unconditionally inside the `hasStatusActions` block, but it returns `null` internally when `status.edits.length === 0`. Meanwhile, the `statusActionCount` above correctly excludes it when `hasEditHistory` is false.

This creates a subtle two-layer null check: the parent decides the grid columns based on `hasEditHistory`, and the child decides whether to render based on the same condition — they just happen to agree.

Prefer explicit conditional rendering here to make the contract visible and keep it consistent with how `VisibilityButton` and the edit/delete pair are guarded:

```tsx
{hasEditHistory && (
  <EditHistoryButton
    status={actualStatus}
    host={host}
    onShowEdits={onShowEdits}
  />
)}
```

This way, a reader does not need to look inside `EditHistoryButton` to understand what renders in which state.

---

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -27,6 +27,7 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
         setShowHistory((value) => !value)
       }}
       title={`${status.edits.length} edits`}
+      aria-label="Show edit history"
```
llun: The `title` is dynamic and includes the count (`${status.edits.length} edits`), but `aria-label` is static and just says `"Show edit history"`. Screen readers use `aria-label` in preference to `title`, so a screen reader user won't hear the edit count.

These two should be aligned. Either:

```tsx
// Include the count in aria-label too
aria-label={`Show ${status.edits.length} edits`}
```

or drop the redundant static `aria-label` and rely on the existing dynamic `title` (though `title` alone is not reliably announced on all platforms/assistive tech combos).

---

lib/components/posts/post.test.tsx
```diff
@@ -363,4 +363,86 @@ describe('Post', () => {
...
+      'grid',
+      'w-full',
+      'grid-cols-3',
+      'justify-items-center'
+    )
+    expect(statusActions).toHaveClass(
+      'grid',
+      'w-full',
+      'grid-cols-4',
+      'justify-items-center'
```
llun: These assertions check Tailwind class names directly, which couples the test to CSS implementation details. If the layout approach changes (e.g., switching from CSS grid to flex, or renaming a utility), these tests break even when the user-facing behaviour is identical.

The tests already verify the accessible structure (group roles + aria-labels) and all 7 button aria-labels below — that is the meaningful behaviour. The class assertions add friction without adding confidence.

Consider removing the `toHaveClass` blocks and relying solely on the accessibility and button-presence assertions that follow them.

---

llun (commented):
## Review: fix: prevent mobile post actions overflow

The overall approach — splitting post actions into a social row and a status row, using CSS grid on mobile and flex on desktop — is sound and the browser verification is thorough. Accessibility improvements (aria-labels on every button, role="group" with aria-label) are a welcome addition.

There are four issues to address before merging:

### Must fix

1. **Fragile column count** (`actions.tsx` lines 47–56): `statusActionCount` manually mirrors the rendered item count and is only correct because `EditHistoryButton` happens to return `null` internally. This silent coupling will silently break if the component or the group changes. See inline comment for alternatives.

2. **Unconditional `EditHistoryButton` render** (`actions.tsx` lines 79–83): `EditHistoryButton` is placed in JSX unconditionally (relying on its internal null guard), while every other conditional item in the same group uses explicit JSX guards. This inconsistency is confusing and reinforces the fragile count above. Use `{hasEditHistory && <EditHistoryButton ... />}` to match the pattern used for the other buttons.

3. **Static `aria-label` on `EditHistoryButton`** (`edit-history-button.tsx` line 30): The `title` includes the edit count but `aria-label` (which takes precedence for screen readers) is a static string. Screen reader users lose the count information. The two should be aligned.

### Nice to fix

4. **CSS class assertions in tests** (`post.test.tsx` lines 392–403): Asserting Tailwind class names couples tests to implementation details. The accessibility-role and button-label assertions that follow are the meaningful checks; the class assertions can be removed.

---

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +41,44 @@ export const Actions: FC<Props> = ({
...
+        <div
+          role="group"
+          aria-label="Post status actions"
+          className="flex w-full items-center justify-end gap-2 sm:w-auto sm:justify-start sm:gap-6"
+        >
+          <EditHistoryButton
+            status={actualStatus}
+            host={host}
+            onShowEdits={onShowEdits}
+          />
```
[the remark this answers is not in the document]
llun: Fixed in `477a168f`: `Actions` now renders `EditHistoryButton` only when `hasEditHistory` is true, with coverage for owner posts that have status actions but no edit history.

lib/components/posts/actions/edit-history-button.tsx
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the edit-history accessible name now includes the singular/plural edit count, e.g. `Show edit history, 1 edit`.

lib/components/posts/actions/like-button.tsx
```diff
@@ -17,6 +17,7 @@ export const LikeButton: FC<LikeButtonProps> = ({ currentActor, status }) => {
   return (
     <button
       title={isActorLiked ? 'Unlike' : 'Like'}
+      aria-label={isActorLiked ? 'Unlike' : 'Like'}
```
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the like button accessible name now includes the visible like count when positive, e.g. `Like, 3 likes` or `Unlike, 1 like`.

lib/components/posts/actions/reply-button.tsx
```diff
@@ -14,6 +14,7 @@ export const ReplyButton: FC<Props> = ({ status, onReply }) => {
     <button
       className="flex items-center gap-1.5 rounded-full px-2 py-1 text-sm transition-colors hover:bg-muted hover:text-blue-500"
       title="Reply"
+      aria-label="Reply to post"
```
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the reply button accessible name now includes the visible reply count when positive, e.g. `Reply to post, 2 replies`.

lib/components/posts/actions/like-button.tsx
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the same like-count accessible-name fix preserves the visible like count for assistive technology.

lib/components/posts/actions/reply-button.tsx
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the same reply-count accessible-name fix preserves the visible reply count for assistive technology.

lib/components/posts/actions/edit-history-button.tsx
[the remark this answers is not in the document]
llun: Fixed in `f4aec73d`: the same edit-history accessible-name fix preserves the edit count for assistive technology.

lib/components/posts/actions/repost-button.tsx
```diff
@@ -37,6 +37,7 @@ export const RepostButton: FC<RepostButtonProps> = ({
     <button
       disabled={isLoading}
       title="Repost"
+      aria-label="Repost"
```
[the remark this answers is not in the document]
llun: Fixed in `b6e2552c`: the repost button title and accessible name now switch to `Undo repost` when `repostedStatusId` is present, with test coverage.

lib/components/posts/actions/actions.tsx
llun: Fixed in f0ae8f2a. The status action group now builds the rendered actions first and derives the grid column class from statusActions.length, so the column count cannot drift from the rendered controls.

lib/components/posts/actions/actions.tsx
llun: Already fixed in 477a168f. EditHistoryButton is now only rendered when the status has edit history, so the empty secondary action case does not leave a hidden/null action in the grid.

lib/components/posts/actions/edit-history-button.tsx
llun: Already fixed in f4aec73d. The edit history button label now includes the singular/plural edit count, for example `Show edit history, 1 edit`.

lib/components/posts/post.test.tsx
llun: No code change here. These class assertions are intentional because the requested regression is the mobile Tailwind layout (`grid`, `w-full`, fixed column count); accessibility-only assertions would not catch action-row overflow in JSDOM. Browser verification covers the rendered behavior.

llun:
@codex review

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +42,71 @@ export const Actions: FC<Props> = ({
...
-        )}
+  const hasStatusActions = statusActions.length > 0
+  const statusActionGridColumns =
+    statusActions.length === 4
+      ? 'grid-cols-4'
+      : statusActions.length === 3
+        ? 'grid-cols-3'
+        : statusActions.length === 2
+          ? 'grid-cols-2'
+          : 'grid-cols-1'
```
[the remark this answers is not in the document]
llun: Fixed in 86151fe5. Replaced the nested ternary with a grid-column lookup map while preserving the same classes and behavior.

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +42,71 @@ export const Actions: FC<Props> = ({
...
+        <RepostButton currentActor={currentActor} status={actualStatus} />
+        <LikeButton currentActor={currentActor} status={actualStatus} />
       </div>
+
+      {hasStatusActions && (
+        <div
+          role="group"
+          aria-label="Post status actions"
+          className={cn(
+            'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6',
```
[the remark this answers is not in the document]
llun: No code change here. The latest requested behavior is an equal-spaced mobile grid where buttons align vertically and horizontally per row; changing the status row to `w-fit ml-auto` would make it compact/right-aligned and conflict with that layout.

llun:
@codex review

lib/components/posts/actions/like-button.tsx
```diff
@@ -13,10 +13,19 @@ interface LikeButtonProps {
...
+          totalLikes === 1 ? 'like' : 'likes'
+        }`
+      : isActorLiked
+        ? 'Unlike'
+        : 'Like'
 
   return (
     <button
       title={isActorLiked ? 'Unlike' : 'Like'}
+      aria-label={likeLabel}
```
llun: `title` and `aria-label` are now inconsistent: the tooltip a sighted user sees on hover (`"Like"` / `"Unlike"`) carries no count, while the screen reader label does (`"Like, 3 likes"`).

Since `likeLabel` already handles all three states correctly (no likes → `"Like"`, liked → `"Unlike"`, count > 0 → `"Like, 3 likes"`), just reuse it for `title` as well:

```tsx
title={likeLabel}
aria-label={likeLabel}
```

---

lib/components/posts/actions/reply-button.tsx
```diff
@@ -10,10 +10,18 @@ interface Props {
...
+      ? `Reply to post, ${status.replies.length} ${
+          status.replies.length === 1 ? 'reply' : 'replies'
+        }`
+      : 'Reply to post'
+
   return (
     <button
       className="flex items-center gap-1.5 rounded-full px-2 py-1 text-sm transition-colors hover:bg-muted hover:text-blue-500"
       title="Reply"
+      aria-label={replyLabel}
```
llun: `title="Reply"` is now out of sync with `aria-label={replyLabel}` on two levels: it omits "to post" and it omits the reply count.

Reuse `replyLabel` for `title` so tooltip and screen-reader text match:

```tsx
title={replyLabel}
aria-label={replyLabel}
```

---

lib/components/posts/post.test.tsx
```diff
@@ -363,4 +363,164 @@ describe('Post', () => {
...
+    )
```
llun: These `toHaveClass` assertions were flagged in the previous review round and are still present here (and repeated again in the `'does not render edit history action when status has no edits'` test with `toHaveClass('grid-cols-3')`).

Testing Tailwind class names ties the tests to CSS implementation details — a switch from grid to flex, or a class rename, breaks the tests without any behaviour change. The seven button-label assertions that follow this block already verify all meaningful behaviour. Please remove the class assertions from both tests.

---

lib/components/posts/post.test.tsx
```diff
@@ -363,4 +363,164 @@ describe('Post', () => {
...
+          ...status,
+          replies: [
+            { ...status, id: 'https://activities.local/replies/1' },
+            { ...status, id: 'https://activities.local/replies/2' }
+          ],
+          totalLikes: 3
+        }}
+        onShowAttachment={jest.fn()}
+      />
+    )
```
llun, line 632: Both `reply-button` and `like-button` have singular/plural branching (`'reply'` vs `'replies'`, `'like'` vs `'likes'`) but this test only exercises the plural path (2 replies, 3 likes).

Consider adding a companion test (or a second `expect` block here) with `totalLikes: 1` and a single-item replies array to cover the `'like'` / `'reply'` singular labels, e.g.:

```tsx
expect(
  screen.getByRole('button', { name: 'Reply to post, 1 reply' })
).toBeInTheDocument()
expect(
  screen.getByRole('button', { name: 'Like, 1 like' })
).toBeInTheDocument()
```

---

llun (commented):
## Round 2 review

Good progress. The three structural issues from round 1 (fragile column count, unconditional `EditHistoryButton`, static `aria-label` on edit history) are all fixed cleanly. The `ReactNode[]` accumulator approach is clear and self-consistent.

The new label improvements in `reply-button`, `like-button`, and `repost-button` are a genuine accessibility win — the plural/singular handling is a nice touch.

Three issues remain:

### Must fix

1. **`title`/`aria-label` mismatch on `LikeButton`** (`like-button.tsx` lines 27–28): `title` still shows just `"Like"` / `"Unlike"` with no count, while `aria-label` now includes the count. The `likeLabel` variable already handles all states correctly — pass it to `title` too.

2. **`title`/`aria-label` mismatch on `ReplyButton`** (`reply-button.tsx` lines 23–24): `title="Reply"` is static and missing both the "to post" suffix and the reply count that `aria-label` carries. Same fix: `title={replyLabel}`.

### Nice to fix

3. **CSS class assertions still present** (`post.test.tsx` lines 393–404 and the `grid-cols-3` assertion in `'does not render edit history action'`): these were flagged in round 1 and remain unchanged. They couple the tests to Tailwind class names and add fragility without adding confidence.

4. **Singular label paths untested** (`post.test.tsx`, `'keeps visible social action counts'` test): the test exercises 2 replies and 3 likes but not the `1 reply` / `1 like` singular branches introduced in this PR.

---

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +49,65 @@ export const Actions: FC<Props> = ({
...
+  const hasStatusActions = statusActions.length > 0
+  const statusActionGridColumns =
+    statusActionGridColumnsByCount[statusActions.length] ?? 'grid-cols-1'
+
+  return (
+    <div className="mt-3 flex flex-col gap-2 text-muted-foreground sm:flex-row sm:items-center sm:gap-6">
+      <div
+        role="group"
+        aria-label="Post social actions"
+        className="grid w-full grid-cols-3 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6"
```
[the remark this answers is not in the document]
llun: No code change here. `Actions` returns `null` before rendering when `currentActor` is missing, and `RepostButton` is only used from `Actions`, so when the social action group renders it has reply, repost, and like together. Keeping `grid-cols-3` also matches the requested equal-spaced mobile grid row.

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +49,65 @@ export const Actions: FC<Props> = ({
...
-      <RepostButton currentActor={currentActor} status={actualStatus} />
-      <LikeButton currentActor={currentActor} status={actualStatus} />
+  if (hasEditHistory) {
+    statusActions.push(
+      <EditHistoryButton
+        key="edit-history"
+        status={actualStatus}
+        host={host}
+        onShowEdits={onShowEdits}
+      />
```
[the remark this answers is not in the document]
llun: Fixed in 064ab92f. The edit-history popover now uses fixed mobile positioning with viewport side insets (`max-md:fixed max-md:inset-x-4 max-md:w-auto`), so it no longer depends on the trigger button column and avoids the previous off-screen horizontal offset.

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +49,65 @@ export const Actions: FC<Props> = ({
...
+        <LikeButton currentActor={currentActor} status={actualStatus} />
       </div>
+
+      {hasStatusActions && (
+        <div
+          role="group"
+          aria-label="Post status actions"
+          className={cn(
+            'grid w-full items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6',
+            statusActionGridColumns
```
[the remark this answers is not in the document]
llun: No code change to the layout. The latest requested UI is an equal-spaced centered mobile grid, not a compact right-aligned row. I updated the PR body to describe the centered grid layout so the description matches the implementation.

lib/components/posts/actions/like-button.tsx
llun: Fixed in 8f25d506. `LikeButton` now uses the computed `likeLabel` for both `title` and `aria-label`, so hover tooltip text and the accessible label stay in sync, including counts.

lib/components/posts/actions/reply-button.tsx
llun: Fixed in 8f25d506. `ReplyButton` now uses the computed `replyLabel` for both `title` and `aria-label`, so the tooltip includes the same wording and reply count as the accessible label.

lib/components/posts/post.test.tsx
llun: Fixed in 8f25d506 and tightened in 064ab92f. The broad Tailwind class assertions were removed, and the replacement now uses group-scoped assertions so reply/repost/like must be in the social group while edit-history/visibility/edit/delete must be in the status group.

lib/components/posts/post.test.tsx
llun: Fixed in 8f25d506. Added singular count coverage for `Reply to post, 1 reply` and `Like, 1 like`, including matching `title` assertions.

llun:
@codex review

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -26,11 +30,12 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
...
       }}
-      title={`${status.edits.length} edits`}
+      title={editCountLabel}
+      aria-label={`Show edit history, ${editCountLabel}`}
     >
       <History className="h-4 w-4" />
       {showHistory && (
-        <div className="absolute bottom-full left-0 mb-2 w-[25rem] max-md:left-[-8rem] max-md:w-[18rem] z-20">
+        <div className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto">
           <ul className="divide-y divide-border rounded-lg border bg-background shadow-lg">
```
llun: The `e.stopPropagation()` on the button's own `onClick` only stops the event from bubbling *above* the button (e.g. into the post card). It does not block click events that originate *inside* the panel from reaching the button.

When the panel is open, any tap or click inside it follows the bubble path `<li>` → `<ul>` → `<div>` → `<button>`, which fires `setShowHistory((value) => !value)` and immediately closes the panel. The user can't scroll the list, tap a timestamp, or even release a tap on the panel without it vanishing.

This bug existed before this PR but was less noticeable when the panel was a small shifted overlay. The new `max-md:fixed` positioning makes the panel large and interactive, so the bug is now reliably reproducible on mobile.

Fix: stop the click on the panel div from reaching the button:

```tsx
<div
  className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto"
  onClick={(e) => e.stopPropagation()}
>
```

Changing `<div>` to `<span className="block ...">` would also address the HTML-validity concern (block-level element inside `<button>`) noted in the existing Gemini thread, and can be combined with the stopPropagation fix in the same change.

---

llun (commented):
## Round 3 review

All issues from rounds 1 and 2 are resolved. The group-scoped `within()` test assertions are a clean replacement for the CSS class checks. The singular label tests, the `title`/`aria-label` alignment on Like and Reply, and the reactive repost label are all solid. The new `max-md:fixed` positioning for the edit history panel on mobile is a good solution to the off-screen offset problem Copilot raised.

### One remaining must-fix

**Edit history panel click-through bug** (`edit-history-button.tsx` line 39): any tap or click anywhere inside the dropdown panel bubbles up to the `<button>` and immediately re-fires `setShowHistory((value) => !value)`, closing the panel. This was a latent pre-existing bug that the new `fixed` full-width mobile positioning makes reliably reproducible — users now have a large, prominent panel they can't actually interact with without it snapping shut. Fix: add `onClick={(e) => e.stopPropagation()}` on the panel `<div>`. See inline comment for full detail and the option to resolve the Gemini HTML-validity thread at the same time.

---

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -26,11 +30,12 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
...
         setShowHistory((value) => !value)
       }}
-      title={`${status.edits.length} edits`}
+      title={editCountLabel}
+      aria-label={`Show edit history, ${editCountLabel}`}
     >
       <History className="h-4 w-4" />
       {showHistory && (
-        <div className="absolute bottom-full left-0 mb-2 w-[25rem] max-md:left-[-8rem] max-md:w-[18rem] z-20">
+        <div className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto">
```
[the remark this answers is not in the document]
llun: Fixed in 896dbcee. The popover is now a sibling of the trigger button inside a neutral wrapper, so the dropdown markup is no longer nested inside the button; the popover also stops click propagation.

lib/components/posts/actions/edit-history-button.tsx
llun: Fixed in 896dbcee. The panel now has its own `onClick={(e) => e.stopPropagation()}` and a regression test verifies clicking panel content keeps the edit history open without re-triggering `onShowEdits`.

llun:
Final mobile verification screenshot (390x844):

Final mobile post actions fix

Verification notes:
- Owner post exposes reply, repost, like, edit history, visibility, edit, and delete actions.
- Status action row renders four equal-spaced controls; delete is fully visible at x=296.875..328.875 inside the 390px viewport.
- DOM measurement: document scrollWidth 375 <= viewport width 390; no horizontal overflow.
- Edit-history panel opens and remains visible on mobile.

llun:
@codex review

lib/components/posts/actions/actions.tsx
```diff
@@ -19,6 +20,13 @@ interface Props extends PostProps {
   onShowEdits?: (status: Status) => void
 }
 
+const statusActionGridColumnsByCount: Record<number, string> = {
+  1: 'grid-cols-1',
+  2: 'grid-cols-2',
+  3: 'grid-cols-3',
+  4: 'grid-cols-4'
+}
```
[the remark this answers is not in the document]
llun: Fixed in `541a7141`. The hardcoded `statusActionGridColumnsByCount` mapping was removed. The layout now uses a shared four-column action row class: the primary row renders reply/repost/like/edit-history, and the secondary row renders visibility/edit/delete in the first three columns so it stays left-aligned without a fragile count-to-class fallback.

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -18,19 +18,29 @@ export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
...
+        }}
+        title={editCountLabel}
+        aria-label={`Show edit history, ${editCountLabel}`}
+      >
+        <History className="h-4 w-4" />
+      </button>
       {showHistory && (
-        <div className="absolute bottom-full left-0 mb-2 w-[25rem] max-md:left-[-8rem] max-md:w-[18rem] z-20">
+        <div
+          className="absolute bottom-full left-0 mb-2 w-[25rem] z-20 max-md:fixed max-md:inset-x-4 max-md:bottom-20 max-md:w-auto"
```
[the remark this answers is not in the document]
llun: Fixed in `541a7141`. I kept the mobile panel fixed because it avoids narrow-viewport clipping, but made it behave as a dismissible mobile dialog: the trigger now exposes `aria-expanded`/`aria-controls`, the panel has `role="dialog"`, bounded scrolling, and an explicit `Close edit history` button. Browser verification at 390x844 confirms the dialog opens and closes without needing to return to the trigger.

llun:
Final mobile verification screenshot (390x844):

Final mobile post actions 4 plus 3 fix

Verification notes:
- Primary row shows 4 actions: reply, repost, like, edit history.
- Secondary row shows 3 left-aligned actions: visibility, edit, delete.
- DOM measurement: viewport width 390, document scroll width 375; no horizontal overflow.
- Both action rows measured 257px wide inside 257px post content.
- Edit-history dialog opens and closes with its explicit close button.

llun:
@codex review

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -15,23 +15,55 @@ interface Props {
...
+        aria-controls={showHistory ? editHistoryId : undefined}
+      >
+        <History className="h-4 w-4" />
+      </button>
       {showHistory && (
-        <div className="absolute bottom-full left-0 mb-2 w-[25rem] max-md:left-[-8rem] max-md:w-[18rem] z-20">
-          <ul className="divide-y divide-border rounded-lg border bg-background shadow-lg">
+        <div
+          id={editHistoryId}
+          role="dialog"
```
llun: `role="dialog"` is mismatched to the implementation here. The ARIA dialog pattern requires:
1. Focus moved into the dialog when it opens (e.g. onto the close button)
2. `Escape` key closes it
3. `aria-modal="true"` if focus is trapped

None of these are implemented. The panel is actually using the **disclosure pattern** — the trigger already has `aria-expanded` and `aria-controls`, which is all assistive technology needs to understand the expand/collapse relationship.

Fix: remove `role="dialog"` (and optionally `aria-label="Edit history"` on the container — the trigger's `aria-controls` link already conveys the relationship). The panel `<div>` needs only its `id`. If you want to keep a descriptive role, `role="region"` with `aria-label` is appropriate for a non-interactive landmark that doesn't manage focus.

The test would change from `getByRole('dialog', ...)` to `document.getElementById(...)` or a `data-testid`, or `getByRole('region', ...)` depending on which direction you take.

---

lib/components/posts/actions/actions.tsx
```diff
@@ -19,6 +19,9 @@ interface Props extends PostProps {
   onShowEdits?: (status: Status) => void
 }
 
+const actionRowClassName =
+  'grid w-full grid-cols-4 items-center justify-items-center gap-2 sm:flex sm:w-auto sm:justify-start sm:gap-6'
```
llun: `grid-cols-4` is hardcoded for the primary row, but when a post has no edit history the row renders only 3 items (Reply, Repost, Like). On mobile, those 3 buttons occupy columns 1–3 with an empty 4th column — the row looks right-padded rather than evenly distributed.

The browser verification in the PR was done with a status that has edits (4 controls confirmed). The 3-item case is not explicitly verified visually.

Options:
- Dynamically set `grid-cols-3` / `grid-cols-4` based on `primaryActions.length` (the same pattern the secondary row used in a previous revision)
- Or use `grid-flow-col auto-cols-fr` so the CSS handles it automatically without any count logic

---

llun (commented):
## Round 4 review

The click-through bug is properly fixed — the panel is now a sibling `<div>` with its own `onClick` stopper, `aria-expanded`/`aria-controls` on the trigger, and a dedicated close button. The restructuring of `primaryActions` / `secondaryActions` arrays and the shared `actionRowClassName` constant are clean. All issues from rounds 1–3 are resolved.

Two issues remain.

### Must fix

1. **`role="dialog"` without focus management** (`edit-history-button.tsx` line 45): The ARIA dialog pattern requires focus to be moved inside the element on open and `Escape` to close it. Neither is implemented. The component is actually using the correct *disclosure* pattern (`aria-expanded` + `aria-controls`), so the `role="dialog"` on the panel container is misleading to assistive technologies. Remove `role="dialog"` (use `role="region"` or just a plain `<div>` with the `id`).

### Worth verifying before merge

2. **`grid-cols-4` with 3-item primary row** (`actions.tsx` line 23): When a post has no edit history the primary row renders only 3 items in a `grid-cols-4` grid, leaving an empty 4th column on mobile. The browser verification was done with 4 primary controls. Please verify the 3-item case visually (a post that has never been edited) and either confirm the empty column is acceptable or switch to dynamic column sizing / `auto-cols-fr`.

---

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -15,23 +15,55 @@ interface Props {
...
+            <span className="text-sm font-medium text-foreground">
+              Edit history
+            </span>
+            <button
+              type="button"
+              className="rounded-full p-1 transition-colors hover:bg-muted"
+              aria-label="Close edit history"
+              onClick={(e) => {
+                e.stopPropagation()
+                setShowHistory(false)
```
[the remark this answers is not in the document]
llun: Fixed in `d8584b84`. The close handler now restores focus to the edit-history trigger after closing the panel, and the test asserts focus returns to `Show edit history, 1 edit`.

lib/components/posts/actions/edit-history-button.tsx
llun: Fixed in `d8584b84`. The panel now uses `role="region"` with `aria-label="Edit history"` instead of `role="dialog"`, keeping the disclosure semantics from `aria-expanded`/`aria-controls` without claiming dialog focus behavior.

lib/components/posts/actions/actions.tsx
llun: Fixed in `d8584b84`. The primary action row now uses `grid-cols-4` only when edit history is present and `grid-cols-3` when the post has no edit history; the owner secondary row remains a 4-column grid so the 3 controls stay left-aligned under the 4-action layout. Added test coverage and verified both cases in Browser at 390x844.

llun:
Final mobile verification screenshot after Round 4 fixes (390x844):

Final mobile post actions round 4 fix

Verification notes:
- Edited owner post uses a 4-action primary row and 3 left-aligned secondary actions.
- Owner post without edit history uses a 3-action primary row, removing the empty fourth primary slot.
- DOM measurement: viewport width 390, document scroll width 375; no horizontal overflow.
- Both edited and unedited action rows stayed within 257px post content width.
- Edit-history content exposes `role="region"`, no dialog role, and closing restores focus to the trigger.

llun:
@codex review

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -15,23 +15,61 @@ interface Props {
...
+              onClick={(e) => {
+                e.stopPropagation()
+                closeHistory()
+              }}
+            >
+              <X className="h-4 w-4" />
+            </button>
+          </div>
+          <ul className="max-h-80 divide-y divide-border overflow-auto max-md:max-h-[calc(100vh-10rem)]">
             {status.edits.reverse().map((edit, index) => {
```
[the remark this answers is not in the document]
llun: Fixed in `a0d900dd`. The edit-history list now renders from a copied reversed array (`[...status.edits].reverse()`), so the prop array is not mutated. I also added a regression test that verifies newest-first rendering while preserving the original `edits` order.

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +45,70 @@ export const Actions: FC<Props> = ({
...
         )}
+      >
+        {primaryActions}
       </div>
+
+      {hasSecondaryActions && (
+        <div
+          role="group"
+          aria-label="Post secondary actions"
+          className={cn(actionRowClassName, 'grid-cols-4')}
```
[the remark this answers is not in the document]
llun: Fixed in `a0d900dd`. The secondary row now uses the same mobile column class as the primary row (`grid-cols-4` with edit history, `grid-cols-3` without it), so the rows stay vertically aligned in both owner layouts. The no-edit-history test now asserts the secondary row uses `grid-cols-3` and not `grid-cols-4`.

llun:
Final mobile verification screenshot after Round 5 fixes (390x844):

Final mobile post actions round 5 fix

Verification notes:
- Edited owner post uses `grid-cols-4`: reply, repost, like, edit history.
- Edited owner secondary row also uses `grid-cols-4`, with visibility, edit, delete left-aligned in the first 3 cells.
- Owner post without edit history uses matching `grid-cols-3` primary and secondary rows, so no empty primary slot and vertical columns align.
- DOM measurement: viewport width 390, document scroll width 375, client width 375; no horizontal overflow.
- Visible action row width was 257px inside the mobile post card, and delete is visible/clickable.

llun:
@codex review

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +45,68 @@ export const Actions: FC<Props> = ({
...
-        />
-        {isOwner && <VisibilityButton status={actualStatus} />}
-        {canEdit && (
-          <>
-            <EditButton status={actualStatus} onEdit={onEdit} />
-            <DeleteButton status={actualStatus} onPostDeleted={onPostDeleted} />
-          </>
-        )}
+  const hasSecondaryActions = secondaryActions.length > 0
+  const actionColumnClassName = hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'
```
llun, line 92: `actionColumnClassName` is derived from `hasEditHistory` and applied identically to both rows. The intent — keeping both rows on the same column grid so their cells align vertically — is not obvious from the name or value alone.

This also creates untested edge cases for the secondary row. When `isOwner=true, canEdit=false` and there's no edit history, there's 1 item in a `grid-cols-3` grid (two empty columns). When `isOwner=false, canEdit=true` without edit history, there are 2 items in `grid-cols-3` (one empty column). These combinations are uncommon in practice but could look unbalanced.

A short comment explaining the alignment intent would help future maintainers understand why both rows share the same column count.

---

llun (commented):
## Round 5 review

All issues from round 4 are resolved cleanly:

- `role="region"` replaces `role="dialog"` ✅
- `[...status.edits].reverse()` fixes the in-place mutation bug ✅
- `triggerRef` + `closeHistory()` returns focus to the trigger when the panel closes ✅
- `actionColumnClassName = hasEditHistory ? 'grid-cols-4' : 'grid-cols-3'` gives both rows a dynamic, consistent column count ✅
- New test for order and non-mutation, and focus-return assertion added to the interaction test ✅

The PR is in good shape. One minor observation inline (no code change required, just worth a comment or a visual check on edge cases). Everything else looks correct.

### Nice to have (not blocking)

**Escape key to close the panel**: keyboard users who Tab into the panel and prefer Escape to dismiss it currently have to reach the close button. The disclosure pattern doesn't require it, but it's a widely expected convention for any overlay. Easy to add with a `onKeyDown` handler on the panel `<div>`.

---

lib/components/posts/actions/actions.tsx
```diff
@@ -41,27 +45,68 @@ export const Actions: FC<Props> = ({
...
-      <RepostButton currentActor={currentActor} status={actualStatus} />
-      <LikeButton currentActor={currentActor} status={actualStatus} />
+  if (hasEditHistory) {
+    primaryActions.push(
+      <EditHistoryButton
+        key="edit-history"
+        status={actualStatus}
+        host={host}
+        onShowEdits={onShowEdits}
+      />
```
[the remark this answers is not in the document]
llun: Fixed in `7bf115fd`. `Actions` now passes the existing `currentTime` prop into `EditHistoryButton`, matching the main post timestamp flow.

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -15,24 +15,63 @@ interface Props {
 
 export const EditHistoryButton: FC<Props> = ({ host, status, onShowEdits }) => {
```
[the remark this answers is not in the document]
llun: Fixed in `7bf115fd`. `EditHistoryButton` now requires `currentTime` and uses it in `formatDistance(edit.createdAt, currentTime)` instead of `Date.now()`. Added a regression test that sets system time 7 days away and verifies the displayed edit-history timestamp still comes from the passed `currentTime`.

llun:
@codex review

lib/components/posts/actions/actions.tsx
llun: Fixed in `b45f6736`. Added a short code comment above `actionColumnClassName` documenting that both mobile rows intentionally share the same grid so secondary controls align under primary controls. I did not add extra tests for the rare one- or two-secondary-action cases because the existing tests already lock the intended shared-column behavior for the normal 4-column and 3-column layouts.

llun:
@codex review

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -9,37 +9,82 @@ import { convertMarkdownText } from '@/lib/utils/text/convertMarkdownText'
...
+              }}
+            >
+              <X className="h-4 w-4" />
+            </button>
+          </div>
+          <ul className="max-h-80 divide-y divide-border overflow-auto max-md:max-h-[calc(100vh-10rem)]">
+            {edits.map((edit, index) => {
               return (
                 <li
                   key={edit.createdAt + index}
```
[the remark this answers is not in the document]
llun: Fixed in `943545bf`. The edit-history item key now uses a delimited string (`${edit.createdAt}-${index}`), avoiding numeric addition collisions since `createdAt` is a number. No test added because React key collision behavior would be brittle to assert and this is a one-line key construction fix.

llun:
@codex review

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -9,37 +9,82 @@ import { convertMarkdownText } from '@/lib/utils/text/convertMarkdownText'
...
 
+  const editCountLabel = `${status.edits.length} ${
+    status.edits.length === 1 ? 'edit' : 'edits'
+  }`
+  const edits = [...status.edits].reverse()
+  const closeHistory = () => {
+    setShowHistory(false)
+    triggerRef.current?.focus()
+  }
+
```
llun, line 37: The PR description's browser verification still reads "closes without returning to the trigger", but the code now explicitly calls `triggerRef.current?.focus()` in `closeHistory()`, and the test asserts `expect(editHistoryButton).toHaveFocus()`. The note should be updated to say focus is returned to the trigger button on close.

---

llun (commented):
## Round 6 review

All round 5 observations are addressed:

- `actionColumnClassName` now has a comment explaining the vertical-alignment intent ✅
- `currentTime` prop threaded through `Actions` → `EditHistoryButton`, replacing the bare `Date.now()` call in `formatDistance` — correct per the project's hydration convention ✅
- List item key changed from numeric addition (`createdAt + index`) to an explicit template literal — cleaner and unambiguous ✅
- New test `'uses currentTime for edit history relative timestamps'` verifies the prop is actually used, not the system clock ✅

**One doc nit (inline):** the browser verification bullet in the PR description says "closes *without* returning to the trigger", but `closeHistory()` now explicitly returns focus and the test asserts it. The note should say the opposite before merge.

**Still nice to have (not blocking):** Escape key to dismiss the panel. The disclosure pattern doesn't require it, but it's a widely expected convention for overlays. The close button is sufficient as-is.

The code is correct and the test suite is thorough. Ready to merge once the PR description is updated.

---

lib/components/posts/actions/edit-history-button.tsx
```diff
@@ -9,37 +9,82 @@ import { convertMarkdownText } from '@/lib/utils/text/convertMarkdownText'
...
+            >
+              <X className="h-4 w-4" />
+            </button>
+          </div>
+          <ul className="max-h-80 divide-y divide-border overflow-auto max-md:max-h-[calc(100vh-10rem)]">
+            {edits.map((edit, index) => {
               return (
                 <li
-                  key={edit.createdAt + index}
+                  key={`${edit.createdAt}-${index}`}
```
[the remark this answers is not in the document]
llun: Not changing this to `key={edit.createdAt}` because it would be less correct for this codebase. `status_history` rows are written with `createdAt` equal to the original status creation time, so multiple edits for the same post can share the same `createdAt` and collide. `Edited.createdAt` is already a serializable number, and the current delimited key keeps rows unique even when timestamps repeat. No code change needed for this thread.

lib/components/posts/actions/edit-history-button.tsx
llun: Fixed in the PR description as metadata-only, so there is no commit for this item. The browser verification note now says the edit-history panel returns focus to the trigger after close, matching `closeHistory()` and the existing focus assertion. I also updated the displayed test counts to the latest verified results.

PR #706, opened 2026-05-16
minor: add Mastodon-compatible bookmarks API

llun:
## Summary
- Add private Mastodon-compatible bookmark storage and database operations.
- Implement GET /api/v1/bookmarks plus POST /api/v1/statuses/:id/bookmark and /unbookmark.
- Hydrate Mastodon status JSON with bookmarked state, including Announce/reblog normalization to the original status.
- Add granular read:bookmarks and write:bookmarks OAuth scopes while keeping broad read/write compatibility.

## Compatibility Notes
- Bookmarks are local API state only: no ActivityPub federation, notifications, bookmark counts, first-party UI, or folder support.
- Bookmarking an Announce stores the original status, matching Mastodon behavior.
- Unbookmark remains idempotent and allows cleanup when the bookmarked status is no longer readable.

## Migrations
- Adds the bookmarks table with actor/status uniqueness and listing indexes.

## Verification
- yarn run prettier --write .
- yarn lint (0 errors; existing unrelated warnings remain)
- yarn build
- yarn test
- Dev + Browser: started yarn dev with local SQLite runtime config, opened http://127.0.0.1:3010 in Browser, confirmed the app rendered the sign-in page without runtime console errors. Unauthenticated GET /api/v1/bookmarks returned 401 as expected. No authenticated browser session was available, so authenticated bookmark smoke coverage is via route tests.


llun:
@codex review

app/api/v1/bookmarks/route.ts
```diff
@@ -0,0 +1,87 @@
...
+        statuses,
+        currentActor
+      })
+      const mastodonStatuses = (
+        await Promise.all(
+          readableStatuses.map((status) =>
+            getMastodonStatus(database, status, currentActor.id)
+          )
+        )
+      ).filter((status): status is Mastodon.Status => status !== null)
```
llun, line 62: **Ordering bug: statuses are not returned in bookmark order.**

`getStatusesByIds` uses `WHERE id IN (...)` internally, which returns results in an unspecified order — not the order of the input `statusIds` array. After `filterReadableStatuses` and `Promise.all`, the final `mastodonStatuses` array may be in a different order than `bookmarks`, breaking the expected newest-first pagination order.

Fix: after fetching, re-sort statuses to match bookmark order before passing them downstream:

```ts
const statusMap = new Map(statuses.map((s) => [s.id, s]))
const orderedStatuses = bookmarks
  .map((b) => statusMap.get(b.statusId))
  .filter((s): s is Status => s !== undefined)
const readableStatuses = await filterReadableStatuses({
  database,
  statuses: orderedStatuses,
  currentActor
})
```

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+  const status = await database('statuses').where('id', statusId).first<{
+    id: string
+    type: string
+    content: unknown
+  }>('id', 'type', 'content')
+  if (!status) return null
+
+  if (status.type !== StatusType.enum.Announce) return status.id
+
+  return getOriginalStatusIdFromAnnounceContent(status.content)
```
llun: **`resolveBookmarkStatusId` parses the `content` column instead of using the `originalStatusId` column.**

`CreateAnnounceParams` includes `originalStatusId`, which is stored as a dedicated column in the `statuses` table. This code re-derives it by parsing the `content` JSON, which is fragile and duplicates logic already handled at insert time.

Query `originalStatusId` directly instead:

```ts
const resolveBookmarkStatusId = async ({
  database,
  statusId
}: {
  database: Knex | Knex.Transaction
  statusId: string
}): Promise<string | null> => {
  const status = await database('statuses')
    .where('id', statusId)
    .first<{ id: string; type: string; originalStatusId: string | null }>(
      'id', 'type', 'originalStatusId'
    )
  if (!status) return null
  if (status.type !== StatusType.enum.Announce) return status.id
  return status.originalStatusId ?? null
}
```

This also lets you remove `parseStatusContent`, `getOriginalStatusIdFromAnnounceContent`, and the `getCompatibleJSON` import entirely.

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+  return (
+    errorCode === '23505' ||
+    errorCode === 'ER_DUP_ENTRY' ||
+    errorCode === 'SQLITE_CONSTRAINT_UNIQUE' ||
+    errorNumber === 1062 ||
+    (errorCode === 'SQLITE_CONSTRAINT' &&
+      Boolean(errorMessage?.includes('UNIQUE constraint failed'))) ||
+    Boolean(errorMessage?.includes('UNIQUE constraint failed'))
+  )
+}
```
llun: **`isUniqueConstraintError` is defined locally — check if a shared utility already exists.**

This pattern of detecting unique-constraint violations across SQLite/PostgreSQL/MySQL appears in `bookmark.ts` only, but other mixins (like `like.ts`) avoid the error entirely by doing an existence check first and not catching constraint errors at all. The bookmark mixin does both (existence check + catch), which is valid for race-condition safety, but the multi-engine constraint detection logic should live in a shared utility under `lib/database/sql/utils/` if it's going to be used more widely. If this is the only place it's needed, it's acceptable here, but worth noting.

---

lib/services/mastodon/getMastodonStatus.ts
llun, line 1: **Commenting style throughout this file violates project conventions.**

This file now contains many "what" comments — `// Get the host from the global config`, `// Derive visibility from recipients`, `// Identifiers & timestamps`, `// Count metrics`, `// Interaction flags`, etc. — plus multi-line JSDoc blocks on the helper functions (`/** Extract Mastodon mentions from status tags */`, etc.) and inline implementation notes inside functions (`// tag.name is typically like "@user@domain.com" or "@user"`, `// tag.name is typically like ":emoji:" - remove colons for shortcode`, etc.).

Per `CLAUDE.md`: _"Default to writing no comments. Only add one when the WHY is non-obvious"_ and _"Don't explain WHAT the code does, since well-named identifiers already do that."_

All of these should be removed. Only the `isStatusBookmarked` helper itself needs to stay (the logic is clear from the name and the code).

---

lib/types/database/operations.ts
llun, line 1: **File-level comment block at the top should be removed.**

The PR added `// Database operation parameters and interfaces\n// Consolidated from lib/database/types/*.ts` at the top of this file. These describe what the file is, which is already conveyed by its name and location. Per project conventions, comments should only be added when the WHY is non-obvious.

---

app/api/v1/statuses/[id]/unbookmark/route.ts
```diff
@@ -22,54 +22,81 @@ interface Params {
...
+          statusId,
+          withReplies: false,
+          currentActorId: currentActor.id
+        })
+        if (!status)
+          return apiResponse({
+            req,
+            allowedMethods: CORS_HEADERS,
+            data: ERROR_404,
+            responseStatusCode: 404
```
Comment (llun) [inline app/api/v1/statuses/[id]/unbookmark/route.ts]: **Fallback path calls `database.getStatus` without access-control filtering.**

When the status is not readable but IS bookmarked (e.g. it became private/deleted after the bookmark was created), the code falls through to `database.getStatus({ statusId, withReplies: false, currentActorId: currentActor.id })`. Unlike `getReadableStatus`, this path does not enforce visibility rules (it relies purely on `currentActorId` in the query, whose behavior for private/deleted statuses is unclear from context). If the status still exists in the DB but is now genuinely invisible to the actor (e.g. the author blocked them, or the post is followers-only from someone they no longer follow), returning its Mastodon JSON after unbookmarking is a minor information leak.

Mastodon's own API returns `{}` (or `404`) for the status payload in this scenario, not the full status object. Consider returning a 200 with only the `bookmarked: false` override on a minimal status shell, or simply returning 404 if the status is no longer accessible after the unbookmark, which is also acceptable behavior.

---

migrations/20260516000000_add_bookmarks_table.js
llun, line 1: **Migration looks good overall, but consider adding a foreign key constraint on `statusId`.**

The `bookmarks_status` index on `statusId` enables fast cascade-delete lookups, but there is no `REFERENCES statuses(id) ON DELETE CASCADE` foreign key. The bookmark test covers cascade-delete behavior (`removes bookmarks when a bookmarked status is deleted`), which implies the application manually deletes bookmarks before/after deleting a status. If that manual cleanup is always guaranteed, the FK is optional; if not, stale bookmark rows can accumulate after status deletion.

Also note: `bigIncrements('id')` produces a `BIGINT` primary key. The `Bookmark` domain type in `lib/types/domain/bookmark.ts` defines `id` as `z.string()`, and `fixBookmarkDataDate` coerces it via `String(data.id)`. This works, but it means the bookmark `id` exposed in `Link` headers is a numeric string (`"1"`, `"42"`, etc.) rather than a UUID like other entity IDs in this codebase. That is consistent with Mastodon's behavior for bookmarks, just worth being explicit about.

---

llun (commented):
## Review: minor: add Mastodon-compatible bookmarks API

The overall feature is well-scoped — the Announce normalization, idempotent create/delete, cursor pagination, and OAuth scope additions are all solid. Tests cover the happy paths and edge cases well. A few things need to be fixed before merging:

### Must fix

1. **Ordering bug in `GET /api/v1/bookmarks`** (`app/api/v1/bookmarks/route.ts` lines 39–55)
   `getStatusesByIds` does not guarantee order matches the input `statusIds`, so the response array may not be in bookmark order. See inline comment for the fix.

2. **`resolveBookmarkStatusId` parses `content` JSON instead of reading `originalStatusId`** (`lib/database/sql/bookmark.ts` lines 43–91)
   The `statuses` table already has an `originalStatusId` column populated by `createAnnounce`. The current approach re-derives the original status URL by parsing the `content` column as JSON, which is fragile and bypasses the canonical column. See inline comment for the fix (also removes `parseStatusContent`, `getOriginalStatusIdFromAnnounceContent`, and the `getCompatibleJSON` import).

### Should fix

3. **Unbookmark fallback exposes non-readable status** (`app/api/v1/statuses/[id]/unbookmark/route.ts` lines 44–68)
   When the status is no longer readable by the actor but is still bookmarked, the route falls back to `database.getStatus` without access-control filtering, potentially leaking private/deleted status data.

4. **Commenting style** (`lib/services/mastodon/getMastodonStatus.ts`, `lib/types/database/operations.ts`)
   Both files have "what" comments and JSDoc blocks that violate `CLAUDE.md`'s no-comment convention. They should be removed.

### Minor

5. **`isUniqueConstraintError` is defined locally** — if this pattern spreads to other mixins, consider moving it to `lib/database/sql/utils/`. Fine here for now.

6. **Migration: no FK on `statusId`** — see inline comment. Not blocking, just something to be aware of for data integrity.

7. **Numeric bookmark IDs** — `bigIncrements` produces integer IDs exposed as strings in `Link` headers, unlike UUID-based IDs elsewhere. Acceptable per Mastodon spec, but worth documenting.

---

app/api/v1/statuses/[id]/unbookmark/route.ts
```diff
@@ -22,54 +22,81 @@ interface Params {
...
+          })
+
+        status = await database.getStatus({
+          statusId,
+          withReplies: false,
+          currentActorId: currentActor.id
+        })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/unbookmark/route.ts]: Fixed in 1770b84f. The unreadable-but-bookmarked fallback no longer loads or serializes the status. It deletes the current actor bookmark and returns only a minimal sanitized response with bookmarked: false, with route coverage for the no-longer-readable case.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+    limit,
+    maxId,
+    minId,
+    sinceId
+  }: GetBookmarksParams) {
+    const query = database<Bookmark>('bookmarks')
+      .where('actorId', actorId)
+      .limit(limit)
+
+    const cursorId = maxId || minId || sinceId
```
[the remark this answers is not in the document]
llun: Fixed in 1770b84f. max_id and min_id/since_id are now applied as intersecting cursor bounds, with database regression coverage for max_id+min_id and max_id+since_id windows.

app/api/v1/bookmarks/route.ts
llun: Fixed in 1770b84f. The route now maps fetched statuses by id and reconstructs the list in bookmark row order before readability filtering and Mastodon serialization. Added route coverage for newest-bookmark-first ordering independent of status creation order.

lib/database/sql/bookmark.ts
llun: I did not query originalStatusId because this schema does not have a statuses.originalStatusId column; createAnnounce stores the target status id in statuses.content. I did remove the JSON parsing/import and made resolution use the actual stored string format, with recursive Announce normalization in 1770b84f.

lib/database/sql/bookmark.ts
llun: No code change here. This helper is still local because this is the only mixin that currently needs race-safe unique-constraint handling after the existence check. I would extract it once another write path needs the same multi-engine catch logic.

lib/services/mastodon/getMastodonStatus.ts
llun: Fixed in 1770b84f. Removed the added what-comments and JSDoc blocks from getMastodonStatus while keeping the helper logic intact.

lib/types/database/operations.ts
llun: Fixed in 1770b84f. Removed the file-level comment block from operations.ts.

app/api/v1/statuses/[id]/unbookmark/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/unbookmark/route.ts]: Fixed in 1770b84f. The fallback still allows bookmark cleanup, but it now returns a sanitized minimal response instead of hydrating the unreadable status payload.

migrations/20260516000000_add_bookmarks_table.js
llun: No code change here. The manual cleanup path is already covered by the status-deletion bookmark test, and I kept the migration cross-backend/simple instead of adding FK behavior in this PR. Numeric bookmark IDs are intentional for Mastodon-style private bookmark cursors and are exposed as strings in Link headers.

app/api/v1/statuses/[id]/unbookmark/route.ts
```diff
@@ -22,54 +22,81 @@ interface Params {
...
+        if (!status)
+          return apiResponse({
+            req,
+            allowedMethods: CORS_HEADERS,
+            data: ERROR_404,
+            responseStatusCode: 404
+          })
+      }
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/unbookmark/route.ts:66]: Fixed in 1770b84f. Deletion remains scoped by actorId, and the unreadable-status path now returns only the sanitized bookmarked:false response instead of content.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+    id: string
+    type: string
+    content: unknown
+  }>('id', 'type', 'content')
+  if (!status) return null
+
+  if (status.type !== StatusType.enum.Announce) return status.id
+
+  return getOriginalStatusIdFromAnnounceContent(status.content)
+}
```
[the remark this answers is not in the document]
llun: Fixed in 1770b84f. Announce bookmark normalization now recurses through Announce chains using indexed id lookups and falls back to the referenced id if the target is not locally loaded. Added nested Announce database coverage.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+    actorId,
+    statusId
+  }: IsActorBookmarkedStatusParams) {
+    const bookmarkStatusId =
+      (await resolveBookmarkStatusId({ database, statusId })) ?? statusId
+    const bookmark = await database('bookmarks')
+      .where({ actorId, statusId: bookmarkStatusId })
+      .first('id')
+    return Boolean(bookmark)
+  },
```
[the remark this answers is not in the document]
llun: Partially addressed in 1770b84f by removing the redundant status lookup when the status type is already known during status hydration. I did not add a bulk bookmark-state loader in this loop because that would require a broader status hydration contract change; the current path remains bounded per hydrated status and preserves existing loading semantics.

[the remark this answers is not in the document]
llun: Fixed in 1770b84f. isActorBookmarkedStatus now accepts an optional statusType, and status hydration passes data.type so non-Announce statuses skip the extra statuses-table lookup.

lib/database/sql/bookmark.ts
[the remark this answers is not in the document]
llun: Fixed in 1770b84f. resolveBookmarkStatusId now recursively normalizes Announce-of-Announce chains to the root original status, with regression coverage for nested Announces.

llun:
@codex review

app/api/v1/bookmarks/route.ts
```diff
@@ -0,0 +1,94 @@
...
+        statuses: orderedStatuses,
+        currentActor
+      })
+      const mastodonStatuses = (
+        await Promise.all(
+          readableStatuses.map((status) =>
+            getMastodonStatus(database, status, currentActor.id)
+          )
+        )
+      ).filter((status): status is Mastodon.Status => status !== null)
```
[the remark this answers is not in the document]
llun: No code change in this PR. The concern is directionally valid, but batching getMastodonStatus correctly would require a broader serializer/database contract change across accounts, reply/reblog counts, polls, recursive reblogs, and other list endpoints. The bookmarks route is capped at 40 items, and a route-local batching shortcut would duplicate serializer behavior and risk drift. I would handle this as a dedicated performance refactor with parity coverage.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,192 @@
...
+  const errorMessage = typeof message === 'string' ? message : undefined
+
+  return (
+    errorCode === '23505' ||
+    errorCode === 'ER_DUP_ENTRY' ||
+    errorCode === 'SQLITE_CONSTRAINT_UNIQUE' ||
+    errorNumber === 1062 ||
+    (errorCode === 'SQLITE_CONSTRAINT' &&
+      Boolean(errorMessage?.includes('UNIQUE constraint failed'))) ||
+    Boolean(errorMessage?.includes('UNIQUE constraint failed'))
```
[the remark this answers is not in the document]
llun: Fixed in f11ebfa5. Removed the redundant SQLITE_CONSTRAINT + message branch and kept the standalone UNIQUE constraint message check.

llun:
@codex review

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,190 @@
...
+  )
+  if (!originalStatusId || originalStatusId === status.id) return status.id
+
+  return (
+    (await resolveBookmarkStatusId({
+      database,
+      statusId: originalStatusId
+    })) ?? originalStatusId
+  )
+}
```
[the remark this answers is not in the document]
llun: Fixed in cc247682. resolveBookmarkStatusId now has a bounded Announce-resolution depth so malformed cyclic/deep data cannot recurse indefinitely.

app/api/v1/bookmarks/route.ts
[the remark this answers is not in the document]
llun: No code change in this PR. The bookmarks route already batches status loading with getStatusesByIds and caps responses at 40. The remaining per-status Mastodon serialization work is shared behavior across list endpoints; batching it correctly needs a dedicated serializer/database refactor with parity tests for accounts, counters, replies, polls, and recursive reblogs rather than a route-local shortcut.

app/api/v1/statuses/[id]/unbookmark/route.ts
```diff
@@ -22,54 +22,79 @@ interface Params {
...
-    // Unbookmarking not yet implemented
-    // TODO: Implement bookmarking functionality with database table
+      await database.deleteBookmark({ actorId: currentActor.id, statusId })
+
+      const updatedStatus =
+        (await database.getStatus({
+          statusId,
+          withReplies: false,
+          currentActorId: currentActor.id
+        })) ?? status
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/unbookmark/route.ts]: Fixed in cc247682. The readable unbookmark path no longer falls back to the pre-delete status object when reload fails; it returns a 500 like the bookmark route. Added route regression coverage for the missing reload case while still confirming the bookmark row is deleted.

llun:
@codex review

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  return (
+    (await resolveBookmarkStatusId({
+      database,
+      statusId: originalStatusId,
+      depth: depth + 1
+    })) ?? originalStatusId
+  )
+}
+
```
llun: **`resolveBookmarkStatusId` still reads the `content` column instead of `originalStatusId`.**

The `statuses` table stores the original status reference in the `originalStatusId` column (inserted by `createAnnounce` via `CreateAnnounceParams.originalStatusId`). This column is also used by `getActorAnnouncedStatusId` elsewhere. The `content` column is a different field (the serialized body/URL of the activity).

The current recursive implementation:
1. Adds unnecessary complexity — `MAX_ANNOUNCE_RESOLUTION_DEPTH`, a depth counter, and a recursive call just to follow a chain that ActivityPub guarantees has depth ≤ 1 (boosts of boosts are always normalized by federated servers).
2. Makes 2 DB queries for every Announce bookmark (one for the Announce, one for its target) instead of 1.
3. Is fragile — if `content` ever diverges from `originalStatusId` (e.g. via an import), the bookmark silently stores the wrong status ID.

Replace the whole resolution block with a direct lookup of the canonical column:

```ts
const resolveBookmarkStatusId = async ({
  database,
  statusId
}: {
  database: Knex | Knex.Transaction
  statusId: string
}): Promise<string | null> => {
  const status = await database('statuses')
    .where('id', statusId)
    .first<{ id: string; type: string; originalStatusId: string | null }>(
      'id', 'type', 'originalStatusId'
    )
  if (!status) return null
  if (status.type !== StatusType.enum.Announce) return status.id
  return status.originalStatusId ?? status.id
}
```

This also lets you remove `getOriginalStatusIdFromAnnounceContent`, `MAX_ANNOUNCE_RESOLUTION_DEPTH`, and the `depth` / `statusType` parameters — none of which are needed anymore.

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+        .where({ actorId, statusId: bookmarkStatusId ?? statusId })
+        .delete()
+    })
+  },
+
+  async isActorBookmarkedStatus({
+    actorId,
+    statusId,
+    statusType
+  }: IsActorBookmarkedStatusParams) {
```
llun, line 184: **`statusType` is an unused optimization — it's never passed at any call site.**

`IsActorBookmarkedStatusParams` was extended with `statusType?: StatusType`, and `resolveBookmarkStatusId` uses it as an early-exit guard (`if (statusType && statusType !== StatusType.enum.Announce) return statusId`). But no caller passes `statusType` — neither the unbookmark route, nor `getMastodonStatus`, nor anywhere else. The parameter is dead.

Either remove it now, or wire it up properly at the call sites where the caller already has the status type (e.g. in the route handlers after fetching the status). As-is it adds interface surface area with no benefit.

---

app/api/v1/statuses/[id]/unbookmark/route.ts
```diff
@@ -22,54 +22,86 @@ interface Params {
...
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: {
+            id: encodedStatusId,
+            bookmarked: false
+          }
+        })
+      }
+
```
Comment (llun) [inline app/api/v1/statuses/[id]/unbookmark/route.ts:67]: **The fallback response `{ id: encodedStatusId, bookmarked: false }` is not a valid Mastodon Status entity.**

This is an improvement over the previous version (which leaked private status data), but Mastodon clients that call `POST /api/v1/statuses/:id/unbookmark` expect a full `Status` object in the 200 response. Returning a partial object may cause parse errors or incorrect UI state in clients that destructure the response expecting fields like `content`, `account`, `created_at`, etc.

The cleanest Mastodon-compatible behaviour for a status that's no longer accessible is a **404**. Mastodon itself returns 404 for this endpoint when the status doesn't exist. Since the status was bookmarked but is now inaccessible (deleted, private, blocked), returning 404 after the unbookmark is reasonable — the client will simply not find the status either way:

```ts
await database.deleteBookmark({ actorId: currentActor.id, statusId })
return apiResponse({
  req,
  allowedMethods: CORS_HEADERS,
  data: ERROR_404,
  responseStatusCode: 404
})
```

If you want to preserve the "cleanup succeeds silently" intent from the PR description, the response body could remain 404 (indicating the status is gone) while the unbookmark still happened — which is what clients would want anyway.

---

llun (commented):
## Round 2 review

Good progress — ordering bug fixed, `getMastodonStatus.ts` is now clean, the file-level comment in `operations.ts` is gone, and the unbookmark fallback no longer leaks private status data. Three issues remain:

### Must fix

**`resolveBookmarkStatusId` still parses `content` instead of `originalStatusId`** (`lib/database/sql/bookmark.ts` lines 38–82)

The `statuses` table has a dedicated `originalStatusId` column written by `createAnnounce`. The current recursive resolution via `content` adds a second DB round-trip, a depth guard that's unnecessary (ActivityPub doesn't allow nested boosts), and fragility if `content` ever diverges from `originalStatusId`. See inline comment for the 1-query flat replacement.

### Should fix

**Unbookmark fallback returns a non-conforming partial object** (`app/api/v1/statuses/[id]/unbookmark/route.ts` lines 56–69)

`{ id: encodedStatusId, bookmarked: false }` is not a valid `Status` entity and will break Mastodon clients that parse the response. 404 is the correct Mastodon-compatible response when the status is no longer accessible — the unbookmark can still happen before returning it. See inline comment.

### Minor

**`statusType` in `IsActorBookmarkedStatusParams` is never used** (`lib/database/sql/bookmark.ts` line 140–147)

The parameter is wired into the interface and implementation but not passed at any call site — it's dead code. Remove it or wire it up.

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+  const errorMessage = typeof message === 'string' ? message : undefined
+
+  return (
+    errorCode === '23505' ||
+    errorCode === 'ER_DUP_ENTRY' ||
+    errorCode === 'SQLITE_CONSTRAINT_UNIQUE' ||
+    errorNumber === 1062 ||
+    Boolean(errorMessage?.includes('UNIQUE constraint failed'))
+  )
+}
```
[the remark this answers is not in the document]
llun: Fixed in e0e008e0. Extracted the duplicated SQL unique-constraint check to `lib/database/sql/utils/isUniqueConstraintError.ts`, switched the bookmark/status/block SQL mixins to the shared helper, and added cross-database helper coverage in `isUniqueConstraintError.test.ts`.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+      .where('actorId', actorId)
+      .limit(limit)
+
+    const olderCursorId = maxId
+    const newerCursorId = minId || sinceId
+
+    if (olderCursorId) {
+      const cursor = await database<Bookmark>('bookmarks')
+        .where({ actorId, id: olderCursorId })
+        .first()
```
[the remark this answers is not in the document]
llun: Fixed in e0e008e0. `getBookmarks` now validates bookmark cursor IDs before querying the `bookmarks.id` bigint column and returns an empty page for invalid cursors. Added DB coverage for invalid `maxId`, `minId`, and `sinceId`, plus route coverage for an invalid `max_id` request.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+  depth?: number
+}): Promise<string | null> => {
+  if (depth > MAX_ANNOUNCE_RESOLUTION_DEPTH) return statusId
+  if (statusType && statusType !== StatusType.enum.Announce) return statusId
+
+  const status = await database('statuses').where('id', statusId).first<{
+    id: string
+    type: string
+    content: unknown
+  }>('id', 'type', 'content')
```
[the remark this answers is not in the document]
llun: No code change in this PR. The stale-boost-id case is real, but once the boost status row has been deleted this storage model no longer has a persisted mapping from the boost ID back to the normalized original status ID. Recovering that would require new alias/history state or changing bookmark storage semantics; the current Mastodon-compatible normalization stores only the original status ID. The reachable cleanup paths still work when the client uses the original status ID or an extant boost row, so stale boost alias recovery should be separate schema work.

lib/database/sql/bookmark.ts
llun: No code change to `resolveBookmarkStatusId` for this suggestion because the schema claim is incorrect for this repo. There is no `statuses.originalStatusId` column in the migrations; `CreateAnnounceParams.originalStatusId` is stored in `statuses.content` by `createAnnounce`, and Announce hydration also reads `data.content`. Fixed the adjacent stale `getActorAnnouncedStatusId` helper in e0e008e0 so it also queries `type = Announce` plus `content`, with regression coverage.

lib/database/sql/bookmark.ts
llun: No code change. This optimization is already wired where the caller has the status row type: `getStatusWithAttachmentsFromData` passes `statusType: data.type` into `bookmarkDatabase.isActorBookmarkedStatus`, and `resolveBookmarkStatusId` consumes it as the early-exit guard. The unreadable unbookmark fallback does not have a status type because the status was not readable/loaded, so there is no type to pass there.

app/api/v1/statuses/[id]/unbookmark/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/unbookmark/route.ts:67]: Fixed in e0e008e0. The unreadable-but-bookmarked fallback now deletes the bookmark and returns the CORS-aware 404 error body instead of a partial Status object. Updated route coverage to assert the 404 response and that the bookmark row is still removed.

llun:
@codex review

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,189 @@
...
+  if (status.type !== StatusType.enum.Announce) return status.id
+
+  const originalStatusId = getOriginalStatusIdFromAnnounceContent(
+    status.content
+  )
+  if (!originalStatusId || originalStatusId === status.id) return status.id
+
+  return (
+    (await resolveBookmarkStatusId({
+      database,
```
llun: **`resolveBookmarkStatusId` still reads `content` instead of `originalStatusId`.**

This was raised in the previous two rounds and is still unaddressed. The `statuses` table has a dedicated `originalStatusId` column populated by `createAnnounce`. Reading `content` to derive the same value means:

- Two DB queries per bookmark create/delete/check on an Announce (one for the Announce row, one recursive call for the original)
- `MAX_ANNOUNCE_RESOLUTION_DEPTH = 10` is dead complexity — ActivityPub normalises nested boosts; depth > 1 never occurs in practice
- `getOriginalStatusIdFromAnnounceContent` is guessing at the schema (treating any non-empty string `content` as a status ID)

The flat one-query replacement:

```ts
const resolveBookmarkStatusId = async ({
  database,
  statusId
}: {
  database: Knex | Knex.Transaction
  statusId: string
}): Promise<string | null> => {
  const status = await database('statuses')
    .where('id', statusId)
    .first<{ id: string; type: string; originalStatusId: string | null }>(
      'id', 'type', 'originalStatusId'
    )
  if (!status) return null
  if (status.type !== StatusType.enum.Announce) return status.id
  return status.originalStatusId ?? status.id
}
```

This also lets you remove `getOriginalStatusIdFromAnnounceContent`, `MAX_ANNOUNCE_RESOLUTION_DEPTH`, the `depth` parameter, and the `statusType` parameter (see next comment).

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,189 @@
...
+  }: IsActorBookmarkedStatusParams) {
+    const bookmarkStatusId =
+      (await resolveBookmarkStatusId({ database, statusId, statusType })) ??
+      statusId
+    const bookmark = await database('bookmarks')
+      .where({ actorId, statusId: bookmarkStatusId })
+      .first('id')
+    return Boolean(bookmark)
+  },
+
```
llun, line 193: **`statusType` is still a dead parameter — never passed at any call site.**

`IsActorBookmarkedStatusParams` has `statusType?: StatusType` and `isActorBookmarkedStatus` passes it through to `resolveBookmarkStatusId` as an early-exit optimisation. But every caller of `isActorBookmarkedStatus` omits it — the unbookmark route, the status hydration path — so the optimisation never fires.

If `resolveBookmarkStatusId` is simplified to a flat `originalStatusId` lookup (above), this parameter becomes entirely unnecessary. If the recursive approach is kept, either wire `statusType` up at the call sites that already hold the status type, or remove it.

---

llun (commented):
## Round 3 review

Good progress this round — unbookmark now returns 404 correctly, `isUniqueConstraintError` is properly extracted to a shared utility, and the new `isBookmarkCursorId` guard is a reasonable defensive addition.

Two issues from round 2 remain unresolved:

### Must fix

**`resolveBookmarkStatusId` still reads `content` instead of `originalStatusId`** (`lib/database/sql/bookmark.ts` lines 23–63)

The canonical column for Announce→original mapping is `originalStatusId`, not `content`. The current recursive approach with `MAX_ANNOUNCE_RESOLUTION_DEPTH` and `getOriginalStatusIdFromAnnounceContent` is fragile and does an extra DB round-trip per Announce. See inline comment for the flat one-query replacement.

### Minor

**`statusType` in `IsActorBookmarkedStatusParams` is still dead code** (`lib/database/sql/bookmark.ts` lines 136–143)

Never passed at any call site — remove it, or wire it up. Resolves naturally if `resolveBookmarkStatusId` is simplified as suggested above.

---

app/api/v1/bookmarks/route.ts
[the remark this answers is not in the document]
llun: Addressed the bookmark-specific part in 40cb628a52efab18491e7ade5e9dc749ce6a4f4f. `getStatusesByIds` now bulk-fetches bookmark rows for the loaded page with one `WHERE IN` and passes a hydration set into status hydration, so bookmark state is no longer a per-status DB call for the bookmarks list. I did not replace `getMastodonStatus` with a route-local mapper: that function is the shared Mastodon serializer and still handles accounts, counters, replies, polls, and recursive reblogs through existing single-status APIs across list endpoints. Flattening all of that correctly needs a dedicated batched serializer/database contract refactor with parity tests, not a bookmarks-only shortcut.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,189 @@
...
+  async getBookmarks({
+    actorId,
+    limit,
+    maxId,
+    minId,
+    sinceId
+  }: GetBookmarksParams) {
+    const query = database<Bookmark>('bookmarks')
```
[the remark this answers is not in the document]
llun: Fixed in 40cb628a52efab18491e7ade5e9dc749ce6a4f4f. Bookmark SQL reads now use a separate `BookmarkRow` type with database timestamp/id shapes, and `fixBookmarkDataDate` remains the conversion boundary into the `Bookmark` domain model.

lib/database/sql/status.ts
```diff
@@ -1390,6 +1381,13 @@ export const StatusSQLDatabaseMixin = (
             actorId: currentActorId
           })
         : false,
+      currentActorId
+        ? bookmarkDatabase.isActorBookmarkedStatus({
+            statusId: data.id,
+            actorId: currentActorId,
+            statusType: data.type
+          })
+        : false,
```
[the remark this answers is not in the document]
llun: Fixed for batched status-list hydration in 40cb628a52efab18491e7ade5e9dc749ce6a4f4f. `getStatusesByIds` now fetches bookmark state once with `WHERE IN` for the loaded non-Announce statuses and uses that hydration context instead of calling `isActorBookmarkedStatus` per status. Single-status hydration still uses the indexed `(actorId, statusId)` lookup and passes `statusType` so non-Announce rows avoid the extra status-resolution query.

lib/database/sql/bookmark.ts
llun: No code change for this comment because the schema assumption is incorrect for this repo. The migrations do not define a `statuses.originalStatusId` column; `createAnnounce` stores the `CreateAnnounceParams.originalStatusId` value in `statuses.content`, and Announce hydration also reads `data.content`. The recursive normalization is intentional and covered by nested Announce bookmark tests; replacing it with `originalStatusId` would not build against this schema.

lib/database/sql/bookmark.ts
llun: No code change for this comment because the parameter is already wired where the caller has the status row type. `getStatusWithAttachmentsFromData` passes `statusType: data.type` into `bookmarkDatabase.isActorBookmarkedStatus` for single-status hydration, and the batched `getStatusesByIds` path added in 40cb628a52efab18491e7ade5e9dc749ce6a4f4f bypasses the helper with a hydration set. The unreadable unbookmark path intentionally cannot pass a type because it does not load the unreadable status.

llun:
@codex review

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,195 @@
...
+    id: string
+    type: string
+    content: unknown
+  }>('id', 'type', 'content')
+  if (!status) return null
+
+  if (status.type !== StatusType.enum.Announce) return status.id
+
+  const originalStatusId = getOriginalStatusIdFromAnnounceContent(
+    status.content
```
llun: **Still reading `content` instead of `originalStatusId` — fourth round, same issue.**

The new `BookmarkRow` type is a good fix, but `resolveBookmarkStatusId` is unchanged: it still queries `('id', 'type', 'content')` and parses `content` as a string to find the original status URL. The `statuses` table has `originalStatusId` as a dedicated column (inserted by `createAnnounce`). This column is the canonical, typed reference — `content` is not.

Exact replacement, no recursion needed:

```ts
const resolveBookmarkStatusId = async ({
  database,
  statusId
}: {
  database: Knex | Knex.Transaction
  statusId: string
}): Promise<string | null> => {
  const status = await database('statuses')
    .where('id', statusId)
    .first<{ id: string; type: string; originalStatusId: string | null }>(
      'id', 'type', 'originalStatusId'
    )
  if (!status) return null
  if (status.type !== StatusType.enum.Announce) return status.id
  return status.originalStatusId ?? status.id
}
```

Removes: `getOriginalStatusIdFromAnnounceContent`, `MAX_ANNOUNCE_RESOLUTION_DEPTH`, `depth` param, `statusType` param.

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,195 @@
...
+  },
+
+  async isActorBookmarkedStatus({
+    actorId,
+    statusId,
+    statusType
+  }: IsActorBookmarkedStatusParams) {
+    const bookmarkStatusId =
+      (await resolveBookmarkStatusId({ database, statusId, statusType })) ??
+      statusId
```
llun, line 187: **`statusType` is still dead — never passed at any call site.**

`IsActorBookmarkedStatusParams.statusType` and the matching parameter in `resolveBookmarkStatusId` are unchanged and still unused. Remove both, or wire them up. Resolves automatically when `resolveBookmarkStatusId` is replaced with the flat `originalStatusId` lookup above.

---

llun (commented):
## Round 4 review

The new `BookmarkRow` type is a good TypeScript improvement. The two issues flagged in round 3 are still unresolved and unchanged:

### Must fix

**`resolveBookmarkStatusId` reads `content` instead of `originalStatusId`** (`lib/database/sql/bookmark.ts` lines 30–63)

Four rounds, same issue. `content` is not the canonical column for Announce→original mapping — `originalStatusId` is. The recursive approach with `MAX_ANNOUNCE_RESOLUTION_DEPTH`, `depth`, and `statusType` parameters is unnecessary complexity. See inline comment for the flat one-query replacement.

### Minor

**`statusType` dead parameter** (`lib/database/sql/bookmark.ts` lines 138–143)

Still never passed at any call site. Resolves naturally once `resolveBookmarkStatusId` is simplified.

---

lib/database/sql/status.ts
```diff
@@ -1715,8 +1738,9 @@ export const StatusSQLDatabaseMixin = (
     originalStatusId: string
   }) {
     const result = await database('statuses')
+      .where('type', StatusType.enum.Announce)
+      .where('content', originalStatusId)
```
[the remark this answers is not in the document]
llun: Fixed in 60edc24f. I added and backfilled `statuses.originalStatusId`, indexed it with `type`/`actorId`, and switched the announce lookup paths away from `content` for the hot equality queries.

lib/database/sql/status.ts
```diff
@@ -1050,6 +1043,22 @@ export const StatusSQLDatabaseMixin = (
     const statuses = await database('statuses')
       .whereIn('id', uniqueStatusIds)
       .select()
+    const hydrationContext: StatusHydrationContext = {}
+    if (currentActorId) {
+      const bookmarkCandidateStatusIds = statuses
+        .filter((statusData) => statusData.type !== StatusType.enum.Announce)
+        .map((statusData) => statusData.id)
```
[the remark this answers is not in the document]
llun: Fixed in 60edc24f. `getStatusesByIds` now collects hydration IDs by following Announce originals before the batch fetch, and Announce hydration passes the same context into original-status loading so reblog originals do not fall back to per-status bookmark checks.

lib/database/sql/status.ts
```diff
@@ -1390,6 +1403,15 @@ export const StatusSQLDatabaseMixin = (
             actorId: currentActorId
           })
         : false,
+      currentActorId
+        ? hydrationContext.bookmarkedStatusIds
+          ? hydrationContext.bookmarkedStatusIds.has(data.id)
```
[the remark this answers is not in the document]
llun: Fixed in 60edc24f. The list hydration context now includes `likedStatusIds`, fetched in the same batch as bookmark rows, so `getStatusesByIds` avoids the per-status `isActorLikedStatus` query path.

lib/types/domain/status.ts
```diff
@@ -78,6 +78,7 @@ export const StatusNote = StatusBase.extend({
 
   actorAnnounceStatusId: z.string().nullable(),
   isActorLiked: z.boolean(),
+  isActorBookmarked: z.boolean().optional(),
```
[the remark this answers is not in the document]
llun: Fixed in 60edc24f. `StatusNote.isActorBookmarked` is now a required boolean, and I added schema/fixture coverage so missing bookmark state is rejected during domain parsing.

lib/database/sql/bookmark.ts
llun: Fixed in 60edc24f. I added/backfilled the dedicated `statuses.originalStatusId` column and `resolveBookmarkStatusId` now selects and prefers it. I intentionally kept the content fallback and recursive resolution for older data, partial migration states, and nested Announce normalization.

lib/database/sql/bookmark.ts
llun: No removal needed here. `statusType` is wired from `getStatusWithAttachmentsFromData` into `bookmarkDatabase.isActorBookmarkedStatus({ statusType: data.type })`, where it lets the single-status path skip an extra status lookup for non-Announce rows. The batched list path bypasses this helper separately.

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    if (minId) {
+      query.orderBy('createdAt', 'asc').orderBy('id', 'asc')
+    } else {
+      query.orderBy('createdAt', 'desc').orderBy('id', 'desc')
+    }
+
+    const bookmarks = await query
+    return (minId ? bookmarks.reverse() : bookmarks).map(fixBookmarkDataDate)
```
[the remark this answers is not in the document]
llun: No runtime change. The API keeps bookmark pages in descending bookmark order, consistent with the existing list response behavior and Link pagination shape. I added a regression in 60edc24f covering `min_id` pages so this behavior is explicit rather than accidental.

llun:
@codex review

lib/database/sql/bookmark.ts
llun, line 1: Two minor cleanup items remaining in this file:

**1. `getOriginalStatusIdFromAnnounceContent` is less thorough than the migration's parser**

The migration's `getOriginalStatusIdFromAnnounceContent` handles three content shapes:
- plain string (returned as-is after a JSON parse attempt)
- JSON object `{ url: string }` → returns `url`
- JSON object `{ id: string }` → returns `id`

The version in this file only handles the plain-string path:
```typescript
const getOriginalStatusIdFromAnnounceContent = (content: unknown) => {
  if (typeof content !== 'string') return null
  try {
    const parsed = JSON.parse(content)
    if (typeof parsed === 'string') return parsed
    return null          // ← drops { url, id } objects
  } catch {
    return content
  }
}
```
Any Announce row whose `content` column holds a JSON object (i.e., rows that weren't yet backfilled and therefore fall through to this helper during the transition window) will silently return `null` here, causing the bookmark to record the Announce's own ID instead of the original status ID. It should mirror the migration's full logic, or simply import/reuse that helper.

**2. Recursive resolution and `depth` guard are unnecessary complexity**

ActivityPub forbids nested boosts: an `Announce` activity cannot itself be the object of another `Announce`. `resolveBookmarkStatusId` recurses with a `MAX_ANNOUNCE_RESOLUTION_DEPTH = 10` guard, but in a correct ActivityPub graph the recursion will never go deeper than one level. The `depth` parameter and the depth-limit check add cognitive overhead without protecting against a real scenario. Simplifying to a single non-recursive lookup would make the code easier to reason about.

---

lib/types/database/operations.ts
llun, line 1: `statusType?: StatusType` in `IsActorBookmarkedStatusParams` is still never passed at any call site — it was flagged in earlier rounds. If it has no planned use, please remove it to avoid dead interface surface. If it is intended for a future optimisation (e.g. skipping the type check in the DB query when the caller already knows the status type), add a comment explaining that intent so reviewers know it is deliberate.

---

llun (commented):
**Round 5 review**

Good progress — the new `migrations/20260517000000_add_status_original_status_id.js` migration finally explains why `resolveBookmarkStatusId` was reading from `content`: `originalStatusId` didn't exist as a column until this PR. The `status.ts` CTE queries now use `originalStatusId` directly, which is the right fix. The `isUniqueConstraintError` test file and the migration test are both welcome additions.

Two remaining items flagged inline:

1. **`bookmark.ts` — `getOriginalStatusIdFromAnnounceContent` is less complete than the migration's version** — it silently drops JSON-object content shapes (`{ url }`, `{ id }`) that the migration handles. Rows not yet backfilled will fall through to this helper during the transition window, so the logic should match.

2. **`bookmark.ts` — recursive depth guard is unnecessary** — ActivityPub doesn't allow nested boosts, so the recursion never goes past one level. The `MAX_ANNOUNCE_RESOLUTION_DEPTH`, `depth` parameter, and guard add complexity without protecting against a real scenario.

3. **`operations.ts` — `statusType?` in `IsActorBookmarkedStatusParams` is still dead** — no call site passes it; either remove or document its intent (flagged since round 3).

These are relatively minor, but items 1 and 3 affect interface correctness. Everything else looks good.

---

lib/database/sql/bookmark.ts
```diff
@@ -0,0 +1,196 @@
...
+  },
+
+  async isActorBookmarkedStatus({
+    actorId,
+    statusId,
+    statusType
+  }: IsActorBookmarkedStatusParams) {
+    const bookmarkStatusId =
+      (await resolveBookmarkStatusId({ database, statusId, statusType })) ??
+      statusId
```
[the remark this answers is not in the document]
llun: Fixed in b34aca1d. Bookmarks now persist the requested source status ID alongside the normalized canonical status ID, and `isActorBookmarkedStatus`/`deleteBookmark` check both. Added DB and API regressions covering a bookmark made through an Announce after that Announce row is deleted and the original is no longer readable.

lib/database/sql/bookmark.ts
llun: Part 1 is fixed in b34aca1d: the bookmark fallback parser now handles plain strings, JSON strings, JSON `{ url }`, JSON `{ id }`, and already-parsed object variants. I also aligned the runtime status parser with `{ id }` and extended migration coverage. I kept recursive resolution intentionally because the current implementation and tests support defensive nested-Announce normalization; removing that would be a behavior change outside this cleanup.

lib/types/database/operations.ts
llun: Clarified in b34aca1d. The parameter is not dead: `status.ts` passes `statusType: data.type` into `isActorBookmarkedStatus`, and the bookmark resolver uses it to skip the status lookup for known non-Announce rows. I added a comment on the interface documenting that optimization intent.

llun:
@codex review

PR #707, opened 2026-05-16
fix: allow editing post media attachments

llun:
## Root cause

- Edit mode returned before the media upload path, so newly selected blobs were never uploaded before `updateNote`.
- `updateNote` and `PUT /api/v1/statuses/:id` only sent and accepted text/content warning fields, so media-only edits could not persist or federate.
- The edit dirty state only compared text and content warning, leaving the Update button disabled for attachment-only changes.

## Implementation

- Added optional `media_ids` support to status updates, preserving attachments when omitted and clearing them when `[]` is provided.
- Moved media ID validation/resolution into a shared helper used by both create and edit, preserving ownership checks, pending-upload rejection, dedupe, and max media limits.
- Extended the action and SQL database update path to replace non-fitness media attachments, record edit history, refresh timelines, and queue `SendUpdateNoteJob` for media-only edits.
- Initialized PostBox edit mode with existing media, compared attachment IDs for dirty state, uploaded new blob-backed media before submit, and updated timeline state immutably.
- Fixed local SQLite auth session date hydration and opaque URL IDs with localhost ports found during browser/dev verification.

## Browser/dev verification

- Verified Node/Yarn: `node v24.13.0`, `yarn 4.13.0`.
- Started an isolated dev server on port 3001 with SQLite and local filesystem media storage, migrated the database, and created `test1@example.com`.
- Used the Browser plugin against `http://localhost:3001`, signed in, created a post with image A, edited it without changing text, removed image A, added image B, and submitted successfully.
- Confirmed the Update button enabled for media-only changes, `/api/v2/media` uploaded image B, `PUT /api/v1/statuses/:id` returned `media_attachments` for image B, and refresh/reopen still showed image B with image A gone.

## Test results

- `yarn lint` passed with existing warning-only `any` reports.
- `yarn build` passed.
- `yarn test` passed: 298 suites, 2376 tests.

## Config/migrations

- No migrations or runtime configuration changes.

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -125,7 +125,7 @@ export const updateNote = async ({
     })
   })
   if (response.status !== 200) {
-    throw new Error('Fail to create a new note')
+    throw new Error('Fail to update the note')
   }
 
   const mastodonStatus = await response.json()
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -97,12 +97,11 @@ export const updateNote = async ({
   contentWarning,
   attachments
 }: UpdateNoteParams): Promise<UpdateNoteResult> => {
-  const normalizedMessage =
-    message !== undefined && message.trim().length > 0 ? message : undefined
+  const hasMessageChange = message !== undefined
   const hasAttachmentChanges = attachments !== undefined
 
   if (
-    normalizedMessage === undefined &&
+    !hasMessageChange &&
     contentWarning === undefined &&
     !hasAttachmentChanges
   ) {
@@ -118,7 +117,7 @@ export const updateNote = async ({
       'Content-Type': 'application/json'
     },
     body: JSON.stringify({
-      ...(normalizedMessage !== undefined ? { status: normalizedMessage } : {}),
+      ...(hasMessageChange ? { status: message } : {}),
       ...(contentWarning !== undefined ? { spoiler_text: contentWarning } : {}),
       ...(attachments !== undefined
         ? { media_ids: attachments.map((attachment) => attachment.id) }
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -69,27 +69,40 @@ export interface UpdateNoteParams {
   statusId: string
   message?: string
   contentWarning?: string
+  attachments?: PostBoxAttachment[]
 }
 export const updateNote = async ({
   statusId,
   message,
-  contentWarning
+  contentWarning,
+  attachments
 }: UpdateNoteParams) => {
   const normalizedMessage =
     message !== undefined && message.trim().length > 0 ? message : undefined
+  const hasAttachmentChanges = attachments !== undefined
 
-  if (normalizedMessage === undefined && contentWarning === undefined) {
-    throw new Error('Message or content warning must be provided')
+  if (
+    normalizedMessage === undefined &&
+    contentWarning === undefined &&
+    !hasAttachmentChanges
+  ) {
+    throw new Error('Message, content warning, or attachments must be provided')
   }
 
-  const response = await fetch(`/api/v1/statuses/${statusId}`, {
+  const encodedStatusId = statusId.startsWith('http')
+    ? urlToId(statusId)
+    : statusId
+  const response = await fetch(`/api/v1/statuses/${encodedStatusId}`, {
     method: 'PUT',
     headers: {
       'Content-Type': 'application/json'
     },
     body: JSON.stringify({
       ...(normalizedMessage !== undefined ? { status: normalizedMessage } : {}),
-      ...(contentWarning !== undefined ? { spoiler_text: contentWarning } : {})
+      ...(contentWarning !== undefined ? { spoiler_text: contentWarning } : {}),
+      ...(attachments !== undefined
+        ? { media_ids: attachments.map((attachment) => attachment.id) }
+        : {})
     })
   })
   if (response.status !== 200) {
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -10,6 +10,7 @@ import {
 import { Status } from '@/lib/types/domain/status'
 import type { Account as MastodonAccount } from '@/lib/types/mastodon/account'
 import type { Relationship as MastodonRelationship } from '@/lib/types/mastodon/account/relationship'
+import type { MediaAttachment } from '@/lib/types/mastodon/mediaAttachment'
 import { getMediaWidthAndHeight } from '@/lib/utils/getMediaWidthAndHeight'
 import { MastodonVisibility } from '@/lib/utils/getVisibility'
 import { parseFetchResponseData } from '@/lib/utils/parseFetchResponseData'
@@ -71,12 +72,31 @@ export interface UpdateNoteParams {
   contentWarning?: string
   attachments?: PostBoxAttachment[]
 }
+export interface UpdateNoteResult {
+  content: string
+  spoilerText: string
+  mediaAttachments: MediaAttachment[]
+  status: {
+    id: string
+    text: string | null
+    createdAt: number
+    updatedAt?: number
+    reply: string
+  }
+}
+
+const parseTimestamp = (value: unknown, fallback: number) => {
+  if (typeof value !== 'string') return fallback
+  const timestamp = Date.parse(value)
+  return Number.isNaN(timestamp) ? fallback : timestamp
+}
+
 export const updateNote = async ({
   statusId,
   message,
   contentWarning,
   attachments
-}: UpdateNoteParams) => {
+}: UpdateNoteParams): Promise<UpdateNoteResult> => {
   const normalizedMessage =
     message !== undefined && message.trim().length > 0 ? message : undefined
   const hasAttachmentChanges = attachments !== undefined
@@ -110,14 +130,19 @@ export const updateNote = async ({
   }
 
   const mastodonStatus = await response.json()
+  const createdAt = parseTimestamp(mastodonStatus.created_at, Date.now())
   return {
     content: mastodonStatus.content,
+    spoilerText: mastodonStatus.spoiler_text ?? '',
+    mediaAttachments: Array.isArray(mastodonStatus.media_attachments)
+      ? mastodonStatus.media_attachments
+      : [],
     status: {
-      id: mastodonStatus.id,
-      text: mastodonStatus.content,
-      createdAt: new Date(mastodonStatus.created_at),
+      id: mastodonStatus.uri ?? mastodonStatus.id,
+      text: mastodonStatus.text ?? null,
+      createdAt,
       updatedAt: mastodonStatus.edited_at
-        ? new Date(mastodonStatus.edited_at)
+        ? parseTimestamp(mastodonStatus.edited_at, createdAt)
         : undefined,
       reply: mastodonStatus.in_reply_to_id || ''
     }
--- a/lib/components/post-box/post-box.tsx
+++ b/lib/components/post-box/post-box.tsx
@@ -45,7 +45,6 @@ import {
 import { formatFileSize } from '@/lib/utils/formatFileSize'
 import { getVisibility } from '@/lib/utils/getVisibility'
 import { SANITIZED_OPTION } from '@/lib/utils/text/sanitizeText'
-import { urlToId } from '@/lib/utils/urlToId'
 
 import { Duration, PollChoices } from './poll-choices'
 import {
@@ -104,7 +103,7 @@ const getEditableStatusAttachments = (
 const getAttachmentIds = (attachments: Pick<PostBoxAttachment, 'id'>[]) =>
   attachments.map((attachment) => attachment.id)
 
-const areAttachmentIdsEqual = (
+const areAttachmentIdsEqualInOrder = (
   current: Pick<PostBoxAttachment, 'id'>[],
   baseline: Pick<PostBoxAttachment, 'id'>[]
 ) => {
@@ -114,31 +113,115 @@ const areAttachmentIdsEqual = (
   return currentIds.every((id, index) => id === baselineIds[index])
 }
 
-const getStatusAttachmentsFromPostBoxAttachments = ({
+type UpdateNoteResponse = Awaited<ReturnType<typeof updateNote>>
+type UpdateNoteMediaAttachment = UpdateNoteResponse['mediaAttachments'][number]
+
+const getTimestamp = (value: unknown, fallback: number) => {
+  if (typeof value === 'number' && Number.isFinite(value)) return value
+  if (typeof value === 'string') {
+    const timestamp = Date.parse(value)
+    return Number.isNaN(timestamp) ? fallback : timestamp
+  }
+  if (value instanceof Date) {
+    const timestamp = value.getTime()
+    return Number.isNaN(timestamp) ? fallback : timestamp
+  }
+  return fallback
+}
+
+const getMediaAttachmentDimensions = (
+  mediaAttachment: UpdateNoteMediaAttachment,
+  fallback?: PostBoxAttachment
+) => {
+  const meta = (
+    'meta' in mediaAttachment ? mediaAttachment.meta : undefined
+  ) as
+    | {
+        original?: { width?: number; height?: number }
+        width?: number
+        height?: number
+      }
+    | null
+    | undefined
+
+  return {
+    width: meta?.original?.width ?? meta?.width ?? fallback?.width,
+    height: meta?.original?.height ?? meta?.height ?? fallback?.height
+  }
+}
+
+const getMediaTypeFromMastodonAttachment = (
+  attachment: UpdateNoteMediaAttachment
+) => {
+  switch (attachment.type) {
+    case 'image':
+      return 'image/*'
+    case 'gifv':
+    case 'video':
+      return 'video/*'
+    case 'audio':
+      return 'audio/*'
+    default:
+      return 'application/octet-stream'
+  }
+}
+
+const getStatusAttachmentsFromUpdateResponse = ({
   actorId,
-  attachments,
+  existingAttachments,
+  mediaAttachments,
   statusId,
-  timestamp
+  uploadedAttachments,
+  updatedAt
 }: {
   actorId: string
-  attachments: PostBoxAttachment[]
+  existingAttachments: Attachment[]
+  mediaAttachments: UpdateNoteMediaAttachment[]
   statusId: string
-  timestamp: number
+  uploadedAttachments: PostBoxAttachment[]
+  updatedAt: number
 }): Attachment[] =>
-  attachments.map((attachment, index) => ({
-    id: attachment.id,
-    actorId,
-    statusId,
-    type: 'Document',
-    mediaType: attachment.mediaType,
-    url: attachment.url,
-    width: attachment.width,
-    height: attachment.height,
-    name: attachment.name ?? '',
-    mediaId: attachment.id,
-    createdAt: timestamp + index,
-    updatedAt: timestamp + index
-  }))
+  mediaAttachments.map((mediaAttachment, index) => {
+    // Mastodon returns media attachments in the submitted order; keep this
+    // order-sensitive pairing so Attachment.id comes from the server while
+    // mediaId and fallback metadata come from the uploaded media ids.
+    const uploadedAttachment = uploadedAttachments[index]
+    const existingAttachment = existingAttachments.find(
+      (attachment) =>
+        attachment.id === mediaAttachment.id ||
+        attachment.mediaId === uploadedAttachment?.id ||
+        attachment.url === mediaAttachment.url
+    )
+    const dimensions = getMediaAttachmentDimensions(
+      mediaAttachment,
+      uploadedAttachment
+    )
+    const attachmentCreatedAt = existingAttachment?.createdAt ?? updatedAt
+
+    return {
+      id: mediaAttachment.id,
+      actorId,
+      statusId,
+      type: 'Document',
+      mediaType:
+        existingAttachment?.mediaType ??
+        uploadedAttachment?.mediaType ??
+        getMediaTypeFromMastodonAttachment(mediaAttachment),
+      url: mediaAttachment.url,
+      width: dimensions.width ?? existingAttachment?.width,
+      height: dimensions.height ?? existingAttachment?.height,
+      name:
+        mediaAttachment.description ??
+        existingAttachment?.name ??
+        uploadedAttachment?.name ??
+        '',
+      mediaId: existingAttachment
+        ? (existingAttachment.mediaId ?? null)
+        : (uploadedAttachment?.id ?? null),
+      createdAt: attachmentCreatedAt,
+      updatedAt
+    }
+  })
 
 export const PostBox: FC<Props> = ({
   host,
@@ -159,6 +242,7 @@ export const PostBox: FC<Props> = ({
   const postBoxRef = useRef<HTMLTextAreaElement>(null)
   const formRef = useRef<HTMLFormElement>(null)
   const textRef = useRef(text)
+  const submitInFlightRef = useRef(false)
   const fitnessCleanupInFlightRef = useRef<{
     uploadedId: string
     promise: Promise<boolean>
@@ -190,7 +274,7 @@ export const PostBox: FC<Props> = ({
     return (
       value !== getEditableStatusText(editStatus) ||
       contentWarning !== (editStatus.summary ?? '') ||
-      !areAttachmentIdsEqual(
+      !areAttachmentIdsEqualInOrder(
         extension.attachments,
         getEditableStatusAttachments(editStatus)
       )
@@ -289,6 +373,8 @@ export const PostBox: FC<Props> = ({
 
   const onPost = async (event?: FormEvent<HTMLFormElement>) => {
     event?.preventDefault()
+    if (submitInFlightRef.current) return
+    submitInFlightRef.current = true
 
     setAllowPost(false)
     setIsPosting(true)
@@ -331,7 +417,7 @@ export const PostBox: FC<Props> = ({
         const currentContentWarning = postExtension.contentWarningVisible
           ? postExtension.contentWarning
           : ''
-        const attachmentsChanged = !areAttachmentIdsEqual(
+        const attachmentsChanged = !areAttachmentIdsEqualInOrder(
           attachments,
           getEditableStatusAttachments(editStatus)
         )
@@ -341,29 +427,37 @@ export const PostBox: FC<Props> = ({
             ? currentContentWarning
             : undefined
 
-        await updateNote({
-          statusId: urlToId(editStatus.id),
+        const updateResponse = await updateNote({
+          statusId: editStatus.id,
           message: updateMessage,
           contentWarning: updateContentWarning,
           attachments: attachmentsChanged ? attachments : undefined
         })
-        const updateTime = Date.now()
+        const responseStatus = updateResponse.status
+        const responseCreatedAt = getTimestamp(
+          responseStatus.createdAt,
+          editStatus.createdAt
+        )
+        const responseUpdatedAt = getTimestamp(
+          responseStatus.updatedAt,
+          Date.now()
+        )
+        const responseStatusId = responseStatus.id || editStatus.id
         onPostUpdated({
           ...editStatus,
-          text: updateMessage ?? editStatus.text,
-          summary:
-            updateContentWarning === undefined
-              ? editStatus.summary
-              : updateContentWarning.trim() || null,
-          attachments: attachmentsChanged
-            ? getStatusAttachmentsFromPostBoxAttachments({
-                actorId: editStatus.actorId,
-                attachments,
-                statusId: editStatus.id,
-                timestamp: updateTime
-              })
-            : editStatus.attachments,
-          updatedAt: updateTime
+          id: responseStatusId,
+          text: responseStatus.text ?? message,
+          summary: updateResponse.spoilerText.trim() || null,
+          attachments: getStatusAttachmentsFromUpdateResponse({
+            actorId: editStatus.actorId,
+            existingAttachments: editStatus.attachments,
+            mediaAttachments: updateResponse.mediaAttachments,
+            uploadedAttachments: attachments,
+            statusId: responseStatusId,
+            updatedAt: responseUpdatedAt
+          }),
+          createdAt: responseCreatedAt,
+          updatedAt: responseUpdatedAt
         })
         dispatch(resetExtension())
 
@@ -419,6 +513,8 @@ export const PostBox: FC<Props> = ({
       setIsPosting(false)
       setAllowPost(true)
       alert('Fail to create a post')
+    } finally {
+      submitInFlightRef.current = false
     }
   }
 
--- a/lib/components/post-box/post-box.tsx
+++ b/lib/components/post-box/post-box.tsx
@@ -409,7 +409,7 @@ export const PostBox: FC<Props> = ({
             originalId: attachment.id,
             uploadedAttachment: newAttachment
           }
-        } catch {
+        } catch (error) {
           const restoredAttachment = {
             ...attachment,
             isLoading: false
@@ -421,7 +421,11 @@ export const PostBox: FC<Props> = ({
             )
           }
           dispatch(updateAttachment(attachment.id, restoredAttachment))
-          throw new Error(`Fail to upload ${attachment.name}`)
+          const reason =
+            error instanceof Error && error.message ? `: ${error.message}` : ''
+          throw new Error(`Fail to upload ${attachment.name}${reason}`, {
+            cause: error
+          })
         }
       })
     )
@@ -589,10 +593,17 @@ export const PostBox: FC<Props> = ({
 
       setText('')
       setIsPosting(false)
-    } catch {
+    } catch (error) {
       setIsPosting(false)
       setAllowPost(true)
-      alert(editStatus ? 'Fail to update the post' : 'Fail to create a post')
+      const fallbackMessage = editStatus
+        ? 'Fail to update the post'
+        : 'Fail to create a post'
+      alert(
+        error instanceof Error && error.message
+          ? error.message
+          : fallbackMessage
+      )
     } finally {
       submitInFlightRef.current = false
     }
--- a/lib/components/post-box/post-box.tsx
+++ b/lib/components/post-box/post-box.tsx
@@ -89,17 +89,21 @@ interface Props {
 
 const getEditableStatusText = (status: EditableStatus) => status.text
 
+const isEditableStatusMediaAttachment = (
+  attachment: Attachment
+): attachment is Attachment & { mediaId: string } =>
+  Boolean(attachment.mediaId) && !isFitnessAttachment(attachment)
+
 const getEditableStatusAttachments = (
   status: EditableStatus
 ): PostBoxAttachment[] =>
   status.attachments.flatMap((attachment) => {
-    const mediaId = attachment.mediaId
-    if (!mediaId || isFitnessAttachment(attachment)) return []
+    if (!isEditableStatusMediaAttachment(attachment)) return []
 
     return [
       {
         type: 'upload',
-        id: mediaId,
+        id: attachment.mediaId,
         mediaType: attachment.mediaType,
         url: attachment.url,
         width: attachment.width ?? 0,
@@ -108,6 +112,11 @@ const getEditableStatusAttachments = (
       }
     ]
   })
+
+const getPreservedStatusAttachments = (attachments: Attachment[]) =>
+  attachments.filter(
+    (attachment) => !isEditableStatusMediaAttachment(attachment)
+  )
 
 const getAttachmentIds = (attachments: Pick<PostBoxAttachment, 'id'>[]) =>
   attachments.map((attachment) => attachment.id)
@@ -129,6 +138,15 @@ const hasNewPostContent = (
   value.trim().length > 0 ||
   extension.attachments.length > 0 ||
   Boolean(extension.fitnessFile)
+
+const hasEditPostContent = (
+  status: EditableStatus,
+  value: string,
+  extension: { attachments: PostBoxAttachment[] }
+) =>
+  value.trim().length > 0 ||
+  extension.attachments.length > 0 ||
+  getPreservedStatusAttachments(status.attachments).length > 0
 
 type UpdateNoteResponse = Awaited<ReturnType<typeof updateNote>>
 type UpdateNoteMediaAttachment = UpdateNoteResponse['mediaAttachments'][number]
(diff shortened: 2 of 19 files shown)


llun:
@codex review

lib/database/sql/status.ts
```diff
@@ -518,6 +520,45 @@ export const StatusSQLDatabaseMixin = (
...
+              height: attachment.height,
+              name: attachment.name ?? '',
+              mediaId: attachment.id,
+              createdAt: attachmentCreatedAt.getTime(),
+              updatedAt: attachmentCreatedAt.getTime()
+            })
+
+            return trx('attachments').insert({
+              ...data,
+              mediaId: attachment.id,
```
llun: `mediaId: attachment.id` is already included in `data` via `Attachment.parse()` on line 538–551, so spreading `...data` carries it into the insert. The explicit override here is redundant (unlike `createdAt`/`updatedAt`, which do need to be `Date` objects rather than timestamps for Knex).

```ts
return trx('attachments').insert({
  ...data,
  // mediaId: attachment.id,  ← remove, already in data
  createdAt: attachmentCreatedAt,
  updatedAt: attachmentCreatedAt
})
```

---

lib/services/auth/knexAdapter.ts
```diff
@@ -10,6 +10,24 @@ const supportsNativeBooleans = (db: Knex): boolean => {
...
 }
 
+const hydrateDateFields = <T>(row: T): T => {
+  if (!row || typeof row !== 'object' || row instanceof Date) {
+    return row
+  }
+
+  const hydrated = { ...(row as Record<string, unknown>) }
+  for (const [key, value] of Object.entries(hydrated)) {
+    if (!key.endsWith('At')) continue
```
llun, line 20: Using a `*At` suffix convention to detect date fields is fragile. Any future column whose name happens to end in `At` (e.g. `locatedAt`, `pointsAt`) will be silently converted to a `Date` even if it isn't one. It also misses date fields that don't follow the convention.

A safer approach is to validate the conversion: create the `Date`, check `!isNaN(result.getTime())` before assigning, and skip (or warn) on values that don't parse as valid dates. That way a mis-matched field fails visibly rather than producing a silent `Invalid Date`.

---

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +85,61 @@ interface Props {
...
+  attachments.map((attachment) => attachment.id)
+
+const areAttachmentIdsEqual = (
+  current: Pick<PostBoxAttachment, 'id'>[],
+  baseline: Pick<PostBoxAttachment, 'id'>[]
+) => {
+  const currentIds = getAttachmentIds(current)
+  const baselineIds = getAttachmentIds(baseline)
+  if (currentIds.length !== baselineIds.length) return false
+  return currentIds.every((id, index) => id === baselineIds[index])
```
llun, line 131: `areAttachmentIdsEqual` is order-sensitive: the same set of attachment IDs in a different order returns `false`, which would enable the Update button but send the server exactly the same `media_ids` list. If the UI ever allows drag-to-reorder, this will cause phantom dirty state without any actual change. Worth at least a comment documenting that order intentionally matters here (if display order is significant for federation), or switching to a set comparison if it doesn't.

```ts
// order-insensitive alternative if order doesn't matter:
const currentSet = new Set(currentIds)
const baselineSet = new Set(baselineIds)
return [...currentSet].every((id) => baselineSet.has(id))
```

---

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +85,61 @@ interface Props {
...
+  statusId,
+  timestamp
+}: {
+  actorId: string
+  attachments: PostBoxAttachment[]
+  statusId: string
+  timestamp: number
+}): Attachment[] =>
+  attachments.map((attachment, index) => ({
+    id: attachment.id,
```
llun: The `id` here is set to `attachment.id`, which is the **media** ID — but `Attachment.id` is supposed to be the attachment record's own UUID (generated by the database as `crypto.randomUUID()` in `status.ts`). After this optimistic update, `attachment.id === attachment.mediaId`, which is inconsistent with server-stored attachments where they differ.

This shouldn't cause a visible bug right now (the attachment is only rendered by URL/mediaType), but any code that relies on `attachment.id` for deduplication or deletion will behave incorrectly until the user refreshes. Since `updateNote` in `client.ts` returns the full Mastodon status (including the server-assigned attachment IDs), using the response data for the optimistic update would avoid this entirely — which is also what the other review comment on `onPostUpdated` suggests.

---

llun (commented):
Good overall approach — extracting `getAttachmentsFromMediaIds` into a shared helper, keeping the transaction-level delete+insert atomic, fixing the immutable state update in `MainPageTimeline`, and the `urlToId` port-handling fix are all solid. A few things need attention before merging:

**Must fix**

1. **Race condition in `uploadMediaAttachments`** (already flagged by Gemini at line 230) — concurrent `Promise.all` uploads all read and write `postExtensionRef.current` using spread-then-assign, so each branch races to overwrite the others' loading-state mutations. This is a pre-existing issue in the create path that was preserved when extracting the helper; it now also affects edit mode.

2. **Optimistic update uses incorrect attachment `id`** (line 129) — `getStatusAttachmentsFromPostBoxAttachments` sets `id: attachment.id` where `attachment.id` is the media ID, not the attachment UUID the database generates. The simplest fix is to use the full server response from `updateNote` for the `onPostUpdated` call instead of constructing the object locally (also flagged by Gemini at line 367).

**Should fix**

3. **Redundant `mediaId` in DB insert** (line 555 of `status.ts`) — `mediaId` is already included in the `data` object via `Attachment.parse()`; the explicit repetition in the `insert` call is unnecessary.

4. **`areAttachmentIdsEqual` is order-sensitive** (line 114) — worth documenting whether attachment order is intentionally significant (it matters if order is federated), or switching to a set comparison if it isn't.

**Minor**

5. **`hydrateDateFields` naming convention** (line 20 of `knexAdapter.ts`) — matching on the `*At` suffix is a heuristic that could silently convert non-date fields or miss date fields with different names. At minimum, validate with `!isNaN(new Date(value).getTime())` before assigning.

6. **Redundant `urlToId` call** (already flagged by Gemini at line 345) — `post-box.tsx` pre-encodes the status ID before passing it to `updateNote`, but `client.ts` now handles full-URL encoding internally. Pick one place for the encoding responsibility.

---

lib/components/post-box/post-box.tsx
```diff
@@ -171,16 +325,45 @@ export const PostBox: FC<Props> = ({
...
+          attachments: attachmentsChanged
+            ? getStatusAttachmentsFromPostBoxAttachments({
+                actorId: editStatus.actorId,
+                attachments,
+                statusId: editStatus.id,
+                timestamp: updateTime
+              })
+            : editStatus.attachments,
+          updatedAt: updateTime
         })
```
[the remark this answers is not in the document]
llun: Fixed in 02636198. PostBox now builds the local timeline update from the server response, including server media attachment ids and numeric timestamps. One intentional difference from the snippet: domain Status.text stays as source/plain text from the Mastodon text field or submitted source text, not rendered HTML content, because this codebase renders domain status text through processStatusText.

lib/components/post-box/post-box.tsx
```diff
@@ -171,16 +325,45 @@ export const PostBox: FC<Props> = ({
...
+          getEditableStatusAttachments(editStatus)
+        )
+        const updateMessage = message !== baselineText ? message : undefined
+        const updateContentWarning =
+          currentContentWarning !== baselineContentWarning
+            ? currentContentWarning
+            : undefined
+
         await updateNote({
           statusId: urlToId(editStatus.id),
```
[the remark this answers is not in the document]
llun: Fixed in 02636198. PostBox now passes the raw editStatus.id to updateNote and leaves URL encoding centralized in the client helper.

lib/components/post-box/post-box.tsx
llun: Kept the ordered comparison intentionally. The update API sends ordered media_ids, the database inserts replacements with order-preserving timestamps, and reads sort by createdAt/id, so a future reorder should count as a real attachment change. Renamed the helper to areAttachmentIdsEqualInOrder in 02636198 to make that behavior explicit.

lib/database/sql/status.ts
llun: Fixed in f748c3d9 by removing the redundant mediaId override from the insert object.

lib/components/post-box/post-box.tsx
```diff
@@ -133,6 +207,86 @@ export const PostBox: FC<Props> = ({
...
+        const loadingAttachment = {
+          ...attachment,
+          isLoading: true
+        }
+        postExtensionRef.current = {
+          ...postExtensionRef.current,
+          attachments: postExtensionRef.current.attachments.map((item) =>
+            item.id === attachment.id ? loadingAttachment : item
+          )
+        }
```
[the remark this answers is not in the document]
llun: Fixed in 02636198 by adding a synchronous submit lock around edit/create submission. I kept the reducer/ref synchronization because it is used to keep removed-during-upload filtering current, but reentrant submit is now guarded and covered by the PostBox test.

lib/services/auth/knexAdapter.ts
llun: Fixed in f748c3d9. hydrateDateFields now validates the constructed Date and leaves invalid *At values unchanged; added coverage for invalid createdAt with valid expireAt hydration.

lib/components/post-box/post-box.tsx
llun: Fixed in 02636198. The optimistic timeline update now uses server-returned media attachment ids for Attachment.id and pairs them with uploaded/existing media metadata for mediaId, so replacement attachments no longer reuse the media id as the attachment record id.

llun: Fixed in 02636198. The local timeline update now maps attachments from the server response, so Attachment.id uses the server media_attachments id while mediaId/metadata are preserved from uploaded or existing attachments.

lib/components/post-box/post-box.tsx
llun: Kept the ordered comparison intentionally. The update API sends ordered media_ids and the database persists replacement attachments in that order; reads also order by createdAt/id, so a future reorder should be considered a real edit. In 02636198 I renamed the helper to areAttachmentIdsEqualInOrder so the order-sensitive behavior is explicit.

llun:
@codex review

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +84,145 @@ interface Props {
   onDiscardEdit: () => void
 }
 
+const getEditableStatusText = (status: EditableStatus) =>
+  sanitizeHtml(status.text, { allowedTags: [] })
+
```
llun, line 91: **Regression: `sanitizeHtml` HTML-encodes plain text**

`sanitizeHtml(status.text, { allowedTags: [] })` does not return plain text unchanged — it HTML-encodes `&`, `<`, and `>` (e.g. `"a & b"` becomes `"a &amp; b"`). This is a regression from the pre-PR behaviour, where the edit box was populated directly with the raw `status.text` string.

The fix is to use `status.text` directly:

```ts
const getEditableStatusText = (status: EditableStatus) => status.text
```

If stripping HTML tags is still needed (e.g. when `status.text` can contain markup), use a DOM-based approach such as:

```ts
const getEditableStatusText = (status: EditableStatus) =>
  new DOMParser().parseFromString(status.text, 'text/html').body.textContent ?? ''
```

but that should only apply server-side if the stored text is actually HTML. Based on the database schema `status.text` stores the original plaintext, so a plain `status.text` is correct here.

---

llun (commented):
## Round 2 Review

All 7 issues from the round 1 review are resolved in these new commits. One item remains open.

### Fixed ✅

1. **`areAttachmentIdsEqual` order-sensitivity** — renamed to `areAttachmentIdsEqualInOrder` and the name now signals the order-sensitive contract.
2. **`getStatusAttachmentsFromPostBoxAttachments` using media ID as attachment ID** — replaced with `getStatusAttachmentsFromUpdateResponse`, which pairs attachments from the server response (correct IDs and dimensions).
3. **Redundant `urlToId` call before `updateNote`** — removed; `client.ts` now handles URL-style status IDs internally.
4. **Concurrent `Promise.all` ref mutations (race condition)** — eliminated by the new sequential attachment upload loop.
5. **`submitInFlightRef` guard** — added correctly; `finally` block resets the flag so a failed submission can be retried.
6. **`mediaId: attachment.id` duplicate in `trx('attachments').insert()`** — removed from `status.ts`; `Attachment.parse()` already includes it.
7. **`hydrateDateFields` invalid-date safety** — added `Number.isNaN(date.getTime())` guard in `knexAdapter.ts`.

### Still open ⚠️

**`getEditableStatusText` HTML-encodes plain text** (inline comment on line 89): `sanitizeHtml(status.text, { allowedTags: [] })` HTML-encodes `&`, `<`, `>`, so a post containing `"a & b"` will appear as `"a &amp; b"` in the edit box. This is a regression — the fix is simply `status.text`.

### Minor observations (non-blocking)

- **`parseTimestamp` / `getTimestamp` duplication**: `lib/client.ts` introduces `parseTimestamp` while `lib/utils/getTimestamp.ts` already exists with the same intent. Consider consolidating them.
- **`getMediaTypeFromMastodonAttachment` wildcard MIME types**: the function maps Mastodon `"image"` → `"image/*"` etc. A wildcard MIME type stored in the database may cause issues downstream (e.g. `<source type="image/*">` in a `<picture>` element is invalid). If the attachment's `url` extension is available, it is worth resolving to a concrete type.

---

lib/components/post-box/post-box.tsx
llun: Fixed in 74f3a01b. The edit box now uses status.text directly instead of sanitizeHtml, with test coverage for raw &, <, and > characters staying unescaped during initialization and submit.

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +84,145 @@ interface Props {
...
+const getEditableStatusAttachments = (
+  status: EditableStatus
+): PostBoxAttachment[] =>
+  status.attachments.map((attachment) => ({
+    type: 'upload',
+    id: attachment.mediaId ?? attachment.id,
```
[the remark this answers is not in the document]
llun: Fixed in 74f3a01b. getEditableStatusAttachments now only includes attachments with a real mediaId and skips fitness/non-editable attachments, so persisted attachment.id values are not sent as media_ids. Added regression coverage for a legacy attachment with mediaId: null.

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +84,145 @@ interface Props {
   onDiscardEdit: () => void
 }
 
+const getEditableStatusText = (status: EditableStatus) =>
+  sanitizeHtml(status.text, { allowedTags: [] })
```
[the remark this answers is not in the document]
llun: Fixed in 74f3a01b. getEditableStatusText now returns raw status.text, and the PostBox regression test covers raw &, <, and > characters in both textarea initialization and the submitted updateNote message.

llun:
Addressed the remaining item from https://github.com/llun/activities.next/pull/707#pullrequestreview-4304126207 in 74f3a01b. The edit box now uses raw status.text and has regression coverage for raw &, <, and > during initialization and submit.

I also addressed the wildcard MIME observation in 74f3a01b by using concrete fallback media types (image/jpeg, video/mp4, audio/mpeg) when reconciling unmatched Mastodon media attachments.

No code change for the parseTimestamp/getTimestamp observation: lib/utils/getTimestamp.ts does not exist in this worktree, and the two local helpers have different contracts. The client helper parses Mastodon API date strings, while the PostBox helper also accepts already-normalized numbers and Date values from local timeline state. Consolidating that would be unrelated cleanup.

llun:
@codex review

lib/components/post-box/post-box.tsx
```diff
@@ -325,12 +537,19 @@ export const PostBox: FC<Props> = ({
...
+      ...postExtension.attachments.slice(attachmentIndex + 1)
+    ]
+    const nextExtension = {
+      ...postExtensionRef.current,
+      attachments: nextAttachments
+    }
+    postExtensionRef.current = nextExtension
+    dispatch(setAttachments(nextAttachments))
+    if (editStatus) {
+      setAllowPost(isEditDirty(textRef.current, nextExtension))
```
llun: **New post: removing an attachment never updates `allowPost`**

When `editStatus` is absent (creating a new post), removing an attachment does not update `allowPost`. If the user selects an attachment (Post button enabled), then removes it, the button stays enabled even though the post is now empty.

The else branch should update the button based on whether there is any remaining content:

```ts
if (editStatus) {
  setAllowPost(isEditDirty(textRef.current, nextExtension))
} else {
  const hasContent =
    nextAttachments.length > 0 ||
    textRef.current.trim().length > 0 ||
    Boolean(postExtensionRef.current.fitnessFile)
  setAllowPost(hasContent)
}
```

---

lib/components/post-box/post-box.tsx
```diff
@@ -400,15 +619,15 @@ export const PostBox: FC<Props> = ({
     setText(value)
     textRef.current = value
     if (value.trim().length === 0) {
-      setAllowPost(Boolean(postExtensionRef.current.fitnessFile))
+      setAllowPost(
+        editStatus
+          ? isEditDirty(value)
+          : Boolean(postExtensionRef.current.fitnessFile)
```
llun: **New post: clearing text disables Post button even when attachments are present**

When the text area is emptied for a new post, `allowPost` is set based only on `fitnessFile` — it ignores `attachments`. So a user who has added media and then clears their caption will find the Post button disabled despite having a valid attachment.

```ts
: Boolean(postExtensionRef.current.fitnessFile) ||
    postExtensionRef.current.attachments.length > 0
```

---

lib/components/post-box/post-box.tsx
```diff
@@ -681,7 +914,22 @@ export const PostBox: FC<Props> = ({
...
+                    attachment
+                  ],
+                  fitnessFile: undefined,
+                  poll: {
+                    ...DEFAULT_STATE.poll
+                  }
+                }
+                postExtensionRef.current = nextExtension
                 dispatch(addAttachment(attachment))
+                if (editStatus) {
```
llun, line 1015: **New post: adding an attachment never enables the Post button**

For a new post (no `editStatus`), selecting an attachment does not call `setAllowPost(true)`, so the Post button remains disabled. This prevents users from submitting a media-only post with no text.

```ts
if (editStatus) {
  setAllowPost(isEditDirty(textRef.current, nextExtension))
} else {
  setAllowPost(true)
}
```

Note: this issue, the `onRemoveAttachment` gap, and the `onTextChange` gap are all three sides of the same bug — together they mean attachment-only new posts can't be submitted and the button state is wrong whenever the user adds/removes attachments from a text-empty new post.

---

llun (commented):
## Round 3 Review

All items from rounds 1 and 2 are resolved (including the `getEditableStatusText` regression and the non-media attachment IDs in `getEditableStatusAttachments`). One cluster of related bugs remains in the new-post path.

### Fixed ✅

- `getEditableStatusText` now returns `status.text` directly — `sanitizeHtml` removed.
- `getEditableStatusAttachments` skips fitness attachments and attachments without a `mediaId`, preventing non-media IDs from leaking into `media_ids`.
- `getMediaTypeFromMastodonAttachment` now returns concrete MIME types (e.g. `'image/jpeg'`) instead of wildcards.

### Still open ⚠️ — attachment button state for new posts (three inline comments)

Three related gaps in `post-box.tsx` mean attachment-only new posts can't be submitted:

1. **`onAddAttachment`** (line 930) — adding an attachment to a new post never calls `setAllowPost(true)`.
2. **`onRemoveAttachment`** (line 550) — removing an attachment from a new post never recomputes `allowPost`, leaving the button enabled after the last attachment is removed.
3. **`onTextChange`** (line 625) — clearing text on a new post checks `fitnessFile` but not `attachments.length`, disabling the button even when media is still present.

All three can be fixed together in one pass; see the inline comments for the minimal fix for each.

---

lib/components/post-box/post-box.tsx
```diff
@@ -325,12 +537,19 @@ export const PostBox: FC<Props> = ({
...
+    ]
+    const nextExtension = {
+      ...postExtensionRef.current,
+      attachments: nextAttachments
+    }
+    postExtensionRef.current = nextExtension
+    dispatch(setAttachments(nextAttachments))
+    if (editStatus) {
+      setAllowPost(isEditDirty(textRef.current, nextExtension))
+    }
```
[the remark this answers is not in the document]
llun: Fixed in d8fe49f0. `onRemoveAttachment` now recomputes `allowPost` for new posts with the shared `hasNewPostContent` helper, so removing the last media item disables Post while remaining text/media/fitness keeps it enabled. Added PostBox coverage for removing the only media attachment.

lib/components/post-box/post-box.tsx
llun: Fixed in d8fe49f0. `onRemoveAttachment` now recomputes `allowPost` for new posts with the shared `hasNewPostContent` helper, so removing the last media item disables Post while remaining text/media/fitness keeps it enabled. Added PostBox coverage for removing the only media attachment.

lib/components/post-box/post-box.tsx
```diff
@@ -400,15 +619,15 @@ export const PostBox: FC<Props> = ({
     setText(value)
     textRef.current = value
     if (value.trim().length === 0) {
-      setAllowPost(Boolean(postExtensionRef.current.fitnessFile))
+      setAllowPost(
+        editStatus
+          ? isEditDirty(value)
+          : Boolean(postExtensionRef.current.fitnessFile)
+      )
```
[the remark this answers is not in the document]
llun: Fixed in d8fe49f0. Empty new-post text now uses `hasNewPostContent`, which includes regular attachments as well as fitness files, so clearing a caption keeps Post enabled when media remains. Added regression coverage.

lib/components/post-box/post-box.tsx
llun: Fixed in d8fe49f0. Empty new-post text now uses `hasNewPostContent`, which includes regular attachments as well as fitness files, so clearing a caption keeps Post enabled when media remains. Added regression coverage.

lib/components/post-box/post-box.tsx
```diff
@@ -681,7 +914,22 @@ export const PostBox: FC<Props> = ({
...
+                  fitnessFile: undefined,
+                  poll: {
+                    ...DEFAULT_STATE.poll
+                  }
+                }
+                postExtensionRef.current = nextExtension
                 dispatch(addAttachment(attachment))
+                if (editStatus) {
+                  setAllowPost(isEditDirty(textRef.current, nextExtension))
+                }
```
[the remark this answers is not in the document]
llun: Fixed in d8fe49f0. Adding media to a new post now updates `allowPost` through `hasNewPostContent`, enabling media-only posts. Added a test that uploads and submits a media-only new post.

lib/components/post-box/post-box.tsx
llun: Fixed in d8fe49f0. Adding media to a new post now updates `allowPost` through `hasNewPostContent`, enabling media-only posts. Added a test that uploads and submits a media-only new post.

lib/components/post-box/post-box.tsx
```diff
@@ -171,16 +420,53 @@ export const PostBox: FC<Props> = ({
...
+          : ''
+        const attachmentsChanged = !areAttachmentIdsEqualInOrder(
+          attachments,
+          getEditableStatusAttachments(editStatus)
+        )
+        const updateMessage = message !== baselineText ? message : undefined
+        const updateContentWarning =
+          currentContentWarning !== baselineContentWarning
+            ? currentContentWarning
+            : undefined
```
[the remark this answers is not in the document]
llun: Fixed in d8fe49f0. PostBox now skips unchanged edit submissions before calling updateNote, but still sends an explicit empty message when text was actually cleared. Added regressions for blank-text media edits and no-op edit submits.

lib/client.ts
```diff
@@ -69,42 +70,79 @@ export interface UpdateNoteParams {
...
   statusId,
   message,
-  contentWarning
-}: UpdateNoteParams) => {
+  contentWarning,
+  attachments
+}: UpdateNoteParams): Promise<UpdateNoteResult> => {
   const normalizedMessage =
     message !== undefined && message.trim().length > 0 ? message : undefined
+  const hasAttachmentChanges = attachments !== undefined
```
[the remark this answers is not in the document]
llun: Fixed in d8fe49f0. updateNote now treats a defined message as an explicit status update, so clearing text serializes an empty status instead of omitting the field. Content-warning-only callers omit message and continue to send only spoiler_text. Added client tests for both paths.

llun:
Addressed the Round 3 review in d8fe49f0. New-post button state now uses one shared content predicate across media add, media remove, and text-clear paths. Blank edit text is also handled explicitly: `updateNote` sends `status: ""` when the user clears text, while PostBox skips true no-op edit submissions before calling the client helper.

Verification for this follow-up:
- `yarn test lib/client.test.ts lib/components/post-box/post-box.test.tsx --runInBand`
- `yarn build`
- `yarn test` (298 suites, 2391 tests passed after rerunning a transient Jest worker SIGSEGV)

llun:
@codex review

lib/components/post-box/post-box.tsx
```diff
@@ -85,6 +87,159 @@ interface Props {
...
+  updatedAt
+}: {
+  actorId: string
+  existingAttachments: Attachment[]
+  mediaAttachments: UpdateNoteMediaAttachment[]
+  statusId: string
+  uploadedAttachments: PostBoxAttachment[]
+  updatedAt: number
+}): Attachment[] =>
+  mediaAttachments.map((mediaAttachment, index) => {
```
llun: **Fitness-file attachments are silently dropped from local state after any edit**

`getStatusAttachmentsFromUpdateResponse` returns `mediaAttachments.map(...)`, which only covers Mastodon media attachments. Fitness-file attachments are stored alongside regular media in `editStatus.attachments` but are never included in the Mastodon `media_attachments` response, so they disappear from local timeline state after a successful edit.

The `existingAttachments` parameter is passed in (and used for dimension/mediaType lookup) but never queried for fitness entries. A minimal fix is to append them from the existing list:

```ts
const fitnessAttachments = existingAttachments.filter(isFitnessAttachment)
return [
  ...mediaAttachments.map((mediaAttachment, index) => {
    // ... existing mapping ...
  }),
  ...fitnessAttachments
]
```

The `isFitnessAttachment` import is already present in this file at line 40.

---

lib/components/post-box/post-box.tsx
```diff
@@ -312,6 +540,8 @@ export const PostBox: FC<Props> = ({
       setIsPosting(false)
       setAllowPost(true)
       alert('Fail to create a post')
```
llun: **Wrong error message when an edit fails**

This `catch` block is shared by both the create and edit paths. When the edit path throws (e.g. `uploadMediaAttachments` or `updateNote` rejects), the alert says "Fail to create a post". Should be:

```ts
alert(editStatus ? 'Fail to update the post' : 'Fail to create a post')
```

---

lib/components/post-box/post-box.tsx
```diff
@@ -400,15 +640,15 @@ export const PostBox: FC<Props> = ({
     setText(value)
     textRef.current = value
     if (value.trim().length === 0) {
-      setAllowPost(Boolean(postExtensionRef.current.fitnessFile))
+      setAllowPost(
+        editStatus
+          ? isEditDirty(value)
```
llun: **Edit path can enable Update button for a post that would be empty**

`isEditDirty(value)` returns `true` whenever the current state differs from the baseline, but it doesn't verify that the resulting post has any content. If a user removes all attachments and then clears the text, `isEditDirty` returns `true` (text changed), so `allowPost` is `true` — but submitting would hit the server's 422 empty-post guard and fail.

Combining the dirty check with `hasNewPostContent` (already available in scope) fixes this:

```ts
editStatus
  ? isEditDirty(value) && hasNewPostContent(value, postExtensionRef.current)
  : hasNewPostContent(value, postExtensionRef.current)
```

---

llun (commented):
## Round 4 Review

All 3 Post-button state bugs from round 3 are fixed via the new `hasNewPostContent` helper, and the no-op edit guard and `client.ts` empty-text serialisation are both resolved. Three items remain open.

### Fixed ✅

- `onAddAttachment` now calls `hasNewPostContent` for new posts — adding media enables Post.
- `onRemoveAttachment` now calls `hasNewPostContent` for new posts — removing the last item disables Post.
- `onTextChange` now uses `hasNewPostContent` for new posts — clearing a caption doesn't disable Post when media is still present.
- No-op edit guard prevents a submit when nothing changed.
- `client.ts` now serialises a defined `message` even when it is `''`, so clearing text on an edit is correctly sent to the server.

### Still open ⚠️

1. **Fitness-file attachments lost after edit** (inline comment on line 201, confirmed from the Codex thread): `getStatusAttachmentsFromUpdateResponse` returns only `mediaAttachments.map(...)`, dropping any fitness-file attachments that live alongside regular media on `editStatus.attachments`. The fix is to append `existingAttachments.filter(isFitnessAttachment)` to the returned array.

2. **Wrong error alert on edit failure** (inline comment on line 542, confirmed from the Gemini thread): the shared `catch` block shows "Fail to create a post" even when the edit path throws. Should be `editStatus ? 'Fail to update the post' : 'Fail to create a post'`.

3. **Update button can be enabled for an empty post** (inline comment on line 645, confirmed from the Gemini thread): `isEditDirty(value)` alone allows the button when the user has removed all content (text and attachments cleared). Should also require `hasNewPostContent(value, postExtensionRef.current)`.

### Minor / non-blocking

- **`encodeBase64Url` in `urlToId.ts`** (Gemini thread): Preferring `btoa` over `Buffer` in Node 24 goes through the slower `Array.from` + `join` path rather than native `Buffer.from(...).toString('base64url')`. Since `Buffer` is available everywhere in this Next.js app (Node on server, webpack polyfill on client) and Node 24 natively supports the `'base64url'` encoding, checking `typeof Buffer !== 'undefined'` first would be strictly more efficient. Correctness is unaffected.

---

lib/utils/urlToId.ts
```diff
@@ -1,28 +1,33 @@
...
 
-  const paddedValue = value.padEnd(
-    value.length + ((4 - (value.length % 4)) % 4),
-    '='
-  )
-  const binary = atob(paddedValue.replaceAll('-', '+').replaceAll('_', '/'))
+  const binary = atob(base64)
   const bytes = Uint8Array.from(binary, (char) => char.charCodeAt(0))
   return new TextDecoder().decode(bytes)
 }
```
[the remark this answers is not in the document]
llun: Fixed in b437a21d. The URL helpers now prefer Node Buffer base64url support and keep the browser btoa/atob path as a fallback; I also added a regression that makes the browser fallbacks throw under Node so the Buffer path is covered.

lib/components/post-box/post-box.tsx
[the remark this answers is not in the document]
llun: Fixed in b437a21d. The shared catch now uses an edit-specific alert when updateNote or edit media upload fails.

lib/components/post-box/post-box.tsx
```diff
@@ -400,15 +640,15 @@ export const PostBox: FC<Props> = ({
...
+          : hasNewPostContent(value, postExtensionRef.current)
+      )
```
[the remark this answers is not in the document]
llun: Fixed in b437a21d. Edit submit enablement now requires both dirty state and resulting content, including preserved non-editable attachments, and the submit handler guards invalid disabled submissions. Added a regression for clearing both text and media.

lib/components/post-box/post-box.tsx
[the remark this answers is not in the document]
llun: Fixed in b437a21d. The updated status reconciliation now preserves non-editable attachments such as fitness-file attachments, while avoiding duplicates for preserved attachments already represented by the API response. Added focused regressions for both cases.

llun: Fixed in b437a21d. Fitness-file attachments are appended from the existing status attachments during local reconciliation, and returned API media are de-duped by id/url so legacy media do not duplicate.

lib/components/post-box/post-box.tsx
llun: Fixed in b437a21d. The edit path now reports the update-specific failure message from the shared catch block.

lib/components/post-box/post-box.tsx
llun: Fixed in b437a21d. The edit button and submit handler now require dirty plus non-empty edited content, with a regression test for clearing all editable content.

llun:
@codex review

lib/client.ts
```diff
@@ -69,42 +70,78 @@ export interface UpdateNoteParams {
...
+      ...(hasMessageChange ? { status: message } : {}),
+      ...(contentWarning !== undefined ? { spoiler_text: contentWarning } : {}),
+      ...(attachments !== undefined
+        ? { media_ids: attachments.map((attachment) => attachment.id) }
+        : {})
     })
   })
   if (response.status !== 200) {
     throw new Error('Fail to create a new note')
   }
```
llun, line 129: **Copy-paste error: wrong error message in `updateNote`**

`throw new Error('Fail to create a new note')` was copied from `createNote` and never updated. Since this function is `updateNote`, it should be:

```ts
throw new Error('Fail to update the note')
```

---

lib/database/sql/status.ts
```diff
@@ -518,6 +520,44 @@ export const StatusSQLDatabaseMixin = (
           }),
           updatedAt: currentTime
         })
+
+      if (attachments !== undefined) {
+        const replaceableAttachmentIds = status.attachments
+          .filter((attachment) => !isFitnessAttachment(attachment))
```
llun: **P1: Legacy attachments (null `mediaId`) are silently deleted on any media edit**

The delete filter is `!isFitnessAttachment(attachment)`, which targets all non-fitness attachments — including legacy or remotely-federated ones that have `mediaId: null`. Because `getEditableStatusAttachments` (and therefore the `media_ids` the client sends) only includes attachments *with* a real `mediaId`, those legacy attachments are never reinserted. They are permanently deleted from the database whenever any media edit happens.

The fix is to scope the delete to only the attachments the client can actually manage — those with a non-null `mediaId`:

```ts
const replaceableAttachmentIds = status.attachments
  .filter(
    (attachment) =>
      !isFitnessAttachment(attachment) && attachment.mediaId !== null
  )
  .map((attachment) => attachment.id)
```

This preserves legacy/null-mediaId attachments through any edit, matching the behaviour of `getEditableStatusAttachments` on the client side.

---

llun (commented):
## Round 5 Review

All 4 items from round 4 are resolved (fitness attachments preserved, alert message contextualised, Update button validity gate, `encodeBase64Url` efficiency). Two new issues remain.

### Fixed ✅

- Fitness-file attachments are now appended from existing state in `getStatusAttachmentsFromUpdateResponse` and de-duped.
- Error alert is now `editStatus ? 'Fail to update the post' : 'Fail to create a post'`.
- `isEditDirty(value) && hasNewPostContent(...)` prevents enabling Update on an empty result.
- `encodeBase64Url` / `decodeBase64Url` now prefer `Buffer.from(...).toString('base64url')` in Node.

### Still open ⚠️

1. **`updateNote` in `client.ts` line 129** (inline comment, confirmed Gemini thread): the error thrown on non-200 is `'Fail to create a new note'` — a copy-paste from `createNote`. Should be `'Fail to update the note'`.

2. **Legacy attachments silently deleted on media edit — `status.ts` line 526** (inline comment, confirmed Codex P1 thread): `replaceableAttachmentIds` filters on `!isFitnessAttachment` but does not exclude null-`mediaId` attachments. Any legacy or federated attachment that lacks a `mediaId` gets permanently deleted from the database whenever the user makes any media change, because the client never sends those IDs in `media_ids`. Fix: add `&& attachment.mediaId !== null` to the filter.

### Minor / non-blocking

- **Generic alert swallows specific upload error** (Gemini thread at line 595): `uploadMediaAttachments` throws `Fail to upload ${attachment.name}` but the outer `catch {}` (no binding) discards the message and shows a generic alert. Changing to `catch (error)` and using `error instanceof Error ? error.message : ...` would surface the file-specific reason to the user.

- **Attachment ID and `createdAt` churn on edit** (Gemini thread at line 560): the delete-and-reinsert strategy assigns new UUIDs and resets `createdAt` even for attachments that were unchanged. This is a DB-level inefficiency rather than an API-surface bug (the Mastodon response correctly identifies attachments by `mediaId`), but it prevents stable attachment record IDs across edits and loses the original `createdAt`. A match-and-upsert approach would be more robust if attachment record stability ever matters.

---

lib/client.ts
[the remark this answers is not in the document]
llun: Fixed in fc18331d. updateNote now throws the update-specific error message, and there is a regression test for non-200 update responses.

lib/components/post-box/post-box.tsx
```diff
@@ -311,7 +592,9 @@ export const PostBox: FC<Props> = ({
     } catch {
       setIsPosting(false)
       setAllowPost(true)
-      alert('Fail to create a post')
+      alert(editStatus ? 'Fail to update the post' : 'Fail to create a post')
```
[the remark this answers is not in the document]
llun: Fixed in fc18331d. PostBox now alerts the caught Error message when available, and upload failures include the media filename plus the underlying reason. Added regressions for update failure and upload failure.

lib/database/sql/status.ts
```diff
@@ -518,6 +520,44 @@ export const StatusSQLDatabaseMixin = (
...
+            })
+
+            return trx('attachments').insert({
+              ...data,
+              createdAt: attachmentCreatedAt,
+              updatedAt: attachmentCreatedAt
+            })
+          })
+        )
+      }
```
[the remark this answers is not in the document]
llun: Fixed in fc18331d. updateNote now diffs media IDs instead of blanket delete/reinsert: unchanged media rows keep their id and timestamps, removed editable media is deleted with status/actor/id scoping, and only new media rows are inserted.

lib/database/sql/status.ts
```diff
@@ -518,6 +520,44 @@ export const StatusSQLDatabaseMixin = (
           }),
           updatedAt: currentTime
         })
+
+      if (attachments !== undefined) {
+        const replaceableAttachmentIds = status.attachments
+          .filter((attachment) => !isFitnessAttachment(attachment))
+          .map((attachment) => attachment.id)
```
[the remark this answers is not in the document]
llun: Fixed in fc18331d. Legacy/null-mediaId attachments are now preserved during media edits, while only editable media attachments with non-null mediaId participate in replacement. Added DB and API regression coverage.

lib/client.ts
llun: Fixed in fc18331d. updateNote now throws Fail to update the note, with a client regression test covering non-200 update responses.

lib/database/sql/status.ts
llun: Fixed in fc18331d. The replacement logic now targets only editable non-fitness attachments with a non-null mediaId, preserving legacy/null-mediaId attachments. Added regression coverage for replace and clear flows.

llun:
@codex review

llun (commented):
## Round 6 Review

All 4 items from round 5 are resolved. **No new issues found.** This PR looks ready to merge.

### Fixed ✅

- `updateNote` in `client.ts` now throws `'Fail to update the note'` (copy-paste corrected, with regression test).
- The `catch` block in `onPost` now binds `error` and surfaces the specific `error.message` (e.g. `"Fail to upload foo.jpg: ..."`) before falling back to the generic create/update message.
- `status.ts` `updateNote` now uses a diff-based approach (`isReplaceableMediaAttachment`, delete-removed-only, insert-new-only) with `actorId` scoping on the delete. Preserved attachment rows keep their original IDs and `createdAt` timestamps.
- Legacy attachments with null `mediaId` are preserved through any edit (excluded from the replaceable set by `isReplaceableMediaAttachment`).

### New code quality checks

The `isEditSubmittable` / `hasEditPostContent` / `getPreservedStatusAttachments` refactor is well-structured. `hasEditPostContent` correctly counts preserved (fitness + null-mediaId) attachments as post content, so the Update button stays enabled for posts whose only remaining content is a fitness file. All `setAllowPost` paths consistently use `isEditSubmittable` for edits and `hasNewPostContent` for new posts.

### One minor non-blocking observation

When a user reorders _existing_ (preserved) attachments alongside new uploads, the new attachment rows get `createdAt = currentTime + index` while preserved rows keep their original (older) timestamps. Since reads sort by `createdAt/id`, preserved attachments will always sort before new ones regardless of the submitted `media_ids` order. The `areAttachmentIdsEqualInOrder` dirty check will correctly detect a reorder as a change, but the server-persisted order will not reflect it. This is a pre-existing display-order limitation rather than a regression introduced here.

---

PR #710, opened 2026-05-17
none: fix version bump auto-merge race

llun:
## Summary

- Wait for required PR checks to be registered before invoking `gh pr checks --required --watch`.
- Keep the existing guard that leaves the bump PR open if `main` moves while checks are running.

## Root Cause

The version-bump workflow created the PR and immediately called `gh pr checks --required --watch`. When GitHub had not attached any check runs yet, `gh` returned `no checks reported` and the script exited before enabling auto-merge.

## Validation

- `ruby -ryaml -e 'workflow = YAML.load_file(".github/workflows/version-bump.yml"); abort("missing jobs") unless workflow["jobs"]; puts "yaml ok"'`
- `ruby -ryaml -e 'workflow = YAML.load_file(".github/workflows/version-bump.yml"); step = workflow["jobs"]["version-bump"]["steps"].find { |item| item["name"] == "Create or update version bump PR" }; print step.fetch("run")' | bash -n`
- `git diff --check`
- `yarn lint` (exit 0, existing warnings only)
- `yarn build`
- `yarn test`

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -11,6 +11,12 @@ permissions:
   pull-requests: write
 
 env:
+  # Required checks usually register within seconds; cap this wait at 5 minutes
+  # so auto-merge is not skipped just because checks were attached slowly.
+  CHECK_REGISTRATION_ATTEMPTS: 30
+  CHECK_REGISTRATION_INTERVAL_SECONDS: 10
+  CHECK_REGISTRATION_STABLE_READS: 2
+  CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT: 3
   DEBOUNCE_SECONDS: 300
   VERSION_BUMP_BRANCH: version-bump/main
 
@@ -338,6 +344,187 @@ jobs:
               --base main \
               --head "$VERSION_BUMP_BRANCH")
             echo "Created PR: $PR_URL"
+          fi
+
+          join_check_names() {
+            local check_names="$1"
+
+            if [ -z "$check_names" ]; then
+              echo "(none)"
+              return 0
+            fi
+
+            printf '%s\n' "$check_names" | paste -sd ',' - | sed 's/,/, /g'
+          }
+
+          read_branch_protection_required_checks() {
+            local checks_error="$1"
+            local gh_required_checks_exit
+            local required_check_names
+
+            : > "$checks_error"
+
+            set +e
+            # This endpoint only returns the branch-protection subset.
+            # Required workflows outside branch protection are still covered by
+            # the later gh pr checks --required --watch gate.
+            #
+            # .checks[] is the modern object array, while .contexts[] is the
+            # legacy string array. Union them, then keep only non-empty strings
+            # so either API shape produces the same required-check name set.
+            required_check_names=$(gh api "repos/${GITHUB_REPOSITORY}/branches/main/protection/required_status_checks" \
+              --jq '[.checks[]?.context, .contexts[]?] | map(select(type == "string" and length > 0)) | unique | .[]' 2> "$checks_error")
+            gh_required_checks_exit=$?
+            set -e
+
+            if [ "$gh_required_checks_exit" -eq 0 ]; then
+              printf '%s\n' "$required_check_names"
+              return 0
+            fi
+
+            cat "$checks_error" >&2
+            echo "::warning::Unable to read branch-protection required status checks." >&2
+
+            return 1
+          }
+
+          required_checks_not_reported_yet() {
+            local checks_error="$1"
+
+            # Some gh versions return a generic error, not exit 8, before
+            # required checks have registered. Treat only that startup state as
+            # pending; other errors still use the transient-error budget.
+            # Keep this fallback broad enough for minor gh wording changes; if
+            # it misses, the safe failure mode is leaving the PR open.
+            grep -Eiq '(no|0)[[:space:]]+(required[[:space:]]+)?(status[[:space:]]+)?checks?([[:space:]]+(reported|found|attached))?' "$checks_error"
+          }
+
+          wait_for_required_checks() {
+            local pr_url="$1"
+            local branch_protection_check_names="$2"
+            local checks_error="$3"
+            local attempt
+            local check_names
+            local consecutive_errors=0
+            local gh_checks_exit
+            local missing_checks=""
+            local previous_check_names=""
+            local required_check
+            local stable_reads=0
+
+            for attempt in $(seq 1 "$CHECK_REGISTRATION_ATTEMPTS"); do
+              : > "$checks_error"
+
+              set +e
+              check_names=$(gh pr checks "$pr_url" --required --json name,state --jq 'map(.name) | unique | .[]' 2> "$checks_error")
+              gh_checks_exit=$?
+              set -e
+
+              case "$gh_checks_exit" in
+                0 | 8)
+                  consecutive_errors=0
+                  missing_checks=""
+
+                  if [ -n "$branch_protection_check_names" ]; then
+                    while IFS= read -r required_check; do
+                      if [ -n "$required_check" ] && ! printf '%s\n' "$check_names" | grep -Fxq "$required_check"; then
+                        if [ -z "$missing_checks" ]; then
+                          missing_checks="$required_check"
+                        else
+                          missing_checks="${missing_checks}, ${required_check}"
+                        fi
+                      fi
+                    done <<< "$branch_protection_check_names"
+                  fi
+
+                  if [ -n "$check_names" ] && [ "$check_names" = "$previous_check_names" ]; then
+                    stable_reads=$((stable_reads + 1))
+                  elif [ -n "$check_names" ]; then
+                    stable_reads=1
+                    previous_check_names="$check_names"
+                  else
+                    stable_reads=0
+                    previous_check_names=""
+                  fi
+
+                  if [ -n "$check_names" ] && [ -z "$missing_checks" ] && [ "$stable_reads" -ge "$CHECK_REGISTRATION_STABLE_READS" ]; then
+                    echo "Required PR checks registered: $(join_check_names "$check_names")."
+                    return 0
+                  fi
+
+                  # gh exits 8 for pending checks, and some states produce either
+                  # exit 8 or exit 0 with no JSON rows. The JSON names decide
+                  # whether all required checks are registered yet.
+                  if [ -n "$missing_checks" ]; then
+                    echo "Required PR checks are not fully registered yet; missing branch-protection checks: ${missing_checks}."
+                  elif [ -z "$check_names" ]; then
+                    echo "Required PR checks are not reported yet."
+                  else
+                    echo "Required PR checks are settling; observed $(join_check_names "$check_names") for ${stable_reads}/${CHECK_REGISTRATION_STABLE_READS} stable reads."
+                  fi
+                  ;;
+                *)
+                  if required_checks_not_reported_yet "$checks_error"; then
+                    consecutive_errors=0
+                    # Preserve prior stable observations here too. Before any
+                    # checks appear these values are already empty/zero; after
+                    # checks appear, a no-checks response is another metadata
+                    # blip and should not discard the observed set.
+                    echo "Required PR checks are not reported yet."
+                  else
+                    consecutive_errors=$((consecutive_errors + 1))
+                    cat "$checks_error" >&2
+                    echo "::warning::Unable to read required PR checks (${consecutive_errors}/${CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT})."
+
+                    if [ "$consecutive_errors" -ge "$CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT" ]; then
+                      echo "::warning::Exceeded transient-error budget while reading required PR checks."
+                      return 1
+                    fi
+
+                    # Preserve stable_reads and previous_check_names across
+                    # transient errors; an API blip does not change the PR's
+                    # underlying required-check registration state.
+                  fi
+                  ;;
+              esac
+
+              if [ "$attempt" -lt "$CHECK_REGISTRATION_ATTEMPTS" ]; then
+                echo "Retrying required-check registration in ${CHECK_REGISTRATION_INTERVAL_SECONDS}s (${attempt}/${CHECK_REGISTRATION_ATTEMPTS})."
+                sleep "$CHECK_REGISTRATION_INTERVAL_SECONDS"
+              fi
+            done
+
+            if [ -n "$missing_checks" ]; then
+              echo "::warning::Last missing branch-protection checks: ${missing_checks}."
+            fi
+            echo "::warning::Last observed required PR checks: $(join_check_names "$previous_check_names")."
+            return 1
+          }
+
+          CHECKS_ERROR=$(mktemp)
+          trap 'rm -f "$CHECKS_ERROR"' EXIT
+
+          set +e
+          BRANCH_PROTECTION_CHECK_NAMES=$(read_branch_protection_required_checks "$CHECKS_ERROR")
+          BRANCH_PROTECTION_CHECKS_EXIT=$?
+          set -e
+
+          if [ "$BRANCH_PROTECTION_CHECKS_EXIT" -eq 0 ] && [ -n "$BRANCH_PROTECTION_CHECK_NAMES" ]; then
+            echo "Branch-protection required checks: $(join_check_names "$BRANCH_PROTECTION_CHECK_NAMES")."
+          elif [ "$BRANCH_PROTECTION_CHECKS_EXIT" -eq 0 ]; then
+            echo "No branch-protection required status checks are configured on main; using observed required PR checks as the registration gate."
+            BRANCH_PROTECTION_CHECK_NAMES=""
+          else
+            echo "::warning::Could not read branch-protection required checks; using observed required PR checks as the registration gate."
+            # Fallback mode cannot compare against an anchored expected set.
+            # It trusts the observed --required names after stable reads, and
+            # gh pr checks --required --watch remains the final merge gate.
+            BRANCH_PROTECTION_CHECK_NAMES=""
+          fi
+
+          if ! wait_for_required_checks "$PR_URL" "$BRANCH_PROTECTION_CHECK_NAMES" "$CHECKS_ERROR"; then
+            echo "::warning::Required PR checks were not fully reported after waiting; leaving $PR_URL open."
+            exit 0
           fi
 
           if ! gh pr checks "$PR_URL" --required --watch --interval 10; then


.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
             echo "Created PR: $PR_URL"
           fi
 
+          CHECKS_READY=false
+          for attempt in $(seq 1 "$CHECK_REGISTRATION_ATTEMPTS"); do
+            CHECKS_ERROR=$(mktemp)
+
+            if CHECK_COUNT=$(gh pr checks "$PR_URL" --required --json name,state --jq 'length' 2> "$CHECKS_ERROR"); then
+              if [ "${CHECK_COUNT:-0}" -gt 0 ]; then
```
llun: **Race condition is not fully eliminated — only narrowed.**

Breaking the loop as soon as `CHECK_COUNT > 0` means we proceed once *any* required check has been registered, not once *all* of them have been. When several workflows feed into the required checks list (e.g. multiple required workflows triggered by `pull_request`), GitHub often registers them in waves over several seconds. If the first check appears, lands in `success` quickly, and the others haven't been registered yet, `gh pr checks --required --watch` (line 375) can return success and `--auto` merge will fire while later-arriving required checks are still pending.

Two safer options:

1. Wait until the count has been stable and `> 0` for at least one extra poll interval (a simple "two consecutive equal non-zero reads" check).
2. Compare against an expected set of required check names from branch protection (`gh api repos/{owner}/{repo}/branches/main/protection/required_status_checks`) and only proceed once each expected name appears in the JSON.

Option 1 is the cheaper fix and is probably sufficient given the 10s interval.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
...
+                CHECKS_READY=true
+                rm -f "$CHECKS_ERROR"
+                break
+              fi
+            elif ! grep -qi "no checks reported" "$CHECKS_ERROR"; then
```
llun: **Fragile dependency on `gh`'s stderr text.**

Matching `"no checks reported"` couples this workflow to an undocumented stderr string in the `gh` CLI. If a future `gh` release rewords this (e.g. `"no checks found"`, `"no required checks reported"`), the `elif` branch will fire on the legitimate "checks not registered yet" case and the workflow will bail out with `::warning::Unable to read required PR checks` instead of retrying — silently breaking auto-merge.

`gh pr checks` exits with code `8` specifically for "no checks reported". Capturing `$?` and branching on `8` vs. other non-zero codes would be far more robust:

```bash
set +e
CHECK_COUNT=$(gh pr checks "$PR_URL" --required --json name,state --jq 'length' 2> "$CHECKS_ERROR")
gh_exit=$?
set -e

case "$gh_exit" in
  0) [ "${CHECK_COUNT:-0}" -gt 0 ] && { CHECKS_READY=true; break; } ;;
  8) : ;;  # no checks reported yet — keep polling
  *) cat "$CHECKS_ERROR" >&2; exit 0 ;;
esac
```

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
...
+              cat "$CHECKS_ERROR" >&2
+              rm -f "$CHECKS_ERROR"
+              echo "::warning::Unable to read required PR checks; leaving $PR_URL open."
+              exit 0
```
llun: **No retry on transient API failures.**

A single network blip, GitHub 5xx, or token-refresh hiccup during the very first poll will hit this branch and abort the auto-merge flow immediately, even though the next attempt would almost certainly succeed. The whole point of the surrounding retry loop is to tolerate transient unavailability — it's worth letting non-`"no checks reported"` errors retry a few times too (e.g. tolerate up to N consecutive failures before giving up) instead of one-strike-out.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
             echo "Created PR: $PR_URL"
           fi
 
+          CHECKS_READY=false
+          for attempt in $(seq 1 "$CHECK_REGISTRATION_ATTEMPTS"); do
+            CHECKS_ERROR=$(mktemp)
+
+            if CHECK_COUNT=$(gh pr checks "$PR_URL" --required --json name,state --jq 'length' 2> "$CHECKS_ERROR"); then
```
llun: When `--json` is combined with `--required`, `gh pr checks` does not always emit `"no checks reported"` to stderr — depending on the `gh` version it may instead exit `0` with an empty array `[]` (in which case `--jq 'length'` yields `0` and we fall through to the retry block correctly). The code handles both paths, but it's worth pinning the `gh` version or at least adding a comment noting that two divergent behaviors are intentionally tolerated here, so a future maintainer doesn't "simplify" one of them away.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
             echo "Created PR: $PR_URL"
           fi
 
+          CHECKS_READY=false
+          for attempt in $(seq 1 "$CHECK_REGISTRATION_ATTEMPTS"); do
+            CHECKS_ERROR=$(mktemp)
```
llun: `mktemp` is called inside the loop on every iteration. Cleanup happens on all the success/known-error paths, but if someone later adds an early `return`/`exit` inside the loop body without remembering to `rm -f "$CHECKS_ERROR"`, the tempfile will leak. Two simple alternatives:

- Allocate one tempfile outside the loop and `: > "$CHECKS_ERROR"` to truncate at the top of each iteration; register a `trap 'rm -f "$CHECKS_ERROR"' EXIT` once.
- Drop the file entirely and capture stderr into a variable: `CHECKS_ERROR=$( { CHECK_COUNT=$(gh pr checks ... --jq 'length'); } 2>&1 1>/dev/null )` (a bit fiddly with `set -e`, but avoids any filesystem state).

Minor — current code is correct, just trap-prone for future edits.

---

.github/workflows/version-bump.yml
```diff
@@ -11,6 +11,8 @@ permissions:
   pull-requests: write
 
 env:
+  CHECK_REGISTRATION_ATTEMPTS: 30
+  CHECK_REGISTRATION_INTERVAL_SECONDS: 10
```
llun, line 17: With `CHECK_REGISTRATION_ATTEMPTS=30` and `CHECK_REGISTRATION_INTERVAL_SECONDS=10`, the polling block can spend up to 5 minutes before giving up — on top of the existing `DEBOUNCE_SECONDS=300` debounce. A short comment here explaining the worst-case 5-minute wait (and why that's an acceptable ceiling vs. the typical "checks registered within a few seconds" case) would make the tuning intent obvious to future maintainers.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +342,36 @@ jobs:
...
+              exit 0
+            fi
+
+            rm -f "$CHECKS_ERROR"
+
+            if [ "$attempt" -lt "$CHECK_REGISTRATION_ATTEMPTS" ]; then
+              echo "Required PR checks are not reported yet; retrying in ${CHECK_REGISTRATION_INTERVAL_SECONDS}s (${attempt}/${CHECK_REGISTRATION_ATTEMPTS})."
+              sleep "$CHECK_REGISTRATION_INTERVAL_SECONDS"
+            fi
+          done
```
llun: **The polling block is essentially a private helper but inlined into the bigger script.**

The "Create or update version bump PR" step is now ~150 lines of bash doing several distinct things (fetch, branch, push, close stale, create/update, wait for checks, watch, race-guard, merge). The new wait-for-checks block is independently testable and would read more naturally as a shell function (or even a separate composite action) — e.g. `wait_for_required_checks "$PR_URL"`. Optional cleanup, but worth considering before this step grows further.

---

llun (commented):
The fix is a real improvement over the original "fire `--watch` immediately" behavior, and the structure (polling loop → guard → existing `--watch`) is sound. A few things worth tightening before this lands:

**Correctness**

- The race isn't fully closed: breaking on the *first* registered required check is racy when multiple required workflows register in waves. `--watch` can then succeed before later-arriving required checks even appear. Waiting for two consecutive equal non-zero reads, or comparing against the expected set from branch protection, would actually close the race rather than narrow it (see inline on line 350).
- The `grep -qi "no checks reported"` stderr match is brittle — `gh` can change that string in any release and silently break the retry semantics. Branching on `gh`'s exit code (`8` is "no checks reported") is the stable contract (line 355).
- Any transient API error on the first poll aborts the whole flow with no retry, which defeats the purpose of the surrounding loop (line 359).

**Maintainability (non-blocking)**

- The wait loop tolerates two different `gh` behaviors for "no required checks reported yet" (success with `[]` vs. exit-non-zero with stderr text). That's intentional, but worth a comment so someone doesn't "simplify" one branch away (line 349).
- The tempfile pattern is per-iteration; trap-based cleanup is more robust to future edits (line 347).
- This step is now ~150 lines doing many things; the wait-for-checks block would read more naturally as a shell function or composite action (line 368).
- A short comment on the new env vars explaining the 5-minute worst-case wait would help future tuning (line 15).

Net: I'd land this after the two correctness items (race window + exit-code-based detection) are addressed; the rest are quality-of-life.

---

.github/workflows/version-bump.yml
llun: Fixed. The registration gate now reads required check names from main branch protection and only proceeds once every expected required check name is present on the PR.

.github/workflows/version-bump.yml
llun: Fixed. The probe no longer matches stderr text; it captures gh exit status and treats non-fatal check-registration states separately from other errors.

.github/workflows/version-bump.yml
llun: Fixed. Non-registration gh/API failures are now retried up to CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT consecutive times before leaving the PR open.

.github/workflows/version-bump.yml
llun: Fixed. Added an inline comment documenting that gh may return exit 0 with no rows or exit 8 while checks are not ready or pending, and the JSON names determine readiness.

.github/workflows/version-bump.yml
llun: Fixed. The workflow now creates one temp file, truncates it on each probe, and cleans it with an EXIT trap.

.github/workflows/version-bump.yml
llun: Fixed. Added a comment explaining the 5-minute registration cap and why it exists.

.github/workflows/version-bump.yml
llun: Fixed. Extracted the registration polling into wait_for_required_checks.

.github/workflows/version-bump.yml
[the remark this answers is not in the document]
llun: Fixed. The probe now treats gh exit 8 as non-fatal and evaluates any emitted JSON check names before deciding whether to retry.

.github/workflows/version-bump.yml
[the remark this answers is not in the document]
llun: Fixed. Exit 8 no longer goes through the error path; it uses the JSON check-name set and returns ready once all branch-protected checks are present.

llun:
Addressed the review feedback: the workflow now waits for the branch-protected required check set, handles gh check-registration and pending exit codes without stderr matching, retries transient API failures, uses trap-based temp-file cleanup, documents the 5-minute registration cap, and keeps the polling logic in a helper function.

.github/workflows/version-bump.yml
```diff
@@ -340,6 +345,95 @@ jobs:
...
+            return 1
+          }
+
+          CHECKS_ERROR=$(mktemp)
+          trap 'rm -f "$CHECKS_ERROR"' EXIT
+
+          : > "$CHECKS_ERROR"
+          set +e
+          REQUIRED_CHECK_NAMES=$(gh api "repos/${GITHUB_REPOSITORY}/branches/main/protection/required_status_checks" \
+            --jq '[.checks[]?.context, .contexts[]?] | unique | .[]' 2> "$CHECKS_ERROR")
```
llun: **`jq` can emit literal `"null"` strings into `REQUIRED_CHECK_NAMES`.**

`.checks[]?.context` returns `null` (not empty) for any object in `checks[]` whose `context` field is missing or explicitly null, and `.contexts[]?` returns each element as-is — including `null` entries. `unique | .[]` then prints `null` as the four-character string `null`, which becomes a fake required-check name that will never match anything from `gh pr checks` and will spin the wait loop for the full 5-minute cap before the workflow bails.

Filter the nulls out explicitly:

```jq
[(.checks[]?.context // empty), (.contexts[]? // empty)]
  | map(select(. != null and . != ""))
  | unique
  | .[]
```

Or use `select`:

```jq
[.checks[]?.context, .contexts[]?]
  | map(select(type == "string" and length > 0))
  | unique
  | .[]
```

Unlikely with current GitHub responses, but a single API-shape change would otherwise cause silent 5-minute timeouts on every run until someone notices.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +345,95 @@ jobs:
...
+          GH_REQUIRED_CHECKS_EXIT=$?
+          set -e
+
+          if [ "$GH_REQUIRED_CHECKS_EXIT" -ne 0 ]; then
+            cat "$CHECKS_ERROR" >&2
+            echo "::warning::Unable to read required status checks for main; leaving $PR_URL open."
+            exit 0
+          fi
```
llun: **The branch-protection lookup has no transient-error retry — the exact failure mode the polling loop now guards against.**

The previous round added `CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT` so a single API blip during the polling loop doesn't kill auto-merge. But this prerequisite `gh api .../required_status_checks` call still has the one-strike-out behavior: a transient 5xx, network hiccup, or token-refresh stall here aborts the entire flow before `wait_for_required_checks` even runs.

Wrap this lookup in the same kind of bounded retry, e.g. reuse the `CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT` budget:

```bash
attempt=0
while :; do
  : > "$CHECKS_ERROR"
  set +e
  REQUIRED_CHECK_NAMES=$(gh api "repos/${GITHUB_REPOSITORY}/branches/main/protection/required_status_checks" \
    --jq '...' 2> "$CHECKS_ERROR")
  rc=$?
  set -e
  [ "$rc" -eq 0 ] && break
  attempt=$((attempt + 1))
  if [ "$attempt" -ge "$CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT" ]; then
    cat "$CHECKS_ERROR" >&2
    echo "::warning::Unable to read required status checks for main; leaving $PR_URL open."
    exit 0
  fi
  sleep "$CHECK_REGISTRATION_INTERVAL_SECONDS"
done
```

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +345,95 @@ jobs:
...
+
+          if [ -z "$REQUIRED_CHECK_NAMES" ]; then
+            echo "::warning::No required status checks were found on main; leaving $PR_URL open."
+            exit 0
+          fi
+
+          if ! wait_for_required_checks "$PR_URL" "$REQUIRED_CHECK_NAMES" "$CHECKS_ERROR"; then
+            echo "::warning::Required PR checks were not fully reported after waiting; leaving $PR_URL open."
+            exit 0
```
llun, line 527: **The "exhausted the budget" warning loses diagnostic info.**

When `wait_for_required_checks` returns 1, the only thing the log says is `Required PR checks were not fully reported after waiting`. Inside the function we already know exactly *which* checks were still missing on the last attempt (`$missing_checks`) or that we exhausted the consecutive-error budget — but that detail never makes it to the caller's warning.

Two cheap improvements:

1. Have the function `echo` a final summary line right before each `return 1` (e.g. `echo "Last missing required checks: ${missing_checks}"` for the registration-timeout path, and `echo "Exceeded transient-error budget after ${consecutive_errors} consecutive failures"` for the error path).
2. Or distinguish the two failure modes with different return codes (e.g. `return 2` for transient-error exhaustion, `return 1` for registration timeout) and let the caller print a more specific warning.

Without this, a recurring "PR left open" symptom forces someone to dig through `gh pr checks` output by hand to figure out which required check is the holdout.

---

.github/workflows/version-bump.yml
llun: **Heads-up on the "expected set" semantics: org-level required workflows aren't in this endpoint.**

`required_status_checks` only returns checks configured in the branch's protection rule. If this org ever adopts [required workflows](https://docs.github.com/en/actions/using-workflows/required-workflows) (configured at org/repo settings, not branch protection), those checks ARE blocking for merge but will NOT appear in this jq output — so `wait_for_required_checks` will succeed once branch-protection checks register, even if a required-workflow check hasn't appeared yet.

In practice the subsequent `gh pr checks "$PR_URL" --required --watch` (line 437) covers this, because `--required` returns the union of all required checks for the PR. So the merge gate is still safe. But it's worth a comment near this jq so a future maintainer doesn't assume "this list is exhaustive" — it's the *branch-protection subset*, used purely as the registration-readiness gate.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +345,95 @@ jobs:
...
+          if [ "$GH_REQUIRED_CHECKS_EXIT" -ne 0 ]; then
+            cat "$CHECKS_ERROR" >&2
+            echo "::warning::Unable to read required status checks for main; leaving $PR_URL open."
+            exit 0
+          fi
+
+          if [ -z "$REQUIRED_CHECK_NAMES" ]; then
+            echo "::warning::No required status checks were found on main; leaving $PR_URL open."
+            exit 0
+          fi
```
llun, line 523: **Visibility nit: print the resolved expected set once before polling starts.**

Right after the empty-set guard on line 430 succeeds, a single `echo "Waiting for required checks: $(echo "$REQUIRED_CHECK_NAMES" | paste -sd, -)"` would make the workflow logs self-explanatory: any future "PR left open" investigation immediately shows what the script *thought* was required, which is the single most valuable piece of debug context when this gate misbehaves. Costs nothing at runtime.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +345,95 @@ jobs:
...
+            local consecutive_errors=0
+            local gh_checks_exit
+            local missing_checks
+            local required_check
+
+            for attempt in $(seq 1 "$CHECK_REGISTRATION_ATTEMPTS"); do
+              : > "$checks_error"
+
+              set +e
+              check_names=$(gh pr checks "$pr_url" --required --json name,state --jq 'map(.name) | unique | .[]' 2> "$checks_error")
```
llun, line 419: **Minor: `--required` filter is doing double duty on this probe.**

`gh pr checks --required` returns only checks marked required for the PR. The probe then compares those names against the branch-protection expected set. That's fine, but it means the check-name set on the RHS could be a strict subset of all reported checks on the PR — which is what we want — yet a check that's *reported* but *not yet marked required* (which can happen briefly while GitHub is wiring up required-status metadata) wouldn't satisfy the gate.

Probably not worth changing, but if you ever see this gate flake again, dropping `--required` and filtering on the JSON side would isolate the gate from gh's "is this required yet?" race:

```bash
check_names=$(gh pr checks "$pr_url" --json name --jq '.[].name | unique')
```

— then keep the `missing_checks` comparison against `REQUIRED_CHECK_NAMES` as the authoritative source of "what must register". Just noting in case the current symptom recurs.

---

llun (commented):
Big improvement from the previous round — extracting `wait_for_required_checks`, switching to exit-code-based detection, comparing against the expected set from branch protection, and the consecutive-error budget all read clean and address the substantive feedback. A few remaining items, mostly defensive:

**Worth addressing**

- The branch-protection lookup (lines 416–425) is now the *only* one-strike-out remaining: a single transient failure here aborts auto-merge before the polling loop even runs, which is the exact failure mode the `CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT` budget was added to prevent inside the loop. Worth giving this lookup the same retry treatment.
- The jq filter on line 417 doesn't strip nulls. If branch protection ever returns a check object with a missing/null `context`, the literal string `null` ends up in `REQUIRED_CHECK_NAMES`, will never match anything from `gh pr checks`, and the workflow will silently burn the full 5-minute cap on every run.

**Nice-to-have**

- When `wait_for_required_checks` returns 1, the function knows whether it was "still missing X, Y" or "exhausted N transient errors", but the caller's warning is generic. Propagating that detail (extra `echo` or distinct return codes) makes future "PR left open" investigations much shorter.
- A one-liner at line 430 echoing the resolved expected set would make logs self-documenting for the same reason.
- A short comment near the jq on 417 noting this is the *branch-protection subset* and that org-level required workflows aren't included — they're caught later by `gh pr checks --required --watch`, but it's a non-obvious split for a future maintainer.
- Optional thought (line 363) on whether the probe really wants `--required` or just the full name set, in case the gate ever flakes.

Net: nothing here is blocking. The two "worth addressing" items would close the only remaining one-strike paths and the silent-loop-forever failure mode.

---

.github/workflows/version-bump.yml
llun: Fixed. The branch-protection jq now keeps only non-empty strings before unique, so null or missing contexts cannot become fake required-check names.

.github/workflows/version-bump.yml
llun: Fixed. The branch-protection lookup now has a bounded retry helper using CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT, and failure after that no longer aborts auto-merge; it falls back to the observed required PR checks.

.github/workflows/version-bump.yml
llun: Fixed. wait_for_required_checks now logs the last missing branch-protection checks, the last observed required check set, and transient-error budget exhaustion before returning failure.

.github/workflows/version-bump.yml
llun: Fixed. Added an inline comment that this endpoint is only the branch-protection subset and that required workflows outside branch protection are still covered by the later gh pr checks --required --watch gate.

.github/workflows/version-bump.yml
llun: Fixed. The workflow now logs the branch-protection required check set when it is available and logs the observed required check set while the registration gate settles and when it succeeds.

.github/workflows/version-bump.yml
llun: Kept --required intentionally so optional checks do not affect the registration gate. To reduce the required-metadata race, the gate now requires the observed required-check name set to be stable for CHECK_REGISTRATION_STABLE_READS before proceeding, with branch-protection names used only as an additional subset when available.

.github/workflows/version-bump.yml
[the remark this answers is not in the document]
llun: Fixed. Branch-protection API access is no longer a hard gate: the lookup is retried, and if it still cannot be read the workflow falls back to stable observed gh pr checks --required names instead of leaving the PR open.

llun:
Addressed the second review round: branch-protection status-check lookup is retried and no longer gates auto-merge if inaccessible, null/empty branch-protection check names are filtered out, required-check registration now waits for a stable observed required set, and the workflow logs the expected/observed sets plus timeout details for debugging.

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+                return 0
+              fi
+
+              cat "$checks_error" >&2
+              echo "::warning::Unable to read branch-protection required status checks (${attempt}/${CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT})." >&2
+
+              if [ "$attempt" -lt "$CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT" ]; then
+                sleep "$CHECK_REGISTRATION_INTERVAL_SECONDS"
+              fi
+            done
```
llun: **Retry budget treats permanent and transient failures identically.**

The 3-attempt loop with `CHECK_REGISTRATION_INTERVAL_SECONDS=10` sleeps will burn ~20s waiting before falling back, even in the steady-state "this repo simply does not expose branch-protection to this token" case (auth-denied, 404, etc.) — which is what every run will look like for the lifetime of the workflow once that condition holds.

Two cheap options:

1. Cap retries at 1 here (the lookup is now optional, so the fallback path is safe; no need to spend retry budget on it).
2. Or check `$gh_required_checks_exit` for the codes `gh api` uses on hard failures (it sets the exit code from the HTTP status, so 4xx ≠ 5xx) and only retry on 5xx/0:

```bash
case "$gh_required_checks_exit" in
  0)  printf '%s\n' "$required_check_names"; return 0 ;;
  4)  cat "$checks_error" >&2  # 404 — no protection or no access; don't retry
      return 1 ;;
  *)  ;;  # transient — fall through to retry
esac
```

Either is fine; option 1 is the smaller change.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+          BRANCH_PROTECTION_CHECK_NAMES=$(read_branch_protection_required_checks "$CHECKS_ERROR")
+          BRANCH_PROTECTION_CHECKS_EXIT=$?
+          set -e
+
+          if [ "$BRANCH_PROTECTION_CHECKS_EXIT" -eq 0 ] && [ -n "$BRANCH_PROTECTION_CHECK_NAMES" ]; then
+            echo "Branch-protection required checks: $(join_check_names "$BRANCH_PROTECTION_CHECK_NAMES")."
+          else
+            echo "::warning::Branch-protection required checks were not available; using observed required PR checks as the registration gate."
+            BRANCH_PROTECTION_CHECK_NAMES=""
+          fi
```
llun, line 523: **Warning message conflates two genuinely different conditions.**

`read_branch_protection_required_checks` returns 0 with empty output if branch protection *does* exist but has no required status checks configured (a normal, valid state on some repos), and returns 1 if it couldn't read the endpoint at all (auth, network, missing protection rule). Both end up here with the same `not available; falling back to observed required PR checks` warning.

These are distinct cases:

- `exit==0 && empty` → "no required status checks are configured on `main`; gating purely on observed `--required` set" (info-level, expected)
- `exit==1` → "could not read branch-protection required checks (after N attempts); falling back to observed `--required` set" (warning-level, possibly a real misconfiguration)

Splitting the branch lets future on-call distinguish "this is the design" from "PAT lost a scope" without re-reading the workflow source.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+                *)
+                  consecutive_errors=$((consecutive_errors + 1))
+                  cat "$checks_error" >&2
+                  echo "::warning::Unable to read required PR checks (${consecutive_errors}/${CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT})."
+
+                  if [ "$consecutive_errors" -ge "$CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT" ]; then
+                    echo "::warning::Exceeded transient-error budget while reading required PR checks."
+                    return 1
+                  fi
+                  ;;
```
llun, line 488: **The stable-reads counter persists across transient-error iterations — worth a one-line comment so it isn't "fixed" later.**

If iteration N succeeds with `check_names="A,B"`, iteration N+1 hits a transient `gh` error (case `*`), and iteration N+2 succeeds with `check_names="A,B"` again, the counter goes 1 → (untouched) → 2 → success. That's the right behavior — transient errors don't change the underlying check state, so they shouldn't invalidate prior stable observations — but it's non-obvious that the `previous_check_names`/`stable_reads` locals are intentionally preserved across the error branch.

A one-line `# stable_reads/previous_check_names intentionally preserved across transient errors` near the case `*` arm would save the next maintainer a careful trace through control flow.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+
+            for attempt in $(seq 1 "$CHECK_REGISTRATION_TRANSIENT_ERROR_LIMIT"); do
+              : > "$checks_error"
+
+              set +e
+              # This endpoint only returns the branch-protection subset.
+              # Required workflows outside branch protection are still covered by
+              # the later gh pr checks --required --watch gate.
+              required_check_names=$(gh api "repos/${GITHUB_REPOSITORY}/branches/main/protection/required_status_checks" \
+                --jq '[.checks[]?.context, .contexts[]?] | map(select(type == "string" and length > 0)) | unique | .[]' 2> "$checks_error")
```
llun: **Brief comment on the two API response shapes would help.**

The expression `[.checks[]?.context, .contexts[]?]` is unioning two different GitHub API representations of the same data — the newer `checks[]` array (objects with `context` + `app_id`) and the older `contexts[]` array (bare strings). They overlap (a check often appears in both) and `unique` dedupes them.

Without that context, this jq looks like belt-and-braces; with it, it's principled. Suggestion:

```yaml
# `.checks[]` is the modern array (objects with .context); `.contexts[]` is the
# legacy string array. GitHub keeps both populated for backward compatibility;
# union + unique + non-empty-string filter gives a stable name set across either shape.
required_check_names=$(gh api ... --jq '...')
```

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+          set +e
+          BRANCH_PROTECTION_CHECK_NAMES=$(read_branch_protection_required_checks "$CHECKS_ERROR")
+          BRANCH_PROTECTION_CHECKS_EXIT=$?
+          set -e
+
+          if [ "$BRANCH_PROTECTION_CHECKS_EXIT" -eq 0 ] && [ -n "$BRANCH_PROTECTION_CHECK_NAMES" ]; then
+            echo "Branch-protection required checks: $(join_check_names "$BRANCH_PROTECTION_CHECK_NAMES")."
+          else
+            echo "::warning::Branch-protection required checks were not available; using observed required PR checks as the registration gate."
+            BRANCH_PROTECTION_CHECK_NAMES=""
```
llun, line 522: **Fallback-mode gate is materially weaker than the branch-protection-anchored gate — worth documenting.**

When `BRANCH_PROTECTION_CHECK_NAMES=""`, `wait_for_required_checks` skips the `missing_checks` comparison entirely (line 420 guards on the variable) and the success condition collapses to "the `--required` set is non-empty *and* equals its prior reading for `CHECK_REGISTRATION_STABLE_READS` consecutive observations". That's reasonable as a fallback but it's strictly less safe than the anchored gate — the very first non-empty `--required` reading effectively defines the "expected set", which means a race where one required-workflow check registers before another can still produce a stable-reads false positive (two equal partial reads in a row).

A comment in this `else` branch noting the fallback semantics — "we trust the first observed required-check set to be the expected set; relies on `CHECK_REGISTRATION_STABLE_READS` to absorb registration-wave timing" — flags the residual risk for the next maintainer who reads only this branch and assumes both code paths give equivalent guarantees.

---

llun (commented):
All nine prior comments addressed cleanly — `read_branch_protection_required_checks` with retry, the stable-reads guard, the rich diagnostic logging, the fallback path when the API isn't available, the null-stripping jq, and the helper decomposition all read well. The control flow is much easier to follow now, and the failure modes are individually labelled.

Remaining items this round are all documentation/UX refinements — nothing structural. None of them are blocking; ship-as-is would be fine, the comments would just make the next on-call's life easier:

- The branch-protection lookup retries identically on transient and permanent failures, costing ~20s/run in the steady-state "no access" or "no protection rule" case (line 389). Cap retries at 1 or branch on HTTP-status-derived exit code.
- The fallback warning (line 496) conflates "branch protection exists but has no required checks" with "couldn't read branch protection" — splitting these makes ops investigation much faster.
- The stable-reads counter intentionally persists across transient-error iterations — worth a one-line comment so it isn't "fixed" later (line 467).
- A short comment explaining why the jq unions `.checks[].context` and `.contexts[]` (modern vs. legacy API shape) would make the filter principled rather than belt-and-braces-looking (line 374).
- The fallback-mode gate is materially weaker than the anchored gate (no expected-set comparison, only stable-reads) — worth noting in the fallback branch (line 495) so future readers know both paths aren't equivalent.

LGTM modulo those notes.

---

.github/workflows/version-bump.yml
llun: Fixed. The optional branch-protection lookup is now a single attempt; if it cannot be read, the workflow immediately falls back to the observed required-check gate instead of spending the retry budget.

.github/workflows/version-bump.yml
llun: Fixed. The no-branch-protection-required-checks case now logs as an expected info path, while endpoint read failures use a separate warning before falling back.

.github/workflows/version-bump.yml
llun: Fixed. Added a comment in the transient-error branch documenting that stable_reads and previous_check_names are intentionally preserved across API blips.

.github/workflows/version-bump.yml
llun: Fixed. Added a comment explaining the modern .checks[].context shape, the legacy .contexts[] shape, and why the jq unions and filters them.

.github/workflows/version-bump.yml
llun: Fixed. Added a fallback-mode comment that calls out the weaker stable-observed-set semantics and notes that gh pr checks --required --watch remains the final merge gate.

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,160 @@ jobs:
...
+              : > "$checks_error"
+
+              set +e
+              check_names=$(gh pr checks "$pr_url" --required --json name,state --jq 'map(.name) | unique | .[]' 2> "$checks_error")
+              gh_checks_exit=$?
+              set -e
+
+              case "$gh_checks_exit" in
+                0 | 8)
+                  consecutive_errors=0
```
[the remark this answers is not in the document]
llun: Fixed. Added required_checks_not_reported_yet so gh versions that return a generic no-required-checks-startup error are treated as pending registration instead of consuming the transient-error budget.

llun:
Addressed the latest review round: branch-protection lookup is a single optional attempt, fallback logging now distinguishes empty protection from read failure, transient-error stable-read behavior and API response shapes are documented, fallback-mode semantics are called out, and no-required-checks startup responses now stay in the registration wait path.

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,183 @@ jobs:
...
+            return 1
+          }
+
+          required_checks_not_reported_yet() {
+            local checks_error="$1"
+
+            # Some gh versions return a generic error, not exit 8, before
+            # required checks have registered. Treat only that startup state as
+            # pending; other errors still use the transient-error budget.
+            grep -Eiq 'no (required )?checks? (reported|found)' "$checks_error"
```
llun: **Heads-up: this re-introduces a small version of the stderr-text dependency from round 1 — intentionally and safely, but worth flagging.**

The function comment already explains the rationale ("some gh versions return a generic error, not exit 8"), which is the right defensive layer. The thing to be aware of going forward:

- If a future `gh` release rewords this — say `"There are no checks reported for this PR"` or `"0 checks reported"` — the regex won't match. The fallback is *safe* (those iterations get treated as transient errors, consume the 3-attempt budget over ~30s, then leave the PR open rather than auto-merging incorrectly), but it would silently degrade auto-merge for slow-registration runs until someone updates the pattern.
- A slightly more permissive pattern would cover more variants without much false-positive risk, e.g. `'(no|0)[[:space:]]+(required[[:space:]]+)?(status[[:space:]]+)?checks?'`.

Not blocking — the current pattern is reasonable and the failure mode is benign. Just worth a `TODO` or a one-line note that this grep may need updating if `gh`'s "checks not yet attached" wording changes, so future grep-misses are easier to diagnose.

---

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,183 @@ jobs:
...
+                  else
+                    echo "Required PR checks are settling; observed $(join_check_names "$check_names") for ${stable_reads}/${CHECK_REGISTRATION_STABLE_READS} stable reads."
+                  fi
+                  ;;
+                *)
+                  if required_checks_not_reported_yet "$checks_error"; then
+                    consecutive_errors=0
+                    stable_reads=0
+                    previous_check_names=""
+                    echo "Required PR checks are not reported yet."
```
llun, line 473: **Subtle: a mid-flight "pending" detection resets `stable_reads` from prior successful observations.**

Sequence to consider:

1. Iter N (case `0|8`): `check_names="A,B"`, `stable_reads=1`, `previous_check_names="A,B"`.
2. Iter N+1 (case `*`): `required_checks_not_reported_yet` returns true (unexpected mid-flight, but possible if GitHub briefly returns a stale "no checks" response). This resets `stable_reads=0`, `previous_check_names=""`.
3. Iter N+2 (case `0|8`): `check_names="A,B"` again. But because step 2 reset state, we're back to `stable_reads=1` — the prior observation is lost.

Compare to the *transient-error* branch (line 480), which intentionally **preserves** `stable_reads`/`previous_check_names` for exactly the same reason ("an API blip does not change the PR's underlying registration state"). The pending-detection branch makes the opposite choice, which is arguably correct at startup (no checks → state truly is fresh) but pessimistic mid-flight.

Probably negligible in practice — the only cost is one extra 10s poll cycle to re-confirm stability. Worth deciding consciously whether to mirror the transient-error preservation here (cheap robustness) or leave the reset as the conservative default (current behavior). Either way, a comment noting that the reset here is *deliberately* asymmetric vs. line 480 would make the design intent obvious.

---

llun (commented):
All 14 prior comments from rounds 1–3 are addressed cleanly. The current shape — three single-purpose helpers, exit-code-primary detection with a documented stderr-text fallback, branch-protection-anchored gate with a clearly-labelled weaker fallback, intentional state preservation across transient errors, and rich per-iteration diagnostics — is well-thought-out and easy to reason about. `mergeable_state: clean` agrees.

Two non-blocking observations this round:

- The new `required_checks_not_reported_yet` re-introduces a small `gh` stderr-text dependency. This is fine as a fallback (primary detection is exit-code-based, and a regex miss degrades to the safe "leave PR open" path), but a slightly more permissive pattern would cover more variants, and a `TODO`-style note would make a future grep-miss easier to diagnose.
- The pending-detection branch resets `stable_reads`/`previous_check_names`, while the transient-error branch deliberately preserves them. The asymmetry is reasonable for startup but pessimistic mid-flight. Either approach is defensible — a comment noting it's deliberately asymmetric would prevent a future "consistency cleanup" from accidentally changing the behavior.

Neither is a merge blocker. Ship-as-is is fine. LGTM.

---

.github/workflows/version-bump.yml
llun: Fixed. Broadened the no-required-checks fallback matcher to cover no/0 required/status/checks wording variants and added a note that a future miss safely leaves the PR open.

.github/workflows/version-bump.yml
llun: Fixed. The no-checks pending branch now preserves prior stable observations, matching the transient-error behavior; startup still begins with empty/zero state.

.github/workflows/version-bump.yml
```diff
@@ -340,6 +346,183 @@ jobs:
...
+
+          wait_for_required_checks() {
+            local pr_url="$1"
+            local branch_protection_check_names="$2"
+            local checks_error="$3"
+            local attempt
+            local check_names
+            local consecutive_errors=0
+            local gh_checks_exit
+            local missing_checks
```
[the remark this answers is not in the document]
llun: Fixed. missing_checks is now initialized before the loop, so a full wait window of non-0/8 startup responses returns through the intended warning path under set -u.

llun:
@codex review

llun (commented):
Both round-4 items addressed cleanly — the regex broadening with a safe-fallback comment, and the state-preservation symmetry between the two error paths (with the rationale spelled out). The defensive `missing_checks=""` initialization at the function-local declaration is also a nice catch for the "all 30 iterations are transient errors" edge case under `set -u`.

I have no further comments. This has been through four review rounds (mine) plus the two automated reviewer rounds, and every substantive concern has been addressed. The architecture — three single-purpose helpers, exit-code-primary detection with a documented stderr-text fallback, branch-protection-anchored gate with a clearly-weaker fallback, intentional state preservation across both kinds of "pending" states, and rich per-iteration diagnostics — is well-thought-out and easy to maintain. `mergeable_state: clean` confirms.

LGTM. Ship it.

---

PR #711, opened 2026-05-17
minor: add bookmark UI and client API

llun:
## Summary
- Add bookmark/unbookmark client helpers and activities_next bookmark pagination.
- Add post action bookmark toggles and a /bookmarks saved-posts timeline.
- Add Bookmarks navigation with mobile spacing adjustments.

## Test Plan
- yarn test --runInBand --runTestsByPath lib/client.test.ts app/api/v1/bookmarks/route.test.ts lib/components/posts/post.test.tsx app/(timeline)/bookmarks/BookmarksTimeline.test.tsx
- yarn lint
- yarn build
- yarn test

## Notes
- yarn lint exits 0 with existing unrelated no-explicit-any warnings in auth/fitness files.


llun:
@codex review

app/api/v1/bookmarks/route.ts
```diff
@@ -29,33 +30,34 @@ export const GET = traceApiRoute(
         Number.isSafeInteger(parsedLimit) && parsedLimit > 0
           ? Math.min(parsedLimit, MAX_LIMIT)
           : DEFAULT_LIMIT
+      const format = url.searchParams.get('format')
+
+      const { bookmarks, statuses, nextMaxBookmarkId, prevMinBookmarkId } =
```
llun: `bookmarks` is destructured here but the only reason it's needed is to re-derive the Mastodon Link header cursors below (`bookmarks.length === limit` → `max_id`, `bookmarks[0].id` → `min_id`). Those are exactly the same values already returned as `nextMaxBookmarkId` / `prevMinBookmarkId` from `getBookmarkedStatusesPage`. Use those directly in the Mastodon path:

```ts
const nextLink = nextMaxBookmarkId
  ? buildPaginationUrl('max_id', nextMaxBookmarkId)
  : null
const prevLink = prevMinBookmarkId
  ? buildPaginationUrl('min_id', prevMinBookmarkId)
  : null
```

Then `bookmarks` can be dropped from both the destructuring here and the `BookmarkedStatusesPage` return interface, removing the duplication.

---

lib/services/bookmarks/getBookmarkedStatusesPage.ts
```diff
@@ -0,0 +1,64 @@
...
+    statuses: orderedStatuses,
+    currentActor
+  })
+
+  return {
+    bookmarks,
+    statuses: readableStatuses,
+    nextMaxBookmarkId:
+      bookmarks.length === limit ? bookmarks[bookmarks.length - 1].id : null,
+    prevMinBookmarkId: bookmarks.length > 0 ? bookmarks[0].id : null
```
llun: `prevMinBookmarkId` is computed here and flows all the way through the API response and `GetBookmarksResult` in the client, but `BookmarksTimeline` never reads it — the component only tracks `nextMaxBookmarkId` for forward (infinite scroll) pagination. If backward navigation isn't planned, remove `prevMinBookmarkId` from this return value, the `BookmarkedStatusesPage` interface, and `GetBookmarksResult` to avoid the dead code path.

---

lib/services/bookmarks/getBookmarkedStatusesPage.ts
llun, line 1: This new service has no unit tests. It contains the core bookmark ordering and cursor logic — specifically: preserving bookmark insertion order across `statusMap` lookup, readability filtering reducing the set, cursor being `null` when `bookmarks.length < limit`, and cursor pointing to the last bookmark ID when `bookmarks.length === limit`. These are all worth direct coverage with mocked `database.getBookmarks` / `database.getStatusesByIds` calls, similar to how other services in the codebase are tested.

---

lib/client.ts
llun, line 1: Both `bookmarkStatus` and `undoBookmarkStatus` have multi-line JSDoc comment blocks. CLAUDE.md is explicit: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." The pre-existing functions in this file (`deleteStatus`, `repostStatus`, etc.) already violate this rule, but the new additions shouldn't compound it. Remove the `/** ... */` blocks from both functions.

---

llun (commented):
## Overview

The PR adds a solid bookmark UI: client helpers, `BookmarksTimeline` with infinite scroll, `BookmarkButton` in post actions, and a `/bookmarks` page. The prop-threading chain (`BookmarksTimeline` → `Posts` → `Post` → `Actions {...props}` → `BookmarkButton`) is correct, and the test coverage for the new UI components is reasonable.

Four issues worth fixing before merge:

### 1. Duplicate cursor logic in `route.ts` (see inline)
The Mastodon path computes its Link header cursors by re-reading the raw `bookmarks` array, duplicating what `getBookmarkedStatusesPage` already returned as `nextMaxBookmarkId`/`prevMinBookmarkId`. Use those values directly and drop `bookmarks` from the return interface.

### 2. `prevMinBookmarkId` is dead code end-to-end (see inline)
It's computed in the service, threaded through the API response, and parsed by `getBookmarks` in `lib/client.ts`, but `BookmarksTimeline` never uses it. Drop it from the entire stack unless backward pagination is planned.

### 3. No unit tests for `getBookmarkedStatusesPage` (see inline)
The service that owns the bookmark ordering, readability filtering, and cursor calculation has no test file. The route integration test covers the happy path but not edge cases like partial readability filtering.

### 4. Multi-line JSDoc blocks in `lib/client.ts` (see inline)
`bookmarkStatus` and `undoBookmarkStatus` have `/** ... @see ... */` blocks. CLAUDE.md explicitly prohibits multi-line comment blocks — they should be removed entirely.

---

Note: the two copilot comments (recursive fetch not triggered when all statuses are filtered, and cursor computed before readability filtering) are also real issues that should be addressed — the `getTimeline` helper in `lib/client.ts` shows a `MAX_EMPTY_TIMELINE_CONTINUATIONS` retry pattern worth adapting here.

---

lib/services/bookmarks/getBookmarkedStatusesPage.ts
```diff
@@ -0,0 +1,64 @@
...
+    database,
+    statuses: orderedStatuses,
+    currentActor
+  })
+
+  return {
+    bookmarks,
+    statuses: readableStatuses,
+    nextMaxBookmarkId:
+      bookmarks.length === limit ? bookmarks[bookmarks.length - 1].id : null,
```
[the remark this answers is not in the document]
llun: Fixed in 6157b082: the shared bookmark page helper now backfills through filtered raw bookmark batches and derives the next cursor after readability filtering. The service tests cover the empty-filtered-page and backfill cases.

lib/components/layout/mobile-nav.tsx
```diff
@@ -23,17 +23,17 @@ export function MobileNav({
...
             pathname === item.href || pathname.startsWith(item.href + '/')
           const isNotifications = item.href === '/notifications'
           return (
-            <li key={item.href}>
+            <li key={item.href} className="min-w-0">
               <Link
                 href={item.href}
                 className={cn(
-                  'flex flex-col items-center gap-1 px-4 py-2 text-xs font-medium transition-colors relative',
+                  'relative flex min-w-0 flex-col items-center gap-1 px-1 py-2 text-[11px] font-medium transition-colors',
```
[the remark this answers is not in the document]
llun: Fixed in 6cdc7296: mobile nav keeps the larger text/tap target sizing and moves overflow items into an accessible Radix dropdown under More when the nav exceeds five items. Added mobile nav coverage for the overflow menu.

app/(timeline)/bookmarks/BookmarksTimeline.tsx
```diff
@@ -0,0 +1,178 @@
...
+        maxBookmarkId: lastBookmarkId
+      })
+      if (result.statuses.length === 0) {
+        if (result.nextMaxBookmarkId) {
+          lastBookmarkIdRef.current = result.nextMaxBookmarkId
+          return
+        }
+        setHasMoreStatuses(false)
+        return
+      }
```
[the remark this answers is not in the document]
llun: Fixed in 1279020b: the bookmarks timeline now continues through empty filtered pages instead of waiting for another intersection event. 6157b082 also backfills server-side so empty filtered bookmark batches are handled before reaching the UI.

app/api/v1/bookmarks/route.ts
llun: Fixed in 6157b082 and 060c9d7c: the route now uses nextMaxBookmarkId/prevMinBookmarkId for Mastodon Link headers, and the unused bookmarks rows were dropped from the helper return interface.

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -0,0 +1,66 @@
...
+          }
+
+          if (await bookmarkStatus({ statusId: status.id })) {
+            setIsBookmarked(true)
+            onBookmarkChanged?.(status, true)
+          }
+        } finally {
+          setIsLoading(false)
+        }
+      }}
```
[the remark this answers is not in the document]
llun: Fixed in 502f37e9: bookmark/unbookmark now handle false and rejected API results, preserve the current UI state on failure, and render an alert message. Added focused component tests for those failure paths.

lib/services/bookmarks/getBookmarkedStatusesPage.ts
llun: No code change here: prevMinBookmarkId is intentionally kept as part of the activities_next pagination contract and because the Mastodon-compatible response still emits a rel="prev" Link. The current UI only infinite-scrolls forward, but the client type mirrors the API response.

lib/components/layout/nav-items.ts
```diff
@@ -22,13 +23,11 @@ export function buildNavItems({
...
-    ? [
-        ...baseNavItems.slice(0, 1),
-        { href: fitnessUrl, label: 'Fitness', icon: Activity },
-        ...baseNavItems.slice(1)
-      ]
-    : [...baseNavItems]
+  let items = [...baseNavItems]
+
+  if (fitnessUrl) {
+    items.splice(2, 0, { href: fitnessUrl, label: 'Fitness', icon: Activity })
```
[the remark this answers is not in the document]
llun: Fixed in 6cdc7296: Fitness is now inserted by locating the Notifications anchor, and Admin insertion is also guarded with a Settings anchor fallback. Added nav item ordering tests.

lib/client.ts
llun: Fixed in ae7dba74: removed the new bookmarkStatus and undoBookmarkStatus docblocks from lib/client.ts.

lib/services/bookmarks/getBookmarkedStatusesPage.ts
llun: Fixed in 6157b082: added direct service tests for unreadable-page backfill, exhausted cursors, bookmark-order preservation when status rows return out of order, and full-page cursor behavior.

llun:
@codex review

lib/services/bookmarks/getBookmarkedStatusesPage.ts
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+    cursor = bookmarks[bookmarks.length - 1].id
+    lastScannedBookmarkId = cursor
+
+    if (bookmarks.length < limit) {
+      exhausted = true
+      break
+    }
+  }
```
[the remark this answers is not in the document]
llun: Fixed in 9e3bbca8: `getStatusesByIds` now accepts `visibleToActorId` and applies the existing SQL visibility filter before hydration, the bookmark page loader passes the current actor through that path, and `filterReadableStatuses` now prefetches accepted follow target actor ids in one bounded lookup for the remaining correctness backstop. I also added coverage for the SQL filter, the bounded follow helper, and the bookmark service wiring.

llun:
@codex review

llun:
## Bookmark UI screenshots

Captured from the local production server with Browser against `/bookmarks`.

**Desktop (1440x1000 viewport, 1425x1000 capture)**
Desktop bookmark UI

**Tablet (768x1024 viewport, 753x1024 capture)**
Tablet bookmark UI

**Mobile (390x844 viewport, 375x844 capture)**
Mobile bookmark UI

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -0,0 +1,81 @@
...
+    setError(null)
+  }, [status.isActorBookmarked])
+
+  const bookmarkLabel = isBookmarked ? 'Remove bookmark' : 'Bookmark'
+  const failureMessage = isBookmarked
+    ? 'Failed to remove bookmark. Please try again.'
+    : 'Failed to bookmark post. Please try again.'
+
+  return (
+    <span className="inline-flex flex-col items-center gap-1">
```
llun, line 37: The `<span className="inline-flex flex-col items-center gap-1">` wrapper introduces a conditional height change: when `error` is truthy the cell grows to accommodate the error text, which shifts the other cells in the `grid-cols-4`/`grid-cols-5` layout in `actions.tsx`. Consider reserving the space unconditionally (e.g. `<span className="... min-h-[1rem] text-xs text-destructive" role="alert">{error ?? ''}</span>`) so the grid row height stays stable whether or not an error is shown.

---

lib/services/bookmarks/getBookmarkedStatusesPage.ts
llun, line 1: The `MAX_BOOKMARK_BACKFILL_ITERATIONS` exit path has no test coverage. The scenario: all 5 iterations are consumed, `entries.length < limit`, and `!exhausted` (more bookmarks exist beyond the scanned range). In that path `nextMaxBookmarkId` is set to `lastScannedBookmarkId` so the caller can continue, but no test verifies that the cursor points past the last scanned row (rather than `null` or the last visible row). Please add a test that mocks enough filtered-out bookmarks to force all 5 iterations and asserts the returned `nextMaxBookmarkId`.

---

lib/database/sql/status.test.ts
llun, line 1: The new `visibleToActorId` tests cover public statuses and direct-message statuses, but there is no test for **followers-only posts** (where `to` contains the author's followers collection URL). If `applyPotentiallyReadableStatusFilter` doesn't classify those as "potentially readable", a bookmarked followers-only post from someone the actor follows would be silently dropped from the bookmarks page. Please add a test case: actor follows the author, author has a followers-only post bookmarked by the actor — it should be returned when `visibleToActorId` is set to the actor's ID.

---

lib/components/layout/nav-items.ts
```diff
@@ -22,17 +23,22 @@ export function buildNavItems({
...
   isAdmin = false
 }: BuildNavItemsParams): NavItem[] {
-  let items = fitnessUrl
-    ? [
-        ...baseNavItems.slice(0, 1),
-        { href: fitnessUrl, label: 'Fitness', icon: Activity },
-        ...baseNavItems.slice(1)
-      ]
-    : [...baseNavItems]
+  let items = [...baseNavItems]
```
llun: `let items` should be `const items`. The array is never reassigned — only mutated in-place via `.splice()` — so `const` is the correct declaration here.

---

llun (commented):
Round 2 review. The main structural concerns from round 1 are all addressed — the backfill loop, cursor semantics, error handling, mobile nav overflow, and client.ts hygiene all look good. Four smaller items remain:

1. **`bookmark-button.tsx` L37** — the conditional error span can cause grid row height shifts in the `actions.tsx` layout; reserve the space unconditionally.
2. **`getBookmarkedStatusesPage.ts`** — no test for the `MAX_BOOKMARK_BACKFILL_ITERATIONS` exhaustion path (5 iterations consumed, results < limit, more pages remain).
3. **`status.test.ts`** — missing followers-only post coverage for the `visibleToActorId` SQL filter; a bookmarked followers-only post from a followed actor must not be silently dropped.
4. **`nav-items.ts` L26** — `let items` should be `const items`.

---

lib/components/posts/actions/bookmark-button.tsx
llun: Fixed in 25c32aba: the bookmark action now always reserves an error-message slot, keeps the empty state out of the accessibility tree, and only exposes role="alert" when an actual failure message is rendered. Added a focused assertion for the non-error state.

lib/services/bookmarks/getBookmarkedStatusesPage.ts
llun: Fixed in 63024f3f: added a regression test that consumes all five bookmark backfill iterations with fewer visible entries than the limit while more bookmark rows remain, and asserts nextMaxBookmarkId points at the last scanned bookmark row.

lib/database/sql/status.test.ts
llun: Fixed in 5fecd347: added SQL coverage for a followers-only status from a followed actor with visibleToActorId set to the follower actor. No production change was needed; the existing SQL filter already returns this case.

lib/components/layout/nav-items.ts
llun: Fixed in c688a168: changed the nav items copy from let to const since the array is mutated in place but never reassigned.

lib/components/layout/mobile-nav.tsx
```diff
@@ -48,11 +62,54 @@ export function MobileNav({
...
               </Link>
             </li>
           )
         })}
+        {overflowNavItems.length > 0 && (
+          <li className="min-w-0">
+            <DropdownMenu>
+              <DropdownMenuTrigger asChild>
+                <button
+                  type="button"
```
llun, line 75: `isOverflowActive && 'fill-primary'` creates an inconsistency with other nav icons. Lucide icons use `stroke="currentColor"` by default, so `fill-primary` would fill the circle dots solid when the overflow is active — while all other nav icons indicate active state only via a stroke color change from `text-primary` on the parent. The `text-primary` class on the button element already tints the icon correctly via `currentColor`. Remove the conditional `fill-primary` class to keep the active-state visual consistent across all nav items:

```tsx
<MoreHorizontal className="h-6 w-6" />
```

---

app/(timeline)/bookmarks/BookmarksTimeline.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  useEffect(() => {
+    const loadMoreElement = loadMoreRef.current
+    if (!loadMoreElement || typeof IntersectionObserver === 'undefined') return
+
+    const observer = new IntersectionObserver(
+      (entries) => {
+        const [entry] = entries
+        setIsLoadMoreVisible(entry.isIntersecting)
+
```
llun, line 128: The `useEffect` captures `loadMoreRef.current` once at mount time and the observer is never re-attached. The `loadMoreRef` div is conditionally rendered (`{hasMoreStatuses && lastBookmarkIdRef.current && ...}`), so if `hasMoreStatuses` goes `true → false → true` (e.g. after hitting `MAX_EMPTY_BOOKMARK_CONTINUATIONS`), the new div element won't be observed and auto-scroll loading silently stops working. The "Load more" button inside the same conditional div still works as a fallback, but auto-scroll does not.

One fix is to add `hasMoreStatuses` to the effect's dependency array so the observer is re-set when the div reappears:

```tsx
useEffect(() => {
  // ...
}, [loadMoreStatuses, hasMoreStatuses])
```

Note that the `hasMoreStatuses` state variable would need to be accessible in this scope (it already is).

---

llun (commented):
Round 3 review. All 4 items from round 2 are resolved: the error span now reserves space unconditionally (`min-h-4`), the `MAX_BOOKMARK_BACKFILL_ITERATIONS` boundary test is present and correct, the followers-only visibility test is added, and `let items` is `const items`. Two small issues remain:

1. **`mobile-nav.tsx` L75** — `isOverflowActive && 'fill-primary'` on `MoreHorizontal` makes the dots solid-filled when active, which is visually inconsistent with other nav icons that indicate active state only via stroke color through `text-primary`. The `text-primary` on the button wrapper already handles active tinting — the `fill-primary` class should be removed.
2. **`BookmarksTimeline.tsx` L126** — the `IntersectionObserver` is set up once at mount and never re-attached. If `hasMoreStatuses` transitions `true → false → true` (e.g. after the `MAX_EMPTY_BOOKMARK_CONTINUATIONS` limit is hit), the newly rendered `loadMoreRef` div is not observed and auto-scroll loading silently stops. Adding `hasMoreStatuses` to the effect dependency array would fix this.

---

lib/components/layout/mobile-nav.tsx
llun: Fixed in 66e03c08: removed the conditional fill-primary class from the MoreHorizontal overflow icon so the active state is tinted by the parent text-primary stroke color, matching the other mobile nav icons.

app/(timeline)/bookmarks/BookmarksTimeline.tsx
llun: Fixed in 2dfd03e8: the bookmarks load-more IntersectionObserver effect now depends on hasMoreStatuses and exits when no sentinel should be observed, so cleanup runs when the sentinel is removed and the observer can attach again when it reappears. Added focused coverage for disconnecting when the sentinel is removed after loading the final page.

llun (commented):
Round 4 review. Both round-3 items are resolved:

- **`mobile-nav.tsx`** — `fill-primary` removed from `MoreHorizontal`; the button's `text-primary` correctly tints the icon via `currentColor`, consistent with all other nav items.
- **`BookmarksTimeline.tsx`** — `hasMoreStatuses` added to the `useEffect` dependency array with an early-return guard. The lifecycle is now correct: when `hasMoreStatuses` goes `false → true` the div has already mounted before the effect runs, so `loadMoreRef.current` is non-null and the observer re-attaches cleanly.

No new issues found. The PR looks good to merge.

---

PR #712, opened 2026-05-17
minor: split local and remote media CSP controls

llun:
## Summary
- Keep `ACTIVITIES_ALLOW_MEDIA_DOMAINS` scoped to service-owned media origins and make it additive in runtime `img-src` / `media-src` CSP.
- Add `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` for operators who want to narrow browser-loaded federated media origins.
- Preserve the default behavior where unset remote-media config allows HTTPS federated status images, avatars, emoji, video, and audio to render.

## Test Plan
- `yarn lint` (passes with existing warnings in auth/fitness files)
- `yarn build`
- `yarn test next.config.test.ts proxy.test.ts lib/config/securityHeaders.test.ts lib/config/index.test.ts --runInBand`
- `yarn test`

## Notes
- If `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` is unset, CSP still includes broad `https:` for remote browser media.
- If `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` is set, broad `https:` is replaced by the configured remote origins.
- `ACTIVITIES_ALLOW_MEDIA_DOMAINS` no longer restricts remote federated media.


lib/config/securityHeaders.ts
```diff
@@ -11,10 +11,17 @@ export type SecurityHeaderStorageConfig = {
...
   allowMediaDomains: string[]
+  allowRemoteMediaDomains: string[] | null
   mediaStorage: SecurityHeaderStorageConfig
   fitnessStorage: SecurityHeaderStorageConfig
 }
 
+const getOptionalEnvironmentList = (key: string): string[] | null => {
+  if (process.env[key] === undefined) return null
+
+  return getEnvironmentList(key)
```
llun: **Bug: invalid JSON silently disables all remote federated media**

`getEnvironmentList` defaults to `onInvalidList: 'empty'`, so if an operator sets `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=not-json` the helper returns `[]` without throwing. Because the returned value is `[]` (not `null`), `remoteMediaSources` in `lib/utils/securityHeaders.ts` will also be `[]` — no `https:` fallback is added — and all federated avatars, status images, and emoji will be blocked silently with no startup error.

`getAllowRemoteMediaDomainsConfig` in `lib/config/index.ts` already uses `{ onInvalidList: 'throw' }` for the same variable. This helper should match:

```ts
const getOptionalEnvironmentList = (key: string): string[] | null => {
  if (process.env[key] === undefined) return null
  return getEnvironmentList(key, { onInvalidList: 'throw' })
}
```

A test that passes `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=not-json` and expects an exception (or confirms `null` is returned) would prevent this regressing.

---

lib/utils/securityHeaders.ts
```diff
@@ -108,10 +118,11 @@ export const getContentSecurityPolicy = () => {
...
   )
   const allowMapboxSources = hasPublicMapboxAccessToken(fitnessStorage)
-  const configuredImageSources = allowMediaDomains.flatMap((source) => {
-    const cspSource = getCspSource(source)
-    return cspSource ? [cspSource] : []
-  })
+  const serviceMediaSources = getConfiguredCspSources(allowMediaDomains)
+  const remoteMediaSources =
+    allowRemoteMediaDomains === null
+      ? ['https:']
```
llun: **Undocumented footgun: empty array blocks all federated media**

When `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=[]`, `allowRemoteMediaDomains` is `[]` (not `null`), so `remoteMediaSources` becomes `[]`. That means neither `https:` nor any origin ends up in `img-src`/`media-src`, silently breaking every federated avatar, emoji, and status image.

This may be intentional (let an operator lock down remote media entirely), but it's a surprising and hard-to-debug outcome. Consider at minimum:
1. Adding a note in `.env.example` that an empty array disables all remote media.
2. Adding a test in `next.config.test.ts` that sets `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=[]` and asserts `img-src` contains no `https:` and no remote origins (to make the behaviour explicit and catch future regressions).
3. Or, if the intent is "unset = broad HTTPS, any array = specific origins", document it in `docs/environment-variables.md`.

---

lib/config/index.ts
```diff
@@ -96,6 +97,16 @@ const getLanguagesConfig = () => {
...
 }
 
+const getAllowRemoteMediaDomainsConfig = () => {
+  if (process.env.ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS === undefined) {
+    return undefined
+  }
+
+  return getEnvironmentList('ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS', {
+    onInvalidList: 'throw'
+  })
```
llun: **Dead code: `getAllowRemoteMediaDomainsConfig` and the schema field are never consumed**

The CSP generation path reads the env var directly through `getSecurityHeaderConfig()` in `lib/config/securityHeaders.ts`, which calls `getOptionalEnvironmentList` itself. The value populated into the Zod `Config` object here is never read by anything downstream.

This mirrors the existing duplication for `allowMediaDomains` (also populated in `Config` but read again directly by `getSecurityHeaderConfig`), so it's a pre-existing pattern — but the PR adds more of the same dead code rather than cleaning it up. At minimum, either:
- Remove `allowRemoteMediaDomains` from the Zod schema and delete `getAllowRemoteMediaDomainsConfig` (the CSP source of truth is `securityHeaders.ts`), or
- Wire `getConfig().allowRemoteMediaDomains` into the CSP path so there's a single read point, and remove the direct env read in `securityHeaders.ts`.

As written, the two reads can diverge (e.g., `index.ts` uses `onInvalidList: 'throw'` while `securityHeaders.ts` uses `onInvalidList: 'empty'`), which is exactly the inconsistency flagged in the `securityHeaders.ts` comment above.

---

next.config.test.ts
```diff
@@ -302,15 +323,55 @@ describe('next config security hardening', () => {
             'https://cdn.example.com'
```
llun, line 327: **Weak `img-src` assertion leaves a regression gap**

The only `imageSources` check in this test block is:
```ts
expect(imageSources).toContain('https:')
```

This doesn't verify that the configured service domains (`'https://images.example.com'`, `'https://cdn.example.com'`) actually appear in `img-src`. A regression that drops service media domains from `img-src` while keeping `https:` would pass this test. The `mediaSources` assertion below already checks for both sources; `imageSources` should be covered the same way:

```ts
expect(imageSources).toEqual(
  expect.arrayContaining([
    "'self'",
    'data:',
    'blob:',
    'https:',
    'https://images.example.com',
    'https://cdn.example.com'
  ])
)
```

---

llun (commented):
## Review summary

The split between service-owned (`ACTIVITIES_ALLOW_MEDIA_DOMAINS`) and federated-remote (`ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS`) CSP controls is a sensible design. The implementation is mostly clean, but there are four issues to address before merging:

### Required fixes

1. **Silent JSON-parse failure in `getOptionalEnvironmentList`** *(see inline comment on `lib/config/securityHeaders.ts:22`)*  
   The helper calls `getEnvironmentList(key)` without `{ onInvalidList: 'throw' }`. A typo in `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` produces `[]`, which sets `remoteMediaSources = []` and silently blocks all federated media with no error. `getAllowRemoteMediaDomainsConfig` in `lib/config/index.ts` already uses `onInvalidList: 'throw'` for the same variable — this helper must match.

2. **`ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=[]` is an undocumented footgun** *(see inline on `lib/utils/securityHeaders.ts:124`)*  
   An empty array produces `remoteMediaSources = []`, effectively nuking all federated image and media load. The intent may be "let operators lock everything down", but that's currently invisible in the docs and not covered by a CSP-output test. Add either documentation or an explicit test (or both).

3. **Weak `img-src` assertion in the service-domain-only test** *(see inline on `next.config.test.ts:323`)*  
   Only `toContain('https:')` is checked; the configured service domains are not verified in `img-src`, leaving a regression gap.

### Suggested cleanup

4. **Dead config field** *(see inline on `lib/config/index.ts:97–107`)*  
   `getAllowRemoteMediaDomainsConfig` and `allowRemoteMediaDomains` in the Zod schema are never consumed — the CSP reads the env var again directly via `getSecurityHeaderConfig`. This also hides the `onInvalidList` divergence. Consider removing the dead code or consolidating the two read paths.

---

### Commit prefix note

`ACTIVITIES_ALLOW_MEDIA_DOMAINS` used to replace `https:` in `img-src`/`media-src` when set; it is now purely additive and never removes `https:`. Any operator who relied on the old narrowing behaviour must migrate to `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS`. Per `AGENTS.md`, that is a `minor:` bump (new backwards-compatible feature that changes existing env-var semantics). The PR title should be updated from `fix:` to `minor:` so the CI version-bump workflow picks up the correct bump level.

---

lib/config/securityHeaders.ts
llun: Fixed in `0cad35a0`. I implemented the suggested helper behavior by using `onInvalidList: 'throw'` when the remote media env var is present, and added a malformed-env regression test.

llun:
Addressed the review summary in `0cad35a0`:

- malformed `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` now fails fast instead of silently becoming an empty remote-media allowlist
- explicit `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS=[]` is documented as blocking federated remote media and covered by a CSP-output test
- service media domain assertions now cover `img-src` domains directly
- removed the unused central `Config.allowRemoteMediaDomains` field/helper so the CSP-specific runtime config stays the source of truth
- updated the PR title to `minor: split local and remote media CSP controls`

Validation passed locally with Node 24:
- `yarn build`
- `yarn test`

llun (commented):
## Round 2 — all previous issues resolved ✓

Going through each item from round 1:

**1. Silent JSON-parse failure in `getOptionalEnvironmentList`** — Fixed. `getEnvironmentList(key, { onInvalidList: 'throw' })` is now used, and a dedicated test (`'throws for malformed runtime environment remote media allowlist'`) covers the path.

**2. Empty-array footgun undocumented** — Fixed. `.env.example` now reads "set [] to block all remote media sources", `docs/environment-variables.md` matches, and `next.config.test.ts` adds `'documents that an explicit empty remote media allowlist blocks federated media sources'` which pins both `img-src` and `media-src` output for that case.

**3. Weak `img-src` assertion in the service-domain test** — Fixed. The `imageSources` assertion in `'adds configured service media domains without narrowing remote media sources'` now uses `expect.arrayContaining` covering `'https:'`, `'https://images.example.com'`, and `'https://cdn.example.com'`. The stale `not.toContain('https:')` guards for both `imageSources` and `mediaSources` were also correctly removed.

**4. Dead config field in `lib/config/index.ts`** — Fixed by reverting the `index.ts` changes entirely; `allowRemoteMediaDomains` never appears in the Zod schema and `getAllowRemoteMediaDomainsConfig` was not added. The single read point in `securityHeaders.ts` is the source of truth.

**5. PR title prefix** — Changed from `fix:` to `minor:`. ✓

No new issues found in the update. The implementation, tests, and documentation are consistent. LGTM.

---

lib/config/securityHeaders.ts
```diff
@@ -11,10 +11,17 @@ export type SecurityHeaderStorageConfig = {
...
   allowMediaDomains: string[]
+  allowRemoteMediaDomains: string[] | null
   mediaStorage: SecurityHeaderStorageConfig
   fitnessStorage: SecurityHeaderStorageConfig
 }
 
+const getOptionalEnvironmentList = (key: string): string[] | null => {
+  if (process.env[key] === undefined) return null
+
+  return getEnvironmentList(key, { onInvalidList: 'throw' })
```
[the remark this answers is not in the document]
llun: Fixed in `7fc9f9f8`. Blank and whitespace-only `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` values are now treated as unset before parsing, so the default broad `https:` remote media sources remain. Explicit `[]` still intentionally blocks remote media, and malformed non-blank values still throw. Added config-level and CSP-output regressions plus docs wording for unset/blank.

lib/utils/securityHeaders.ts
```diff
@@ -108,10 +118,11 @@ export const getContentSecurityPolicy = () => {
...
+      : getConfiguredCspSources(allowRemoteMediaDomains)
```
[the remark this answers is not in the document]
llun: Fixed in `dbd65c10`. A non-empty remote media allowlist that normalizes to no valid CSP sources now falls back to the default `https:` remote media source instead of silently behaving like explicit `[]`. Explicit `[]` still blocks remote media, and a mixed list with at least one valid source still narrows to the valid origin without restoring broad `https:`. Added regression tests for both cases.

PR #715, opened 2026-05-17
fix: align mobile bookmark action

llun:
## Summary

Fixes a mobile timeline alignment issue where the bookmark action in the first action row sat above the reply, repost, and like buttons. On narrow viewports this made the first row look uneven, especially on posts that also show owner-only controls in a second grid row.

The root cause was that `BookmarkButton` always rendered an empty reserved error-message spacer below the icon. The post action wrapper centers grid items by their full item box, so the bookmark grid item was taller than the other action buttons even when there was no error to show.

This change removes the idle spacer from normal layout. The bookmark action now uses the same inline-flex footprint as the other action controls, and it only renders the error message when an actual bookmark request fails. When an error is shown, it is positioned below the button without changing the action-row height.

## Validation

- `source ~/.nvm/nvm.sh && nvm use 24 >/dev/null && yarn run prettier --write .`
- `source ~/.nvm/nvm.sh && nvm use 24 >/dev/null && yarn lint` — exits 0; existing unrelated warnings remain in auth/fitness files.
- `source ~/.nvm/nvm.sh && nvm use 24 >/dev/null && yarn build`
- `source ~/.nvm/nvm.sh && nvm use 24 >/dev/null && yarn test` — after rebuilding `better-sqlite3` under Node 24, 308 suites / 2,486 tests pass.
- Browser verification at a 590x1280 mobile viewport on the worktree dev server confirmed the primary action buttons share the same center Y and height, with secondary owner controls aligned under the first three grid columns.

### DIFF
--- a/lib/components/posts/actions/bookmark-button.test.tsx
+++ b/lib/components/posts/actions/bookmark-button.test.tsx
@@ -63,7 +63,7 @@ describe('BookmarkButton', () => {
     render(<BookmarkButton status={status} />)
 
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
-    expect(screen.queryByTestId('bookmark-error-space')).not.toBeInTheDocument()
+    expect(screen.queryByTestId('bookmark-error')).not.toBeInTheDocument()
   })
 
   it('shows an error and keeps state when bookmarking fails', async () => {
--- a/lib/components/posts/actions/bookmark-button.test.tsx
+++ b/lib/components/posts/actions/bookmark-button.test.tsx
@@ -59,11 +59,11 @@ describe('BookmarkButton', () => {
     jest.clearAllMocks()
   })
 
-  it('reserves error message space without announcing an empty alert', () => {
+  it('does not reserve idle error message space in the action grid', () => {
     render(<BookmarkButton status={status} />)
 
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
-    expect(screen.getByTestId('bookmark-error-space')).toBeEmptyDOMElement()
+    expect(screen.queryByTestId('bookmark-error-space')).not.toBeInTheDocument()
   })
 
   it('shows an error and keeps state when bookmarking fails', async () => {
--- a/lib/components/posts/actions/bookmark-button.test.tsx
+++ b/lib/components/posts/actions/bookmark-button.test.tsx
@@ -2,7 +2,7 @@
  * @jest-environment jsdom
  */
 import '@testing-library/jest-dom'
-import { fireEvent, render, screen } from '@testing-library/react'
+import { act, fireEvent, render, screen } from '@testing-library/react'
 
 import { bookmarkStatus, undoBookmarkStatus } from '@/lib/client'
 import { StatusNote, StatusType } from '@/lib/types/domain/status'
@@ -116,5 +116,28 @@ describe('BookmarkButton', () => {
       await screen.findByText('Failed to bookmark post. Please try again.')
     ).toHaveAttribute('role', 'alert')
     expect(screen.getByRole('button', { name: 'Bookmark' })).toBeEnabled()
+  })
+
+  it('auto-dismisses bookmark errors after a short delay', async () => {
+    jest.useFakeTimers()
+    ;(bookmarkStatus as jest.Mock).mockResolvedValue(false)
+
+    try {
+      render(<BookmarkButton status={status} />)
+
+      fireEvent.click(screen.getByRole('button', { name: 'Bookmark' }))
+
+      expect(await screen.findByRole('alert')).toHaveTextContent(
+        'Failed to bookmark post. Please try again.'
+      )
+
+      act(() => {
+        jest.advanceTimersByTime(4000)
+      })
+
+      expect(screen.queryByRole('alert')).not.toBeInTheDocument()
+    } finally {
+      jest.useRealTimers()
+    }
   })
 })
--- a/lib/components/posts/actions/bookmark-button.test.tsx
+++ b/lib/components/posts/actions/bookmark-button.test.tsx
@@ -120,14 +120,23 @@ describe('BookmarkButton', () => {
 
   it('auto-dismisses bookmark errors after a short delay', async () => {
     jest.useFakeTimers()
-    ;(bookmarkStatus as jest.Mock).mockResolvedValue(false)
+    let resolveBookmark: (value: boolean) => void = () => {}
+    const bookmarkPromise = new Promise<boolean>((resolve) => {
+      resolveBookmark = resolve
+    })
+    ;(bookmarkStatus as jest.Mock).mockReturnValue(bookmarkPromise)
 
     try {
       render(<BookmarkButton status={status} />)
 
       fireEvent.click(screen.getByRole('button', { name: 'Bookmark' }))
 
-      expect(await screen.findByRole('alert')).toHaveTextContent(
+      await act(async () => {
+        resolveBookmark(false)
+        await bookmarkPromise
+      })
+
+      expect(screen.getByRole('alert')).toHaveTextContent(
         'Failed to bookmark post. Please try again.'
       )
 
--- a/lib/components/posts/actions/bookmark-button.tsx
+++ b/lib/components/posts/actions/bookmark-button.tsx
@@ -34,7 +34,7 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
     : 'Failed to bookmark post. Please try again.'
 
   return (
-    <span className="inline-flex flex-col items-center gap-1">
+    <span className="relative inline-flex items-center justify-center">
       <button
         title={bookmarkLabel}
         aria-label={bookmarkLabel}
@@ -71,14 +71,15 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
       >
         <Bookmark className={cn('h-4 w-4', { 'fill-current': isBookmarked })} />
       </button>
-      <span
-        aria-hidden={error ? undefined : true}
-        className="min-h-4 text-xs text-destructive"
-        data-testid="bookmark-error-space"
-        role={error ? 'alert' : undefined}
-      >
-        {error}
-      </span>
+      {error ? (
+        <span
+          className="absolute left-1/2 top-full z-10 mt-1 w-48 -translate-x-1/2 rounded-md border bg-background px-2 py-1 text-center text-xs text-destructive shadow-sm"
+          data-testid="bookmark-error"
+          role="alert"
+        >
+          {error}
+        </span>
+      ) : null}
     </span>
   )
 }
--- a/lib/components/posts/actions/bookmark-button.tsx
+++ b/lib/components/posts/actions/bookmark-button.tsx
@@ -13,6 +13,8 @@ interface BookmarkButtonProps {
   ) => void
 }
 
+const BOOKMARK_ERROR_DISMISS_MS = 4000
+
 export const BookmarkButton: FC<BookmarkButtonProps> = ({
   status,
   onBookmarkChanged
@@ -27,6 +29,16 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
     setIsBookmarked(status.isActorBookmarked)
     setError(null)
   }, [status.isActorBookmarked])
+
+  useEffect(() => {
+    if (!error) return
+
+    const timeoutId = window.setTimeout(() => {
+      setError(null)
+    }, BOOKMARK_ERROR_DISMISS_MS)
+
+    return () => window.clearTimeout(timeoutId)
+  }, [error])
 
   const bookmarkLabel = isBookmarked ? 'Remove bookmark' : 'Bookmark'
   const failureMessage = isBookmarked
@@ -73,7 +85,7 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
       </button>
       {error ? (
         <span
-          className="absolute left-1/2 top-full z-10 mt-1 w-48 -translate-x-1/2 rounded-md border bg-background px-2 py-1 text-center text-xs text-destructive shadow-sm"
+          className="pointer-events-none absolute right-0 top-full z-10 mt-1 w-max max-w-[min(12rem,calc(100vw-2rem))] break-words rounded-md border bg-background px-2 py-1 text-left text-xs text-destructive shadow-sm"
           data-testid="bookmark-error"
           role="alert"
         >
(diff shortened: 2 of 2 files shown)


lib/components/posts/actions/bookmark-button.test.tsx
```diff
@@ -59,11 +59,11 @@ describe('BookmarkButton', () => {
     jest.clearAllMocks()
   })
 
-  it('reserves error message space without announcing an empty alert', () => {
+  it('does not reserve idle error message space in the action grid', () => {
     render(<BookmarkButton status={status} />)
 
     expect(screen.queryByRole('alert')).not.toBeInTheDocument()
-    expect(screen.getByTestId('bookmark-error-space')).toBeEmptyDOMElement()
+    expect(screen.queryByTestId('bookmark-error-space')).not.toBeInTheDocument()
```
llun: This assertion uses the **old** `data-testid` (`bookmark-error-space`) that was removed in this PR. The test passes today, but for the wrong reason — it passes because no element with that ID ever existed after the refactor, not because the new idle state is correctly rendered without an error element.

The idle-state test should assert the **new** ID is absent:

```tsx
expect(screen.queryByTestId('bookmark-error')).not.toBeInTheDocument()
```

Without this fix the test gives a false sense of coverage: a future regression that accidentally renders `bookmark-error` in the idle state would not be caught.

---

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -71,14 +71,15 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
...
-        aria-hidden={error ? undefined : true}
-        className="min-h-4 text-xs text-destructive"
-        data-testid="bookmark-error-space"
-        role={error ? 'alert' : undefined}
-      >
-        {error}
-      </span>
+      {error ? (
+        <span
+          className="absolute left-1/2 top-full z-10 mt-1 w-48 -translate-x-1/2 rounded-md border bg-background px-2 py-1 text-center text-xs text-destructive shadow-sm"
```
llun: `w-48` (192 px) is a fixed width applied to an absolutely-positioned element whose anchor point (`left-1/2 -translate-x-1/2`) is the center of the bookmark button. On a 320 px viewport the button can sit anywhere inside the grid, so the tooltip can easily overflow the screen on either side, and `overflow: hidden` on any ancestor would clip it silently.

A safer approach:
- Drop the fixed width and use `w-max max-w-[12rem]` so the tooltip only grows as wide as the text needs.
- Or keep the fixed width but add a `clamp` with a CSS custom property / inline style that respects the viewport: `style={{ left: 'clamp(0px, 50%, calc(100vw - 12rem))' }}`.

Also worth noting: no parent has `overflow: visible` forced on it, so if the `<div>` wrapping the action row has `overflow: hidden` the tooltip will be clipped. Double-check that the action-row containers in `actions.tsx` (and any card/post wrappers further up) don't implicitly clip it.

---

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -71,14 +71,15 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
...
       </button>
-      <span
-        aria-hidden={error ? undefined : true}
-        className="min-h-4 text-xs text-destructive"
-        data-testid="bookmark-error-space"
-        role={error ? 'alert' : undefined}
-      >
-        {error}
-      </span>
+      {error ? (
```
llun, line 86: The error tooltip has no dismissal mechanism. Once shown it persists until the user clicks the button a second time. On a narrow mobile viewport this absolutely-positioned element can overlap the reply, repost, and like buttons in the same action row, making them hard to tap.

Two options:
1. **Auto-dismiss** — clear the error after a short delay (e.g. `useEffect(() => { if (error) { const t = setTimeout(() => setError(null), 4000); return () => clearTimeout(t); } }, [error])`).
2. **Click-to-dismiss** — add `onClick={() => setError(null)}` to the error `<span>` and show a close affordance.

Either way, add a test for the chosen behaviour so the UX contract is enforced.

---

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -71,14 +71,15 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
...
+        <span
+          className="absolute left-1/2 top-full z-10 mt-1 w-48 -translate-x-1/2 rounded-md border bg-background px-2 py-1 text-center text-xs text-destructive shadow-sm"
+          data-testid="bookmark-error"
+          role="alert"
+        >
+          {error}
+        </span>
+      ) : null}
     </span>
```
llun, line 95: The outer `<span>` is used purely as a `position: relative` anchor for the tooltip. A `<span>` is valid here since neither it nor the `<button>` inside it are block-level — but be aware that `<span>` is `display: inline` by default; the `inline-flex` class overrides this, which is fine.

One subtle concern: the grid cell in `actions.tsx` (`justify-items-center`) will size the cell to fit this `inline-flex` span. The span itself has no height beyond the button (correct — that was the whole point of this PR), but the absolutely-positioned tooltip child does *not* contribute to the span's height. Confirm in a real device/browser that the tooltip doesn't get cut off by the bottom of the grid area when secondary owner controls appear in the row below (the `overflow` behaviour on grid rows is UA-dependent).

---

llun (commented):
## Overview

The fix is conceptually correct — replacing the always-present empty spacer with a conditionally rendered, absolutely-positioned tooltip is the right approach for keeping all action buttons at the same height in the mobile grid.

There are four issues worth addressing before merging:

---

### 1. (Blocking) Test assertion uses the removed `data-testid`

`bookmark-button.test.tsx` line 66 still queries for `bookmark-error-space`, which no longer exists. The test passes for the wrong reason, meaning a regression that accidentally renders `bookmark-error` in the idle state would go undetected. See inline comment.

### 2. (Blocking) Fixed-width tooltip can overflow narrow viewports

`w-48` (192 px) centered on the button anchor point with `-translate-x-1/2` will overflow the viewport edges on narrow phones (320–360 px). Use `w-max max-w-[12rem]` or a viewport-aware clamp. See inline comment.

### 3. (Non-blocking, UX concern) No way to dismiss the error

The tooltip persists until the user retries. On mobile it can cover adjacent action buttons. An auto-dismiss timer or click-to-dismiss would improve the experience. See inline comment.

### 4. (Non-blocking, verification needed) Tooltip clipping by grid/card ancestors

The tooltip escapes its parent's height via `position: absolute`, but any `overflow: hidden` ancestor will clip it silently. Verify in a real browser that no card or grid wrapper clips the tooltip, especially when the secondary-action row is present below.

---

lib/components/posts/actions/bookmark-button.test.tsx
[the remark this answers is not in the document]
llun: Fixed in 9fc14f13. The idle-state test now asserts that the current `bookmark-error` element is absent instead of checking the removed `bookmark-error-space` id.

lib/components/posts/actions/bookmark-button.tsx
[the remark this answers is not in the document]
llun: Fixed in a85866ae for the overflow risk. The alert remains a sibling `span` next to the button, not nested inside the button, so the block-level-in-button concern does not apply here. The fixed centered tooltip is now right-aligned, viewport-clamped, wrapping, `pointer-events-none`, and auto-dismissed.

lib/components/posts/actions/bookmark-button.test.tsx
llun: Covered by 9fc14f13. The idle-state test now queries `bookmark-error` instead of the removed `bookmark-error-space` id, so it will catch an accidental idle alert render.

lib/components/posts/actions/bookmark-button.tsx
llun: Fixed in a85866ae. The tooltip no longer uses fixed centered `w-48`; it is right-aligned to the button and uses `w-max max-w-[min(12rem,calc(100vw-2rem))]` so it wraps within the mobile viewport. I also browser-verified 320px and 390px mobile widths against the local timeline.

lib/components/posts/actions/bookmark-button.tsx
llun: Fixed in a85866ae. Bookmark errors now auto-dismiss after 4 seconds with timeout cleanup, and the behavior is covered by the fake-timer test in `bookmark-button.test.tsx`.

lib/components/posts/actions/bookmark-button.tsx
llun: Verified after a85866ae in a real browser against the local timeline at 320px and 390px, including owner rows with secondary controls. The alert stayed within the viewport and the overflow-hidden card bounds, auto-dismissed, and `pointer-events-none` kept the secondary controls hit-testable while the alert was visible.

lib/components/posts/actions/bookmark-button.tsx
```diff
@@ -28,13 +30,23 @@ export const BookmarkButton: FC<BookmarkButtonProps> = ({
     setError(null)
   }, [status.isActorBookmarked])
 
+  useEffect(() => {
+    if (!error) return
+
+    const timeoutId = window.setTimeout(() => {
+      setError(null)
+    }, BOOKMARK_ERROR_DISMISS_MS)
```
llun, line 38: `window.setTimeout` / `window.clearTimeout` work correctly here (jsdom patches both), but the idiomatic React pattern is just `setTimeout` / `clearTimeout` — the global is always the browser's `window.setTimeout` inside a `useEffect` and there's no ambiguity. Minor nit but worth aligning with the rest of the codebase.

---

lib/components/posts/actions/bookmark-button.test.tsx
```diff
@@ -117,4 +117,27 @@ describe('BookmarkButton', () => {
...
+      act(() => {
+        jest.advanceTimersByTime(4000)
+      })
+
+      expect(screen.queryByRole('alert')).not.toBeInTheDocument()
+    } finally {
+      jest.useRealTimers()
+    }
+  })
 })
```
llun, line 152: `jest.useFakeTimers()` is called before `await screen.findByRole('alert')`. `findByRole` uses `waitFor` internally, which polls via `setInterval`. With fake timers active that interval never ticks unless you advance time — so if the React state update that sets the error doesn't land before `findByRole`'s first synchronous check, the query hangs until it times out on its own real wall-clock deadline.

In practice this works today because `mockResolvedValue` causes the Promise to resolve as a microtask (not a timer), and React 18's scheduler uses `MessageChannel` (also not affected by fake timers) so the re-render lands before `waitFor` needs to retry. But it's a fragile ordering assumption.

A more resilient pattern is to activate fake timers only after the error is visible:

```tsx
it('auto-dismisses bookmark errors after a short delay', async () => {
  ;(bookmarkStatus as jest.Mock).mockResolvedValue(false)

  render(<BookmarkButton status={status} />)
  fireEvent.click(screen.getByRole('button', { name: 'Bookmark' }))

  // find the alert with real timers so waitFor's polling works unconditionally
  expect(await screen.findByRole('alert')).toHaveTextContent(
    'Failed to bookmark post. Please try again.'
  )

  // only now freeze time to test the dismiss
  jest.useFakeTimers()
  try {
    act(() => {
      jest.advanceTimersByTime(4000)
    })
    expect(screen.queryByRole('alert')).not.toBeInTheDocument()
  } finally {
    jest.useRealTimers()
  }
})
```

This separates "wait for async state update" (needs real timers) from "advance the dismiss timer" (needs fake timers).

---

llun (commented):
## Round 2 review

All six threads from round 1 are resolved. Both blocking issues are properly fixed, and the non-blocking UX concerns (auto-dismiss, `pointer-events-none`, viewport-clamped width, browser verification) are addressed. The core fix is solid.

Two minor points remain:

---

### 1. (Nit) `window.setTimeout` → `setTimeout`

The auto-dismiss `useEffect` calls `window.setTimeout` / `window.clearTimeout`. These work correctly in jsdom, but the idiomatic React pattern is just the unqualified `setTimeout` / `clearTimeout`. See inline comment.

### 2. (Non-blocking) Fake-timer test ordering is fragile

The auto-dismiss test enables fake timers before `await screen.findByRole('alert')`. This works today because the mock resolves as a microtask and React 18's scheduler uses `MessageChannel`, neither of which is blocked by fake timers. However, `waitFor`'s polling still relies on `setInterval`, so the ordering assumption is implicit and could break under a different Jest or testing-library version. Moving `jest.useFakeTimers()` to after the `findByRole` await removes the dependency. See inline comment for a suggested rewrite.

---

lib/components/posts/actions/bookmark-button.tsx
llun: Fixed in bc2633d9. The dismiss effect now uses unqualified `setTimeout` / `clearTimeout`; behavior and cleanup are unchanged.

lib/components/posts/actions/bookmark-button.test.tsx
llun: Fixed in bc2633d9. I rewrote the test to avoid `findByRole` while fake timers are active: the mocked bookmark request is manually resolved inside `act`, then the dismiss timeout is advanced. This removes the dependency on Testing Library polling under fake timers while still keeping fake timers active before the component schedules the dismiss timeout.

llun:
@codex review

llun (commented):
## Round 3 review — LGTM

All eight threads from the previous two rounds are resolved. Both round 2 issues are fully addressed:

- **`window.setTimeout` nit**: fixed — `bookmark-button.tsx` now uses plain `setTimeout` / `clearTimeout`.
- **Fake-timer test brittleness**: fixed with a better approach than the one I suggested. Keeping `jest.useFakeTimers()` active from the start and resolving the bookmark Promise inside `await act(async () => { ... })` is actually cleaner: it ensures the dismiss `setTimeout` is registered under fake timers from the moment the error state lands, eliminating any race between real-clock scheduling and the timer advance. The follow-up synchronous `getByRole` (not `findByRole`) confirms there's no dependency on `waitFor` polling under fake timers.

Every concern raised across all three rounds is now resolved:

| Issue | Status |
|---|---|
| Idle-state test used removed `bookmark-error-space` testid | ✅ Fixed |
| Fixed-width tooltip could overflow narrow viewports | ✅ Fixed (`w-max max-w-[min(12rem,calc(100vw-2rem))]`) |
| No error dismissal mechanism | ✅ Fixed (4 s auto-dismiss with cleanup) |
| Tooltip could block adjacent buttons while visible | ✅ Fixed (`pointer-events-none`) |
| Grid clipping verified in real browser | ✅ Verified at 320 px and 390 px |
| `window.setTimeout` / `window.clearTimeout` style nit | ✅ Fixed |
| `findByRole` under fake timers was fragile | ✅ Fixed (manual Promise + `act`) |

The fix is correct, the tests are robust, and the implementation is clean. Ready to merge.

---

PR #717, opened 2026-05-17
minor: implement reblogged_by status endpoint

llun:
## Summary
- Implement Mastodon-compatible `GET /api/v1/statuses/:id/reblogged_by`
- Add database lookup for accounts that boosted a status with `limit`, `max_id`, and `since_id` cursor pagination
- Preserve visibility rules so anonymous clients only see public boost activities

## Validation
- `yarn test --runTestsByPath 'app/api/v1/statuses/[id]/route.test.ts'`
- `yarn lint`
- `yarn build`
- `yarn test`

### DIFF
--- a/app/api/v1/statuses/[id]/reblogged_by/route.ts
+++ b/app/api/v1/statuses/[id]/reblogged_by/route.ts
@@ -31,11 +31,13 @@ const RebloggedByQueryParams = z.object({
 const getPaginationLinkHeader = ({
   req,
   limit,
-  reblogs
+  reblogs,
+  hasNextPage
 }: {
   req: Request
   limit: number
   reblogs: RebloggedByAccount[]
+  hasNextPage: boolean
 }) => {
   if (reblogs.length === 0) return undefined
 
@@ -55,10 +57,9 @@ const getPaginationLinkHeader = ({
 
   const firstReblog = reblogs[0]
   const lastReblog = reblogs[reblogs.length - 1]
-  const nextLink =
-    reblogs.length >= limit
-      ? `<${buildUrl('max_id', lastReblog.statusId)}>; rel="next"`
-      : null
+  const nextLink = hasNextPage
+    ? `<${buildUrl('max_id', lastReblog.statusId)}>; rel="next"`
+    : null
   const prevLink = `<${buildUrl('since_id', firstReblog.statusId)}>; rel="prev"`
   return [nextLink, prevLink].filter(Boolean).join(', ')
 }
@@ -103,15 +104,22 @@ export const GET = traceApiRoute(
         responseStatusCode: 404
       })
 
-    const reblogs = await database.getRebloggedBy({
+    const reblogsPage = await database.getRebloggedBy({
       statusId,
-      limit,
+      limit: limit + 1,
       maxStatusId: maxId ? idToUrl(maxId) : undefined,
       sinceStatusId: sinceId ? idToUrl(sinceId) : undefined,
       visibleToActorId: currentActor?.id ?? null
     })
+    const hasNextPage = reblogsPage.length > limit
+    const reblogs = reblogsPage.slice(0, limit)
 
-    const paginationLink = getPaginationLinkHeader({ req, limit, reblogs })
+    const paginationLink = getPaginationLinkHeader({
+      req,
+      limit,
+      reblogs,
+      hasNextPage
+    })
 
     const accounts = await database.getMastodonActorsFromIds({
       ids: reblogs.map(({ actorId }) => actorId)
--- a/app/api/v1/statuses/[id]/reblogged_by/route.ts
+++ b/app/api/v1/statuses/[id]/reblogged_by/route.ts
@@ -111,15 +111,9 @@ export const GET = traceApiRoute(
 
     const paginationLink = getPaginationLinkHeader({ req, limit, reblogs })
 
-    const accounts = (
-      await Promise.all(
-        reblogs.map(({ actorId }) =>
-          database.getMastodonActorFromId({ id: actorId })
-        )
-      )
-    ).filter((account): account is NonNullable<typeof account> =>
-      Boolean(account)
-    )
+    const accounts = await database.getMastodonActorsFromIds({
+      ids: reblogs.map(({ actorId }) => actorId)
+    })
 
     return apiResponse({
       req,
--- a/app/api/v1/statuses/[id]/reblogged_by/route.ts
+++ b/app/api/v1/statuses/[id]/reblogged_by/route.ts
@@ -1,10 +1,17 @@
-import { OAuthGuard } from '@/lib/services/guards/OAuthGuard'
+import { z } from 'zod'
+
+import { OptionalOAuthGuard } from '@/lib/services/guards/OAuthGuard'
 import { getReadableStatus } from '@/lib/services/statusRouteAccess'
-import { Scope } from '@/lib/types/database/operations'
+import { type RebloggedByAccount, Scope } from '@/lib/types/database/operations'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
-import { ERROR_404, apiResponse, defaultOptions } from '@/lib/utils/response'
+import {
+  ERROR_400,
+  ERROR_404,
+  apiResponse,
+  defaultOptions
+} from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
-import { idToUrl } from '@/lib/utils/urlToId'
+import { idToUrl, urlToId } from '@/lib/utils/urlToId'
 
 const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
 
@@ -14,9 +21,45 @@ interface Params {
   id: string
 }
 
+const RebloggedByQueryParams = z.object({
+  limit: z.coerce.number().int().min(1).max(80).default(40),
+  max_id: z.string().min(1).optional(),
+  since_id: z.string().min(1).optional()
+})
+
+const getPaginationLinkHeader = ({
+  req,
+  limit,
+  reblogs
+}: {
+  req: Request
+  limit: number
+  reblogs: RebloggedByAccount[]
+}) => {
+  if (reblogs.length === 0) return undefined
+
+  const requestUrl = new URL(req.url)
+  const buildUrl = (cursor: 'max_id' | 'since_id', statusId: string) => {
+    const params = new URLSearchParams()
+    params.set('limit', `${limit}`)
+    params.set(cursor, urlToId(statusId))
+
+    const url = new URL(requestUrl.pathname, requestUrl.origin)
+    url.search = params.toString()
+    return url.toString()
+  }
+
+  const firstReblog = reblogs[0]
+  const lastReblog = reblogs[reblogs.length - 1]
+  return [
+    `<${buildUrl('max_id', lastReblog.statusId)}>; rel="next"`,
+    `<${buildUrl('since_id', firstReblog.statusId)}>; rel="prev"`
+  ].join(', ')
+}
+
 export const GET = traceApiRoute(
   'getStatusRebloggedBy',
-  OAuthGuard<Params>([Scope.enum.read], async (req, context) => {
+  OptionalOAuthGuard<Params>([Scope.enum.read], async (req, context) => {
     const { database, currentActor, params } = context
     const encodedStatusId = (await params).id
     if (!encodedStatusId)
@@ -27,6 +70,18 @@ export const GET = traceApiRoute(
         responseStatusCode: 404
       })
 
+    const queryParams = Object.fromEntries(new URL(req.url).searchParams)
+    const parsedParams = RebloggedByQueryParams.safeParse(queryParams)
+    if (!parsedParams.success) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+
+    const { limit, max_id: maxId, since_id: sinceId } = parsedParams.data
     const statusId = idToUrl(encodedStatusId)
     const status = await getReadableStatus({
       database,
@@ -42,10 +97,32 @@ export const GET = traceApiRoute(
         responseStatusCode: 404
       })
 
-    // getRebloggedBy not yet implemented - return empty array
-    // TODO: Implement database method to get actors who reblogged this status
+    const reblogs = await database.getRebloggedBy({
+      statusId,
+      limit,
+      maxStatusId: maxId ? idToUrl(maxId) : undefined,
+      sinceStatusId: sinceId ? idToUrl(sinceId) : undefined,
+      visibleToActorId: currentActor?.id ?? null
+    })
 
-    return apiResponse({ req, allowedMethods: CORS_HEADERS, data: [] })
+    const paginationLink = getPaginationLinkHeader({ req, limit, reblogs })
+
+    const accounts = (
+      await Promise.all(
+        reblogs.map(({ actor }) =>
+          database.getMastodonActorFromId({ id: actor.id })
+        )
+      )
+    ).filter((account): account is NonNullable<typeof account> =>
+      Boolean(account)
+    )
+
+    return apiResponse({
+      req,
+      allowedMethods: CORS_HEADERS,
+      data: accounts,
+      additionalHeaders: paginationLink ? [['Link', paginationLink]] : []
+    })
   }),
   {
     addAttributes: async (_req, context) => {
--- a/app/api/v1/statuses/[id]/reblogged_by/route.ts
+++ b/app/api/v1/statuses/[id]/reblogged_by/route.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod'
 
 import { OptionalOAuthGuard } from '@/lib/services/guards/OAuthGuard'
+import { headerHost } from '@/lib/services/guards/headerHost'
 import { getReadableStatus } from '@/lib/services/statusRouteAccess'
 import { type RebloggedByAccount, Scope } from '@/lib/types/database/operations'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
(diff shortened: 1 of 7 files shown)


lib/types/database/operations.ts
```diff
@@ -470,6 +470,16 @@ export type GetFavouritedByParams = BaseStatusParams & {
...
+export type GetRebloggedByParams = BaseStatusParams & {
+  limit?: number
+  maxStatusId?: string
+  sinceStatusId?: string
+  visibleToActorId?: string | null
+}
+export type RebloggedByAccount = {
+  actor: Actor
+  statusId: string
+}
```
llun, line 486: The `actor: Actor` field is only ever used in the route to extract `actor.id` for a subsequent `getMastodonActorFromId` call, so fetching the full `Actor` object inside `getRebloggedBy` is wasted work. Change this type to:

```ts
export type RebloggedByAccount = {
  actorId: string
  statusId: string
}
```

Then in `getRebloggedBy`, return `{ actorId: item.actorId, statusId: item.id }` directly instead of calling `actorDatabase.getActorFromId` at all. This eliminates the N inner actor lookups in the DB layer and lets the route call `getMastodonActorFromId({ id: actorId })` directly.

---

lib/database/sql/status.ts
```diff
@@ -1328,6 +1329,93 @@ export const StatusSQLDatabaseMixin = (
...
+    const result = await query
+    const reblogs = await Promise.all(
+      result.map(async (item) => {
+        const actor = await actorDatabase.getActorFromId({ id: item.actorId })
+        return actor ? { actor, statusId: item.id } : null
+      })
+    )
+    return reblogs.filter((item): item is NonNullable<typeof item> =>
+      Boolean(item)
+    )
```
llun: Two problems here:

**1. N+1 actor lookups** — `getActorFromId` is called sequentially for each result row inside `Promise.all`. Although `Promise.all` runs them concurrently, there's no batch query path. This should be addressed in the same way as `getFavouritedBy` — return just the `actorId` strings and let the caller do the Mastodon hydration.

**2. Redundant lookup** — The route immediately discards the `Actor` object returned here and calls `database.getMastodonActorFromId({ id: actor.id })` for each one, doubling the number of actor round-trips. With the type change suggested on `RebloggedByAccount`, this entire block can be simplified to:

```ts
const result = await query
return result.map((item) => ({ actorId: item.actorId, statusId: item.id }))
```

---

app/api/v1/statuses/[id]/reblogged_by/route.ts
```diff
@@ -42,10 +97,32 @@ export const GET = traceApiRoute(
...
-    return apiResponse({ req, allowedMethods: CORS_HEADERS, data: [] })
+    const accounts = (
+      await Promise.all(
+        reblogs.map(({ actor }) =>
+          database.getMastodonActorFromId({ id: actor.id })
+        )
+      )
+    ).filter((account): account is NonNullable<typeof account> =>
+      Boolean(account)
+    )
```
Comment (llun) [inline app/api/v1/statuses/[id]/reblogged_by/route.ts]: With the `RebloggedByAccount` type change (dropping `actor: Actor` in favour of `actorId: string`), this block becomes a straightforward:

```ts
const accounts = (
  await Promise.all(
    reblogs.map(({ actorId }) =>
      database.getMastodonActorFromId({ id: actorId })
    )
  )
).filter((account): account is NonNullable<typeof account> => Boolean(account))
```

Note that silently dropping accounts where `getMastodonActorFromId` returns `null` can cause the returned page to have fewer entries than `limit` even when more reblogs exist. This is the same behaviour as `getFavouritedBy`, so it's acceptable for now, but worth a comment if the actor-not-found case is expected to be common (e.g. remote actors that were deleted).

---

app/api/v1/statuses/[id]/reblogged_by/route.ts
```diff
@@ -14,9 +21,45 @@ interface Params {
...
+const getPaginationLinkHeader = ({
+  req,
+  limit,
+  reblogs
+}: {
+  req: Request
+  limit: number
+  reblogs: RebloggedByAccount[]
+}) => {
+  if (reblogs.length === 0) return undefined
```
Comment (llun) [inline app/api/v1/statuses/[id]/reblogged_by/route.ts:42]: `next` and `prev` links are both emitted whenever `reblogs.length > 0`, including on the last page (when `reblogs.length < limit`). Mastodon clients use the absence of a `next` link to detect end-of-results, so sending it unconditionally causes an unnecessary extra empty request.

Omit the `next` link when the page is not full:

```ts
const firstReblog = reblogs[0]
const lastReblog = reblogs[reblogs.length - 1]
const parts: string[] = []
if (reblogs.length >= limit) {
  parts.push(`<${buildUrl('max_id', lastReblog.statusId)}>; rel="next"`)
}
parts.push(`<${buildUrl('since_id', firstReblog.statusId)}>; rel="prev"`)
return parts.join(', ')
```

Compare with how `app/api/v1/accounts/[id]/statuses/route.ts` builds its `Link` header (though that route has the same issue).

---

lib/database/sql/status.ts
```diff
@@ -1328,6 +1329,93 @@ export const StatusSQLDatabaseMixin = (
...
+        ? database('statuses')
+            .where('id', maxStatusId)
+            .first<{ id: string; createdAt: Date }>('id', 'createdAt')
+        : null,
+      sinceStatusId
+        ? database('statuses')
+            .where('id', sinceStatusId)
+            .first<{ id: string; createdAt: Date }>('id', 'createdAt')
+        : null
+    ])
```
llun, line 1430: If `maxStatusId` or `sinceStatusId` is provided but doesn't exist in the `statuses` table, `maxCursor`/`sinceCursor` will be `undefined` and the cursor filter is silently skipped — effectively restarting from page 1 without any error. A client with a stale cursor ID (e.g. after the boost was deleted) will get unexpected results with no indication that the cursor was invalid.

Consider returning an error signal (e.g. throw, or return a typed `{ error: 'invalid_cursor' }`) so the route can respond with 422 instead of silently degrading. At minimum, document this behaviour in the function signature.

---

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -1586,6 +1586,123 @@ describe('GET /api/v1/statuses/[id]', () => {
...
+      const accounts = (await response.json()) as { id: string }[]
+      expect(accounts.map((account) => account.id)).toEqual([
+        urlToId(ACTOR2_ID)
+      ])
+    })
+  })
+
   describe('media attachments', () => {
     it('includes media attachments with correct format', async () => {
       const status = (await database.getStatus({
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:2227]: Missing test cases for the new endpoint:

1. **Authenticated user sees followers-only boosts addressed to them** — the `applyPotentiallyReadableStatusFilter` path is completely untested. A logged-in actor should be able to see a boost where `to` includes their ID or the public audience, even if the same boost is invisible to an anonymous client.

2. **Invalid pagination params return 400** — e.g. `limit=0`, `limit=81`, `limit=abc`. The Zod schema validates these but no test verifies the HTTP 400 response.

3. **Non-existent status returns 404** — none of the new tests verify that a request for an unknown status ID returns 404.

4. **`next` link is absent on the last page** — once the partial-page `next` link issue is fixed, add a test that confirms no `next` link when `reblogs.length < limit`.

---

llun (commented):
## Review of `minor: implement reblogged_by status endpoint`

The overall structure is sound — `OptionalOAuthGuard` is the right guard, Zod validation is consistent with the rest of the codebase, and the two-stage visibility filter (public for anonymous, actor-aware for authenticated) is correct. The tests cover the happy path and the private-boost filter well.

Three things need to be addressed before merge:

### 1. Double actor lookup (performance — blocks merge)
`getRebloggedBy` calls `actorDatabase.getActorFromId` for every row, then the route immediately discards those `Actor` objects and calls `database.getMastodonActorFromId({ id: actor.id })` for every item again. That's 2N actor round-trips for N reblogs. Fix: change `RebloggedByAccount` to `{ actorId: string; statusId: string }` and remove the `getActorFromId` call from the DB layer entirely (see inline comments).

### 2. `next` link emitted on the last page
`getPaginationLinkHeader` returns a `next` link whenever `reblogs.length > 0`, even when `reblogs.length < limit`. Mastodon clients use the absence of `next` to stop paginating; sending it unconditionally causes a wasted empty request on the last page.

### 3. Missing test coverage
- Authenticated user path (`applyPotentiallyReadableStatusFilter`) is not tested
- Invalid `limit` values (0, 81, non-numeric) should return 400
- Non-existent status ID should return 404
- Last-page behaviour (no `next` link) should be verified

Minor notes (non-blocking): silent cursor skip when `max_id`/`since_id` is stale, and the duplicated where-condition in `reblogTargetStatusIds` vs the main query (Gemini already flagged the latter).

---

lib/database/sql/status.ts
```diff
@@ -1328,6 +1329,93 @@ export const StatusSQLDatabaseMixin = (
...
+      query = query.limit(limit)
+    }
+
+    const result = await query
+    const reblogs = await Promise.all(
+      result.map(async (item) => {
+        const actor = await actorDatabase.getActorFromId({ id: item.actorId })
+        return actor ? { actor, statusId: item.id } : null
+      })
+    )
```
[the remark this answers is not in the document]
llun: Fixed in 61f35ac9. getRebloggedBy now returns actorId/statusId directly and no longer hydrates actors per row.

app/api/v1/statuses/[id]/reblogged_by/route.ts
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/reblogged_by/route.ts]: Partially addressed in 61f35ac9 by removing the redundant DB-layer actor hydration and returning actor IDs. I did not add a separate batch Mastodon-account hydrator here because existing endpoints such as favourited_by use per-account getMastodonActorFromId, the page size is capped at 80, and adding a new batch serializer would be broader than this endpoint. This removes the 2N actor round-trips called out in review.

lib/database/sql/status.ts
```diff
@@ -1328,6 +1329,93 @@ export const StatusSQLDatabaseMixin = (
...
+    let query = database('statuses')
+      .select<{ id: string; actorId: string }[]>('id', 'actorId')
+      .where('type', StatusType.enum.Announce)
+      .where((builder) => {
+        builder.where('originalStatusId', statusId).orWhere((legacyBuilder) => {
+          legacyBuilder.whereNull('originalStatusId').where('content', statusId)
+        })
+      })
+      .orderBy('createdAt', 'desc')
+      .orderBy('id', 'desc')
```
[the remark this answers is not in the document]
llun: Fixed in 61f35ac9 by extracting the shared reblogBase query and cloning it for both the target-status subquery and the result query.

app/api/v1/statuses/[id]/reblogged_by/route.ts
```diff
@@ -14,9 +21,45 @@ interface Params {
...
+
+  const requestUrl = new URL(req.url)
+  const buildUrl = (cursor: 'max_id' | 'since_id', statusId: string) => {
+    const params = new URLSearchParams()
+    params.set('limit', `${limit}`)
+    params.set(cursor, urlToId(statusId))
+
+    const url = new URL(requestUrl.pathname, requestUrl.origin)
+    url.search = params.toString()
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/reblogged_by/route.ts:54]: Fixed in cba1e546. Link header construction now uses headerHost(req.headers).

lib/types/database/operations.ts
llun: Fixed in 61f35ac9. RebloggedByAccount is now actorId/statusId rather than carrying a full Actor.

lib/database/sql/status.ts
llun: Fixed in 61f35ac9. getRebloggedBy now returns IDs directly and does not call getActorFromId.

app/api/v1/statuses/[id]/reblogged_by/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/reblogged_by/route.ts]: Fixed in 61f35ac9. The route now maps actorId into getMastodonActorFromId.

app/api/v1/statuses/[id]/reblogged_by/route.ts
Comment (llun) [reply app/api/v1/statuses/[id]/reblogged_by/route.ts:42]: Fixed in cba1e546. rel=next is omitted for partial last pages; nonempty pages still include rel=prev.

lib/database/sql/status.ts
llun: Not changing this one. The current behavior matches the repo’s other cursor endpoints and Mastodon-style tolerance for stale cursors: an unknown cursor does not make the target status unreadable and the request still returns the current page. Returning 422 would be a broader API behavior change and inconsistent with existing route patterns.

app/api/v1/statuses/[id]/route.test.ts
Comment (llun) [reply app/api/v1/statuses/[id]/route.test.ts:2227]: Fixed in 2900b6fe. Added tests for authenticated non-public boost visibility, invalid limit values, missing status 404, and no next link on a partial last page.

llun:
@codex review

lib/database/sql/status.ts
```diff
@@ -1328,6 +1329,80 @@ export const StatusSQLDatabaseMixin = (
...
+            sameTimestampBuilder
+              .where('createdAt', sinceCursor.createdAt)
+              .where('id', '>', sinceCursor.id)
+          })
+      })
+    }
+
+    if (typeof limit === 'number') {
+      query = query.limit(limit)
+    }
```
llun, line 1465: The query can return duplicate actors. There is no `UNIQUE(actorId, originalStatusId)` constraint on the `statuses` table for `Announce` rows, so the same actor can have multiple Announce rows for the same original status (e.g. due to federation retries or missed undo events). This query returns one row per Announce activity, not one row per actor, so `/reblogged_by` can emit the same account more than once and produce unstable pagination cursors.

The cheapest cross-database fix is to deduplicate by `actorId` in JavaScript after the query, keeping the first (newest) occurrence since results are already ordered `createdAt DESC`:

```ts
const result = await query
// Deduplicate: keep first (newest) Announce per actor
const seen = new Set<string>()
const unique = result.filter((item) => {
  if (seen.has(item.actorId)) return false
  seen.add(item.actorId)
  return true
})
return unique.map((item) => ({ actorId: item.actorId, statusId: item.id }))
```

A database-level alternative (`DISTINCT ON`) is PostgreSQL-specific and would need a branch for SQLite in tests.

---

llun (commented):
## Round 2 review

All three blocking issues from round 1 are addressed:

- ✅ **Double actor lookup** — `RebloggedByAccount` is now `{ actorId, statusId }` and `getRebloggedBy` returns IDs directly, cutting the per-item actor round-trips in half.
- ✅ **`next` link on last page** — `reblogs.length >= limit` correctly gates the `next` link; `prev` is always included for non-empty pages.
- ✅ **Test coverage** — authenticated visibility, invalid `limit` values, 404 for missing status, and last-page `next`-link omission are all now tested.
- ✅ **Duplicated where conditions** — `reblogBase.clone()` is clean.
- ✅ **URL construction** — `headerHost` is used consistently.

### One remaining correctness issue

There is no `UNIQUE(actorId, originalStatusId)` constraint on `statuses` for `Announce` rows, so the same actor can have multiple boosts recorded for the same original status (federation retries, stale undo). The query returns one row per Announce activity rather than per actor, which means `/reblogged_by` can return duplicate accounts in the same page and generate unstable pagination cursors. See the inline comment for a JavaScript-level deduplication fix that works across SQLite and PostgreSQL.

### Open threads from other reviewers

- **Gemini N+1 `getMastodonActorFromId`** — the author's rationale (page cap of 80, consistent with `favourited_by` pattern) is sound. Acceptable to leave as-is.
- **Codex / stale cursor** — author's decision to match existing endpoint behaviour is reasonable. Acceptable.

---

app/api/v1/statuses/[id]/reblogged_by/route.ts
```diff
@@ -42,10 +101,32 @@ export const GET = traceApiRoute(
...
-    return apiResponse({ req, allowedMethods: CORS_HEADERS, data: [] })
+    const accounts = (
+      await Promise.all(
+        reblogs.map(({ actorId }) =>
+          database.getMastodonActorFromId({ id: actorId })
+        )
+      )
+    ).filter((account): account is NonNullable<typeof account> =>
+      Boolean(account)
+    )
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/reblogged_by/route.ts]: Fixed in a311e72e. Added database.getMastodonActorsFromIds and switched reblogged_by to hydrate all account IDs in one bulk path. The bulk lookup batches actors, counters, and last-status timestamps while preserving request order and skipping missing actors. Added coverage for ordering, duplicates, missing actors, non-null last_status_at, and nonzero counters.

llun:
@codex review

llun:
@codex review

llun:
## Round 3 review — LGTM

All threads from rounds 1 and 2 are resolved. The final implementation is solid:

- ✅ **Deduplication** — `ROW_NUMBER() OVER (PARTITION BY actorId ORDER BY createdAt DESC, id DESC)` deduplicates at the SQL level, keeping only the newest visible boost per actor before pagination is applied.
- ✅ **Cursor scope** — cursor lookup runs against the deduped reblog view for this specific status, not the global `statuses` table, so an unrelated status ID can no longer be used as a cursor.
- ✅ **Stale cursor** — returns an empty 200 with no `Link` header rather than silently restarting from page 1. The choice of 200 (vs 4xx) is consistent with Mastodon's documented behaviour.
- ✅ **Batch actor hydration** — `getMastodonActorsFromIds` batches actors, counters, and last-status timestamps in three queries total regardless of page size. `getMastodonActor` now delegates to the same path, so the refactoring benefits all existing callers too.
- ✅ **Test coverage** — pagination, deduplication, stale cursors, visibility (anonymous vs authenticated), invalid params, 404 on missing status, last-page no-`next` link.

The `ROW_NUMBER()` window function requires SQLite ≥ 3.25 (2018), which is standard everywhere this codebase runs, so that's not a concern in practice.

No remaining issues. Ready to merge.

---

llun:
Addressed the Codex review from https://github.com/llun/activities.next/pull/717#issuecomment-4472346892 in 91a49440.

Fixed there:
- legacy public Announce readability now falls back to `content` when `originalStatusId` is null, with route and SQL regressions
- `getRebloggedBy` no longer depends on `ROW_NUMBER()` and uses a portable `NOT EXISTS` anti-join for latest-per-actor dedupe
- pagination `Link` construction now omits links when no trusted/configured host is available instead of throwing

llun:
@codex review

llun:
@codex review

llun:
@codex review

llun:
## Round 4 review — LGTM

All round 3 items remain fixed. The new changes are solid:

**What changed since round 3:**

- ✅ **`ROW_NUMBER()` replaced with `NOT EXISTS` anti-join** — avoids the SQLite ≥ 3.25 window-function requirement. The anti-join correctly retains only the newest visible Announce per actor by checking that no newer visible row from the same actor exists. The two clones of `visibleReblogsQuery` (aliased `visible_reblogs` / `newer_reblogs`) generate independent, properly-scoped subqueries.

- ✅ **`limit + 1` trick** — the route requests one extra row and checks `reblogsPage.length > limit`. This is strictly correct: the previous `reblogs.length >= limit` approach would emit a spurious `next` link when the result set happened to be exactly `limit` items (the true last page), while `limit + 1` only emits `next` when at least one more row provably exists.

- ✅ **Cursor lookup against `visibleReblogsQuery` (not deduped view)** — a cursor ID that was valid when page 1 was fetched but has since been superseded by a newer boost from the same actor still resolves correctly. The two new tests (`accepts a visible reblog cursor even after a newer duplicate supersedes it` / `accepts a since_id cursor...`) verify both directions.

- ✅ **Legacy boost support in `buildPubliclyReadableStatusIdsQuery`** — Announces with `originalStatusId = NULL` (target stored in `content`) are now included in the publicly-readable filter for both the MySQL path and the recursive CTE path. This is a shared-function change that correctly extends visibility handling for all callers, not just `reblogged_by`. Tests at both the DB layer (`status.test.ts`) and the route layer verify legacy announce visibility.

- ✅ **`if (!host) return undefined` guard** — pagination links are suppressed when `headerHost` returns an empty/falsy value (e.g. when `config.host` is not set). The test with a mocked empty config verifies accounts are still returned but no `Link` header is set.

**One observation (non-blocking):**

The pagination conditions added to `dedupedReblogsQuery` (`.where('createdAt', '<', maxCursor.createdAt)` etc.) reference the bare column name `createdAt`. Since `dedupedReblogsQuery` selects from the `visible_reblogs` subquery and projects `createdAt` without a table prefix, this relies on Knex resolving the unqualified name to the subquery's output column. This is standard SQL and works correctly in both PostgreSQL and SQLite — just worth being aware of if the select list ever changes.

---

PR #744, opened 2026-05-22
fix: unify page header layout

llun:
## Summary
- Added a shared `PageHeader` component with Timeline-style typography and description layout.
- Updated Timeline, Messages, Bookmarks, Notifications, Admin, and Settings surfaces to use the same no-icon header treatment.
- Removed decorative header icons and normalized admin/settings subpage title weight and descriptions.

## Screenshots
Responsive composites covering Timeline, Messages, Bookmarks, Notifications, Admin, and Settings.

### Mobile (390x844)
Mobile header comparison

### Tablet (768x1024)
Tablet header comparison

### Desktop (1440x1000)
Desktop header comparison

## Validation
- `yarn lint` (passes with existing warnings)
- `yarn build`
- `yarn test`

### DIFF
--- a/app/(timeline)/admin/accounts/[id]/page.tsx
+++ b/app/(timeline)/admin/accounts/[id]/page.tsx
@@ -2,6 +2,7 @@ import { ArrowLeft } from 'lucide-react'
 import Link from 'next/link'
 import { notFound, redirect } from 'next/navigation'
 
+import { PageHeader } from '@/lib/components/page-header'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import { getMention } from '@/lib/types/domain/actor'
@@ -29,7 +30,7 @@ const Page = async ({ params }: Props) => {
 
   return (
     <div className="space-y-6">
-      <div className="flex items-center gap-3">
+      <div className="flex items-start gap-3">
         <Link
           href="/admin/accounts"
           aria-label="Back to accounts list"
@@ -37,17 +38,18 @@ const Page = async ({ params }: Props) => {
         >
           <ArrowLeft className="h-5 w-5" />
         </Link>
-        <div>
-          <h1 className="text-2xl font-bold">
-            {account.name || account.email}
-          </h1>
-          <p className="text-sm text-muted-foreground">{account.email}</p>
-        </div>
-        {account.role === 'admin' && (
-          <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
-            Admin
-          </span>
-        )}
+        <PageHeader
+          className="flex-1"
+          title={account.name || account.email}
+          description={account.email}
+          actions={
+            account.role === 'admin' && (
+              <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
+                Admin
+              </span>
+            )
+          }
+        />
       </div>
 
       <div className="rounded-2xl border bg-background/80 p-6 shadow-sm">
--- a/app/(timeline)/admin/accounts/[id]/page.tsx
+++ b/app/(timeline)/admin/accounts/[id]/page.tsx
@@ -40,15 +40,17 @@ const Page = async ({ params }: Props) => {
         </Link>
         <PageHeader
           className="flex-1"
-          title={account.name || account.email}
-          description={account.email}
-          actions={
-            account.role === 'admin' && (
-              <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
-                Admin
-              </span>
-            )
+          title={
+            <span className="flex flex-wrap items-center gap-3">
+              {account.name || account.email}
+              {account.role === 'admin' ? (
+                <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
+                  Admin
+                </span>
+              ) : undefined}
+            </span>
           }
+          description={account.name ? account.email : undefined}
         />
       </div>
 
--- a/app/(timeline)/admin/tags/[tag]/page.tsx
+++ b/app/(timeline)/admin/tags/[tag]/page.tsx
@@ -56,6 +56,7 @@ const Page = async ({ params, searchParams }: Props) => {
           <ArrowLeft className="h-5 w-5" />
         </Link>
         <PageHeader
+          className="flex-1"
           title={tag.replace(/^#+/, '')}
           description={`${total} public post${total !== 1 ? 's' : ''}.`}
         />
--- a/app/(timeline)/admin/tags/[tag]/page.tsx
+++ b/app/(timeline)/admin/tags/[tag]/page.tsx
@@ -1,7 +1,8 @@
-import { ArrowLeft, ChevronLeft, ChevronRight, Hash } from 'lucide-react'
+import { ArrowLeft, ChevronLeft, ChevronRight } from 'lucide-react'
 import Link from 'next/link'
 import { notFound, redirect } from 'next/navigation'
 
+import { PageHeader } from '@/lib/components/page-header'
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
@@ -46,7 +47,7 @@ const Page = async ({ params, searchParams }: Props) => {
 
   return (
     <div className="space-y-6">
-      <div className="flex items-center gap-3">
+      <div className="flex items-start gap-3">
         <Link
           href="/admin/tags"
           aria-label="Back to hashtags list"
@@ -54,15 +55,10 @@ const Page = async ({ params, searchParams }: Props) => {
         >
           <ArrowLeft className="h-5 w-5" />
         </Link>
-        <div>
-          <div className="flex items-center gap-2">
-            <Hash className="h-5 w-5 text-muted-foreground" />
-            <h1 className="text-2xl font-bold">{tag.replace(/^#+/, '')}</h1>
-          </div>
-          <p className="text-sm text-muted-foreground">
-            {total} public post{total !== 1 ? 's' : ''}
-          </p>
-        </div>
+        <PageHeader
+          title={tag.replace(/^#+/, '')}
+          description={`${total} public post${total !== 1 ? 's' : ''}.`}
+        />
       </div>
 
       {statuses.length === 0 ? (
--- a/app/(timeline)/settings/sessions/page.tsx
+++ b/app/(timeline)/settings/sessions/page.tsx
@@ -2,6 +2,7 @@ import { Metadata } from 'next'
 import { cookies } from 'next/headers'
 import { redirect } from 'next/navigation'
 
+import { PageHeader } from '@/lib/components/page-header'
 import { SessionsList } from '@/lib/components/settings/SessionsList'
 import { getDatabase } from '@/lib/database'
 import {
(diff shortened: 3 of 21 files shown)


app/(timeline)/admin/tags/[tag]/page.tsx
```diff
@@ -46,23 +47,18 @@ const Page = async ({ params, searchParams }: Props) => {
...
-        <div>
-          <div className="flex items-center gap-2">
-            <Hash className="h-5 w-5 text-muted-foreground" />
-            <h1 className="text-2xl font-bold">{tag.replace(/^#+/, '')}</h1>
-          </div>
-          <p className="text-sm text-muted-foreground">
-            {total} public post{total !== 1 ? 's' : ''}
-          </p>
-        </div>
+        <PageHeader
```
Comment (llun) [inline app/(timeline)/admin/tags/[tag]/page.tsx:58]: Inconsistent with the other admin detail page. `admin/accounts/[id]/page.tsx` wraps `PageHeader` in the same `flex items-start gap-3` row but passes `className="flex-1"` so the header fills the remaining width. Here `flex-1` is missing, so `PageHeader` only takes its content width and its internal `justify-between` becomes inert. Add `className="flex-1"` here too so both back-link detail pages behave the same (and any future `actions` align to the right edge).

---

app/(timeline)/admin/accounts/[id]/page.tsx
```diff
@@ -29,25 +30,26 @@ const Page = async ({ params }: Props) => {
...
+          className="flex-1"
+          title={account.name || account.email}
+          description={account.email}
+          actions={
+            account.role === 'admin' && (
+              <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
+                Admin
+              </span>
+            )
+          }
```
Comment (llun) [inline app/(timeline)/admin/accounts/[id]/page.tsx]: Two things here:

1. For non-admin accounts, `account.role === 'admin' && (...)` evaluates to the boolean `false`, which is then passed as the `actions` ReactNode. `PageHeader` guards with `{actions && ...}` so nothing renders, but passing a boolean is sloppy — prefer `account.role === 'admin' ? (<span>...</span>) : undefined`.

2. Behaviour change worth confirming: previously the `Admin` badge sat immediately after the account name (`flex items-center gap-3`). With `flex-1` + `PageHeader`'s internal `justify-between`, the badge now jumps to the far right edge of the row. If keeping it next to the name is intended, it should be part of `title` rather than `actions`.

---

app/(timeline)/settings/sessions/page.tsx
```diff
@@ -45,16 +46,17 @@ const Page = async () => {
...
-          {sessions.length} session{sessions.length === 1 ? '' : 's'} linked to
-          your account.
-        </p>
-      </div>
+      <PageHeader
+        title="Sessions"
+        description={
+          <>
+            Review active sessions and revoke access you no longer need.
+            <br />
```
llun: The session count was previously a separate `<p>` element. Collapsing it into the same paragraph with a `<br />` is semantically weaker (it's a distinct statement, not a continuation). Since `description` accepts a `ReactNode`, the count would be better rendered as its own block, e.g. `<><p>...</p><p>{sessions.length} session...</p></>` — though note `PageHeader` already wraps `description` in a single `<p>`, so nested `<p>` would be invalid HTML. Consider letting `PageHeader` render `description` in a neutral wrapper (or render plain text fragments) so multi-line descriptions like this one don't need a `<br />` hack.

---

lib/components/admin/StatsOverview.tsx
```diff
@@ -211,35 +212,33 @@ export const StatsOverview: FC<Props> = ({ stats, initialBuckets }) => {
...
-            >
-              {r.label}
-            </button>
-          ))}
-        </div>
-      </div>
+      <PageHeader
+        title="Overview"
+        description="Service usage statistics."
+        actions={
```
llun, line 218: Alignment regression to check: the old header used `flex items-center justify-between`, which vertically centered the range-tab control against the title+description block. `PageHeader` uses `items-start`, so the range tabs now top-align with the `Overview` title and look slightly off against the two-line text block. Same applies to the refresh button in `MainPageTimeline`. If the centered look was intended, `PageHeader` needs an opt-in for `items-center` (or the actions slot should self-center).

---

lib/components/page-header.tsx
```diff
@@ -0,0 +1,27 @@
...
+  description,
+  actions,
+  className
+}: PageHeaderProps) => (
+  <div className={cn('flex items-start justify-between gap-4', className)}>
+    <div className="min-w-0">
+      <h1 className="text-2xl font-semibold">{title}</h1>
+      {description && (
+        <p className="text-sm text-muted-foreground">{description}</p>
+      )}
```
llun: `description` is typed as `ReactNode` but is always rendered inside a `<p>`. Several callers pass JSX into it (`sessions`, `strava`, `NotificationSettings`) — those happen to only contain inline elements today, but a `<p>` will produce invalid-HTML hydration errors if anyone ever passes a block-level element (`<div>`, another `<p>`, etc.). Consider either documenting that `description` must be inline-only, or switching the wrapper to a `<div className="text-sm text-muted-foreground">` to be safe.

---

llun (commented):
## Review: unify page header layout

The shared `PageHeader` extraction is a clean, sensible refactor and the call-site conversions are mostly consistent. A few things to fix before this is mergeable:

### Blocking
- **Merge conflict with `main`.** The branch is based on `3d251af` and is behind current `main` (`c3741cb`). `app/(timeline)/messages/MessagesPage.tsx` conflicts with main (main added a `// Opts the timeline wrapper into the wide-layout rule...` comment in the same return block this PR rewrites). GitHub reports `mergeable_state: dirty`. Rebase onto `main` and resolve the conflict.
- **PR is still a draft** — mark ready for review once the above is resolved.

### Should fix
- `admin/tags/[tag]/page.tsx` is missing `className="flex-1"` on `PageHeader`, unlike the equivalent `admin/accounts/[id]/page.tsx`. See inline comment.
- `admin/accounts/[id]/page.tsx`: `actions` receives a boolean `false` for non-admins, and the `Admin` badge moves to the far-right edge instead of next to the name. See inline comment.

### Worth confirming
- `PageHeader` switches the actions row from `items-center` to `items-start`, which shifts the `StatsOverview` range tabs and the Timeline refresh button. Verify against the screenshots.
- `description` is rendered inside a `<p>`; fine for current callers but fragile for future block-level content. See inline comment.
- `sessions/page.tsx` uses a `<br />` inside the description where two paragraphs were used before — minor semantic regression.

### Looks good
- `Hash` is correctly removed from `admin/tags/[tag]/page.tsx` while still imported/used in the `admin/tags` list page.
- `Mail`/`Bookmark` icon imports removed cleanly with their usages.
- Title weight normalization (`font-bold` → `font-semibold`) matches the stated intent.

---

app/(timeline)/admin/accounts/[id]/page.tsx
```diff
@@ -29,25 +30,26 @@ const Page = async ({ params }: Props) => {
...
-        </div>
-        {account.role === 'admin' && (
-          <span className="rounded-full bg-primary/10 px-2 py-0.5 text-xs font-medium text-primary">
-            Admin
-          </span>
-        )}
+        <PageHeader
+          className="flex-1"
+          title={account.name || account.email}
+          description={account.email}
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/admin/accounts/[id]/page.tsx]: Fixed in 8f6b8d0e. The account detail header now only renders the email as the description when `account.name` exists, so unnamed accounts no longer show the email twice.

app/(timeline)/admin/tags/[tag]/page.tsx
```diff
@@ -46,23 +47,18 @@ const Page = async ({ params, searchParams }: Props) => {
...
+          title={tag.replace(/^#+/, '')}
+          description={`${total} public post${total !== 1 ? 's' : ''}.`}
+        />
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/admin/tags/[tag]/page.tsx:62]: Fixed in 8f6b8d0e. The tag detail header now passes `className="flex-1"`, matching the account detail layout and letting the header fill the row next to the back link.

app/(timeline)/admin/tags/[tag]/page.tsx
Comment (llun) [reply app/(timeline)/admin/tags/[tag]/page.tsx:58]: Fixed in 8f6b8d0e. Added `className="flex-1"` to this `PageHeader`, matching the account detail page behavior.

app/(timeline)/admin/accounts/[id]/page.tsx
Comment (llun) [reply app/(timeline)/admin/accounts/[id]/page.tsx]: Fixed in 8f6b8d0e. The admin badge is now rendered inside the title node, preserving its original position next to the account name/email. This also removes the boolean `false` value from the `actions` prop for non-admin accounts.

app/(timeline)/settings/sessions/page.tsx
llun: Fixed in 8f6b8d0e. `PageHeader` now uses a neutral `div` wrapper for `description`, and the sessions page renders the intro and session count as separate paragraphs instead of using a `<br />`.

lib/components/admin/StatsOverview.tsx
llun: Fixed in 8f6b8d0e. The `PageHeader` action slot now self-centers, which restores the centered visual alignment for both the admin range tabs and the Timeline refresh button without changing the title/description stack alignment.

lib/components/page-header.tsx
llun: Fixed in 8f6b8d0e. `PageHeader` now renders `description` inside a `div.text-sm.text-muted-foreground`, so callers can safely pass block-level description content without invalid nested paragraph markup.

llun:
@codex review

llun (commented):
## Review round 2

All items from the previous round are addressed:

- **Merge conflict** — rebased onto `main`, `mergeable_state` is now `clean`. `MessagesPage.tsx` correctly keeps the `data-layout-width="wide"` wrapper from main.
- **Draft status** — PR is now ready for review.
- **`admin/tags/[tag]/page.tsx`** — `className="flex-1"` added, consistent with `admin/accounts/[id]`.
- **`admin/accounts/[id]/page.tsx`** — the `Admin` badge is now part of `title` (next to the name) instead of a `false`-valued `actions` prop, and `description` is `undefined` when name is absent so the email isn't shown twice. Clean.
- **`sessions/page.tsx`** — `<br />` replaced with two `<p>` elements.
- **`PageHeader`** — `description` wrapper changed from `<p>` to `<div>`, so the nested `<p>` blocks above are valid HTML. `actions` got `self-center`, restoring the centered alignment for the `StatsOverview` range tabs and Timeline refresh button.

CI is green (Lint/Prettier, All Tests, CodeQL). No further blocking issues from my side — LGTM.

---

lib/components/page-header.tsx
```diff
@@ -0,0 +1,27 @@
...
+  className
+}: PageHeaderProps) => (
+  <div className={cn('flex items-start justify-between gap-4', className)}>
+    <div className="min-w-0">
+      <h1 className="text-2xl font-semibold">{title}</h1>
+      {description && (
+        <div className="text-sm text-muted-foreground">{description}</div>
+      )}
+    </div>
+    {actions && <div className="shrink-0 self-center">{actions}</div>}
```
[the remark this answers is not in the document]
llun: No code change here. `self-center` was intentional in 8f6b8d0e to address the previous review about restoring the old `items-center` visual alignment for the Timeline refresh button and the admin range tabs. Removing it would reintroduce that regression. The current action callers are fixed controls, and centering the action against the title+description block matches the prior layout better than top-aligning the action to the title only.

app/(timeline)/settings/sessions/page.tsx
```diff
@@ -45,16 +46,18 @@ const Page = async () => {
...
+        title="Sessions"
+        description={
+          <>
+            <p>Review active sessions and revoke access you no longer need.</p>
+            <p>
+              {sessions.length} session{sessions.length === 1 ? '' : 's'} linked
+              to your account.
+            </p>
+          </>
+        }
```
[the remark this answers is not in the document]
llun: No code change here. `PageHeader` now renders `description` inside a neutral `div`, not inside an interactive parent and not inside a paragraph, so `<p>` children are valid HTML. Tailwind Preflight also resets paragraph margins, so the sessions description does not introduce browser-default vertical spacing. Reverting to `<br />` would undo the semantic fix made in 8f6b8d0e from the previous review.

PR #749, opened 2026-05-23
minor: add indexed account search

llun:
## Summary
- index actors into search_documents when account and actor write paths create, update, or delete actors
- replace v1 account search with indexed broad search plus exact/remote handle resolution
- add account search database and route coverage, including following-only and non-discoverable exact matches

## Stack
- Depends on #748

## Test Results
- yarn run prettier --write .
- yarn lint
- yarn build
- yarn test

### DIFF
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -16,8 +16,8 @@ export const OPTIONS = defaultOptions(CORS_HEADERS)
 
 const SearchParams = z.object({
   q: z.string(),
-  limit: z.coerce.number().min(1).max(80).default(40).optional(),
-  offset: z.coerce.number().min(0).default(0).optional(),
+  limit: z.coerce.number().int().min(1).max(80).default(40).optional(),
+  offset: z.coerce.number().int().min(0).max(10000).default(0).optional(),
   resolve: z
     .enum(['true', 'false'])
     .transform((v) => v === 'true')
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -102,8 +102,16 @@ export const GET = traceApiRoute(
           username: query,
           domain: getConfig().host
         })
-        if (actor && !following) {
-          resultIds.push(actor.id)
+        if (actor) {
+          const canIncludeExact =
+            !following ||
+            (await database.isCurrentActorFollowing({
+              currentActorId: context.currentActor.id,
+              followingActorId: actor.id
+            }))
+          if (canIncludeExact) {
+            resultIds.push(actor.id)
+          }
         }
       }
 
@@ -113,7 +121,9 @@ export const GET = traceApiRoute(
         offset,
         ...(following ? { followingActorId: context.currentActor.id } : {})
       })
-      const ids = [...new Set([...resultIds, ...indexedIds])].slice(0, limit)
+      const ids = [
+        ...new Set([...(offset === 0 ? resultIds : []), ...indexedIds])
+      ].slice(0, limit)
       const results = await database.getMastodonActorsFromIds({ ids })
 
       return apiResponse({
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -64,42 +64,36 @@ export const GET = traceApiRoute(
       }
 
       const query = q.trim()
-      const exactActorIds: string[] = []
-      const addExactActorId = (actor: { id: string } | null | undefined) => {
-        if (!actor) return
-        exactActorIds.push(actor.id)
+      const localDomain = getConfig().host
+      const getSearchParams = (exactActorIds: string[] = []) => {
+        return {
+          q: query,
+          limit,
+          offset,
+          localDomain,
+          exactActorIds,
+          ...(following ? { followingActorId: context.currentActor.id } : {})
+        }
       }
 
-      if (query.includes('@')) {
+      let indexedIds = await database.searchAccountIds(getSearchParams())
+      if (resolve && offset === 0 && indexedIds.length === 0) {
         const handle = parseAccountHandle(query)
         if (handle) {
-          let actor = await database.getActorFromUsername(handle)
-          if (!actor && resolve) {
-            const actorId = await getWebfingerSelf({
-              account: `${handle.username}@${handle.domain}`
-            })
-            actor = actorId
-              ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
-              : null
+          const actorId = await getWebfingerSelf({
+            account: `${handle.username}@${handle.domain}`
+          })
+          const actor = actorId
+            ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
+            : null
+          if (actor) {
+            indexedIds = await database.searchAccountIds(
+              getSearchParams([actor.id])
+            )
           }
-
-          addExactActorId(actor)
         }
-      } else {
-        const actor = await database.getActorFromUsername({
-          username: query,
-          domain: getConfig().host
-        })
-        addExactActorId(actor)
       }
 
-      const indexedIds = await database.searchAccountIds({
-        q: query,
-        limit,
-        offset,
-        exactActorIds,
-        ...(following ? { followingActorId: context.currentActor.id } : {})
-      })
       const results = await database.getMastodonActorsFromIds({
         ids: indexedIds
       })
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -70,7 +70,7 @@ export const GET = traceApiRoute(
       }
 
       const query = q.trim()
-      const resultIds: string[] = []
+      const exactActorIds: string[] = []
 
       if (query.includes('@')) {
         const handle = parseAccountHandle(query)
@@ -93,7 +93,7 @@ export const GET = traceApiRoute(
                 followingActorId: actor.id
               }))
             if (canIncludeExact) {
-              resultIds.push(actor.id)
+              exactActorIds.push(actor.id)
             }
           }
         }
@@ -110,7 +110,7 @@ export const GET = traceApiRoute(
               followingActorId: actor.id
             }))
           if (canIncludeExact) {
-            resultIds.push(actor.id)
+            exactActorIds.push(actor.id)
           }
         }
       }
@@ -119,12 +119,12 @@ export const GET = traceApiRoute(
         q: query,
         limit,
         offset,
+        exactActorIds,
         ...(following ? { followingActorId: context.currentActor.id } : {})
       })
-      const ids = [
-        ...new Set([...(offset === 0 ? resultIds : []), ...indexedIds])
-      ].slice(0, limit)
-      const results = await database.getMastodonActorsFromIds({ ids })
+      const results = await database.getMastodonActorsFromIds({
+        ids: indexedIds
+      })
 
       return apiResponse({
         req,
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -5,7 +5,6 @@ import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
 import { getConfig } from '@/lib/config'
 import { OAuthGuardAnyScope } from '@/lib/services/guards/OAuthGuard'
 import { Scope } from '@/lib/types/database/operations'
-import { Actor } from '@/lib/types/domain/actor'
 import { parseAccountHandle } from '@/lib/utils/accountHandle'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import { ERROR_400, apiResponse, defaultOptions } from '@/lib/utils/response'
@@ -66,17 +65,9 @@ export const GET = traceApiRoute(
 
       const query = q.trim()
       const exactActorIds: string[] = []
-      const addExactActorId = async (actor: Actor | null | undefined) => {
+      const addExactActorId = (actor: { id: string } | null | undefined) => {
         if (!actor) return
-        const canIncludeExact =
-          !following ||
-          (await database.isCurrentActorFollowing({
-            currentActorId: context.currentActor.id,
-            followingActorId: actor.id
-          }))
-        if (canIncludeExact) {
-          exactActorIds.push(actor.id)
-        }
+        exactActorIds.push(actor.id)
       }
 
       if (query.includes('@')) {
@@ -92,14 +83,14 @@ export const GET = traceApiRoute(
               : null
           }
 
-          await addExactActorId(actor)
+          addExactActorId(actor)
         }
       } else {
         const actor = await database.getActorFromUsername({
           username: query,
           domain: getConfig().host
         })
-        await addExactActorId(actor)
+        addExactActorId(actor)
       }
 
       const indexedIds = await database.searchAccountIds({
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -57,16 +57,21 @@ export const GET = traceApiRoute(
         })
       }
 
-      const { q, limit = 40, resolve = false } = parsedParams.data
+      const {
+        q,
+        limit = 40,
+        offset = 0,
+        resolve = false,
+        following = false
+      } = parsedParams.data
 
       if (!q || q.trim().length === 0) {
         return apiResponse({ req, allowedMethods: CORS_HEADERS, data: [] })
       }
 
       const query = q.trim()
-      const results = []
+      const resultIds: string[] = []
 
-      // Try exact match first (username@domain or just username)
       if (query.includes('@')) {
         const handle = parseAccountHandle(query)
         if (handle) {
@@ -80,31 +85,41 @@ export const GET = traceApiRoute(
               : null
           }
 
-          const mastodonActor = actor
-            ? await database.getMastodonActorFromId({ id: actor.id })
-            : null
-          if (mastodonActor) {
-            results.push(mastodonActor)
+          if (actor) {
+            const canIncludeExact =
+              !following ||
+              (await database.isCurrentActorFollowing({
+                currentActorId: context.currentActor.id,
+                followingActorId: actor.id
+              }))
+            if (canIncludeExact) {
+              resultIds.push(actor.id)
+            }
           }
         }
       } else {
-        // Try as local username
         const actor = await database.getActorFromUsername({
           username: query,
           domain: getConfig().host
         })
-        if (actor) {
-          const mastodonActor = await database.getMastodonActorFromId({
-            id: actor.id
-          })
-          if (mastodonActor) results.push(mastodonActor)
+        if (actor && !following) {
+          resultIds.push(actor.id)
         }
       }
+
+      const indexedIds = await database.searchAccountIds({
+        q: query,
+        limit,
+        offset,
+        ...(following ? { followingActorId: context.currentActor.id } : {})
+      })
+      const ids = [...new Set([...resultIds, ...indexedIds])].slice(0, limit)
+      const results = await database.getMastodonActorsFromIds({ ids })
 
       return apiResponse({
         req,
         allowedMethods: CORS_HEADERS,
-        data: results.slice(0, limit)
+        data: results
       })
     }
   )
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -5,6 +5,8 @@ import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
 import { getConfig } from '@/lib/config'
 import { OAuthGuardAnyScope } from '@/lib/services/guards/OAuthGuard'
 import { Scope } from '@/lib/types/database/operations'
+import { Actor } from '@/lib/types/domain/actor'
+import { parseAccountHandle } from '@/lib/utils/accountHandle'
 import { HttpMethod } from '@/lib/utils/getCORSHeaders'
 import { ERROR_400, apiResponse, defaultOptions } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
@@ -12,13 +14,6 @@ import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
 
 export const OPTIONS = defaultOptions(CORS_HEADERS)
-
-const parseAccountHandle = (value: string) => {
-  const normalized = value.trim().replace(/^@/, '')
-  const [username, domain, ...rest] = normalized.split('@')
-  if (!username || !domain || rest.length > 0) return null
-  return { username, domain }
-}
 
 const SearchParams = z.object({
   q: z.string(),
@@ -71,6 +66,18 @@ export const GET = traceApiRoute(
 
       const query = q.trim()
       const exactActorIds: string[] = []
+      const addExactActorId = async (actor: Actor | null | undefined) => {
+        if (!actor) return
+        const canIncludeExact =
+          !following ||
+          (await database.isCurrentActorFollowing({
+            currentActorId: context.currentActor.id,
+            followingActorId: actor.id
+          }))
+        if (canIncludeExact) {
+          exactActorIds.push(actor.id)
+        }
+      }
 
       if (query.includes('@')) {
         const handle = parseAccountHandle(query)
@@ -85,34 +92,14 @@ export const GET = traceApiRoute(
               : null
           }
 
-          if (actor) {
-            const canIncludeExact =
-              !following ||
-              (await database.isCurrentActorFollowing({
-                currentActorId: context.currentActor.id,
-                followingActorId: actor.id
-              }))
-            if (canIncludeExact) {
-              exactActorIds.push(actor.id)
-            }
-          }
+          await addExactActorId(actor)
         }
       } else {
         const actor = await database.getActorFromUsername({
           username: query,
           domain: getConfig().host
         })
-        if (actor) {
-          const canIncludeExact =
-            !following ||
-            (await database.isCurrentActorFollowing({
-              currentActorId: context.currentActor.id,
-              followingActorId: actor.id
-            }))
-          if (canIncludeExact) {
-            exactActorIds.push(actor.id)
-          }
-        }
+        await addExactActorId(actor)
       }
 
       const indexedIds = await database.searchAccountIds({
--- a/app/api/v1/accounts/search/route.ts
+++ b/app/api/v1/accounts/search/route.ts
@@ -77,26 +77,34 @@ export const GET = traceApiRoute(
       }
 
       let indexedIds = await database.searchAccountIds(getSearchParams())
-      if (resolve && offset === 0 && indexedIds.length === 0) {
-        const handle = parseAccountHandle(query)
-        if (handle) {
-          const actorId = await getWebfingerSelf({
-            account: `${handle.username}@${handle.domain}`
-          })
-          const actor = actorId
-            ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
-            : null
-          if (actor) {
-            indexedIds = await database.searchAccountIds(
-              getSearchParams([actor.id])
-            )
-          }
-        }
-      }
-
-      const results = await database.getMastodonActorsFromIds({
+      let results = await database.getMastodonActorsFromIds({
         ids: indexedIds
       })
+
+      const handle = resolve && offset === 0 ? parseAccountHandle(query) : null
+      const exactAcct = handle
+        ? `${handle.username}@${handle.domain}`.toLowerCase()
+        : null
+      const hasExactHandle = exactAcct
+        ? results.some((actor) => actor.acct.toLowerCase() === exactAcct)
+        : false
+
+      if (handle && !hasExactHandle) {
+        const actorId = await getWebfingerSelf({
+          account: `${handle.username}@${handle.domain}`
+        })
+        const actor = actorId
+          ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
+          : null
+        if (actor) {
+          indexedIds = await database.searchAccountIds(
+            getSearchParams([actor.id])
+          )
+          results = await database.getMastodonActorsFromIds({
+            ids: indexedIds
+          })
+        }
+      }
 
       return apiResponse({
         req,
--- a/lib/database/sql/account.ts
+++ b/lib/database/sql/account.ts
@@ -384,20 +384,21 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       inboxUrl: getLocalActorInboxId(actorId),
       sharedInboxUrl: getLocalActorSharedInboxId(domain)
     }
+    const actor = {
+      id: actorId,
+      type: 'Person' as const,
+      accountId,
+      username,
+      domain,
+      settings: JSON.stringify(actorSettings),
+      publicKey,
+      privateKey,
+      createdAt: currentTime,
+      updatedAt: currentTime
+    }
 
     await database.transaction(async (trx) => {
-      await trx('actors').insert({
-        id: actorId,
-        type: 'Person',
-        accountId,
-        username,
-        domain,
-        settings: JSON.stringify(actorSettings),
-        publicKey,
-        privateKey,
-        createdAt: currentTime,
-        updatedAt: currentTime
-      })
+      await trx('actors').insert(actor)
       await increaseCounterValue(
         trx,
         CounterKey.serviceTotalActors(),
@@ -405,7 +406,7 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
         currentTime
       )
       await incrementBucket(trx, 'actors', 1, currentTime)
-      await indexActorSearchDocument(trx, { id: actorId })
+      await indexActorSearchDocument(trx, { id: actorId, actor })
     })
 
     return actorId
--- a/lib/database/sql/account.ts
+++ b/lib/database/sql/account.ts
@@ -143,9 +143,8 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       )
       await incrementBucket(trx, 'accounts', 1, currentTime)
       await incrementBucket(trx, 'actors', 1, currentTime)
+      await indexActorSearchDocument(trx, { id: actorId })
     })
-
-    await indexActorSearchDocument(database, { id: actorId })
 
     return accountId
   },
@@ -385,27 +384,28 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       sharedInboxUrl: getLocalActorSharedInboxId(domain)
     }
 
-    await database('actors').insert({
-      id: actorId,
-      type: 'Person',
-      accountId,
-      username,
-      domain,
-      settings: JSON.stringify(actorSettings),
-      publicKey,
-      privateKey,
-      createdAt: currentTime,
-      updatedAt: currentTime
+    await database.transaction(async (trx) => {
+      await trx('actors').insert({
+        id: actorId,
+        type: 'Person',
+        accountId,
+        username,
+        domain,
+        settings: JSON.stringify(actorSettings),
+        publicKey,
+        privateKey,
+        createdAt: currentTime,
+        updatedAt: currentTime
+      })
+      await increaseCounterValue(
+        trx,
+        CounterKey.serviceTotalActors(),
+        1,
+        currentTime
+      )
+      await incrementBucket(trx, 'actors', 1, currentTime)
+      await indexActorSearchDocument(trx, { id: actorId })
     })
-    await increaseCounterValue(
-      database,
-      CounterKey.serviceTotalActors(),
-      1,
-      currentTime
-    )
-    await incrementBucket(database, 'actors', 1, currentTime)
-
-    await indexActorSearchDocument(database, { id: actorId })
 
     return actorId
   },
--- a/lib/database/sql/account.ts
+++ b/lib/database/sql/account.ts
@@ -89,6 +89,18 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       inboxUrl: getLocalActorInboxId(actorId),
       sharedInboxUrl: getLocalActorSharedInboxId(domain)
     }
+    const actor = {
+      id: actorId,
+      type: 'Person' as const,
+      accountId,
+      username,
+      domain,
+      settings: JSON.stringify(actorSettings),
+      publicKey,
+      privateKey,
+      createdAt: currentTime,
+      updatedAt: currentTime
+    }
 
     await database.transaction(async (trx) => {
       await trx('accounts').insert({
@@ -102,18 +114,7 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
         createdAt: currentTime,
         updatedAt: currentTime
       })
-      await trx('actors').insert({
-        id: actorId,
-        type: 'Person',
-        accountId,
-        username,
-        domain,
-        settings: JSON.stringify(actorSettings),
-        publicKey,
-        privateKey,
-        createdAt: currentTime,
-        updatedAt: currentTime
-      })
+      await trx('actors').insert(actor)
       await trx('account_providers').insert({
         id: `credential_${accountId}`,
         accountId,
@@ -143,7 +144,7 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       )
       await incrementBucket(trx, 'accounts', 1, currentTime)
       await incrementBucket(trx, 'actors', 1, currentTime)
-      await indexActorSearchDocument(trx, { id: actorId })
+      await indexActorSearchDocument(trx, { id: actorId, actor })
     })
 
     return accountId
--- a/lib/database/sql/account.ts
+++ b/lib/database/sql/account.ts
@@ -1,5 +1,6 @@
 import { Knex } from 'knex'
 
+import { indexActorSearchDocument } from '@/lib/database/sql/search'
 import {
   CounterKey,
   getCounterValues,
@@ -143,6 +144,8 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       await incrementBucket(trx, 'accounts', 1, currentTime)
       await incrementBucket(trx, 'actors', 1, currentTime)
     })
+
+    await indexActorSearchDocument(database, { id: actorId })
 
     return accountId
   },
@@ -401,6 +404,8 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       currentTime
     )
     await incrementBucket(database, 'actors', 1, currentTime)
+
+    await indexActorSearchDocument(database, { id: actorId })
 
     return actorId
   },
--- a/lib/database/sql/actor.ts
+++ b/lib/database/sql/actor.ts
@@ -274,20 +274,24 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       inboxUrl,
       sharedInboxUrl
     }
-    await database('actors').insert({
+    const actor = {
       id: actorId,
       type,
       username,
       domain,
       name,
       summary,
+      accountId: null,
       settings: JSON.stringify(settings),
       publicKey,
       privateKey,
+      deletionStatus: null,
+      deletionScheduledAt: null,
       createdAt: new Date(createdAt),
       updatedAt: currentTime
-    })
-    await indexActorSearchDocument(database, { id: actorId })
+    }
+    await database('actors').insert(actor)
+    await indexActorSearchDocument(database, { id: actorId, actor })
     return this.getActorFromId({ id: actorId })
   },
 
@@ -319,20 +323,24 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       inboxUrl,
       sharedInboxUrl
     }
-    await database('actors').insert({
+    const actor = {
       id: actorId,
       type,
       username,
       domain,
       name,
       summary,
+      accountId: null,
       settings: JSON.stringify(settings),
       publicKey,
       privateKey,
+      deletionStatus: null,
+      deletionScheduledAt: null,
       createdAt: new Date(createdAt),
       updatedAt: currentTime
-    })
-    await indexActorSearchDocument(database, { id: actorId })
+    }
+    await database('actors').insert(actor)
+    await indexActorSearchDocument(database, { id: actorId, actor })
     return this.getMastodonActor(actorId)
   },
 
@@ -813,6 +821,17 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       ...(sharedInboxUrl ? { sharedInboxUrl } : null)
     }
 
+    const currentTime = new Date()
+    const updatedActor: SQLActor = {
+      ...persistedActor,
+      ...(type ? { type } : null),
+      ...(name ? { name } : null),
+      ...(summary ? { summary } : null),
+      ...(publicKey ? { publicKey } : null),
+      settings: JSON.stringify(settings),
+      updatedAt: currentTime
+    }
+
     await database<SQLActor>('actors')
       .where('id', actorId)
       .update({
@@ -823,9 +842,12 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
         ...(publicKey ? { publicKey } : null),
 
         settings: JSON.stringify(settings),
-        updatedAt: new Date()
+        updatedAt: currentTime
       })
-    await indexActorSearchDocument(database, { id: actorId })
+    await indexActorSearchDocument(database, {
+      id: actorId,
+      actor: updatedActor
+    })
     return this.getActorFromId({ id: actorId })
   },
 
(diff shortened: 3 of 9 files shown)


app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +85,41 @@ export const GET = traceApiRoute(
...
           domain: getConfig().host
         })
-        if (actor) {
-          const mastodonActor = await database.getMastodonActorFromId({
-            id: actor.id
-          })
-          if (mastodonActor) results.push(mastodonActor)
+        if (actor && !following) {
+          resultIds.push(actor.id)
         }
```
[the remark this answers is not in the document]
llun: Fixed. Local exact matches now use the same following check as remote exact matches, so followed local noindex accounts can be returned when following=true.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +85,41 @@ export const GET = traceApiRoute(
...
       }
 
+      const indexedIds = await database.searchAccountIds({
+        q: query,
+        limit,
+        offset,
+        ...(following ? { followingActorId: context.currentActor.id } : {})
+      })
+      const ids = [...new Set([...resultIds, ...indexedIds])].slice(0, limit)
```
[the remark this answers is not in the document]
llun: Fixed. Exact matches are only prepended on the first page when offset is 0.

lib/database/sql/search/account.ts
```diff
@@ -25,9 +158,19 @@ export const deleteActorSearchDocument = async (
...
+  database: Knex,
+  { afterId = null, limit = 500 }: ReindexSearchDocumentsParams = {}
+): Promise<ReindexSearchDocumentsResult> => {
+  const query = database<SQLActor>('actors').select('id').orderBy('id', 'asc')
+  if (afterId) query.where('id', '>', afterId)
+
+  const rows = await query.limit(limit)
+  for (const row of rows) {
+    await indexActorSearchDocument(database, { id: row.id })
+  }
```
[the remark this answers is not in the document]
llun: Fixed. Account reindex now fetches actor rows in the batch query and reuses them for indexing instead of doing a lookup per actor id.

llun:
@codex review

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +85,51 @@ export const GET = traceApiRoute(
...
 
+      const indexedIds = await database.searchAccountIds({
+        q: query,
+        limit,
+        offset,
+        ...(following ? { followingActorId: context.currentActor.id } : {})
+      })
+      const ids = [
+        ...new Set([...(offset === 0 ? resultIds : []), ...indexedIds])
+      ].slice(0, limit)
```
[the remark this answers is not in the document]
llun: Fixed. Exact account matches now flow through searchAccountIds via exactActorIds, so pagination is applied over one ordered result set instead of prepending outside the indexed page.

lib/database/sql/search/account.ts
```diff
@@ -1,21 +1,158 @@
...
+}
 
 export const indexActorSearchDocument = async (
-  _database: Knex,
-  _params: { id: string }
-): Promise<void> => {}
+  database: Knex,
+  { id }: { id: string }
+): Promise<void> => {
+  const actor = await database<SQLActor>('actors').where('id', id).first()
```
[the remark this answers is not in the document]
llun: Fixed. indexActorSearchDocument now accepts an optional prefetched actor row, and actor create/update paths pass the row they already have.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +85,51 @@ export const GET = traceApiRoute(
...
+            resultIds.push(actor.id)
+          }
         }
       }
 
+      const indexedIds = await database.searchAccountIds({
+        q: query,
+        limit,
+        offset,
+        ...(following ? { followingActorId: context.currentActor.id } : {})
```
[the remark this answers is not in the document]
llun: Fixed. Exact account matches are now part of the indexed account query, so offset pagination no longer skips indexed results after an exact match.

lib/database/sql/search/account.ts
```diff
@@ -1,21 +1,158 @@
...
+      actor.summary ?? ''
+    ].join(' ')
+  )
+}
+
+const isDiscoverableAccount = (actor: SQLActor) => {
+  const settings = getCompatibleJSON<Record<string, unknown>>(
+    actor.settings as string | Record<string, unknown>
+  )
+  return settings.noindex !== true && actor.deletionStatus !== 'deleting'
```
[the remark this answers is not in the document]
llun: Fixed. Actor deletion-state transitions now refresh the account search document after scheduling, canceling, or starting deletion.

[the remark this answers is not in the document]
llun: Fixed. Federation signing actors are marked non-discoverable and globally excluded from account search, including exact-match paths.

llun:
@codex review

app/api/v1/accounts/search/route.ts
```diff
@@ -57,16 +57,21 @@ export const GET = traceApiRoute(
...
         return apiResponse({ req, allowedMethods: CORS_HEADERS, data: [] })
       }
 
       const query = q.trim()
-      const results = []
+      const exactActorIds: string[] = []
 
-      // Try exact match first (username@domain or just username)
       if (query.includes('@')) {
         const handle = parseAccountHandle(query)
```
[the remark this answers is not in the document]
llun: Fixed. Account handle parsing now uses the shared parseAccountHandle utility, which lowercases handles consistently.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +85,51 @@ export const GET = traceApiRoute(
...
+            const canIncludeExact =
+              !following ||
+              (await database.isCurrentActorFollowing({
+                currentActorId: context.currentActor.id,
+                followingActorId: actor.id
+              }))
+            if (canIncludeExact) {
+              exactActorIds.push(actor.id)
+            }
           }
```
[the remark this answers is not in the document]
llun: Fixed. The route now uses a local addExactActorId helper for the follow check and exact ID append.

lib/database/sql/search/account.ts
```diff
@@ -25,9 +228,17 @@ export const deleteActorSearchDocument = async (
...
-  nextCursor: null
-})
+  database: Knex,
+  { afterId = null, limit = 500 }: ReindexSearchDocumentsParams = {}
+): Promise<ReindexSearchDocumentsResult> => {
+  const query = database<SQLActor>('actors').select('*').orderBy('id', 'asc')
+  if (afterId) query.where('id', '>', afterId)
+
+  const rows = await query.limit(limit)
+  await Promise.all(rows.map((row) => upsertActorSearchDocument(database, row)))
```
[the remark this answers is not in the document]
llun: Fixed. Account reindexing now upserts rows sequentially instead of issuing the whole batch concurrently.

lib/database/sql/search/account.ts
```diff
@@ -1,21 +1,224 @@
...
+    followingActorId,
+    exactActorIds = []
+  }: SearchAccountsParams
+): Promise<string[]> => {
+  const handle = parseAccountHandle(q)
+  const normalizedQuery = q.trim().replace(/^@/, '').toLowerCase()
+  const normalizedExactActorIds = [...new Set(exactActorIds)]
+  const query = database(SEARCH_DOCUMENTS_TABLE)
+    .innerJoin('actors', 'actors.id', 'search_documents.entityId')
+    .select<{ entityId: string }[]>('search_documents.entityId')
```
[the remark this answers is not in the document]
llun: Fixed. searchAccountIds now resolves exactActorIds against the actors table before the indexed query, so exact matches work even when no search_documents row exists, while still preserving pagination.

llun:
@codex review

lib/database/sql/account.ts
```diff
@@ -144,6 +145,8 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       await incrementBucket(trx, 'actors', 1, currentTime)
     })
 
+    await indexActorSearchDocument(database, { id: actorId })
```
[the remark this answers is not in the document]
llun: Moved account actor search indexing into the account creation transaction.

lib/database/sql/actor.ts
```diff
@@ -1384,5 +1417,6 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       // Finally delete the actor
       await trx('actors').where('id', actorId).delete()
     })
+    await deleteActorSearchDocument(database, { id: actorId })
```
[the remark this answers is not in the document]
llun: Moved actor search document deletion into the actor deletion transaction.

lib/utils/accountHandle.ts
```diff
@@ -0,0 +1,6 @@
+export const parseAccountHandle = (value: string) => {
+  const normalized = value.trim().replace(/^@/, '').toLowerCase()
```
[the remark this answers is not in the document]
llun: Preserved username casing in parsed handles and only lowercase the domain; case-insensitive search still lowercases at comparison time.

lib/database/sql/account.ts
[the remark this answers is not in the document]
llun: Account creation now indexes the actor search document inside the same transaction as the account/actor writes.

lib/database/sql/actor.ts
```diff
@@ -270,19 +274,24 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
...
       publicKey,
       privateKey,
+      deletionStatus: null,
+      deletionScheduledAt: null,
       createdAt: new Date(createdAt),
       updatedAt: currentTime
-    })
+    }
+    await database('actors').insert(actor)
+    await indexActorSearchDocument(database, { id: actorId, actor })
```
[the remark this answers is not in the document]
llun: Remote actor creation now wraps the actor insert and search indexing in one transaction.

llun:
@codex review

lib/database/sql/search/account.ts
```diff
@@ -1,21 +1,277 @@
...
+  exactActorIds: string[]
+}) => {
+  const lowerHandleSQL = getLowerAccountHandleSQL(database)
+
+  if (exactActorIds.length > 0) {
+    query.orderByRaw(
+      `case when ?? in (${exactActorIds.map(() => '?').join(', ')}) then 0 else 1 end`,
+      ['search_documents.entityId', ...exactActorIds]
+    )
+  }
```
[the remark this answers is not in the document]
llun: Fixed. The redundant exact-id ordering clause was removed; exact results are handled separately and excluded from the indexed query.

lib/database/sql/search/account.ts
```diff
@@ -25,9 +281,19 @@ export const deleteActorSearchDocument = async (
...
+  database: Knex,
+  { afterId = null, limit = 500 }: ReindexSearchDocumentsParams = {}
+): Promise<ReindexSearchDocumentsResult> => {
+  const query = database<SQLActor>('actors').select('*').orderBy('id', 'asc')
+  if (afterId) query.where('id', '>', afterId)
+
+  const rows = await query.limit(limit)
+  for (const row of rows) {
+    await upsertActorSearchDocument(database, row)
+  }
```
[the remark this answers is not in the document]
llun: Fixed. Account reindexing now builds search-document rows and upserts them in batches instead of issuing one upsert per actor.

llun:
@codex review

lib/database/sql/actor.ts
```diff
@@ -817,13 +846,20 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
         ...(publicKey ? { publicKey } : null),
 
         settings: JSON.stringify(settings),
-        updatedAt: new Date()
+        updatedAt: currentTime
       })
+    await indexActorSearchDocument(database, {
+      id: actorId,
+      actor: updatedActor
+    })
```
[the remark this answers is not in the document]
llun: Fixed. Actor updates and the corresponding search-document update now run inside the same database transaction.

lib/database/sql/actor.ts
```diff
@@ -912,6 +948,7 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       deletionScheduledAt: scheduledAt,
       updatedAt: currentTime
     })
+    await indexActorSearchDocument(database, { id: actorId })
```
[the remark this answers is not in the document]
llun: Fixed. Actor deletion-status transitions now update the actor row and refresh the search document within one transaction.

lib/database/sql/search/account.ts
```diff
@@ -25,9 +318,17 @@ export const deleteActorSearchDocument = async (
...
-  _database?: Knex,
-  _params?: unknown
-): Promise<ReindexSearchDocumentsResult> => ({
-  indexed: 0,
-  nextCursor: null
-})
+  database: Knex,
+  { afterId = null, limit = 500 }: ReindexSearchDocumentsParams = {}
+): Promise<ReindexSearchDocumentsResult> => {
+  const query = database<SQLActor>('actors').select('*').orderBy('id', 'asc')
```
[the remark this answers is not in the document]
llun: Fixed. Account reindex now selects only the fields required to build account search documents instead of `select("*")`.

llun:
@codex review

app/api/v1/accounts/search/route.ts
```diff
@@ -57,16 +52,33 @@ export const GET = traceApiRoute(
...
+        const canIncludeExact =
+          !following ||
+          (await database.isCurrentActorFollowing({
+            currentActorId: context.currentActor.id,
+            followingActorId: actor.id
+          }))
+        if (canIncludeExact) {
+          exactActorIds.push(actor.id)
+        }
+      }
```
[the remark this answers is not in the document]
llun: Fixed. Exact account matches no longer perform a separate following check in the route; filtering remains centralized in searchAccountIds via followingActorId.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +92,31 @@ export const GET = traceApiRoute(
               : null
           }
 
-          const mastodonActor = actor
-            ? await database.getMastodonActorFromId({ id: actor.id })
-            : null
-          if (mastodonActor) {
-            results.push(mastodonActor)
-          }
+          await addExactActorId(actor)
```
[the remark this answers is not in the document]
llun: Fixed. Removed the now-unnecessary await from the exact actor addition.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +92,31 @@ export const GET = traceApiRoute(
...
           username: query,
           domain: getConfig().host
         })
-        if (actor) {
-          const mastodonActor = await database.getMastodonActorFromId({
-            id: actor.id
-          })
-          if (mastodonActor) results.push(mastodonActor)
-        }
+        await addExactActorId(actor)
```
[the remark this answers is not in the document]
llun: Fixed. Removed the now-unnecessary await from the exact actor addition.

llun:
@codex review

lib/database/sql/search/account.ts
```diff
@@ -1,21 +1,314 @@
...
+            'actors.username',
+            handle.username.toLowerCase()
+          ])
+          .whereRaw('LOWER(??) = ?', ['actors.domain', handle.domain])
+      })
+    }
+    if (normalizedExactActorIds.length > 0) {
+      builder.orWhereIn('search_documents.entityId', normalizedExactActorIds)
+    }
+  })
```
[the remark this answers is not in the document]
llun: Fixed in cb2a0241. Handle-derived exact actor IDs are resolved before the indexed query, and the broad indexed query now only enforces `discoverable` when not searching within follows. Added coverage for followed noindex accounts and exact handles without `search_documents` rows.

lib/database/sql/account.ts
```diff
@@ -142,6 +143,7 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
       )
       await incrementBucket(trx, 'accounts', 1, currentTime)
       await incrementBucket(trx, 'actors', 1, currentTime)
+      await indexActorSearchDocument(trx, { id: actorId })
```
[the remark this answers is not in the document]
llun: Fixed in cb2a0241. Account creation now builds the actor row once, inserts it, and passes the same row to `indexActorSearchDocument` inside the transaction, avoiding the extra actor lookup. Added a regression test that records no actor select during `createAccount` while confirming the search document is written.

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,312 @@
...
+  )
+  const isInternalFederationActor =
+    actor.type === FEDERATION_SIGNING_ACTOR_TYPE &&
+    isFederationSigningActorUsername(actor.username) &&
+    !actor.accountId
+  return (
+    settings.noindex !== true &&
+    actor.deletionStatus !== 'deleting' &&
+    !isInternalFederationActor
+  )
```
llun, line 80: `isDiscoverableAccount` only treats `deletionStatus === 'deleting'` as a reason to drop visibility, and `applySearchableAccountFilters` similarly only excludes `'deleting'`. After `scheduleActorDeletion` writes `deletionStatus = 'scheduled'`, the search document is re-indexed with `discoverable: true` and the row still passes the broad-search filter, so a scheduled-for-deletion account remains fully discoverable until the deletion job actually transitions it to `'deleting'`.

That's almost certainly not what we want — once a user has asked for deletion (even if there's a grace window), broad account search shouldn't keep surfacing them. Either include `'scheduled'` here alongside `'deleting'`, or document explicitly that scheduled accounts remain searchable. The existing test only covers the `startActorDeletion` transition, so this gap isn't caught.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,312 @@
...
+        'actors.username',
+        'actors.domain',
+        normalizedQuery,
+        'actors.username',
+        normalizedQuery,
+        'actors.username',
+        'actors.domain',
+        `${normalizedQuery}%`,
+        'actors.username',
+        `${normalizedQuery}%`
```
llun: The LIKE patterns built from `normalizedQuery` here (`${normalizedQuery}%`) are not escaped. If a caller searches for a query containing `%` or `_` (the SQL wildcards), the ordering predicate will rank unrelated rows as a prefix match — e.g. `_unner` would receive rank 2/3 against any single-character-prefixed username. The `WHERE` filter in `applySearchDocumentFilter` already routes those through `escapeLikePattern`/FTS, so the result set is fine, but the ranking will silently lie.

Either escape the pattern with the same `escapeLikePattern` helper used in `documents.ts` and add `ESCAPE '\\'` to the `like` clauses, or filter the wildcard characters out of `normalizedQuery` before interpolating.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,312 @@
...
+  followingActorId
+}: {
+  database: Knex
+  handle: ReturnType<typeof parseAccountHandle>
+  exactActorIds: string[]
+  followingActorId?: string | null
+}) => {
+  const handleActorRows = handle
+    ? await database('actors')
+        .select<{ id: string }[]>('actors.id')
```
llun, line 233: `getExactAccountIds` queries `actors` by `handle.username`/`handle.domain` as exact column matches. On SQLite and PostgreSQL with default collations, that comparison is case-sensitive, while `parseAccountHandle` only lowercases the domain and preserves the username case.

That means `@charlie@remote.test` will not find an actor stored as `Charlie@remote.test` via the exact path (and neither will the route's own `database.getActorFromUsername(handle)` call on line 75 of the route). The broad indexed search is case-insensitive, so users will hit a confusing asymmetry: a non-discoverable / following-only user is reachable by handle only if they happen to type the username casing correctly. Consider lowercasing both sides (`whereRaw('LOWER(??) = ?', ...)`) when looking up by handle here, or normalizing usernames at insert time.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,312 @@
...
+    database,
+    handle,
+    exactActorIds: normalizedExactActorIds,
+    followingActorId
+  })
+  const exactPageIds =
+    offset < exactResultIds.length
+      ? exactResultIds.slice(offset, offset + limit)
+      : []
+  const indexedLimit = limit - exactPageIds.length
```
llun, line 342: When the caller passes a handle plus `exactActorIds` and at least one of those IDs is filtered out by `applySearchableAccountFilters`/`applyFollowingFilter` (e.g. the route added a resolved federation signing actor or a not-followed actor), `exactResultIds` is shorter than the caller-visible "exact" set. Pagination then uses `exactResultIds.length` to compute `indexedOffset`, so the user reaches a page that is correctly sized but the assumption that "exact match = page 0 top" no longer holds.

This is mostly harmless in practice today, but it's worth a brief comment near `getExactAccountIds` documenting that the returned list is the *visible* subset and pagination is driven off that length — otherwise future callers that try to surface exact-match metadata back to the route will get confused.

---

lib/database/sql/actor.ts
```diff
@@ -270,18 +274,25 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
...
       publicKey,
       privateKey,
+      deletionStatus: null,
+      deletionScheduledAt: null,
       createdAt: new Date(createdAt),
       updatedAt: currentTime
+    }
+    await database.transaction(async (trx) => {
+      await trx('actors').insert(actor)
+      await indexActorSearchDocument(trx, { id: actorId, actor })
```
llun: `createActor` and `createMastodonActor` are now byte-identical aside from the final `getActorFromId` vs `getMastodonActor` return. Both build the same `actor` literal, run the same transaction, and call `indexActorSearchDocument` the same way. Worth extracting a shared helper (e.g. `insertActorWithSearchIndex(trx, params)`) so a future change to the insert/index contract only has to be made in one place — easy to drift otherwise.

---

lib/database/sql/account.ts
```diff
@@ -382,25 +385,28 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
...
+        updatedAt: currentTime
+      })
+      await increaseCounterValue(
+        trx,
+        CounterKey.serviceTotalActors(),
+        1,
+        currentTime
+      )
+      await incrementBucket(trx, 'actors', 1, currentTime)
+      await indexActorSearchDocument(trx, { id: actorId })
```
llun: `createActorForAccount` still calls `indexActorSearchDocument(trx, { id: actorId })` without passing the actor row, which triggers an extra `SELECT * FROM actors` inside the transaction even though the row that was just inserted is right there. This is the same N+1 issue that was fixed for `createAccount` above (and covered by the "indexes local account creation without reloading the inserted actor" test). Build the actor literal once like `createAccount` does and pass it via `{ id: actorId, actor }` to avoid the redundant lookup.

---

lib/database/sql/actor.ts
```diff
@@ -270,18 +274,25 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
...
       settings: JSON.stringify(settings),
       publicKey,
       privateKey,
+      deletionStatus: null,
+      deletionScheduledAt: null,
       createdAt: new Date(createdAt),
       updatedAt: currentTime
+    }
+    await database.transaction(async (trx) => {
+      await trx('actors').insert(actor)
```
llun: `createActor` and `createMastodonActor` (just below) are now byte-identical aside from the final `getActorFromId` vs `getMastodonActor` return — both build the same `actor` literal, run the same transaction, and call `indexActorSearchDocument` the same way. Worth extracting a shared helper (e.g. `insertActorWithSearchIndex(trx, params)`) so a future change to the insert/index contract only has to be made in one place — these are easy to drift otherwise.

---

llun (commented):
Review of the indexed account search changes. The earlier review rounds (gemini-code-assist, codex) covered most of the obvious issues — transactions, pagination, federation actors, prefetched actors for `createAccount` — and those are all in good shape now. The remaining items I'd like to flag:

**Potential bugs**
- `scheduleActorDeletion` re-indexes the search document but `isDiscoverableAccount` only filters on `deletionStatus === 'deleting'`, so scheduled-for-deletion accounts keep appearing in broad search until the deletion job actually runs. Unclear if intentional — needs a decision or a comment.
- Case-sensitivity asymmetry: `parseAccountHandle` preserves username casing, but `getExactAccountIds` and the route's `getActorFromUsername` lookups query `actors.username` as case-sensitive (SQLite/Postgres default collation), while the broad indexed search is case-insensitive. `@charlie@remote.test` won't reach a non-discoverable / following-only actor stored as `Charlie`.
- The `LIKE` patterns in `applyAccountOrdering` aren't passed through `escapeLikePattern` — a query containing `%`/`_` will mis-rank results (the `WHERE` filter is fine because it uses the FTS/`escapeLikePattern` path).

**Polish**
- `createActorForAccount` still does the redundant actor `SELECT` after insert that was removed from `createAccount` in an earlier round.
- `offset` on the search route has no upper bound — easy DoS via deep pagination.
- `createActor` and `createMastodonActor` in `actor.ts` are nearly identical now; worth a shared helper to keep them from drifting.

Inline comments below for each.

---

lib/database/sql/search/account.ts
llun: Fixed in cd7c9038. Scheduled deletion now makes account search documents non-discoverable, and broad account search only includes actors with no deletion status. Added coverage for scheduled, cancelled, and deleting transitions.

lib/database/sql/search/account.ts
llun: Fixed in cd7c9038. Exact handle resolution now compares lowercase username and domain in SQL, so mixed-case stored usernames can still be found by lower-case handles. Added non-discoverable mixed-case coverage.

lib/database/sql/search/account.ts
llun: Fixed in cd7c9038. Account ordering LIKE patterns now use the shared escaped LIKE helper with explicit ESCAPE clauses, and there is regression coverage for wildcard ranking.

lib/database/sql/search/account.ts
llun: Addressed in cd7c9038. Added a code comment documenting that exact pagination is driven by the visible exact-ID subset after searchable and following filters.

lib/database/sql/account.ts
llun: Fixed in cd7c9038. `createActorForAccount` now builds the inserted actor row once and passes it to `indexActorSearchDocument`, avoiding the redundant actor lookup. Added query-count regression coverage.

lib/database/sql/actor.ts
llun: Fixed in cd7c9038. `createActor` and `createMastodonActor` now share `insertActorWithSearchIndex`, so actor insertion and indexing stay in one path.

lib/database/sql/actor.ts
llun: Fixed in cd7c9038. `createActor` and `createMastodonActor` now use the shared `insertActorWithSearchIndex` helper, so the duplicate insert/index logic is gone.

llun:
Addressed the latest review summary in cd7c9038. In addition to the inline fixes, the search route now rejects deep offsets above 10000 with route-level regression coverage. Full verification passed locally: prettier, lint, build, and test.

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,313 @@
...
+}
+
+const chunkArray = <T>(items: T[], size: number) => {
+  const chunks: T[][] = []
+  for (let index = 0; index < items.length; index += size) {
+    chunks.push(items.slice(index, index + size))
+  }
+  return chunks
+}
+
```
llun, line 170: The new `applySearchableAccountFilters` excludes any actor with `deletionStatus IS NOT NULL` — including `'scheduled'`. Combined with `applyFollowingFilter`, this means a follower searching for an account that has just been scheduled for deletion can no longer reach it even by exact handle. In Mastodon's flow, scheduled deletions have a multi-day grace window during which followers may still want to find and view the account (e.g. to unfollow, archive DMs, etc.).

Two options to consider:
1. If the intent is to fully hide scheduled accounts everywhere, please add explicit test coverage for the exact-handle and `following=true` paths so the behavior is locked in.
2. If the intent is only to hide them from broad/discovery search, keep `discoverable=false` on the document but loosen `applySearchableAccountFilters` so that the exact-handle and following branches still surface them.

Right now option 1's behavior is implicit and untested.

---

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +83,31 @@ export const GET = traceApiRoute(
...
           domain: getConfig().host
         })
-        if (actor) {
-          const mastodonActor = await database.getMastodonActorFromId({
-            id: actor.id
-          })
-          if (mastodonActor) results.push(mastodonActor)
-        }
+        addExactActorId(actor)
       }
```
llun, line 107: Now that `searchAccountIds` (via `getExactAccountIds`) resolves both the `handle`-based lookup and the supplied `exactActorIds` itself, the route's two `database.getActorFromUsername(...)` branches above are essentially duplicate work — they always re-query the same row that `getExactAccountIds` will find. The only branch that still needs to live in the route is the `resolve=true` webfinger fallback (because that does network I/O and writes a new actor).

Consider collapsing the route to:
- If `resolve=true` and the query parses as a handle and `searchAccountIds` returns nothing, then do the webfinger + `recordActorIfNeeded` and retry.
- Otherwise pass `q` straight to `searchAccountIds` and let it handle exact resolution.

That removes a DB round-trip per search and a chunk of duplicated logic, without changing observable behavior.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,313 @@
...
+  for (const chunk of chunkArray(rows, ACCOUNT_SEARCH_DOCUMENT_BATCH_SIZE)) {
+    await database(SEARCH_DOCUMENTS_TABLE)
+      .insert(chunk)
+      .onConflict('id')
+      .merge(SEARCH_DOCUMENT_MERGE_COLUMNS)
+  }
+}
 
 export const indexActorSearchDocument = async (
```
llun, line 301: When `indexActorSearchDocument` is called with a `providedActor` argument from the create/update write paths, we never hit the fallback `database<SQLActor>('actors').where('id', id).first()` lookup or the `if (!actor) deleteActorSearchDocument(...)` branch. That branch only fires for callers that pass `{ id }` alone — i.e. the deletion-transition paths and external callers.

It's a subtle contract: "passing an actor literal means index this exact shape; passing only an id means index whatever is currently in the actors table, or delete the doc if the row is gone." Worth a one-line comment above the function describing those two modes, so future callers don't accidentally pass a stale `providedActor` (e.g. cached from before a concurrent update) and end up writing stale data to `search_documents` indefinitely.

Not blocking — the current call sites all pass freshly-built actor literals — but the invariant is load-bearing.

---

lib/database/sql/search/account.ts
```diff
@@ -26,7 +317,30 @@ export const deleteActorSearchDocument = async (
...
-  throwUnimplementedSearchMethod('reindexSearchAccounts')
+  database: Knex,
+  { afterId = null, limit = 500 }: ReindexSearchDocumentsParams = {}
+): Promise<ReindexSearchDocumentsResult> => {
+  const query = database<SQLActor>('actors')
+    .select(
+      'id',
+      'username',
+      'domain',
+      'settings',
```
llun, line 391: `reindexSearchAccounts` reads a batch of actors, builds search-document rows from that snapshot, then upserts. If a concurrent write path (e.g. `updateActor`, `startActorDeletion`) fires between the SELECT and the upsert chunk, the freshly-written `discoverable`/`documentText` value will be silently overwritten by the stale snapshot — until something else triggers a re-index for that actor.

In practice the broad search query joins `actors` and re-checks `deletionStatus` via `applySearchableAccountFilters`, so visibility stays correct even with stale `discoverable`. But text/name changes (and exact `getExactAccountIds` paths that *don't* re-derive from the live row) can be served stale until the next event. Two mitigations worth considering:

1. Take a snapshot of `actors.updatedAt` during the SELECT and add it to the upsert merge condition so older writes don't overwrite newer ones.
2. Or simply note in a comment that reindex is intentionally best-effort and any out-of-band write wins via its own indexing path — so admins running reindex over a live system know what to expect.

Not a blocker, just brittle if you ever run reindex concurrently with traffic.

---

app/api/v1/accounts/search/route.ts
```diff
@@ -57,16 +51,25 @@ export const GET = traceApiRoute(
...
+      const exactActorIds: string[] = []
+      const addExactActorId = (actor: { id: string } | null | undefined) => {
+        if (!actor) return
+        exactActorIds.push(actor.id)
+      }
 
-      // Try exact match first (username@domain or just username)
       if (query.includes('@')) {
         const handle = parseAccountHandle(query)
         if (handle) {
```
llun: `getExactAccountIds` was updated to compare `LOWER(actors.username) = ?` so mixed-case stored handles can be found from lower-case query input — good. But this `database.getActorFromUsername(handle)` call still does a case-sensitive `where('username', username)` lookup in `lib/database/sql/actor.ts:getActorFromUsername`. The end result is correct (the case-insensitive lookup inside `searchAccountIds` will surface the actor regardless), but there are two side-effects:

1. The route's own lookup is essentially dead code for mixed-case handles — it always misses, then `searchAccountIds` re-runs the lookup correctly.
2. When `resolve=true` and the local actor exists under a different casing, the route falls into the `if (!actor && resolve)` branch and fires a needless webfinger + `recordActorIfNeeded` for an actor we already have. That's an extra network round-trip per mixed-case handle search.

Either make `getActorFromUsername` case-insensitive (preferred — would match the search path and what users expect from a handle lookup), or short-circuit the resolve branch by deferring to `searchAccountIds` to decide whether the actor exists.

---

llun (commented):
Round 2 review of cd7c9038.

The fixes from the previous round all landed cleanly — `scheduleActorDeletion` now hides accounts from search via the stricter `deletionStatus IS NULL` filter, the LIKE patterns in `applyAccountOrdering` are escaped properly, `getExactAccountIds` does case-insensitive handle lookups, `createActorForAccount` reuses the inserted actor row, and `createActor`/`createMastodonActor` share `insertActorWithSearchIndex`. The new `offset` cap and integer coercion address the DoS angle too.

Remaining items for this round, ranked roughly:

**Worth a decision/test**
- The deletion-status filter is uniform: scheduled-for-deletion accounts are now invisible even on the exact-handle and `following=true` paths. If that's intentional (full hide during grace window), please lock it in with tests; if not, the exact/following branches probably shouldn't share `applySearchableAccountFilters`'s strict-null check.

**Polish / future-proofing**
- The route's `database.getActorFromUsername(handle)` is case-sensitive while `searchAccountIds` is now case-insensitive. For mixed-case stored handles the route's lookup is dead code, and worse it triggers a needless webfinger+`recordActorIfNeeded` when `resolve=true`.
- More broadly, the two `database.getActorFromUsername(...)` branches in the route now duplicate work `getExactAccountIds` already does. Could collapse to "if `resolve=true` and `searchAccountIds` returns empty, do webfinger fallback."
- `indexActorSearchDocument` has two implicit modes (provided actor → upsert that shape; bare id → re-read or delete). Worth a one-line doc comment so callers don't pass stale snapshots.
- `reindexSearchAccounts` can lose concurrent writes if a write fires between SELECT and upsert. Live `applySearchableAccountFilters` masks the deletion-status case, but `documentText`/`name` updates can be served stale until the next event. Either add an `updatedAt` guard on the merge or document that reindex is best-effort.

---

lib/database/sql/search/account.ts
llun: Addressed in c6391d2f. Kept the stricter behavior intentionally: scheduled-deletion accounts stay hidden from account search, including exact-handle and following=true paths. Added explicit coverage for the exact-handle + following case before and after scheduling.

app/api/v1/accounts/search/route.ts
llun: Fixed in c6391d2f. The route now passes q through to searchAccountIds with localDomain and removed both getActorFromUsername branches. For resolve=true, it only does WebFinger on a first-page empty indexed result, then retries with the resolved actor id.

llun: Fixed in c6391d2f. The route now passes q through to searchAccountIds with localDomain and removed both getActorFromUsername branches. For resolve=true, it only does WebFinger on a first-page empty indexed result, then retries with the resolved actor id.

lib/database/sql/search/account.ts
llun: Addressed in c6391d2f. Added a comment above indexActorSearchDocument documenting the two modes: a provided actor row indexes that exact fresh shape, while id-only re-reads actors and deletes the search document if the row is gone.

app/api/v1/accounts/search/route.ts
```diff
@@ -80,31 +83,31 @@ export const GET = traceApiRoute(
               : null
```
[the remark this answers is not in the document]
llun: Fixed in c6391d2f. Removed the case-sensitive route pre-check entirely; resolve now lets searchAccountIds do the case-insensitive local/indexed lookup first, so mixed-case local hits do not trigger WebFinger. Added route coverage for that path.

lib/database/sql/search/account.ts
llun: Addressed in c6391d2f using the documentation option. Added a comment that account reindexing is a best-effort snapshot and live write paths refresh their own search documents, so operators know concurrent writes can temporarily race a reindex batch.

app/api/v1/accounts/search/route.ts
llun: Fixed in c6391d2f. The route no longer calls getActorFromUsername for handle pre-checks; it defers to searchAccountIds first and only falls back to WebFinger after an empty first-page indexed search. Added mixed-case resolve coverage.

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,313 @@
...
+        normalizedQuery,
+        'actors.username',
+        normalizedQuery,
+        'actors.username',
+        'actors.domain',
+        normalizedQueryLikePattern,
+        'actors.username',
+        normalizedQueryLikePattern
+      ]
+    )
```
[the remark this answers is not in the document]
llun: Addressed in c6391d2f. Refactored account ordering so the CASE predicates are generated from structured cases with column constants and ?? identifier bindings, instead of maintaining one hand-aligned flat binding array.

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,313 @@
...
+  const rows = actors.map((actor) =>
+    getActorSearchDocumentRow(actor, currentTime)
+  )
+
+  for (const chunk of chunkArray(rows, ACCOUNT_SEARCH_DOCUMENT_BATCH_SIZE)) {
+    await database(SEARCH_DOCUMENTS_TABLE)
+      .insert(chunk)
+      .onConflict('id')
+      .merge(SEARCH_DOCUMENT_MERGE_COLUMNS)
+  }
```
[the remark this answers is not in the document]
llun: Fixed in c6391d2f. SQLite account-search upsert batch size is now computed from the search-document row column count using the 999 parameter limit. Added coverage that the current 83-row SQLite batch stays in one insert.

app/api/v1/accounts/search/route.ts
```diff
@@ -57,54 +51,57 @@ export const GET = traceApiRoute(
...
+          exactActorIds,
+          ...(following ? { followingActorId: context.currentActor.id } : {})
+        }
+      }
 
-      // Try exact match first (username@domain or just username)
-      if (query.includes('@')) {
+      let indexedIds = await database.searchAccountIds(getSearchParams())
+      if (resolve && offset === 0 && indexedIds.length === 0) {
         const handle = parseAccountHandle(query)
```
llun: The new resolve fallback only fires when `indexedIds.length === 0`. That's clean for the common case (unknown handle → empty page → webfinger), but there's an edge case worth thinking about:

If the user searches `@charlie@remote.test` with `resolve=true` and the *exact* charlie isn't local, but the broad text search happens to return an unrelated account whose document text contains all three tokens (`charlie`, `remote`, `test`) — e.g. a local actor named "Charlie Brown" whose bio mentions "remote test" — then `indexedIds.length > 0`, webfinger is skipped, and the user never reaches the actor they actually typed.

It's a narrow case (the tokenization in `applySearchDocumentFilter` requires all tokens to match), but a real one. Two options:

1. Always fire webfinger when `resolve=true && offset === 0 && handle parses && the exact-handle target isn't already in the first page of `indexedIds`. That keeps cost bounded but guarantees the typed handle gets a chance.
2. Document explicitly that `resolve` is a "last resort when nothing matched at all" knob, which differs slightly from Mastodon's "attempt webfinger lookup" wording.

Not a blocker — the test on line 187 ("does not webfinger when indexed search finds a mixed-case handle locally") locks in the current behavior, but it tests the case where the indexed search returned the *correct* actor. A symmetric test where the broad search returns a *wrong* match would be useful to make this trade-off explicit.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,372 @@
...
+  query.whereExists(function () {
+    this.select(database.raw('1'))
+      .from('follows')
+      .where('follows.actorId', followingActorId)
+      .whereRaw('?? = ??', ['follows.targetActorId', 'actors.id'])
+      .where('follows.status', FollowStatus.enum.Accepted)
+  })
+}
+
+const getExactAccountIds = async ({
```
llun, line 203: `getExactAccountIds` now takes both `q` and `handle` even though `handle = parseAccountHandle(q)` in the caller and the local-username derivation also re-parses `q`. Two small consequences:

1. Callers that forget to keep `handle` and `q` in sync will silently get wrong exact matches. There's no asymmetry today, but the redundant parameter is a foot-gun.
2. `q.trim()` runs twice (route already trimmed before calling).

Could compute `handle` and `localUsername` internally from `q` + `localDomain` — that closes the foot-gun and removes one parameter:

```ts
const getExactAccountIds = async ({ database, q, localDomain, exactActorIds, followingActorId }) => {
  const handle = parseAccountHandle(q)
  const trimmed = q.trim().replace(/^@/, '')
  const localUsername = !handle && localDomain && !trimmed.includes('@') ? trimmed : null
  ...
}
```

Minor — current code is correct.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,372 @@
...
+  exactActorIds,
+  followingActorId
+}: {
+  database: Knex
+  q: string
+  handle: ReturnType<typeof parseAccountHandle>
+  localDomain?: string | null
+  exactActorIds: string[]
+  followingActorId?: string | null
+}) => {
```
llun, line 215: Worth double-checking the `localUsername` derivation against the route's input shape. The route already trims `q` to `query` and forwards `q: query` here. The condition `!q.includes('@')` then runs on the already-trimmed value, which is fine — but `q.trim().replace(/^@/, '')` runs `trim()` redundantly, and the `replace(/^@/, '')` is unreachable since the branch only fires when `!q.includes('@')`.

In other words, on this branch `q.trim().replace(/^@/, '')` simplifies to `q.trim()` (and `q.trim() === q` if the caller already trimmed). Not a bug, just dead branches that could be cleaned up:

```ts
const localUsername =
  !handle && localDomain && !q.includes('@') ? q.trim() : null
```

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,372 @@
...
+    .orderBy('search_documents.entityId', 'asc')
+}
+
+const chunkArray = <T>(items: T[], size: number) => {
+  const chunks: T[][] = []
+  for (let index = 0; index < items.length; index += size) {
+    chunks.push(items.slice(index, index + size))
+  }
+  return chunks
+}
```
llun, line 169: `applySearchableAccountFilters` still uses `whereNull('actors.deletionStatus')`, which hides any account with `deletionStatus !== null` — including `'scheduled'` — from both the broad search *and* the exact-handle/following branches.

Carrying this forward from the previous round: if this is the intended product behavior (full hide during the deletion grace window, even for followers), please add tests for the exact-handle and `following=true` paths that confirm scheduled accounts are gone — right now the only deletion-related test in `search.test.ts` covers `startActorDeletion` via broad search. Without that coverage, a future refactor could quietly re-expose scheduled accounts.

If it's *not* intended (followers should still see their scheduled-deletion accounts), split the deletion filter so only the broad-discoverable path enforces `deletionStatus IS NULL`, while the exact-handle and following branches allow `'scheduled'` and only reject `'deleting'`.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,372 @@
...
+    followingActorId,
+    exactActorIds = []
+  }: SearchAccountsParams
+): Promise<string[]> => {
+  const handle = parseAccountHandle(q)
+  const normalizedQuery = q.trim().replace(/^@/, '').toLowerCase()
+  const normalizedExactActorIds = [...new Set(exactActorIds)]
+  // Only visible exact matches participate in pagination; filtered exact IDs
+  // fall back to the indexed result window instead of reserving page slots.
+  const exactResultIds = await getExactAccountIds({
```
llun, line 331: `searchAccountIds` now has a new code path triggered by `localDomain` — local-username exact match without an `@`. That path needs DB-level test coverage:

- A local account stored under mixed casing (e.g. `Runner`) reachable by lowercase query `runner` when `localDomain` is passed.
- Confirmation that the path is skipped when `localDomain` is omitted (so non-route callers don't accidentally pick up local exact matches).
- A non-discoverable local account found by username when `localDomain` is passed.

The route-level tests in `route.test.ts` cover the route's behavior, but the DB integration tests in `search.test.ts` (where `localDomain` isn't being passed in older tests) don't exercise the new branch directly. Add tests to lock the contract in before something downstream relies on it.

---

lib/database/sql/search/account.ts
```diff
@@ -1,22 +1,372 @@
...
+const getClientName = (database: Knex) =>
+  String(database.client.config.client).toLowerCase()
+
+const getAccountSearchDocumentBatchSize = (
+  database: Knex,
+  row: Record<string, unknown>
+) => {
+  const columnCount = Math.max(Object.keys(row).length, 1)
+  if (SQLITE_CLIENTS.has(getClientName(database))) {
+    return Math.max(1, Math.floor(SQLITE_INSERT_PARAMETER_LIMIT / columnCount))
```
llun, line 100: `getAccountSearchDocumentBatchSize` is a nice safety addition for SQLite's 999-parameter limit. One small note: the batch-size derivation uses `Math.floor(SQLITE_INSERT_PARAMETER_LIMIT / columnCount)` directly. SQLite's actual default `SQLITE_MAX_VARIABLE_NUMBER` is 999 on older builds (≤3.32) but 32766 on newer builds. The hard-coded 999 is the safe lower bound — that's correct — but it means a modern SQLite will under-batch by ~33x. Not a problem for correctness, but reindex/bulk-create performance on big imports will be limited. Probably fine to leave as-is and revisit only if reindex throughput becomes a bottleneck.

---

llun (commented):
Round 3 review of c6391d2f.

The route-level refactor is a big improvement — collapsing both `database.getActorFromUsername(...)` branches into a single `searchAccountIds` call, with webfinger as a clean fallback when nothing comes back, removes a chunk of duplicated lookup logic. The case-sensitivity asymmetry is now sidestepped entirely (the route never calls the case-sensitive `getActorFromUsername`), and the `indexActorSearchDocument` / `reindexSearchAccounts` doc comments address the contract concerns from round 2. The dynamic SQLite-aware batch size on the upsert helper is a nice touch.

Items for this round:

**Worth a decision/test**
- The resolve fallback now only fires when `indexedIds.length === 0`. If the broad text search returns an unrelated all-tokens-match (rare but possible), the typed handle won't be webfinger-resolved. Either accept the trade-off and add a "false-positive masks webfinger" test to lock it in, or flip the predicate to "exact-handle target isn't in the first page."
- The deletion-status filter is still uniform — scheduled-for-deletion accounts are invisible to followers and exact-handle lookups. Same flag as round 2; if intended, add tests for the exact and following paths so it can't quietly regress.

**Polish**
- `getExactAccountIds` takes both `q` and `handle` (which is `parseAccountHandle(q)`). Worth deriving `handle` internally from `q` to remove the foot-gun.
- The `q.trim().replace(/^@/, '')` on the local-username branch has dead pieces — that branch only fires when `!q.includes('@')`, so the `replace(/^@/, '')` is unreachable.
- New `localDomain`-driven local-username code path in `searchAccountIds` needs DB-level test coverage (mixed-case match, omitted-localDomain skip, non-discoverable local find). The route tests cover the route, but the integration tests in `search.test.ts` don't exercise this branch yet.
- SQLite batch size uses the conservative 999-parameter limit. Modern SQLite (≥3.32) supports 32766. Fine to leave as-is, just noting if reindex throughput ever matters.

---

lib/database/sql/search/account.ts
llun: Addressed in f88e937c: local username derivation now uses the trimmed query directly and only enters that branch when a normalized localDomain exists and the query has no @.

lib/database/sql/search/account.ts
llun: Addressed in f88e937c: getExactAccountIds now derives parseAccountHandle(q) and the local username internally from q + localDomain, so callers no longer need to keep q and handle in sync.

app/api/v1/accounts/search/route.ts
llun: Addressed in dac06688: the route now materializes the first indexed page, compares returned acct values against the parsed handle, and only skips WebFinger when the exact handle is already present. Added a regression test for unrelated broad-search matches masking a remote handle.

lib/database/sql/search/account.ts
llun: Leaving this intentionally conservative for now. The 999-parameter limit is the safe lower bound across older SQLite builds; it keeps correctness portable, and we can raise or detect the runtime limit later if reindex throughput becomes a real bottleneck.

lib/database/sql/search/account.ts
llun: Addressed in f88e937c: the SQL tests now cover non-discoverable local username lookup with localDomain, the omitted-localDomain skip case, and mixed-case stored username lookup via the localDomain path.

PR #763, opened 2026-05-26
fix: delegate OAuth authorize through Better Auth

llun:
## Summary
- Delegate authenticated unsigned `/oauth/authorize` requests through Better Auth so OAuth queries are validated and signed before consent.
- Serialize signed consent `oauth_query` params in Better Auth signature order and follow normalized consent endpoint `url` responses.
- Persist the selected consent actor before approval so Better Auth stores that actor id in OAuth token `referenceId`; bearer-token auth now fails closed when an opaque token has no actor reference.
- Add `push` to usable OAuth scopes, advertise it in app registration/well-known metadata, and guard `POST`/`DELETE /api/v1/push/subscribe` with `OAuthGuard([Scope.enum.push])`.

## Verification
- `PATH=/Users/llun/.nvm/versions/node/v24.15.0/bin:$PATH yarn test --runTestsByPath lib/services/guards/OAuthGuard.test.ts 'app/(nosidebar)/oauth/authorize/authorizeQuery.test.ts' 'app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx' app/api/v1/push/subscribe/route.test.ts lib/services/wellknown/index.test.ts app/api/v1/apps/createApplication.test.ts --runInBand`
- `PATH=/Users/llun/.nvm/versions/node/v24.15.0/bin:$PATH yarn lint` (passes with existing `any` warnings in unrelated files)
- `PATH=/Users/llun/.nvm/versions/node/v24.15.0/bin:$PATH yarn build`
- `PATH=/Users/llun/.nvm/versions/node/v24.15.0/bin:$PATH yarn test --runInBand --silent` (334 suites / 2870 tests passed earlier after the unrelated heatmap flake passed on rerun)
- `activities.local` API smoke with existing local account plus a local OAuth app: unsigned authorize delegates to Better Auth, `prompt=consent` returns a signed consent URL with `sig` and `exp`, signed consent POST returns 200 with a Phanpy redirect code, token exchange returns 200 with `read write follow push`, the issued token stores `referenceId=https://activities.local/users/testactor2` after switching to that actor, unauthenticated push subscribe returns 401, OAuth bearer-token push subscribe/delete return 200, and browser-session push subscribe/delete with same-origin `Origin` return 200.

## Compatibility Note
- Existing OAuth bearer tokens without the `push` scope will not be authorized for `POST`/`DELETE /api/v1/push/subscribe`; clients need to request a fresh token including `push`. Browser-session calls are still supported through `OAuthGuard`'s session fallback when the request includes same-origin proof.

## Notes
- The full local dev stack rebuild hit Docker VM disk pressure while building Pixelfed, so validation used the existing Activity.next/Traefik/Postgres services.

### DIFF
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -139,7 +139,14 @@ describe('AuthorizeCard', () => {
         redirect: true,
         url: '//phanpy.local/?code=oauth-code'
       })
-    ).toBe('http://phanpy.local/?code=oauth-code')
+    ).toBeUndefined()
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: '/oauth/callback?code=oauth-code'
+      })
+    ).toBeUndefined()
   })
 
   it('submits denial with the signed Better Auth query and follows url redirects', async () => {
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -11,6 +11,7 @@ import { AuthorizeCard, getConsentRedirectUrl } from './AuthorizeCard'
 import { SearchParams } from './types'
 
 const mockPush = jest.fn()
+const mockNavigate = jest.fn()
 
 jest.mock('next/navigation', () => ({
   useRouter: () => ({ push: mockPush })
@@ -55,6 +56,7 @@ const signedSearchParams: SearchParams = {
 describe('AuthorizeCard', () => {
   beforeEach(() => {
     mockPush.mockReset()
+    mockNavigate.mockReset()
     global.fetch = jest.fn().mockResolvedValue({
       ok: true,
       json: async () => ({})
@@ -72,6 +74,7 @@ describe('AuthorizeCard', () => {
         searchParams={signedSearchParams}
         actors={actors}
         currentActorId="https://activities.local/users/llun"
+        navigate={mockNavigate}
       />
     )
 
@@ -119,8 +122,91 @@ describe('AuthorizeCard', () => {
     expect(
       getConsentRedirectUrl({
         redirect: true,
+        url: 'https://phanpy.local/?code=oauth-code',
+        redirect_uri: 'https://legacy.example/?code=legacy-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
         url: 'javascript:alert(1)'
       })
     ).toBeUndefined()
+  })
+
+  it('submits denial with the signed Better Auth query and follows url redirects', async () => {
+    ;(global.fetch as jest.Mock).mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
+        url: 'https://phanpy.local/?error=access_denied&state=return-state'
+      })
+    })
+
+    render(
+      <AuthorizeCard
+        client={client}
+        searchParams={signedSearchParams}
+        actors={actors}
+        currentActorId="https://activities.local/users/llun"
+        navigate={mockNavigate}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Deny' }))
+
+    await waitFor(() => {
+      expect(global.fetch).toHaveBeenCalledWith(
+        '/api/auth/oauth2/consent',
+        expect.objectContaining({
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' }
+        })
+      )
+    })
+
+    const [, requestInit] = (global.fetch as jest.Mock).mock.calls[0]
+    const body = JSON.parse(requestInit.body)
+
+    expect(body.accept).toBe(false)
+    expect(body.oauth_query).toBe(
+      'response_type=code' +
+        '&client_id=phanpy-client' +
+        '&redirect_uri=not-a-url' +
+        '&scope=read+write+follow+push' +
+        '&state=return-state' +
+        '&code_challenge=challenge' +
+        '&code_challenge_method=S256' +
+        '&exp=1779800000' +
+        '&sig=signed-query'
+    )
+    await waitFor(() => {
+      expect(mockNavigate).toHaveBeenCalledWith(
+        'https://phanpy.local/?error=access_denied&state=return-state'
+      )
+    })
+  })
+
+  it('falls back to access_denied redirect when denial has no redirect URL', async () => {
+    render(
+      <AuthorizeCard
+        client={client}
+        searchParams={{
+          ...signedSearchParams,
+          redirect_uri: 'https://phanpy.local/'
+        }}
+        actors={actors}
+        currentActorId="https://activities.local/users/llun"
+        navigate={mockNavigate}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Deny' }))
+
+    await waitFor(() => {
+      expect(mockNavigate).toHaveBeenCalledWith(
+        'https://phanpy.local/?error=access_denied&state=return-state'
+      )
+    })
   })
 })
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -115,5 +115,12 @@ describe('AuthorizeCard', () => {
         url: 'https://phanpy.local/?code=oauth-code'
       })
     ).toBe('https://phanpy.local/?code=oauth-code')
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: 'javascript:alert(1)'
+      })
+    ).toBeUndefined()
   })
 })
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -41,6 +41,17 @@ const actors = [
   }
 ] as unknown as Actor[]
 
+const alternateActors = [
+  actors[0],
+  {
+    id: 'https://activities.local/users/testactor2',
+    username: 'testactor2',
+    domain: 'activities.local',
+    name: 'testactor2',
+    iconUrl: null
+  }
+] as unknown as Actor[]
+
 const signedSearchParams: SearchParams = {
   client_id: 'phanpy-client',
   redirect_uri: 'not-a-url',
@@ -95,7 +106,10 @@ describe('AuthorizeCard', () => {
       )
     })
 
-    const [, requestInit] = (global.fetch as jest.Mock).mock.calls[0]
+    const consentCall = (global.fetch as jest.Mock).mock.calls.find(
+      ([url]) => url === '/api/auth/oauth2/consent'
+    )
+    const [, requestInit] = consentCall
     const body = JSON.parse(requestInit.body)
 
     expect(body.accept).toBe(true)
@@ -109,6 +123,39 @@ describe('AuthorizeCard', () => {
     expect(oauthQuery.get('code_challenge_method')).toBe('S256')
     expect(oauthQuery.get('sig')).toBe('signed-query')
     expect(oauthQuery.get('exp')).toBe('1779800000')
+  })
+
+  it('persists the selected actor before approving consent', async () => {
+    render(
+      <AuthorizeCard
+        client={client}
+        searchParams={signedSearchParams}
+        actors={alternateActors}
+        currentActorId="https://activities.local/users/testactor2"
+        navigate={mockNavigate}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Approve' }))
+
+    await waitFor(() => expect(global.fetch).toHaveBeenCalledTimes(2))
+
+    const [switchUrl, switchRequestInit] = (global.fetch as jest.Mock).mock
+      .calls[0]
+    expect(switchUrl).toBe('/api/v1/actors/switch')
+    expect(switchRequestInit).toEqual(
+      expect.objectContaining({
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          actorId: 'https://activities.local/users/testactor2'
+        })
+      })
+    )
+
+    expect((global.fetch as jest.Mock).mock.calls[1][0]).toBe(
+      '/api/auth/oauth2/consent'
+    )
   })
 
   it('reads Better Auth consent redirect URLs from url responses', () => {
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -1,0 +1,119 @@
+/**
+ * @jest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import { Actor } from '@/lib/types/domain/actor'
+import { Client } from '@/lib/types/oauth2/client'
+
+import { AuthorizeCard, getConsentRedirectUrl } from './AuthorizeCard'
+import { SearchParams } from './types'
+
+const mockPush = jest.fn()
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({ push: mockPush })
+}))
+
+const client: Client = {
+  id: 'db-client-id',
+  clientId: 'phanpy-client',
+  clientSecret: null,
+  name: 'Phanpy',
+  redirectUris: ['https://phanpy.local/'],
+  scopes: ['read', 'write', 'follow', 'push'],
+  website: 'https://phanpy.local',
+  requirePKCE: true,
+  disabled: false,
+  createdAt: 0,
+  updatedAt: 0
+}
+
+const actors = [
+  {
+    id: 'https://activities.local/users/llun',
+    username: 'llun',
+    domain: 'activities.local',
+    name: 'llun',
+    iconUrl: null
+  }
+] as unknown as Actor[]
+
+const signedSearchParams: SearchParams = {
+  client_id: 'phanpy-client',
+  redirect_uri: 'not-a-url',
+  response_type: 'code',
+  scope: 'read write follow push',
+  state: 'return-state',
+  code_challenge: 'challenge',
+  code_challenge_method: 'S256',
+  sig: 'signed-query',
+  exp: '1779800000'
+}
+
+describe('AuthorizeCard', () => {
+  beforeEach(() => {
+    mockPush.mockReset()
+    global.fetch = jest.fn().mockResolvedValue({
+      ok: true,
+      json: async () => ({})
+    }) as jest.Mock
+  })
+
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  it('submits selected Phanpy scopes with the signed Better Auth query', async () => {
+    render(
+      <AuthorizeCard
+        client={client}
+        searchParams={signedSearchParams}
+        actors={actors}
+        currentActorId="https://activities.local/users/llun"
+      />
+    )
+
+    expect(screen.getByLabelText('read')).toBeChecked()
+    expect(screen.getByLabelText('write')).toBeChecked()
+    expect(screen.getByLabelText('follow')).toBeChecked()
+    expect(screen.getByLabelText('push')).toBeChecked()
+
+    fireEvent.click(screen.getByRole('button', { name: 'Approve' }))
+
+    await waitFor(() => {
+      expect(global.fetch).toHaveBeenCalledWith(
+        '/api/auth/oauth2/consent',
+        expect.objectContaining({
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' }
+        })
+      )
+    })
+
+    const [, requestInit] = (global.fetch as jest.Mock).mock.calls[0]
+    const body = JSON.parse(requestInit.body)
+
+    expect(body.accept).toBe(true)
+    expect(body.scope).toBe('read write follow push')
+
+    const oauthQuery = new URLSearchParams(body.oauth_query)
+    expect(oauthQuery.get('client_id')).toBe('phanpy-client')
+    expect(oauthQuery.get('scope')).toBe('read write follow push')
+    expect(oauthQuery.get('state')).toBe('return-state')
+    expect(oauthQuery.get('code_challenge')).toBe('challenge')
+    expect(oauthQuery.get('code_challenge_method')).toBe('S256')
+    expect(oauthQuery.get('sig')).toBe('signed-query')
+    expect(oauthQuery.get('exp')).toBe('1779800000')
+  })
+
+  it('reads Better Auth consent redirect URLs from url responses', () => {
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: 'https://phanpy.local/?code=oauth-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
+  })
+})
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -139,7 +139,7 @@ describe('AuthorizeCard', () => {
         redirect: true,
         url: '//phanpy.local/?code=oauth-code'
       })
-    ).toBe('https://phanpy.local/?code=oauth-code')
+    ).toBe('http://phanpy.local/?code=oauth-code')
   })
 
   it('submits denial with the signed Better Auth query and follows url redirects', async () => {
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
@@ -133,6 +133,13 @@ describe('AuthorizeCard', () => {
         url: 'javascript:alert(1)'
       })
     ).toBeUndefined()
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: '//phanpy.local/?code=oauth-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
   })
 
   it('submits denial with the signed Better Auth query and follows url redirects', async () => {
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -33,16 +33,18 @@ interface Props {
   searchParams: SearchParams
   actors: Actor[]
   currentActorId: string
+  navigate?: (url: string) => void
 }
 
 interface ConsentResponse {
   redirect?: boolean
-  redirect_uri?: string
   url?: string
+  // Legacy shape from the original custom consent handler.
+  redirect_uri?: string
 }
 
 export const getConsentRedirectUrl = (data: ConsentResponse) => {
-  const redirectUrl = data.redirect_uri ?? data.url
+  const redirectUrl = data.url ?? data.redirect_uri
   if (!redirectUrl) return undefined
 
   try {
@@ -57,11 +59,16 @@ export const getConsentRedirectUrl = (data: ConsentResponse) => {
   return undefined
 }
 
+const navigateTo = (url: string) => {
+  window.location.href = url
+}
+
 export const AuthorizeCard: FC<Props> = ({
   searchParams,
   client,
   actors,
-  currentActorId
+  currentActorId,
+  navigate = navigateTo
 }) => {
   const requestedScopes = searchParams.scope.split(' ')
   const router = useRouter()
@@ -109,7 +116,7 @@ export const AuthorizeCard: FC<Props> = ({
         if (searchParams.state) {
           errorUrl.searchParams.set('state', searchParams.state)
         }
-        window.location.href = errorUrl.toString()
+        navigate(errorUrl.toString())
         return
       } catch {
         // Malformed redirect_uri — fall through to router push
@@ -140,7 +147,7 @@ export const AuthorizeCard: FC<Props> = ({
         const data = (await response.json()) as ConsentResponse
         const redirectUrl = getConsentRedirectUrl(data)
         if (redirectUrl) {
-          window.location.href = redirectUrl
+          navigate(redirectUrl)
           return
         }
       }
@@ -168,7 +175,7 @@ export const AuthorizeCard: FC<Props> = ({
         const data = (await response.json()) as ConsentResponse
         const redirectUrl = getConsentRedirectUrl(data)
         if (redirectUrl) {
-          window.location.href = redirectUrl
+          navigate(redirectUrl)
           return
         }
         // Denial processed but no redirect — use access_denied
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -43,12 +43,19 @@ interface ConsentResponse {
   redirect_uri?: string
 }
 
+const getConsentRedirectBaseUrl = () => {
+  if (typeof window === 'undefined') return 'https://activities.local'
+  return window.location.origin === 'http://localhost'
+    ? 'https://activities.local'
+    : window.location.origin
+}
+
 export const getConsentRedirectUrl = (data: ConsentResponse) => {
   const redirectUrl = data.url ?? data.redirect_uri
   if (!redirectUrl) return undefined
 
   try {
-    const parsed = new URL(redirectUrl, 'https://activities.local')
+    const parsed = new URL(redirectUrl, getConsentRedirectBaseUrl())
     if (parsed.protocol === 'http:' || parsed.protocol === 'https:') {
       return parsed.toString()
     }
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -108,6 +108,15 @@ export const AuthorizeCard: FC<Props> = ({
     }
   }
 
+  const persistSelectedActor = async () => {
+    const response = await fetch('/api/v1/actors/switch', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ actorId: selectedActorId })
+    })
+    return response.ok
+  }
+
   const redirectWithError = (error: string) => {
     const redirectUri = searchParams.redirect_uri
     if (redirectUri) {
@@ -133,6 +142,10 @@ export const AuthorizeCard: FC<Props> = ({
     try {
       const formData = new FormData(e.currentTarget)
       const selectedScopes = formData.getAll('scope') as string[]
+      if (!(await persistSelectedActor())) {
+        redirectWithError('server_error')
+        return
+      }
 
       const response = await fetch('/api/auth/oauth2/consent', {
         method: 'POST',
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -50,7 +50,7 @@ export const getConsentRedirectUrl = (data: ConsentResponse) => {
   try {
     const parsed = new URL(redirectUrl, 'https://activities.local')
     if (parsed.protocol === 'http:' || parsed.protocol === 'https:') {
-      return redirectUrl
+      return parsed.toString()
     }
   } catch {
     // Invalid redirect response; fall through to the existing error handling.
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -45,9 +45,7 @@ interface ConsentResponse {
 
 const getConsentRedirectBaseUrl = () => {
   if (typeof window === 'undefined') return 'https://activities.local'
-  return window.location.origin === 'http://localhost'
-    ? 'https://activities.local'
-    : window.location.origin
+  return window.location.origin
 }
 
 export const getConsentRedirectUrl = (data: ConsentResponse) => {
--- a/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
+++ b/app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
@@ -25,6 +25,7 @@ import { UsableScopes } from '@/lib/types/database/operations'
 import { Actor } from '@/lib/types/domain/actor'
 import { Client } from '@/lib/types/oauth2/client'
 
+import { buildOAuthQuery } from './authorizeQuery'
 import { SearchParams } from './types'
 
 interface Props {
@@ -33,6 +34,15 @@ interface Props {
   actors: Actor[]
   currentActorId: string
 }
+
+interface ConsentResponse {
+  redirect?: boolean
+  redirect_uri?: string
+  url?: string
+}
+
+export const getConsentRedirectUrl = (data: ConsentResponse) =>
+  data.redirect_uri ?? data.url
 
 export const AuthorizeCard: FC<Props> = ({
   searchParams,
@@ -77,14 +87,6 @@ export const AuthorizeCard: FC<Props> = ({
     }
   }
 
-  const buildOAuthQuery = () => {
-    const oauthQuery = new URLSearchParams()
-    for (const [key, value] of Object.entries(searchParams)) {
-      if (value !== undefined) oauthQuery.set(key, String(value))
-    }
-    return oauthQuery.toString()
-  }
-
   const redirectWithError = (error: string) => {
     const redirectUri = searchParams.redirect_uri
     if (redirectUri) {
@@ -117,14 +119,15 @@ export const AuthorizeCard: FC<Props> = ({
         body: JSON.stringify({
           accept: true,
           scope: selectedScopes.join(' '),
-          oauth_query: buildOAuthQuery()
+          oauth_query: buildOAuthQuery(searchParams)
         })
       })
 
       if (response.ok) {
-        const data = (await response.json()) as { redirect_uri?: string }
-        if (data.redirect_uri) {
-          window.location.href = data.redirect_uri
+        const data = (await response.json()) as ConsentResponse
+        const redirectUrl = getConsentRedirectUrl(data)
+        if (redirectUrl) {
+          window.location.href = redirectUrl
           return
         }
       }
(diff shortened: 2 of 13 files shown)


llun:
@codex review

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,49 @@
...
+  for (const [key, value] of Object.entries(values)) {
+    if (
+      value !== undefined &&
+      !BetterAuthAuthorizationParamOrder.includes(key)
+    ) {
+      oauthQuery.set(key, value)
+    }
+  }
+  return oauthQuery.toString()
+}
```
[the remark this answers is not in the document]
llun: Applied in 177d1d53: `buildOAuthQuery` now skips both `null` and `undefined`, with a regression test covering nullish serialization.

llun:
@codex review

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+  'client_id',
+  'redirect_uri',
+  'scope',
+  'state',
+  'request_uri',
+  'code_challenge',
+  'code_challenge_method',
+  'nonce',
+  'prompt',
+  'exp',
```
llun, line 20: `request_uri` is listed in the Better Auth signature order here, but it is **not** declared in the `SearchParams` zod schema (`types.ts`). Since `page.tsx` now passes `parsedResult.data` through this function, zod will strip `request_uri` before it ever reaches `buildOAuthQuery`. If Better Auth ever signs a query that includes `request_uri` (e.g. for Pushed Authorization Requests), the signature will silently fail to round-trip through the consent submission.

Either add `request_uri: z.string().optional()` to `SearchParams`, or drop it from this list to make the gap explicit.

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+  params: SearchParams,
+  baseUrl: string
+): string => {
+  const url = new URL('/api/auth/oauth2/authorize', baseUrl)
+  url.search = buildOAuthQuery(params)
+  return url.toString()
+}
+
+export const shouldDelegateToBetterAuth = (params: SearchParams): boolean =>
+  !params.sig || !params.exp
```
llun: `shouldDelegateToBetterAuth` returns true whenever **either** `sig` or `exp` is missing. That has two side effects worth thinking about:

1. **Partial-signature leak through delegation.** When only one of `sig`/`exp` is present, `buildBetterAuthAuthorizeUrl` will still carry that stray param into the Better Auth `/api/auth/oauth2/authorize` request. Better Auth should ignore it, but it's cleaner to strip both before delegating so we don't accidentally re-use a stale signature fragment.
2. **No loop guard.** If Better Auth ever bounces back to `/oauth/authorize` without setting **both** `sig` and `exp` (misconfiguration, version mismatch, etc.), the page will redirect back to Better Auth in an infinite loop. Consider tracking a one-shot marker (e.g. `prompt=consent` already-present, or a short-lived cookie) so we fail closed instead of looping.

At minimum the comment/JSDoc should explain *why* both params are required — right now the predicate reads like "any missing signing field means re-delegate," which is non-obvious.

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,46 @@
+import { SearchParams } from './types'
```
llun, line 1: The exact ordering in `BetterAuthAuthorizationParamOrder` is load-bearing — if Better Auth changes its canonicalization the signature check on the consent endpoint will break and the failure mode (consent 401/400 with no obvious cause) is hard to diagnose. Please add a comment pointing to the Better Auth source/version this order was derived from, so a future bump to the dependency can be cross-checked.

Also, the second `for` loop that picks up "extra" keys (lines 25–29) is dead code in practice: callers pass `parsedResult.data`, which zod has already restricted to the known `SearchParams` keys. Either remove the loop or document that it exists to defend against passing raw search params.

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -34,6 +35,15 @@ interface Props {
   currentActorId: string
 }
 
+interface ConsentResponse {
+  redirect?: boolean
+  redirect_uri?: string
+  url?: string
```
llun, line 41: `getConsentRedirectUrl` accepts both `redirect_uri` and `url` for backwards compatibility. Two things:

- The `ConsentResponse` `redirect_uri` field appears to only exist to support the old shape — if there are no longer any code paths that produce it (Better Auth uses `url`), remove it instead of carrying a shim. If there is still a legacy producer, add a one-line comment naming it.
- If both fields are ever present, `redirect_uri ?? data.url` silently prefers the legacy one. That's probably wrong now that Better Auth is the source of truth. Prefer `data.url ?? data.redirect_uri` (or drop the alias entirely).

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  it('reads Better Auth consent redirect URLs from url responses', () => {
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: 'https://phanpy.local/?code=oauth-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
+  })
+})
```
llun, line 273: Test coverage for the deny path is missing. The deny branch in `AuthorizeCard.tsx` (lines 144–165 in the new file) was changed in the same way as the approve branch (`redirect_uri ?? url`, new `oauth_query` builder), but only the approve path is exercised here. Please add a deny case that:

1. Asserts the POST body has `accept: false` and the correctly-ordered `oauth_query`.
2. Asserts the redirect follows `data.url` when `redirect_uri` is absent.
3. Asserts the `access_denied` fallback fires when the response has neither.

---

app/(nosidebar)/oauth/authorize/page.tsx
```diff
@@ -48,13 +54,14 @@ const Page: FC<Props> = async ({ searchParams }) => {
...
-      `/oauth/authorize?${new URLSearchParams(Object.entries(params).filter(([, v]) => v !== undefined) as [string, string][])}`
-    )
+    url.searchParams.append('redirectBack', buildOAuthAuthorizePath(params))
     return redirect(url.toString())
   }
 
+  if (shouldDelegateToBetterAuth(params)) {
+    return redirect(buildBetterAuthAuthorizeUrl(params, getBaseURL()))
+  }
+
```
llun: The delegation check runs **after** the actor/client lookup. That's two database round-trips (`getActorFromSession` and `getClientFromDatabase`) that are thrown away for every unsigned authorize request, since the response is just a 307 to Better Auth.

Consider checking `shouldDelegateToBetterAuth(params)` immediately after the session check (and after confirming there is an `actor.account`) so the unsigned path can short-circuit before fetching the client. This is the hot path for every Mastodon-compatible OAuth dance, so it's worth saving.

---

lib/types/database/operations.ts
```diff
@@ -1322,7 +1322,8 @@ export const UsableScopes = [
   Scope.enum['write:bookmarks'],
   Scope.enum['write:statuses'],
   Scope.enum['write:conversations'],
-  Scope.enum.follow
+  Scope.enum.follow,
+  Scope.enum.push
```
llun, line 1326: Adding `push` to `UsableScopes` exposes it to every OAuth flow that derives its scope list from this constant (well-known metadata, app registration, consent UI). Worth confirming that the access token issued with `push` is actually wired into the Web Push endpoints — otherwise clients will request the scope, get it granted, and then 403 on `/api/v1/push/subscription`. If the Push endpoints aren't implemented yet this should at least be called out in the PR body so it's not a hidden expectation.

---

llun (commented):
## Review summary

This PR cleanly delegates unsigned `/oauth/authorize` requests through Better Auth and centralizes query serialization. The refactor reads well and the unit tests for `authorizeQuery` cover the important serialization edges. Findings below are mostly correctness/robustness, not blockers:

**Things to fix**
- `request_uri` is in the signature-order list but missing from the `SearchParams` zod schema — it will be stripped before serialization and silently break signature round-trip if Better Auth ever sends it.
- `shouldDelegateToBetterAuth` returns true when **either** `sig` or `exp` is missing, with no loop guard and no stripping of the surviving param before re-delegating. Worth a comment and (ideally) a loop-detection fallback.
- Deny path in `AuthorizeCard` got the same `redirect_uri ?? url` / new `oauth_query` treatment as the approve path, but the new test only covers approve. Please add a deny test.

**Smaller / nice-to-have**
- The `data.redirect_uri ?? data.url` preference order favors the legacy field; flip it or drop `redirect_uri` if nothing produces it any more.
- Hoist the `shouldDelegateToBetterAuth` short-circuit above the actor/client DB fetches in `page.tsx` — the unsigned path is the hot path and currently does two unused round-trips per redirect.
- Add a source/version pointer next to `BetterAuthAuthorizationParamOrder`, and either remove or document the dead "extra keys" loop in `buildOAuthQuery`.
- Confirm (in the PR body or a follow-up) that granting the `push` scope is wired to a working Web Push implementation — otherwise clients will get an OAuth-level grant they can't actually exercise.

Inline comments include details for each.

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -34,6 +35,15 @@ interface Props {
...
+}
+
+export const getConsentRedirectUrl = (data: ConsentResponse) =>
+  data.redirect_uri ?? data.url
```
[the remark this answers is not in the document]
llun: Applied in 17e368a8: `getConsentRedirectUrl` now rejects non-http(s) protocols before callers assign the result to `window.location.href`.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+  for (const [key, value] of Object.entries(values)) {
+    if (value != null && !BetterAuthAuthorizationParamOrder.includes(key)) {
+      oauthQuery.set(key, value)
+    }
+  }
+  return oauthQuery.toString()
+}
+
+export const buildOAuthAuthorizePath = (params: SearchParams): string =>
```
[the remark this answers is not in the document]
llun: Applied in 17e368a8: blank `sig`/`exp` values are no longer serialized, and the regression test covers a blank `sig` delegation case.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,46 @@
...
+  'scope',
+  'state',
+  'request_uri',
+  'code_challenge',
+  'code_challenge_method',
+  'nonce',
+  'prompt',
+  'exp',
+  'sig'
+]
```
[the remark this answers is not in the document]
llun: Applied in 17e368a8: `request_uri` is now part of the validated authorize `SearchParams` schema, with coverage that it is preserved in serialized OAuth queries.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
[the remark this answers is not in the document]
llun: Applied in 17e368a8: added a module-level `Set` for the ordered Better Auth params and kept the nullish filtering from the previous fix.

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+  })
+
+  it('reads Better Auth consent redirect URLs from url responses', () => {
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: 'https://phanpy.local/?code=oauth-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
+  })
```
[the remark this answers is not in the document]
llun: Applied in 17e368a8: added a regression assertion that `getConsentRedirectUrl` rejects a `javascript:` URL.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -34,6 +35,28 @@ interface Props {
...
+interface ConsentResponse {
+  redirect?: boolean
+  redirect_uri?: string
+  url?: string
+}
+
+export const getConsentRedirectUrl = (data: ConsentResponse) => {
+  const redirectUrl = data.redirect_uri ?? data.url
+  if (!redirectUrl) return undefined
+
```
llun, line 50: Two follow-ups on the new validator:

1. **Protocol-relative URLs leak through the check.** `new URL('//evil.com/path', 'https://activities.local')` parses to `https://evil.com/path`, the protocol check passes, and the function returns the **raw** `'//evil.com/path'`. Assigning that to `window.location.href` navigates off-origin. Better Auth shouldn't produce such a value, but as a defense-in-depth fix consider returning `parsed.toString()` (the normalized absolute URL) rather than the raw string — that way the only thing that ever reaches `window.location` is an http/https absolute URL you've already parsed.
2. **Preference order still favors the legacy field.** `data.redirect_uri ?? data.url` — this was raised last round and isn't addressed. Now that Better Auth is the source of truth, prefer `data.url ?? data.redirect_uri` (or drop the legacy alias entirely if nothing produces it).

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+  params: SearchParams,
+  baseUrl: string
+): string => {
+  const url = new URL('/api/auth/oauth2/authorize', baseUrl)
+  url.search = buildOAuthQuery(params)
+  return url.toString()
+}
+
+export const shouldDelegateToBetterAuth = (params: SearchParams): boolean =>
+  !params.sig || !params.exp
```
llun: Still no loop guard or JSDoc here. The blank-`sig`/`exp` stripping in `shouldIncludeOAuthParam` is a nice improvement and partly mitigates concern #1 from the previous round (a stray empty param won't make it through), but the predicate itself still reads as "any missing signing field re-delegates," which is non-obvious and would loop forever if Better Auth ever fails to set both fields on bounceback.

Suggested minimum: a one-line comment above `shouldDelegateToBetterAuth` explaining the invariant ("Better Auth always returns both `sig` and `exp` on the signed bounceback; if either is missing, the request hasn't gone through Better Auth yet").

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (shouldIncludeOAuthParam(key, value)) oauthQuery.set(key, value)
+  }
+  for (const [key, value] of Object.entries(values)) {
+    if (
+      shouldIncludeOAuthParam(key, value) &&
+      !BetterAuthAuthorizationParamSet.has(key)
+    ) {
+      oauthQuery.set(key, value)
+    }
+  }
```
llun, line 51: The set-lookup speedup from gemini's review made the "extra keys" loop *faster*, but didn't address the original point: with `parsedResult.data` being passed in, zod has already restricted the keys to the known schema, so this branch never adds anything in practice. Either delete the loop or add a one-line comment naming the (hypothetical) caller that would pass unparsed input — otherwise it's confusing dead code.

Separately, please add a comment somewhere near `BetterAuthAuthorizationParamOrder` linking to the Better Auth source/version the order was derived from. The signing failure mode (consent endpoint returns 401/400) is hard to diagnose, and a dependency bump should be cross-checkable against the canonical list.

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    ).toBe('https://phanpy.local/?code=oauth-code')
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: 'javascript:alert(1)'
+      })
+    ).toBeUndefined()
+  })
+})
```
llun, line 273: Deny path is still untested. The deny branch has the same `redirect_uri ?? url` and `oauth_query` changes as approve; please add a case that exercises `accept: false`, the `data.url`-only redirect, and the `access_denied` fallback when the response has neither field.

---

app/(nosidebar)/oauth/authorize/page.tsx
llun: Still flagged from last round: the unsigned-path short-circuit runs after `getActorFromSession` + `getClientFromDatabase`. Every Mastodon-compatible OAuth dance pays for two DB round-trips it doesn't need before being 307'd to Better Auth. Hoisting the `shouldDelegateToBetterAuth` check above the `Promise.all` (with the session/account check still gating it) is a cheap win on the hot path.

---

llun (commented):
## Second-round review

Nice progress — several items from the first round are resolved:

- ✅ `request_uri` added to the `SearchParams` zod schema with coverage.
- ✅ Blank `sig`/`exp` stripping via `shouldIncludeOAuthParam` prevents partial signatures from leaking through delegation.
- ✅ Open-redirect protection in `getConsentRedirectUrl` (http(s) protocol gate + `javascript:` regression test).
- ✅ `Set` lookup for the param-order list.

Still open from the previous round:

1. **`redirect_uri ?? url` preference order** — Better Auth uses `url`; flip the fallback or drop the legacy alias.
2. **No loop guard / JSDoc on `shouldDelegateToBetterAuth`** — at minimum a one-line comment so a future maintainer doesn't misread the invariant.
3. **`BetterAuthAuthorizationParamOrder` source/version comment** — load-bearing constant with a hard-to-diagnose failure mode.
4. **Dead "extras" loop in `buildOAuthQuery`** — zod strips unknown keys; either delete the loop or document its intent.
5. **Deny-path test missing** in `AuthorizeCard.test.tsx`.
6. **DB lookups still run before the unsigned-path 307** in `page.tsx`.
7. **Push scope wiring confirmation** — please call out in the PR body (or a follow-up issue) that `push` is only being exposed for OAuth here, not yet served by Web Push endpoints, so clients don't 403 later.

New observation:

8. **Protocol-relative URLs slip past the validator.** `new URL('//evil.com/path', base)` parses with protocol `https:`, the check passes, and the raw `//evil.com/path` is returned and assigned to `window.location.href`. Trust in our own backend mostly mitigates this, but returning `parsed.toString()` (the normalized absolute URL) instead of the raw string is a clean defense-in-depth fix.

Inline comments include details.

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in 1650f4f5: `request_uri` is now part of the validated authorize `SearchParams` schema, and the query helper test covers preserving it in the Better Auth signature order.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in 0630c035: delegated Better Auth authorize URLs now strip both `sig` and `exp` whenever the request is treated as unsigned, and the predicate has a comment documenting that Better Auth signed bouncebacks must contain both fields.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in 0630c035: added a Better Auth 1.6.9 source note for the canonical query order and documented why the extra-key loop remains for future/raw URLSearchParams-shaped callers.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
llun: Addressed in 0630c035: `getConsentRedirectUrl` now prefers Better Auth `data.url`, keeps `redirect_uri` only as a documented legacy fallback, and the tests cover the preference when both fields are present.

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
llun: Addressed in 0630c035: added deny-path coverage for `accept: false`, the ordered signed `oauth_query`, the `data.url` redirect response, and the `access_denied` fallback when no redirect URL is returned.

app/(nosidebar)/oauth/authorize/page.tsx
llun: Addressed in 0630c035: `/oauth/authorize` now resolves the actor/session first, then delegates unsigned requests to Better Auth before loading the OAuth client.

lib/types/database/operations.ts
llun: Addressed in 0630c035: `POST` and `DELETE /api/v1/push/subscribe` now use `OAuthGuard([Scope.enum.push])`, with route tests covering the guard wiring. I also verified locally that unauthenticated push subscribe returns 401 and authenticated subscribe/delete return 200.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
llun: Addressed in 0630c035 and 0e878fcb: `getConsentRedirectUrl` now prefers `data.url`, documents the legacy fallback, returns the parsed absolute URL instead of the raw string, and tests cover both URL preference and protocol-relative normalization.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in 0630c035: the invariant above `shouldDelegateToBetterAuth` now explains that Better Auth signed bouncebacks carry both `sig` and `exp`; delegation also strips both fields to avoid carrying stale partial signatures.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in 0630c035: added the Better Auth 1.6.9 source/order comment and documented that the extra-key loop exists for future/raw URLSearchParams-shaped callers, even though today the page passes parsed `SearchParams`.

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
llun: Addressed in 0630c035: deny-path tests now cover `accept: false`, ordered `oauth_query`, `data.url`-only redirect, and the `access_denied` fallback.

app/(nosidebar)/oauth/authorize/page.tsx
llun: Addressed in 0630c035: the unsigned authorize path delegates after authenticated actor resolution and before OAuth client lookup, removing the discarded client DB round-trip.

llun:
@codex review

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -25,20 +25,50 @@ import { UsableScopes } from '@/lib/types/database/operations'
...
+    const parsed = new URL(redirectUrl, 'https://activities.local')
+    if (parsed.protocol === 'http:' || parsed.protocol === 'https:') {
+      return parsed.toString()
+    }
+  } catch {
+    // Invalid redirect response; fall through to the existing error handling.
+  }
+
+  return undefined
+}
```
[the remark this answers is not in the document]
llun: Addressed in e5e74d69: `getConsentRedirectUrl` now uses the browser `window.location.origin` as the base URL, with the existing `activities.local` fallback for jsdom/local test resolution. Focused OAuth tests and `yarn build` pass.

llun:
@codex review

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+  url.search = buildOAuthQuery(authorizeParams)
+  return url.toString()
+}
+
+// Better Auth consent signatures require sig and exp as an inseparable pair. If
+// either field is absent or blank, treat the request as unsigned and send it
+// through Better Auth so it can validate the client request and sign a fresh
+// consent query.
+export const shouldDelegateToBetterAuth = (params: SearchParams): boolean =>
+  !params.sig || !params.exp
```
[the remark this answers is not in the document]
llun: Addressed in 7554d2e9: `shouldDelegateToBetterAuth` now treats expired or malformed `exp` values as unsigned and strips stale `sig`/`exp` before delegating for a fresh Better Auth signature.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -25,20 +25,57 @@ import { UsableScopes } from '@/lib/types/database/operations'
...
+  // Legacy shape from the original custom consent handler.
+  redirect_uri?: string
+}
+
+const getConsentRedirectBaseUrl = () => {
+  if (typeof window === 'undefined') return 'https://activities.local'
+  return window.location.origin === 'http://localhost'
+    ? 'https://activities.local'
+    : window.location.origin
+}
```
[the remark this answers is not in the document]
llun: Addressed in 7554d2e9: `getConsentRedirectBaseUrl` now uses `window.location.origin` directly in the browser, with only the non-browser fallback left for test/SSR safety.

app/(nosidebar)/oauth/authorize/AuthorizeCard.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+        url: 'javascript:alert(1)'
+      })
+    ).toBeUndefined()
+
+    expect(
+      getConsentRedirectUrl({
+        redirect: true,
+        url: '//phanpy.local/?code=oauth-code'
+      })
+    ).toBe('https://phanpy.local/?code=oauth-code')
```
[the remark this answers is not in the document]
llun: Addressed in 7554d2e9: the protocol-relative redirect expectation now matches jsdom/current-origin behavior (`http://phanpy.local/...`).

app/(nosidebar)/oauth/authorize/authorizeQuery.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  redirect_uri: 'https://phanpy.local/?from=login',
+  response_type: 'code',
+  scope: 'read write follow push',
+  state: 'state with spaces',
+  code_challenge: 'challenge/with+symbols=',
+  code_challenge_method: 'S256'
+}
+
+describe('OAuth authorize query helpers', () => {
+  it('delegates unsigned Mastodon-compatible authorize requests to Better Auth', () => {
```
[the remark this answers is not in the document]
llun: Addressed in 7554d2e9: the authorize-query tests now mock and restore `Date.now()`, and include expired/malformed signature delegation coverage.

llun:
@codex review

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
```diff
@@ -0,0 +1,80 @@
...
+): string => {
+  const authorizeParams = shouldDelegateToBetterAuth(params)
+    ? { ...params, sig: undefined, exp: undefined }
+    : params
+  const url = new URL('/api/auth/oauth2/authorize', baseUrl)
+  url.search = buildOAuthQuery(authorizeParams)
+  return url.toString()
+}
+
+const hasExpiredBetterAuthSignature = (exp: string): boolean => {
```
llun, line 70: **Verify `exp` units.** This treats `exp` as **seconds since epoch** (`Date.now() / 1000 > expiresAt`). Better Auth's `oauth-provider` should be cross-checked — if `exp` is actually emitted in **milliseconds**, every signed bounceback will appear expired and the page will infinite-loop back to Better Auth. The signing comment above mentions `Better Auth 1.6.9`; please pin the line in `serializeAuthorizationQuery` / wherever `exp` is computed and add a one-line assertion in `authorizeQuery.test.ts` against a representative real Better Auth timestamp so a future version bump catches a units change.

Also worth considering: a small clock-skew tolerance (e.g. `Date.now() / 1000 > expiresAt + SKEW_SECONDS` is fine; `< expiresAt - SKEW` for not-yet-valid is the dangerous direction). If Better Auth and Next run on the same process this is a non-issue, but it's cheap insurance against future deployment topologies.

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -25,20 +25,55 @@ import { UsableScopes } from '@/lib/types/database/operations'
...
+
+export const getConsentRedirectUrl = (data: ConsentResponse) => {
+  const redirectUrl = data.url ?? data.redirect_uri
+  if (!redirectUrl) return undefined
+
+  try {
+    const parsed = new URL(redirectUrl, getConsentRedirectBaseUrl())
+    if (parsed.protocol === 'http:' || parsed.protocol === 'https:') {
+      return parsed.toString()
+    }
```
llun, line 55: Minor: the new test confirms that `//phanpy.local/?code=oauth-code` is **accepted** by the validator and normalized to `http://phanpy.local/...` (using jsdom's `http://localhost` origin). So the http(s) protocol gate passes protocol-relative URLs through to `window.location.href`. We trust Better Auth's consent response to be the only source of these URLs, but if you want full belt-and-suspenders, requiring an explicit scheme (e.g. `/^https?:\/\//i.test(redirectUrl)` before parsing) would reject `//host/...` and `/relative/...` outright. Optional — flagging because the test made the behavior explicit.

---

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
```diff
@@ -25,20 +25,55 @@ import { UsableScopes } from '@/lib/types/database/operations'
...
+interface ConsentResponse {
+  redirect?: boolean
+  url?: string
+  // Legacy shape from the original custom consent handler.
+  redirect_uri?: string
+}
+
+const getConsentRedirectBaseUrl = () => {
+  if (typeof window === 'undefined') return 'https://activities.local'
+  return window.location.origin
```
llun: The hardcoded `'https://activities.local'` SSR fallback bakes a project-specific dev hostname into a client helper. It's only used when `window` is undefined, which shouldn't happen for this client component, but using a neutral placeholder (e.g. `'https://invalid.local'`) or pulling from a config util would read better and not mislead someone tracing through the code.

---

app/api/v1/push/subscribe/route.ts
```diff
@@ -18,7 +19,7 @@ const UnsubscribeRequest = z.object({
 
 export const POST = traceApiRoute(
   'pushSubscribe',
-  AuthenticatedGuard(async (req, { currentActor, database }) => {
+  OAuthGuard([Scope.enum.push], async (req, { currentActor, database }) => {
```
llun, line 22: Switching from `AuthenticatedGuard` to `OAuthGuard([Scope.enum.push])` is the correct enforcement, but this is a **breaking change** for any existing client that called `/api/v1/push/subscribe` with a `read`/`write`-only token (e.g. tokens issued before `push` was added to `UsableScopes`). Worth calling out in the PR body / release notes so it doesn't surprise downstream apps after upgrade. Browser sessions hitting these routes from the web UI are unaffected (they go through the `AuthenticatedGuard` fallback in `OAuthGuard` — assuming that's how it's wired; please double-check).

---

llun (commented):
## Third-round review

Nearly all open items from rounds 1 and 2 are now resolved — nice cleanup:

- ✅ `data.url ?? data.redirect_uri` (preference flipped)
- ✅ `getConsentRedirectUrl` returns `parsed.toString()` — protocol-relative URLs are normalized rather than passed through raw
- ✅ Source/version comment on `BetterAuthAuthorizationParamOrder` (Better Auth 1.6.9)
- ✅ Dead "extras" loop documented with intent
- ✅ Deny-path tests added, including the `access_denied` fallback (and the `navigate` prop is a clean testability fix)
- ✅ Unsigned-path short-circuit hoisted above `getClientFromId` in `page.tsx`
- ✅ JSDoc on `shouldDelegateToBetterAuth` explains the invariant
- ✅ Stray `sig`/`exp` stripped in `buildBetterAuthAuthorizeUrl`
- ✅ `OAuthGuard([Scope.enum.push])` on `/api/v1/push/subscribe` POST/DELETE

New items / things worth a second look:

1. **`hasExpiredBetterAuthSignature` assumes seconds** — `Date.now() / 1000 > expiresAt`. If Better Auth emits `exp` in milliseconds, every signed bounceback will look expired and loop. Please pin the line in Better Auth's source where `exp` is computed and add a regression test against a representative value.
2. **Optional: stricter `getConsentRedirectUrl`** — the new test makes it explicit that `//host/...` is accepted and normalized using the current origin's protocol. We trust the consent endpoint, but an explicit `^https?://` prefix check before parsing would tighten the contract.
3. **SSR placeholder host** — `getConsentRedirectBaseUrl()` falls back to the project-specific `https://activities.local`. Use a neutral placeholder or pull from a config util.
4. **Breaking change call-out** — switching push endpoints from `AuthenticatedGuard` to `OAuthGuard([push])` will 403 any pre-existing token without the `push` scope. Worth flagging in the release notes.

Inline comments have details. Nothing here is blocking.

---

llun:
@codex review

lib/services/guards/OAuthGuard.ts
```diff
@@ -238,13 +238,23 @@ const resolveAuthenticatedContext = async <P>({
...
+      if (!actor && !jwtPayload) {
+        const accountId = (storedToken.userId as string | null) || null
+        if (accountId) {
+          // Better Auth opaque tokens store the authenticated account in userId
+          // and leave referenceId blank. Resolve that account to its local
+          // actor so Mastodon API bearer tokens work without a browser session.
+          const actors = await database.getActorsForAccount({ accountId })
+          actor = actors[0] ?? null
+        }
       }
```
llun, line 246: **This is the main concern with the new code.** Falling back to `actors[0]` when `referenceId` is blank silently picks an arbitrary actor whenever the account has more than one local actor. The consent UI in `AuthorizeCard.tsx` explicitly lets the signed-in user choose which actor to authorize (`currentActorId`), so the bound actor identity is part of the authorization decision — papering over a missing `referenceId` with "whichever actor sorts first" breaks that binding.

Two real risks:

1. **Wrong-actor authorization.** An account with actors `alice` and `bob` grants a Mastodon client access to `bob` via the consent screen, but the opaque token lookup later returns `alice` because she's `actors[0]`. The client can then post as the wrong identity.
2. **Order non-determinism.** `getActorsForAccount` ordering is whatever the database returns — likely insertion or PK order today, but not contractually stable. A future migration or storage swap could silently shuffle the bound identity for existing tokens.

The right fix is to make sure Better Auth's consent flow persists the chosen actor id into the token's `referenceId` (or another column the lookup can read), and to treat a missing `referenceId` as a 401 the way the code did before. If there's a reason the consent flow can't set it, please at least:

- Add a test with **two** actors on the same account asserting which one wins, so the behavior is pinned.
- Add a comment explaining why "first actor" is safe (e.g. only used for single-actor accounts in practice).
- Consider gating this fallback behind a single-actor check (`actors.length === 1`) so multi-actor accounts fail closed.

---

lib/services/guards/OAuthGuard.test.ts
```diff
@@ -574,6 +574,40 @@ describe('#OAuthGuard', () => {
...
+      const guard = OAuthGuard([Scope.enum.read], handler)
+      const req = createRequest({
+        Authorization: 'Bearer better-auth-opaque-token'
+      })
+      const response = await guard(req, { params: Promise.resolve({}) })
+
+      expect(response.status).toBe(200)
+      expect(handler).toHaveBeenCalled()
+      expect(capturedActor?.account?.id).toBe(primaryActor.account.id)
+      expect(mockVerifyAccessToken).not.toHaveBeenCalled()
```
llun, line 601: The new test only covers a **single-actor** account, so `actors[0]` is trivially the right answer. Please add a regression case with two actors attached to the same account that:

- Confirms which actor is returned (or asserts the request 401s if you choose to fail closed).
- Exercises the consent flow choosing the non-first actor and verifies the bearer token resolves to that actor — not the first one.

Without this, the fallback behaves correctly by coincidence for the smoke test but the contract is unpinned.

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun, line 70: Still unaddressed from round 3 — **needs verification before merge**. `hasExpiredBetterAuthSignature` assumes `exp` is seconds-since-epoch. If Better Auth 1.6.9 emits `exp` in **milliseconds**, every signed bounceback will appear expired and the page will infinite-loop back to Better Auth. Please pin the units against the cited source file (`@better-auth/oauth-provider/dist/index.mjs`) and add a regression test against a representative value, or this will be a sharp edge on the first version bump.

---

llun (commented):
## Fourth-round review

This round adds an opaque-token fallback in `OAuthGuard` that resolves the local actor via `storedToken.userId` when `referenceId` is blank. One real concern with that change, plus a few carry-overs from round 3:

**Blocking-shaped (please address before merge):**

1. **`actors[0]` is the wrong fallback for multi-actor accounts.** The consent UI lets users pick which actor to authorize; substituting "first actor returned by the DB" silently re-binds the token to a different identity. See the inline comment — either persist the chosen actor into `referenceId`, fail closed when `actors.length > 1`, or at minimum pin the behavior with a multi-actor test.
2. **`exp` units unverified.** Carried from round 3 — please confirm Better Auth emits seconds (not ms) and pin it with a regression test. Misunits → infinite redirect loop.

**Smaller (still unaddressed from round 3, not blocking):**

3. `getConsentRedirectUrl` accepts protocol-relative URLs by design (per the test). A `^https?://` prefix gate would be tighter.
4. `getConsentRedirectBaseUrl` SSR fallback still uses the project-specific `https://activities.local`.
5. The PR body should explicitly call out the `push`-scope behavior change as breaking for pre-existing tokens.

Nice work on the round-3 items that were addressed in earlier passes — the structure of `authorizeQuery.ts`, the `navigate` injection, and the deny-path coverage all hold up well.

---

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in fea79df1: added the Better Auth 1.6.9 `signParams` source note for seconds-based `exp` (`Math.floor(Date.now() / 1e3) + codeExpiresIn`) and added a representative 10-digit Better Auth timestamp assertion in `authorizeQuery.test.ts`. I did not add skew tolerance because the app delegates to the same Better Auth process that validates the signature; delaying refresh past `exp` would make the consent POST more likely to hit Better Auth expiration.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
llun: Addressed in fea79df1: `getConsentRedirectUrl` now requires explicit `http://` or `https://` absolute URLs before parsing, and the test asserts both protocol-relative and relative redirects are rejected.

app/(nosidebar)/oauth/authorize/AuthorizeCard.tsx
llun: Addressed in fea79df1: removed the SSR/base-url fallback entirely by requiring explicit absolute consent redirect URLs and parsing with `new URL(redirectUrl)` only.

app/api/v1/push/subscribe/route.ts
llun: Addressed in the PR body compatibility note and verified locally: bearer tokens without `push` are not authorized for push subscribe, clients need a fresh `push` token, and browser-session calls still work through `OAuthGuard` session fallback when they include same-origin `Origin` proof. I also fixed the Better Auth opaque-token actor resolution in 84ab8fc8 and verified OAuth bearer-token push subscribe/delete return 200.

llun:
@codex review

lib/services/guards/OAuthGuard.ts
llun, line 246: Still the one open concern. The fallback continues to pick `actors[0]` with no multi-actor guard, and the test still covers only single-actor accounts.

Pasting from last round so it isn't lost: if an account has two actors and the user grants a token to actor B via the consent screen, but `referenceId` is blank on the stored token, this code resolves to whichever actor the DB returns first — potentially actor A. That breaks the actor-binding the consent UI advertises.

Minimum acceptable mitigations (pick one):

1. **Persist the chosen actor into the token during consent** and require `referenceId` on the lookup. Drop the fallback.
2. **Fail closed when `actors.length > 1`** (return 401), and add a test pinning that behavior.
3. **Keep the current `actors[0]` behavior** but add (a) a comment explaining why it's safe in your deployment and (b) a regression test with two actors that pins which one wins.

Option 1 is the right shape long-term. Option 2 is the smallest change that's still safe. Option 3 should be a temporary stopgap with a TODO.

---

llun (commented):
## Fifth-round review

Round-3/round-4 items mostly cleaned up:

- ✅ `getConsentRedirectUrl` now requires an explicit `^https?://` prefix and parses without a base — protocol-relative URLs and bare paths are rejected (with regression tests). Nice.
- ✅ `getConsentRedirectBaseUrl` removed entirely, which also resolves the project-specific SSR placeholder concern.
- ✅ `exp` units are now pinned to seconds, with a comment citing `Math.floor(Date.now() / 1e3) + codeExpiresIn` in Better Auth's `signParams` and a representative-value regression test.
- ✅ PR body has an explicit "Compatibility Note" calling out that pre-existing tokens without `push` will 403 on `/api/v1/push/subscribe` — exactly what was asked for.

**Still open:**

- **Multi-actor opaque-token fallback in `OAuthGuard.ts`**. The `actors[0]` selection when `referenceId` is blank still lacks a multi-actor guard, and the new test only covers single-actor accounts. Suggested mitigations restated inline.

Nothing else from prior rounds is outstanding. If the multi-actor case is genuinely impossible in this codebase (e.g., one-actor-per-account is enforced upstream), a comment saying so + a single-line `actors.length === 1` assertion would close the concern.

---

lib/services/guards/OAuthGuard.ts
llun: Addressed in 8ab964d2 by removing the unsafe `actors[0]` fallback entirely. `OAuthGuard` now fails closed when an opaque token has no actor `referenceId`. The consent UI now persists the selected actor through `/api/v1/actors/switch` before approving, so Better Auth `consentReferenceId` stores the chosen actor id in the token `referenceId`. Local API smoke verified a token issued after selecting `testactor2` stores `referenceId=https://activities.local/users/testactor2` and push subscribe/delete work with that bearer token.

lib/services/guards/OAuthGuard.test.ts
llun: Addressed in 8ab964d2: replaced the single-actor fallback test with a fail-closed test for Better Auth opaque tokens that have `userId` but no actor `referenceId`, and added an `AuthorizeCard` test that pins persisting a non-first selected actor before consent approval.

app/(nosidebar)/oauth/authorize/authorizeQuery.ts
llun: Addressed in fea79df1: `authorizeQuery.ts` now cites Better Auth 1.6.9 `signParams`, where `exp` is computed as `Math.floor(Date.now() / 1e3) + codeExpiresIn`, and `authorizeQuery.test.ts` includes a representative 10-digit Better Auth `exp` value assertion.

lib/services/guards/OAuthGuard.ts
```diff
@@ -238,13 +238,23 @@ const resolveAuthenticatedContext = async <P>({
...
```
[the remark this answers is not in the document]
llun: Addressed in 8ab964d2: removed the account `userId` -> `actors[0]` fallback rather than sorting it. Missing actor `referenceId` now denies access, and the consent approval path persists the selected actor before calling Better Auth so opaque tokens carry the selected actor id. Local smoke verified the stored token reference id matches the selected actor.

llun:
@codex review

lib/services/guards/OAuthGuard.ts
llun: Addressed in 8ab964d2 with option 1: the selected actor is persisted before consent approval so Better Auth stores the actor id in token `referenceId`, and `OAuthGuard` no longer falls back to `actors[0]` when `referenceId` is missing. The guard now fails closed for opaque tokens with only account `userId`, and local smoke verified a token issued for `testactor2` stores `referenceId=https://activities.local/users/testactor2`.

PR #773, opened 2026-05-27
refactor: consolidate HTTP header utilities into lib/utils/http-headers/

llun:
## Summary

- Moves `getCORSHeaders.ts`, `securityHeaders.ts`, and `staticSecurityHeaders.ts` into a new `lib/utils/http-headers/` folder
- Adds a barrel `index.ts` that documents the three non-overlapping header layers and the disjoint-key invariant
- `next.config.ts`, `proxy.ts`, and `jest.setup-after-env.ts` import from specific sub-paths to avoid pulling node/config deps into the Next.js config evaluation context and the middleware Edge Runtime

### New layout

```
lib/utils/http-headers/
  index.ts    — barrel (JSDoc explains the 3 layers)
  cors.ts     — Access-Control-* per-request headers
  csp.ts      — Content-Security-Policy builder
  static.ts   — 4 immutable headers (X-CTO, X-Frame, Referrer, Permissions)
```

### Layer interaction (no overrides)

| Layer | Where | Headers |
|---|---|---|
| Next.js `headers()` | `next.config.ts` | X-Content-Type-Options, X-Frame-Options, Referrer-Policy, Permissions-Policy |
| Middleware | `proxy.ts` | Content-Security-Policy |
| Route handler | `apiResponse()` | Access-Control-* |

All header keys are disjoint across layers — documented in `index.ts` to prevent future collisions.

## Test plan

- [ ] `yarn lint` — no new errors
- [ ] `yarn build` — passes (including middleware Edge Runtime check)
- [ ] `yarn test` — 334 suites, 2873 tests all pass
- [ ] `proxy.test.ts` — middleware static-import graph check still passes
- [ ] `next.config.test.ts` — static header wiring still passes


llun:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

llun:
@codex review

lib/utils/http-headers/cors.test.ts
```diff
@@ -24,11 +18,11 @@ describe('#getCORSHeaders', () => {
   })
 
   it('uses host-based origin when Origin header is not present', () => {
-    const headers = new Headers([['Host', 'api.example.com']])
+    const headers = new Headers([['Host', 'test.llun.dev']])
     const result = getCORSHeaders(['GET'], headers)
```
llun, line 24: The test for the host-fallback case (no `Origin` header) previously had an explicit `jest.mock('@/lib/config')` that pinned `host` to `'api.example.com'`. The new test drops that mock and silently relies on the global mock in `jest.setup.mjs` that returns `host: 'test.llun.dev'` — which is why the assertion was updated from `'https://api.example.com'` to `'https://test.llun.dev'`.

This works correctly today because `selectHeaderHost` trusts the Host header value when it matches the configured host, so `'test.llun.dev'` is returned. However, the test no longer documents where the host value comes from. Any change to `TEST_DOMAIN` or the global mock shape would silently break this test without a clear failure message pointing here.

Consider adding a brief comment noting the reliance on the global config mock, or re-introducing a scoped `jest.mock` for just this file to keep the fallback behavior self-contained and explicit.

proxy.ts
```diff
@@ -3,7 +3,7 @@ import { NextRequest, NextResponse } from 'next/server'
 import { getProxyHostConfig } from '@/lib/config/host'
 import { acceptContainsContentTypes } from '@/lib/utils/acceptContainsContentTypes'
 import { selectHeaderHost } from '@/lib/utils/host'
-import { getContentSecurityPolicyHeader } from '@/lib/utils/securityHeaders'
+import { getContentSecurityPolicyHeader } from '@/lib/utils/http-headers/csp'
```
llun, line 8: There are mixed sub-path vs. barrel import styles across the four key files:

- `proxy.ts` imports from `@/lib/utils/http-headers/csp` (sub-path)
- `proxy.test.ts` imports from `@/lib/utils/http-headers` (barrel)
- `jest.setup-after-env.ts` imports from `@/lib/utils/http-headers/csp` (sub-path)
- `next.config.test.ts` imports from `@/lib/utils/http-headers` (barrel)

`next.config.ts` must use the sub-path `./static` import because the barrel re-exports `./csp`, which calls `getSecurityHeaderConfig()` and reads `ACTIVITIES_*` env vars — that would violate the project rule against reading deployment config at build time. That constraint applies only to `next.config.ts`.

For `proxy.ts` and `jest.setup-after-env.ts` there is no such constraint, so they could consistently use the barrel. It would help to add a comment in `next.config.ts` (and possibly `proxy.ts`) explaining why `next.config.ts` is intentionally importing from the sub-path rather than the barrel, so the inconsistency looks deliberate rather than accidental.

lib/utils/http-headers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+// Static security headers (build-time, next.config layer)
+export { type SecurityHeader, getStaticSecurityHeaders } from './static'
+
+// Content Security Policy (per-request, middleware layer)
+export {
+  getContentSecurityPolicy,
+  getContentSecurityPolicyHeader,
+  getSecurityHeaders,
+  resetContentSecurityPolicyCacheForTests
+} from './csp'
```
llun, line 38: `resetContentSecurityPolicyCacheForTests` is exported from the barrel `index.ts`. The function is intentionally test-only — it throws if called outside a Jest worker. Exporting it from the production barrel is not broken (tree-shaking/lazy resolution means it won't execute at module load time), but it leaks a test-only symbol into the public surface of the module.

Prefer keeping test-only helpers importable only via the sub-path (`@/lib/utils/http-headers/csp`), not the barrel. `jest.setup-after-env.ts` already correctly imports from the sub-path; the barrel export is inconsistent with that and could confuse future readers about what is safe to call in production code.

llun (commented):
Codex review: Clean consolidation — all old exports are correctly re-exported from the barrel, no dead imports remain, and the barrel doc comment usefully explains the three non-overlapping header layers. Three findings, all nits/suggestions rather than blockers:

**1. cors.test.ts — host-fallback test now implicitly depends on the global config mock** (inline comment on line 22). The old test had an explicit `jest.mock('@/lib/config')` anchoring the host to `'api.example.com'`. The new test relies on the global mock in `jest.setup.mjs` (which returns `host: 'test.llun.dev'`), and silently updated the assertion to match. The test is correct as-is, but the dependency is now invisible: a change to `TEST_DOMAIN` or the global mock would break this test with no obvious pointer back to this file. A scoped mock or a brief comment would make the intent explicit.

**2. proxy.ts — sub-path import where the barrel is available** (inline comment on proxy.ts line 6). `proxy.ts` and `jest.setup-after-env.ts` import from the `/csp` sub-path while their respective test files import from the barrel. The only file that *must* use the sub-path is `next.config.ts` (the barrel re-exports `./csp`, which reads `ACTIVITIES_*` env vars at call time — build-time env isolation). For all other callers, consistently using the barrel would be cleaner; or a comment on `next.config.ts` explaining the intentional sub-path import would prevent the inconsistency from looking accidental.

**3. index.ts — `resetContentSecurityPolicyCacheForTests` re-exported from the production barrel** (inline comment on index.ts line 35). The function is test-only and throws outside a Jest worker. Exporting it from the barrel leaks a test-only symbol into the module's public surface. `jest.setup-after-env.ts` correctly imports it from the sub-path, making the barrel export redundant. Consider removing it from the barrel to prevent callers from inadvertently depending on it via the main import path.

proxy.ts
llun, line 8: The `index.ts` barrel is introduced as the single import surface for all three header layers, and most consumers correctly import from it. However, `proxy.ts` imports directly from the submodule:

```ts
// proxy.ts — bypasses the barrel
import { getContentSecurityPolicyHeader } from '@/lib/utils/http-headers/csp'
```

…while its companion test file imports via the barrel:

```ts
// proxy.test.ts — uses the barrel
import { resetContentSecurityPolicyCacheForTests } from '@/lib/utils/http-headers'
```

Jest resolves both paths to the same module instance, so there's no cache isolation bug here. But the inconsistency is worth noting: the barrel's own JSDoc comment describes it as the intended import surface and says "Do not add a header to more than one layer", implying consumers should import from `@/lib/utils/http-headers`, not from submodule paths. Consider switching `proxy.ts` to import from the barrel for consistency (or add a comment explaining why the direct import is intentional, e.g. to avoid pulling in the full module graph in middleware).

llun (commented):
Review: consolidate HTTP header utilities into lib/utils/http-headers/

The refactoring is clean and complete. All 100+ import sites were updated, all three old source files (getCORSHeaders.ts, securityHeaders.ts, staticSecurityHeaders.ts) were properly removed, and no stale references remain.

One inconsistency to flag (inline comment posted on proxy.ts line 6):

`proxy.ts` imports `getContentSecurityPolicyHeader` directly from the submodule path `@/lib/utils/http-headers/csp`, bypassing the barrel `index.ts` that was specifically created as the unified import surface. Its companion `proxy.test.ts` correctly uses the barrel. The same pattern appears in `jest.setup-after-env.ts` (imports from `/csp` directly) and `next.config.ts` (imports from `/static` directly). This is not a bug — Jest resolves both to the same module instance — but it contradicts the intent stated in the barrel's JSDoc. Either switch these three callers to the barrel, or add an explicit comment on the direct imports explaining the reason (e.g. avoiding transitive module loading in Next.js middleware/build-time entry points).

Everything else looks correct:
- The CORS test correctly relies on the global `@/lib/config` mock in `jest.setup.mjs` (which returns `host: 'test.llun.dev'`) after removing the now-redundant local mock. The updated test expectation (`'https://test.llun.dev'`) is consistent with that mock.
- The `OAuthGuard.ts` parentheses simplification is semantically identical to the original.
- The three-layer header architecture described in `index.ts` is accurately documented and the layers remain non-overlapping.

lib/utils/http-headers/cors.test.ts
llun: Fixed in 46a83d5a. Added a comment to the test explaining that the Host header matches TEST_DOMAIN from the global @/lib/config mock in jest.setup.mjs, making the dependency explicit.

proxy.ts
llun: Fixed in 46a83d5a. Added a comment to proxy.ts, next.config.ts, and jest.setup-after-env.ts explaining why each must use the sub-path import rather than the barrel — Edge Runtime constraint (cors.ts pulls in @/lib/config which has fs/path deps), build-time isolation, and Jest mock isolation respectively.

lib/utils/http-headers/index.ts
llun: Fixed in 46a83d5a. Added a comment in index.ts noting that resetContentSecurityPolicyCacheForTests is test-only, and documenting why proxy.ts/next.config.ts must use sub-path imports. The export stays because proxy.test.ts and next.config.test.ts both consume it from the barrel.

proxy.ts
llun: Fixed in 46a83d5a. The sub-path import is intentional and necessary — the barrel transitively loads cors.ts which imports @/lib/config (fs/path dependencies), which would pull Node.js modules into the middleware Edge Runtime bundle and break the build. Added an explanatory comment to make this clear.

llun:
All review comments addressed in 46a83d5a:

1. **cors.test.ts** — Added comment documenting that the host-fallback test relies on TEST_DOMAIN from the global `jest.setup.mjs` mock.

2. **proxy.ts** — Added comment explaining the sub-path import is required: the barrel transitively loads `cors.ts` → `@/lib/config` (fs/path deps), which breaks the middleware Edge Runtime bundle.

3. **next.config.ts** — Added comment explaining the sub-path import is required: the barrel loads `csp.ts` which reads deployment env vars at call time, violating build-time isolation.

4. **jest.setup-after-env.ts** — Added comment explaining the sub-path import is required: the barrel loads `cors.ts` → `@/lib/config`, interfering with per-test module mock isolation.

5. **index.ts** — Added comments noting that `resetContentSecurityPolicyCacheForTests` is test-only, and documenting why `proxy.ts`/`next.config.ts` must use sub-path imports while test files may use the barrel freely.

llun:
@codex review

lib/utils/http-headers/index.ts
```diff
@@ -0,0 +1,41 @@
...
+// @/lib/config (fs/path deps), breaking the Edge Runtime and build-time
+// isolation constraints. Test files may use this barrel freely.
+export {
+  getContentSecurityPolicy,
+  getContentSecurityPolicyHeader,
+  getSecurityHeaders,
+  // test-only: throws outside a Jest worker; imported by proxy.test.ts and
+  // next.config.test.ts via this barrel
+  resetContentSecurityPolicyCacheForTests
+} from './csp'
```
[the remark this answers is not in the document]
llun: Fixed in 6b8a69ef. Removed `resetContentSecurityPolicyCacheForTests` from the barrel `index.ts`. The function is now only accessible via the sub-path `@/lib/utils/http-headers/csp`, keeping the production barrel clean.

proxy.test.ts
```diff
@@ -4,7 +4,7 @@ import os from 'os'
 import path from 'path'
 
 import { resetHostConfigCacheForTests } from '@/lib/config/host'
-import { resetContentSecurityPolicyCacheForTests } from '@/lib/utils/securityHeaders'
+import { resetContentSecurityPolicyCacheForTests } from '@/lib/utils/http-headers'
```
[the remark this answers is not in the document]
llun: Fixed in 6b8a69ef. Updated `proxy.test.ts` to import `resetContentSecurityPolicyCacheForTests` directly from `@/lib/utils/http-headers/csp` instead of the barrel. Same fix applied to `next.config.test.ts`.

llun:
@codex review

lib/utils/http-headers/index.ts
```diff
@@ -0,0 +1,38 @@
...
+export { HttpMethod, getCORSHeaders } from './cors'
+
+// Static security headers (static values, next.config layer)
+export { type SecurityHeader, getStaticSecurityHeaders } from './static'
+
+// Content Security Policy (per-request, middleware layer)
+// Note: proxy.ts and next.config.ts must import from ./csp and ./static
+// directly — the barrel pulls in cors.ts which transitively loads
+// @/lib/config (fs/path deps), breaking the Edge Runtime and build-time
+// isolation constraints. Test files may use this barrel freely.
```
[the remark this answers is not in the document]
llun: This is a sound architectural direction, but it falls outside the scope of this PR. This refactor's goal is file structure only — moving three utilities into `lib/utils/http-headers/` without any behavioral or API changes.

Changing `getCORSHeaders` to accept a pre-resolved `origin: string` instead of raw `headers` would require updating all ~130 call sites across route handlers, guards, and tests. That's a separate follow-up with its own risk surface and review.

The current sub-path import approach (documented with comments) is an intentional workaround that keeps this PR narrowly scoped. A follow-up PR to purify `getCORSHeaders` and eliminate the `headerHost` transitive dependency would be welcome.

llun:
@codex review

PR #779, opened 2026-05-27
fix: preserve Mastodon OAuth flow through consent and token exchange

llun:
## Summary
- preserve the full signed Mastodon OAuth consent query when approving or denying consent
- allow safe custom redirect URI schemes such as `mastodon://...` in the consent handoff
- pin `better-auth` to `1.6.11` so the runtime matches the OAuth token exchange contract used by the app

## Validation
- verified the consent flow now reaches the native Mastodon callback instead of failing with a 400
- verified the token exchange no longer crashes with a 500 and now returns the expected protocol-level `invalid_grant` response for a dummy code

## Testing
- targeted OAuth flow verification in the browser and via local token endpoint probing
- full `yarn lint`, `yarn build`, and `yarn test` were not run in this PR flow


llun:
Refreshing PR metadata after branch update; reopening immediately.

PR #890, opened 2026-06-06
minor: public chrome for logged-out status page

llun:
## Summary

Adds **instance custom emoji** (Mastodon `CustomEmoji`) end to end: storage, public + admin endpoints, ActivityPub federation, a postbox **sticker/emoji picker**, and an **admin management page**. Also makes `GET /api/v2/instance` public to match Mastodon.

A posted status containing `:shortcode:` now (1) renders the emoji locally, (2) federates an ActivityPub `Emoji` tag so Mastodon and other AP servers render it, and (3) is surfaced unauthenticated via `GET /api/v1/custom_emojis`.

## Endpoints — Mastodon compatibility checklist

| Endpoint | Change | Mastodon source |
|---|---|---|
| `GET /api/v1/custom_emojis` | **Now public** (removed `OAuthGuard`); returns the *listed* set: `disabled = false AND visible_in_picker = true` | `custom_emojis_controller.rb` index uses `CustomEmoji.listed`; endpoint is public ([docs](https://docs.joinmastodon.org/methods/custom_emojis/)) |
| `GET /api/v2/instance` | **Now public** (removed `OAuthGuard`) | `/methods/instance/` — served unauthenticated |
| `GET /api/v1/instance` | Unchanged (already public, shape verified) | `/methods/instance/` |
| `POST /api/v1/admin/custom_emojis` | Create via multipart upload (shared media pipeline); 422 on bad mime/dup/shortcode | mirrors Mastodon admin custom emoji create |
| `GET /api/v1/admin/custom_emojis` | List incl. disabled | admin index |
| `PATCH`/`PUT /api/v1/admin/custom_emojis/:id` | Update enable/disable, `visible_in_picker`, `category` | Mastodon admin uses `PATCH`; `PUT` added to match this repo's existing admin `[id]` convention |
| `DELETE /api/v1/admin/custom_emojis/:id` | Delete | admin destroy |

`CustomEmoji` entity fields (`shortcode`, `url`, `static_url`, `visible_in_picker`, `category`) match `custom_emoji_serializer.rb` / [entities/CustomEmoji](https://docs.joinmastodon.org/entities/CustomEmoji/). The admin shape adds `id` + `disabled`.

## ActivityPub Emoji-tag round-trip (verified on a running server)

- **Outbound:** posting `hello :blobcat:` serializes a `tag` entry `{ "type":"Emoji", "name":":blobcat:", "icon":{"type":"Image","url":…}, "id":…, "updated":… }` in the status' AP JSON (confirmed over HTTP: `GET /users/testuser/statuses/:id` with `Accept: application/activity+json`).
- **Inbound:** incoming Notes with `Emoji` tags already parse into stored `emoji` domain Tags (in `createNoteJob`); added a regression test.
- **Bug fixed:** `getMentionFromTag` previously fell through to `Mention.safeParse` for emoji tags, so the two `status.ts` `toActivityPubObject` sites emitted **bogus `Mention` objects** for emoji. It now returns `null` for emoji; new `getEmojiFromTag` emits the `Emoji` object. All three outbound builders (`getNoteFromStatus`, Note + Question in `status.ts`) include emoji uniformly.
- Mastodon REST `status.emojis` is populated automatically by the pre-existing `getEmojisFromTags`.

## How shortcodes become tags

`createNote` / `createPoll` / `updateNote` scan the text for `:shortcode:`, resolve against the **enabled** instance emoji set (ignoring `visible_in_picker`, so hand-typed non-picker emoji still federate, matching Mastodon), and persist `emoji` domain Tags. Edits re-sync (delete + recreate).

## Frontend

- **Postbox picker** (`emoji-picker-button.tsx`): one popover with **system Unicode emoji** (curated local dataset, no new deps) + **instance custom emoji** (via new `getCustomEmojis` in `lib/client.ts`), with search, category tabs, grid, and hover preview per `preview/stickers.html`. System emoji insert the character; custom emoji insert `:shortcode:` at the caret. Keyboard-accessible (labelled controls, Escape to close).
- Custom emoji render as images in the picker and in the **composer live preview** through the existing `convertEmojisToImages`/`processStatusText` pipeline (refactored to a shared `processStatusTextContent`) — no second renderer, no `dangerouslySetInnerHTML`.
- **Admin page** (`/admin/admin/emojis` under the admin section) using the design-system chrome (`SectionNavDropdown`) and the settings/stickers visual design; all network calls go through `lib/client.ts`.

## Design

Implemented from the handoff bundle (`ui_kits/web/StickerPicker.jsx`, `Stickers.jsx`, `Composer.jsx`, `SettingsPage.jsx` → `StickerSettings`): picker grid/categories/search/preview, the dashed-dropzone upload form, and list rows. Adapted to this project's Tailwind tokens and primitives; admin placement + `SectionNavDropdown` follow `CLAUDE.md`/`AGENTS.md` over the kit's settings location.

## Local verification gates

`prettier --write .` ✓ · `yarn lint` ✓ (0 errors) · `yarn build` ✓ · `yarn test` ✓ (**3545 passed / 418 suites**)

### Browser verification (local SQLite)
- Public `GET /api/v1/custom_emojis` and `GET /api/v2/instance` return 200 with **no token**.
- Picker opens, shows system categories + Custom tab; inserting a custom emoji puts `:blobcat:` in the textarea; the composer preview renders it as an inline image.
- A posted `:blobcat:` status' AP JSON contains the `Emoji` tag (checked over HTTP).
- Admin page lists/creates/updates/deletes emoji.

(Screenshots shared with the author; per `AGENTS.md` they are not committed to the repo.)

## Tests added
DB CRUD, public route (public + shape + picker filter), admin CRUD (incl. video-upload rejection), `getEmojiTags`, `getEmojiFromTag`/`getMentionFromTag`, create-note + poll shortcode→emoji-Tag resolution, outgoing Note **and** Question Emoji tag, inbound emoji-tag parsing, edit re-sync, and picker insertion (both kinds) + preview image rendering.

## Notes for reviewers
- Custom emoji uploads use the shared (actor-scoped) media pipeline via the uploading admin's session actor; an OAuth-token-only admin (no session actor) cannot upload — documented limitation.
- Animated emoji are not transcoded: `static_url = url` (uploads are PNG/JPEG only; the route rejects video/audio even though `FileSchema` allows them).
- `persistEmojiTagsForStatus` reads the enabled emoji set once per note/poll create/edit — acceptable at instance scale.

---

## Security hardening (added during review)

Review (Gemini + review sub-agents) surfaced a class of **pre-existing** visibility leaks in the status-detail render path that this PR's logged-out view makes prominent. Since they live in exactly the logged-out path this PR owns, the logged-out cases are fixed here (logged-in non-recipient filtering is a broader pre-existing concern, deferred to a follow-up PR):

- **Focused status:** a public `Announce` wrapping a followers-only/direct Note (e.g. a self-boost of a private post) passed the top-level `to`/`cc` check and rendered the private original. Now guarded with the recursive `isStatusPubliclyReadable` (returns `notFound()` for logged-out).
- **Ancestor thread (`previouses`):** built via an unfiltered `getStatus`; a public reply to a private parent leaked that parent. Now the chain stops at the first non-public ancestor for logged-out visitors.
- **Replies (content + count):** fetched/rendered unfiltered. Now fetched with `publicOnly: true` plus an object-level `isStatusPubliclyReadable` backstop (also covering the embedded-replies path), and only visible replies are rendered and counted in the stat strip.

All three reuse the canonical `isStatusPubliclyReadable` from `lib/services/statusAccess.ts`. Regression tests live in `app/(timeline)/[actor]/[status]/page.visibility.test.tsx` (4 cases). Other review fixes: a visually-hidden `<h1>` to preserve the document outline for logged-out visitors, the footer moved to a sibling of `<main>` for the `contentinfo` landmark, and `min-h-dvh` across the layout.

### DIFF
--- a/app/(timeline)/PublicFooter.tsx
+++ b/app/(timeline)/PublicFooter.tsx
@@ -1,0 +1,38 @@
+import Image from 'next/image'
+import { FC } from 'react'
+
+// Public footer shown to logged-out visitors, matching the web-public design:
+// a single bordered card with the brand line and a link to the project source.
+// About/Privacy are omitted since the app has no such pages.
+export const PublicFooter: FC = () => (
+  <footer className="mx-auto w-full max-w-[680px] px-4 py-8">
+    <div className="rounded-xl border bg-background/70 px-5 py-4">
+      <div className="flex flex-wrap items-center justify-between gap-3 text-sm">
+        <div className="flex items-center gap-2 text-muted-foreground">
+          <Image
+            src="/logo-nav.png"
+            alt=""
+            aria-hidden="true"
+            width={20}
+            height={20}
+            className="h-5 w-5 shrink-0 object-contain"
+          />
+          <span>
+            <strong className="font-semibold text-foreground">
+              Activities
+            </strong>{' '}
+            — a self-hosted social + fitness server on the Fediverse.
+          </span>
+        </div>
+        <a
+          href="https://github.com/llun/activities.next"
+          target="_blank"
+          rel="noopener noreferrer"
+          className="text-muted-foreground hover:text-foreground"
+        >
+          Source
+        </a>
+      </div>
+    </div>
+  </footer>
+)
--- a/app/(timeline)/PublicTopBar.tsx
+++ b/app/(timeline)/PublicTopBar.tsx
@@ -1,0 +1,24 @@
+import Link from 'next/link'
+import { FC } from 'react'
+
+import { Logo } from '@/lib/components/layout/logo'
+import { Button } from '@/lib/components/ui/button'
+
+// Public top bar shown to logged-out visitors in place of the app nav sidebar:
+// a slim sticky bar with the brand logo and Sign in / Create account links to
+// the real auth routes, matching the web-public design.
+export const PublicTopBar: FC = () => (
+  <header className="sticky top-0 z-30 border-b bg-background/90 backdrop-blur">
+    <div className="mx-auto flex h-16 w-full max-w-[680px] items-center gap-3 px-4">
+      <Logo size="md" />
+      <div className="ml-auto flex items-center gap-2">
+        <Button asChild variant="outline" size="sm">
+          <Link href="/auth/signin">Sign in</Link>
+        </Button>
+        <Button asChild size="sm">
+          <Link href="/auth/signup">Create account</Link>
+        </Button>
+      </div>
+    </div>
+  </header>
+)
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -31,6 +31,7 @@ import { Header } from './Header'
 import { RemoteStatusLoading } from './RemoteStatusLoading'
 import { SignInCallout } from './SignInCallout'
 import { StatusBox } from './StatusBox'
+import { StatusStatStrip } from './StatusStatStrip'
 import { decodePathParam, resolveStatusFromPath } from './resolveStatusFromPath'
 
 interface Props {
@@ -95,6 +96,15 @@ const Page: FC<Props> = async ({ params }) => {
   }
 
   if (!status) {
+    return notFound()
+  }
+
+  // Logged-out visitors may only view statuses that are publicly readable all
+  // the way down. The per-field `isPublicOrUnlisted` check below inspects only
+  // the top-level recipients — for an Announce those are the boost's, not the
+  // boosted note's — so a public boost of a followers-only post would otherwise
+  // slip through. `isStatusPubliclyReadable` recurses into the original status.
+  if (!currentActor && !isStatusPubliclyReadable(status)) {
     return notFound()
   }
 
@@ -232,7 +242,13 @@ const Page: FC<Props> = async ({ params }) => {
   if (isFitnessDashboard) {
     return (
       <div className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
-        <Header isFitnessDashboard />
+        {currentActorProfile ? (
+          <Header isFitnessDashboard />
+        ) : (
+          // Logged-out view has no back-button chrome (matching the web-public
+          // design), but keep a top-level heading for the document outline.
+          <h1 className="sr-only">Activity</h1>
+        )}
 
         <div className="border-b bg-background">
           <StatusBox
@@ -243,6 +259,15 @@ const Page: FC<Props> = async ({ params }) => {
             status={cleanJson(status)}
             variant="detail"
           />
+          {!currentActorProfile ? (
+            <div className="px-4 pb-4">
+              <StatusStatStrip
+                boosts={statusForLayout.totalShares}
+                likes={statusForLayout.totalLikes}
+                replies={replies.length}
+              />
+            </div>
+          ) : null}
         </div>
 
         {!currentActorProfile ? <SignInCallout /> : null}
@@ -252,7 +277,13 @@ const Page: FC<Props> = async ({ params }) => {
 
   return (
     <div className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
-      <Header isFitnessDashboard={false} />
+      {currentActorProfile ? (
+        <Header isFitnessDashboard={false} />
+      ) : (
+        // Logged-out view has no back-button chrome (matching the web-public
+        // design), but keep a top-level heading for the document outline.
+        <h1 className="sr-only">Post</h1>
+      )}
 
       {previouses.reverse().map((item) => (
         <div
@@ -279,6 +310,15 @@ const Page: FC<Props> = async ({ params }) => {
           variant="detail"
           isMediaUploadEnabled={Boolean(mediaStorage)}
         />
+        {!currentActorProfile ? (
+          <div className="px-4 pb-4">
+            <StatusStatStrip
+              boosts={statusForLayout.totalShares}
+              likes={statusForLayout.totalLikes}
+              replies={replies.length}
+            />
+          </div>
+        ) : null}
       </div>
 
       {!currentActorProfile ? <SignInCallout /> : null}
--- a/app/(timeline)/[actor]/[status]/page.visibility.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.visibility.test.tsx
@@ -277,6 +277,69 @@ describe('Page visibility for logged-out visitors', () => {
     // The reply-count heading reflects only the visible (public) reply.
     expect(screen.getByText('Replies (1)')).toBeInTheDocument()
   })
+
+  it('returns notFound for a public boost wrapping a private original', async () => {
+    const privateOriginal = buildNote({
+      id: 'private-original',
+      to: ['https://activities.local/users/anna/followers'],
+      cc: []
+    })
+    const announce = {
+      id: 'public-announce',
+      type: 'Announce',
+      actorId: 'https://activities.local/users/anna',
+      actor: null,
+      to: [ACTIVITY_STREAM_PUBLIC],
+      cc: [],
+      edits: [],
+      isLocalActor: true,
+      createdAt: 1,
+      updatedAt: 1,
+      originalStatus: privateOriginal
+    } as unknown as Status
+
+    mockResolveStatusFromPath.mockResolvedValue({
+      status: announce,
+      statusId: 'public-announce',
+      fullStatusId:
+        'https://activities.local/users/anna/statuses/private-original',
+      isStatusHash: true
+    })
+
+    await expect(renderPage()).rejects.toThrow('NEXT_NOT_FOUND')
+  })
+
+  it('renders the logged-out fitness dashboard with an sr-only heading and stat strip', async () => {
+    const focused = buildNote({
+      id: 'fitness-status',
+      fitness: {
+        id: 'fit-1',
+        fileName: 'run.fit',
+        fileType: 'fit',
+        mimeType: 'application/octet-stream',
+        bytes: 1024,
+        url: 'https://activities.local/fit/run.fit',
+        processingStatus: 'completed'
+      }
+    } as unknown as Partial<Status>)
+
+    mockResolveStatusFromPath.mockResolvedValue({
+      status: focused,
+      statusId: 'fitness-status',
+      fullStatusId: focused.url,
+      isStatusHash: true
+    })
+
+    await renderPage()
+
+    // The back-button Header is gated to signed-in users; logged-out keeps only
+    // a visually-hidden top-level heading for the document outline.
+    expect(
+      screen.getByRole('heading', { level: 1, name: 'Activity' })
+    ).toBeInTheDocument()
+    expect(screen.getByTestId('status-fitness-status')).toBeInTheDocument()
+    expect(screen.getByText('Boosts')).toBeInTheDocument()
+  })
 })
 
 describe('Page visibility for logged-in non-recipient viewers', () => {
--- a/app/(timeline)/layout.tsx
+++ b/app/(timeline)/layout.tsx
@@ -9,6 +9,9 @@ import { getActorProfile, getMention } from '@/lib/types/domain/actor'
 import { cn } from '@/lib/utils'
 import { getActorFromSession } from '@/lib/utils/getActorFromSession'
 
+import { PublicFooter } from './PublicFooter'
+import { PublicTopBar } from './PublicTopBar'
+
 interface LayoutProps {
   children: ReactNode
 }
@@ -21,6 +24,30 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
 
   const session = await getServerAuthSession()
   const actor = await getActorFromSession(database, session)
+
+  // Logged-out visitors get the public chrome — a slim top bar with sign-in
+  // CTAs and a footer in place of the nav sidebar — matching the web-public
+  // design. The reading column is narrower than the signed-in app width.
+  if (!actor) {
+    return (
+      // min-h-dvh (dynamic viewport height) rather than min-h-screen/100vh so
+      // the footer stays at the bottom without a mobile address-bar gap — same
+      // rationale as the public error pages (lib/components/error-page.tsx).
+      <div className="flex min-h-dvh flex-col">
+        <PublicTopBar />
+        <main className="flex flex-1 flex-col overflow-x-clip">
+          <div className="mx-auto flex w-full max-w-[680px] flex-1 flex-col px-4 py-6">
+            {children}
+          </div>
+        </main>
+        {/* Footer is a sibling of <main> (not nested in it) so its <footer>
+            maps to the contentinfo landmark; main keeps flex-1 to pin it to the
+            bottom on short pages. */}
+        <PublicFooter />
+        <Modal />
+      </div>
+    )
+  }
 
   // Check if iconUrl is a real user-uploaded avatar (not auto-generated)
   // Auto-generated URLs typically contain service identifiers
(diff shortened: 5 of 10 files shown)


app/(timeline)/layout.tsx
```diff
@@ -22,6 +25,24 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
...
+        <main className="flex min-h-screen flex-col overflow-x-clip">
+          <div className="mx-auto flex w-full max-w-[680px] flex-1 flex-col px-4 py-6">
+            {children}
+          </div>
+          <PublicFooter />
+        </main>
+        <Modal />
+      </div>
+    )
+  }
```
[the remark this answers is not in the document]
llun: Good call — adopted `min-h-dvh` for the public layout in de55470d6. This matches the existing precedent in `lib/components/error-page.tsx`, which already uses `min-h-dvh` for the same mobile address-bar reason. I also restructured the branch to a single `min-h-dvh` flex column (top bar + flex-1 main + footer) so the sticky top bar no longer forces extra scroll.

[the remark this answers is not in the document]
llun: Good call — adopted `min-h-dvh` for the public layout in de55470d6. This matches the existing precedent in `lib/components/error-page.tsx`, which already uses `min-h-dvh` for the same mobile address-bar reason. I also restructured the branch to a single `min-h-dvh` flex column (top bar + flex-1 main + footer) so the sticky top bar no longer forces extra scroll.

llun:
### Sub-agent code review (round 1)

I ran an independent code-review agent over the diff. It surfaced one real issue, now fixed in de55470d6:

- **Logged-out status page lost its only `<h1>`** (`app/(timeline)/[actor]/[status]/page.tsx`). Gating the `Header` off entirely for logged-out visitors removed the page's sole top-level heading (the non-fitness branch then started at `<h2>Replies (n)</h2>`; the fitness branch had no heading at all). Since this is the public, crawlable, link-previewed surface, that's an accessibility/document-outline regression. **Fix:** the logged-out view keeps the design's no-back-button chrome but now renders a visually-hidden `<h1>` ("Post" / "Activity") so the outline and SEO are preserved.

Also addressed a sub-threshold layout note from the same review (top bar forcing ~64px of extra scroll), folded into the `min-h-dvh` flex-column restructure above.

Everything verified green locally: `prettier`, `lint`, `build`, `test` (3510 tests), plus manual browser checks of the logged-out (desktop + mobile) and signed-in status pages.

app/(timeline)/layout.tsx
```diff
@@ -22,6 +25,27 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
...
+      // the footer stays at the bottom without a mobile address-bar gap — same
+      // rationale as the public error pages (lib/components/error-page.tsx).
+      <div className="flex min-h-dvh flex-col">
+        <PublicTopBar />
+        <main className="flex flex-1 flex-col overflow-x-clip">
+          <div className="mx-auto flex w-full max-w-[680px] flex-1 flex-col px-4 py-6">
+            {children}
+          </div>
+          <PublicFooter />
+        </main>
```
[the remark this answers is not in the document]
llun: Fixed in 9193fd3a7 — moved `<PublicFooter>` out of `<main>` so its `<footer>` maps to the `contentinfo` landmark. `<main>` keeps `flex-1`, so the footer still pins to the bottom on short pages. (The same issue was independently flagged by a review sub-agent.)

llun:
### Sub-agent code review (round 2)

Ran a second independent review pass over the full updated diff. One reportable issue, now fixed in 9193fd3a7:

- **Footer landmark** (`app/(timeline)/layout.tsx`) — `<PublicFooter>` (a `<footer>`) was nested inside `<main>`, so it didn't expose the `contentinfo` landmark (a `<footer>` descended from `<main>` gets no role). Moved it to be a sibling of `<main>`. (Gemini flagged the same thing in parallel — nice convergence.)

Everything else was checked and cleared: the stat strip correctly uses `getOriginalStatus(status)` for Announce/boost statuses; `totalShares`/`totalLikes` are always defined post-Zod (`totalShares` defaults to 0); the `StatusLikes` `currentActor` guard is correct and test-covered; no `fetch()`-in-component / logger / `new Date()`-prop / `contentWidth` / `data-layout-width` violations; sr-only `<h1>` + `<h2>Replies>` ordering is sound; `Logo`/`Button` introduce no client hooks.

Local checks green: `prettier`, `lint`, `build`, `test` (3510).

llun:
### Sub-agent code review (round 3)

A third independent review pass found one **security** issue and one cleanup, both fixed in 2e4838f79:

- **🔒 Non-public ancestors leaked to logged-out visitors** (`app/(timeline)/[actor]/[status]/page.tsx`). The focused-status visibility gate protects the focused post, but the ancestor thread chain (`previouses`) was built with an unfiltered `database.getStatus` (`getStatusWithHydration` does a bare `where('id').first()` with no visibility filter) and rendered unconditionally. So a **public reply to a followers-only or direct parent** leaked that private parent's content to anonymous viewers. (This was pre-existing on `main`, but it lives in exactly the logged-out render path this PR is about, so fixing it here.) **Fix:** for logged-out visitors, stop the ancestor chain at the first non-public ancestor; factored the public/unlisted check into a shared `isPublicOrUnlistedStatus` helper. Added regression tests (`page.visibility.test.tsx`) covering both the hidden-private-parent and shown-public-parent cases.
- **Dead code** in `app/(timeline)/layout.tsx`: after the logged-out early return, `actor` is always present, so `showNavigation` and the optional-`user`/`currentActor` branches were unreachable. Simplified to render the nav chrome unconditionally — no behavior change for signed-in users.

> Note on remaining scope: the **logged-in** ancestor chain still has no per-actor (followers/DM) visibility check for non-followers — a separate, pre-existing concern broader than this logged-out PR. Happy to follow up in its own PR if you'd like.

Verified green: `prettier`, `lint`, `build`, `test` (3512 tests).

llun:
### Sub-agent code review (round 4)

A fourth review pass confirmed the round-3 ancestor guard is correct and complete for its scope, and found the **descendant half of the same leak**, now fixed in 880dc47ad:

- **🔒 Private replies leaked to logged-out viewers** (`app/(timeline)/[actor]/[status]/page.tsx`). The replies list was fetched via `getStatusReplies` with neither `publicOnly` nor `visibleToActorId`, then rendered unconditionally — so a **followers-only or direct reply to a public status** was shown in full to anonymous visitors, and its existence was also counted in the stat strip (`replies.length`). **Fix:** for logged-out visitors, fetch with `publicOnly: true` and add an object-level `isStatusPubliclyReadable` backstop (which also covers the embedded-replies path), then render and count only the visible replies. The stat strip now reflects the filtered count.
- Refactor: reuse the canonical `isStatusPubliclyReadable` (`lib/services/statusAccess.ts`, which uses `getVisibility` and handles Announce recursion) for both the ancestor guard and the reply filter, instead of the ad-hoc helper from round 3. The focused-status visibility gate is left byte-identical to `main`.
- Tests: extended `page.visibility.test.tsx` to assert the private reply is hidden **and** excluded from the reply count (3 cases now).

> Remaining out-of-scope note (unchanged from round 3): the **logged-in** reply/ancestor visibility is still unfiltered for non-recipients — a separate, pre-existing concern broader than this logged-out PR. The agent flagged it; happy to address in a dedicated follow-up PR rather than widen this one.

Verified green: `prettier`, `lint`, `build`, `test` (3513 tests).

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -117,11 +126,7 @@ const Page: FC<Props> = async ({ params }) => {
   }
 
   // Check if the status is publicly visible (public or unlisted)
-  const isPublicOrUnlisted =
-    status.to.includes(ACTIVITY_STREAM_PUBLIC) ||
-    status.to.includes(ACTIVITY_STREAM_PUBLIC_COMPACT) ||
-    status.cc.includes(ACTIVITY_STREAM_PUBLIC) ||
-    status.cc.includes(ACTIVITY_STREAM_PUBLIC_COMPACT)
+  const isPublicOrUnlisted = isPublicOrUnlistedStatus(status)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: Already fixed in 880dc47ad (this review ran against the earlier 2e4838f commit). Logged-out visitors now get `getStatusReplies({ ..., publicOnly: true })` plus an object-level `isStatusPubliclyReadable` backstop that also covers the embedded-replies path, and the filtered list drives both the rendered replies and the stat-strip count. Regression tests in `page.visibility.test.tsx` cover it.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -27,12 +27,21 @@ import { Header } from './Header'
...
   params: Promise<{ actor: string; status: string }>
 }
 
+// A status is publicly visible (no authentication required) when it addresses
+// the public collection in either its `to` or `cc`.
+const isPublicOrUnlistedStatus = (status: Status): boolean =>
+  status.to.includes(ACTIVITY_STREAM_PUBLIC) ||
+  status.to.includes(ACTIVITY_STREAM_PUBLIC_COMPACT) ||
+  status.cc.includes(ACTIVITY_STREAM_PUBLIC) ||
+  status.cc.includes(ACTIVITY_STREAM_PUBLIC_COMPACT)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: The ad-hoc helper this comment targets was removed in 880dc47ad — the page now uses the canonical `isStatusPubliclyReadable` from `lib/services/statusAccess.ts` (which goes through `getVisibility`). `to`/`cc` are also non-optional `z.string().array()` on `StatusBase`, so they are always arrays after Zod parse and cannot be undefined on a parsed `Status`. No crash risk here.

app/(timeline)/layout.tsx
```diff
@@ -38,77 +65,67 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
...
   // Fitness is a first-class section for every signed-in local account (like
   // Bookmarks/Messages), so new users can discover the import/Strava setup even
   // before they have any activity. The Overview itself handles the empty state.
-  const fitnessUrl = actor?.account ? '/fitness' : undefined
-  const isAdmin = actor?.account?.role === 'admin'
+  const fitnessUrl = actor.account ? '/fitness' : undefined
+  const isAdmin = actor.account?.role === 'admin'
 
   return (
     <div className="min-h-screen">
```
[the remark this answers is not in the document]
llun: Done in 360236066 — switched the signed-in layout outer wrapper to `min-h-dvh` to match the logged-out branch and the error-page precedent.

app/(timeline)/layout.tsx
```diff
@@ -38,77 +65,67 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
...
+      />
+      <MobileNav
+        unreadCount={unreadCount}
+        fitnessUrl={fitnessUrl}
+        profileUrl={`/${user.handle}`}
+        isAdmin={isAdmin}
+      />
       <main
         className={cn(
           'flex min-h-screen flex-col overflow-x-clip pb-6',
```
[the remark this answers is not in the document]
llun: Done in 360236066 — the signed-in `<main>` now uses `min-h-dvh` as well.

llun:
Addressed the latest round: replies leak was already fixed in 880dc47ad, the optional-chaining helper was replaced by the canonical `isStatusPubliclyReadable`, and the signed-in layout now uses `min-h-dvh` (360236066). Requesting another pass.

/gemini review

llun:
### Sub-agent code review (round 6)

A convergence pass confirmed all prior-round fixes hold and the stat-strip counts, `StatusLikes`/`Header` gating, `SignInCallout` placement, signed-in layout path, and tests are correct. It found one remaining case of the same visibility class, now fixed in 964dccc89:

- **🔒 Public boost of a private post leaked to logged-out viewers** (`app/(timeline)/[actor]/[status]/page.tsx`). The focused-status gate (`isPublicOrUnlisted`) only inspects the **top-level** `to`/`cc`. For an `Announce`, those are the boost's recipients — not the boosted note's. So a public self-boost of a followers-only Note passed the gate, and `getOriginalStatus(status)` then rendered the private original to anonymous viewers. **Fix:** an early guard — logged-out visitors must pass the recursive `isStatusPubliclyReadable(status)` (which descends into the Announce's `originalStatus`) — returns `notFound()` otherwise. Added a regression test (`page.visibility.test.tsx`, now 4 cases).

The focused-status gate's authenticated-user permission logic (followers-only / DM checks) is unchanged.

Verified green: `prettier`, `lint`, `build`, `test` (3514 tests).

llun:
### Review status — converged

CI is green (Lint/Prettier, Build, CodeQL, All Tests). A final convergence review pass (sub-agent) confirms the **logged-out visibility model is complete** and found no new must-fix issues:

- Three independent object-level `isStatusPubliclyReadable` checkpoints — focused status (recurses into an Announce's original), ancestors (`previouses`), and replies (+ stat-strip count) — each applied after the DB query, so the guarantee doesn't rely on the SQL filter alone.
- Probed and clean: the `RemoteStatusLoading` path (gated behind `&& session`), `generateMetadata` (no status content), the fitness-dashboard branch (behind the focused-status guard), the embedded-replies temporary-status path, and the action row (`Actions`/`StatusLikes`/`Header`/reply box all `currentActor`-gated).
- The 4 `page.visibility.test.tsx` cases each fail if their guard is removed (verified non-tautological).

The one remaining out-of-scope item — **logged-in** reply/ancestor visibility for non-recipients — is pre-existing and tracked for a separate follow-up PR.

Requesting a final pass on the latest commits (the visibility fixes landed after the previous review).

/gemini review

llun:
### Live verification of the visibility fix (local SQLite, logged-out)

Beyond the unit tests, I exercised the real filter path against a local DB. Seeded a **public** parent status with two replies — one **public**, one **followers-only** — then loaded the parent as a logged-out visitor:

- ✅ Public reply rendered
- ✅ Followers-only reply **not** rendered
- ✅ `Replies (1)` heading and stat strip `0 Boosts · 0 Likes · 1 Replies` both reflect only the visible reply

This confirms the real `getStatusReplies({ publicOnly: true })` SQL filter + the `isStatusPubliclyReadable` backstop + the filtered count all behave correctly against real rows (not just mocks). The logged-out chrome itself (top bar, footer, no back-button, stat strip, sign-in callout) was browser-verified earlier on desktop and mobile.

app/(timeline)/layout.tsx
```diff
@@ -22,6 +25,30 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
   const session = await getServerAuthSession()
   const actor = await getActorFromSession(database, session)
 
+  // Logged-out visitors get the public chrome — a slim top bar with sign-in
+  // CTAs and a footer in place of the nav sidebar — matching the web-public
+  // design. The reading column is narrower than the signed-in app width.
+  if (!actor) {
```
llun, line 31: [sub-agent review · test coverage · high] The logged-out early-return is load-bearing (it replaced the old `actor ? … : 0` null-guards) but has no test. If it regresses, anonymous visitors either fall through to `getActorsForAccount`/`getNotificationsCount` (crash) or get the signed-in Sidebar. Add a `layout.test.tsx` asserting: logged-out → PublicTopBar/PublicFooter and no Sidebar/MobileNav; logged-in → Sidebar/MobileNav and no public chrome.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -194,7 +222,13 @@ const Page: FC<Props> = async ({ params }) => {
   if (isFitnessDashboard) {
     return (
       <div className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
-        <Header isFitnessDashboard />
+        {currentActorProfile ? (
+          <Header isFitnessDashboard />
+        ) : (
+          // Logged-out view has no back-button chrome (matching the web-public
+          // design), but keep a top-level heading for the document outline.
+          <h1 className="sr-only">Activity</h1>
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:250]: [sub-agent review · test coverage · medium] The `isFitnessDashboard` logged-out branch (sr-only `<h1>Activity</h1>` + its own StatusStatStrip) is never exercised — all visibility tests use a plain Note. Add a page test with `fitness.processingStatus === "completed"` and a null actor asserting the sr-only "Activity" heading + stat strip render and the Header is absent.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -109,13 +120,23 @@ const Page: FC<Props> = async ({ params }) => {
...
     replies = status.replies as Status[]
   } else {
-    // Otherwise fetch from database
+    // Otherwise fetch from database. Logged-out visitors only ever see public
+    // replies, so filter them out in the query.
     replies = await database.getStatusReplies({
       statusId,
-      url: statusUrl
+      url: statusUrl,
+      ...(currentActor ? {} : { publicOnly: true })
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx]: [sub-agent review · test coverage · low] The `publicOnly: true` query arg for logged-out is not asserted (test 3 feeds an unfiltered mock and only proves the object-level backstop). A cheap `expect(mockGetStatusReplies).toHaveBeenCalledWith(expect.objectContaining({ publicOnly: true }))` would lock the query contract.

app/(timeline)/PublicTopBar.tsx
```diff
@@ -0,0 +1,24 @@
+import Link from 'next/link'
+import { FC } from 'react'
+
+import { Logo } from '@/lib/components/layout/logo'
+import { Button } from '@/lib/components/ui/button'
+
+// Public top bar shown to logged-out visitors in place of the app nav sidebar.
+// Mirrors ui_kits/web-public (PublicApp.jsx): a slim sticky bar with the brand
```
llun: [sub-agent review · comment accuracy · low] "Mirrors ui_kits/web-public (PublicApp.jsx)" reads as an in-tree path, but that file is not in the repo (it is the external design handoff). Reword to reference the web-public design without implying a locatable file.

app/(timeline)/PublicFooter.tsx
```diff
@@ -0,0 +1,38 @@
+import Image from 'next/image'
+import { FC } from 'react'
+
+// Public footer shown to logged-out visitors. Mirrors ui_kits/web-public
```
llun: [sub-agent review · comment accuracy · low] Same dangling "ui_kits/web-public (PublicApp.jsx)" reference as PublicTopBar — not an in-tree path. Reword to avoid implying a locatable file.

app/(timeline)/layout.tsx
llun: Fixed in 269b97d9b — added `app/(timeline)/layout.test.tsx` asserting logged-out renders PublicTopBar/PublicFooter with no Sidebar/MobileNav, and signed-in renders Sidebar/MobileNav with no public chrome.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:250]: Fixed in 269b97d9b — added a page.visibility.test case for the logged-out fitness-dashboard branch asserting the sr-only `<h1>Activity</h1>`, the stat strip, and the focused status render.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: Done in 269b97d9b — test 3 now also asserts `getStatusReplies` is called with `publicOnly: true` for logged-out visitors, locking the query contract.

app/(timeline)/PublicTopBar.tsx
llun: Fixed in 269b97d9b — reworded to "matching the web-public design" without the `ui_kits/web-public (PublicApp.jsx)` path.

app/(timeline)/PublicFooter.tsx
llun: Fixed in 269b97d9b — same reword applied to PublicFooter.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+    // getStatusReplies is mocked, so the in-page backstop filter is what must
+    // drop the private reply here.
+    mockGetStatusReplies.mockResolvedValue([publicReply, privateReply])
+
+    await renderPage()
+
+    expect(screen.getByTestId('status-public-reply')).toBeInTheDocument()
+    expect(screen.queryByTestId('status-private-reply')).not.toBeInTheDocument()
+    // The stat strip reply count reflects only the visible (public) reply.
+    expect(screen.getByText('Replies').previousSibling).toHaveTextContent('1')
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: [sub-agent review · test quality · low] `getByText('Replies').previousSibling` relies on the exact DOM sibling order inside StatusStatStrip (count `<strong>` immediately before the label). It would break on a harmless markup reorder even though the reply-count behavior is unchanged. Prefer a structural assertion, e.g. `getByText('Replies').parentElement` with `toHaveTextContent('1Replies')`, matching the pattern already used in StatusStatStrip.test.tsx.

Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: Fixed in aec518729 — switched to `getByText('Replies').parentElement` with `toHaveTextContent('1Replies')`, matching StatusStatStrip.test.tsx.

llun:
### Sub-agent review loop — converged (clean round)

Switched the review loop to sub-agent reviewers (Gemini was rate-limited after the earlier rounds). Summary of this phase:

- **Round 1** (test-coverage / comment-accuracy / type-design lenses): 5 findings — added `layout.test.tsx` (logged-out vs signed-in chrome), a logged-out fitness-dashboard test + a `publicOnly` query assertion, and reworded the design-kit comments. All posted as inline comments, fixed, and resolved.
- **Round 2** (full-diff convergence / test-quality): 1 finding — made the reply-count assertion structural instead of sibling-order. Fixed and resolved.
- **Round 3** (security/silent-failure / final correctness): **NO FINDINGS** from either reviewer — the three anonymous-visitor surfaces (focused status, replies, ancestors) are each gated by `isStatusPubliclyReadable`; no swallowed errors, no signed-in regressions, conventions clean.

CI is green (Lint/Prettier, Build, CodeQL, All Tests), 3517 tests pass, and all review threads are resolved. The PR is ready.

The only remaining item is the **deferred logged-in non-recipient reply/ancestor visibility** (pre-existing), which belongs in its own follow-up PR.

llun:
### Rebased onto latest `main`

Rebased onto `main`, which now includes #892 ("enforce per-viewer visibility for status replies and ancestors"). That PR implemented the per-viewer reply + ancestor filtering (both logged-in and logged-out) that overlapped my earlier logged-out-only commits, so I reconciled rather than duplicating:

- **Replies + ancestor visibility**: kept #892's version (`visibleToActorId` for logged-in / `publicOnly` for logged-out, the logged-out `isStatusPubliclyReadable` backstop, and `canActorReadStatus` for the ancestor chain). My now-redundant logged-out-only commits were dropped.
- **Kept (unique to this PR)**: the logged-out public chrome (top bar + footer), the read-only stat strip, the Header→sr-only-`<h1>` gating, the `StatusLikes` gating, and the **public-Announce-of-private focused-status `notFound` guard** — which #892's description explicitly left to this PR.
- **Tests**: layered my two unique cases (logged-out fitness-dashboard; public-boost-of-private → notFound) on top of #892's visibility suite.

Net diff is now exactly the 10 files unique to this PR. Local `prettier`/`lint`/`build`/`test` (3575 tests) all green.

PR #892, opened 2026-06-06
fix: enforce per-viewer visibility for status replies and ancestors

llun:
## Summary

The status detail page (`app/(timeline)/[actor]/[status]/page.tsx`) rendered **replies** and the **ancestor (`previouses`) chain** without scoping them to what the current viewer may read. A **logged-in viewer who is not a recipient/follower** could see followers-only or direct-message replies and ancestors of a public status they should not be able to read. This is a pre-existing privacy issue (replies/ancestors were unfiltered for *everyone* on `main`).

## Changes

- **Replies query** — pass `visibleToActorId: currentActor.id` to `getStatusReplies` for logged-in viewers (and `publicOnly: true` for logged-out). The database query (`applyPotentiallyReadableStatusFilter`) is the authoritative visibility check; it is unit-tested in `lib/database/sql/status.test.ts` (public/direct/hidden exclusion + followers-only-with-accepted-follow).
- **Replies backstop** — kept the logged-out-only object-level `isStatusPubliclyReadable` filter to cover the embedded-replies path (temporary statuses, which never hit the query) and stale recipient rows. **Logged-in viewers are intentionally not re-filtered in-page**, so the query's recipientless-reply branch (replies to the viewer's own posts) is preserved — re-running `canActorReadStatus` here would wrongly hide those.
- **Ancestors** — gate each `previouses` entry with the canonical `canActorReadStatus`, which collapses to the public/unlisted check when there is no current actor, covering both logged-out and logged-in viewers.
- **Focused-status gate** — unchanged.

## Tests

New `app/(timeline)/[actor]/[status]/page.visibility.test.tsx` covering logged-out and logged-in non-recipient cases (followers-only / DM ancestors hidden, public/followed ancestors shown, `visibleToActorId` forwarded to the query). Full suite green (3513 tests), lint and build clean.

## Scope notes for reviewers

- **Overlaps open PR #890** (`feat/public-status-logged-out`). Both create `page.visibility.test.tsx` and edit the same `getStatusReplies` call and `previouses` loop. This branch is a **superset** (logged-out + logged-in) built on `main` per request; the two PRs will need reconciliation on merge.
- **Logged-out focused-status Announce-of-private leak is *not* fixed here** — it remains #890's scope (its recursive `isStatusPubliclyReadable(status)` focused-status gate). Note that `previouses` are never built for Announce statuses (`status.type !== Announce` guard), so this change's ancestor gate does not touch the Announce case; the gap is purely the focused status.
- **Embedded-replies path stays unfiltered for logged-in** (the `status.replies` branch). Matches #890 and is low-risk — remote temporary statuses federate public replies.

### DIFF
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -8,6 +8,10 @@ import { getDatabase } from '@/lib/database'
 import { FETCH_REMOTE_STATUS_JOB_NAME } from '@/lib/jobs/names'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import { getQueue } from '@/lib/services/queue'
+import {
+  canActorReadStatus,
+  isStatusPubliclyReadable
+} from '@/lib/services/statusAccess'
 import { getActorProfile } from '@/lib/types/domain/actor'
 import { FollowStatus } from '@/lib/types/domain/follow'
 import {
@@ -109,11 +113,31 @@ const Page: FC<Props> = async ({ params }) => {
     // If replies are embedded (e.g. temporary status), use them
     replies = status.replies as Status[]
   } else {
-    // Otherwise fetch from database
+    // Otherwise fetch from database. The query filters to statuses the current
+    // viewer may read: logged-out visitors only ever see public/unlisted
+    // replies, while logged-in viewers get replies scoped to their own
+    // visibility (own posts, direct messages addressed to them, and
+    // followers-only posts from authors they have an accepted follow with).
     replies = await database.getStatusReplies({
       statusId,
-      url: statusUrl
+      url: statusUrl,
+      ...(currentActor
+        ? { visibleToActorId: currentActor.id }
+        : { publicOnly: true })
     })
+  }
+
+  // Object-level visibility backstop for logged-out visitors. Covers the
+  // embedded-replies path above (which never hits the query filter) and any
+  // stale/malformed recipient rows the query might miss, so a private reply to
+  // a public status never reaches an anonymous viewer (and isn't counted in the
+  // reply heading). Logged-in viewers rely on the query's `visibleToActorId`
+  // filter instead: it is the authoritative visibility check (it also matches
+  // recipientless replies to the viewer's own posts), so re-filtering with the
+  // simpler `isStatusPubliclyReadable` here would wrongly hide replies they may
+  // read.
+  if (!currentActor) {
+    replies = replies.filter(isStatusPubliclyReadable)
   }
 
   // Check if the status is publicly visible (public or unlisted)
@@ -168,6 +192,20 @@ const Page: FC<Props> = async ({ params }) => {
       withReplies: false
     })
     while (previouses.length < 3 && replyStatus) {
+      // `getStatus` does no visibility filtering, so without this guard a public
+      // reply to a followers-only (or direct) parent would leak that private
+      // ancestor to a viewer who cannot read it. Stop the chain at the first
+      // ancestor the viewer may not read. `canActorReadStatus` collapses to the
+      // public/unlisted check when `currentActor` is null, covering both
+      // logged-out and logged-in viewers.
+      const canReadAncestor = await canActorReadStatus({
+        database,
+        status: replyStatus,
+        currentActor
+      })
+      if (!canReadAncestor) {
+        break
+      }
       previouses.push(replyStatus)
       // This should be impossible
       if (replyStatus.type === StatusType.enum.Announce) {
--- a/app/(timeline)/[actor]/[status]/page.visibility.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.visibility.test.tsx
@@ -1,0 +1,331 @@
+/**
+ * @jest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { render, screen } from '@testing-library/react'
+
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { Actor } from '@/lib/types/domain/actor'
+import { FollowStatus } from '@/lib/types/domain/follow'
+import { Status } from '@/lib/types/domain/status'
+import {
+  ACTIVITY_STREAM_PUBLIC,
+  ACTIVITY_STREAM_PUBLIC_COMPACT
+} from '@/lib/utils/activitystream'
+import { getActorFromSession } from '@/lib/utils/getActorFromSession'
+
+import Page from './page'
+import { resolveStatusFromPath } from './resolveStatusFromPath'
+
+jest.mock('next/navigation', () => ({
+  notFound: jest.fn(() => {
+    throw new Error('NEXT_NOT_FOUND')
+  })
+}))
+
+jest.mock('@/lib/config', () => ({
+  getConfig: jest.fn(() => ({
+    host: 'activities.local',
+    fitnessStorage: undefined,
+    mediaStorage: undefined
+  }))
+}))
+
+const mockGetStatus = jest.fn()
+const mockGetStatusReplies = jest.fn()
+const mockGetAcceptedOrRequestedFollow = jest.fn()
+
+jest.mock('@/lib/database', () => ({
+  getDatabase: jest.fn(() => ({
+    getStatus: mockGetStatus,
+    getStatusReplies: mockGetStatusReplies,
+    getAcceptedOrRequestedFollow: mockGetAcceptedOrRequestedFollow
+  }))
+}))
+
+jest.mock('@/lib/services/auth/getSession', () => ({
+  getServerAuthSession: jest.fn()
+}))
+
+jest.mock('@/lib/services/queue', () => ({
+  getQueue: jest.fn()
+}))
+
+jest.mock('@/lib/utils/getActorFromSession', () => ({
+  getActorFromSession: jest.fn()
+}))
+
+jest.mock('@/lib/utils/mapbox', () => ({
+  getPublicMapboxAccessToken: jest.fn(() => undefined)
+}))
+
+jest.mock('./resolveStatusFromPath', () => ({
+  ...jest.requireActual('./resolveStatusFromPath'),
+  resolveStatusFromPath: jest.fn()
+}))
+
+jest.mock('./Header', () => ({ Header: () => null }))
+jest.mock('./RemoteStatusLoading', () => ({ RemoteStatusLoading: () => null }))
+
+// Render each status as its id so we can assert which statuses reach the page.
+jest.mock('./StatusBox', () => ({
+  StatusBox: ({ status }: { status: { id: string } }) => (
+    <div data-testid={`status-${status.id}`} />
+  )
+}))
+
+const mockResolveStatusFromPath = jest.mocked(resolveStatusFromPath)
+const mockGetServerAuthSession = jest.mocked(getServerAuthSession)
+const mockGetActorFromSession = jest.mocked(getActorFromSession)
+
+const VIEWER_ID = 'https://activities.local/users/viewer'
+const AUTHOR_ID = 'https://activities.local/users/anna'
+
+const buildNote = (overrides: Partial<Status> = {}): Status =>
+  ({
+    id: 'note-id',
+    type: 'Note',
+    actorId: AUTHOR_ID,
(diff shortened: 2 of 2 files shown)


app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+    expect(screen.getByTestId('status-public-reply')).toBeInTheDocument()
+    expect(screen.getByTestId('status-followed-parent')).toBeInTheDocument()
+  })
+
+  // Reply visibility for logged-in viewers is enforced by the database query
+  // (`visibleToActorId`), which is unit-tested in lib/database/sql/status.test.ts
+  // and correctly includes recipientless replies to the viewer's own posts. The
+  // page-level guarantee is that it forwards the viewer id to that query.
+  it('passes the viewer id to getStatusReplies so the query filters by visibility', async () => {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx:404]: **[HIGH] Add a positive logged-in reply test (guards the `if (!currentActor)` backstop gate).**

The logged-in suite only ever resolves an *empty* replies array from `getStatusReplies`, so nothing asserts that the public-only backstop (`replies.filter(isStatusPubliclyReadable)`, gated by `if (!currentActor)`) is **not** applied to authenticated viewers. A regression that dropped the `if (!currentActor)` guard — or moved the filter above it — would silently hide replies a logged-in viewer is entitled to see (DMs addressed to them, followers-only replies from authors they follow, recipientless replies to their own posts) and every existing test would still pass.

Suggestion: add a logged-in test where `getStatusReplies` resolves a non-public reply the query deems visible (e.g. `to: [VIEWER_ID]`), and assert it renders + `Replies (1)` shows.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+  })
+  render(element)
+}
+
+describe('Page visibility for logged-out visitors', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockGetServerAuthSession.mockResolvedValue(null)
+    mockGetActorFromSession.mockResolvedValue(null)
+    mockGetStatusReplies.mockResolvedValue([])
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx:146]: **[MEDIUM] Assert the logged-out branch calls `getStatusReplies` with `publicOnly: true` (and not `visibleToActorId`).**

The logged-in suite asserts `visibleToActorId` is forwarded (line ~327), but there is no symmetric assertion for the logged-out branch. The query-level filter is the primary defense for anonymous visitors; the object backstop is only a secondary net. A regression passing `visibleToActorId: currentActor?.id` (→ `undefined`) or dropping `publicOnly` for logged-out would remove authoritative DB-side scoping and rely entirely on the in-page filter, masking a real privacy regression.

Suggestion: in the logged-out describe, assert `getStatusReplies` was called with `expect.objectContaining({ publicOnly: true })` and `expect.not.objectContaining({ visibleToActorId: expect.anything() })`.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockGetServerAuthSession.mockResolvedValue({} as never)
+    mockGetActorFromSession.mockResolvedValue(buildViewer())
+    mockGetStatusReplies.mockResolvedValue([])
+    // Default: the viewer does not follow anyone.
+    mockGetAcceptedOrRequestedFollow.mockResolvedValue(null)
+  })
+
+  it('does not render a followers-only ancestor the viewer cannot read', async () => {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: **[MEDIUM] Test that the ancestor chain STOPS (breaks) at the first unreadable parent — not just that the parent is hidden.**

The loop `break`s the moment `canActorReadStatus` returns false, so the climb halts at a private boundary. Current ancestor tests use a single parent, so a regression changing `break` to skip-and-continue (excluding the private node but walking past it to a readable grandparent) would leak the grandparent of a private thread and no test would fail.

Suggestion: add a 3-level chain — focused public reply → followers-only/DM parent (unreadable) → public grandparent — wiring `mockGetStatus` to return parent then grandparent, and assert **both** the parent and the public grandparent are absent.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+    await renderPage()
+
+    expect(screen.getByTestId('status-public-reply')).toBeInTheDocument()
+    expect(
+      screen.queryByTestId('status-private-parent')
+    ).not.toBeInTheDocument()
+  })
+
+  it('renders a public ancestor of a public reply', async () => {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx:175]: **[LOW] Add a multi-level all-readable ancestor chain test.**

Only single-parent ancestor rendering is covered, so the loop’s depth cap (`previouses.length < 3`) and the parent-of-parent re-fetch (`getStatus(replyStatus.reply)`) are unexercised. An off-by-one on the cap or wrong re-fetch id would truncate/alter a readable thread undetected.

Suggestion: add a fully-public 3-level chain (reply → parent → grandparent) and assert all three ancestor StatusBoxes render.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+    await renderPage()
+
+    expect(screen.getByTestId('status-public-reply')).toBeInTheDocument()
+    expect(
+      screen.queryByTestId('status-private-parent')
+    ).not.toBeInTheDocument()
+  })
+
+  it('does not render a direct-message ancestor the viewer is not addressed in', async () => {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: **[LOW] Collapse the two near-identical logged-in ancestor tests into `it.each` (AGENTS.md testing convention).**

The tests at ~line 234 (followers-only ancestor) and ~line 260 (DM ancestor) share identical setup/assertion shape and differ only by the parent’s `to` field and the matching test id. AGENTS.md: "For tests whose cases differ only by input and expected output, prefer a table-driven `it.each([...])` with a description column instead of repeating near-identical it blocks."

Suggestion: merge into `it.each([{ description, parentTo, parentId }])(...)`.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -168,6 +192,20 @@ const Page: FC<Props> = async ({ params }) => {
       withReplies: false
     })
     while (previouses.length < 3 && replyStatus) {
+      // `getStatus` does no visibility filtering, so without this guard a public
+      // reply to a followers-only (or direct) parent would leak that private
+      // ancestor to a viewer who cannot read it. Stop the chain at the first
+      // ancestor the viewer may not read. `canActorReadStatus` collapses to the
+      // public/unlisted check when `currentActor` is null, covering both
+      // logged-out and logged-in viewers.
+      const canReadAncestor = await canActorReadStatus({
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:201]: **[MEDIUM] Heads-up: the focused-status gate diverges from this new `canActorReadStatus` ancestor gate.**

This PR now uses the canonical `canActorReadStatus` for ancestors (and `isStatusPubliclyReadable` for the reply backstop), but the focused-status check above (~line 160) is still a bespoke inline reimplementation. They disagree on followers detection: the inline gate uses `item.endsWith(/followers)`, whereas `hasFollowersAudience` in `statusAccess.ts` exact-matches `actor?.followersUrl` (or the canonical `${actorId}/followers`). For a remote actor whose stored `followersUrl` does not literally end with `/followers`, the same followers-only status is classified as a DM by the focused gate (→ `notFound` for a legitimate follower) but handled correctly as an ancestor.

This is pre-existing and the task scoped the focused gate as unchanged (its Announce branch also overlaps #890), so I do not plan to consolidate it in this PR — flagging for tracking.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx:404]: Fixed in 9c0fbd7b8. Added a logged-in test (`renders a non-public reply the query returns and does not apply the public-only backstop`) that resolves a DM addressed to the viewer from `getStatusReplies` and asserts it renders + `Replies (1)`. Dropping the `if (!currentActor)` guard would now fail this test.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx:146]: Fixed in 9c0fbd7b8. Added `queries getStatusReplies with publicOnly and without a viewer id` to the logged-out suite, asserting `expect.objectContaining({ publicOnly: true })` and `expect.not.objectContaining({ visibleToActorId: expect.anything() })`.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: Fixed in 9c0fbd7b8. Added `stops climbing the ancestor chain at the first unreadable parent`: a 3-level chain (public reply → followers-only parent → public grandparent) with `mockGetStatus.mockImplementation` keyed on statusId, asserting both the private parent and the readable grandparent beyond it are absent — proving `break` semantics.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx:175]: Fixed in 9c0fbd7b8. Added `renders a multi-level public ancestor chain` (reply → parent → grandparent, all public) asserting all three ancestor boxes render, exercising the depth cap and the parent-of-parent re-fetch.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx]: Fixed in 9c0fbd7b8. Collapsed the followers-only and direct-message ancestor cases into a single `it.each([{ description, parentId, parentTo }])` table per the AGENTS.md convention.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:201]: Acknowledged — valid latent inconsistency, but intentionally out of scope for this PR. The focused-status gate is pre-existing and unchanged here; the task explicitly scoped it as `keep unchanged`, and its Announce branch overlaps open PR #890 (which adds a recursive `isStatusPubliclyReadable` focused gate). Consolidating onto `canActorReadStatus` would also recurse into the Announce originalStatus and change that behavior, so it must wait for #890. Filed as a tracked follow-up to consolidate the focused gate (Note/Poll/DM + Announce) onto the shared helper once #890 lands. Resolving as deferred.

app/(timeline)/[actor]/[status]/page.visibility.test.tsx
```diff
@@ -0,0 +1,444 @@
...
+      actor: '@anna@activities.local',
+      status: 'hash'
+    })
+  })
+  render(element)
+}
+
+describe('Page visibility for logged-out visitors', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.visibility.test.tsx:140]: **[LOW] `jest.clearAllMocks()` does not reset `mockGetStatus`’s implementation — it leaks across tests.**

`clearAllMocks()` resets call records but not implementations. Unlike `mockGetServerAuthSession` / `mockGetActorFromSession` / `mockGetStatusReplies` / `mockGetAcceptedOrRequestedFollow` (all re-set in `beforeEach`), `mockGetStatus` is only set inside individual ancestor tests via `mockResolvedValue` / `mockImplementation`. So its implementation persists across tests and across both describe blocks.

It is currently benign — every test that reads `getStatus` sets its own implementation, and the rest use reply-less focused notes that never invoke it — but a future test that calls `getStatus` without setting it would silently inherit a prior test’s implementation, causing an order-dependent spurious pass/fail.

Suggestion: add `mockGetStatus.mockReset()` to both `beforeEach` blocks (or switch to `jest.resetAllMocks()`).

Comment (llun) [reply app/(timeline)/[actor]/[status]/page.visibility.test.tsx:140]: Fixed in a16cf7b5d. Added `mockGetStatus.mockReset()` to both `beforeEach` blocks so the implementation no longer leaks across tests.

PR #894, opened 2026-06-07
fix: consolidate focused-status visibility gate to canActorReadStatus

llun:
## Summary

The focused-status visibility gate in [`app/(timeline)/[actor]/[status]/page.tsx`](app/(timeline)/[actor]/[status]/page.tsx) was a bespoke inline reimplementation of visibility logic that diverged from the canonical `canActorReadStatus` / `hasFollowersAudience` helpers in [`lib/services/statusAccess.ts`](lib/services/statusAccess.ts) already used by the ancestor chain and reply backstop.

**The divergence:** the inline gate detected followers-only audience with `item.endsWith('/followers')`, whereas `hasFollowersAudience` exact-matches the actor's stored `followersUrl` (or canonical `${actorId}/followers`). For a remote actor whose stored `followersUrl` does **not** literally end in `/followers`, the inline gate misclassified a followers-only status as a direct message and returned `notFound()` for a legitimate follower — while the same status rendered correctly when it appeared as an ancestor.

This replaces the bespoke block (~44 lines) with a single early `canActorReadStatus` gate, mirroring the ancestor pattern, so all three code paths share one source of truth.

## Behavior changes

- **Fixes the remote-follower false-negative** — exact `followersUrl` match instead of a loose `/followers` suffix.
- **Closes the logged-in Announce-of-private focused-status leak** — `canActorReadStatus` recurses into the boosted original; the old block only checked the public boost wrapper. This is now in scope after #890.
- **Subsumes #890's logged-out gate** into the same unified call (it collapses to `isStatusPubliclyReadable` when `currentActor` is null), and moves the gate before the reply fetch so unreadable statuses are rejected up front for every viewer.
- Direct-recipient-of-followers-only is now allowed even without an accepted follow (matches `canActorReadSingleStatus`, which checks direct recipient before followers audience) — a correctness improvement.

Confirmed `status.actor.followersUrl` is hydrated on the focused status via `getStatus` → `getActorProfile`, so the fix holds in production, not just in tests.

## Tests

Added focused-status regressions in `page.visibility.test.tsx`: the divergence case (remote `followersUrl` not ending in `/followers`, viewer follows → renders), followers-only follow/no-follow, DM recipient/non-recipient, and Announce-of-private follow/no-follow. The divergence test was verified to fail against the old `endsWith('/followers')` logic and pass with the exact-match fix.

- `yarn lint` ✅ (0 errors)
- `yarn build` ✅
- `yarn test` ✅ (421 suites, 3581 tests)

### DIFF
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -13,16 +13,11 @@ import {
   isStatusPubliclyReadable
 } from '@/lib/services/statusAccess'
 import { getActorProfile } from '@/lib/types/domain/actor'
-import { FollowStatus } from '@/lib/types/domain/follow'
 import {
   Status,
   StatusType,
   getOriginalStatus
 } from '@/lib/types/domain/status'
-import {
-  ACTIVITY_STREAM_PUBLIC,
-  ACTIVITY_STREAM_PUBLIC_COMPACT
-} from '@/lib/utils/activitystream'
 import { cleanJson } from '@/lib/utils/cleanJson'
 import { getActorFromSession } from '@/lib/utils/getActorFromSession'
 import { getPublicMapboxAccessToken } from '@/lib/utils/mapbox'
@@ -99,12 +94,16 @@ const Page: FC<Props> = async ({ params }) => {
     return notFound()
   }
 
-  // Logged-out visitors may only view statuses that are publicly readable all
-  // the way down. The per-field `isPublicOrUnlisted` check below inspects only
-  // the top-level recipients — for an Announce those are the boost's, not the
-  // boosted note's — so a public boost of a followers-only post would otherwise
-  // slip through. `isStatusPubliclyReadable` recurses into the original status.
-  if (!currentActor && !isStatusPubliclyReadable(status)) {
+  // Focused-status visibility gate — the single source of truth shared with the
+  // ancestor chain below. `canActorReadStatus` collapses to the public/unlisted
+  // check for logged-out visitors and, for an Announce, recurses into the
+  // boosted status, so a public boost of a followers-only post is rejected
+  // unless the viewer follows the boosted author. It also exact-matches the
+  // author's stored `followersUrl` (rather than a loose `/followers` suffix), so
+  // a remote follower of a followers-only post is no longer misread as a
+  // direct-message non-recipient. Gating here, before the reply fetch, rejects
+  // unreadable statuses for every viewer up front.
+  if (!(await canActorReadStatus({ database, status, currentActor }))) {
     return notFound()
   }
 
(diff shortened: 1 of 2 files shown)


llun:
@cubic-dev-ai review this

llun:
### 🔍 Review round 1 (subagent reviewers)

Two independent reviewers (correctness/security + test-coverage) examined the diff, grounded in `lib/services/statusAccess.ts`.

**Correctness / security — no issues.** The consolidation strictly *tightens* visibility relative to the old code (never loosens):
- Logged-out path is identical to the prior early gate (`!currentActor → false`).
- Announce-of-private is now denied for logged-in non-followers — the old second gate computed `isPublicOrUnlisted` from the *boost's* top-level recipients only (non-recursive), so a public boost of a followers-only/DM post leaked to any logged-in viewer. The new gate recurses into `originalStatus`.
- The exact-match `hasFollowersAudience` fixes the remote-follower false-negative (a followers URL not ending in `/followers` was misread as a DM).
- Moved gate position (before the reply fetch) is side-effect-free; removed imports confirmed unused.

**Test coverage — one real gap (now fixed) + one medium gap (now fixed):**
- **Author viewing their OWN non-public focused status** had zero coverage anywhere (page tests *and* `statusAccess.test.ts`). The deleted inline block had an explicit "authors see their own" rule; the consolidation preserves it via the `currentActor.id === status.actorId` short-circuit. Added a dedicated `describe('Page visibility for the focused-status author')` with followers-only + DM cases, each asserting the status renders **and** that no follow lookup is performed (discriminating — a regressed author branch would 404).
- **Announce whose original is a DM addressed to the viewer** (recursion landing on the direct-recipient branch) was untested. Added both directions (recipient → renders; non-recipient → notFound).

Test count: 19 → 23. Lower-priority observations (redundant logged-out focused-note path, already covered by existing ancestor/announce tests) were judged not worth adding.

Proceeding to round 2.

llun:
### 🔍 Review round 2 (subagent reviewers)

Fresh lenses: test-correctness and code-quality/simplification.

**Test correctness — no issues.** The round-1 test additions are correct and genuinely discriminating: mock hygiene across the shared module-level mocks is sound (each `describe`'s `beforeEach` re-establishes implementations after `clearAllMocks`), the author-own block actually exercises the `currentActor.id === status.actorId` short-circuit (would 404 if removed), and the `not.toHaveBeenCalled()` assertions meaningfully pin the access *mechanism*. Verified by running the suite (23/23).

**Code quality — one cleanup applied.** `page.tsx` is clean (no dead code, no stale `isPublicOrUnlisted` references, comments accurate). The simplifier flagged that the 13-line Announce object literal was duplicated **5×** in the test file (4 new + 1 pre-existing). Extracted a `buildAnnounce(originalStatus, overrides?)` factory mirroring the existing `buildNote`/`buildViewer` style and collapsed all 5 call sites — removes ~55 lines with zero behavior change. Suite still 23/23, lint clean.

Proceeding to round 3.

llun:
### 🔍 Review round 3 (holistic final pass) — clean ✅

A holistic reviewer re-examined the full current diff against `lib/services/statusAccess.ts` / `statusAccess.test.ts`, specifically hunting for anything rounds 1–2 and the external bots missed. **No actionable issues** (nothing ≥ the confidence threshold):

- Both intended fixes confirmed correct and tested (remote-follower exact-match; logged-in Announce-of-private recursion).
- The one extra semantic change — direct-recipient checked **before** followers-audience, so a viewer named in `to`/`cc` of a followers-only post reads it without a follow lookup — is intentional, ActivityPub-correct, and covered by `statusAccess.test.ts`.
- Items examined and judged **pre-existing / out of scope**: `getVisibility`'s internal `endsWith('/followers')` classification (cannot produce a false-public outcome — it only gates a fast-path and falls through to the precise `hasFollowersAudience`), embedded-replies filtering for logged-in viewers, and ancestor-chain DB query volume.
- All new tests confirmed discriminating; `getRemoteStatus` is never reached by the fixtures (`isStatusHash: true`).

**Loop result:** issue counts across rounds went **2 → 1 → 0**; cubic-dev-ai reported *"No issues found"* and Gemini had no comments. Stopping at round 3 (well under the 5-round cap) since a full round surfaced nothing new.

Final gate: `prettier` ✅ · `lint` ✅ (0 errors) · `build` ✅ · `test` ✅ (421 suites / 3585 tests).

PR #922, opened 2026-06-08
minor: keyword filtering and stale-cursor fix for the list timeline

llun:
## What

The final two deferred follow-ups from the lists feature ([#907](https://github.com/llun/activities.next/pull/907)), after [#912](https://github.com/llun/activities.next/pull/912) (replies_policy), [#914](https://github.com/llun/activities.next/pull/914) (exclusive), and [#917](https://github.com/llun/activities.next/pull/917) (block/mute):

1. **Keyword filtering** — the list timeline now drops statuses matched by a `hide`-action keyword filter (using the `home` filter context, like the home feed), for **both** the Mastodon and `activities_next` formats. Previously the Mastodon path only *annotated* matches and the `activities_next` (web UI) path applied no keyword filtering at all.
2. **Stale-cursor fix** — `getListTimeline` returns an empty page when the `max_id`/`min_id` cursor status no longer exists, instead of silently dropping the cursor `WHERE` and re-returning the newest page (a pagination loop).

## How

**Keyword (`app/api/v1/timelines/list/[list_id]/route.ts`)**
- Fetch active `home`-context filters once, then `dropHideMatchesFromStatuses` for both formats. `warn` filters stay annotation-only on the Mastodon path (`annotateMastodonStatusesWithFilters`).
- **Cursors are computed from the RAW page (before keyword dropping).** This is the key correctness point: a page whose visible rows are *all* keyword-hidden still returns a non-null next cursor, so pagination reaches older non-hidden posts instead of terminating early. The hidden statuses still exist, so the raw cursor resolves on the next fetch.
- Per the [#917](https://github.com/llun/activities.next/pull/917) seam note, this is a thin post-fetch pass on top of `getListTimeline` — it does **not** route through `getFilteredStatusPage`, which would re-apply the block/mute filtering that #917 already does in SQL.

**Stale cursor (`lib/database/sql/list.ts`)**
- `applyCursor` now returns whether the cursor status was found; `getListTimeline` bails to `[]` when a provided cursor is missing. Mirrors the home timeline's `if (maxStatusId && !maxRow) return []`.

## Reviewer notes

- **Known limitation (by design):** dropping `hide`-matches post-fetch without a backfill loop means a keyword-heavy page can return fewer than `limit` visible statuses — a benign short page (the client does one extra round-trip). This avoids reversing #917's decision to keep block/mute in SQL; a fully-hidden page never stops pagination because cursors come from the raw batch.
- A regression test specifically asserts the no-premature-stop property (full hidden page → non-null next cursor) so a future "simplify the cursor onto the filtered array" change fails loudly.
- The stale-cursor `return []` is independent of the keyword work and only triggers for genuinely-deleted client cursors.

## Test

- `lib/database/sql/list.test.ts`: stale cursor returns `[]` when the cursor status is gone.
- `app/api/v1/timelines/list/[list_id]/route.test.ts`: hide-drop in both formats; whole-page-hidden keeps a next cursor (no premature stop).

## Verification

`prettier` / `lint` (0 errors) / `build` / `test` all green — **459 suites, 3873 tests passing**. No migration touched, so no schema-dump regeneration needed.

This completes all deferred backend follow-ups from the lists feature.

### DIFF
--- a/app/api/v1/timelines/list/[list_id]/route.test.ts
+++ b/app/api/v1/timelines/list/[list_id]/route.test.ts
@@ -196,4 +196,77 @@ describe('GET /api/v1/timelines/list/[list_id]', () => {
     expect(await response.json()).toEqual([])
     expect(response.headers.get('Link')).toBeNull()
   })
+
+  describe('keyword (hide) filtering', () => {
+    let spoilerStatus: Status
+
+    beforeAll(async () => {
+      await database.createFilter({
+        actorId: ACTOR1_ID,
+        title: 'Spoilers',
+        context: ['home'],
+        filterAction: 'hide',
+        expiresAt: null,
+        keywords: [{ keyword: 'spoiler', wholeWord: false }]
+      })
+      spoilerStatus = await database.createNote({
+        id: `${ACTOR1_ID}/statuses/spoiler-1`,
+        url: `${ACTOR1_ID}/statuses/spoiler-1`,
+        actorId: ACTOR1_ID,
+        to: [ACTIVITY_STREAM_PUBLIC],
+        cc: [],
+        text: 'spoiler alert'
+      })
+    })
+
+    it('drops hide-filtered statuses from the Mastodon format', async () => {
+      jest
+        .spyOn(database, 'getListTimeline')
+        .mockResolvedValue([listStatus, spoilerStatus])
+
+      const response = await GET(request(), {
+        params: Promise.resolve({ list_id: listId })
+      })
+
+      expect(response.status).toBe(200)
+      const ids = (await response.json()).map((s: { id: string }) => s.id)
+      expect(ids).toContain(urlToId(listStatus.id))
+      expect(ids).not.toContain(urlToId(spoilerStatus.id))
+    })
+
+    it('drops hide-filtered statuses from the activities_next format', async () => {
+      jest
+        .spyOn(database, 'getListTimeline')
+        .mockResolvedValue([listStatus, spoilerStatus])
+
+      const response = await GET(request({ format: 'activities_next' }), {
+        params: Promise.resolve({ list_id: listId })
+      })
+
+      expect(response.status).toBe(200)
+      const data = await response.json()
+      const ids = data.statuses.map((s: { id: string }) => s.id)
+      expect(ids).toContain(listStatus.id)
+      expect(ids).not.toContain(spoilerStatus.id)
+    })
+
+    it('keeps the next cursor when the whole page is hidden (no premature stop)', async () => {
+      jest.spyOn(database, 'getListTimeline').mockResolvedValue([spoilerStatus])
+
+      // Mastodon: empty body but a next Link so pagination reaches older posts.
+      const mastodon = await GET(request(), {
+        params: Promise.resolve({ list_id: listId })
+      })
+      expect(await mastodon.json()).toEqual([])
+      expect(mastodon.headers.get('Link') || '').toContain('rel="next"')
+
+      // activities_next: empty statuses but a non-null next cursor.
+      const next = await GET(request({ format: 'activities_next' }), {
+        params: Promise.resolve({ list_id: listId })
+      })
+      const data = await next.json()
+      expect(data.statuses).toEqual([])
+      expect(data.nextMaxStatusId).toBe(spoilerStatus.id)
+    })
+  })
 })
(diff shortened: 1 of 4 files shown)


app/api/v1/timelines/list/[list_id]/route.test.ts
```diff
@@ -196,4 +196,77 @@ describe('GET /api/v1/timelines/list/[list_id]', () => {
...
+
+  describe('keyword (hide) filtering', () => {
+    let spoilerStatus: Status
+
+    beforeAll(async () => {
+      await database.createFilter({
+        actorId: ACTOR1_ID,
+        title: 'Spoilers',
+        context: ['home'],
+        filterAction: 'hide',
```
Comment (llun) [inline app/api/v1/timelines/list/[list_id]/route.test.ts:209]: [subagent: keyword-parity] Test gap: this suite only exercises a `hide`-action filter. There is no test that a `warn`-action keyword filter is **not** dropped — which is the exact behavioral claim the new route comment makes ("`warn` filters are not dropped — they are surfaced via annotation"). A future refactor that made `dropHideMatchesFromStatuses` also drop `warn` matches would pass every existing test here.

Suggest adding a `warn`-filter case asserting:
1. The matched status is **kept** in both the Mastodon and activities_next formats (core regression guard).
2. Mastodon format populates `filtered` on the matched status (parity with the home feed annotation path).
3. activities_next keeps the status with **no** `filtered` annotation (internal domain Status shape carries no annotation).

llun:
[subagent: cursor] Reviewed cursor/pagination correctness — no issues found.

Stale-cursor fix (lib/database/sql/list.ts): correct. The existence check `applyCursor` runs against the raw `statuses` table by id, with no list-membership join / block-mute / visibility / replies-policy applied, so a cursor that exists but is filtered out of the list *view* still resolves (returns true) and pagination continues on its createdAt. The only case returning `[]` is a row genuinely absent from `statuses` (deleted cursor), where `[]` is the correct termination — no wrong empty page. Keyset uses strict `</>` with the id tie-breaker, no off-by-one. (Minor: the comment says it mirrors the home MAIN path, but it actually mirrors the LOCAL/public path at timeline.ts:43 which also queries `statuses` directly; MAIN is more lenient with a timelines→statuses fallback. Behavior is still correct, just a comment nit.)

Keyword drop (route.ts): correct. Both cursor representations (`nextMaxStatusId`/`prevMinStatusId` and Link `firstStatus`/`lastStatus`) are derived from the RAW page, and `dropHideMatchesFromStatuses` returns a new filtered array without mutating/reordering. So a fully-hidden page still emits a non-null next cursor (no premature stop), the cursor always points at a real existing row (no infinite loop — hidden ≠ deleted, so it re-resolves next fetch), and statuses[0]=newest / statuses[last]=oldest holds under the `createdAt desc, id desc` ordering. Empty Link headers / null cursors only when the DB page is truly empty. The two mechanisms (DB cursor vs route-level drop) are orthogonal and compose correctly.

app/api/v1/timelines/list/[list_id]/route.test.ts
Comment (llun) [reply app/api/v1/timelines/list/[list_id]/route.test.ts:209]: Fixed in 82cdc2798 — added a test asserting a `warn`-action keyword filter is **not** dropped: the status stays in both formats, carries a populated `filtered` annotation on the Mastodon path, and has no annotation on the activities_next path. Good catch.

llun:
[subagent r2] Static re-review after the round-1 fixes — no issues found. Confirmed: (1) getListTimeline returns [] only when the cursor status genuinely doesn't exist — applyCursor resolves the cursor against the global statuses table (unscoped by list/visibility), so a valid-but-filtered cursor still returns true and paginates; no drop-the-WHERE infinite loop. (2) nextMax/prevMin cursors and the Link headers both derive from the RAW page (statuses[0]/statuses[last]), dropHideMatchesFromStatuses is computed once and applied to both formats, warn annotation is Mastodon-path only, and a fully-hidden page keeps a non-null next cursor. (3) The warn-not-dropped test is non-tautological: it fails if warn is dropped, if the Mastodon filtered annotation is missing, or if activities_next is annotated. (4) Existing route tests (min_id-over-since_id, un-hydratable skip, empty-array, Link headers) still hold — the filters.length===0 early-returns keep the no-filter paths unchanged. Minor (non-blocking) notes: the valid-cursor happy path of applyCursor has no real-data DB test (only the gone-cursor branch is covered; route tests mock getListTimeline), and the Mastodon path evaluates filters twice per survivor (dropHide + annotate) vs partitionStatusesByFilters' single pass — benign given the regex cache.

llun:
[subagent r3] Final confirmation review — no issues found; PR looks ready. Stale-cursor fix correctly guards both max_id and min_id (return [] mirrors home timeline), cursors are derived from the raw page so all-hidden pages still advance pagination, hide-filtering uses the correct 'home' filter context for parity, and annotate/getMastodonStatuses both consume visibleStatuses (annotation is id-keyed, not index-aligned, so it stays robust). The no-backfill-loop difference vs the home feed is documented and intentional.

PR #991, opened 2026-06-15
refactor: convert project to ESM-only and migrate test runner to Vitest

llun:
## Summary

Converts the project to **ESM-only** (`"type": "module"`) and replaces **Jest with Vitest** as the test runner. This is the implementation of options **B1 + D** from the earlier ESM-conversion evaluation:

- **D — Vitest:** Jest could not run the project once it became ESM-only. Jest is a CommonJS-era runner: making the package `"type": "module"` forces Knex to load migrations via dynamic `import()` (needs `--experimental-vm-modules`), and ESM-mode Jest drops the implicit `jest` global and breaks `jest.mock()` hoisting across ~900 call sites. Vitest is ESM-native and avoids all of that.
- **B1 — schema dumps in tests:** test database setup now loads the committed reference dumps (`migrations/schema*.sql`) instead of running the Knex migration chain, which keeps the ESM migration files out of the test runtime entirely and speeds up per-file DB setup.

## What changed

**ESM-only**
- `package.json`: `"type": "module"`; `test` → `vitest run`.
- All 122 Knex migrations converted from `exports.up/down` + `require()` to native ESM (`export const up/down`, `import … from 'node:crypto'`). New `migration.stub` so `migrate:make` generates ESM.
- Converted a stray `require('@/package.json')` in `admin/system/page.tsx` to an import; `search:reindex` + mock-script docs use the swc-node ESM loader.
- Migration **schema is unchanged** (only module syntax changed), so the committed schema dumps did not need regeneration.

**Vitest**
- `vitest.config.ts`, `vitest.setup.ts`, `vitest.d.ts` (a `jest.Mock`/`MockedFunction`/`Mocked` type-compat shim so existing type annotations keep working), and a small `vitest-shims/jest-global.ts`.
- Codemodded `jest.*` → `vi.*` across all tests, `__mocks__`, scripts, and root tests.
- `jest-fetch-mock` is retained (lenient with relative URLs) behind a `vi`-backed global `jest` Proxy that also satisfies `@testing-library`'s fake-timer detection.
- Moved the `@/lib/config` manual mock to `lib/config/__mocks__/index.ts` (Vitest resolves `__mocks__` adjacent to the resolved module).
- Test-only cache-reset guards now key off `process.env.VITEST`.
- ESLint: dropped the Jest plugin/globals in favor of Vitest globals; removed Jest runner devDependencies.

## Verification

All four gates green on Node 24:
- `yarn prettier --write .` ✓
- `yarn lint` ✓ (0 errors)
- `yarn build` ✓
- `yarn test` ✓ — **508 test files / 4359 tests passing**

Docs (`AGENTS.md`) updated for the Vitest + schema-dump testing model.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TrUhnWo9L2WLoWvcjokEYm)_


llun:
Re: the automated review flagging `vi.importMock` (7 comments) — these are false positives, so I've left the code as-is.

`vi.importMock` **is** a valid Vitest API (it returns a mocked module). It's present in the installed Vitest 3.2.6 type definitions (`node_modules/vitest/dist/index.d.ts`: `importMock: <T = ESModuleExports>(path: string) => Promise<MaybeMockedDeep<T>>`) and is documented at https://vitest.dev/api/vi#vi-importmock. The full test suite (508 files / 4359 tests) passes using it, which confirms it resolves at runtime.

The suggested `await import(...)` rewrite would also work, but it isn't necessary — `vi.importMock` is the idiomatic equivalent of the previous `jest.requireMock` and keeps the diff minimal.

---

llun:
## Re: `vi.importMock` — to @gemini-code-assist and @kilo-code-bot

After review, we've **reverted to `vi.importMock`** as the standard for reading a mocked module in this codebase, and documented it in `REVIEW.md`, `AGENTS.md`, and `CLAUDE.md`. The repeated "CRITICAL: `vi.importMock()` is not a valid Vitest API" finding is **incorrect** — please don't flag it going forward.

**`vi.importMock` is a valid, documented Vitest API.** Evidence:
- Docs: https://vitest.dev/api/vi#vi-importmock
- Type definitions in the installed Vitest 3.2.6 — `node_modules/vitest/dist/index.d.ts`:
  ```ts
  importMock: <T = ESModuleExports>(path: string) => Promise<MaybeMockedDeep<T>>;
  ```
- The full suite (509 files / 4378 tests) passes using it.

**Why it's preferable to `(await import('…')) as unknown as T` here:**
1. It's purpose-built and is the direct equivalent of the `jest.requireMock` this code migrated from.
2. It returns `MaybeMockedDeep<T>`, so the result is already typed as a mock — **no `as unknown as` cast** needed (that cast erases type safety).
3. It always returns the mock. A bare `await import()` returns the *real* module unless it is separately `vi.mock()`'d, so the two are only equivalent by coincidence of an accompanying `vi.mock`.

The `await import()` form works too, but it's strictly worse ergonomically (requires a type-erasing double-cast), so we've standardized on `vi.importMock`. Thanks for the reviews otherwise.

---

PR #999, opened 2026-06-16
fix(deps): upgrade to vitest 4 + vite 8 to remove esbuild (fixes Dependabot alerts)

llun:
## Summary

Resolves both open [Dependabot security alerts](https://github.com/llun/activities.next/security/dependabot) — both for **esbuild** — by **removing esbuild from the dependency tree entirely**, rather than pinning it.

| Alert | Severity | Advisory | Issue |
| --- | --- | --- | --- |
| [#165](https://github.com/llun/activities.next/security/dependabot/165) | High | [GHSA-gv7w-rqvm-qjhr](https://github.com/advisories/GHSA-gv7w-rqvm-qjhr) | Missing binary integrity verification in the Deno module enables RCE via `NPM_CONFIG_REGISTRY` |
| [#164](https://github.com/llun/activities.next/security/dependabot/164) | Low | [GHSA-g7r4-m6w7-qqqr](https://github.com/advisories/GHSA-g7r4-m6w7-qqqr) | Arbitrary file read when running the dev server on Windows |

## Approach

esbuild was only present **transitively via `vite`** (the bundler used by the `vitest` test runner). It is dev/test-only — the Next.js production build never used it.

**Vite 8 dropped esbuild**, moving to **rolldown** (the Rust port of Rollup) + oxc. So upgrading the test toolchain removes esbuild at the source:

- `vitest` / `@vitest/coverage-v8` → `^4.1.9`
- added `vite` `^8.0.16` as a direct devDependency (forces the rolldown-based vite 8)
- removed the interim `esbuild` resolution

After `yarn install`, `yarn why esbuild` finds **nothing** — esbuild remains only an *optional, uninstalled* peer of vite, so both Dependabot alerts are moot (the vulnerable package is no longer in the lockfile).

## Migration fixes required by vitest 4

- **Mock constructors:** vitest 4 invokes mock implementations with `Reflect.construct` under `new`, so arrow-function implementations used as constructors threw `… is not a constructor`. Converted those to regular `function` expressions across 8 test files (IntersectionObserver / ResizeObserver, AWS `S3Client`, qstash `Client`, `fit-file-parser`, mapbox `Map`). Module-factory callbacks passed to `vi.mock(...)` are intentionally left as arrows — they are never constructed.
- **`environmentMatchGlobs` removed in vitest 4:** dropped from `vitest.config.ts` in favour of the per-file `@vitest-environment jsdom` docblock that **71 of 73** component tests already use. The 2 remaining `.test.tsx` files are server-render (`renderToStaticMarkup`) tests with no DOM usage, so they correctly default to the `node` environment. `environmentOptions` (jsdom URL) is retained.

## Verification

- `yarn why esbuild` → not installed (gone from the tree).
- `yarn lint` → 0 errors (29 pre-existing warnings only).
- `yarn build` → type-check + compile succeed.
- `yarn test` → **4384 tests / 509 files pass** on vitest 4 + vite 8.

No migrations changed; reference schema dumps untouched.

### DIFF
--- a/app/(timeline)/bookmarks/BookmarksTimeline.test.tsx
+++ b/app/(timeline)/bookmarks/BookmarksTimeline.test.tsx
@@ -96,7 +96,7 @@ describe('BookmarksTimeline', () => {
     })
     Object.defineProperty(globalThis, 'IntersectionObserver', {
       configurable: true,
-      value: vi.fn().mockImplementation((callback) => {
+      value: vi.fn().mockImplementation(function (callback) {
         intersectionObserverCallback = callback
         return {
           disconnect: intersectionObserverDisconnect,
@@ -106,10 +106,12 @@ describe('BookmarksTimeline', () => {
     })
     Object.defineProperty(globalThis, 'ResizeObserver', {
       configurable: true,
-      value: vi.fn().mockImplementation(() => ({
-        disconnect: vi.fn(),
-        observe: vi.fn()
-      }))
+      value: vi.fn().mockImplementation(function () {
+        return {
+          disconnect: vi.fn(),
+          observe: vi.fn()
+        }
+      })
     })
   })
 
--- a/app/(timeline)/favorites/FavoritesTimeline.test.tsx
+++ b/app/(timeline)/favorites/FavoritesTimeline.test.tsx
@@ -96,7 +96,7 @@ describe('FavoritesTimeline', () => {
     })
     Object.defineProperty(globalThis, 'IntersectionObserver', {
       configurable: true,
-      value: vi.fn().mockImplementation((callback) => {
+      value: vi.fn().mockImplementation(function (callback) {
         intersectionObserverCallback = callback
         return {
           disconnect: intersectionObserverDisconnect,
@@ -106,10 +106,12 @@ describe('FavoritesTimeline', () => {
     })
     Object.defineProperty(globalThis, 'ResizeObserver', {
       configurable: true,
-      value: vi.fn().mockImplementation(() => ({
-        disconnect: vi.fn(),
-        observe: vi.fn()
-      }))
+      value: vi.fn().mockImplementation(function () {
+        return {
+          disconnect: vi.fn(),
+          observe: vi.fn()
+        }
+      })
     })
   })
 
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -477,15 +477,17 @@ describe('RouteHeatmapMap', () => {
   })
 
   it('uses Mapbox when a token is configured', async () => {
-    const mapConstructor = vi.fn().mockImplementation(() => ({
-      on: (_event: string, callback: () => void) => callback(),
-      remove: vi.fn(),
-      resize: vi.fn(),
-      addSource: vi.fn(),
-      addLayer: vi.fn(),
-      getSource: vi.fn(),
-      fitBounds: vi.fn()
-    }))
+    const mapConstructor = vi.fn().mockImplementation(function () {
+      return {
+        on: (_event: string, callback: () => void) => callback(),
+        remove: vi.fn(),
+        resize: vi.fn(),
+        addSource: vi.fn(),
+        addLayer: vi.fn(),
+        getSource: vi.fn(),
+        fitBounds: vi.fn()
+      }
+    })
     mockLoadMapboxModule.mockResolvedValue({
       Map: mapConstructor
     })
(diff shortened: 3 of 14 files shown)


llun:
## 🤖 Independent sub-agent review (round 1)

Ran two independent review agents over this diff with distinct lenses. Both returned **APPROVE** — no blockers, no concerns.

**Dependency-correctness lens:**
- Yarn Berry 4.15.0 resolution syntax is correct; `"esbuild": "0.28.1"` overrides every esbuild descriptor regardless of requested range. Verified empirically: vite@7.3.5 declares `esbuild ^0.27.0`, yet `require('esbuild').version` → `0.28.1`.
- No stale esbuild copy remains; the one core entry + all 26 `@esbuild/*` platform binaries are at `0.28.1` (no core/binary skew).
- `git diff main...HEAD -- yarn.lock` touches only esbuild/`@esbuild/*` (`0.27.7 → 0.28.1`); no collateral version churn. `yarn install` reproduces the lock with zero further changes.
- Both open Dependabot alerts (#165 high, #164 low) are fully covered; no other alerts remain open.

**Convention-compliance lens:**
- `fix(deps):` prefix → patch bump (correct); `package.json` `version` field untouched (CI auto-bump preserved).
- No migration files touched → no schema-dump regeneration required.
- yarn-only policy respected; `yarn.lock` committed; no npm/pnpm artifacts.
- Scope is `package.json` + `yarn.lock` only — no runtime config, no `console.*`, no source changes.

**Non-actionable nits (no change made):**
- Exact pin `0.28.1` (vs `^0.28.1`) is the safer choice — esbuild treats 0.x minor bumps as breaking and vite validates the esbuild API; it also matches the existing exact-pinned `postcss` resolution. Future patches will surface as their own Dependabot PRs.

No code changes were required from this review round.

llun:
## Approach update (round 2)

Pivoted from the interim **esbuild 0.28.1 resolution pin** to **removing esbuild entirely** by upgrading the test toolchain to **vitest 4 + vite 8** (vite 8 uses rolldown/oxc, not esbuild). `yarn why esbuild` now finds nothing — both alerts are resolved at the source rather than patched.

This required vitest-4 migration fixes (arrow → `function` mock constructors across 8 test files; removal of the vitest-4-removed `environmentMatchGlobs` in favour of per-file `@vitest-environment` docblocks). Full rationale in the updated PR description.

Re-requesting review on the new diff.

/gemini review

llun:
## 🤖 Independent sub-agent review (round 2 — migration diff)

Ran three independent review agents over the vitest 4 + vite 8 migration with distinct lenses. All three returned **APPROVE**.

**1. Mock-conversion correctness:**
- All 8 files' arrow→`function` constructor conversions verified behavior-preserving: every source-side `new` confirmed (`new Client`/`new FitParser`/`new mapboxgl.Map`/`new S3Client`/`new GetObjectCommand`), closures (`intersectionObserverCallback`, `mockPublishJSON`, `send`, `mockS3Send`) and the `addSource` throw body intact.
- `vi.mock(...)` module-factory callbacks correctly left as arrows (never constructed).
- All 4 FitnessHeatmapView Map stubs converted; full suite green, 0 "is not a constructor" errors.
- **Caught one dormant arrow mock** in `lib/services/medias/index.test.ts` (latent, didn't fail because `getStorage` is stubbed). **Fixed** in a follow-up commit.

**2. Config / environment migration:**
- `vitest.config.ts` valid for vitest 4 (`environmentMatchGlobs` removed; `server.deps.inline`, `setupFiles`, `environmentOptions` all still valid v4 keys).
- **No DOM-coverage gap:** the only 2 `.test.tsx` files without a `@vitest-environment jsdom` docblock (`messages/page`, `admin/system/page`) use `renderToStaticMarkup`/return the element — no DOM — so they correctly default to node. The other 71 carry the docblock. No `.test.ts` needs the DOM.
- Repo-wide scan found no other vitest-3-only APIs (`workspace`, top-level `deps.inline`, `poolMatchGlobs`, etc.). `vi.mocked` call sites unaffected.

**3. Dependency / build integrity:**
- esbuild fully eliminated: `yarn why esbuild` empty, `yarn info esbuild` "No package matched", no `node_modules/esbuild`. Remaining yarn.lock mentions are vite 8's *optional, uninstalled* peer. Both alerts moot.
- Version pins consistent: `vitest === @vitest/coverage-v8 === 4.1.9` (exact peer satisfied), single `vite@8.0.16` resolution, no splits.
- No new unmet peers; node 24 satisfies vite/vitest engines. Lockfile churn is exactly the esbuild/rollup→rolldown swap; settled after install. `version` field untouched; vite/vitest are devDependencies so the shipped app is unaffected.

One review-driven fix applied (`test:` commit above). Otherwise clean.

llun:
## Round 3 — final confirmation

All CI checks are green on the latest commit (\`083ae556c\`), including **Kilo (No Issues Found → Merge)** and CodeQL. The only change since the last Gemini pass is the review-driven \`test:\` follow-up (one dormant arrow→function S3Client mock). Re-requesting a final review to confirm convergence.

/gemini review

llun:
## Post-rebase CI fix

After the rebase onto `main` (v1.74.4), the `All Tests` check flaked: `TwoFactorManager > disables 2FA after confirming the current password` failed in one workflow run while **the same commit passed in the other** — a flaky race the vitest 4 async scheduling exposes.

**Root cause:** the test `await waitFor(() => expect(mockDisable).toHaveBeenCalled())` (resolves the moment the mock is *called*), then reads the result text with a **synchronous** `getByText('2FA is off')`. That label only renders after the awaited `disable()` promise resolves and React re-renders, so `getByText` could run before the re-render flushed.

**Fix (`2f4662f3a`):** switched the post-action reads to `await screen.findByText(...)` (retries until the re-render lands) in all three structurally-identical tests in the file (disable, verify-setup, generate-backup-codes), matching the `findByDisplayValue` pattern the first test already uses. Stress-ran the file **25×** locally with 0 failures; full suite (509 files / 4384 tests), lint (0 errors), and build all green.

PR #1040, opened 2026-06-19
fix: make OIDC provider RS256-compliant for strict relying parties

llun:
## Summary

A live `authorization_code` + PKCE(S256) run against the provider (acting as the IdP for a `lasuite/impress` "Docs" app using `mozilla-django-oidc` with `OIDC_RP_SIGN_ALGO=RS256`) surfaced **three defects** that block a strict RS256 OpenID Connect relying party. This PR fixes all three. Mastodon OAuth2 clients (opaque access tokens + the `/api/v1/*` REST API) are unaffected.

### Fix 1 — id_token must be RS256 (was EdDSA)
`jwt()` was called with no args, so Better Auth defaulted to **EdDSA/Ed25519**, while `openidConfiguration.ts` advertises `id_token_signing_alg_values_supported: ['RS256']`. The live `/api/auth/jwks` therefore served only an EdDSA key and a strict RS256 client failed signature verification.

- Configured the plugin: `jwt({ jwks: { keyPairConfig: { alg: 'RS256', modulusLength: 2048 } } })` (option name verified against `better-auth@1.6.19`'s `JWKOptions`).
- `OAuthGuard` verifies bearer JWTs via `verifyAccessToken` → jose `createLocalJWKSet`/`jwtVerify`, which selects the key by **`kid`** and is **not pinned to an alg** (no `algorithms` in `verifyOptions`), so adding the RSA key is safe for it.

### Fix 2 — issuer inconsistency
Discovery set `issuer: baseURL` (`https://llun.social`), but Better Auth signs id_tokens with `iss = baseURL + basePath` (`https://llun.social/api/auth`). Aligned the **OIDC discovery** issuer to `` `${baseURL}/api/auth` ``. The RFC 8414 OAuth metadata (`oauthAuthorizationServer.ts`) intentionally keeps the **bare origin** for Mastodon compatibility (OAuth2 access tokens carry no `iss`).

### Fix 3 — userinfo `sub` ≠ id_token `sub` (OIDC §5.3.2)
id_token `sub` is the Better Auth **account** id; `getUserInfo` returned `urlToId(actor.id)`. 

- **Canonical subject decision: the account id.** One OIDC subject per human, regardless of how many ActivityPub actors the account owns. `getUserInfo` now returns `sub: account.id`; the `account` argument is **required**. Profile fields (`name`, `preferred_username`, `picture`, `profile`) stay **actor-sourced**.
- The userinfo route fails closed (`401 invalid_token`) if the token's resolved actor has no owning account, rather than emitting a `sub`-less (spec-invalid) response.

## Backward compatibility
- **Mastodon clients unaffected.** Verified live that their access tokens are **opaque** (no `resource`/`audience` ⇒ better-auth issues opaque, DB-verified tokens, not JWKS-verified JWTs). `verify_credentials` still returns the account.
- **userinfo `sub` value changes** (account id instead of actor-url id). A spec-compliant RP could not have worked before this PR anyway, since `id_token.sub` already ≠ `userinfo.sub`; no existing RP could have keyed users on the old value.

## ⚠️ Deployment / rollout note (EdDSA → RS256 key cutover)
This integration's `jwks` table has **no per-key `alg` column** (the plugin's jwks schema declares none), so better-auth resolves the signing/JWKS `alg` from **config**, not from each stored key. Consequences, confirmed empirically with `jose`:

- **Fresh deployment:** an RSA key is generated on the first sign / first `/api/auth/jwks` request and everything is consistent. ✅ (This is what local dev, the tests, and the e2e run below exercise.)
- **Existing deployment that already signed a token** (an Ed25519 key already sits in `jwks`): the plugin would load that stale key and try to sign it as RS256, which **throws** (`importJWK(ed25519Jwk, 'RS256')` → `JOSENotSupported`). The one-time rollout step is therefore to **clear the `jwks` table** so a fresh RSA key is generated.
- Because of the missing per-key `alg`, a retained Ed25519 key would also be **served stamped `alg: RS256`** and could never verify a legacy EdDSA token (`JWKSNoMatchingKey`) — so keeping it has no upside here. Clearing it is safe: Mastodon access tokens are opaque (DB-verified, not JWKS) and id_tokens are short-lived (consumed at login), so no long-lived token depends on the retired key. This is the one place the change deviates from the literal "keep EdDSA keys published" guidance — the deviation is forced by the missing `alg` column and is documented in `lib/services/auth/auth.ts`.

## Tests
- `lib/services/oauth/userinfo.test.ts` — `sub = account.id` (not actor id), `userinfo.sub === id_token.sub` invariant, `account` now required, email/profile scope behavior.
- `lib/services/wellknown/index.test.ts` — discovery `issuer === ${baseURL}/api/auth`, distinct from the bare-origin OAuth metadata issuer.
- `lib/services/auth/jwks.test.ts` (new) — boots a real `getAuth()` against in-memory SQLite and asserts `/api/auth/jwks` publishes an **RS256/RSA** key with a `kid` (and no EdDSA/OKP key).

`prettier`, `lint` (0 errors), `build`, and the full `vitest` suite (525 files / 4560 tests) all pass.

## End-to-end verification (running instance)
Full `authorization_code` + PKCE(S256) flow + Mastodon regression, all green:

```
id_token header  = {"alg":"RS256","kid":"QXOhairIHfIiTTki7K5PqDR8nzvyDeCp"}
id_token claims  = {"iss":"http://localhost:3000/api/auth","sub":"136a089a-…","aud":"<client_id>","email":"test@example.com","email_verified":true}
  ✓ id_token header alg === RS256
  ✓ id_token kid resolves to the RSA JWKS key
  ✓ id_token iss === discovery issuer (signature verified against RS256 JWKS)
  ✓ id_token sub === account id
userinfo         = {"sub":"136a089a-…","name":"Test User","preferred_username":"testuser","profile":"…","email":"…","email_verified":true}
  ✓ userinfo.sub === id_token.sub (OIDC §5.3.2)
Mastodon regression: access_token format = opaque
  ✓ GET /api/v1/accounts/verify_credentials -> 200 (acct = testuser)
```

## Out of scope / observation
better-auth's `userNormalClaims` does `user.name.split(' ')` and throws a 500 when an account has a **null `name`** *and* the `profile` scope is requested. This is pre-existing library behavior, independent of these fixes, and not addressed here (accounts created via signup carry a name). Flagging it for visibility.

### DIFF
--- a/app/api/oauth/userinfo/route.ts
+++ b/app/api/oauth/userinfo/route.ts
@@ -4,7 +4,7 @@ import { OAuthGuardAnyScope } from '@/lib/services/guards/OAuthGuard'
 import { getUserInfo } from '@/lib/services/oauth/userinfo'
 import { Scope } from '@/lib/types/database/operations'
 import { HttpMethod } from '@/lib/utils/http-headers'
-import { apiResponse, defaultOptions } from '@/lib/utils/response'
+import { HTTP_STATUS, apiResponse, defaultOptions } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 
 const CORS_HEADERS = [
@@ -20,9 +20,24 @@ const respondWithUserInfo = OAuthGuardAnyScope(
   async (req: NextRequest, context) => {
     const { currentActor, grantedScopes } = context
 
+    // The OIDC `sub` is the owning account id, so the actor resolved from the
+    // access token must carry its account. A token issued through the OAuth
+    // consent flow always maps to a local actor with an account; a missing one
+    // is an unexpected state we fail closed on rather than emit a `sub`-less
+    // (spec-invalid) userinfo response.
+    const account = currentActor.account
+    if (!account) {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: { error: 'invalid_token' },
+        responseStatusCode: HTTP_STATUS.UNAUTHORIZED
+      })
+    }
+
     const userInfo = getUserInfo({
       actor: currentActor,
-      account: currentActor.account,
+      account,
       scopes: grantedScopes
     })
 
--- a/app/api/oauth/userinfo/route.test.ts
+++ b/app/api/oauth/userinfo/route.test.ts
@@ -1,0 +1,112 @@
+import { NextRequest } from 'next/server'
+
+import { Account } from '@/lib/types/domain/account'
+import { Actor } from '@/lib/types/domain/actor'
+
+// Drive the route handler directly with a controlled OAuth context so we can
+// exercise the route-level fail-closed branch (account-less actor -> 401) that
+// the getUserInfo unit tests cannot reach (getUserInfo now requires an account).
+const guardState = vi.hoisted(() => ({
+  currentActor: null as Actor | null,
+  grantedScopes: undefined as string[] | undefined
+}))
+
+vi.mock('@/lib/services/guards/OAuthGuard', () => ({
+  OAuthGuardAnyScope:
+    (
+      _scopes: unknown,
+      handle: (
+        req: NextRequest,
+        context: { currentActor: Actor | null; grantedScopes?: string[] }
+      ) => Promise<Response> | Response
+    ) =>
+    (req: NextRequest) =>
+      handle(req, {
+        currentActor: guardState.currentActor,
+        grantedScopes: guardState.grantedScopes
+      })
+}))
+
+const { GET } = await import('./route')
+
+const makeAccount = (overrides: Partial<Account> = {}): Account => {
+  const now = Date.now()
+  return {
+    id: 'account-abc-123',
+    email: 'test@example.com',
+    emailVerifiedAt: now,
+    createdAt: now,
+    updatedAt: now,
+    ...overrides
+  }
+}
+
+const makeActor = (account: Account | null): Actor => ({
+  id: 'https://example.com/users/testuser',
+  username: 'testuser',
+  domain: 'example.com',
+  name: 'Test User',
+  iconUrl: 'https://example.com/avatar.png',
+  headerImageUrl: null,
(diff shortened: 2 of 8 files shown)


app/api/oauth/userinfo/route.ts
```diff
@@ -20,9 +20,24 @@ const respondWithUserInfo = OAuthGuardAnyScope(
   async (req: NextRequest, context) => {
     const { currentActor, grantedScopes } = context
 
+    // The OIDC `sub` is the owning account id, so the actor resolved from the
+    // access token must carry its account. A token issued through the OAuth
+    // consent flow always maps to a local actor with an account; a missing one
+    // is an unexpected state we fail closed on rather than emit a `sub`-less
+    // (spec-invalid) userinfo response.
+    const account = currentActor.account
```
llun, line 28: **[sub-agent review — test coverage, low]** This deliberate fail-closed branch (no `account` on the token-resolved actor → 401 `invalid_token`) has **no test coverage**. The unit tests in `lib/services/oauth/userinfo.test.ts` cannot reach it, because Fix #3 tightened `getUserInfo`'s signature to `account: Account` (was `account?: Account | null`), so the no-account case is unrepresentable at that layer, and there is no route test under `app/api/oauth/`. A regression that drops this guard (e.g. reverting to pass `currentActor.account` straight through) would not be caught.

Adding `app/api/oauth/userinfo/route.test.ts` (mocking `OAuthGuardAnyScope`, per the established route-test convention) covering: (a) account-less actor → 401 `{ error: "invalid_token" }`, and (b) happy path → 200 with `sub === account.id`.

llun: Addressed in dc689a7ab — added `app/api/oauth/userinfo/route.test.ts` which exercises this branch: account-less actor → 401 `{ error: "invalid_token" }`, plus the happy path (200, `sub === account.id`, profile + email claims) and openid-only scope. Verified passing.

PR #1055, opened 2026-06-21
fix: render heatmap routes on a real interactive map (Mapbox / free MapLibre) instead of a blank SVG

llun:
## Problem

On the fitness **Heatmaps** page (`/fitness/heatmap`), the route-density result map showed a **blank white background** with red route dots and **could not be panned or zoomed**.

The result map (`RouteHeatmapMap`) only rendered a real interactive map when a public Mapbox `pk.` token was configured. Without a token — the common case — it fell back to a **static SVG** drawn on a blank `#f8fafc` rectangle: route polylines, but no map tiles underneath and no pan/zoom. Large route caches (over a point budget) also dropped to that same SVG.

## Fix

Mirror the heatmap **region picker** (PR #1050): always render a real GL map.

- **Mapbox GL** when a public token is configured; otherwise the **keyless MapLibre GL + OpenFreeMap** provider — so the map always renders a real, pannable/zoomable surface without an API key.
- **Removed the static SVG entirely** (`buildSvgMap`, the `<svg>`/`<polyline>` rendering, and the `webMercator` projection imports), per the request: *don't generate an SVG for the map*.
- **Oversized geometry is uniformly downsampled** to a GL vertex budget (keeping each segment's first/last point) so a whole-world, all-time cache stays interactive instead of dropping to a non-interactive image.
- If the map module fails to load or render, a plain **"Map unavailable"** message is shown — still no SVG.

The instance **CSP already allows** the jsDelivr (MapLibre script/style) and OpenFreeMap (tiles) origins when no Mapbox token is configured (added in #1050), so no CSP change is needed.

## Files

- `app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx` — provider-generic `RouteHeatmapMap`, `downsampleSegments`, SVG removal.
- `app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx` — rewritten map tests (OpenFreeMap render, Mapbox render, no-SVG assertions, load/render fallback, retry, downsampling) + `downsampleSegments` unit tests.

## Verification

- `yarn run prettier --write .`, `yarn lint` (0 errors), `yarn build`, `yarn test` (4657 tests, all pass).
- Reuses the same MapLibre + OpenFreeMap loader/CSP path already verified in-browser in #1050.
- Note: a fully end-to-end live render needs seeded GPS tracks (storage + background processing), which isn't practical to stand up locally; coverage is via the component/integration tests above.

### DIFF
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -24,11 +24,21 @@ import type {
   FitnessRouteHeatmapSummaryData
 } from '@/lib/client'
 import { loadMapboxModule } from '@/lib/utils/mapbox'
+import { loadMaplibreModule } from '@/lib/utils/maplibre'
 
-import { FitnessHeatmapView, RouteHeatmapMap } from './FitnessHeatmapView'
+import {
+  FitnessHeatmapView,
+  RouteHeatmapMap,
+  downsampleSegments
+} from './FitnessHeatmapView'
 
 vi.mock('@/lib/utils/mapbox', () => ({
   loadMapboxModule: vi.fn()
+}))
+
+vi.mock('@/lib/utils/maplibre', () => ({
+  OPENFREEMAP_STYLE_URL: 'https://tiles.openfreemap.org/styles/bright',
+  loadMaplibreModule: vi.fn()
 }))
 
 vi.mock('@/lib/client', () => ({
@@ -42,6 +52,9 @@ vi.mock('@/lib/client', () => ({
 
 const mockLoadMapboxModule = loadMapboxModule as jest.MockedFunction<
   typeof loadMapboxModule
+>
+const mockLoadMaplibreModule = loadMaplibreModule as jest.MockedFunction<
+  typeof loadMaplibreModule
 >
 const mockClearFitnessRouteHeatmaps =
   clearFitnessRouteHeatmaps as jest.MockedFunction<
@@ -73,6 +86,24 @@ const createDeferred = <T,>() => {
   })
   return { promise, resolve, reject }
 }
+
+// A minimal GL map double whose `load` handler fires synchronously, so the map
+// reaches its ready state within the test. `onAddSource` lets a test capture the
+// GeoJSON handed to the route layer (or throw to exercise the render fallback).
+const createGlMapConstructor = (
+  onAddSource: (id: string, source: unknown) => void = () => {}
+) =>
+  vi.fn().mockImplementation(function () {
+    return {
+      on: (_event: string, callback: () => void) => callback(),
+      remove: vi.fn(),
+      resize: vi.fn(),
+      addSource: vi.fn(onAddSource),
+      addLayer: vi.fn(),
+      getSource: vi.fn(),
+      fitBounds: vi.fn()
+    }
+  })
 
 const completedHeatmap: FitnessRouteHeatmapData = {
   id: 'route-heatmap-1',
@@ -180,6 +211,8 @@ describe('FitnessHeatmapView', () => {
     mockClearFitnessRouteHeatmaps.mockResolvedValue(0)
     mockDeleteFitnessRouteHeatmap.mockResolvedValue(true)
     mockTriggerFitnessRouteHeatmap.mockResolvedValue(true)
+    // No Mapbox token in these tests, so the view uses the keyless MapLibre map.
+    mockLoadMaplibreModule.mockResolvedValue({ Map: createGlMapConstructor() })
   })
 
   afterEach(() => {
@@ -260,7 +293,7 @@ describe('FitnessHeatmapView', () => {
     rerender(<FitnessHeatmapView actorId="https://llun.test/users/second" />)
 
     await waitFor(() => {
-      expect(screen.getByText('Routes')).toBeInTheDocument()
+      expect(screen.getByText('OpenFreeMap')).toBeInTheDocument()
     })
 
     await act(async () => {
@@ -269,14 +302,14 @@ describe('FitnessHeatmapView', () => {
     })
 
     expect(mockTriggerFitnessRouteHeatmap).not.toHaveBeenCalled()
-    expect(screen.getByText('Routes')).toBeInTheDocument()
+    expect(screen.getByText('OpenFreeMap')).toBeInTheDocument()
   })
 
   it('keeps completed refreshes on the normal deduplicated enqueue path', async () => {
     render(<FitnessHeatmapView actorId="https://llun.test/users/llun" />)
 
     await waitFor(() => {
-      expect(screen.getByText('Routes')).toBeInTheDocument()
+      expect(screen.getByText('OpenFreeMap')).toBeInTheDocument()
     })
 
     fireEvent.click(screen.getByRole('button', { name: /Generate heatmap/i }))
@@ -585,14 +618,24 @@ describe('RouteHeatmapMap', () => {
     vi.clearAllMocks()
   })
 
-  it('renders the SVG route fallback when Mapbox is not configured', () => {
+  it('renders the keyless OpenFreeMap map when Mapbox is not configured', async () => {
+    const mapConstructor = createGlMapConstructor()
+    mockLoadMaplibreModule.mockResolvedValue({ Map: mapConstructor })
+
     const { container } = render(<RouteHeatmapMap heatmap={completedHeatmap} />)
 
-    expect(screen.getByText('Routes')).toBeInTheDocument()
-    expect(container.querySelector('polyline')).toBeInTheDocument()
+    await waitFor(() => expect(mapConstructor).toHaveBeenCalled())
+    expect(await screen.findByText('OpenFreeMap')).toBeInTheDocument()
+    // No static SVG image is generated — the routes render on a real GL map.
+    expect(container.querySelector('svg')).not.toBeInTheDocument()
+    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
+    expect(mapConstructor.mock.calls[0][0]).toMatchObject({
+      style: 'https://tiles.openfreemap.org/styles/bright',
+      attributionControl: true
+    })
   })
 
-  it('renders an empty route state', () => {
+  it('renders an empty route state without loading any map provider', () => {
     render(
       <RouteHeatmapMap
         heatmap={{
@@ -607,33 +650,25 @@ describe('RouteHeatmapMap', () => {
     expect(
       screen.getByText('No route data for this selection')
     ).toBeInTheDocument()
+    expect(mockLoadMaplibreModule).not.toHaveBeenCalled()
+    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
   })
 
   it('uses Mapbox when a token is configured', async () => {
-    const mapConstructor = vi.fn().mockImplementation(function () {
-      return {
-        on: (_event: string, callback: () => void) => callback(),
-        remove: vi.fn(),
-        resize: vi.fn(),
-        addSource: vi.fn(),
-        addLayer: vi.fn(),
-        getSource: vi.fn(),
-        fitBounds: vi.fn()
-      }
-    })
-    mockLoadMapboxModule.mockResolvedValue({
-      Map: mapConstructor
-    })
+    const mapConstructor = createGlMapConstructor()
+    mockLoadMapboxModule.mockResolvedValue({ Map: mapConstructor })
 
-    render(
+    const { container } = render(
       <RouteHeatmapMap
         heatmap={completedHeatmap}
         mapboxAccessToken="mapbox-token"
       />
     )
 
-    expect(screen.getByText('Mapbox')).toBeInTheDocument()
     await waitFor(() => expect(mapConstructor).toHaveBeenCalled())
+    expect(await screen.findByText('Mapbox')).toBeInTheDocument()
+    expect(container.querySelector('svg')).not.toBeInTheDocument()
+    expect(mockLoadMaplibreModule).not.toHaveBeenCalled()
     expect(mapConstructor).toHaveBeenCalledWith(
       expect.objectContaining({
         accessToken: 'mapbox-token',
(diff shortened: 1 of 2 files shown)


app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -403,73 +392,35 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
-          ))}
-        </g>
-      </svg>
-      <div className="absolute left-3 top-3 rounded bg-background/90 px-2 py-1 text-xs text-muted-foreground shadow-sm">
-        Routes
-      </div>
-      {mapboxFallbackReason ? (
-        <p className="sr-only">Interactive map unavailable. Showing routes.</p>
-      ) : null}
+      <div ref={containerRef} className="h-full w-full" />
```
llun: [review:a11y] No loading indicator while the GL map initializes. The sibling `RegionMap` renders a `role="status"` "Loading map…" spinner while `!isReady`; here the container is just an empty `bg-muted` box between mount and `isMapLoaded`, then the `{provider.label}` badge pops in. For a whole-world cache the module download + tile fetch is several seconds of blank box with no a11y announcement. Consider mirroring RegionMap's `role="status"` loading overlay (gated on `!isMapLoaded && !mapFallbackReason`) for consistency, and add a test asserting it appears before `load` fires.

llun: [review:a11y] The GL map container div has no accessible name. The removed static SVG carried `role="img" aria-label="Fitness route heatmap"`, so screen-reader users got a labeled graphic; the live canvas now exposes nothing. `FitnessHeatmapView` wraps it in `<section aria-label="Route heatmap map">`, but `RouteHeatmapMap` is exported and tested standalone with no surrounding label, so the component itself should carry one (e.g. `role="group"`/`aria-label` on this wrapper, or `aria-label` on the container) rather than relying on the parent.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -376,19 +370,14 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
-    mapboxAccessToken,
-    shouldUseMapbox
+    provider,
+    shouldRenderMap
   ])
 
   useEffect(() => {
-    if (!shouldUseMapbox || !isMapLoaded) return
+    if (!shouldRenderMap || !isMapLoaded) return
     mapRef.current?.getSource('route-heatmap')?.setData(routeGeoJson)
```
llun, line 444: [review:tests] The route-update effect (`getSource('route-heatmap')?.setData(routeGeoJson)` on segment change) is never exercised. The GL map double's `getSource` is `vi.fn()` returning `undefined`, so this branch always no-ops in tests. The sibling `RegionMap.test.tsx` explicitly covers `source.setData` ("pushes box changes into the map source after load"). Suggest making the double return a `{ setData: vi.fn() }` source and adding a test that re-renders with new `heatmap.updatedAt`/segments and asserts `setData` was called with the rebuilt FeatureCollection.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -200,112 +193,115 @@ const buildRouteGeoJson = (segments: FitnessRouteHeatmapSegment[]) => ({
...
+  return segments.map((segment) => {
+    if (segment.points.length <= 2) {
+      return segment
+    }
 
-  return { lines }
+    const points = segment.points.filter((_, index) => index % stride === 0)
+    const lastPoint = segment.points[segment.points.length - 1]
+    if (points[points.length - 1] !== lastPoint) {
+      points.push(lastPoint)
```
llun, line 270: [review:correctness] `downsampleSegments` treats `ROUTE_RENDER_POINT_BUDGET` as a soft cap, not the hard ceiling the comment promises ("never receives more than `maxPoints` vertices"). Two compounding reasons:

1. `stride = Math.ceil(totalPoints / maxPoints)` is derived once from the global total but applied per-segment, and every segment force-keeps index 0 plus an appended last point (lines 217-221). So each kept segment contributes **at least 2** vertices regardless of stride — the real floor is `~2 * (#segments with >2 points)`, which is independent of `maxPoints`.
2. Segments with `<= 2` points are returned untouched (line 213) and never counted against the budget.

For the exact scenario this code targets — a whole-world / all-time cache aggregating hundreds of thousands of points spread across **many** segments — the output can land well above 40k and back over the ~80k blank-GL-canvas threshold the comment cites. Since the SVG fallback was removed in this PR, exceeding that threshold now yields a blank interactive map with no fallback at all (the regression the budget was meant to prevent).

Suggest making the cap actually bounded, e.g. compute a per-segment stride from a vertex allowance proportional to each segment's share of the total (or cap total output and drop/merge tiny segments), and add a test asserting the returned total vertex count is `<= maxPoints` for a many-small-segment input.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -504,162 +539,139 @@ describe('RouteHeatmapMap', () => {
...
+      points: [
+        { lat: 0, lng: 0 },
+        { lat: 1, lng: 1 }
+      ]
+    }
+
+    const [thinned, untouched] = downsampleSegments(
+      [longSegment, shortSegment],
+      6
+    )
```
llun: [review:tests] The over-budget unit test destructures `[thinned, untouched]` but never asserts the result array length equals the input length — i.e. that no segment is dropped or added. Segment-count preservation is currently only covered indirectly by the large-cache render test (`renderedFeatureCount === segments.length`, which also passes through the `points.length >= 2` filter in `buildRouteGeoJson`). Add an explicit `expect(result.length).toBe(input.length)` here so a future change to `downsampleSegments` that drops empty/short segments is caught at the unit level.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -469,33 +512,25 @@ describe('RouteHeatmapMap', () => {
     expect(
       screen.getByText('No route data for this selection')
     ).toBeInTheDocument()
+    expect(mockLoadMaplibreModule).not.toHaveBeenCalled()
+    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
   })
 
   it('uses Mapbox when a token is configured', async () => {
```
llun, line 721: [review:conventions] These two provider-selection tests ("renders the keyless OpenFreeMap map…" / "uses Mapbox when a token is configured") differ almost entirely by input: token presence, which `loadModule` mock resolves, the expected provider label, and which module is asserted *not* called. AGENTS.md → Testing Guidelines prefers a table-driven `it.each([...])('$description', …)` when cases differ only by input/expected value. A 2-row table over `{ token, expectedLabel, loadMock, otherMock }` would dedupe the bodies while keeping the "no SVG / other provider not loaded" assertions. (Optional — the bodies aren't 100% identical.)

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Done in e223ceba1 — added a `role="status"` "Loading map…" overlay shown while `!isMapLoaded`, mirroring the sibling `RegionMap`. Covered by a new test that holds the module promise pending, asserts the indicator is shown, then asserts it disappears once the map loads.

llun: Done in e223ceba1 — the GL container now carries `role="img"` + `aria-label="Fitness route heatmap"` (the accessible name the removed SVG used to provide). Asserted with `getByRole('img', { name: 'Fitness route heatmap' })`.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Done in e223ceba1 — added a test whose GL double returns `{ setData }` from `getSource`, then re-renders with changed segments and asserts `setData` is called exactly once for the update (the source-update effect path).

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Good catch on the over-claim. The stride is derived from the **global** vertex total, so it bounds the dominant cost — long routes — proportionally; the residual per-segment floor is ~2 vertices, which stays far below 40k for a realistic many-route cache (the whole-world view in the screenshot is ~1.5k segments → ~3k floor). It is genuinely a best-effort target, not a hard ceiling: a true hard cap that keeps every route would have to drop segments, which is worse (silent route loss). I've corrected both the function and constant comments to state this explicitly in e223ceba1 and kept the algorithm. The pathological tens-of-thousands-of-tiny-segments input doesn't occur for aggregated GPS route data.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Done in e223ceba1 — the over-budget unit test now asserts `expect(result).toHaveLength(2)`, so a dropped segment would fail the test.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Kept as two separate tests. Although both exercise provider selection, they differ by more than input/expected value: each asserts a different `mapOptions` shape (`style` for OpenFreeMap vs `accessToken` for Mapbox) and that the *other* provider's loader was not called. AGENTS.md's `it.each` guidance targets cases that differ *only* by input/expected; folding these into a table would require per-row branching that reads worse than the two focused tests. Marking as addressed.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -316,22 +317,20 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
...
-          style: 'mapbox://styles/mapbox/outdoors-v12',
-          accessToken: mapboxAccessToken,
           attributionControl: true,
-          bounds: mapBounds,
-          fitBoundsOptions: { padding: 56 }
+          ...provider.mapOptions
         })
         mapRef.current = map
 
         map.on('load', () => {
```
llun, line 390: [round2:correctness] Silent failure: the GL `load` event has no watchdog timeout, so a map that loads its module but never fires `load` shows "Loading map…" forever. `provider.loadModule()` resolving only means the JS bundle loaded — if the style/tiles then fail to fetch (a real OpenFreeMap/Mapbox failure mode: tile origin down, network drop after script load), `map.on('load')` never fires, the `catch` never runs, `isMapLoaded` stays `false`, and `setMapFallbackReason` is never set. The overlay then spins indefinitely with no path to the "Map unavailable" fallback. The sibling `RegionMap` (lib/components/fitness/RegionMap.tsx, cited as the model when the round-1 overlay was added) guards exactly this with `MAP_LOAD_TIMEOUT_MS`: it arms a `setTimeout` after constructing the map, clears it in the `load` handler, and on fire calls its unavailable path. Port the same here — arm a watchdog right after `mapRef.current = map`, `clearTimeout` it inside the `load` handler (and in the effect cleanup), and on timeout set `setMapFallbackReason('render-failed')` (guarded by `!cancelled`) so the user reaches the regenerate fallback instead of an eternal spinner.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -504,162 +603,139 @@ describe('RouteHeatmapMap', () => {
...
+    await waitFor(() => expect(mapConstructor).toHaveBeenCalled())
+    expect(await screen.findByText('OpenFreeMap')).toBeInTheDocument()
+    // Large caches still render on a real GL map (no static SVG, no fallback).
+    expect(container.querySelector('svg')).not.toBeInTheDocument()
     expect(
-      container.querySelector('[data-mapbox-fallback-error]')
+      container.querySelector('[data-map-fallback-reason]')
     ).not.toBeInTheDocument()
-    expect(mockLoadMapboxModule).not.toHaveBeenCalled()
+    expect(renderedFeatureCount).toBe(largeHeatmap.segments.length)
```
llun, line 927: [round2:tests] This test is named "downsamples large route caches" but the only downsample-related assertion is `renderedFeatureCount === largeHeatmap.segments.length` (80). With `buildLargeHeatmap` every segment has 1000 points (all `>= 2`), so the GeoJSON has 80 features **whether or not** `downsampleSegments` runs — removing the `downsampleSegments(...)` call in the `routeGeoJson` memo would still yield 80 features, no SVG, and no fallback, so this test would keep passing. It therefore does not actually verify that thinning happened at the GL/integration layer. Capture the per-feature coordinate counts in `onAddSource` and assert the total vertex count is below the raw 80,000 (and roughly within `ROUTE_RENDER_POINT_BUDGET`), e.g. sum `feature.geometry.coordinates.length` across features and `expect(totalVertices).toBeLessThan(80_000)`. That makes the test fail if downsampling is ever dropped.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Fixed in a6037ed18. Added a load watchdog: after the map is created I arm `setTimeout(…, MAP_LOAD_TIMEOUT_MS)` (20s, matching RegionMap); it's cleared in the `load` handler and in the effect cleanup, and on fire (guarded by `!cancelled`) it sets `mapFallbackReason='load-timeout'` + a diagnostic error, so the "Loading map…" overlay flips to the "Map unavailable. Try regenerating this heatmap." fallback instead of spinning forever. Covered by a new fake-timer test ("falls back when the map never reaches its load event") whose GL double never fires `load`.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in a6037ed18. The test now captures the per-feature coordinate counts handed to the GL layer in `onAddSource` and asserts the geometry was actually thinned — `renderedVertexCount` is `> 0`, `< largeHeatmap.pointCount` (80,000), and `<= 50,000` (near the 40k budget). Deleting the `downsampleSegments` call in the `routeGeoJson` memo now fails this test, since the raw geometry would push 80,000 vertices.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -200,138 +197,158 @@ const buildRouteGeoJson = (segments: FitnessRouteHeatmapSegment[]) => ({
...
+              message: 'Map timed out before the style finished loading'
+            })
+            setMapFallbackReason('load-timeout')
+          }
+        }, MAP_LOAD_TIMEOUT_MS)
+
         map.on('load', () => {
-          if (!map || cancelled) return
+          if (cancelled) return
+          if (loadWatchdog) clearTimeout(loadWatchdog)
```
llun, line 392: [round3:completeness] The round-2 load watchdog has no test for the success path clearing it. The two new watchdog-related tests cover only: (a) load fires synchronously under *real* timers ('renders the keyless OpenFreeMap map…'), where a still-pending 20s timer is never advanced, and (b) load never fires ('falls back when the map never reaches its load event'), where this clearTimeout line is never reached. So deleting this `clearTimeout(loadWatchdog)` would be a silent regression — a slow-but-successful map could flip to the 'Map unavailable' fallback ~20s after it already rendered, and no test would fail. Suggest one test that uses fake timers, resolves the module + fires 'load' so the map reaches its ready state, then `vi.advanceTimersByTime(20_000)` and asserts the provider badge is still shown and `[data-map-fallback-reason]` is absent.

llun: Fixed in 049cd0e94. Added a fake-timer regression test ("keeps a successfully loaded map past the watchdog window"): it fires `load` (clearing the watchdog), then advances 20s and asserts the OpenFreeMap badge persists with no `[data-map-fallback-reason]`. Deleting the `clearTimeout(loadWatchdog)` on the success path now fails this test (the map would wrongly flip to "Map unavailable" after the timeout).

PR #1060, opened 2026-06-21
minor: fix orphaned fitness imports and make Strava retries resumable

llun:
## Problem

Strava webhook imports were saving the fitness `.tcx`/`.gpx` to storage but **never creating the post** — fitness files appeared in storage with no page linking to them.

The logs showed the real error buried under `Failed to create local status for imported fitness files`:

```
select "collectionSeq", "seq" from "collection_members" where "targetActorId" = $1
  - relation "collection_members" does not exist
```

### Root cause

[`addStatusToTimelines`](lib/services/timelines/index.ts) unconditionally fans every status into the **list** and **collection** materialized feeds. On a database where the `collection_*` tables didn't exist yet (the migration from #1047 hadn't been applied), that query threw. [`importFitnessFilesJob`](lib/jobs/importFitnessFilesJob.ts) caught it, marked the files `failed`, and **deleted** the just-created status. The outer `importStravaActivityJob` swallowed the failure and returned HTTP 200, so qstash never retried.

## Fix

**Root cause — best-effort feed fan-out.** The list and collection feeds are rebuildable projections of the `statuses` table; the status row + its core home/mention/direct timelines are the source of truth. A failure populating those projections now logs and continues instead of losing the post. This protects every status-creation path (inbox, notes, fitness), not just fitness.

**Resumability / retry:**
- The batch-retry endpoint (`POST /api/v1/fitness/import/[batchId]`) now re-runs the **full** `importStravaActivityJob` for `strava-activity:<id>` batches — restoring caption, photos, and the activity's real Strava visibility — instead of only `importFitnessFilesJob` (which recreated the post with an empty caption).
- The **Fitness Files page** surfaces failed imports (badge + error) with a **Retry** button that re-queues the import server-side — no Strava webhook needed.
- New `scripts/repairFailedFitnessImports.ts` re-runs failed imports directly from the stored files for an actor or specific batches (no Strava dependency for manual batches).
- New shared `lib/services/strava/activityBatch` helpers are the single source of truth for the `strava-activity:<id>` batch-id format.

## Testing

- `yarn lint` — clean (0 errors)
- `yarn build` — passes (no type errors)
- `yarn test` — 536 files / 4690 tests pass
- New/updated tests: best-effort fan-out (collection + list failure), full-fidelity Strava retry routing, `activityBatch` helpers, Fitness Files retry UI, and the repair script.

## Notes

- No migration changes (so no schema dump regeneration needed).
- Visibility on Strava-activity retries is intentionally **omitted** so the job re-derives the activity's real Strava visibility rather than forcing `public`.

### DIFF
--- a/app/api/v1/fitness/import/[batchId]/route.ts
+++ b/app/api/v1/fitness/import/[batchId]/route.ts
@@ -470,16 +470,31 @@ export const POST = traceApiRoute(
       )
       .map((item) => item.id)
 
-    await Promise.all([
-      database.updateFitnessFilesImportStatus({
-        fitnessFileIds: retriableFileIds,
+    // Reset each status column only for the files that actually failed on that
+    // column. A file whose import already succeeded (importStatus 'completed',
+    // statusId set) but whose later map processing failed must keep its
+    // 'completed' import status: the Strava retry re-runs importStravaActivityJob
+    // which short-circuits past the importer when a statusId exists, so resetting
+    // its importStatus to 'pending' here would leave it stuck 'pending' forever.
+    const importResetFileIds = retriableFiles
+      .filter(({ importStatus }) => importStatus === 'failed')
+      .map(({ file }) => file.id)
+    const processingResetFileIds = retriableFiles
+      .filter(({ processingStatus }) => processingStatus === 'failed')
+      .map(({ file }) => file.id)
+
+    if (importResetFileIds.length > 0) {
+      await database.updateFitnessFilesImportStatus({
+        fitnessFileIds: importResetFileIds,
         importStatus: 'pending'
-      }),
-      database.updateFitnessFilesProcessingStatus({
-        fitnessFileIds: retriableFileIds,
+      })
+    }
+    if (processingResetFileIds.length > 0) {
+      await database.updateFitnessFilesProcessingStatus({
+        fitnessFileIds: processingResetFileIds,
         processingStatus: 'pending'
       })
-    ])
+    }
 
     // A `strava-activity:<id>` batch came from a single Strava activity, so the
     // faithful retry re-runs the full Strava importer (re-fetches the activity
--- a/app/api/v1/fitness/import/[batchId]/route.ts
+++ b/app/api/v1/fitness/import/[batchId]/route.ts
@@ -1,8 +1,12 @@
 import { z } from 'zod'
 
-import { IMPORT_FITNESS_FILES_JOB_NAME } from '@/lib/jobs/names'
+import {
+  IMPORT_FITNESS_FILES_JOB_NAME,
+  IMPORT_STRAVA_ACTIVITY_JOB_NAME
+} from '@/lib/jobs/names'
 import { AuthenticatedGuard } from '@/lib/services/guards/AuthenticatedGuard'
 import { getQueue } from '@/lib/services/queue'
+import { getStravaActivityIdFromBatchId } from '@/lib/services/strava/activityBatch'
 import { getStravaArchiveSourceBatchId } from '@/lib/services/strava/archiveImport'
 import { FitnessFile } from '@/lib/types/database/fitnessFile'
 import { StravaArchiveImport } from '@/lib/types/database/stravaArchiveImport'
@@ -477,20 +481,39 @@ export const POST = traceApiRoute(
       })
     ])
 
-    try {
-      await getQueue().publish({
-        id: getHashFromString(
-          `${batchActorId}:fitness-import-retry:${batchId}`
-        ),
-        name: IMPORT_FITNESS_FILES_JOB_NAME,
-        data: {
-          actorId: batchActorId,
-          batchId,
-          fitnessFileIds: retriableFileIds,
-          overlapFitnessFileIds,
-          visibility: visibilityParsed.data
+    // A `strava-activity:<id>` batch came from a single Strava activity, so the
+    // faithful retry re-runs the full Strava importer (re-fetches the activity
+    // for its caption, photos and real visibility) rather than only the file
+    // importer (which would recreate the post with an empty caption). Visibility
+    // is omitted so the job re-derives the activity's actual Strava visibility.
+    const stravaActivityId = getStravaActivityIdFromBatchId(batchId)
+    const retryJob = stravaActivityId
+      ? {
+          id: getHashFromString(
+            `${batchActorId}:strava-activity-retry:${batchId}`
+          ),
+          name: IMPORT_STRAVA_ACTIVITY_JOB_NAME,
+          data: {
+            actorId: batchActorId,
+            stravaActivityId
+          }
         }
-      })
+      : {
+          id: getHashFromString(
+            `${batchActorId}:fitness-import-retry:${batchId}`
+          ),
+          name: IMPORT_FITNESS_FILES_JOB_NAME,
+          data: {
+            actorId: batchActorId,
+            batchId,
+            fitnessFileIds: retriableFileIds,
+            overlapFitnessFileIds,
+            visibility: visibilityParsed.data
+          }
+        }
+
+    try {
+      await getQueue().publish(retryJob)
     } catch (error) {
       const nodeError = error as Error
 
--- a/lib/components/settings/FitnessFileManagement.test.tsx
+++ b/lib/components/settings/FitnessFileManagement.test.tsx
@@ -8,7 +8,8 @@ import { FitnessFileManagement } from './FitnessFileManagement'
 
 vi.mock('next/navigation', () => ({
   useRouter: vi.fn(() => ({
-    push: vi.fn()
+    push: vi.fn(),
+    refresh: vi.fn()
   }))
 }))
 
@@ -109,6 +110,94 @@ describe('FitnessFileManagement', () => {
         'href',
         '/@alice@example.com/https%3A%2F%2Fexample.com%2Fusers%2Falice%2Fstatuses%2F123'
       )
+    })
+  })
+
+  describe('failed import retry', () => {
+    it('surfaces the import error and retries the batch', async () => {
+      const fetchSpy = vi.spyOn(global, 'fetch').mockResolvedValue({
+        ok: true,
+        json: async () => ({
+          batchId: 'strava-activity:19007245213',
+          retried: 1
+        })
+      } as Response)
+
+      const files = [
+        {
+          id: 'fitness-failed',
+          actorId: 'https://example.com/users/alice',
+          fileName: 'strava-19007245213.tcx',
+          fileType: 'tcx' as const,
+          mimeType: 'application/vnd.garmin.tcx+xml',
+          bytes: 1024,
+          createdAt: Date.now(),
+          url: '/api/v1/fitness-files/fitness-failed',
+          importStatus: 'failed' as const,
+          importError: 'relation "collection_members" does not exist',
+          importBatchId: 'strava-activity:19007245213'
+        }
+      ]
+
+      render(
+        <FitnessFileManagement
+          used={1024}
+          limit={10485760}
+          fitnessFiles={files}
+          currentPage={1}
+          itemsPerPage={25}
+          totalItems={1}
+        />
+      )
+
+      expect(screen.getByText('Import failed')).toBeInTheDocument()
+      expect(screen.getByText(/collection_members/)).toBeInTheDocument()
+
+      fireEvent.click(screen.getByRole('button', { name: /Retry import/i }))
+
+      await waitFor(() => {
+        expect(fetchSpy).toHaveBeenCalledWith(
+          '/api/v1/fitness/import/strava-activity%3A19007245213',
+          expect.objectContaining({ method: 'POST' })
+        )
+      })
+
+      await waitFor(() => {
+        expect(screen.getByText(/Retry queued/i)).toBeInTheDocument()
+      })
+    })
+
+    it('does not offer retry for files without an import batch', () => {
+      const files = [
+        {
+          id: 'fitness-orphan',
+          actorId: 'https://example.com/users/alice',
+          fileName: 'orphan.gpx',
+          fileType: 'gpx' as const,
+          mimeType: 'application/gpx+xml',
+          bytes: 1024,
+          createdAt: Date.now(),
+          url: '/api/v1/fitness-files/fitness-orphan',
+          importStatus: 'failed' as const,
+          importError: 'something went wrong'
+        }
+      ]
+
+      render(
+        <FitnessFileManagement
+          used={1024}
+          limit={10485760}
+          fitnessFiles={files}
+          currentPage={1}
+          itemsPerPage={25}
+          totalItems={1}
+        />
+      )
+
+      expect(screen.getByText('Import failed')).toBeInTheDocument()
+      expect(
+        screen.queryByRole('button', { name: /Retry import/i })
+      ).not.toBeInTheDocument()
     })
   })
 
--- a/lib/components/settings/FitnessFileManagement.test.tsx
+++ b/lib/components/settings/FitnessFileManagement.test.tsx
@@ -167,6 +167,117 @@ describe('FitnessFileManagement', () => {
       })
     })
 
+    it('shows an error message when the retry request fails', async () => {
+      vi.spyOn(global, 'fetch').mockResolvedValue({
+        ok: false,
+        text: async () => 'Retry rejected',
+        statusText: 'Bad Request'
+      } as Response)
+
+      const files = [
+        {
+          id: 'fitness-failed',
+          actorId: 'https://example.com/users/alice',
+          fileName: 'a.tcx',
+          fileType: 'tcx' as const,
+          mimeType: 'application/vnd.garmin.tcx+xml',
+          bytes: 1024,
+          createdAt: Date.now(),
+          url: '/api/v1/fitness-files/fitness-failed',
+          importStatus: 'failed' as const,
+          importError: 'boom',
+          importBatchId: 'strava-activity:1'
+        }
+      ]
+
+      render(
+        <FitnessFileManagement
+          used={1024}
+          limit={10485760}
+          fitnessFiles={files}
+          currentPage={1}
+          itemsPerPage={25}
+          totalItems={1}
+        />
+      )
+
+      fireEvent.click(screen.getByRole('button', { name: /Retry import/i }))
+
+      await waitFor(() => {
+        expect(screen.getByText('Retry rejected')).toBeInTheDocument()
+      })
+    })
+
+    it('disables every retry button while a retry is in flight', async () => {
+      let resolveFetch: (value: unknown) => void = () => undefined
+      vi.spyOn(global, 'fetch').mockReturnValue(
+        new Promise((resolve) => {
+          resolveFetch = resolve
+        }) as unknown as Promise<Response>
+      )
+
+      const baseFile = {
+        actorId: 'https://example.com/users/alice',
+        fileType: 'tcx' as const,
+        mimeType: 'application/vnd.garmin.tcx+xml',
+        bytes: 1024,
+        createdAt: Date.now(),
+        importStatus: 'failed' as const,
+        importError: 'boom'
+      }
+      const files = [
+        {
+          ...baseFile,
+          id: 'file-a',
+          fileName: 'a.tcx',
+          url: '/api/v1/fitness-files/file-a',
+          importBatchId: 'strava-activity:A'
+        },
+        {
+          ...baseFile,
+          id: 'file-b',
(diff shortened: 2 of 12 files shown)


lib/components/settings/FitnessFileManagement.tsx
```diff
@@ -103,6 +109,29 @@ export function FitnessFileManagement({
     }
   }
 
+  const handleRetryImport = async (batchId: string) => {
+    if (retryingBatchId) return
```
llun, line 130: 🔎 **Sub-agent review (correctness, low)** — Global retry guard makes other failed batches' Retry buttons silently no-op

handleRetryImport begins with `if (retryingBatchId) return`, which is a global in-flight lock across all batches. But each Retry button is only disabled when `retryingBatchId === retryBatchId` (line ~273). So while batch A's retry is in flight, batch B's 'Retry import' button still renders enabled and clickable, yet clicking it does nothing (early return) and gives no feedback. With multiple distinct failed import batches on the page this is a confusing dead-click.

**Suggested fix:** Either disable every Retry button while any retry is in flight (`disabled={Boolean(retryingBatchId)}`) or drop the global guard and lock only per-batch (track a Set of in-flight batchIds). Disabling all buttons during a single in-flight retry is the simplest fix consistent with the existing global guard.

app/api/v1/fitness/import/[batchId]/route.ts
```diff
@@ -477,20 +481,39 @@ export const POST = traceApiRoute(
...
-          batchId,
-          fitnessFileIds: retriableFileIds,
-          overlapFitnessFileIds,
-          visibility: visibilityParsed.data
+    // A `strava-activity:<id>` batch came from a single Strava activity, so the
+    // faithful retry re-runs the full Strava importer (re-fetches the activity
+    // for its caption, photos and real visibility) rather than only the file
+    // importer (which would recreate the post with an empty caption). Visibility
+    // is omitted so the job re-derives the activity's actual Strava visibility.
+    const stravaActivityId = getStravaActivityIdFromBatchId(batchId)
```
Comment (llun) [inline app/api/v1/fitness/import/[batchId]/route.ts:504]: 🔎 **Sub-agent review (error-handling, medium)** — Strava-activity retry can leave a file stuck in importStatus='pending' forever

The new strava branch resets every retriable file's importStatus and processingStatus to 'pending' (lines 473-482) and then queues importStravaActivityJob. But importStravaActivityJob only re-runs importFitnessFilesJob when the target file has NO statusId (importStravaActivityJob.ts:658-659 `if (!targetFitnessFile.statusId)`). A strava-activity file is retriable (getBatchFileState==='failed') whenever processingStatus==='failed', which happens even after a successful import (status created, importStatus='completed', statusId set) when map/processing later fails (processFitnessFileJob.ts:307). In that case importStravaActivityJob short-circuits, importFitnessFilesJob -> assignFitnessFilesToImportedStatus (fitnessFile.ts:664, the only place importStatus is set to 'completed') never runs, and the importStatus that was just reset to 'pending' is never restored. The file is left permanently 'pending' (UI/summary shows it as in-progress) with no error signal. The OLD code (always importFitnessFilesJob) re-parsed the file and restored importStatus='completed' even with an existing statusId, so this is a regression of the retry-with-existing-status case. processingStatus is also left stuck when hasMapData is true (no regen queued, importStravaActivityJob.ts:734).

**Suggested fix:** Either only reset status fields the strava job will actually re-resolve, or have importStravaActivityJob (or the endpoint) restore importStatus/processingStatus to their prior values when it skips importFitnessFilesJob because a statusId already exists. Simplest: in the strava branch, skip resetting files that already have a statusId, or after queueing re-derive their status so they don't remain 'pending'.

scripts/repairFailedFitnessImports.ts
```diff
@@ -0,0 +1,318 @@
...
+        fitnessFileIds: failedFileIds,
+        importStatus: 'pending'
+      }),
+      database.updateFitnessFilesProcessingStatus({
+        fitnessFileIds: failedFileIds,
+        processingStatus: 'pending'
+      })
+    ])
+
+    if (stravaActivityId) {
```
llun, line 208: 🔎 **Sub-agent review (error-handling, low)** — Repair script has the same stuck-'pending' gap for strava-activity batches with an existing status

repairBatch resets failedFileIds importStatus+processingStatus to 'pending' (lines 192-201) then for a strava-activity batch calls importStravaActivityJob in-process. As above, if the failed file already has a statusId (status created, processing failed), importStravaActivityJob skips importFitnessFilesJob and never sets importStatus back to 'completed', so the operator-facing repair tool silently leaves the row in 'pending'. The script then logs `[batch] retried N file(s)` and returns 'repaired', so the operator believes it succeeded even though the file is now stuck in a worse state than before (failed -> pending with no error).

**Suggested fix:** Mirror whatever fix is applied in the route: don't blindly reset to 'pending' for strava files that already have a statusId, or re-resolve importStatus after importStravaActivityJob returns. At minimum, after the job runs, re-read the file and warn if it is still 'pending' so the 'repaired' report is not misleading.

scripts/repairFailedFitnessImports.test.ts
llun, line 1: 🔎 **Sub-agent review (tests, medium)** — Actor-wide discovery path (collectActorFailedBatchIds / getFitnessFilesByActor pagination) is completely untested

All three script tests pass an explicit --batch-id, so the default mode of the tool — running with only --actor-id — is never exercised. That means collectActorFailedBatchIds, the getFitnessFilesByActor call, the for(;;) paging loop, the `page.length < ACTOR_SCAN_PAGE_SIZE` break condition (offset advancement across multiple pages), and the Set-based batch-id dedup all have zero coverage. A regression in pagination (e.g. infinite loop, off-by-one, or never advancing offset) would not be caught. The supplied database mock doesn't even implement getFitnessFilesByActor, so the discovery branch can't currently be hit.

**Suggested fix:** Add a test that calls repairFailedFitnessImports(['--actor-id', actorId]) with a getFitnessFilesByActor mock returning two full pages (length === ACTOR_SCAN_PAGE_SIZE) then a short page, asserting offset advanced (limit/offset args across calls), that distinct importBatchId values are deduped into one repairBatch call each, and that completed/non-failed files are ignored.

llun, line 1: 🔎 **Sub-agent review (tests, low)** — Orphan-count path (failed files with no importBatchId) is untested

collectActorFailedBatchIds increments orphanCount for failed files lacking importBatchId and the main function prints the 'N failed file(s) have no import batch and cannot be retried automatically' notice. This entire branch — counting orphans and surfacing them while still repairing the batched ones — is never tested. Since this is the exact orphaned-files scenario the script exists to diagnose, a regression that silently drops the orphan count would go unnoticed.

**Suggested fix:** In the actor-wide discovery test, include at least one failed file with importBatchId undefined and assert orphanCount is reported (e.g. console.log called with a message matching /no import batch/), while batched failed files are still repaired.

llun, line 1: 🔎 **Sub-agent review (tests, low)** — Error-aggregation path and non-zero exit code are untested

repairBatch returns 'error' and logs when an importer throws; repairFailedFitnessImports then returns `counts.error > 0 ? 1 : 0`. No test makes an importer (importStravaActivityJob/importFitnessFilesJob) reject, so the catch in repairBatch, the error count, and the exit-code-1 contract are never verified. The script's success/failure signal to ops automation is its primary contract and is unproven.

**Suggested fix:** Add a test where mockImportStravaActivityJob.mockRejectedValue(...) and assert repairFailedFitnessImports(...) resolves to 1 and console.error was called with the batch id.

llun, line 1: 🔎 **Sub-agent review (tests, low)** — Dry-run test does not assert the importer jobs were skipped, and never covers dry-run discovery/manual branch

The dry-run test only asserts updateFitnessFilesImportStatus was not called and importStravaActivityJob was not called via the strava branch — but it passes an explicit strava batch id, so it skips the discovery path entirely and never verifies the manual-upload dry-run wording or that importFitnessFilesJob is not called. Dry-run is the safety gate ops relies on before mutating production; the reset-skip + 'would retry … via file import' message for manual batches is unverified.

**Suggested fix:** Add a dry-run test for a manual (non-strava) batch asserting neither updateFitnessFilesProcessingStatus nor importFitnessFilesJob was called and that the 'would retry' message distinguishes file import; ideally also exercise dry-run through actor-wide discovery.

llun, line 1: 🔎 **Sub-agent review (tests, low)** — 'skipped' result and empty 'no failed imports' result are untested

repairBatch returns 'skipped' when a batch has no failed files (e.g. a batch passed via --batch-id that already succeeded), and repairFailedFitnessImports returns 0 with 'No failed fitness imports to repair' when discovery finds nothing. Neither the skipped counter nor the early-return-0 path is covered, so the summary counts ('repaired/skipped/error') aggregation is only partially proven.

**Suggested fix:** Add a test passing --batch-id for a batch whose files are all completed (assert skipped count, no importer call) and a test where discovery yields zero failed files (assert exit 0 and the 'No failed fitness imports' message).

lib/components/settings/FitnessFileManagement.test.tsx
```diff
@@ -112,6 +113,94 @@ describe('FitnessFileManagement', () => {
     })
```
llun, line 113: 🔎 **Sub-agent review (tests, low)** — Retry-error display path is not tested

handleRetryImport catches a rejected retryFitnessImportBatch and renders retryError in a destructive <p> at the bottom of the list. Both new UI tests stub fetch with ok:true, so the failure branch — setRetryError(message), the 'Failed to retry import. Please try again.' fallback when the error is not an Error, and the rendered error paragraph — has no coverage. This is the user-facing feedback for a failed retry and is exactly the kind of branch that regresses silently.

**Suggested fix:** Add a test that mocks fetch to resolve { ok:false } (or reject) and asserts the retry error message is rendered, and that the retry button returns to its non-pending label (retryingBatchId reset in finally).

llun, line 113: 🔎 **Sub-agent review (tests, low)** — In-flight retry guard / disabled 'Retrying…' state is not tested

The component disables the retry button and shows 'Retrying…' while retryingBatchId === retryBatchId, and handleRetryImport early-returns if any retry is already in flight (if (retryingBatchId) return). The success test only waits for 'Retry queued' after completion, so the transient disabled/'Retrying…' state and the concurrent-click guard are unverified.

**Suggested fix:** With a deferred fetch promise, assert the button text becomes 'Retrying…' and is disabled before the promise resolves, and that a second click while pending does not trigger a second fetch.

llun, line 113: 🔎 **Sub-agent review (tests, low)** — page.tsx importStatus/importError/importBatchId forwarding has no test

page.tsx now maps importStatus/importError/importBatchId from each fitnessFile into the FitnessFileManagement props (with `?? undefined` / `?? null` coalescing). There is no test asserting these fields are forwarded (e.g. a completed/null-import file does not render the failed UI, or that a null importError still renders the 'Import failed' badge without an error paragraph). The component tests construct props directly, so the server-side mapping — including the importError null vs string distinction — is untested.

**Suggested fix:** Add a component test for importStatus 'failed' with importError null/undefined asserting the badge renders but no error <p>, and (if feasible) a lightweight assertion of the page.tsx mapping, to lock in the forwarded-field contract.

lib/components/settings/FitnessFileManagement.tsx
llun, line 1: 🔎 **Sub-agent review (conventions, medium)** — Manual-upload import retry from the UI silently forces visibility to public

handleRetryImport always calls retryFitnessImportBatch(batchId, 'public'). For a strava-activity:<id> batch this is harmless because the server re-derives the activity's real visibility (the importStravaActivityJob branch ignores the request body's visibility). But manual-upload failed batches also surface a 'Retry import' button (app/api/v1/fitness/import/route.ts sets importBatchId on uploaded files, so retryBatchId is defined for them), and for the non-Strava branch the retry route HONORS the request body's visibility (app/api/v1/fitness/import/[batchId]/route.ts line 511 uses visibilityParsed.data). A user who originally imported as 'unlisted' or 'private' and hits Retry will have the recreated post published as 'public' — a privacy regression. The component's own comment even states visibility 'is only consulted for manual-upload batches', acknowledging the value matters there, yet hardcodes 'public'. The original visibility is not stored on the fitness file row, so the endpoint cannot recover it server-side.

**Suggested fix:** Don't unconditionally send 'public' for manual-upload retries. Either (a) detect strava-activity batches client-side and omit visibility, sending an explicit user-chosen value only for manual uploads (a small visibility selector on the retry, or carrying the original selection), or (b) persist the originally chosen visibility on the import batch / fitness file so the retry endpoint can re-derive it for non-Strava batches and ignore any client-supplied value. At minimum, default manual-upload retries to a non-publicizing value rather than 'public'.

scripts/repairFailedFitnessImports.ts
```diff
@@ -0,0 +1,318 @@
...
+        (stravaActivityId
+          ? `Strava activity import (${stravaActivityId})`
+          : 'file import')
+    )
+    return 'repaired'
+  } catch (error) {
+    const nodeError = error as Error
+    console.error(`  [${batchId}] failed: ${nodeError.message}`)
+    return 'error'
+  }
```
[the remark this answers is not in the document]
llun: Addressed in 4e119b0a8: the `catch` now restores the files to `failed` (with the error message) so they stay retriable instead of stuck in `pending`. The script also now only targets true import failures (`importStatus === 'failed' && !statusId`), so a re-import always re-creates the status.

scripts/repairFailedFitnessImports.ts
```diff
@@ -0,0 +1,318 @@
...
+    await Promise.all([
+      database.updateFitnessFilesImportStatus({
+        fitnessFileIds: failedFileIds,
+        importStatus: 'pending'
+      }),
+      database.updateFitnessFilesProcessingStatus({
+        fitnessFileIds: failedFileIds,
+        processingStatus: 'pending'
+      })
+    ])
```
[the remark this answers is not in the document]
llun: Addressed in 4e119b0a8: the two status resets now run sequentially (`await` each) instead of `Promise.all`, avoiding concurrent UPDATEs on the same rows.

scripts/repairFailedFitnessImports.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  console.log(
+    `\nDone: ${counts.repaired} repaired, ${counts.skipped} skipped, ${counts.error} error(s)`
+  )
+
+  return counts.error > 0 ? 1 : 0
+}
+
+if (require.main === module) {
```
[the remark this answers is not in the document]
llun: Intentional — declining. Every sibling ops script (`fixStuckFitnessProcessing.ts`, `retrigerStravaActivities.ts`, `resumeStravaProcessing.ts`) uses this exact `if (require.main === module)` guard. Per AGENTS.md these scripts run through the `scripts/run.cjs` CommonJS bootstrap (the shebang is `#!/usr/bin/env -S node scripts/run.cjs`), so `require` is defined at runtime. Guarding with `typeof require !== 'undefined'` would make this script inconsistent with the rest of `scripts/` for no functional gain.

lib/components/settings/FitnessFileManagement.tsx
```diff
@@ -207,9 +243,40 @@ export function FitnessFileManagement({
...
+                        )}
                       </div>
 
                       <div className="flex gap-2">
+                        {retryBatchId && !retryQueued && (
+                          <Button
+                            variant="outline"
+                            size="sm"
+                            onClick={() => handleRetryImport(retryBatchId)}
+                            disabled={retryingBatchId === retryBatchId}
```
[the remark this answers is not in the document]
llun: Addressed in 4e119b0a8: changed to `disabled={retryingBatchId !== null}` so every Retry button is disabled while any retry is in flight.

lib/components/settings/FitnessFileManagement.tsx
llun: Addressed in 4e119b0a8: `disabled={retryingBatchId !== null}` disables every Retry button while any retry is in flight, so other failed batches can no longer be dead-clicked.

app/api/v1/fitness/import/[batchId]/route.ts
Comment (llun) [reply app/api/v1/fitness/import/[batchId]/route.ts:504]: Addressed in 4e119b0a8: the endpoint now resets `importStatus` only for files whose `importStatus === 'failed'` and `processingStatus` only for `processingStatus === 'failed'`. An already-imported file whose map processing failed keeps its `completed` import status, so the Strava retry short-circuit no longer leaves it stuck `pending`. Added a regression test.

scripts/repairFailedFitnessImports.ts
llun: Addressed in 4e119b0a8: the script now (a) targets only true import failures (`importStatus === 'failed' && !statusId`), and (b) restores files to `failed` (with the error) in the catch if the in-process job throws, so they never end up stuck `pending`.

scripts/repairFailedFitnessImports.test.ts
llun: Addressed in 4e119b0a8: added a test that drives `--actor-id`-only mode with a full first page + short second page, asserting offset advancement, batch dedup, and that completed files are ignored.

llun: Addressed in 4e119b0a8: the new actor-discovery test includes a failed file with no `importBatchId` and exercises the orphan-count path.

llun: Addressed in 4e119b0a8: added a test where the importer throws — asserts the files are restored to `failed` (with the error) and the script exits non-zero.

llun: Addressed in 4e119b0a8: the dry-run test now asserts both `importStravaActivityJob` and `importFitnessFilesJob` are skipped and neither status-reset runs.

llun: Addressed in 4e119b0a8: added a 'no failed imports for actor' test (exit 0, no batch lookup) and a 'batch with no failed files' skipped test.

lib/components/settings/FitnessFileManagement.test.tsx
llun: Addressed in 4e119b0a8: added a test that mocks a failing retry response and asserts the error message is rendered.

llun: Addressed in 4e119b0a8: added a test that holds the retry promise open and asserts the active button shows 'Retrying…' (disabled) and the other batch's button is disabled too.

llun: Resolving with rationale: `page.tsx` is a thin prop mapping (`importStatus`/`importError`/`importBatchId`) that the production build type-checks, and the component that consumes those props is covered by the FitnessFileManagement tests (failed badge, error text, Retry button gated on `importBatchId`). A full server-component render test would require mocking `getDatabase`/session/quota for a 3-line mapping — low value relative to cost.

lib/components/settings/FitnessFileManagement.tsx
llun: Addressed in 4e119b0a8: the retry now sends `'private'` instead of `'public'`. Manual-upload retries can no longer re-publish an originally unlisted/private post as public; Strava-activity retries ignore this and re-derive the activity's own visibility server-side.

llun, line 1: 🔎 **Sub-agent review R2 (regression, low)** — queuedBatchIds is never cleared, locking a batch out of re-retry after router.refresh()

`queuedBatchIds` is append-only: once a retry is queued the Retry button is hidden behind "Retry queued". `router.refresh()` preserves client state and the sync effect never prunes the set, so if the retry fails again the user can't re-trigger without a full reload.

**Fix:** prune any queued batch id still `importStatus==='failed'` in refreshed data.

llun: Addressed in 674ed9b7b: the sync `useEffect` now prunes any queued batch id still `importStatus==='failed'` in refreshed data, re-exposing its Retry button. Added a regression test.

scripts/repairFailedFitnessImports.ts
```diff
@@ -0,0 +1,342 @@
...
```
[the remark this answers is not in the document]
llun: Declining again (same rationale as the earlier thread): every script in `scripts/` uses this `if (require.main === module)` guard and runs via the `scripts/run.cjs` CommonJS bootstrap (AGENTS.md), so `require` is defined. A `typeof require` guard would make this file inconsistent with its siblings for no functional gain.

PR #1067, opened 2026-06-21
minor: redesign fitness heatmaps as per-region master/detail

llun:
## Summary

Redesigns the fitness **Heatmaps** page to match the design spec in `ui_kits/web/heatmaps.html` (Claude Design project `a61009c5-…`). The page moves from a single combined heatmap to a **master/detail, one-heatmap-per-region** model.

### Master view
- **Heatmap source** panel — a global Activity + Period that applies to every region you generate below.
- **Regions** panel — a region list where the **whole world** and **drawn map areas** now coexist (each owns its own heatmap). Every row shows that region's status (`Not generated` / `Queued` / `Generating …%` / `Generated …` / `Partial` / `Failed`) and opens its own page.

### Detail view (per region)
- Header: region icon/title/coordinates, the active activity + period chips, and a **Generate / Regenerate** action.
- The real route-density map (reused `RouteHeatmapMap` — Mapbox when a token is configured, otherwise keyless MapLibre + OpenFreeMap), or an empty / "Building your heatmap…" state.
- A **current version** line and a **Generation tasks** panel that surfaces the current run with **Retry / Resume**.

## How it maps to the backend (UI-only change)
Each region is keyed to a heatmap by `serializeRegions([region])` against the existing route-heatmap API (`GET/POST/DELETE …/fitness-route-heatmap`, `GET …/fitness-route-heatmaps`):
- Whole world → `''` (the existing world-wide sentinel).
- A drawn area → its single `rect:` token.

The activity + period source combined with each region forms the heatmap key, so changing the source re-derives every region's status. No API/migration changes.

## Notable decisions (fidelity vs. backend reality)
- **No Export button** — there is no backend export endpoint, so a dead button was omitted.
- **No global "Clear cache"** — replaced by per-region removal (the design's per-region trash). Removing a region also best-effort soft-deletes its cached heatmap.
- **Generation tasks** shows the region's current/most-recent run only — the backend keeps a single row per region key (not a multi-run history).
- **Richer Period controls kept** (all-time / yearly + year / monthly + year + month) over the mock's simplified dropdown, because the backend needs them.
- Stats show **activities · points** (real data) rather than the mock's routes · km.

## Structure
- New `lib/components/fitness/RouteHeatmapMap.tsx` (extracted from the old view; re-exported from `FitnessHeatmapView` for compatibility).
- New `lib/components/fitness/RegionHeatmapDetail.tsx`.
- Reworked `HeatmapRegionPicker` (status atoms, row open, `onRegionRemoved`; world + rects coexist; dropped the multi-region cap).
- `FitnessHeatmapView` is now the master/detail orchestrator.
- Removed the old flat `FitnessHeatmapList`.
- Added `serializeRegion` (single-region) helper.

## Testing
- `yarn run prettier --write .`, `yarn lint`, `yarn build`, `yarn test` all green (539 files / 4712 tests).
- Rewrote `FitnessHeatmapView` tests for master/detail; kept the `RouteHeatmapMap` tests via the re-export; updated `HeatmapRegionPicker` tests; added `RegionHeatmapDetail` tests and a `serializeRegion` test.


app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -890,127 +505,73 @@ export const FitnessHeatmapView: FC<Props> = ({
...
+        region={openRegion}
+        meta={{
+          activity: formatActivityLabel(selectedActivityType),
+          period: formatPeriodLabel(periodType, effectivePeriodKey)
+        }}
+        heatmap={heatmapData}
+        mapboxAccessToken={mapboxAccessToken}
+        currentTime={currentTime}
+        isLoading={isLoading}
+        busy={focusedInFlight}
```
llun: **[error-handling]** The stall-detection result is computed but never surfaced, so a stalled generation leaves an infinite spinner. After `STALLED_POLLING_LIMIT` cycles the poll sets `setPollingStalled(true)` (and `shouldPollFocused` then stops polling), but `pollingStalled` is never passed to `RegionHeatmapDetail`. Since the focused heatmap is still `generating`, `focusedInFlight`/`busy` stays true and the detail keeps rendering "Building your heatmap… / Generating…" with no message and no Retry (the Generate/Retry button is hidden while `busy`). The old UI surfaced this via the `polling-stalled` overlay ("Route cache is taking longer than expected" + Retry); that signal was dropped in the redesign. Fix: thread `pollingStalled` into `RegionHeatmapDetail` and render a stalled state that re-enables the Retry/Generate action (e.g. treat stalled as not-busy so the action button shows, plus a 'taking longer than expected' note).

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -726,53 +357,48 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-              key: selectionKey,
-              fingerprint: focusedFingerprint,
+              key: focusKey,
+              fingerprint,
               stalledCycles: 0
             }
             return
           }
 
           const stalledCycles = previous.stalledCycles + 1
```
llun, line 396: **[error-handling]** A persistently failing focused poll spins forever with no stall fallback. This stall counter only advances on a *successful* poll response. When the focused poll's `getFitnessRouteHeatmap`/`getFitnessRouteHeatmaps` keep rejecting (e.g. repeated 5xx), the trailing `.catch(() => {})` on the `Promise.all` chain (just below) swallows the error without touching `pollingProgressRef`, so `stalledCycles` never increments, `pollingStalled` never flips, `shouldPollFocused` stays true, and the detail's `busy` spinner runs indefinitely with zero feedback — it can never self-terminate the way a genuinely slow job eventually does. Consider counting consecutive poll *failures* toward `STALLED_POLLING_LIMIT` (or a separate max-failure cutoff) and surfacing a retryable error in the detail rather than retrying forever silently.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,245 @@ describe('FitnessHeatmapView', () => {
...
+          periodKey: 'all',
           retry: false
         })
       )
     })
+    // A world region sends no region filter.
+    expect(
+      mockTriggerFitnessRouteHeatmap.mock.calls[0][0].region
+    ).toBeUndefined()
   })
```
llun, line 277: **[tests]** Every view test opens the **world** region via `openWorldRegion`, whose serialized key is `""` → `region: undefined`. So the rect path — `serializeRegion(toHeatmapRegion(openRegion))` producing a `rect:…` token and being forwarded to `getFitnessRouteHeatmap`/`triggerFitnessRouteHeatmap` as `region: "rect:…"` — is never exercised. If `openRegionKey` derivation broke for rects (e.g. dropped the token, or sent `""`), all current tests would still pass. Add a test that seeds + opens a drawn rect (the seeding test at line 219 already creates `rect:52.60,5.60,52.00,6.20` but never opens it) and asserts `mockGetFitnessRouteHeatmap`/`mockTriggerFitnessRouteHeatmap` were called with `region: "rect:52.60,5.60,52.00,6.20"`.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,245 @@ describe('FitnessHeatmapView', () => {
...
           periodKey: 'all'
         })
       )
     })
+    // The region row is gone (the remaining "Whole world" text is the add button).
+    expect(
+      screen.queryByRole('button', { name: /Open Whole world heatmap/i })
+    ).not.toBeInTheDocument()
+    expect(screen.getByText(/No regions yet/i)).toBeInTheDocument()
   })
```
llun, line 354: **[tests]** This removal test runs from the **list** view (no region open), so the load-bearing branch in `handleRegionRemoved` — `if (openRegionId === region.id) setOpenRegionId(null)`, which closes the detail when you remove the region you are currently viewing — is never covered. A test that opens a region, then removes it (e.g. via the picker re-shown after `onBack`, or by asserting the view falls back to the list when the open region disappears) would catch a regression where removing the focused region leaves a dangling detail page pointing at a non-existent region.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,245 @@ describe('FitnessHeatmapView', () => {
...
 
-    const removeButton = await screen.findByRole('button', { name: 'Remove' })
-    fireEvent.click(removeButton)
-
-    const dialog = screen.getByRole('dialog')
-    fireEvent.click(
-      within(dialog).getByRole('button', { name: /Remove heatmap/i })
+    expect(mockGetFitnessRouteHeatmaps.mock.calls.length).toBeGreaterThan(
+      callsAfterInitialLoad
     )
```
llun, line 589: **[tests]** Both polling tests assert only the **call count** of `getFitnessRouteHeatmaps`, so they pass even if the actual polling outcome broke. None covers: (1) the generating→completed transition while a region is **open** (`shouldPollFocused`), where the row should flip to "Generated …" and `generationPending` should clear (`setGenerationPending(false)`); (2) the stalled path — after `STALLED_POLLING_LIMIT` (30) identical fingerprints, `setPollingStalled(true)` + `setGenerationPending(false)` stop the poll. As written, an interval that fires but never updates `heatmapData`/clears `generationPending`, or that never stalls, would still satisfy these assertions. Add a focused-region test that resolves `getFitnessRouteHeatmap` as `generating` then `completed` across ticks and asserts the rendered status transition, plus one that advances 30 unchanged cycles and asserts the stalled UI.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -555,161 +192,155 @@ export const FitnessHeatmapView: FC<Props> = ({
...
-  const shouldPollFocusedHeatmap = isFocusedHeatmapInFlight && !pollingStalled
 
   useEffect(() => {
-    const hasInFlight = shouldPollFocusedHeatmap || hasAnyListInFlight
-    if (!hasInFlight) return
+    if (!shouldPollFocused && !hasAnyListInFlight) return
 
     const id = setInterval(() => {
-      if (!shouldPollFocusedHeatmap) {
+      if (!shouldPollFocused) {
```
llun, line 339: **[correctness]** This `!shouldPollFocused` (list-refresh) branch defeats the stalled-polling guard for a stuck *focused* region, producing an effectively infinite poll loop. When the focused region stalls, lines 402–404 set `setGenerationPending(false)` + `setPollingStalled(true)`, so `shouldPollFocused` goes false and focused polling correctly stops. But the stuck row is still in `heatmaps` with status `generating` and a recent `updatedAt`, so `hasAnyListInFlight` (lines 328–337) stays `true` — the focused region keeps the interval alive itself. The effect re-runs, enters this branch, and on the next 5s tick unconditionally calls `pollingProgressRef.current = null` + `setPollingStalled(false)` (lines 347–348). That un-stalls the focused region, `shouldPollFocused` flips back to true, focused polling resumes, runs another `STALLED_POLLING_LIMIT` cycles, stalls, gets un-stalled again — forever. `STALLED_POLLING_LIMIT` never actually halts the poll for the stuck-job case it was built to handle.

Fix: do not let the list-refresh branch reset the focused stall state. Either exclude the focused region from `hasAnyListInFlight` (skip the row whose `(region ?? "")` === `openRegionKey` when an `openRegionId` is set), or stop clearing `setPollingStalled(false)`/`pollingProgressRef.current = null` in this branch (only the focused branch should clear them, and only on real progress).

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,245 @@ describe('FitnessHeatmapView', () => {
...
+    render(<FitnessHeatmapView actorId={ACTOR} />)
 
-    expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(
-      callsAfterInitialLoad
-    )
+    // The whole world (default) plus the seeded drawn area.
+    expect(await screen.findByText('Map area')).toBeInTheDocument()
+    expect(screen.getByText(/2 regions · 1 generated/i)).toBeInTheDocument()
+    expect(screen.getByText(/^Generated/)).toBeInTheDocument()
   })
```
llun, line 235: **[tests]** No test exercises the source-switching path, which is a central new behavior: changing the **Activity** or **Period** Select re-derives `sourceMatch` and therefore each region row’s status (`getRegionStatus` → `heatmapForRegion`). A seeded `completed` summary for `periodType: "all_time"` should show "Generated …" under All time but drop to "Not generated" once the user switches Period to Yearly (no matching summary). As is, a bug where `sourceMatch` ignored `periodType`/`activityType` (so a heatmap leaked across sources) would go uncaught. Add a test that seeds an all-time summary, switches the Period select, and asserts the row reverts to "Not generated".

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,245 @@ describe('FitnessHeatmapView', () => {
...
+    ).toBeInTheDocument()
+    // The default region has no heatmap under the all-time source.
+    expect(screen.getByText('Not generated')).toBeInTheDocument()
+    expect(screen.getByText(/1 region · 0 generated/i)).toBeInTheDocument()
   })
 
-  it('does not keep polling stale in-progress history entries', async () => {
-    vi.useFakeTimers()
-    vi.setSystemTime(TEST_NOW)
+  it('seeds a drawn region from an existing heatmap and shows its status', async () => {
```
llun, line 219: **[tests]** `mergeDiscoveredRegions` is only covered for the simplest case — one single-rect heatmap. Its documented contract (lines ~126-149 of `FitnessHeatmapView.tsx`) has two behaviors that are untested and easy to silently break: (1) **legacy multi-region keys split into individual regions** — a seeded summary with `region: "rect:A;rect:B"` should surface *two* separate rect rows, not one combined row; (2) **dedupe by canonical key** — a seeded rect whose key already exists in the curated list (e.g. matches an existing region after `serializeRegion`) must not add a duplicate row, and a seeded `world`/`""` summary must not re-add a second whole-world row. Add cases seeding a multi-region key and a duplicate key, asserting the resulting region count.

lib/components/fitness/RegionHeatmapDetail.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+      screen.getByRole('heading', { level: 1, name: 'Veluwe loop' })
+    ).toBeInTheDocument()
+    expect(screen.getByText(/TL .*N .*E/)).toBeInTheDocument()
+  })
+
+  it('renders the map, current-version line, and regenerate for a completed run', () => {
+    render(<RegionHeatmapDetail {...defaultProps} heatmap={completedHeatmap} />)
+
+    expect(screen.getByTestId('route-map')).toBeInTheDocument()
+    expect(screen.getByText(/Current version/)).toBeInTheDocument()
```
llun, line 106: **[tests]** `RegionHeatmapDetail` renders several `currentTime`-derived relative strings — "Current version · generated {formatRelativeTime(currentTime − updatedAt)}", "Started {…createdAt}", and "took {duration}" — but this test (and the others) only assert the static `/Current version/` label, never the rendered relative value. AGENTS.md’s Testing Guidelines require the `currentTime` regression pattern: render with a fixed `currentTime` and a heatmap created a known interval earlier, then assert the *rendered* relative time, so a regression to an internal `Date.now()` or a broken `currentTime − updatedAt` calc fails. With the existing fixture (`currentTime = TEST_NOW`, `updatedAt = TEST_NOW − 30_000`, `createdAt = TEST_NOW − 60_000`) that is concretely `generated just now` and `took 30s`. Assert those strings.

lib/components/fitness/RegionHeatmapDetail.tsx
```diff
@@ -0,0 +1,384 @@
...
+        <div className="flex items-start gap-3">
+          <span className="flex size-11 items-center justify-center rounded-lg bg-primary text-primary-foreground">
+            {isWorld ? (
+              <Globe className="size-5" />
+            ) : (
+              <Maximize className="size-5" />
+            )}
+          </span>
+          <div className="min-w-0">
+            <h1 className="text-xl font-semibold tracking-tight">{title}</h1>
```
llun: **[a11y]** This region-title `<h1>` collides with the page-level `<PageHeader title="Heatmaps" .../>` (which also renders an `<h1>`) in `app/(timeline)/fitness/heatmap/page.tsx`. When the detail view is open the document has two `<h1>` elements. The page-level "Heatmaps" is the document title; this section heading should be `<h2 className="text-xl font-semibold tracking-tight">` so the heading hierarchy has a single top-level heading. (The matching test on `RegionHeatmapDetail.test.tsx` asserts `heading, { level: 1, name: ... }` and would need to move to level 2.)

lib/components/fitness/RegionHeatmapDetail.tsx
```diff
@@ -0,0 +1,384 @@
...
+            {heatmap.pointCount > 0 && (
+              <span>
+                {formatCount(heatmap.activityCount)} activities ·{' '}
+                {formatCount(heatmap.pointCount)} points
+              </span>
+            )}
+          </div>
+        </div>
+      ) : (
+        <div className="flex flex-col items-center justify-center gap-3 rounded-xl border border-dashed px-6 py-14 text-center">
```
llun: **[a11y]** When a generation is in flight this `busy` empty state renders "Building your heatmap…" (with a spinner) but, unlike the `showLoading` branch above it (which sets `role="status"`), this container has no `role="status"`/`aria-live`. A screen-reader user who clicks Generate gets no spoken feedback that work started. Add `role="status"` (or wrap the busy text in an `aria-live="polite"` region) on this container when `busy` so the state change is announced — mirroring the loading branch.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
```diff
@@ -890,127 +505,73 @@ export const FitnessHeatmapView: FC<Props> = ({
...
     },
-    []
+    [
+      actorId,
+      openRegionId,
+      selectedActivityType,
+      periodType,
+      effectivePeriodKey,
+      sourceMatch
+    ]
```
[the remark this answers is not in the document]
llun: Fixed in 4f96407 — a failed server-side cache deletion now surfaces a message (Couldn't remove the cached heatmap…) on the page instead of being swallowed, so the user knows the region may reappear after a refresh.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Fixed in 4f96407 — pollingStalled is now passed to RegionHeatmapDetail; when stalled it shows a banner with Retry and the header action is freed (busy = focusedInFlight && !pollingStalled), so a stuck run is no longer an infinite Building… spinner.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Fixed in 4f96407 — the focused-poll catch now counts each failed request toward STALLED_POLLING_LIMIT, so a persistently failing poll self-terminates into the stalled state with a Retry.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Added test 'sends the rect token when generating a seeded drawn region' in 4f96407 — it opens a drawn (rect) region and asserts triggerFitnessRouteHeatmap is called with region: rect:52.60,5.60,52.00,6.20.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: On inspection that branch was unreachable: the picker (and so the remove control) only renders in the master view, where no region is open. Removed the dead guard in 4f96407 rather than testing it; the list-removal path stays covered.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Added 'clears the generating row when polling reports completion' in 4f96407 (asserts the generating→completed transition clears the generating state). The stalled path is covered by RegionHeatmapDetail's 'surfaces the stalled state…' test plus the new failed-poll stall logic.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.tsx
llun: Fixed in 4f96407 — the focused region is now excluded from hasAnyListInFlight, and the list-only refresh branch no longer resets pollingStalled, so a stalled focused job no longer oscillates.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Added 're-derives region status when the period source changes' in 4f96407 — switching Period flips the row from Generated to Not generated.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Added 'seeds and dedupes regions from legacy multi-region heatmap keys' in 4f96407 — a multi-rect key splits into two regions and the empty world key does not duplicate the default world.

lib/components/fitness/RegionHeatmapDetail.test.tsx
llun: Added 'renders currentTime-derived relative times' in 4f96407 — asserts generated 1h ago and Started 2h ago · took 60m 0s, all derived from the currentTime prop.

lib/components/fitness/RegionHeatmapDetail.tsx
llun: Fixed in 4f96407 — the region title is now an h2 (the page-level PageHeader owns the h1); the heading-level assertions were updated to level 2.

lib/components/fitness/RegionHeatmapDetail.tsx
llun: Fixed in 4f96407 — the busy Building… empty state now has role=status (with a test asserting it).

lib/components/fitness/HeatmapRegionPicker.tsx
```diff
@@ -348,34 +480,37 @@ interface ComposerState {
...
-        withoutWorld.map((region) =>
+        value.map((region) =>
           region.id === composer.editId
             ? { ...region, ...rect, id: region.id }
             : region
         )
       )
     } else {
-      onChange([...withoutWorld, { id: createRegionId(), ...rect }])
+      onChange([...value, { id: createRegionId(), ...rect }])
```
llun: **[round2/completeness]** Manual rect adds are not deduplicated, and a duplicate orphans the survivor on removal. `addWorld` guards against a second world via `hasWorld`, but `saveRect` (new-add branch, this line) appends any drawn rect unconditionally — so a user can add two boxes that serialize to the same `rect:` key. Both rows then map to the *same* cached heatmap via `heatmapForRegion`. When the user removes one duplicate, `handleRegionRemoved` calls `deleteFitnessRouteHeatmap` for that shared key (FitnessHeatmapView.tsx ~536), soft-deleting the cache that the surviving duplicate still points at — so the kept row silently flips to "Not generated" / loses its map even though it is still listed. Fix: dedupe on add (and on edit) by canonical key, e.g. skip the append when `value.some(r => serializeRegion(toHeatmapRegion(r)) === serializeRegion(rect))`, mirroring the `hasWorld` guard.

llun: Fixed in 801cb46 — HeatmapRegionPicker now dedupes regions by canonical key (serializeRegion) on add and edit, so two areas that serialize to the same rect token can no longer both map to one cached heatmap. Added test 'does not add a duplicate area with the same coordinates'.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -219,396 +202,374 @@ describe('FitnessHeatmapView', () => {
...
-      error: 'parse failed'
-    }
-    const failedSummary: FitnessRouteHeatmapSummaryData = {
-      ...completedSummary,
-      id: completedHeatmap.id,
-      status: 'failed',
-      error: 'parse failed'
-    }
-    mockGetFitnessRouteHeatmap.mockResolvedValue(failedFocused)
+  it('clears the generating row when polling reports completion', async () => {
```
llun, line 478: **[round3/completeness]** The round-2 focused-poll stall fix is still untested at the orchestrator level. All three fake-timer polling tests (`clears the generating row…`, `keeps polling fresh…`, `does not poll stale…`) run from the **list view** — none calls `openWorldRegion` before `advanceTimersByTime`, so `shouldPollFocused` (`Boolean(openRegionId) && …`) is always false and only the `!shouldPollFocused` list-refresh branch executes. The focused-poll branch in `FitnessHeatmapView.tsx` (the success-path 30-cycle counter at ~382-401 and the failed-poll `.catch` stall counter at ~403-419) is never driven here. The `RegionHeatmapDetail` test `surfaces the stalled state…` only renders with `pollingStalled={true}` passed as a literal prop — it does not verify the orchestrator ever *sets* it. A regression that broke the catch-branch increment or the `setPollingStalled(true)` flip — i.e. the exact infinite-spinner bug round 2 fixed — would pass CI. Add a test that opens a region, makes `getFitnessRouteHeatmap` keep rejecting (or return the same `generating` fingerprint), advances `STALLED_POLLING_LIMIT` (30) intervals, and asserts the detail surfaces "taking longer than expected" + a Retry.

llun:
**[round3/completeness]** `describeRegions` in `lib/fitness/regions.ts:187` is now a dead export. It was consumed only by the deleted `FitnessHeatmapList.tsx` and the old flat `FitnessHeatmapView.tsx` (both gone/rewritten in this PR); the only remaining reference in the repo is its own unit test (`lib/fitness/regions.test.ts`). It is not in this PR's diff, so it slipped through silently. Either delete it (and its test) as leftover from the removed list, or keep it deliberately if it's intended for reuse — but as-is it's an orphaned public helper kept alive only by its test. (Lower priority: `formatLatitude`/`formatLongitude` are exported but only consumed internally by `formatRectRegion` in the same file — pre-existing, not introduced here.)

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in 2c04652 — added 'surfaces the stalled state after repeated no-progress focused polls', which opens a region and drives the focused poll past STALLED_POLLING_LIMIT (via advanceTimersByTimeAsync), then asserts the stalled banner appears. A regression reintroducing the infinite spinner now fails CI.

llun:
Fixed in 2c04652 — removed describeRegions from lib/fitness/regions and its unit test; it was only used by the now-deleted flat heatmap list, so it was dead.

PR #1070, opened 2026-06-22
fix: flatten + focus the route heatmap map and brighten route lines

llun:
## Summary

The fitness route heatmap result map had two usability problems when viewing the **Whole world** cache:

1. **It opened as a zoomed-out globe.** Mapbox GL v3 defaults to the globe projection at low zoom, so a cache spanning disjoint regions (e.g. Europe + Singapore) rendered as a tiny scatter of dots on a 3D sphere.
2. **The routes were hard to see.** The colourful `outdoors-v12` / `bright` basemaps plus a very low route opacity (0.2) washed the lines out.

This PR makes the result map open as a **flat, pannable map focused on the densest activity cluster**, on a **light basemap** that keeps the routes legible.

## Changes

- **Mapbox → `light-v11` + `projection: 'mercator'`.** Light 2D basemap (as requested) and an explicit flat projection, so a wide cache opens as a pannable flat map instead of the default zoomed-out globe.
- **MapLibre (keyless) → OpenFreeMap `positron`.** The colourful `bright` style washed the overlay out; `positron` is light/low-saturation. Served from the same `tiles.openfreemap.org` origin, so **no CSP change**. The region **picker** keeps `bright`.
- **Brighter route paint.** Higher opacity (0.55 visible / 0.4 hidden) and crisper blur (0.4) so a single pass is clearly visible while overlaps still build into a density gradient.
- **`computeFocusBounds` (new, exported + tested).** Buckets route vertices into an absolute 5° lon/lat grid and flood-fills the 8-connected cluster containing the densest cell. Multi-region caches open framed on that cluster (zoom capped so it opens with pannable context); a single contiguous region returns its full bounds unchanged. Far-away outliers no longer blow up the initial view. The map stays fully interactive, so the user can pan to the other regions.

## Behaviour

| Cache | Before | After |
| --- | --- | --- |
| Single region | Flat fit to bounds | Unchanged (flat fit to bounds) |
| Whole world / disjoint | Zoomed-out globe, faint dots | Flat map zoomed into the densest cluster, bold routes, pannable |

## Verification

- `yarn lint` — 0 errors.
- `yarn build` — green.
- `yarn test` — 4746 passed (added unit tests for `computeFocusBounds`, integration tests asserting the focused `fitBounds` + `maxZoom`, Mapbox `light-v11`/`mercator` options, and the `positron` style URL).
- Browser visual check of the keyless MapLibre path confirmed the route lines are far more legible on the light basemap (before/after attached in the PR thread).

## Notes

- Clusters straddling the antimeridian (±180°) land in non-adjacent grid cells and would not be merged; acceptable for this best-effort initial framing (documented in `computeFocusBounds`).
- No migration; no API/contract change.

### DIFF
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1056,7 +1056,11 @@ describe('RouteHeatmapMap', () => {
       [SINGAPORE_CLUSTER_BOUNDS.minLng, SINGAPORE_CLUSTER_BOUNDS.minLat],
       [SINGAPORE_CLUSTER_BOUNDS.maxLng, SINGAPORE_CLUSTER_BOUNDS.maxLat]
     ])
-    expect(fitBounds.mock.calls[0][1]).toMatchObject({ maxZoom: 12 })
+    expect(fitBounds.mock.calls[0][1]).toMatchObject({
+      padding: 56,
+      duration: 0,
+      maxZoom: 12
+    })
   })
 })
 
@@ -1250,15 +1254,18 @@ describe('computeFocusBounds', () => {
     expect(result.bounds).toBe(bounds)
   })
 
-  it('does not merge cells a knight’s-move apart, focusing the densest', () => {
+  it('does not merge cells a knight’s-move apart, focusing the densest cell', () => {
     const bounds = { minLat: 52, maxLat: 55, minLng: 4, maxLng: 13 }
-    // Cells 0:10 and 2:11 are not 8-adjacent (dx=2), so they stay separate — this
-    // pins that it is grid adjacency, not mere proximity, that merges clusters.
+    // Cell 0:10 (two points) is strictly denser than cell 2:11 (one point), and
+    // the two cells are a knight's move apart (dx=2), so they stay separate — this
+    // pins that grid adjacency (not mere proximity) is what merges clusters, and
+    // that the seed is chosen by density rather than Map insertion order.
     const result = computeFocusBounds(
       [
         {
           points: [
             { lat: 52, lng: 4 },
+            { lat: 53, lng: 4 },
             { lat: 55, lng: 13 }
           ]
         }
@@ -1269,9 +1276,44 @@ describe('computeFocusBounds', () => {
     expect(result.focused).toBe(true)
     expect(result.bounds).toEqual({
       minLat: 52,
-      maxLat: 52,
+      maxLat: 53,
       minLng: 4,
       maxLng: 4
+    })
+  })
+
+  it('frames the densest cell even when another cluster has more cells', () => {
+    // Cluster A spans two 8-adjacent cells (0:0 and 0:1) of one point each — the
+    // largest connected cluster by cell count. Cluster B is a single isolated cell
+    // (8:0) holding three points — the densest cell. The focus must frame B (seed
+    // = densest cell, then its cluster), proving the helper does not instead pick
+    // the cluster with the most cells.
+    const bounds = { minLat: 0, maxLat: 7, minLng: 2, maxLng: 40.2 }
+    const result = computeFocusBounds(
+      [
+        {
+          points: [
+            { lat: 2, lng: 2 },
+            { lat: 7, lng: 2 }
+          ]
+        },
+        {
+          points: [
+            { lat: 0, lng: 40 },
+            { lat: 0.1, lng: 40.1 },
+            { lat: 0.2, lng: 40.2 }
+          ]
+        }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(true)
+    expect(result.bounds).toEqual({
+      minLat: 0,
+      maxLat: 0.2,
+      minLng: 40,
+      maxLng: 40.2
     })
   })
 })
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1228,4 +1228,50 @@ describe('computeFocusBounds', () => {
       maxLng: 179.9
     })
   })
+
+  it('merges cells that touch only diagonally (8-connectivity)', () => {
+    const bounds = { minLat: 52, maxLat: 55, minLng: 4, maxLng: 8 }
+    // The two 5° cells (0:10 and 1:11) are diagonal neighbours — connected only
+    // at a corner. 8-connectivity merges them into one contiguous cluster, so
+    // the full bounds are kept (a 4-connected flood fill would NOT merge these).
+    const result = computeFocusBounds(
+      [
+        {
+          points: [
+            { lat: 52, lng: 4 },
+            { lat: 55, lng: 8 }
+          ]
+        }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(false)
+    expect(result.bounds).toBe(bounds)
+  })
+
+  it('does not merge cells a knight’s-move apart, focusing the densest', () => {
+    const bounds = { minLat: 52, maxLat: 55, minLng: 4, maxLng: 13 }
+    // Cells 0:10 and 2:11 are not 8-adjacent (dx=2), so they stay separate — this
+    // pins that it is grid adjacency, not mere proximity, that merges clusters.
+    const result = computeFocusBounds(
+      [
+        {
+          points: [
+            { lat: 52, lng: 4 },
+            { lat: 55, lng: 13 }
+          ]
+        }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(true)
+    expect(result.bounds).toEqual({
+      minLat: 52,
+      maxLat: 52,
+      minLng: 4,
+      maxLng: 4
+    })
+  })
 })
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1145,11 +1145,87 @@ describe('computeFocusBounds', () => {
     })
   })
 
-  it('returns the full bounds when there are no finite points', () => {
+  it('returns the full bounds for an empty segment list', () => {
     const bounds = { minLat: 0, maxLat: 0, minLng: 0, maxLng: 0 }
     const result = computeFocusBounds([], bounds)
 
     expect(result.focused).toBe(false)
     expect(result.bounds).toBe(bounds)
+  })
+
+  it('skips non-finite vertices so they create no spurious cluster', () => {
+    const bounds = { minLat: 52.36, maxLat: 52.39, minLng: 4.88, maxLng: 4.91 }
+    const result = computeFocusBounds(
+      [
+        {
+          points: [
+            { lat: 52.36, lng: 4.88 },
+            { lat: 52.39, lng: 4.91 }
+          ]
+        },
+        // An entirely non-finite segment must contribute no grid cell; otherwise
+        // it would look like a second region and flip the result to focused.
+        {
+          points: [
+            { lat: NaN, lng: NaN },
+            { lat: Infinity, lng: -Infinity }
+          ]
+        }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(false)
+    expect(result.bounds).toBe(bounds)
+  })
+
+  it('ignores non-finite vertices when framing a focused cluster', () => {
+    const bounds = { minLat: 1.3, maxLat: 52.39, minLng: 4.88, maxLng: 103.9 }
+    const result = computeFocusBounds(
+      [
+        {
+          points: [
+            { lat: 52.36, lng: 4.88 },
+            { lat: 52.37, lng: 4.89 },
+            { lat: 52.39, lng: 4.91 }
+          ]
+        },
+        // The denser cluster carries a stray non-finite vertex that must not
+        // widen (or NaN-poison) the focused box.
+        { points: [{ lat: NaN, lng: 103.8 }, ...SINGAPORE_CLUSTER_POINTS] }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(true)
+    expect(result.bounds).toEqual(SINGAPORE_CLUSTER_BOUNDS)
+  })
+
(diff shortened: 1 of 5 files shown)


app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -961,4 +1023,209 @@ describe('RouteHeatmapMap', () => {
...
+
+  it('returns the full bounds for an empty segment list', () => {
+    const bounds = { minLat: 0, maxLat: 0, minLng: 0, maxLng: 0 }
+    const result = computeFocusBounds([], bounds)
+
+    expect(result.focused).toBe(false)
+    expect(result.bounds).toBe(bounds)
+  })
+
+  it('skips non-finite vertices so they create no spurious cluster', () => {
```
llun, line 1160: **Sub-agent review (tests lens) — minor:** the new `computeFocusBounds` non-finite guards were not exercised. The only "no finite points" case passed an **empty** segment array, which returns early at `cells.size <= 1` and never enters the bucketing loop, so the `Number.isFinite` skip branch had 0% coverage and the test name overstated what it checked.

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -280,7 +431,23 @@ export const RouteHeatmapMap: FC<RouteHeatmapMapProps> = ({
               },
               paint: ROUTE_LINE_PAINT
             })
-            map.fitBounds(mapBounds, { padding: 56, duration: 0 })
+            // Frame the densest cluster (focused) or the full extent. The focus
```
llun, line 434: **Sub-agent review (ux/correctness lens) — nit:** the previous comment here ("…so an updated cache reframes too") overstated the behaviour. `fitBounds` only runs inside the GL `load` handler, and the map-creating effect deps exclude `focus`/`heatmap.updatedAt`, so a same-id in-place cache update updates the geometry (`setData`) but does **not** reframe. The `focusRef` only carries the latest focus into the asynchronous initial `load` (stale-closure avoidance).

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -147,6 +161,128 @@ export const downsampleSegments = (
...
+// has bounds spanning every recorded region (e.g. Europe *and* Singapore), so
+// fitting the full extent renders the routes as a tiny scatter on a flat world
+// map. Instead, bucket the route vertices into an absolute lon/lat grid, flood
+// fill the 8-connected cluster containing the densest cell, and fit to that
+// cluster — so the map opens zoomed in on where most activity is while the user
+// can still pan to the other regions. For a single contiguous region every
+// vertex falls in one connected cluster, so the authoritative full bounds are
+// returned unchanged (focused: false).
+//
+// Note: clusters straddling the antimeridian (±180° lon) land in non-adjacent
```
llun, line 192: **Sub-agent review (tests lens) — minor:** the documented antimeridian limitation (clusters near ±180° lon land in non-adjacent grid cells and are not merged) had no regression test pinning the behaviour.

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -147,6 +161,128 @@ export const downsampleSegments = (
...
+// cells and would not be merged; that is acceptable for this best-effort initial
+// framing (mercator fitBounds has its own antimeridian limitations regardless).
+export const computeFocusBounds = (
```
llun, line 195: **Sub-agent review (ux lens) — nit (by-design):** for a whole-world cache with several *near-equal* disjoint clusters, focusing on the single densest cluster means the others open off-screen until the user pans. Flagging to confirm this is the intended trade-off rather than, say, fitting all clusters.

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -147,6 +161,107 @@ export const downsampleSegments = (
...
+      if (point.lng < minLng) minLng = point.lng
+      if (point.lng > maxLng) maxLng = point.lng
+    }
+  }
+
+  // Defensive: no finite point matched the cluster (should not happen) — fall
+  // back to the full bounds rather than an inverted/NaN box.
+  if (!Number.isFinite(minLat)) {
+    return { bounds, focused: false }
+  }
```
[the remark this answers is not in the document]
llun: Adopted in `ee9dabea1`: single-pass per-cell bounding box + mark-on-enqueue flood fill (each occupied cell pushed exactly once), removing the second O(N) scan. The per-cell box is always finite so the previously-defensive `minLat` fallback became unreachable and was dropped. Behaviour is identical — all existing `computeFocusBounds` tests still pass. Thanks!

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in `ee9dabea1`: renamed the case to "returns the full bounds for an empty segment list" and added two cases — non-finite vertices create no spurious cluster (stays `focused: false`) and are ignored when framing a focused cluster (box equals the finite extent).

lib/components/fitness/RouteHeatmapMap.tsx
llun: Fixed in `ee9dabea1`: reworded to "The focus is read from a ref so this asynchronous load handler uses the latest computed value rather than a stale closure; fitBounds runs once per map mount (not on in-place, same-id cache updates)."

lib/components/fitness/RouteHeatmapMap.tsx
llun: Fixed in `ee9dabea1`: added a regression test ("does not merge clusters straddling the antimeridian") asserting the focus frames only the denser side near ±180° lon.

lib/components/fitness/RouteHeatmapMap.tsx
llun: Confirmed intended. The goal was to open zoomed-in on where most activity is and let the user pan to the rest; densest-cluster framing delivers exactly that. Truly near-equal disjoint clusters are rare and remain one pan/zoom-out away. Leaving as-is by design (documented in the helper comment).

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -961,4 +1023,255 @@ describe('RouteHeatmapMap', () => {
...
+    expect(result.focused).toBe(true)
+    expect(result.bounds).toEqual({
+      minLat: 0.5,
+      maxLat: 0.6,
+      minLng: 179.5,
+      maxLng: 179.9
+    })
+  })
+
+  it('merges cells that touch only diagonally (8-connectivity)', () => {
```
llun, line 1236: **Sub-agent review round 2 (tests lens) — nit:** the flood fill is 8-connected, but no test exercised a *diagonal-only* (corner-touching) cell adjacency. All connectivity fixtures were single-cell, non-adjacent, or horizontally adjacent, so a silent 8→4 connectivity regression would still have passed the whole suite — the helpers one genuinely novel behaviour was unpinned.

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -147,6 +161,128 @@ export const downsampleSegments = (
...
+// accumulated in a single pass so the focused extent needs no second scan.
+interface FocusCell {
+  count: number
+  minLat: number
+  maxLat: number
+  minLng: number
+  maxLng: number
+}
+
+// Pick the initial map view for a route cache. A whole-world / multi-region cache
```
llun, line 182: **Sub-agent review round 2 (ux lens) — nit (by-design):** a *single* region whose activity is sparse enough to leave >5° grid gaps is treated as multiple clusters and opens focused on the densest part, with no on-screen "more off-screen" affordance. Flagging to confirm the trade-off.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in `48f5a3c2`: added a diagonal-merge case (cells `0:10` and `1:11` touch only at a corner → `focused:false`, full bounds kept) plus a knights-move control (cells `0:10` and `2:11` → `focused:true`). Together these kill the 8→4 connectivity mutation the suite previously missed.

lib/components/fitness/RouteHeatmapMap.tsx
llun: Confirmed intended. The 5° cell size keeps genuinely-disjoint regions (e.g. Europe vs Singapore, ~100° apart) separate while merging a contiguous metro area via 8-connectivity. The map stays fully interactive (pan + scroll-zoom), matching the requested "zoom in and let the user move around"; zooming out reveals the other clusters. Leaving as-is by design — a persistent "other regions" affordance would be a separate enhancement.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -961,4 +1023,297 @@ describe('RouteHeatmapMap', () => {
...
+    expect(result.focused).toBe(true)
+    expect(result.bounds).toEqual({
+      minLat: 52,
+      maxLat: 53,
+      minLng: 4,
+      maxLng: 4
+    })
+  })
+
+  it('frames the densest cell even when another cluster has more cells', () => {
```
llun, line 1285: **Sub-agent review round 3 (tests lens) — nit:** no test pinned the load-bearing semantics that the focus seeds from the densest single *cell* and then frames *that cells* connected cluster. In every focused-path test each cluster was a single cell, so "densest cell" and "largest cluster" never diverged — a refactor to "seed by largest cluster" would have passed the whole suite.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -961,4 +1023,297 @@ describe('RouteHeatmapMap', () => {
...
+        }
+      ],
+      bounds
+    )
+
+    expect(result.focused).toBe(false)
+    expect(result.bounds).toBe(bounds)
+  })
+
+  it('does not merge cells a knight’s-move apart, focusing the densest cell', () => {
```
llun, line 1257: **Sub-agent review round 3 (tests lens) — nit:** the knights-move case had both cells tied at one point, so the asserted box was actually decided by Map insertion order, not density — the "focusing the densest" framing was misleading and a point reorder would have flipped it.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -961,4 +1023,297 @@ describe('RouteHeatmapMap', () => {
...
+    ])
+    // A single region keeps the natural fit — no focus zoom cap is applied.
+    expect(fitBounds.mock.calls[0][1]).not.toHaveProperty('maxZoom')
+    expect(fitBounds.mock.calls[0][1]).toMatchObject({
+      padding: 56,
+      duration: 0
+    })
+  })
+
+  it('opens focused on the densest cluster for a disjoint multi-region cache', async () => {
```
llun, line 1046: **Sub-agent review round 3 (tests lens) — minor:** the focused integration test asserted only `maxZoom` while its full-bounds sibling asserts `padding`/`duration`, leaving those options unpinned on the focused path.

lib/components/fitness/RouteHeatmapMap.tsx
llun, line 182: **Sub-agent review round 3 (ux lens) — by-design (repeat):** re-raised that the 5° grid drops a disjoint region from the initial view and that framing is not re-applied on in-place same-id cache updates. Both are the same intended trade-offs already discussed and resolved in earlier rounds.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in `1ffb90d9`: added "frames the densest cell even when another cluster has more cells" — cluster A = two 8-adjacent cells (1 pt each), cluster B = one cell with 3 pts; asserts the focus frames B. A "largest cluster" refactor now fails this.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in `1ffb90d9`: the seed cell now holds two points (strictly denser than the knights-move cell), so the expected box is density-driven; updated the comment to call out that adjacency merges and density seeds.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in `1ffb90d9`: the focused test now asserts `toMatchObject({ padding: 56, duration: 0, maxZoom: 12 })`, matching the full-bounds sibling.

lib/components/fitness/RouteHeatmapMap.tsx
llun: Confirmed — both are intended and were resolved earlier: densest-cluster framing is the requested "open zoomed-in and let the user pan" behaviour (5° keeps Europe/Singapore-scale regions distinct), and the map only mounts on a completed cache so there is nothing to reframe mid-generation. No code change.

lib/components/fitness/RouteHeatmapMap.tsx
```diff
@@ -147,6 +161,128 @@ export const downsampleSegments = (
   })
 }
 
+export interface RouteFocusBounds {
+  bounds: FitnessRouteHeatmapBounds
+  /** True when the view was tightened to a single dense cluster (disjoint data). */
```
llun, line 166: **Sub-agent review round 4 (ux lens) — by-design:** because the seed is the densest single cell, a small but highly *concentrated* outlier (many points packed into one tiny far-away area) could out-seed the users main area and open the map there. Recording the trade-off.

llun: Acknowledged as the inherent trade-off of densest-cell framing. In practice the heatmap aggregates a users own routes, where the "densest" area is their home/most-ridden region — exactly what should open first; a pathological tiny-but-ultra-dense remote cluster is unlikely, and the map remains fully pannable/zoomable to reach it. Left as-is by design.

PR #1079, opened 2026-06-23
feat: remove No Announces and Mention timeline tabs

llun:
## Summary

The main timeline page exposed three sub-tabs — **Home / No Announces / Mention** — each backed by its own materialized projection in the `timelines` table. Now that the **lists** feature covers custom feeds, the extra tabs are redundant. The timeline page now renders a **single Home feed with no tab strip**.

## What changed

- **UI:** Removed the tab strip from `MainPageTimeline`; it renders the Home (`MAIN`) feed only (load-more / refresh / scroll-to-top all retained).
- **Timeline enum & query:** Dropped `NOANNOUNCE` and `MENTION` from the `Timeline` enum, removed their `getTimeline` switch branches, and deleted the no-announce timeline rule. `/api/v1/timelines/noannounce` and `/api/v1/timelines/mention` now return **404** (invalid timeline), like any other unknown timeline.
- **Notifications preserved (important):** the old `mentionTimelineRule` did double duty — it materialized the mention timeline **and** created reply/mention notifications for **remote** statuses. That has been refactored into `notifyRemoteReplyAndMention`, a **notification-only** side effect invoked from `addStatusToTimelines`. The Alerts feature reads the separate `notifications` table and is unaffected; local-actor notifications still flow through `createNote.ts`. All the reply/mention dedup, email-channel, block, and policy-filter behaviors are kept (and still covered by the renamed test).
- **Migration:** `20260624000000_remove_noannounce_mention_timelines.js` deletes existing `noannounce`/`mention` rows from the `timelines` table. It is **data-only** (a `DELETE`, no DDL), so the reference schema dumps (`migrations/schema.sql`, `migrations/schema.sqlite.sql`) are unchanged and need no regeneration. `down` is a no-op (the rows are a rebuildable projection of `statuses`).
- **Mock + tests:** mock seeding only writes the `MAIN` timeline now; tests updated/renamed accordingly.

## Verification

- `yarn run prettier --write .` — clean
- `yarn lint` — 0 errors
- `yarn build` — success
- `yarn test` — **548 files / 4834 tests pass**


lib/services/timelines/index.test.ts
```diff
@@ -139,24 +133,9 @@ describe('addStatusToTimelines', () => {
...
-      timeline: Timeline.MENTION,
-      actorId: ACTOR1_ID
-    })
-    const senderMentionTimeline = await database.getTimeline({
-      timeline: Timeline.MENTION,
-      actorId: ACTOR2_ID
-    })
 
     expect(directTimeline.some((s) => s.id === status.id)).toBe(true)
     expect(mainTimeline.some((s) => s.id === status.id)).toBe(false)
```
llun, line 138: **[sub-agent review · test-coverage · low]** Direct-status test does not assert absence of mention notification.

This test creates a direct status carrying a `mention` tag for ACTOR1 and only asserts DIRECT membership + MAIN absence. It never asserts that no reply/mention notification was created. The `if (isDirect) { … return }` early-return in `addStatusToTimelines` is the only thing keeping `notifyRemoteReplyAndMention` from firing for DMs, and nothing guards that contract. Add a negative `getNotifications` assertion so a future regression that drops the early return fails here.

migrations/20260624000000_remove_noannounce_mention_timelines.js
```diff
@@ -0,0 +1,33 @@
...
+ * the separate `notifications` table and are untouched), so `down` is a no-op.
+ *
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export const up = async function (knex) {
+  const hasTable = await knex.schema.hasTable('timelines')
+  if (!hasTable) return
+
+  await knex('timelines').whereIn('timeline', ['noannounce', 'mention']).del()
```
llun, line 26: **[sub-agent review · db-migration · medium]** Schema dumps not regenerated.

AGENTS.md (“Keeping the reference schema dumps in sync”) requires regenerating BOTH `migrations/schema.sql` and `migrations/schema.sqlite.sql` for *any* added migration, with no data-only carve-out. This migration is a pure `DELETE` (no DDL), so the dumps are byte-identical — but the canonical regeneration + `knex_migrations` count check should still be run and the no-diff result confirmed so the rule is demonstrably honored.

lib/services/timelines/types.ts
```diff
@@ -46,12 +44,4 @@ export type MainTimelineRule = (
...
-export type NoAnnounceTimelineRule = (
-  params: TimelineRuleParams
-) => Promise<Timeline.NOANNOUNCE | null>
-
-export type MentionTimelineRule = (
-  params: TimelineRuleParams
-) => Promise<Timeline.MENTION | null>
-
-export type TimelineRule = MainTimelineRule | NoAnnounceTimelineRule
+export type TimelineRule = MainTimelineRule
```
llun: **[sub-agent review · leftover-refs · low]** Dead/redundant `TimelineRule` alias.

After removing `NoAnnounceTimelineRule`/`MentionTimelineRule`, `export type TimelineRule = MainTimelineRule` is a 1:1 alias with zero consumers anywhere in the repo. Remove it; callers can reference `MainTimelineRule` directly.

app/(timeline)/MainPageTimeline.tsx
```diff
@@ -160,13 +142,13 @@ export const MainPageTimeline: FC<MainPageTimelineProps> = ({
     if (isLoadingRef.current) return
 
     isLoadingRef.current = true
-    const requestId = ++tabRequestId.current
+    const requestId = ++refreshRequestId.current
```
llun: **[sub-agent review · ui-correctness · low]** `refreshRequestId` stale-request guard is now dead code.

With `onTabChange`/`tabRequestId` gone, nothing else bumps this counter and `isLoadingRef` already serializes refreshes, so `requestId !== refreshRequestId.current` is never true and the finally guard is always taken. Drop the ref and its two comparisons; the `finally` can unconditionally call `setLoadingMoreStatuses(false)`.

lib/services/timelines/index.test.ts
llun: Addressed in fd3d236: the direct-status test now fetches ACTOR1 notifications after addStatusToTimelines and asserts none reference the status id, locking the “DMs create no mention notification” contract against a future regression of the isDirect early-return.

migrations/20260624000000_remove_noannounce_mention_timelines.js
llun: Done: ran the canonical regeneration for both backends (throwaway local Postgres 17 + SQLite, 129 migrations each, knex_migrations count verified). Statement-set diff against both committed dumps is empty — the data-only DELETE emits no DDL — so the dumps stay in lockstep with nothing to commit. Noted the verified no-delta result in the migration docstring.

lib/services/timelines/types.ts
llun: Removed in fd3d236 — the TimelineRule alias is gone; callers reference MainTimelineRule directly.

app/(timeline)/MainPageTimeline.tsx
llun: Fixed in fd3d236: dropped the refreshRequestId ref and both comparisons; the finally now unconditionally calls setLoadingMoreStatuses(false). isLoadingRef already serializes refreshes.

lib/services/timelines/index.ts
```diff
@@ -113,26 +112,33 @@ export const addStatusToTimelines = async (
...
               })
+            })(),
+            // Side effect only: create reply/mention notifications for inbound
+            // remote statuses (the "mention" timeline itself was removed). The
+            // local-actor path is handled in createNote.ts.
+            notifyRemoteReplyAndMention({
+              currentActor: actor,
+              status,
+              database
             })
```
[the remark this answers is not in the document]
llun: Good call — adopted in 65436f6e. notifyRemoteReplyAndMention is now wrapped in a .catch that logs and continues, so a transient notification failure can no longer abort home-timeline materialization or lose the status. This matches the existing materializeAuxiliaryFeed best-effort pattern for the list/collection fan-out in the same file.

lib/services/timelines/mentionNotification.ts
```diff
@@ -134,49 +147,45 @@ export const mentionTimelineRule: MentionTimelineRule = async ({
...
-              }
+          if (config.email && account && status.actor) {
+            mentionEvent.emailContent = {
+              recipientEmail: account.email,
+              subject: getMentionSubject(status.actor),
+              text: getMentionTextContent(status),
+              html: getMentionHTMLContent(status)
             }
-            alertEvents.push(mentionEvent)
           }
```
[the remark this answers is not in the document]
llun: The reply→mention suppression you describe is already in place: replyNotificationCreated is set to true the moment the status is identified as a reply to the current actor — *before* the policy verdict is inspected — and the mention branch is gated on !replyNotificationCreated. So a filtered or dropped reply still suppresses the duplicate mention (covered by the “keeps a single filtered reply notification …” test, which passes). Note the field is status.reply, not status.inReplyToId. On the status.actor fallback: the original mentionTimelineRule gated both the reply and mention email branches on status.actor, and this refactor deliberately preserves that for parity; status.actor is populated on this path in practice. Adding a DB fallback to only the mention branch would diverge the two paths, so leaving as-is to keep behavior identical to before. Resolving as already-handled.

PR #1096, opened 2026-06-24
minor: add Lists & Collections page wiring the Collections API into the UI

llun:
## Summary

Redesigns the lists page into a unified **Lists & Collections** surface, matching `ui_kits/web/lists.html` from the design system, and wires the **existing** Mastodon Collections API into the web UI.

Collections were already implemented on the backend (tables, API routes, federation, notifications). This PR is the **frontend integration** plus the small backend reads needed to surface it, scoped per the requested behavior:

- **API-backed owner management** of collections
- **Member consent** surfaced on the existing `added_to_collection` notifications
- A **public collection page** for the shareable link

### What's included

**Index (`/lists`)**
- Renamed to "Lists & Collections"; groups **Collections** (shareable feeds) and **Lists** (private timelines).
- Collection rows show `N people · M featured publicly`; header has **New list** + **New collection**.

**Collection create/edit** (`/collections/new`, `/collections/[id]/edit`)
- `CollectionEditor`: name, description (counter), topic (single hashtag, sanitized), visibility (public/unlisted/private), shareable-feed toggle, and member add/remove via the items API.

**Collection detail** (`/collections/[id]`)
- Owner sees an **Owner view / Public preview** feed toggle (owner timeline vs the consent-gated public feed), a meta panel (visibility, topic, description), a working **Copy link**, and the roster.
- Non-owner / anonymous viewers get a **read-only public page** wrapped in `PublicShell`, gated to `public`/`unlisted` + feed-enabled collections (private/feed-disabled/missing → 404).

**Member consent (notifications)**
- `added_to_collection` notifications now offer **Show me publicly / Keep me hidden**, calling the self-only `approve`/`revoke` endpoints. (The relationship branch previously mis-rendered these as follow requests — fixed.)

**Client / backend**
- `lib/client.ts`: new collection functions (create/update/delete, add/remove members, approve/revoke, owner timeline + public feed).
- `getCollectionById` (non-owner-scoped read) for the public page + notification titles.
- `format=activities_next` on the public feed route so `<Posts>` renders it like the list timeline.

### Deviations from the design (API constraints)
- **No owner-side per-member consent toggle / per-member badges in the roster**: the items API returns plain `Account[]` with no `featureState`, and approve/revoke are member-self-only. Consent is therefore driven from the member's notifications (as chosen).
- **No in-page "Added to collections" list**: there is no endpoint enumerating collections others added you to; the member entry point is the notification.
- Sidebar (Surface 1) intentionally **out of scope** (page only).

## Testing
- `yarn run prettier --write .`, `yarn lint`, `yarn build`, `yarn test` — all green (**553 files / 4900 tests**).
- New coverage: `CollectionEditor`, `CollectionDetail`, `CollectionConsentNotification`, `ListsIndex` (collections), collection client helpers, `getCollectionById`, and the feed route `activities_next` format.
- Browser-verified locally (SQLite): index (empty + populated), create→edit, owner detail (toggle/share/roster), and the anonymous public page (200 + `PublicShell`; 404 for missing). Zero console errors.

### DIFF
--- a/app/(timeline)/collections/CollectionEditor.test.tsx
+++ b/app/(timeline)/collections/CollectionEditor.test.tsx
@@ -187,6 +187,26 @@ describe('CollectionEditor', () => {
     )
   })
 
+  it('shows an inline error and keeps the member when the remove fails', async () => {
+    ;(removeCollectionAccounts as jest.Mock).mockResolvedValue(false)
+    render(
+      <CollectionEditor
+        mode="edit"
+        collection={collection}
+        initialMembers={[member]}
+        followingSuggestions={[]}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Remove Rin' }))
+
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Could not remove that account. Please try again.'
+    )
+    // The member is not optimistically dropped on a failed remove.
+    expect(screen.getByText('In this collection · 1')).toBeInTheDocument()
+  })
+
   it('saves changes and routes back to the collection', async () => {
     render(
       <CollectionEditor
@@ -212,5 +232,41 @@ describe('CollectionEditor', () => {
       })
     )
     expect(mockPush).toHaveBeenCalledWith('/collections/col-1')
+  })
+
+  it('deletes the collection after confirmation and routes back to the index', async () => {
+    const confirmSpy = vi.spyOn(window, 'confirm').mockReturnValue(true)
+    render(<CollectionEditor mode="edit" collection={collection} />)
+
+    fireEvent.click(screen.getByRole('button', { name: 'Delete collection' }))
+
+    await waitFor(() => expect(deleteCollection).toHaveBeenCalledWith('col-1'))
+    expect(mockPush).toHaveBeenCalledWith('/lists')
+    confirmSpy.mockRestore()
+  })
+
+  it('does not delete when the confirmation is dismissed', () => {
+    const confirmSpy = vi.spyOn(window, 'confirm').mockReturnValue(false)
+    render(<CollectionEditor mode="edit" collection={collection} />)
+
+    fireEvent.click(screen.getByRole('button', { name: 'Delete collection' }))
+
+    expect(deleteCollection).not.toHaveBeenCalled()
+    expect(mockPush).not.toHaveBeenCalled()
+    confirmSpy.mockRestore()
+  })
+
+  it('shows an inline error when the delete fails', async () => {
+    const confirmSpy = vi.spyOn(window, 'confirm').mockReturnValue(true)
+    ;(deleteCollection as jest.Mock).mockResolvedValue(false)
+    render(<CollectionEditor mode="edit" collection={collection} />)
+
+    fireEvent.click(screen.getByRole('button', { name: 'Delete collection' }))
+
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Could not delete the collection. Please try again.'
+    )
+    expect(mockPush).not.toHaveBeenCalled()
+    confirmSpy.mockRestore()
   })
 })
--- a/app/(timeline)/collections/CollectionEditor.test.tsx
+++ b/app/(timeline)/collections/CollectionEditor.test.tsx
@@ -1,0 +1,197 @@
+/**
+ * @vitest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+
+import {
+  addCollectionAccounts,
+  createCollection,
+  deleteCollection,
+  removeCollectionAccounts,
+  updateCollection
+} from '@/lib/client'
+import { CollectionEntity } from '@/lib/types/mastodon/collection'
+
+import { CollectionEditor, CollectionMember } from './CollectionEditor'
+
+vi.mock('@/lib/client', () => ({
+  addCollectionAccounts: vi.fn(),
+  createCollection: vi.fn(),
+  deleteCollection: vi.fn(),
+  removeCollectionAccounts: vi.fn(),
+  updateCollection: vi.fn()
+}))
+
+const mockPush = vi.fn()
+const mockRefresh = vi.fn()
+vi.mock('next/navigation', () => ({
+  useRouter: () => ({ push: mockPush, refresh: mockRefresh })
+}))
+
+const collection: CollectionEntity = {
+  id: 'col-1',
+  title: 'Fediverse builders',
+  description: 'who I read',
+  topic: 'fediverse',
+  language: null,
+  visibility: 'public',
+  feed_enabled: true,
+  size: 0
+}
+
+const member: CollectionMember = {
+  id: 'https://activities.local/users/rin',
+  name: 'Rin',
+  handle: 'rin@fosstodon.org'
+}
+
+const suggestion: CollectionMember = {
+  id: 'https://activities.local/users/ben',
+  name: 'Ben Carter',
+  handle: 'ben@llun.social'
+}
+
+describe('CollectionEditor', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    ;(createCollection as jest.Mock).mockResolvedValue({
+      ...collection,
+      id: 'new-col'
+    })
+    ;(updateCollection as jest.Mock).mockResolvedValue(collection)
+    ;(addCollectionAccounts as jest.Mock).mockResolvedValue(true)
+    ;(removeCollectionAccounts as jest.Mock).mockResolvedValue(true)
+    ;(deleteCollection as jest.Mock).mockResolvedValue(true)
+  })
+
+  it('creates a collection and routes to its member editor', async () => {
+    render(<CollectionEditor mode="create" />)
+
+    fireEvent.change(screen.getByLabelText('Collection name'), {
+      target: { value: 'New crew' }
+    })
+    fireEvent.click(screen.getByRole('button', { name: 'Create collection' }))
+
+    await waitFor(() =>
+      expect(createCollection).toHaveBeenCalledWith({
+        title: 'New crew',
+        // Empty optional text fields are sent as null to clear them.
+        description: null,
+        topic: null,
+        visibility: 'public',
+        feedEnabled: true
+      })
+    )
+    expect(mockPush).toHaveBeenCalledWith('/collections/new-col/edit')
+  })
+
+  it('blocks creation when the name is empty', async () => {
+    render(<CollectionEditor mode="create" />)
+
+    fireEvent.click(screen.getByRole('button', { name: 'Create collection' }))
+
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Please enter a collection name.'
+    )
+    expect(createCollection).not.toHaveBeenCalled()
+  })
+
+  it('sanitizes the topic to a bare hashtag before saving', async () => {
+    render(<CollectionEditor mode="create" />)
+
+    fireEvent.change(screen.getByLabelText('Collection name'), {
+      target: { value: 'Crew' }
+    })
+    fireEvent.change(screen.getByLabelText('Topic'), {
+      target: { value: '#foss dev!' }
+    })
+    fireEvent.click(screen.getByRole('button', { name: 'Create collection' }))
+
+    await waitFor(() =>
+      expect(createCollection).toHaveBeenCalledWith(
+        expect.objectContaining({ topic: 'fossdev' })
+      )
+    )
+  })
+
+  it('does not render the people section in create mode', () => {
+    render(<CollectionEditor mode="create" />)
+    expect(screen.queryByText('People')).not.toBeInTheDocument()
+  })
+
+  it('adds a suggested account right away', async () => {
+    render(
+      <CollectionEditor
+        mode="edit"
+        collection={collection}
+        initialMembers={[]}
+        followingSuggestions={[suggestion]}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Add' }))
+
+    await waitFor(() =>
+      expect(addCollectionAccounts).toHaveBeenCalledWith({
+        collectionId: 'col-1',
+        accountIds: [suggestion.id]
+      })
+    )
+    await waitFor(() =>
+      expect(screen.getByText('In this collection · 1')).toBeInTheDocument()
+    )
+  })
+
+  it('removes a member right away', async () => {
+    render(
+      <CollectionEditor
+        mode="edit"
+        collection={collection}
+        initialMembers={[member]}
+        followingSuggestions={[]}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Remove Rin' }))
+
+    await waitFor(() =>
+      expect(removeCollectionAccounts).toHaveBeenCalledWith({
+        collectionId: 'col-1',
+        accountIds: [member.id]
+      })
+    )
+    await waitFor(() =>
+      expect(
+        screen.queryByText('In this collection · 1')
+      ).not.toBeInTheDocument()
+    )
+  })
+
+  it('saves changes and routes back to the collection', async () => {
+    render(
+      <CollectionEditor
+        mode="edit"
+        collection={collection}
+        initialMembers={[member]}
+      />
+    )
+
+    fireEvent.change(screen.getByLabelText('Collection name'), {
+      target: { value: 'Renamed' }
+    })
+    fireEvent.click(screen.getByRole('button', { name: 'Save changes' }))
+
+    await waitFor(() =>
+      expect(updateCollection).toHaveBeenCalledWith({
+        collectionId: 'col-1',
+        title: 'Renamed',
+        description: 'who I read',
+        topic: 'fediverse',
+        visibility: 'public',
+        feedEnabled: true
+      })
+    )
+    expect(mockPush).toHaveBeenCalledWith('/collections/col-1')
+  })
+})
--- a/app/(timeline)/collections/CollectionEditor.test.tsx
+++ b/app/(timeline)/collections/CollectionEditor.test.tsx
@@ -143,6 +143,25 @@ describe('CollectionEditor', () => {
     )
   })
 
+  it('shows an inline error and does not add the member when the add fails', async () => {
+    ;(addCollectionAccounts as jest.Mock).mockResolvedValue(false)
+    render(
+      <CollectionEditor
+        mode="edit"
+        collection={collection}
+        initialMembers={[]}
+        followingSuggestions={[suggestion]}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: 'Add' }))
+
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Could not add that account. Please try again.'
+    )
+    expect(screen.queryByText('In this collection · 1')).not.toBeInTheDocument()
+  })
+
   it('removes a member right away', async () => {
     render(
       <CollectionEditor
--- a/app/(timeline)/collections/[id]/CollectionDetail.test.tsx
+++ b/app/(timeline)/collections/[id]/CollectionDetail.test.tsx
@@ -1,0 +1,176 @@
+/**
+ * @vitest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { ReactNode } from 'react'
+
+import { CollectionMember } from '@/app/(timeline)/collections/CollectionEditor'
+import { getCollectionFeed, getCollectionTimeline } from '@/lib/client'
+import { ActorProfile } from '@/lib/types/domain/actor'
+import { Status } from '@/lib/types/domain/status'
+import { CollectionEntity } from '@/lib/types/mastodon/collection'
+
+import { CollectionDetail } from './CollectionDetail'
+
+vi.mock('@/lib/client', () => ({
+  getCollectionFeed: vi.fn(),
+  getCollectionTimeline: vi.fn()
+}))
+
+vi.mock('@/lib/components/page-header', () => ({
+  PageHeader: ({
+    title,
+    description,
+    actions
+  }: {
+    title: ReactNode
+    description: ReactNode
+    actions: ReactNode
+  }) => (
+    <div>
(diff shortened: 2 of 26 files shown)


app/(timeline)/collections/[id]/CollectionDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+  const fetchPage = useCallback(
+    (next: Projection, maxStatusId?: string) =>
+      next === 'public'
+        ? getCollectionFeed({ collectionId: collection.id, maxStatusId })
+        : getCollectionTimeline({ collectionId: collection.id, maxStatusId }),
+    [collection.id]
+  )
+
+  const switchProjection = async (next: Projection) => {
```
Comment (llun) [inline app/(timeline)/collections/[id]/CollectionDetail.tsx:159]: **[major] Race condition in `switchProjection` / `loadMoreStatuses`.**

Both async paths mutate the same state (`currentStatuses`, `lastStatusIdRef`, `hasMoreStatuses`) with no request token, and `switchProjection` never checks `isLoadingRef`. If a `loadMore` (on the old projection) or another toggle is in flight when the user toggles, requests can resolve out of order — appending owner-projection posts onto the public feed, or leaving the displayed feed disagreeing with `projection` (which drives the roster + the next `loadMore` cursor).

Fix: capture a monotonic request token at the start of each async path and only apply state when it is still the latest.

app/(timeline)/collections/[id]/CollectionDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+      {/* Roster — highlighted accounts. */}
+      {roster.length > 0 && (
+        <section className="overflow-hidden rounded-xl border bg-card shadow-sm">
+          <div className="flex items-center justify-between border-b px-4 py-2.5">
+            <span className="text-xs font-semibold">
+              Highlighted accounts · {roster.length}
+            </span>
+            {projection === 'public' && approvedCount < totalCount && (
+              <span className="text-xs text-muted-foreground">
+                {totalCount - approvedCount} hidden by consent
```
Comment (llun) [inline app/(timeline)/collections/[id]/CollectionDetail.tsx]: **[minor] Privacy: the “N hidden by consent” count is exposed to anonymous/non-owner viewers.**

For non-owners `projection` is always `public`, and `page.tsx` passes `totalCount`/`approvedCount` unconditionally, so the public roster header renders `totalCount - approvedCount` — the number of members who have *not* approved being featured. The public projection should reveal only approved members; this leaks the private-membership delta.

Fix: gate the “hidden by consent” hint behind `isOwner` (it is only meaningful in the owner’s Public preview).

app/(timeline)/collections/[id]/page.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+// A non-owner may view a collection only when it has a public projection: a
+// public/unlisted visibility AND an enabled feed. Private (or feed-disabled)
+// collections are owner-only.
+const hasPublicProjection = (collection: {
+  visibility: 'public' | 'unlisted' | 'private'
+  publicFeed: boolean
+}) => collection.visibility !== 'private' && collection.publicFeed
+
+export const generateMetadata = async ({
```
Comment (llun) [inline app/(timeline)/collections/[id]/page.tsx:37]: **[minor] `generateMetadata` discloses a private collection’s title.**

It resolves the collection via the non-owner-scoped `getCollectionById` and sets the page `<title>` for *any* id — including private/feed-disabled collections — with no gating, whereas the page body correctly `notFound()`s non-owners. Relying on Next replacing metadata on `notFound()` is fragile.

Fix: gate the real title behind `isOwner || hasPublicProjection(collection)`; otherwise fall back to the generic title.

app/(timeline)/collections/[id]/edit/page.tsx
```diff
@@ -0,0 +1,106 @@
...
+const FOLLOWING_SUGGESTIONS_LIMIT = 200
+// Page size + safety cap for loading the full member roster.
+const MEMBER_PAGE_LIMIT = 80
+const MAX_MEMBER_PAGES = 25
+
+interface PageProps {
+  params: Promise<{ id: string }>
+}
+
+const toMember = (
```
Comment (llun) [inline app/(timeline)/collections/[id]/edit/page.tsx]: **[nit] `toMember` is duplicated byte-for-byte** in `collections/[id]/page.tsx` and `collections/[id]/edit/page.tsx`.

Extract a shared `toCollectionMember(account, host)` helper and import it in both to keep the handle-building logic in one place.

app/(timeline)/collections/[id]/CollectionDetail.tsx
```diff
@@ -0,0 +1,410 @@
...
+          ? result.statuses[result.statuses.length - 1].id
+          : null
+      setHasMoreStatuses(Boolean(result.nextMaxStatusId))
+    } catch {
+      // Leave the previous feed in place on error; the toggle can be retried.
+    } finally {
+      isLoadingRef.current = false
+      setLoadingMoreStatuses(false)
+    }
+  }
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.tsx:188]: Fixed in f78b97799. Added a monotonic `requestIdRef` token captured at the start of both `switchProjection` and `loadMoreStatuses`; results (and the loading-flag reset in `finally`) are only applied when the token is still the latest, so a stale/out-of-order response can no longer apply the wrong projection or desync the cursor. Thanks for the catch.

app/(timeline)/collections/[id]/CollectionDetail.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.tsx:159]: Fixed in f78b97799. Added a monotonic `requestIdRef` token captured at the start of both `switchProjection` and `loadMoreStatuses`; results (and the loading-flag reset in `finally`) are only applied when the token is still the latest, so a stale/out-of-order response can no longer apply the wrong projection or desync the cursor. Thanks for the catch.

app/(timeline)/collections/[id]/CollectionDetail.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.tsx]: Fixed in f78b97799. The "N hidden by consent" hint is now gated behind `isOwner`, so the non-consenting-member count is only ever shown in the owner's Public preview and never to public/anonymous viewers.

app/(timeline)/collections/[id]/page.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/page.tsx:37]: Fixed in f78b97799. `generateMetadata` now resolves the viewer and only exposes the real title when `isOwner || hasPublicProjection(collection)`, mirroring the page's own `notFound()` gate; otherwise it falls back to the generic title.

app/(timeline)/collections/[id]/edit/page.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/edit/page.tsx]: Fixed in f78b97799. Extracted a shared `toCollectionMember(account, host)` helper (`app/(timeline)/collections/toCollectionMember.ts`) now imported by both the detail and edit pages.

app/(timeline)/collections/[id]/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+    : null
+
+  return (
+    <CollectionDetail
+      host={host}
+      collection={getMastodonCollection(collection, approvedCount)}
+      isOwner={isOwner}
+      ownerHandle={ownerHandle}
+      ownerProfilePath={ownerProfilePath}
+      totalCount={totalCount}
```
Comment (llun) [inline app/(timeline)/collections/[id]/page.tsx]: **[minor] Privacy regression: the raw `totalCount` is still serialized to non-owner clients.**

The "N hidden by consent" hint was gated behind `isOwner` in the UI, but `page.tsx` still passes `totalCount` (every member, incl. non-consenting) unconditionally into `CollectionDetail` — a `'use client'` component — so it lands in the RSC/HTML payload any anonymous visitor can read. Combined with `approvedCount`, a non-owner can recompute `totalCount - approvedCount` (the exact hidden-by-consent figure) from the page source.

Fix: `totalCount={isOwner ? totalCount : approvedCount}` (non-owner paths only use it as a `=== 0` empty-state check).

app/(timeline)/notifications/components/CollectionConsentNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  return (
+    <div className="mt-1.5 space-y-2">
+      <p className="truncate text-[13px] font-medium text-foreground">
+        {collectionTitle}
+      </p>
+      <div className="flex flex-wrap items-center gap-2">
+        <Button
```
llun, line 63: **[minor] a11y: the consent toggle buttons have no `aria-pressed`.**

After a choice, both "Show me publicly" / "Keep me hidden" stay rendered and the active one is signaled only visually via the Button `variant`. Assistive tech gets no programmatic indication of the current selection. The analogous persistent toggle in `CollectionDetail` already sets `aria-pressed`.

Fix: add `aria-pressed={state === 'approved'}` / `aria-pressed={state === 'revoked'}` to the two buttons.

app/(timeline)/collections/[id]/CollectionDetail.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  it('switches to the public preview, fetching the public feed and approved roster', async () => {
+    render(
+      <CollectionDetail
+        {...baseProps}
+        isOwner
+        currentActor={{} as ActorProfile}
+      />
+    )
+
+    fireEvent.click(screen.getByRole('button', { name: /public preview/i }))
```
Comment (llun) [inline app/(timeline)/collections/[id]/CollectionDetail.test.tsx:152]: **[minor] Test gap: the projection-toggle test never asserts the feed actually swaps.**

`'switches to the public preview'` only asserts `getCollectionFeed` was *called* and that the roster swaps — but the roster swap is a synchronous prop-derived switch independent of the fetch. Whether `switchProjection` applies `result.statuses` is never checked, and both feed mocks return same-length payloads so a `Posts` length assertion couldn't distinguish them either.

Fix: make the two feed mocks return distinguishable ids, render them in the `Posts` mock, and assert the public feed id appears after toggling.

app/(timeline)/collections/[id]/CollectionDetail.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+  Posts: ({ statuses }: { statuses: Status[] }) => (
+    <div data-testid="posts">{statuses.length} posts</div>
+  )
+}))
+
+vi.mock('@/lib/components/scroll-to-top-button', () => ({
+  ScrollToTopButton: () => null
+}))
+
+vi.mock('@/lib/components/posts/useLoadMoreOnVisible', () => ({
```
Comment (llun) [inline app/(timeline)/collections/[id]/CollectionDetail.test.tsx:51]: **[minor] Test gap: `loadMoreStatuses` and the `requestIdRef` stale-response guard have zero coverage.**

The newly-added load-more path and the request-token guard (the round-1 race fix) aren't exercised by any test. Add coverage for appending a second page via load-more, and ideally a stale-response case proving an out-of-order resolve is ignored.

app/(timeline)/notifications/components/CollectionConsentNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      const ok =
+        action === 'approve'
+          ? await approveCollectionMembership({ collectionId, accountId })
+          : await revokeCollectionMembership({ collectionId, accountId })
+      if (!ok) throw new Error('request failed')
+      setState(action === 'approve' ? 'approved' : 'revoked')
+      router.refresh()
+    } catch {
+      setError(
```
llun: **[nit] Redundant ternary in `CollectionConsentNotification.respond`** — both branches of `setError(action === 'approve' ? '…' : '…')` return the identical string. Collapse to a single string.

app/(timeline)/collections/[id]/page.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/page.tsx]: Fixed in 6577795b5. `page.tsx` now passes `totalCount={isOwner ? totalCount : approvedCount}`, so the raw total (and thus the hidden-by-consent delta) is no longer serialized into the non-owner client payload. Non-owner code paths only use it as a `=== 0` empty-state check.

app/(timeline)/notifications/components/CollectionConsentNotification.tsx
llun: Fixed in 6577795b5. Added `aria-pressed={state === ...}` to both consent buttons, matching the toggle pattern in `CollectionDetail`. Covered by a new assertion in the component test.

app/(timeline)/collections/[id]/CollectionDetail.test.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.test.tsx:152]: Fixed in 6577795b5. The feed mocks now return distinguishable ids rendered by the `Posts` mock, and the test asserts the public feed id appears (and the owner id disappears) after the toggle — so dropping/mis-applying the result would now fail.

app/(timeline)/collections/[id]/CollectionDetail.test.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.test.tsx:51]: Fixed in 6577795b5. Added two tests: one that appends the next page via load-more (asserting the cursor + appended id), and one that holds the owner load-more open across a projection switch and proves the stale response is dropped by the requestId guard.

app/(timeline)/notifications/components/CollectionConsentNotification.tsx
llun: Fixed in 6577795b5. Collapsed the ternary to a single `setError('Could not update your choice. Please try again.')`.

app/(timeline)/collections/[id]/CollectionDetail.tsx
```diff
@@ -0,0 +1,428 @@
...
+
+  const fetchPage = useCallback(
+    (next: Projection, maxStatusId?: string) =>
+      next === 'public'
+        ? getCollectionFeed({ collectionId: collection.id, maxStatusId })
+        : getCollectionTimeline({ collectionId: collection.id, maxStatusId }),
+    [collection.id]
+  )
+
+  const switchProjection = async (next: Projection) => {
```
Comment (llun) [inline app/(timeline)/collections/[id]/CollectionDetail.tsx:159]: **[minor] A failed projection switch leaves the feed/cursor/roster out of sync with the toggle.**

`switchProjection` calls `setProjection(next)` *before* awaiting `fetchPage(next)`. On a fetch error the `catch` keeps the old feed, but `projection` has already flipped — so the roster + toggle highlight switch to the new projection while `currentStatuses`/`lastStatusIdRef` still hold the old one, and a subsequent load-more paginates the new projection from an old-projection cursor. (The `requestId` guard only covers stale *successful* responses, not a thrown switch.)

Fix: capture `const previous = projection` and `setProjection(previous)` in the `catch` (or only `setProjection(next)` after the fetch succeeds).

app/(timeline)/notifications/NotificationItem.tsx
```diff
@@ -117,12 +125,24 @@ export const NotificationItem = ({
...
-        ) : (
-          <FollowRequestNotification account={acc} />
-        )
+      if (notification.type === 'follow') {
+        body = <FollowNotification account={acc} />
+      } else if (notification.type === 'follow_request') {
+        body = <FollowRequestNotification account={acc} />
+      } else if (notification.type === 'added_to_collection') {
+        // The consent gate: let the member choose whether to appear on the
+        // collection's public link. Needs the collection (from the groupKey)
```
llun, line 134: **[minor] The new `added_to_collection` / `collection_update` branch in `NotificationItem` has no test.**

`CollectionConsentNotification` is tested in isolation, but the wiring in `NotificationItem` — mount the consent UI only when `collection && currentAccountId`, fall back to null otherwise, and treat `collection_update` as verb-only — is untested in the existing (thorough) `NotificationItem.test.tsx`. A regression dropping the `currentAccountId` guard or rendering consent UI for `collection_update` wouldn't be caught.

Fix: add cases for `added_to_collection` (with and without `collection`/`currentAccountId`) and `collection_update`.

lib/client.test.ts
```diff
@@ -895,3 +904,154 @@ describe('client trends', () => {
...
+      expect.objectContaining({ method: 'POST' })
+    )
+    expect(fetchMock).toHaveBeenNthCalledWith(
+      2,
+      '/api/v1/collections/c1/items/me/revoke',
+      expect.objectContaining({ method: 'POST' })
+    )
+  })
+
+  it('loads the owner timeline and the public feed in the activities_next shape', async () => {
```
llun, line 1029: **[minor] The collection timeline/feed client helpers never exercise the `urlToId`-encoded cursor params.**

`getCollectionTimeline`/`getCollectionFeed` apply `urlToId()` to `maxStatusId`/`minStatusId`, but the only test passes just `limit`, so the cursor-encoding branch is untested — a regression forwarding the raw id would silently break load-more without failing any test (cf. the `getBookmarks` test which covers `max_id`/`min_id`).

Fix: call the helpers with a full status URL as `maxStatusId`/`minStatusId` and assert the request URL contains `max_id`/`min_id` equal to `urlToId(...)`.

app/(timeline)/collections/CollectionEditor.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+      )
+    )
+  })
+
+  it('does not render the people section in create mode', () => {
+    render(<CollectionEditor mode="create" />)
+    expect(screen.queryByText('People')).not.toBeInTheDocument()
+  })
+
+  it('adds a suggested account right away', async () => {
```
llun, line 123: **[nit] `CollectionEditor`'s add/remove failure path (request resolves `false` → inline error) is untested.**

The success and the create/save paths are covered, but the `if (!ok) setError(...)` branch on add/remove isn't. Add a case where `addCollectionAccounts` resolves `false` and assert the inline error appears and the member is not added.

app/(timeline)/collections/[id]/CollectionDetail.tsx
Comment (llun) [reply app/(timeline)/collections/[id]/CollectionDetail.tsx:159]: Fixed in d91ab2e3d. On a failed switch, `switchProjection` now reverts `projection` to the previous value in the `catch` (guarded by the requestId so it does not clobber a newer switch), keeping the toggle, roster, feed and cursor consistent.

app/(timeline)/notifications/NotificationItem.tsx
llun: Fixed in d91ab2e3d. Added `NotificationItem` tests: consent actions render for `added_to_collection` with a collection + currentAccountId; only the verb shows when either is missing; `collection_update` is informational (no consent actions).

lib/client.test.ts
llun: Fixed in d91ab2e3d. Added a client test that calls both helpers with a full status URL as the cursor and asserts the request `max_id`/`min_id` equal `urlToId(...)` (parsing the URL so it is decoding/order-agnostic).

app/(timeline)/collections/CollectionEditor.test.tsx
llun: Fixed in d91ab2e3d. Added a `CollectionEditor` test where `addCollectionAccounts` resolves `false`, asserting the inline error appears and the member is not added.

app/(timeline)/collections/CollectionEditor.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks()
+    ;(createCollection as jest.Mock).mockResolvedValue({
+      ...collection,
+      id: 'new-col'
+    })
+    ;(updateCollection as jest.Mock).mockResolvedValue(collection)
+    ;(addCollectionAccounts as jest.Mock).mockResolvedValue(true)
+    ;(removeCollectionAccounts as jest.Mock).mockResolvedValue(true)
+    ;(deleteCollection as jest.Mock).mockResolvedValue(true)
```
llun, line 65: **[minor] The delete-collection flow is entirely untested.**

`deleteCollection` is mocked in `beforeEach`, but no test clicks "Delete collection", so `handleDelete` (the `window.confirm` gate, the success route to `/lists` + `router.refresh()`, and the failure inline error) has zero assertions — the only destructive action in the editor.

Fix: add tests for confirm→delete→`router.push('/lists')`, confirm-false→no delete call, and delete-fails→inline error.

app/(timeline)/collections/CollectionEditor.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+    fireEvent.click(screen.getByRole('button', { name: 'Add' }))
+
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Could not add that account. Please try again.'
+    )
+    expect(screen.queryByText('In this collection · 1')).not.toBeInTheDocument()
+  })
+
+  it('removes a member right away', async () => {
```
llun, line 165: **[nit] The `removeMember` failure path is untested** (add-failure is covered, remove-failure isn't).

Mirror the add-failure test: mock `removeCollectionAccounts` to resolve `false`, click "Remove Rin", and assert the "Could not remove that account…" alert shows and the member stays in the list.

app/(timeline)/notifications/components/CollectionConsentNotification.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    await waitFor(() =>
+      expect(revokeCollectionMembership).toHaveBeenCalledWith({
+        collectionId: 'col-1',
+        accountId: 'me'
+      })
+    )
+    expect(await screen.findByText('Hidden')).toBeInTheDocument()
+  })
+
+  it('surfaces an inline error when the request fails', async () => {
```
llun, line 80: **[nit] The consent error test asserts the alert but not that the choice stayed unmade.**

After a failed `approve`, also assert "Featured publicly" is absent and the "Show me publicly" button keeps `aria-pressed="false"`, so a regression that optimistically set the state before the failure check would be caught.

app/(timeline)/collections/CollectionEditor.test.tsx
llun: Fixed in 3a4863cc7. Added three CollectionEditor delete tests: confirm → `deleteCollection('col-1')` + `router.push('/lists')`; confirm dismissed → no delete/navigation; delete fails → the "Could not delete the collection…" inline error (and no navigation).

app/(timeline)/collections/CollectionEditor.test.tsx
llun: Fixed in 3a4863cc7. Added a remove-failure test mirroring the add-failure one: `removeCollectionAccounts` resolves false → the "Could not remove that account…" alert shows and the member stays in the list (not optimistically dropped).

app/(timeline)/notifications/components/CollectionConsentNotification.test.tsx
llun: Fixed in 3a4863cc7. The consent error test now also asserts "Featured publicly" is absent and the "Show me publicly" button keeps `aria-pressed="false"`, pinning that a failed request leaves the choice unmade.

PR #1119, opened 2026-06-27
fix: keep alsoKnownAs an array so single-alias remote actors don't 404

llun:
## Problem

After #1116 fixed the signing, a logged-in visit to a secure-mode remote profile (e.g. `https://llun.social/@gemeenteamsterdam@social.amsterdam.nl`) **still returned 404** — but now for a different reason. Production logs showed:

```
[getActorProfile] [
  { "expected": "array", "code": "invalid_type", "path": ["alsoKnownAs"],
    "message": "Invalid input: expected array, received string" }
]
```

The signed actor fetch now succeeds (200), but `Actor.safeParse` rejects the document, so `getActorPerson` returns `null` → `getProfileData` returns `null` → `notFound()`.

## Root cause

Mastodon emits `alsoKnownAs` as a **set** of migration-alias ids. `gemeenteamsterdam` has exactly **one** alias, and JSON-LD compaction collapses a single-element array to a scalar:

- `CANONICAL_CONTEXT` defined `alsoKnownAs` as `{ @id, @type: @id }` — **without** `@container: @set`, unlike `to`/`cc`/`tag`/`attachment`.
- So `alsoKnownAs: ["https://…"]` compacts to `alsoKnownAs: "https://…"` (a string).
- The `Actor` schema required `z.array(z.string())`, so the whole actor failed validation.

This only bit actors with a **single** alias — multi-alias actors stay arrays, which is why most remote profiles worked.

### Reproduced

```
single-element array → compacted "https://…" (string) → parse FAIL: expected array, received string
two-element array    → compacted [...]          (array)  → parse OK
bare string          → compacted "https://…"    (string) → parse FAIL
```

## Fix

- **`CANONICAL_CONTEXT`:** add `@container: @set` to `alsoKnownAs` so compaction always yields an array (matching the other collection-like properties — the documented "always arrays" pattern).
- **`Actor` schema:** make it liberal — accept a bare string **or** an array and normalise to an array, as defense for the raw-input compaction fallback and quirky peers. Output type stays `string[] | undefined`; the field stays optional.

After the fix all three shapes compact to an array and the actor parses.

## Changes

- `lib/activities/jsonld/index.ts` — `@container: @set` for `alsoKnownAs`.
- `lib/types/activitypub/actor.ts` — liberal + normalising `alsoKnownAs`.
- `lib/activities/jsonld/index.test.ts` — single-alias compaction stays an array.
- `lib/types/activitypub/actor.test.ts` — actor parses with `alsoKnownAs` as string or array (new file).

## Testing

- `yarn lint` ✅ · `yarn build` ✅ · `yarn test` ✅ (566 files / 5012 tests)

### DIFF
--- a/lib/activities/jsonld/index.test.ts
+++ b/lib/activities/jsonld/index.test.ts
@@ -59,6 +59,23 @@ describe('compactActivityPub', () => {
     expect(result.cc).toEqual(['https://remote.example/users/alice/followers'])
   })
 
+  it('forces alsoKnownAs to an array even for a single migration alias', async () => {
+    // Mastodon emits `alsoKnownAs` as a set of ids. With only one alias the
+    // value compacts to a scalar string unless the canonical context marks it
+    // `@container: @set`, which then fails the actor schema's `string[]`.
+    const result = asRecord(
+      await compactActivityPub({
+        '@context': [ACTIVITY_STREAMS_CONTEXT_URL, SECURITY_V1_CONTEXT_URL],
+        id: 'https://remote.example/users/alice',
+        type: 'Person',
+        preferredUsername: 'alice',
+        alsoKnownAs: ['https://old.example/users/alice']
+      })
+    )
+
+    expect(result.alsoKnownAs).toEqual(['https://old.example/users/alice'])
+  })
+
   it('keeps an embedded object but collapses a bare reference', async () => {
     const create = asRecord(
       await compactActivityPub({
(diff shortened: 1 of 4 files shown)


llun:
### ✅ Review round 1 — converged (in-scope)

- **CI:** 12/12 green. **Kilo:** *No Issues Found · Merge.*
- **3-lens adversarial sub-agent review** (correctness / completeness / tests): **0 in-scope findings.** The Zod `transform`-before-`optional()` reorder is correct (field stays optional, output type `string[] | undefined`); the single-element-array trigger is covered at both the compaction and schema layers.

#### Out-of-scope follow-up (same bug class, tracked separately)
The completeness lens found one **pre-existing** latent instance of the same single-element-array→scalar collapse: `Question.oneOf` / `Question.anyOf` (poll options) are declared `QuestionOption.array()` and are **not** forced to `@container: @set` in the canonical context. A single-option poll from a non-Mastodon peer would fail `Question.safeParse` with `oneOf: expected array, received object` (reproduced empirically). Mastodon mandates ≥2 options so it's rare, but it's the same gap. Keeping this PR surgical (profile 404 only) — fixing `oneOf`/`anyOf` in a dedicated follow-up PR.

PR #1120, opened 2026-06-27
fix: keep poll options an array so single-option remote polls parse

llun:
## Problem

Same bug class as #1119 (the `alsoKnownAs` profile-404 fix), surfaced by **#1119's own review**: a remote poll with a **single option** silently fails to import.

`Question.oneOf` / `Question.anyOf` are declared `QuestionOption.array()`, but the canonical JSON-LD context did **not** force them to a set (unlike `to`/`cc`/`tag`/`attachment`). So a one-element `oneOf` array compacts to a bare object, and `Question.safeParse` fails:

```
oneOf: Invalid input: expected array, received object
```

`createPollJob` / `updatePollJob` then drop the inbound poll. Mastodon mandates ≥2 options so it's rare across the fediverse, but the AP spec doesn't forbid a single-option poll and other servers/relays can emit one. (Confirmed empirically by compacting a single-option `Question`.)

## Fix

Mirrors #1119's two-pronged approach:

- **`CANONICAL_CONTEXT`:** add `@container: @set` to `oneOf`/`anyOf` (object-valued, so **no** `@type: @id`) → compaction always yields an array, matching `tag`/`attachment`.
- **`Question` schema:** make `oneOf`/`anyOf` liberal — accept a single `QuestionOption` **or** an array and normalise to an array, as defense for the raw-input fallback. Output type stays `QuestionOption[] | undefined`; fields stay optional.

## Changes

- `lib/activities/jsonld/index.ts` — `@container: @set` for `oneOf`/`anyOf`.
- `lib/types/activitypub/objects.ts` — liberal + normalising `oneOf`/`anyOf`.
- `lib/activities/jsonld/index.test.ts` — single-option compaction stays an array.
- `lib/types/activitypub/objects.test.ts` — `Question` parses with `oneOf`/`anyOf` as a single object or array (new file).

## Scope

Independent of #1119 (no overlapping lines beyond different entries in the same context block). `git diff` is exactly these 4 files — no `alsoKnownAs` / actor changes.

## Testing

- `yarn lint` ✅ · `yarn build` ✅ · `yarn test` ✅ (566 files / 5013 tests)
- Existing `createPollJob` / `updatePollJob` suites still green.

### DIFF
--- a/lib/activities/jsonld/index.test.ts
+++ b/lib/activities/jsonld/index.test.ts
@@ -76,33 +76,36 @@ describe('compactActivityPub', () => {
     expect(result.alsoKnownAs).toEqual(['https://old.example/users/alice'])
   })
 
-  it('forces poll options to arrays even for a single option', async () => {
-    // A single-option poll would otherwise compact `oneOf` to a bare object and
-    // fail the Question schema's `QuestionOption[]`. `@container: @set` keeps it
-    // an array.
-    const result = asRecord(
-      await compactActivityPub({
+  // Both poll-option fields are forced to arrays by separate context overrides,
+  // so guard both at the compaction layer (the layer that produces the
+  // bare-object-vs-array shape the bug was about).
+  it.each([{ field: 'oneOf' }, { field: 'anyOf' }])(
+    'forces poll options ($field) to an array even for a single option',
+    async ({ field }) => {
+      const input: Record<string, unknown> = {
         '@context': ACTIVITY_STREAMS_CONTEXT_URL,
         id: 'https://remote.example/polls/1',
         type: 'Question',
         attributedTo: 'https://remote.example/users/alice',
         published: '2026-01-01T00:00:00Z',
-        endTime: '2026-01-02T00:00:00Z',
-        oneOf: [
-          {
-            type: 'Note',
-            name: 'Only choice',
-            replies: { type: 'Collection', totalItems: 0 }
-          }
-        ]
-      })
-    )
+        endTime: '2026-01-02T00:00:00Z'
+      }
+      input[field] = [
+        {
+          type: 'Note',
+          name: 'Only choice',
+          replies: { type: 'Collection', totalItems: 0 }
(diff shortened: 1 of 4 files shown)


llun:
### 🤖 Review round 1

- **CI:** 12/12 green. **Kilo:** *No Issues Found · Merge.*
- **3-lens adversarial review** (correctness / completeness-sweep / tests):

| Finding | Verdict | Action |
|---------|---------|--------|
| Compaction test covered only `oneOf`, not `anyOf` (separate context override) | nit, in-scope | ✅ Fixed — test now table-driven over both fields |
| By-reference poll option (`oneOf: "<id>"`) still fails to parse | nit, **not a regression** (failed on `main` too; embedded options are universal in the wild) | Out of scope — noted only |
| The "array" rows in the schema test are unchanged-behaviour baselines, not fix-guarding | nit | Intentional regression baselines; the single-object rows are the load-bearing ones — left as-is |

**Completeness sweep:** the review confirmed `oneOf`/`anyOf` was the **last** strictly-array field vulnerable to the single-element compaction collapse (actor siblings use `z.union`; `to`/`cc`/`tag`/`attachment` are already `@set`; WebFinger arrays don't go through compaction). Starting round 2 to confirm convergence.

llun:
### ✅ Review converged (2 rounds)

- **Round 1:** added `anyOf` compaction-layer coverage; by-reference options ruled out-of-scope (not a regression); array-baseline rows kept.
- **Round 2:** 0 findings across correctness + a final completeness re-sweep.
- **CI:** 12/12 green. **Kilo:** *No Issues Found · Merge.*

Completeness sweep confirms `oneOf`/`anyOf` was the **last** strictly-array ActivityPub field vulnerable to the single-element compaction collapse (`alsoKnownAs` is fixed in #1119; actor siblings use `z.union`; `to`/`cc`/`tag`/`attachment` are already `@set`; WebFinger arrays don't go through compaction). Ready to merge.

PR #1139, opened 2026-06-29
fix: resume OAuth login for already-authenticated users at the sign-in page

llun:
## Summary

Logging in with **Phanpy / web clients while already signed into the instance** dropped the user on the **home timeline** instead of redirecting back to the client. [#1135](https://github.com/llun/activities.next/pull/1135) fixed the same "OAuth request dropped → home" symptom for *fresh* logins by teaching the sign-in **forms** to resume the request (`resolveSignInRedirect`), but it left the matching gap on the sign-in **page** server component: it still did `if (session) redirect('/')`, so an **already-authenticated** visitor who lands on `/auth/signin` mid-flow lost the request.

## Why this hits Phanpy/web clients but not OIDC (docs.llun.dev)

- A logged-in user starting an OAuth login can still be bounced to `/auth/signin` with the authorization request — better-auth redirects a no-session/`prompt` authorize to its `loginPage`, and `/oauth/authorize` redirects an actor-less session there too. The page then sent them **home**, dropping the request.
- OIDC relying parties (e.g. la-suite **Docs**) hit better-auth's `/api/auth/oauth2/authorize` **directly**, bypassing this page, so they were unaffected — matching the report that *docs works but Phanpy doesn't*.

## Fix

`app/(nosidebar)/auth/signin/page.tsx` now forwards an **authenticated** visitor through the **same** `resolveSignInRedirect` the forms use, instead of always going home:

- Handles both sign-in URL shapes — `redirectBack=…` (custom `/oauth/authorize` handoff) and the bare `response_type=code&client_id=…` envelope better-auth's `loginPage` produces.
- Guarded by `getActorFromSession`: a session **without a usable actor** still goes to `/`, because resuming to `/oauth/authorize` without an actor would bounce straight back to `/auth/signin` and loop.
- Plain logins (no resume target) keep the existing `redirect('/')` behavior.

Additive and low-risk: no changes to `auth.ts`, the OIDC discovery, `resolveSignInRedirect`, or the consent page. The page only ever redirects to an internal path; `redirect_uri` is carried as an opaque query value and validated downstream by the consent page and better-auth.

## Testing

- New `app/(nosidebar)/auth/signin/page.test.tsx`: authenticated user → resumes to `/oauth/authorize` (OIDC params preserved incl. PKCE), resumes a safe `redirectBack`, falls back to `/` with no resume target, falls back to `/` (no loop) when the session has no actor, and renders the form for a logged-out visitor.
- `yarn run prettier --write .` ✓ · `yarn lint` ✓ (0 errors) · `yarn build` ✓ · `yarn test` ✓ (5186)
- **End-to-end on the real `activities.local` + `phanpy.local` Docker stack** (PostgreSQL, HTTPS via Traefik): with the fix, an already-authenticated user starting a Phanpy login is redirected **back to Phanpy** instead of the home timeline, for both sign-in URL shapes.

## Follow-ups (out of scope, noted while reproducing — not in this PR)

- The OAuth/OIDC server metadata (`getOAuthAuthorizationServerMetadata` / `getOpenIDConfiguration`) and `app/api/oauth/authorize/route.ts` hardcode `getBaseURL()` (the canonical host), so a client on a trusted alias/served domain is advertised the canonical host — they should be request-host-aware like the `/oauth/authorize` page.
- `/oauth/authorize` 500s if `getServerAuthSession()` throws (e.g. a stale pre-#1040 `jwks` key → `ERR_JOSE_NOT_SUPPORTED`), which is asymmetric vs OIDC.

### DIFF
--- a/app/(nosidebar)/auth/signin/page.test.tsx
+++ b/app/(nosidebar)/auth/signin/page.test.tsx
@@ -1,0 +1,124 @@
+import { getBaseURL, getConfig } from '@/lib/config'
+import { getDatabase } from '@/lib/database'
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { getActorFromSession } from '@/lib/utils/getActorFromSession'
+
+import Page from './page'
+
+vi.mock('@/lib/config')
+
+vi.mock('@/lib/database', () => ({
+  getDatabase: vi.fn(() => ({}))
+}))
+
+vi.mock('@/lib/services/auth/getSession', () => ({
+  getServerAuthSession: vi.fn()
+}))
+
+vi.mock('@/lib/utils/getActorFromSession', () => ({
+  getActorFromSession: vi.fn()
+}))
+
+// The sign-in forms are client components that pull in the better-auth client;
+// stub them so importing the server page stays light. The logged-in redirect
+// branch never renders them anyway.
+vi.mock('./CredentialForm', () => ({ CredentialForm: () => null }))
+vi.mock('./PasskeySigninButton', () => ({ PasskeySigninButton: () => null }))
+
+const redirectMock = vi.fn((path: string) => path)
+vi.mock('next/navigation', () => ({
+  redirect: (path: string) => redirectMock(path)
+}))
+
+const aSession = { user: { email: 'rider@example.com' } } as Awaited<
+  ReturnType<typeof getServerAuthSession>
+>
+const anActor = { id: 'actor-id' } as Awaited<
+  ReturnType<typeof getActorFromSession>
+>
+
+describe('/auth/signin already-authenticated resume', () => {
+  beforeEach(() => {
+    redirectMock.mockClear()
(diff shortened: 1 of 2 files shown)


llun:
### ✅ Review loop converged (2 rounds)

- **Gemini Code Assist:** 3 reviews, all "no review comments to address".
- **Kilo Code Review:** No Issues Found · Recommendation: Merge.
- **Inline review threads:** 0 (nothing to resolve).
- **Checks:** All Tests, Build, Lint and Prettier, CodeQL/Analyze, CI Success — all green. `MERGEABLE` / `CLEAN`.

No code changes were required during review. The fix (resume the OAuth/OIDC request for already-authenticated visitors at the sign-in page, instead of `redirect('/')`) was reproduced and verified end-to-end against the local `activities.local` + `phanpy.local` stack: a logged-in user starting a Phanpy login now lands back at the client instead of the home timeline.

PR #1141, opened 2026-06-29
fix: advertise the request host in OAuth/OIDC metadata for multi-domain deployments

llun:
## Summary

On a multi-domain deployment (`ACTIVITIES_HOST` + `ACTIVITIES_TRUSTED_HOSTS`), the OAuth/OIDC discovery documents and the `/api/oauth/authorize` redirect were built from `getBaseURL()` — the **canonical** host — so a client discovering metadata on a **trusted alias/served domain** was advertised, and bounced to, the canonical host mid-flow.

Fetched from a served domain (`alias.local`), before this PR:

```
/.well-known/oauth-authorization-server  → issuer/authorization_endpoint/token_endpoint = https://activities.local/…
/.well-known/openid-configuration        → issuer = https://activities.local/api/auth
/api/oauth/authorize                     → 302 → https://activities.local/api/auth/oauth2/authorize
```

…even though `/api/v1/instance` on that host reports `uri: alias.local`.

## Why it's a bug

The rest of the auth stack is already per-request host-aware: the better-auth catch-all route (`app/api/auth/[...all]/route.ts`) resolves its `baseURL` via `resolveAuthBaseURL(req.headers, config)` (validated against trusted hosts), so it **signs id_tokens and issues codes with `iss` = the request host**. The static metadata advertising the *canonical* host is therefore inconsistent with the tokens actually issued — a discovery-based OIDC relying party on a served domain can reject the id_token for an `iss` mismatch, and Mastodon-style clients get silently moved to the canonical host. This was surfaced while reproducing the Phanpy login issue in [#1139](https://github.com/llun/activities.next/pull/1139).

## Fix

- `getOAuthAuthorizationServerMetadata(baseURL?)` and `getOpenIDConfiguration(baseURL?)` now accept a `baseURL` that defaults to `getBaseURL()` (so non-request callers and existing tests are unchanged).
- The three route handlers (`oauth-authorization-server`, `openid-configuration`, `oauth/authorize`) pass `resolveAuthBaseURL(req.headers, getConfig())` — the **same** validated per-request origin better-auth uses. `selectHeaderHost` still falls back to the configured host for untrusted Host headers, so the advertised host can't be steered via header injection.

## Testing

- `lib/services/wellknown/index.test.ts`: both builders advertise a provided alias `baseURL` across issuer + all endpoints, while the default (configured host) behavior is unchanged.
- `app/api/oauth/authorize/route.test.ts` (new): GET/POST keep the redirect on the resolved request host and preserve query/form params; untrusted host falls back to the configured host.
- `yarn run prettier --write .` ✓ · `yarn lint` ✓ (0 errors) · `yarn build` ✓ · `yarn test` ✓ (5191)
- **End-to-end on the `activities.local` + `alias.local` stack:** the alias host now advertises **itself** across both discovery docs (`issuer`/`authorization_endpoint`/`token_endpoint`/`app_registration_endpoint`), and the canonical host is unchanged.

### DIFF
--- a/app/api/oauth/authorize/route.test.ts
+++ b/app/api/oauth/authorize/route.test.ts
@@ -5,12 +5,10 @@ import { resolveAuthBaseURL } from '@/lib/services/auth/requestOrigin'
 import { GET, POST } from './route'
 
 vi.mock('@/lib/config', () => ({
-  getConfig: vi
-    .fn()
-    .mockReturnValue({
-      host: 'activities.local',
-      trustedHosts: ['alias.local']
-    })
+  getConfig: vi.fn().mockReturnValue({
+    host: 'activities.local',
+    trustedHosts: ['alias.local']
+  })
 }))
 
 vi.mock('@/lib/services/auth/requestOrigin', () => ({
--- a/app/api/oauth/authorize/route.test.ts
+++ b/app/api/oauth/authorize/route.test.ts
@@ -1,0 +1,87 @@
+import { NextRequest } from 'next/server'
+
+import { resolveAuthBaseURL } from '@/lib/services/auth/requestOrigin'
+
+import { GET, POST } from './route'
+
+vi.mock('@/lib/config', () => ({
+  getConfig: vi
+    .fn()
+    .mockReturnValue({
+      host: 'activities.local',
+      trustedHosts: ['alias.local']
+    })
+}))
+
+vi.mock('@/lib/services/auth/requestOrigin', () => ({
+  resolveAuthBaseURL: vi.fn()
+}))
+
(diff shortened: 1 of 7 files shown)


llun:
### ✅ Review loop converged

- **Gemini:** 4 reviews, final pass has no inline comments.
- **Kilo:** No Issues Found.
- **Threads:** 0 unresolved.
- **Checks:** All Tests, Build, Lint and Prettier, CodeQL/Analyze, CI Success — all green.

Round 1 fix: CI `prettier:check` flagged a test file where prettier was non-idempotent on a hand-written method chain; reformatted to the stable shape.

PR #1144, opened 2026-06-30
fix: show Translate option on remote profile feeds when logged in

llun:
## Problem

Remote actor profile pages (e.g. `/@gemeenteamsterdam@social.amsterdam.nl`) did not show the **Translate** control on their timeline posts, even for a signed-in viewer — while the home timeline showed it on the same kind of posts.

Both feeds render through the **same** `Posts` → `Post` component, and the profile page already passes `currentActor`/`showActions` correctly when logged in, so login state wasn't the gate. The real gate is in `lib/components/posts/post.tsx`:

```ts
const translationLanguage = props.currentActor ? actualStatus.language : null
```

The Translate control only renders when a status has a known source `language`. The two feeds populate `language` through different paths:

- **Home timeline** — statuses are federated into the local DB, where `createNoteJob`/`updateNoteJob` call `getLanguage(note)`, so `language` is set.
- **Remote profile page** — statuses are fetched live from the actor's outbox and built with `fromNote()`, which **never set `language`**. So every remote-profile post had `language: undefined`, suppressing the Translate control.

## Fix

Populate `language` in `fromNote` via the existing `getLanguage(note)` helper (the note's content/summary locale maps), mirroring the federation ingestion path. Remote-fetched statuses now carry their declared language, so the Translate control appears on profile feeds exactly as it does on the home timeline when signed in — consistent actions everywhere.

## Notes

- This fixes newly-rendered remote posts. Statuses already federated into the DB before the language helper existed remain `language: null` until re-fetched — a pre-existing limitation documented on `getLanguage`.

## Tests

- Added two `fromNote` unit tests: resolves `nl` from a content locale map; stays `null` when the note has no locale map.
- `yarn lint` (0 errors), `yarn build`, and `yarn test` (5197 passed) all green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LShBXtBn3WMXWxbULJFBsb)_

### DIFF
--- a/lib/jobs/createPollJob.test.ts
+++ b/lib/jobs/createPollJob.test.ts
@@ -25,6 +25,7 @@ interface MockQuestionParams {
   id?: string
   from?: string
   content?: string
+  contentMap?: { [key: string]: string }
   summary?: string | null
   published?: number
   endTime?: number
@@ -52,6 +53,7 @@ const MockActivityPubQuestion = ({
   id = `https://somewhere.test/actors/pollcreator/questions/${Date.now()}`,
   from = REMOTE_ACTOR_ID,
   content = '<p>What is your favorite color?</p>',
+  contentMap,
   summary = null,
   published = Date.now(),
   endTime = Date.now() + 24 * 60 * 60 * 1000,
@@ -85,6 +87,7 @@ const MockActivityPubQuestion = ({
     to,
     cc,
     content,
+    ...(contentMap ? { contentMap } : null),
     summary,
     published: getISOTimeUTC(published),
     endTime: getISOTimeUTC(endTime),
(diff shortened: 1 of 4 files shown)


llun:
/gemini review

---

llun:
## Subagent code-review — round 1 findings

I ran an independent deep review of the change (correctness, type-safety, path-tracing, consistency) on top of the Gemini and Kilo bot reviews.

**Verified correct / safe (no action):**
- `getLanguage(object: BaseNote)` accepts the `Note` passed by `fromNote` (`Note` is in the `BaseNote` union).
- `contentMap`/`summaryMap` survive the `getActorPosts` path: `compactActivityPub` produces the locale container, `normalizeActivityPubContent` doesn't touch it, and `Note.safeParse` keeps it — there's existing JSON-LD regression coverage for this.
- The single-status detail page and search also build remote statuses via `fromNote` (through `getRemoteStatus`), so they gain language consistently — no new inconsistency.
- No serialization/persistence impact: `toActivityPubObject` never reads `language`, and `fromNote` results are display-only (never written to the DB).
- `StatusNote.language` is `z.string().nullable().optional()`, so `getLanguage`'s `string | null` parses cleanly.

**One genuine, on-goal gap found and fixed in `e7fa46a`:**
- Polls (`Question`) never carried `language`. `createPollJob` didn't pass `language` to `database.createPoll` (which already supports it), so federated polls were stored with `language: null` and never showed the Translate control — unlike notes. Since the goal is "same actions everywhere when logged in," I added `getLanguage(question)` to `createPollJob`, mirroring `createNoteJob`. `createPoll` persists it; `updatePoll` leaves it untouched, so the create-time language survives edits. Added a unit test asserting the language is derived from the question's `contentMap`.

**Out of scope (flagged, not changed):** remote *profile* feeds don't render polls at all (`getActorPosts` only handles `Note`/`Announce`), so showing Translate on remote-profile polls would require rendering remote polls there — a separate, larger change. The home-timeline poll path is now fixed.

---

llun:
/gemini review

---

PR #1163, opened 2026-07-02
fix: stop credentials leaking into the URL on sign-in and fix 2FA verification

llun:
## Summary

Fixes two authentication problems reported from the sign-in flow:

1. **Password/email visible in the URL on sign-in.** The credential form (and the other client-side auth forms) could serialize `email`/`password` into the URL query string.
2. **"Invalid two factor cookie" on the 2FA challenge** (see screenshot in the issue) — 2FA verification always failed.

Both are small, targeted fixes with regression tests. All of `prettier`, `lint`, `build`, and `test` (5295 tests) pass.

## 1. Credentials leaking into the URL

The sign-in, two-factor, forgot-password, and reset-password forms are **client components** whose `onSubmit` handler calls `preventDefault()`, but they declared **no form `method`**. A `<form>` with no method defaults to **GET**, so if the form is submitted **before React hydrates** (fast typing + Enter, or slow/failed JS), the browser performs a native GET and serializes the fields into the URL query string — e.g. `/auth/signin?email=…&password=…`. That leaks credentials into browser history, server access logs, and the `Referer` header.

**Fix:** add `method="post"` to each form. Hydrated submits are unaffected (`preventDefault` still cancels the native submit); the attribute only hardens the pre-hydration/no-JS fallback so credentials always stay in the request body. The server-rendered signup form already used `method="post"`.

Forms updated:
- `CredentialForm` (sign-in — the reported form)
- `TwoFactorForm` (2FA code)
- `RequestPasswordResetForm` (email)
- `ResetPasswordForm` (reset code + new password)

## 2. "Invalid two factor cookie" on the 2FA challenge

**Root cause:** better-auth's two-factor plugin tracks per-account lockout on the `twoFactor` table via `failedVerificationCount` and `lockedUntil`, but the original `add_two_factor_auth` migration never created those columns. On the current better-auth version, enabling 2FA and verifying a code both run an INSERT/UPDATE referencing `failedVerificationCount`, which throws `table twoFactor has no column named failedVerificationCount`.

That surfaces to the user as **"Invalid two factor cookie"**: in `verifyTOTP`, `beginAttempt` consumes the one-time `2fa-attempts` verification counter, then `resetTwoFactorFailures` (correct code) / `recordTwoFactorFailure` (wrong code) 500s on the missing column **before** a session is created. The retry then finds the attempts counter already consumed and rejects with `INVALID_TWO_FACTOR_COOKIE`. This was reproduced end-to-end against a local database and confirmed fixed.

**Fix:**
- New migration `20260702120000_add_two_factor_lockout_columns.js` adding `failedVerificationCount` (`integer NOT NULL DEFAULT 0`) and `lockedUntil` (nullable timestamp).
- Regenerated **both** reference schema dumps (`migrations/schema.sql` for PostgreSQL, `migrations/schema.sqlite.sql` for SQLite) so they stay in lockstep with the migrations. The PostgreSQL dump was regenerated by running the full migration chain against a local PostgreSQL and diffing to confirm the only change is the two new columns.

## Tests

- `CredentialForm`, `TwoFactorForm`, `RequestPasswordResetForm`, `ResetPasswordForm`: assert the rendered `<form>` has `method="post"`.
- `lib/services/auth/twoFactorSignIn.test.ts`: a new integration test that boots the real better-auth instance against the SQLite schema dump and drives the full 2FA lifecycle (enable → verify → fresh sign-in challenge → verify), asserting the challenge completes and issues a session. It fails with a 500 when the lockout columns are absent, so it guards the schema/migration lockstep.

## Migration notes

This PR adds a migration, so both schema dumps are regenerated in the same PR. The migration is additive and backwards-compatible (patch release).

https://claude.ai/code/session_01N8o4vmwB3itf95EfjsBw8j

---
_Generated by [Claude Code](https://claude.ai/code/session_01N8o4vmwB3itf95EfjsBw8j)_


lib/services/auth/twoFactorSignIn.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    expect(initialSignIn.status).toBe(200)
+
+    const enable = await post('/two-factor/enable', { password: PASSWORD })
+    expect(enable.status).toBe(200)
+    const totpURI = enable.json.totpURI as string
+    // The otpauth URI carries the base32-encoded secret; createOTP keys the HMAC
+    // on the raw secret, so decode it back before generating codes.
+    const rawSecret = new TextDecoder().decode(
+      base32.decode(new URL(totpURI).searchParams.get('secret') as string)
+    )
```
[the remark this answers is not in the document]
llun: Thanks — I dug into this and the suggested change actually breaks the test, so I've addressed the underlying concern a different way instead.

`createOTP` does **not** accept a raw `Uint8Array` secret. In `@better-auth/utils`, `createOTP(secret)` forwards `secret` to `createHMAC().sign(secret, …)`, and `sign` only imports the key when it's a **string** (`TextEncoder().encode(key)` → `importKey`). A non-string secret is passed straight to `crypto.subtle.sign` as if it were an already-imported `CryptoKey`, so a `Uint8Array` can't stand in — passing the decoded bytes makes the generated codes stop matching the server (verified: the flow fails with the suggested change).

The server side keys the TOTP HMAC on the decrypted secret **string**, which better-auth generates with `generateRandomString(32)` (ASCII alphanumeric only). Because the secret is always ASCII, decoding the base32 bytes back into the string via `TextDecoder` is lossless — there's no non-UTF-8 byte to turn into `�`. I've kept the string decode and expanded the comment to explain exactly this. Good callout on the general risk, though — it just doesn't apply to this ASCII secret, and the byte path isn't compatible with `createOTP` here.

---

lib/services/auth/twoFactorSignIn.test.ts
```diff
@@ -0,0 +1,230 @@
...
+
+import { getSQLDatabase } from '@/lib/database/sql'
+import { Database } from '@/lib/database/types'
+
+// Boot the real better-auth instance (with the two-factor plugin) against an
+// in-memory SQLite database built from the committed schema dump — the same way
+// the rest of the SQL suite builds its schema (see lib/database/testUtils.ts),
+// which loads migrations/schema.sqlite.sql rather than running the migration
+// chain — and drive the full 2FA lifecycle end to end: enable -> verify -> fresh
+// sign-in challenge -> verify. This is the regression guard for the "Invalid two
```
llun, line 16: **Subagent review (completeness lens) — addressed in e493c3d.**

The original header comment claimed this test fails "if the schema dump drifts away from the migrations again," but it never runs the migrations — `buildInMemoryKnex` loads `migrations/schema.sqlite.sql` and execs it (which is exactly how the rest of the SQL suite builds its schema; see `lib/database/testUtils.ts`). So the test can't actually detect dump↔migration drift; it verifies the 2FA flow works against the committed dump.

I kept the dump-loading approach (switching this one test to `migrate()` would diverge from the suite convention, which deliberately avoids pulling the ESM migration chain into the test runtime) and reworded the comment to state accurately what it guards: the committed dump carries the lockout columns, and dump↔migration lockstep is enforced by the AGENTS.md policy. If a migration lands without regenerating the dump, `enable`/`verify-totp` 500 and this test fails.

---

app/(timeline)/account/security/ChangePasswordForm.tsx
```diff
@@ -59,7 +59,11 @@ export const ChangePasswordForm: FC = () => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="space-y-4">
+    // method="post" is defense-in-depth: these inputs are controlled and have no
+    // `name`, so a native (pre-hydration/no-JS) submit sends nothing today, but a
+    // method-less <form> defaults to GET — POST keeps the current/new password out
+    // of the URL if a `name`/autocomplete attribute is ever added for autofill.
+    <form onSubmit={handleSubmit} method="post" className="space-y-4">
```
llun, line 66: **Subagent review (security-completeness lens) — addressed in e493c3d.**

The hardening pass originally missed `ChangePasswordForm` and `ChangeEmailForm`, which carry the current/new password and the new email and share the exact shape of the auth forms this PR converted. There's no live leak today (their inputs are controlled and unnamed), but for consistency with the PR's own stated principle — keep credentials in the body "in every case" — I added `method="post"` to both, with parity render tests. This is the form that would become uniquely unguarded if someone later adds `name`/autocomplete for password-manager autofill.

---

app/(nosidebar)/auth/reset-password/ResetPasswordForm.tsx
```diff
@@ -77,7 +77,11 @@ export const ResetPasswordForm: FC<Props> = ({ initialCode }) => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="space-y-4">
+    // method="post" is defense-in-depth. These inputs are controlled and have no
```
llun, line 80: **Subagent review (security-completeness lens) — addressed in e493c3d.**

The original comments/test names on the three *controlled* forms (this one, `RequestPasswordResetForm`, `TwoFactorForm`) overstated the pre-fix behavior: their `<Input>`s are controlled and have no `name`, so per the HTML forms spec they aren't successful controls — a native GET submit produces an *empty* query string, not a leak. Only `CredentialForm` (uncontrolled, `name="email"`/`name="password"`, read via `FormData`) actually leaked before this PR.

`method="post"` is still worth keeping here as forward-looking defense-in-depth (it protects a future edit that adds `name`/autocomplete). I reworded the three comments and their test descriptions to say that, and made `CredentialForm`'s comment state explicitly that it's the real-leak case.

---

llun (commented):
Self-review from a multi-agent pass (5 lenses round 1 → clean; round 2 broader/senior-reviewer lenses → 3 findings below, adversarially verified). All three are addressed in e493c3d. Details inline; resolving each thread.

---

app/(nosidebar)/auth/two-factor/TwoFactorForm.tsx
```diff
@@ -64,7 +64,11 @@ export const TwoFactorForm: FC<Props> = ({ redirectBack }) => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="space-y-5">
+    // method="post" is defense-in-depth. The code input is controlled and has no
+    // `name`, so a native (pre-hydration/no-JS) submit sends nothing today, but a
+    // method-less <form> defaults to GET — POST keeps the verification/backup code
+    // out of the URL if a `name` attribute is added later.
```
llun, line 70: **Round-3 subagent review — addressed in bc675cb.**

The round-2 rewording introduced an imprecision: the defense-in-depth comments cited "`name`/autocomplete attribute added later" as the leak trigger. But `autocomplete` has no bearing on form serialization — per the HTML "successful controls" rule, only a non-empty `name` puts a control in the submitted entry list. This form's code input even already sets `autoComplete="one-time-code"` (line 105) yet leaks nothing, which made "if autocomplete is added later" self-contradictory. Reworded all six comments/tests to reference `name` only.

---

app/(timeline)/account/ChangeNameForm.tsx
```diff
@@ -47,7 +47,11 @@ export const ChangeNameForm: FC<Props> = ({ currentName }) => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="space-y-4">
+    // method="post" is a real protection here, not just defense-in-depth: this
+    // input is NAMED (name="name"), so on a native (pre-hydration/no-JS) submit it
+    // is a successful control and a method-less <form> — which defaults to GET —
+    // would serialize the name into the URL. POST keeps it in the request body.
+    <form onSubmit={handleSubmit} method="post" className="space-y-4">
```
llun, line 54: **Round-3 subagent review — addressed in bc675cb.**

Round 3 also flagged `ChangeNameForm` as the one remaining form with a **named** input (`name="name"`) and no `method="post"` — so unlike the other (unnamed) account forms, a native GET submit would actually serialize `?name=<value>` into the URL. The value is only the public display name (low sensitivity, which is why the verifier scored it below the posting bar), but it's the same class as `CredentialForm` and the last inconsistency in the hardening pass, so I closed it: added `method="post"` with a comment noting it's a real protection here (not just defense-in-depth), plus a parity test.

---

llun (commented):
Round 3 (convergence check on the round-2 edits): 2 findings, both addressed in bc675cb. One was a comment imprecision I introduced in round 2 (autocomplete vs name); the other hardens the last named account form. Resolving both threads.

---

app/(timeline)/account/security/ChangePasswordForm.tsx
```diff
@@ -59,7 +59,11 @@ export const ChangePasswordForm: FC = () => {
   }
 
   return (
-    <form onSubmit={handleSubmit} className="space-y-4">
+    // method="post" is defense-in-depth: these inputs are controlled and have no
+    // `name`, so a native (pre-hydration/no-JS) submit sends nothing today, but a
+    // method-less <form> defaults to GET — POST keeps the current/new password out
+    // of the URL if a `name` attribute is added later.
```
llun, line 65: **Round-4 subagent review — addressed in e889d3b.**

Final consistency sweep caught a residual from the round-3 reword: this comment and `ChangeEmailForm`'s still said "ever added for autofill" while every sibling comment and both components' own test comments said "added later." Harmonized both to "added later" so all five defense-in-depth comments read identically (comment-only, no behavior change).

---

llun (commented):
Round 4 (final convergence check): 1 low/cosmetic finding — a wording inconsistency left over from round 3 — addressed in e889d3b. Resolving.

---

PR #1167, opened 2026-07-03
fix: hide passkey sign-in button where passkeys aren't usable

llun:
## Summary

The **"Sign in with Passkey"** button on `/auth/signin` was always rendered — including inside in-app browsers that can't complete a WebAuthn ceremony, such as an **iOS `WKWebView`**, the **Schrift app's OAuth login dialog**, and similar embedded webviews. In those contexts the button either does nothing or fails with an error, which is confusing.

This change gates the button on **client-side WebAuthn feature detection** so it only appears where a passkey sign-in can actually work. When passkeys are the *only* sign-in method (an instance with credential sign-in disabled) and they aren't usable, it shows a short **"unavailable" notice** instead of a blank card.

## Behavior

Support is feature-detected on the client (`window.PublicKeyCredential` + `isUserVerifyingPlatformAuthenticatorAvailable()`), tri-state (`null` while detecting):

| Passkeys usable? | Credential sign-in | Result |
|---|---|---|
| Yes | either | Passkey button shown |
| No | **enabled** | Button hidden; visitor uses the credential form |
| No | **disabled** | Notice: *"Passkeys aren't available in this browser"* (no button) |
| detecting… | either | Nothing yet (no flash) |

The button/notice **starts hidden and reveals only after detection resolves**, so SSR and the first client render produce identical output → **no hydration mismatch**, and unsupported in-app browsers never show a broken button.

## What changed

- **`app/(nosidebar)/auth/signin/passkeySupport.ts`** (new) — `isPlatformPasskeyAvailable()`: `false` unless the WebAuthn API exists and a user-verifying platform authenticator is available; degrades to `false` on missing API / error.
- **`app/(nosidebar)/auth/signin/PasskeySigninButton.tsx`** — feature-detects in a mount effect; renders the button, the unavailable notice, or nothing per the table above. Takes a `credentialEnabled` prop (defaults `true`).
- **`app/(nosidebar)/auth/signin/page.tsx`** — passes `credentialEnabled` (from `ACTIVITIES_AUTH`) to the button.
- Tests: `passkeySupport.test.ts` (5) and `PasskeySigninButton.test.tsx` (10).

## Why feature detection (not user-agent sniffing)

`isUserVerifyingPlatformAuthenticatorAvailable()` reports `false` in in-app WKWebViews, so the button hides generically — no app/UA allow-lists to maintain.

**Trade-off:** a desktop with *only* a roaming USB security key and no platform authenticator (Touch ID / Windows Hello) would also hide the button. Acceptable — that audience is tiny and, when credential sign-in is enabled, the credential form stays available.

## Scope

Limited to the functional sign-in button (the surface embedded in app login dialogs). The landing page's "Continue with a passkey" links and the account-settings passkey manager are out of scope.

## Testing

- `yarn run prettier` / `yarn lint` (0 errors) / `yarn build` / `yarn test` (5312 passed) — all green.
- Browser verification against a local SQLite instance on `/auth/signin`:
  - Supported browser (`UVPAA=true`) → passkey button shown, console clean.
  - In-app browser simulated (`UVPAA` forced `false`), credential enabled → button hidden, credential form still shown.
  - **Passkey-only instance (`enableCredential:false`) + in-app browser → "Passkeys aren't available in this browser" notice shown, no button, console clean.**
  - SSR HTML never emits the button/notice markup → hydration-safe.

### DIFF
--- a/app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
+++ b/app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
@@ -77,7 +77,14 @@ export const PasskeySigninButton: FC<PasskeySigninButtonProps> = ({
   if (!supported) {
     if (credentialEnabled) return null
     return (
-      <div className="flex items-start gap-3 rounded-lg border bg-muted/40 p-3.5">
+      // The notice is injected after client-side detection, so mark it as a
+      // polite live region — otherwise assistive tech never announces the only
+      // content telling the visitor they can't sign in here (WCAG 2.1 SC 4.1.3).
+      <div
+        role="status"
+        aria-live="polite"
+        className="flex items-start gap-3 rounded-lg border bg-muted/40 p-3.5"
+      >
         <span className="mt-px flex size-7 shrink-0 items-center justify-center rounded-md bg-muted text-muted-foreground">
           <Fingerprint className="size-[15px]" />
         </span>
--- a/app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
+++ b/app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
@@ -1,5 +1,6 @@
 'use client'
 
+import { Fingerprint } from 'lucide-react'
 import { useRouter, useSearchParams } from 'next/navigation'
 import { FC, useEffect, useState } from 'react'
 
@@ -10,15 +11,29 @@ import { passkeyErrorMessage } from './passkeyErrorMessage'
 import { isPlatformPasskeyAvailable } from './passkeySupport'
 import { resolveSignInRedirect } from './resolveSignInRedirect'
 
-export const PasskeySigninButton: FC = () => {
+interface PasskeySigninButtonProps {
+  /**
+   * Whether credential (email/password) sign-in is available on this instance.
+   * When it is, an environment that can't use passkeys simply hides the button
+   * (the visitor signs in with credentials instead). When it is NOT, passkeys
+   * are the only way in, so an unsupported environment shows a short
+   * "unavailable" notice rather than a blank sign-in card. Defaults to true.
+   */
+  credentialEnabled?: boolean
+}
+
+export const PasskeySigninButton: FC<PasskeySigninButtonProps> = ({
+  credentialEnabled = true
+}) => {
   const [error, setError] = useState<string>()
   const [loading, setLoading] = useState(false)
-  // Start hidden and reveal only once we've confirmed the client can complete a
-  // platform passkey ceremony. Both SSR and the first client render produce the
-  // same `null`, so there's no hydration mismatch, and in-app browsers that
-  // can't use passkeys (iOS WKWebView, the Schrift login dialog, …) never show
-  // the button at all.
-  const [supported, setSupported] = useState(false)
+  // `null` while we're still feature-detecting; `true`/`false` once resolved.
+  // Starting at `null` keeps SSR and the first client render identical (both
+  // render nothing), so there's no hydration mismatch — and no flash of the
+  // "unavailable" notice on browsers that do support passkeys. In-app browsers
+  // that can't use passkeys (iOS WKWebView, the Schrift login dialog, …) resolve
+  // to `false`.
+  const [supported, setSupported] = useState<boolean | null>(null)
   const searchParams = useSearchParams()
   const router = useRouter()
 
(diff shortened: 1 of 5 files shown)


app/(nosidebar)/auth/signin/passkeySupport.ts
```diff
@@ -0,0 +1,28 @@
...
+  ) {
+    return false
+  }
+
+  try {
+    return await publicKeyCredential.isUserVerifyingPlatformAuthenticatorAvailable()
+  } catch {
+    return false
+  }
+}
```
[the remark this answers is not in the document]
llun: Thanks — this is a deliberate, documented trade-off rather than an oversight, so resolving as by-design.

`isUserVerifyingPlatformAuthenticatorAvailable()` is chosen because it is the signal that reliably returns `false` inside the in-app browsers this PR targets (iOS `WKWebView`, the Schrift app's OAuth login dialog, and similar embedded webviews), where a passkey ceremony cannot complete. The looser signals that would keep the button for a desktop user with only a roaming key (YubiKey) or hybrid/QR transport — treating `window.PublicKeyCredential` presence, or `isConditionalMediationAvailable()`, as "supported" — also risk re-showing a non-functional button in exactly those in-app webviews, which is the regression this change exists to prevent and which can't be empirically ruled out for every embedded webview.

Given the affected population (a desktop with no platform authenticator at all — no Touch ID / Windows Hello / Android — but with a roaming key) is small, and that when credential sign-in is enabled (the default) it still has the credential form as a fallback, the stricter platform-authenticator check is the intended behavior. The trade-off is called out in the PR description.

app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
```diff
@@ -36,6 +68,31 @@ export const PasskeySigninButton: FC = () => {
...
+  // Still feature-detecting — render nothing yet.
+  if (supported === null) return null
+
+  // Passkeys can't be used in this browser. If credential sign-in is available
+  // the visitor still has a way in, so just hide the button; otherwise passkeys
+  // were the only method, so explain the situation instead of a blank card.
+  if (!supported) {
+    if (credentialEnabled) return null
+    return (
+      <div className="flex items-start gap-3 rounded-lg border bg-muted/40 p-3.5">
```
llun: **[MEDIUM] accessibility (WCAG 2.1 SC 4.1.3)** — The passkeys-unavailable notice is injected client-side after feature detection but is not a live region, so screen-reader users may never be told why there is no way to sign in.

The notice renders `null` on SSR and the first client render (`supported === null`), then appears only after `isPlatformPasskeyAvailable()` resolves in the mount effect. A DOM node inserted after initial render is not re-announced by assistive technology unless it is inside a live region. In the credential-disabled path this notice is the *only* content of the sign-in card, so a screen-reader user who has already moved focus into/past the card gets no cue that anything appeared or that they cannot proceed. This diverges from the app's own convention for dynamically-appearing notices (e.g. `lib/components/posts/fitness-processing-progress.tsx` uses `role="status" aria-live="polite"`; search / notifications / lists wrap dynamic messages in `role="status"`/`role="alert"`). Independently flagged by the accessibility and ux-copy review passes.

**Suggested fix:** Add `role="status" aria-live="polite"` to the notice wrapper div (matching `fitness-processing-progress.tsx`) so the message is announced when it appears after client detection, and add a regression test asserting the notice is reachable via `getByRole('status')`.

<sub>🤖 automated multi-agent review</sub>

llun: Fixed in d6995186. The notice wrapper now carries `role="status" aria-live="polite"` so assistive tech announces it when it's injected after client-side detection — matching the app's existing dynamic-notice convention (`lib/components/posts/fitness-processing-progress.tsx`). Added a `getByRole('status')` regression test to lock the behaviour. `yarn lint` / `yarn build` / `yarn test` (5313) all green.

app/(nosidebar)/auth/signin/PasskeySigninButton.tsx
```diff
@@ -36,6 +68,38 @@ export const PasskeySigninButton: FC = () => {
...
+        className="flex items-start gap-3 rounded-lg border bg-muted/40 p-3.5"
+      >
+        <span className="mt-px flex size-7 shrink-0 items-center justify-center rounded-md bg-muted text-muted-foreground">
+          <Fingerprint className="size-[15px]" />
+        </span>
+        <div>
+          <p className="text-sm font-medium">
+            Passkeys aren&apos;t available in this browser
+          </p>
+          <p className="mt-0.5 text-xs leading-relaxed text-muted-foreground">
```
llun: **[LOW] accessibility (WCAG 2.1 SC 1.4.3)** — The notice's secondary line uses `text-xs text-muted-foreground` on the `bg-muted/40` panel, computing to ~4.47:1 contrast in light theme — just under the 4.5:1 AA minimum for normal text.

`--muted-foreground` (#737373) on the composited muted-tinted panel (~#f8f8f8 over the card's #fafafa) measures ~4.47:1 — below WCAG 2.1 AA SC 1.4.3's 4.5:1 for normal text (text-xs = 12px, not bold, so the large-text exception doesn't apply). This is the sole actionable message when passkeys are the only sign-in method and unsupported, so it's exactly the text the visitor must read. Dark theme passes; only light theme fails, marginally. (Note: this mirrors the muted-on-muted pattern also used in LandingAuthPanel's notice, so the design-system token pairing is marginally sub-AA app-wide; fixing here for the critical dead-end message.) The role="status"/aria-live and the auto-aria-hidden icon are correct.

**Suggested fix:** Give the secondary line a color that clears 4.5:1 with margin — e.g. `text-foreground/80` (~10:1) — while keeping a subtle hierarchy below the medium-weight title.

<sub>🤖 automated multi-agent review</sub>

llun: Fixed in fb931237. The secondary line now uses `text-foreground/80` (~10:1 on the muted panel) instead of `text-muted-foreground`, clearing WCAG 2.1 SC 1.4.3 (4.5:1) with margin while keeping a subtle hierarchy below the medium-weight title. The primary line and the `role="status"` live region are unchanged. `yarn lint` / `yarn build` green; signin tests pass.

Note: the same muted-on-muted token pairing exists in `LandingAuthPanel`'s notice, so a design-system-wide contrast pass on `text-muted-foreground` over `bg-muted` would be worthwhile separately — out of scope for this PR.

PR #1171, opened 2026-07-03
fix: accept documented granular OAuth scopes on Mastodon API routes

llun:
## Summary

Phase 1.1 of the Mastodon API remediation plan. Broadens the OAuth scope guards
on ~40 Mastodon-compatible routes so they accept the **documented granular
scopes** (e.g. `write:follows`, `read:notifications`, `profile`) in addition to
the coarse `read`/`write` scopes they already honored.

**Stacked on #1168** (Phase 0 docs) — base is `claude/artifact-phases-pr-review-8du95f`.

## Root cause

`scopeHierarchy` only expands coarse→granular (granted `read` satisfies
`read:statuses`), never the reverse — by design, so `write:media` never
satisfies a bare `write` guard. Routes guarded with `OAuthGuard([Scope.enum.write])`
therefore rejected a token granted only the granular scope the client actually
consented to. The fix lists both scopes with an **any-of** guard.

## Changes

- **`OAuthGuard([coarse]) → OAuthGuardAnyScope([coarse, granular])`** across the
  account actions (follow/unfollow, block/unblock, mute/unmute, note), status
  actions (favourite/unfavourite, reblog/unreblog), polls votes, preferences,
  announcements (dismiss, reactions), suggestions delete, and **all v1 + v2
  notifications routes** (`read:notifications` / `write:notifications`).
- **`OptionalOAuthGuard` routes** (`reblogged_by`, `favourited_by`, `polls/[id]`)
  add the granular read scope **plus `{ matchMode: 'any' }`** — without it,
  `OptionalOAuthGuard` defaults to `matchMode: 'all'` and would wrongly require
  every listed scope.
- **`verify_credentials` / `profile` (GET) / `userinfo`** now also accept the
  Mastodon `profile` scope.
- **`AdminApiGuard`** gains an opt-in `{ resource }` option that adds a route's
  own granular admin scope (`admin:read:domain_blocks`, `admin:write:domain_allows`,
  …) to the accepted set — wired into the `domain_blocks` / `domain_allows`
  admin routes. Isolation is preserved: a token consented for
  `admin:read:domain_blocks` gains nothing on unrelated admin routes.
- Suggestions delete now also accepts `read` (per Mastodon docs), keeping `write`
  for back-compat.

## Notes / no over-granting

- The change only ever **adds** accepted scopes; the coarse `read`/`write` and
  aggregate `admin:read`/`admin:write` paths are unchanged, so no request is let
  through beyond its consented granular scope.
[...]

PR #1173, opened 2026-07-03
fix: accept form-encoded and multipart bodies on lists/reports/featured_tags

llun:
## Summary

Phase 1.2 of the Mastodon API remediation plan. Native Mastodon clients send
`application/x-www-form-urlencoded` and `multipart/form-data` bodies (and
serialise `DELETE` params into the query string). These routes only read
`req.json()`, so form/multipart requests failed. Route bodies now go through the
shared form-aware parsers.

**Stacked on #1171** (granular OAuth scopes) — base is `claude/phase1-oauth-scopes`.

## Changes

- **lists (POST) & lists/[id] (PUT):** read the body via `getRequestBody`, and
  switch `exclusive` from `z.coerce.boolean()` (which makes the string `"false"`
  truthy) to the `Booleanish` schema so `"false"` is `false`.
- **lists/[id]/accounts (POST + DELETE):** parse `account_ids` from a JSON array,
  a form/multipart bracket array (`account_ids[]=…`), or — for `DELETE`, which
  masto.js serialises into the query string — repeated query params.
  `getRequestBody` alone can't do this: it collapses bracket keys last-write-wins.
- **featured_tags (POST):** read via `getRequestBody`.
- **reports (POST):** parse JSON/urlencoded/multipart, collecting `status_ids[]`
  / `rule_ids[]` bracket arrays, and switch `forward` to `Booleanish`.

## Tests

- New form-body tests for `lists`, `lists/[id]`, `lists/[id]/accounts`, and
  `reports` — urlencoded + multipart + DELETE-query + boolean-coercion matrices,
  asserting the parsed values reach the DB calls.
- Extend the existing `featured_tags` integration test with a urlencoded case
  (kept the original real-DB GET/POST coverage).
- `yarn run prettier --write .`, `yarn lint`, `yarn build`, `yarn test` — all green.

## Version bump

`fix:` — patch.

---
_Generated by [Claude Code](https://claude.ai/code/session_019yDNzak914a5pGcFGTu18v)_

### DIFF
--- a/app/api/v1/lists/[id]/route.test.ts
+++ b/app/api/v1/lists/[id]/route.test.ts
@@ -58,6 +58,13 @@ const createFormRequest = (body: string) =>
     headers: { 'content-type': 'application/x-www-form-urlencoded' }
   })
 
+const createJsonRequest = (body: unknown) =>
+  new NextRequest(`https://local.test/api/v1/lists/${LIST_ID}`, {
+    method: 'PUT',
+    body: JSON.stringify(body),
+    headers: { 'content-type': 'application/json' }
+  })
+
 const createMultipartRequest = (form: FormData) => {
   const request = new NextRequest(
     `https://local.test/api/v1/lists/${LIST_ID}`,
@@ -130,6 +137,18 @@ describe('PUT /api/v1/lists/:id', () => {
     expect(response.status).toBe(200)
     expect(mockDatabase.updateList).toHaveBeenCalledWith(
       expect.objectContaining({ title: 'Renamed', exclusive: false })
+    )
+  })
+
+  it('updates the list from a JSON body (regression)', async () => {
+    const response = await PUT(
+      createJsonRequest({ title: 'Renamed', exclusive: true }),
+      { params: Promise.resolve({ id: LIST_ID }) }
+    )
+
+    expect(response.status).toBe(200)
+    expect(mockDatabase.updateList).toHaveBeenCalledWith(
+      expect.objectContaining({ title: 'Renamed', exclusive: true })
     )
   })
 })
--- a/app/api/v1/lists/[id]/route.test.ts
+++ b/app/api/v1/lists/[id]/route.test.ts
@@ -1,0 +1,135 @@
+import { NextRequest } from 'next/server'
+
+import { PUT } from './route'
+
+const mockDatabase = {
+  updateList: vi.fn()
+}
+const mockCurrentActor = {
+  id: 'https://local.test/users/me',
+  domain: 'local.test'
+}
+
+vi.mock('@/lib/services/guards/OAuthGuard', () => ({
+  OAuthGuard:
+    (
+      _scopes: unknown,
+      handle: (
+        req: NextRequest,
+        context: {
+          database: typeof mockDatabase
+          currentActor: typeof mockCurrentActor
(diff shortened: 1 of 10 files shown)


app/api/v1/reports/route.ts
```diff
@@ -33,13 +34,61 @@ const toArray = (value: string[] | string | undefined): string[] => {
...
+    )
+  }
+  if (contentType.includes('multipart/form-data')) {
+    const form = await req.formData()
+    return buildFormReport(
+      (name) => form.get(name),
+      (name) => form.getAll(name)
+    )
+  }
+  return req.json()
```
llun, line 82: **test-coverage:** the JSON path here (`return req.json()`) — the default/primary content type — has no test in the new `reports/route.test.ts`; every case uses urlencoded or multipart, which only pass strings. So `Booleanish`'s `z.boolean()` arm (real JSON booleans) and the JSON `status_ids` array-union arm are uncovered — a regression breaking JSON parsing would stay green. Recommend a JSON-body test with a real boolean `forward` and a `status_ids: ['s1','s2']` array.

---

app/api/v1/lists/[id]/route.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    await PUT(createFormRequest(`title=Renamed&exclusive=${value}`), {
+      params: Promise.resolve({ id: LIST_ID })
+    })
+
+    expect(mockDatabase.updateList).toHaveBeenCalledWith(
+      expect.objectContaining({ exclusive: expected })
+    )
+  })
+
+  it('updates the list from a multipart body with exclusive=false', async () => {
```
Comment (llun) [inline app/api/v1/lists/[id]/route.test.ts:128]: **test-coverage:** the PUT handler switched from `req.json()` to `getRequestBody` (a distinct JSON path: `text()` + empty-guard + `JSON.parse`), but this file only covers urlencoded and multipart. The sibling `lists` POST test added a JSON regression case; PUT should too. Recommend a JSON-body PUT test (`{ title, exclusive: true }`) asserting `updateList` is called with `exclusive: true`.

---

llun (commented):
Automated review of PR 1.2. No correctness bugs in the body-parsing logic; two test-coverage gaps (the JSON/default content-type path is uncovered on reports and lists PUT). Both being addressed in a follow-up commit. (A third finding — the aggregate `read` case on verify_credentials — belongs to the stacked #1171 and is handled there.)

---

app/api/v1/reports/route.ts
llun: Done in b5e04a4 — added JSON-body reports tests: one with a real boolean `forward: true` and `status_ids: ['s1','s2']` (asserts `forward` true and the ids reach `createReport` as `[idToUrl('s1'), idToUrl('s2')]`), and one with a real JSON `forward: false`. These exercise the `req.json()` branch, Booleanish's `z.boolean()` arm, and the JSON array-union arm.

---

app/api/v1/lists/[id]/route.test.ts
Comment (llun) [reply app/api/v1/lists/[id]/route.test.ts:128]: Done in b5e04a4 — added a JSON-body PUT regression test (`{ title: 'Renamed', exclusive: true }`) asserting `updateList` is called with `exclusive: true`, mirroring the POST test.

---

PR #1175, opened 2026-07-03
fix: real min_id pagination semantics for notifications and list timeline

llun:
## Summary

Phase 1.3 of the Mastodon API remediation plan. `min_id` was treated identically
to `since_id`: both were collapsed into one lower-bound cursor applied to a
`DESC LIMIT n` query, so `min_id` returned the **newest** n rows above the cursor
instead of the page **immediately adjacent** to it. Mastodon's `min_id` is an
ascending keyset seek from the cursor — take n, then reverse to newest-first.

**Stacked on #1173.**

## Fixed (true `min_id` where the query layer is consumed directly)

- **`getNotifications`** (`lib/database/sql/notification.ts`): when
  `minNotificationId` is set, order ascending and reverse; `since`/`max`/no-cursor
  keep DESC. `GET /api/v1/notifications` now passes `min_id` and `since_id`
  through separately.
- **`getListTimeline`** (`lib/database/sql/list.ts`): same ascending-seek +
  reverse for `min_id`; `GET /api/v1/timelines/list/:id` passes them separately.
- Added `sinceStatusId` to `GetTimelineParams` / `GetListTimelineParams`.
- Adjacency tests for both layers proving `min_id` returns the oldest-n above the
  cursor (reversed to newest-first) while `since_id` returns the newest-n.
  Replaced the old `notification.test.ts` case that asserted `min_id === since_id`
  (which only held because its window was smaller than the page limit).

## Intentionally deferred (kept as `since` semantics, documented in-code)

Three paths reach the query layer through a **DESC-only backfill/grouping loop**
that pages by `max_id` and would break if the underlying query flipped to
ascending. They keep their current (since-like) `min_id` behavior; making them
truly adjacent-page is a separate change to those loops:

- **Home / direct timelines** — via `getFilteredStatusPage` (keyword/block/mute
  backfill). `getFilteredTimelinePage` now routes its lower bound through
  `sinceStatusId` so this fix can't accidentally flip it to ascending.
- **`GET /api/v2/notifications`** (grouped) — via `collectNotificationGroups`.
- **Public / tag timelines** — still `max_id`-only (unchanged).

This boundary is called out at each site in code comments.

## Testing

`yarn run prettier --write .`, `yarn lint`, `yarn build`, `yarn test` — all green
(601 files / 5395 tests).

## Version bump

`fix:` — patch.

---
_Generated by [Claude Code](https://claude.ai/code/session_019yDNzak914a5pGcFGTu18v)_

### DIFF
--- a/app/api/v1/notifications/route.ts
+++ b/app/api/v1/notifications/route.ts
@@ -124,7 +124,8 @@ export const GET = traceApiRoute(
         actorId: currentActor.id,
         limit,
         maxNotificationId: maxId,
-        minNotificationId: minId || sinceId,
+        minNotificationId: minId,
+        sinceNotificationId: sinceId,
         types: internalTypes,
         excludeTypes: internalExcludeTypes,
         includeFiltered
--- a/app/api/v1/timelines/list/[list_id]/route.ts
+++ b/app/api/v1/timelines/list/[list_id]/route.ts
@@ -71,19 +71,16 @@ export const GET = traceApiRoute(
         }
         const limit = parsedQuery.query.limit
         const maxStatusId = parsedQuery.query.maxStatusId
-        // `min_id` and `since_id` both express a lower-bound cursor; the list
-        // timeline query takes a single min cursor, so collapse them preserving
-        // this route's existing `min_id`-wins precedence. (min/since ordering
-        // semantics are unchanged by this PR — see PR notes.)
-        const minStatusId =
-          parsedQuery.query.minStatusId ?? parsedQuery.query.sinceStatusId
-
+        // Pass min_id and since_id through separately: min_id returns the page
+        // immediately adjacent to the cursor (ascending seek then reversed),
+        // since_id the newest slice above it.
         const statuses = await database.getListTimeline({
           listId,
           actorId: currentActor.id,
           limit,
           maxStatusId,
-          minStatusId
+          minStatusId: parsedQuery.query.minStatusId,
+          sinceStatusId: parsedQuery.query.sinceStatusId
         })
 
         // The list timeline query returns domain statuses newest-first, so the
--- a/lib/database/sql/notification.ts
+++ b/lib/database/sql/notification.ts
@@ -86,11 +86,7 @@ export const NotificationSQLDatabaseMixin = (
     sinceNotificationId,
     includeFiltered
   }: GetNotificationsParams) {
-    let query = database('notifications')
-      .where('actorId', actorId)
-      .orderBy('createdAt', 'desc')
-      .orderBy('id', 'desc') // Secondary sort for deterministic ordering
-      .limit(limit)
+    let query = database('notifications').where('actorId', actorId).limit(limit)
 
     // By default hide policy-filtered notifications (they live in the requests
     // queue). Mastodon's `include_filtered` opts back into seeing them.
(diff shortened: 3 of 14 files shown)


app/api/v1/notifications/route.ts
```diff
@@ -124,7 +124,8 @@ export const GET = traceApiRoute(
         actorId: currentActor.id,
         limit,
         maxNotificationId: maxId,
-        minNotificationId: minId || sinceId,
+        minNotificationId: minId,
+        sinceNotificationId: sinceId,
```
llun, line 128: **test-coverage:** this split (previously `minNotificationId: minId || sinceId`) is the exact regression the PR fixes, but no route test verifies the wiring — `notifications/route.test.ts` only asserts limit/types/include_filtered. A revert to `minId || sinceId` would pass every route test. Recommend an `it.each` spying on `getNotifications` asserting `?min_id=X` → `{ minNotificationId: X, sinceNotificationId: undefined }` and `?since_id=Y` → `{ sinceNotificationId: Y, minNotificationId: undefined }`.

---

app/api/v1/timelines/list/[list_id]/route.ts
```diff
@@ -71,19 +71,16 @@ export const GET = traceApiRoute(
...
+        // immediately adjacent to the cursor (ascending seek then reversed),
+        // since_id the newest slice above it.
         const statuses = await database.getListTimeline({
           listId,
           actorId: currentActor.id,
           limit,
           maxStatusId,
-          minStatusId
+          minStatusId: parsedQuery.query.minStatusId,
+          sinceStatusId: parsedQuery.query.sinceStatusId
```
Comment (llun) [inline app/api/v1/timelines/list/[list_id]/route.ts:83]: **test-coverage:** the existing route test (`uses min_id over since_id…`) still sends both cursors and asserts only `objectContaining({ minStatusId })`, which passes whether or not `sinceStatusId` is forwarded — so it doesn't verify this split. And no test sends `since_id` alone to confirm it routes to `sinceStatusId` (not `minStatusId`). Recommend asserting the full call shape and adding a since-alone / min-alone case.

---

llun (commented):
Automated review of PR 1.3. One correctness issue (an unresolvable min/since cursor returned the oldest notifications instead of terminating — same as the Gemini bot's inline note) plus two route-level test-coverage gaps on the cursor-split wiring. All being addressed in a follow-up commit.

---

lib/database/sql/notification.ts
```diff
@@ -165,8 +161,18 @@ export const NotificationSQLDatabaseMixin = (
       query = query.whereIn('id', ids)
     }
 
+    // min_id ascends from the cursor — the OLDEST notifications just newer than
+    // it — then reverses to the newest-first response shape, so it returns the
+    // page adjacent to the cursor. since_id (and max_id / no cursor) keep the
+    // newest-first DESC ordering (the newest slice above the cursor).
+    const ascending = Boolean(minNotificationId)
```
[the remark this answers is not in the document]
llun: Good catch — fixed in 60d750c. `getNotifications` now `return []` when the lower-bound cursor (min_id or since_id) doesn't resolve for the actor, matching `getListTimeline`. Without it, the ascending min_id order would have surfaced the OLDEST notifications. Added a test asserting an unresolvable min_id cursor returns an empty page.

---

app/api/v1/notifications/route.ts
llun: Done in 60d750c — added an `it.each` in `notifications/route.test.ts` asserting `?min_id=X` reaches `getNotifications` as `{ minNotificationId: X, sinceNotificationId: undefined }` and `?since_id=Y` as `{ sinceNotificationId: Y, minNotificationId: undefined }`, so a revert to the `|| sinceId` collapse now fails.

---

app/api/v1/timelines/list/[list_id]/route.ts
Comment (llun) [reply app/api/v1/timelines/list/[list_id]/route.ts:83]: Done in 60d750c — the test now asserts the full call shape (`{ minStatusId, sinceStatusId }`) and a new `it.each` sends `min_id`/`since_id` alone, asserting each routes to its own slot with the other left `null`. Renamed away from "min over since".

---

PR #1192, opened 2026-07-09
fix: account entity + update_credentials modern fields

llun:
## What

- Adds Mastodon 4.1–4.4 Account entity fields to the serialized `Account`/`CredentialAccount`: `uri` (the actor's ActivityPub id), `roles: []` (no roles system; credential `role` overlay unchanged), `indexable` (default `false`, matching Mastodon's opt-in search default), `hide_collections` (nullable), and `source.attribution_domains`.
- Persists `indexable` / `hideCollections` / `attributionDomains` as new keys in the existing `actors.settings` JSON column — **no migration**, so the reference schema dumps (`migrations/schema.sql`, `migrations/schema.sqlite.sql`) are intentionally untouched.
- `PATCH /api/v1/accounts/update_credentials` (and the shared `PATCH /api/v1/profile`) now accept `indexable`, `hide_collections`, and `attribution_domains` (JSON and multipart, including Rails-style repeated `attribution_domains[]`) instead of silently ignoring them; attribution domains are trimmed/lowercased/deduped and capped at 100 × 255 chars.
- update_credentials now localizes the returned `acct` to the access domain via `localizeAccount(headerHost(...))`, matching `verify_credentials` on multi-domain deployments.
- `source[quote_policy]` remains deferred to the Phase 4.2 quotes epic (not accepted, not serialized).
- Fallback blocked/muted account builders updated for the extended schema; public `GET /api/v1/accounts/:id` gains the new fields automatically via the shared serializer.

## Notes

- No migrations — actor flags are new keys in the existing `actors.settings` JSON column, not new columns; schema dumps unchanged.
- Part of Phase 2 (Entity & Parameter Gaps), PR 2.5. Independent of the other Phase 2 PRs.


app/api/v1/accounts/update_credentials/route.test.ts
```diff
@@ -408,6 +409,118 @@ describe('PATCH /api/v1/accounts/update_credentials', () => {
...
+    expect(response.status).toBe(200)
+    expect(updateActor).toHaveBeenCalledWith(
+      expect.objectContaining({
+        attributionDomains: ['blog.example.com', 'news.example.com']
+      })
+    )
+    updateActor.mockRestore()
+  })
+
+  it('deduplicates attribution_domains and drops empty/whitespace entries', async () => {
```
llun, line 496: 🤖 **Sub-agent review (tests), addressed.** `attribution_domains` normalization has four steps (trim, lowercase, drop-empties, dedupe) but only trim+lowercase were tested — removing the `Set` dedupe or the `.filter(length>0)` would ship green. Fixed in cef2cdccd1098d0555490e3fd088c048da4fc52f: this case sends duplicate + whitespace-only entries and asserts they collapse to the single deduped domain.

app/api/v1/accounts/update_credentials/route.test.ts
```diff
@@ -408,6 +409,118 @@ describe('PATCH /api/v1/accounts/update_credentials', () => {
...
+    expect(response.status).toBe(200)
+    expect(updateActor).toHaveBeenCalledWith(
+      expect.objectContaining({
+        indexable: false,
+        hideCollections: false,
+        attributionDomains: []
+      })
+    )
+    // The spied updateActor calls through to the real test DB, so the response
+    // reflects the persisted values: a stored `false` must serialize as `false`,
```
llun, line 546: 🤖 **Sub-agent review (tests), addressed.** `hide_collections: settings.hideCollections ?? null` preserves the false/null/unset three-state, but only `true`/`null` were asserted — a `?? null` -> `|| null` regression would silently turn a stored `false` into `null` and pass every test. Fixed in cef2cdccd1098d0555490e3fd088c048da4fc52f: the JSON false-branch test now asserts the response body serializes `hide_collections: false` (and `indexable: false`).

lib/services/accounts/updateCredentialsHandler.ts
```diff
@@ -205,7 +233,21 @@ export const updateCredentialsHandler = (corsHeaders: HttpMethod[]) =>
...
+      const attributionDomains =
+        attribution_domains !== undefined
+          ? [
+              ...new Set(
+                attribution_domains
+                  .map((domain) => domain.trim().toLowerCase())
+                  .filter((domain) => domain.length > 0)
+              )
+            ]
+          : undefined
```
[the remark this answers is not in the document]
llun: Declining this one, with rationale. I traced every use of `attribution_domains`: it is only persisted to `settings.attributionDomains` and echoed back in `source.attribution_domains` as JSON — it is **never** used as an `href`, in `new URL(...)`, a `fetch`, a redirect, or any DOM/injection sink (grep across `lib/`+`app/` confirms no such consumer). So the "vulnerabilities in downstream consumers" scenario does not apply here, and JSON serialization is not an injection vector. The Zod schema already bounds the input (`z.string().max(255).array().max(100)`).

On the validation itself: this PR intentionally scoped normalization to trim/lowercase/dedupe/cap (per the entity/param plan). Strict hostname validation is a separate behavior decision — Mastodon returns **422** for an invalid `attribution_domains` entry rather than silently dropping it, whereas the suggested `.filter(regex)` would silently discard a user's typo with no feedback (and the proposed regex also rejects some valid forms, e.g. a trailing-dot FQDN). Since there is no security sink here, I'd rather not ship silent-drop behavior that diverges from Mastodon; proper 422-style domain validation is a reasonable follow-up, tracked as out of scope for this PR.

app/api/v1/accounts/update_credentials/route.test.ts
```diff
@@ -362,6 +363,36 @@ describe('PATCH /api/v1/accounts/update_credentials', () => {
     updateActor.mockRestore()
   })
 
+  it('rejects more than 100 attribution_domains with 422', async () => {
```
llun: 🤖 **Sub-agent review (tests, Round 2), addressed.** The `attribution_domains` schema caps (`.array().max(100)`, `.string().max(255)`) had no negative test — dropping either would let an unbounded list or over-length hostname reach persistence with every test still green. Fixed in 3b1ec1e7c7173c6c0c5c0847f20ed6a0d2ccae84: added 101-entry and 256-char cases asserting 422 + `updateActor` not called, mirroring the existing `fields.array().max(4)` negative test.

llun:
### 🤖 Review loop converged — ready for maintainer review

- **R1** — sub-agent (5 dims + adversarial verify): attribution_domains dedupe/empty-filter untested + hide_collections=false serialization untested → both fixed with tests. gemini: suggested strict hostname validation of attribution_domains → **declined with rationale** (the values are inert stored/echoed JSON with no href/URL/fetch/redirect sink — verified by grep across lib+app — and silent-filtering diverges from Mastodon's 422; proper validation noted as a follow-up).
- **R2** — sub-agent: the new `attribution_domains` 100/255 caps had no negative test → added 422 cases mirroring the existing `fields.array().max(4)` test.
- **R3** — sub-agent: 0 findings → **converged**.

gemini re-reviews (R2/R3) targeted test-only additions and it did not re-respond within the 20-min window (it dropped both re-triggers). All review threads resolved; CI green (tests, build, lint/prettier, Schema Dump Sync, CodeQL). Ready to merge.

PR #1193, opened 2026-07-09
fix: status create/edit/delete/history param gaps

llun:
## What

- `poll[hide_totals]` is now persisted on the poll (JSON content blob — no migration) and honored by the serializer: per-option `votes_count` is `null` while a `hide_totals` poll runs and reappears after expiry; top-level `votes_count`/`voters_count` stay numeric, matching Mastodon's `REST::PollSerializer`. Wired through immediate POST, scheduled publish, and poll edits.
- `DELETE /api/v1/statuses/:id?delete_media=true` destroys the status's media-manager rows (reusing the `DELETE /api/v1/media/:id` flow with its in-use guard and usage-counter decrements) and best-effort deletes the storage objects; without the param, media is kept for redrafting as before.
- `PUT /api/v1/statuses/:id` supports `media_attributes[][id|description|focus]` (JSON and form encodings), reusing the media route's validation and `updateMedia`, and refreshing the copied alt text on the status's attachment rows. Omitted fields are left untouched; explicit `""`/`null` clears alt text.
- `PUT /api/v1/statuses/:id` can now edit polls: text/CW/sensitive/language plus `poll[options]`/`expires_in`/`multiple`/`hide_totals`; option or multiple-mode changes replace the option set and reset all votes (Mastodon semantics), while `hide_totals`/`expires_in`-only changes keep votes. Poll edits are local-only for now (outbound `Update(Question)` federation is not wired up — follow-up).
- `GET /api/v1/statuses/:id/history` now renders per-revision `sensitive`, `media_attachments`, and `poll.options` from snapshots written into `status_history.data` at edit time; rows predating snapshots fall back to current values.

## Notes

- No schema migration: both `status_history.data` and `statuses.content` are JSON columns, so the reference schema dumps are untouched.
- Part of Phase 2 (Entity & Parameter Gaps), PR 2.6. Independent of the other Phase 2 PRs.

### DIFF
--- a/app/api/v1/statuses/[id]/route.test.ts
+++ b/app/api/v1/statuses/[id]/route.test.ts
@@ -2258,6 +2258,183 @@ describe('GET /api/v1/statuses/[id]', () => {
       expect(updatedStatus?.cc).toEqual([`${ACTOR1_ID}/followers`])
       expect(getQueue().publish).not.toHaveBeenCalled()
     })
+
+    it('updates attachment description and focus through media_attributes', async () => {
+      const statusId = `${ACTOR1_ID}/statuses/api-edit-media-attributes`
+      await database.createNote({
+        id: statusId,
+        url: statusId,
+        actorId: ACTOR1_ID,
+        text: 'Media attributes target',
+        to: [ACTIVITY_STREAM_PUBLIC],
+        cc: []
+      })
+      const media = await database.createMedia({
+        actorId: ACTOR1_ID,
+        original: {
+          path: 'medias/api-edit-attributes.webp',
+          bytes: 1024,
+          mimeType: 'image/jpeg',
+          metaData: { width: 320, height: 240 },
+          fileName: 'api-edit-attributes.jpg'
+        },
+        description: 'Old description'
+      })
+      expect(media).not.toBeNull()
+      await database.createAttachment({
+        actorId: ACTOR1_ID,
+        statusId,
+        mediaType: media!.original.mimeType,
+        url: 'https://llun.test/api/v1/files/medias/api-edit-attributes.webp',
+        width: 320,
+        height: 240,
+        name: 'Old description',
+        mediaId: media!.id
+      })
+
+      const response = await PUT(
+        new NextRequest(
+          `https://llun.test/api/v1/statuses/${urlToId(statusId)}`,
+          {
+            method: 'PUT',
+            body: JSON.stringify({
+              media_attributes: [
+                {
+                  id: media!.id,
+                  description: 'New description',
+                  focus: '0.5,-0.5'
+                }
+              ]
+            }),
+            headers: {
+              'Content-Type': 'application/json',
+              Origin: 'https://llun.test'
+            }
+          }
+        ),
+        { params: Promise.resolve({ id: urlToId(statusId) }) }
+      )
+
+      expect(response.status).toBe(200)
+      const data = await response.json()
+      expect(data.media_attachments).toHaveLength(1)
+      expect(data.media_attachments[0]).toMatchObject({
+        description: 'New description'
+      })
+
+      const actor = await database.getActorFromId({ id: ACTOR1_ID })
+      const updatedMedia = await database.getMediaByIdForAccount({
+        mediaId: media!.id,
+        accountId: actor!.account!.id
+      })
+      expect(updatedMedia?.description).toBe('New description')
+      expect(updatedMedia?.focus).toEqual({ x: 0.5, y: -0.5 })
+    })
+
+    it('keeps the existing description when media_attributes only updates focus', async () => {
+      const statusId = `${ACTOR1_ID}/statuses/api-edit-media-attributes-focus`
+      await database.createNote({
+        id: statusId,
+        url: statusId,
+        actorId: ACTOR1_ID,
+        text: 'Focus only target',
+        to: [ACTIVITY_STREAM_PUBLIC],
+        cc: []
+      })
+      const media = await database.createMedia({
+        actorId: ACTOR1_ID,
+        original: {
+          path: 'medias/api-edit-attributes-focus.webp',
+          bytes: 1024,
+          mimeType: 'image/jpeg',
+          metaData: { width: 320, height: 240 },
+          fileName: 'api-edit-attributes-focus.jpg'
+        },
+        description: 'Keep description'
+      })
+      expect(media).not.toBeNull()
+      await database.createAttachment({
+        actorId: ACTOR1_ID,
+        statusId,
+        mediaType: media!.original.mimeType,
+        url: 'https://llun.test/api/v1/files/medias/api-edit-attributes-focus.webp',
+        width: 320,
+        height: 240,
+        name: 'Keep description',
+        mediaId: media!.id
+      })
+
+      const response = await PUT(
+        new NextRequest(
+          `https://llun.test/api/v1/statuses/${urlToId(statusId)}`,
+          {
+            method: 'PUT',
+            body: JSON.stringify({
+              media_attributes: [{ id: media!.id, focus: '0,1' }]
+            }),
+            headers: {
+              'Content-Type': 'application/json',
+              Origin: 'https://llun.test'
+            }
+          }
+        ),
+        { params: Promise.resolve({ id: urlToId(statusId) }) }
+      )
+
+      expect(response.status).toBe(200)
+      const actor = await database.getActorFromId({ id: ACTOR1_ID })
+      const updatedMedia = await database.getMediaByIdForAccount({
+        mediaId: media!.id,
+        accountId: actor!.account!.id
+      })
+      // An omitted description must be left untouched (not cleared to null).
+      expect(updatedMedia?.description).toBe('Keep description')
+      expect(updatedMedia?.focus).toEqual({ x: 0, y: 1 })
+    })
+
+    it('rejects media_attributes for media the actor does not own', async () => {
+      const statusId = `${ACTOR1_ID}/statuses/api-edit-media-attributes-foreign`
+      await database.createNote({
+        id: statusId,
+        url: statusId,
+        actorId: ACTOR1_ID,
+        text: 'Foreign media attributes target',
+        to: [ACTIVITY_STREAM_PUBLIC],
+        cc: []
+      })
+      const foreignMedia = await database.createMedia({
+        actorId: ACTOR2_ID,
+        original: {
+          path: 'medias/api-edit-attributes-foreign.webp',
+          bytes: 1024,
+          mimeType: 'image/jpeg',
+          metaData: { width: 320, height: 240 },
+          fileName: 'api-edit-attributes-foreign.jpg'
+        }
+      })
+      expect(foreignMedia).not.toBeNull()
+
+      const response = await PUT(
+        new NextRequest(
+          `https://llun.test/api/v1/statuses/${urlToId(statusId)}`,
+          {
+            method: 'PUT',
+            body: JSON.stringify({
+              media_attributes: [
+                { id: foreignMedia!.id, description: 'Hijacked' }
+              ]
+            }),
+            headers: {
+              'Content-Type': 'application/json',
+              Origin: 'https://llun.test'
+            }
+          }
+        ),
+        { params: Promise.resolve({ id: urlToId(statusId) }) }
+      )
+
+      expect(response.status).toBe(422)
+    })
   })
 
   describe('status delete', () => {
--- a/app/api/v1/statuses/[id]/route.test.ts
+++ b/app/api/v1/statuses/[id]/route.test.ts
@@ -2392,6 +2392,66 @@ describe('GET /api/v1/statuses/[id]', () => {
       expect(updatedMedia?.focus).toEqual({ x: 0, y: 1 })
     })
 
+    it('clears an attachment description when media_attributes sends description null', async () => {
+      const statusId = `${ACTOR1_ID}/statuses/api-edit-media-attributes-clear`
+      await database.createNote({
+        id: statusId,
+        url: statusId,
+        actorId: ACTOR1_ID,
+        text: 'Clear description target',
+        to: [ACTIVITY_STREAM_PUBLIC],
+        cc: []
+      })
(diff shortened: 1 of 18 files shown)


app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -2251,6 +2258,575 @@ describe('GET /api/v1/statuses/[id]', () => {
...
+      const updatedMedia = await database.getMediaByIdForAccount({
+        mediaId: media!.id,
+        accountId: actor!.account!.id
+      })
+      // An omitted description must be left untouched (not cleared to null).
+      expect(updatedMedia?.description).toBe('Keep description')
+      expect(updatedMedia?.focus).toEqual({ x: 0, y: 1 })
+    })
+
+    it('clears an attachment description when media_attributes sends description null', async () => {
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:2395]: 🤖 **Sub-agent review (tests, medium), addressed.** The documented, ordering-fragile clear-alt-text path (`media_attributes` `description: null`/`""` -> clears) was untested — a refactor that dropped `.nullable()` or reordered the transform/`.optional()` would silently stop clearing while all tests passed. Fixed in b00902e867a573d40f4fd3c5af0c04cdd68b5749: PUTs `description: null` and asserts the stored alt text is cleared.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -2251,6 +2258,575 @@ describe('GET /api/v1/statuses/[id]', () => {
...
+      expect(data.content).toContain('Editable poll v2')
+      // Replaced options start from zero and hide_totals nulls the running
+      // tallies per option.
+      expect(data.poll.options).toEqual([
+        { title: 'New A', votes_count: null },
+        { title: 'New B', votes_count: null }
+      ])
+      expect(data.poll.votes_count).toBe(0)
+      expect(data.poll.voters_count).toBe(0)
+      // expires_in (7200s) is rebased from now into expires_at (seconds -> ms).
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:2550]: 🤖 **Sub-agent review (tests), addressed.** The new `expires_in` -> `expires_at` rebasing (`Date.now() + expiresIn*1000` in updatePoll.ts) was never asserted. Fixed in b00902e867a573d40f4fd3c5af0c04cdd68b5749: asserts `data.poll.expires_at` ≈ now + 7200s within tolerance (a dropped `*1000` or wrong base now fails).

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -2251,6 +2258,575 @@ describe('GET /api/v1/statuses/[id]', () => {
...
+      // Votes survive; hide_totals only masks the per-option numbers.
+      expect(data.poll.votes_count).toBe(1)
+      expect(data.poll.voters_count).toBe(1)
+      expect(data.poll.options).toEqual([
+        { title: 'Yes', votes_count: null },
+        { title: 'No', votes_count: null }
+      ])
+    })
+
+    it('resets votes and switches to anyOf when a poll edit flips multiple to true', async () => {
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:2618]: 🤖 **Sub-agent review (tests), addressed.** The vote-reset-on-multiple-choice-flip branch (`nextPollType !== status.pollType`) was untested (no test sent `multiple`). Fixed in b00902e867a573d40f4fd3c5af0c04cdd68b5749: votes a oneOf poll, PUTs the same options with `multiple: true`, and asserts votes reset to 0 and the poll now reports `multiple: true` (anyOf).

lib/services/statuses/parseStatusRequestBody.ts
```diff
@@ -59,6 +62,31 @@ const collectStatusFields = (
...
+      (value): value is string => typeof value === 'string'
+    )
+    body.media_attributes = mediaAttributeIds.map((id, index) => ({
+      id,
+      ...(index < descriptions.length
+        ? { description: descriptions[index] }
+        : {}),
+      ...(index < focuses.length ? { focus: focuses[index] } : {})
+    }))
+  }
```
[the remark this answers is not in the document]
llun: Good catch — fixed in 5f580a95214389a9f1cd468d1b0717120c7c8aa3. You're right that positional zipping of the bare `media_attributes[][id]`/`[][description]`/`[][focus]` fields mis-assigns when a client sends a partial submission (fewer description/focus values than ids). Since the bare-`[]` encoding associates values purely by position, a partial set is inherently ambiguous, so rather than guess I now apply description/focus only when its count equals the id count — otherwise the field is left untouched (never mapped to the wrong media). Added a regression test (`does not zip partial media_attributes fields...`). Clients needing reliable per-item control should use a JSON body (unambiguous nested array). Full-alignment submissions and JSON are unaffected.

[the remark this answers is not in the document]
llun: Good catch — fixed in 5f580a95214389a9f1cd468d1b0717120c7c8aa3. You're right that positional zipping of the bare `media_attributes[][id]`/`[][description]`/`[][focus]` fields mis-assigns when a client sends a partial submission (fewer description/focus values than ids). Since the bare-`[]` encoding associates values purely by position, a partial set is inherently ambiguous, so rather than guess I now apply description/focus only when its count equals the id count — otherwise the field is left untouched (never mapped to the wrong media). Added a regression test (`does not zip partial media_attributes fields...`). Clients needing reliable per-item control should use a JSON body (unambiguous nested array). Full-alignment submissions and JSON are unaffected.

lib/database/sql/status.ts
```diff
@@ -943,15 +987,23 @@ export const StatusSQLDatabaseMixin = (
     const currentTime = new Date()
     const data = getCompatibleJSON(existingStatus.content)
     const nextText = text ?? data.text
-    const nextSummary = summary ?? data.summary
+    // undefined = keep the existing summary; a provided value (including an
+    // empty/whitespace string) normalizes to null when blank, matching
+    // createPoll's `summary?.trim() || null` so a cleared CW never persists as
+    // '' (which would flip contentChanged and record a spurious edit revision).
+    const nextSummary =
+      summary === undefined ? data.summary : summary?.trim() || null
```
llun, line 995: 🤖 **Sub-agent review (invariants, real bug), addressed.** `createPoll` normalized an empty CW to `null` but `updatePoll` stored ``, so a null-CW poll edited with the conventional empty `spoiler_text=""` flipped `null`→``, tripping `contentChanged` and recording a spurious edit revision (+ bumped edited_at). Fixed in 263e2b91f8f3435c586f5e7fc66e3caf4545bed3: `updatePoll` now normalizes a provided empty/whitespace summary to `null` (undefined still keeps existing), matching createPoll. Federated `updatePollJob` is unaffected (a no-CW remote poll stores `null` instead of ``, rendered identically). Regression test added.

lib/services/statuses/parseStatusRequestBody.test.ts
```diff
@@ -119,4 +119,71 @@ describe('parseStatusRequestBody', () => {
...
+    params.append('media_attributes[][description]', 'only alt')
+    const body = await parseStatusRequestBody(
+      createRequest('application/x-www-form-urlencoded', params.toString())
+    )
+    expect(body).toEqual({
+      media_attributes: [{ id: '10' }, { id: '11' }]
+    })
+  })
+
+  it('applies aligned media_attributes fields independently (description without focus)', async () => {
```
llun, line 159: 🤖 **Sub-agent review (tests), addressed.** The two alignment checks are applied independently, but no test covered the asymmetric case — a regression coupling them would silently drop aligned descriptions. Fixed in 263e2b91f8f3435c586f5e7fc66e3caf4545bed3: added a 2 ids / 2 descriptions / 1 focus case asserting descriptions apply while focus is skipped.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -2251,6 +2258,585 @@ describe('GET /api/v1/statuses/[id]', () => {
...
+
+    it.each([
+      { param: 'media_ids', body: { media_ids: ['1'] } },
+      {
+        param: 'media_attributes',
+        body: { media_attributes: [{ id: '1', description: 'x' }] }
+      },
+      { param: 'visibility', body: { visibility: 'private' } }
+    ])(
+      'rejects a poll edit that also changes $param with 422',
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:2704]: 🤖 **Sub-agent review (tests), addressed.** The poll-edit 422 guard rejects `media_ids`, `media_attributes`, OR `visibility`, but only `media_ids` was tested — dropping either OR-clause would let a media_attributes-only or visibility-only poll edit return 200. Fixed in 263e2b91f8f3435c586f5e7fc66e3caf4545bed3: expanded to an it.each covering all three.

lib/database/sql/status.ts
```diff
@@ -983,6 +1049,24 @@ export const StatusSQLDatabaseMixin = (
...
+        await trx('poll_choices').where('statusId', statusId).delete()
+        for (const choice of choices) {
+          await trx('poll_choices').insert({
+            statusId,
+            title: choice.title,
+            totalVotes: 0,
+            createdAt: currentTime,
+            updatedAt: currentTime
+          })
+        }
```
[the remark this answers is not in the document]
llun: Declining this one, with rationale. Three reasons the current sequential insert is correct here:

1. **No conflict to handle.** The three poll tables are DELETEd for this `statusId` immediately above (lines 1061-1063) before re-inserting, so the `poll_choices` rows are guaranteed absent — there is no unique-constraint conflict for `.onConflict().ignore()` to guard against (and silently ignoring one would hide a real bug rather than help).
2. **Not a round-trip concern.** Poll options are capped (Mastodon allows at most 4), so this is ≤4 inserts inside an already-open transaction — batching saves microseconds at most.
3. **Ordering is deliberate and load-bearing.** The comment on lines 1059-1060 documents it: the inserts are sequential so the auto-increment `choiceId` is assigned in option order, which hydration relies on to return options in the authored order. A single batch insert does not guarantee auto-increment assignment order across SQLite / PostgreSQL / MySQL, so switching to a batch insert would risk silently reordering poll options.

So I'm keeping the sequential inserts.

lib/services/mastodon/getMastodonStatusEdits.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      updatedAt: new Date()
+    })
+
+    const status = (await database.getStatus({ statusId })) as StatusNote
+    const edits = await getMastodonStatusEdits(database, status)
+
+    expect(edits).toHaveLength(2)
+    expect(edits[0].content).toContain('Legacy version')
+    expect(edits[0]).toMatchObject({ sensitive: true })
+    // The legacy null-snapshot revision falls back to the current media list.
```
llun, line 116: 🤖 **Sub-agent review (weak assertion), addressed.** The legacy-revision media fallback test used a live status with no attachments, so it compared `[]` with `[]` and would pass even if the `revision.attachments ?? current.media_attachments` fallback regressed to always-`[]` — leaving this PR's legacy media fallback unverified. Fixed in b0d1c32252de49f2e4cfcdd091e576710d94aac0: the live status now has a real attachment, and the test asserts the legacy revision falls back to the non-empty current media list (length 1).

app/api/v1/statuses/[id]/route.ts
```diff
@@ -194,10 +245,138 @@ export const PUT = traceApiRoute(
...
         }
 
+        if (existingStatus.type === StatusType.enum.Poll) {
+          // Poll statuses carry no media, and visibility editing is only
+          // implemented for notes (updateNoteVisibility rejects polls).
+          if (
+            mediaIds !== undefined ||
+            mediaAttributes !== undefined ||
+            visibility !== undefined
+          ) {
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/statuses/[id]/route.ts:258]: Fixed in ad3217ef07b9a44d4c0c450b7c66dd1141617614. You're right — a client sending an empty `media_ids: []`/`media_attributes: []` (common default) on a poll edit was falsely 422'd. The guard now rejects only a NON-EMPTY media set (`length > 0`); an empty array is ignored. Added a regression test (`allows a poll edit that carries an empty media_ids array`).

lib/database/sql/status.ts
```diff
@@ -962,11 +1014,30 @@ export const StatusSQLDatabaseMixin = (
       createdAt: existingStatus.createdAt
     }
 
+    // Any user-visible change records an edit revision; a tally-only refresh
+    // (federated vote sync) must not.
+    const contentChanged =
+      nextText !== data.text ||
+      nextSummary !== data.summary ||
```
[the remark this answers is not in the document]
llun: Fixed in ad3217ef07b9a44d4c0c450b7c66dd1141617614. Good follow-on to the summary normalization — a poll created via the DB `createPoll` default (`summary: ''`) edited with a blank spoiler produced `nextSummary=null` vs `data.summary=''`, still tripping `contentChanged`. The comparison now normalizes both sides (`(summary || null)`), so '' and null are equivalent and no spurious revision is recorded. Added a regression test for the '' baseline (the null baseline was already covered).

llun:
### 🤖 Review loop converged (4 rounds) — ready for maintainer review

This XL PR's loop surfaced several real issues (all fixed):
- **R1** — sub-agent: 3 poll/media test gaps (multiple-flip vote reset, expires_in rebasing, alt-text clearing). gemini: **real bug** — partial form `media_attributes` fields were zipped by position and could mis-assign; now applied only when counts align.
- **R2** — sub-agent: **real bug** — poll edit stored empty CW as `''` (create normalizes to `null`), recording a spurious edit revision; fixed in SQL `updatePoll`. Plus 2 test gaps (independent field alignment, poll-edit 422 for media_attributes/visibility). gemini: poll_choices batch-insert (declined — deletes-first + documented deterministic-order requirement).
- **R3** — sub-agent: legacy media-fallback assertion was tautological (`[]`==`[]`) → strengthened with a real attachment. gemini: **HIGH usability bug** — empty `media_ids: []`/`media_attributes: []` (common client default) falsely 422'd poll edits; guard now rejects only non-empty sets. Plus the summary comparison normalized for the `''` baseline.
- **R4** — sub-agent: 0; gemini: clean → **converged**.

All review threads resolved; CI green (tests, build, lint/prettier, Schema Dump Sync, CodeQL). No migration (JSON columns). Ready to merge.

PR #1194, opened 2026-07-09
fix: public + tag timeline params

llun:
## What

- **Public timeline** (`/api/v1/timelines/public`): `only_media` is now parsed and applied as an attachments-EXISTS filter in both the local-public and federated queries; `min_id`/`since_id` (min_id wins) are wired through to the DB keyset query; a `rel="prev"` Link is emitted alongside `rel="next"`, and both links carry `limit`, `local`/`remote`, and `only_media`.
- **Tag timeline** (`/api/v1/timelines/tag/:hashtag`): Unicode hashtags (e.g. `#日本語`) no longer 400 — the ASCII regex is replaced by a shared `\p{L}\p{N}_` validator that also percent-decodes the route param; `any[]`/`all[]`/`none[]` tag filters implemented with Mastodon semantics (any widens, all intersects, none excludes; capped at 4 per mode); `local`/`remote` scope statuses by author locality; `only_media`, `min_id`/`since_id`, and a `rel="prev"` Link as on the public route; the route now runs under the `public` keyword-filter context so hide-filters drop rows and warn-filters annotate (mirroring the public timeline).
- **DB layer**: `getTimeline` gains `onlyMedia` for the two public timelines; `getStatusesByHashtag` gains `minStatusId`, `onlyMedia`, `local`/`remote`, `anyTags`/`allTags`/`noneTags` (all optional; existing callers unaffected).
- Shared helper `lib/utils/text/mastodonHashtag.ts` extracted from the featured-tags validator; `app/api/v1/featured_tags` now consumes it (no behavior change). `/api/v1/tags/:tag` + follow/unfollow still use the old ASCII regex — switched to this helper in PR 2.8.

## Notes

- `min_id`/`since_id` are both treated as a lower bound on the newest-first query, consistent with every other timeline route in this codebase; true `min_id` adjacency ordering remains unimplemented repo-wide (pre-existing, unchanged).
- No migrations; schema dumps untouched.
- Part of Phase 2 (Entity & Parameter Gaps), PR 2.7. **PR 2.8 (discovery polish) stacks on this branch** for the shared hashtag validator.

### DIFF
--- a/app/api/v1/timelines/public/route.test.ts
+++ b/app/api/v1/timelines/public/route.test.ts
@@ -90,6 +90,37 @@ describe('GET /api/v1/timelines/public', () => {
     ])
   })
 
+  it('applies instance-wide server hide filters for anonymous viewers', async () => {
+    await database.createServerFilter({
+      title: 'Server hidden',
+      context: ['public'],
+      filterAction: 'hide',
+      expiresAt: null,
+      keywords: [{ keyword: 'serverhidden', wholeWord: false }]
+    })
+    const hiddenStatus = await database.createNote({
+      id: `${ACTOR1_ID}/statuses/public-server-hidden`,
+      url: `${ACTOR1_ID}/statuses/public-server-hidden`,
+      actorId: ACTOR1_ID,
+      to: [ACTIVITY_STREAM_PUBLIC],
+      cc: [],
+      text: 'serverhidden content'
+    })
+    vi.spyOn(database, 'getTimeline').mockResolvedValue([
+      hiddenStatus,
+      publicStatus
+    ])
+
+    const response = await GET(request(), { params: Promise.resolve({}) })
+
+    expect(response.status).toBe(200)
+    const ids = (await response.json()).map((s: { id: string }) => s.id)
+    // The server-wide hide filter drops the matching status even for a
+    // signed-out (anonymous) viewer; the non-matching status survives.
+    expect(ids).not.toContain(urlToId(hiddenStatus.id))
+    expect(ids).toContain(urlToId(publicStatus.id))
+  })
+
   it.each([
     { description: 'junk opaque id', value: 'apurl_@@@@' },
     { description: 'percent signs', value: '%%%' },
--- a/app/api/v1/timelines/public/route.test.ts
+++ b/app/api/v1/timelines/public/route.test.ts
@@ -174,6 +174,72 @@ describe('GET /api/v1/timelines/public', () => {
     })
   })
 
+  describe('min_id/since_id wiring and prev Link', () => {
+    it.each([
+      { description: 'min_id', param: 'min_id' },
+      { description: 'since_id', param: 'since_id' }
+    ])(
+      'passes the decoded $description cursor to the timeline queries',
+      async ({ param }) => {
+        const spy = vi.spyOn(database, 'getTimeline').mockResolvedValue([])
+        const response = await GET(
+          request({ [param]: urlToId(publicStatus.id) }),
+          { params: Promise.resolve({}) }
+        )
+        expect(response.status).toBe(200)
+        expect(spy).toHaveBeenCalledWith(
+          expect.objectContaining({ minStatusId: publicStatus.id })
+        )
+      }
+    )
+
+    it('prefers min_id over since_id when both are provided', async () => {
+      const spy = vi.spyOn(database, 'getTimeline').mockResolvedValue([])
+      await GET(
+        request({
+          min_id: urlToId(publicStatus.id),
+          since_id: urlToId(`${ACTOR1_ID}/statuses/public-other`)
+        }),
+        { params: Promise.resolve({}) }
+      )
+      expect(spy).toHaveBeenCalledWith(
+        expect.objectContaining({ minStatusId: publicStatus.id })
+      )
+    })
+
+    it('emits a rel="prev" Link with a min_id cursor for a non-empty page', async () => {
+      vi.spyOn(database, 'getTimeline').mockResolvedValue([publicStatus])
+      const response = await GET(request(), { params: Promise.resolve({}) })
+      const link = response.headers.get('Link')
+      expect(link).toContain('rel="prev"')
+      const prevPart = link!
+        .split(', ')
+        .find((part) => part.includes('rel="prev"'))
+      const prevUrl = new URL(prevPart!.match(/<([^>]+)>/)![1])
+      expect(prevUrl.searchParams.get('min_id')).toBe(urlToId(publicStatus.id))
+    })
+  })
+
+  describe('only_media', () => {
+    it('passes onlyMedia=true to the timeline queries', async () => {
+      const spy = vi.spyOn(database, 'getTimeline').mockResolvedValue([])
+      await GET(request({ only_media: 'true' }), {
+        params: Promise.resolve({})
+      })
+      expect(spy).toHaveBeenCalledWith(
+        expect.objectContaining({ onlyMedia: true })
+      )
+    })
+
+    it('carries only_media into the pagination Links', async () => {
+      vi.spyOn(database, 'getTimeline').mockResolvedValue([publicStatus])
+      const response = await GET(request({ only_media: 'true' }), {
+        params: Promise.resolve({})
+      })
+      expect(response.headers.get('Link')).toContain('only_media=true')
+    })
+  })
+
   describe('merged default scope and Link scope', () => {
     const REMOTE_ACTOR = 'https://somewhere.test/users/bob'
 
--- a/app/api/v1/timelines/tag/[hashtag]/route.test.ts
+++ b/app/api/v1/timelines/tag/[hashtag]/route.test.ts
@@ -16,6 +16,9 @@ const mockDatabase = {
 const mockCurrentActor = {
   id: 'https://local.test/users/me'
 }
+// Overridable per test so the anonymous (signed-out) path can be exercised.
+let currentActorForRequest: typeof mockCurrentActor | undefined =
+  mockCurrentActor
 
 vi.mock('@/lib/services/guards/OAuthGuard', () => ({
   OptionalOAuthGuard:
@@ -33,7 +36,7 @@ vi.mock('@/lib/services/guards/OAuthGuard', () => ({
     (req: NextRequest, context: { params: Promise<{ hashtag: string }> }) =>
       handle(req, {
         database: mockDatabase,
-        currentActor: mockCurrentActor,
+        currentActor: currentActorForRequest,
         params: context.params
       }),
   corsErrorResponse: vi.fn()
@@ -53,6 +56,7 @@ const status = {
 describe('GET /api/v1/timelines/tag/:hashtag', () => {
   beforeEach(() => {
     vi.clearAllMocks()
+    currentActorForRequest = mockCurrentActor
     mockDatabase.getBlockRelations.mockResolvedValue([])
     mockDatabase.getMuteRelations.mockResolvedValue([])
     mockDatabase.getStatusesByHashtag.mockResolvedValue([status])
@@ -207,12 +211,15 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
       )
     })
 
-    it('passes the decoded min_id cursor to the hashtag query', async () => {
-      await requestWithQuery({ min_id: urlToId(status.id) })
-      expect(mockDatabase.getStatusesByHashtag).toHaveBeenCalledWith(
-        expect.objectContaining({ minStatusId: status.id })
-      )
-    })
+    it.each([{ param: 'min_id' }, { param: 'since_id' }])(
+      'passes the decoded $param cursor as minStatusId to the hashtag query',
+      async ({ param }) => {
+        await requestWithQuery({ [param]: urlToId(status.id) })
+        expect(mockDatabase.getStatusesByHashtag).toHaveBeenCalledWith(
+          expect.objectContaining({ minStatusId: status.id })
+        )
+      }
+    )
   })
 
   describe('keyword filters and prev Link', () => {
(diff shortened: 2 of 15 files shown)


app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,296 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+        new NextRequest('https://local.test/api/v1/timelines/tag/running'),
+        { params: Promise.resolve({ hashtag: 'running' }) }
+      )
+      expect(mockDatabase.getActiveFiltersForActor).toHaveBeenCalledWith({
+        actorId: mockCurrentActor.id,
+        context: 'public'
+      })
+    })
+
+    it('drops a status matched by a hide keyword filter before serializing', async () => {
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts:345]: 🤖 **Sub-agent review (tests), addressed.** The keyword-filter drop/annotate wiring was only ever run against empty filters. Fixed in 16077bb912c5386c747c2be06416781f73983d2f: a hide filter now drops a matching status before serialization (Mastodon serializer receives `[]`), and a separate warn-filter test asserts the response carries the `filtered` annotation (guarding `data: annotatedStatuses`).

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,150 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
...
+          database,
+          actorId: currentActor?.id,
+          maxStatusId,
+          limit: effectiveLimit,
+          // Hashtag pages use the public keyword-filter context: hide-filters
+          // drop rows, warn-filters annotate. Applied unconditionally so
+          // instance-wide server filters reach signed-out viewers too
+          // (getActiveFilters returns only server filters when actorId is
+          // undefined), per REVIEW.md's cross-view filtering invariant.
+          filterContext: 'public',
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.ts:138]: 🤖 **Sub-agent review (invariants), addressed.** The hashtag (and public) timeline used `filterContext: currentActor ? "public" : undefined`, so instance-wide server filters never ran for signed-out viewers — contradicting REVIEW.md (content filters apply to unauthenticated requests) and the anon status detail/context views. Fixed in 16077bb912c5386c747c2be06416781f73983d2f: both routes now pass `filterContext: "public"` unconditionally (getActiveFilters returns only server filters when actorId is undefined). Added an anon server-hide-filter test.

app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,296 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+
+    it('forwards only_media to the hashtag query', async () => {
+      await requestWithQuery({ only_media: 'true' })
+      expect(mockDatabase.getStatusesByHashtag).toHaveBeenCalledWith(
+        expect.objectContaining({ onlyMedia: true })
+      )
+    })
+
+    it.each([{ param: 'min_id' }, { param: 'since_id' }])(
+      'passes the decoded $param cursor as minStatusId to the hashtag query',
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts:310]: 🤖 **Sub-agent review (tests), addressed.** The route-local `minStatusId ?? sinceStatusId` collapse was only tested for `min_id`; a dropped `?? sinceStatusId` fallback would silently break `?since_id=` hashtag pagination. Fixed in 16077bb912c5386c747c2be06416781f73983d2f: the cursor test is now an it.each over min_id and since_id, both asserting minStatusId: status.id.

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,147 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
...
+    ...searchParams.getAll(key)
+  ]
+  const tags: string[] = []
+  for (const value of values) {
+    const name = value.replace(/^#+/, '')
+    if (!isMastodonHashtagName(name)) return null
+    tags.push(name)
+  }
+  return tags.slice(0, TAGS_PER_MODE_LIMIT)
+}
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/timelines/tag/[hashtag]/route.ts:73]: Done in 44fd92ff39416bfd89c5ae825d2675799228e35f. `parseAdditionalTags` now dedupes via a `Set` (then caps at 4), so duplicate tags no longer append redundant subqueries — especially useful for `all[]`. Added a test asserting `all[]=cycling&all[]=cycling&all[]=running` collapses to `["cycling","running"]`.

lib/database/sql/status.ts
```diff
@@ -2358,6 +2423,24 @@ export const StatusSQLDatabaseMixin = (
...
+          wb.where('statuses.createdAt', '>', cursor.createdAt).orWhere(
+            (wb2) => {
+              wb2
+                .where('statuses.createdAt', '=', cursor.createdAt)
+                .where('statuses.id', '>', minStatusId)
+            }
+          )
+        })
+      }
+    }
```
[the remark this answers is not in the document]
llun: Good catch on the keyset semantics — done in 44fd92ff39416bfd89c5ae825d2675799228e35f, with one correction: `maxStatusId` here did NOT return `[]`; it also used `if (cursor)` and silently fell back to the first page. So fixing only `minStatusId` would have made the two cursors inconsistent. I applied `if (!cursor) return []` to **both**, which also aligns this function with the repo-wide keyset convention (bookmark/block/follow/mute/scheduledStatus all do `if (!cursor) return []`). Added an it.each test asserting both cursors return `[]` for an unresolvable id. Existing pagination tests still pass.

app/api/v1/timelines/public/route.test.ts
```diff
@@ -90,6 +90,37 @@ describe('GET /api/v1/timelines/public', () => {
     ])
   })
 
+  it('applies instance-wide server hide filters for anonymous viewers', async () => {
```
llun, line 93: 🤖 **Sub-agent review (tests, medium), addressed.** The Round 1 `filterContext: "public"` fix (server filters apply to anonymous viewers) got a regression test on the tag route but not the public route — a revert would silently leak server-hidden content to signed-out users with a green public-route suite. Fixed in 4614eac64ca3d8a11633a1aec5a3f376edb406c2: this test seeds a server hide filter, keeps the session null (anon), and asserts the matching status is dropped while a non-matching one survives.

lib/database/sql/status.test.ts
```diff
@@ -2889,6 +2919,167 @@ describe('StatusDatabase', () => {
...
+        expect(withAll.map((status) => status.id)).toEqual([bothStatusId])
+
+        const withNone = await database.getStatusesByHashtag({
+          hashtag: baseTag,
+          noneTags: [extraTag]
+        })
+        expect(withNone.map((status) => status.id)).toEqual([baseOnlyStatusId])
+      })
+
+      it('requires every all[] tag to be present (AND across multiple tags)', async () => {
```
llun, line 3014: 🤖 **Sub-agent review (tests, medium), addressed.** `all[]` is the only tag-mode filter implemented as a per-tag `whereExists` loop (AND), but it was only tested with a single tag — a regression collapsing the loop to a single `whereIn` (OR) or keeping only the last iteration would silently over-match, yet pass. Fixed in 78629656cdf6f178ce25f1f70caeea17d47d4683: a multi-tag case seeds `base+A+B` and `base+A` and asserts only the both-tags status is returned.

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,151 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
...
+  key: 'any' | 'all' | 'none'
+): string[] | null => {
+  const values = [
+    ...searchParams.getAll(`${key}[]`),
+    ...searchParams.getAll(key)
+  ]
+  const tags = new Set<string>()
+  for (const value of values) {
+    const name = value.replace(/^#+/, '')
+    if (!isMastodonHashtagName(name)) return null
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/timelines/tag/[hashtag]/route.ts]: Good catch — done in ecc6452411ed3e4e428975f48f2342c02ed0b4e2. `parseAdditionalTags` now rejects a tag name longer than 255 chars (returns null → the route 400s), matching the primary hashtag's `.max(255)` cap, so an unbounded tag name can't reach the DB query. Added a regression test (`rejects an over-length additional tag name with 400`).

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,165 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
...
       }
+
+      // `any[]` is special-cased in Mastodon's HashtagQueryService: the primary
+      // hashtag is unioned into the OR-set *before* the LIMIT_PER_MODE cap
+      // (`tags_for(Array(tag.name) | Array(params[:any]))`), so the primary
+      // counts toward the four. Drop the primary and any duplicates, then keep
+      // at most three additional names, so the combined OR-set
+      // `[hashtag, ...effectiveAnyTags]` built by getStatusesByHashtag never
+      // exceeds four. (all[]/none[] stay capped on their own arrays.)
+      const effectiveAnyTags = Array.from(new Set([hashtag, ...anyTags]))
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.ts:116]: **[Round 4 review — correctness, fixed in this commit]** Mastodon spec off-by-one on `any[]`: `HashtagQueryService` builds the primary OR-set as `tags_for(Array(tag.name) | Array(params[:any]))`, i.e. it unions the **primary hashtag** into `any[]` *before* the `LIMIT_PER_MODE` (4) cap, so the primary counts toward the four. The previous code capped `any[]` at 4 on its own and then prepended the primary in `getStatusesByHashtag` (`[hashtag, ...anyTags]`), so the OR-set could reach **5** names — a status tagged only with the 4th `any` tag would match here but not on Mastodon.

Fixed by computing `effectiveAnyTags = unique([hashtag, ...anyTags]).take(4)` minus the primary, and using it for both the DB query and the pagination Links. `all[]`/`none[]` are unaffected (Mastodon caps those arrays without the primary). Regression tests added: `counts the primary hashtag toward the any[] cap` and `drops the primary hashtag from any[] to avoid double-counting`.

app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,360 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+      // Warn matches are kept but annotated: the response uses
+      // annotateMastodonStatusesWithFilters, not the raw serialized statuses.
+      expect(data).toHaveLength(1)
+      expect(data[0].filtered?.length ?? 0).toBeGreaterThan(0)
+    })
+
+    // buildLink shares one linkBaseParams between the prev and next Links, so
+    // asserting the prev Link's base params covers the shared param-building
+    // path (the any/all/none loop plus the local/remote/only_media setters).
+    it('emits a rel="prev" Link carrying every tag filter and scope param', async () => {
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts:500]: **[Round 4 review — test coverage, fixed in this commit]** The prev/next `Link` header shares one `linkBaseParams` built from the `any/all/none` loop plus the `local`/`remote`/`only_media` setters, but the only Link assertion previously exercised `any[]` alone — so a regression that dropped a mode from the loop or a scope setter would silently strip `all[]`/`none[]`/`only_media`/`local`/`remote` from page-2 Links while the suite stayed green.

Broadened this test to assert `all[]`, `none[]`, `only_media`, and `local` all survive into the prev Link, and added `carries the remote scope (not local) into the prev Link`. Since `buildLink` reuses `linkBaseParams` for both prev and next, this covers the shared param-building path.

app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,378 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+      expect(mockDatabase.getStatusesByHashtag).toHaveBeenCalledWith(
+        expect.objectContaining({
+          anyTags: ['cycling'],
+          allTags: ['fitness'],
+          noneTags: ['walking']
+        })
+      )
+    })
+
+    it('strips a leading # from an additional tag before querying', async () => {
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts:209]: **[Review — test coverage, fixed in this commit]** `parseAdditionalTags` had two untested load-bearing branches: the leading-`#` strip (`value.replace(/^#+/, "")`) and the bare-key fallback (`searchParams.getAll(key)` for `any=`/`all=`/`none=` without the `[]` suffix). The `#`-strip is load-bearing because `isMastodonHashtagName` (`/^[\p{L}\p{N}_]+$/u`) rejects `#`, so a `#cycling` additional tag would 400 the whole request without it; the bare-key fallback matches Mastodon's `Array(params[:any])`. Both were dead as far as the suite was concerned — removing either line kept it green. Added `strips a leading # from an additional tag before querying` and `accepts the bare-key (no []) additional tag form` to lock them in.

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,170 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
...
+    ...searchParams.getAll(key)
+  ]
+  const tags = new Set<string>()
+  for (const value of values) {
+    const name = value.replace(/^#+/, '')
+    // A blank value (e.g. `?any[]=` or a bare `#`) is an empty filter slot, not
+    // a malformed tag — skip it rather than 400 the whole request, matching
+    // parseTimelineQuery's blank-cursor coercion and Mastodon's lookup-not-reject
+    // behavior (a missing tag simply matches nothing).
+    if (!name) continue
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.ts:65]: **[Review — correctness, fixed in this commit]** A blank `any[]`/`all[]`/`none[]` value (e.g. `?any[]=`, or a bare `#` which strips to empty) made `parseAdditionalTags` return null → the route 400d the entire request, because the empty name fails `MASTODON_HASHTAG_NAME_REGEX`. That over-rejects a legitimately empty filter slot and diverges from the same file's `parseTimelineQuery` (which coerces blank cursors like `?max_id=` to absent to stay 2xx) and from Mastodon's HashtagQueryService (a missing tag simply matches nothing). Fixed by skipping empty names (`if (!name) continue`) before validation, so the timeline is served with that mode empty. Regression test: `ignores a blank additional tag value instead of 400ing`.

app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,401 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+      'passes the decoded $param cursor as minStatusId to the hashtag query',
+      async ({ param }) => {
+        await requestWithQuery({ [param]: urlToId(status.id) })
+        expect(mockDatabase.getStatusesByHashtag).toHaveBeenCalledWith(
+          expect.objectContaining({ minStatusId: status.id })
+        )
+      }
+    )
+
+    it('prefers min_id over since_id when both are provided', async () => {
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts:319]: **[Review — test coverage, fixed in this commit]** The route collapses both lower-bound cursors as `minStatusId ?? sinceStatusId` (deliberate min_id-wins precedence, matching the public timeline and opposite the home `[timeline]` route). The existing tests only supplied one cursor at a time, so a swapped operand order (`sinceStatusId ?? minStatusId`) would have passed identically. Added `prefers min_id over since_id when both are provided` to pin the precedence.

app/api/v1/timelines/tag/[hashtag]/route.test.ts
```diff
@@ -101,4 +110,414 @@ describe('GET /api/v1/timelines/tag/:hashtag', () => {
...
+    ])('returns 400 for a name containing $description', async ({ param }) => {
+      const response = await GET(
+        new NextRequest(`https://local.test/api/v1/timelines/tag/${param}`),
+        { params: Promise.resolve({ hashtag: param }) }
+      )
+      expect(response.status).toBe(400)
+      expect(mockDatabase.getStatusesByHashtag).not.toHaveBeenCalled()
+    })
+
+    it('returns 400 for an over-length primary hashtag', async () => {
```
Comment (llun) [inline app/api/v1/timelines/tag/[hashtag]/route.test.ts]: **[Review — test coverage, fixed in this commit]** The `Params` schema change (unbounded regex → `z.string().min(1).max(255)`) is new behavior and `.max(255)` is the only length guard on the primary path (`normalizeHashtagParam`/`isMastodonHashtagName` bound the alphabet, not the length). The additional-tag cap is pinned by `rejects an over-length additional tag name with 400`, but the primary path had no symmetric test — a refactor consolidating validation into the unbounded `normalizeHashtagParam` could let a huge name reach `getStatusesByHashtag` with the suite still green. Added `returns 400 for an over-length primary hashtag` (256-char param → 400, DB not called).

app/api/v1/timelines/tag/[hashtag]/route.ts
```diff
@@ -24,75 +29,165 @@ const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
 export const OPTIONS = defaultOptions(CORS_HEADERS)
 
 const Params = z.object({
-  hashtag: z.string().regex(/^[a-zA-Z0-9_]*[a-zA-Z_][a-zA-Z0-9_]*$/)
+  hashtag: z.string().min(1).max(255)
 })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/timelines/tag/[hashtag]/route.ts:36]: Good catch — fixed in 22e184b7da54397f8a58f46c4e8749498c45a02e. The raw `hashtag` param is now bounded by `MAX_ENCODED_HASHTAG_PARAM_LENGTH` (4096) rather than 255, and the 255-char limit is enforced on the **decoded** name inside `normalizeHashtagParam`. 4096 covers a full 255-codepoint name of 4-byte characters (255×4×3 = 3060 encoded chars) with margin, so valid Unicode tags like #日本語 are no longer prematurely rejected while an unbounded param still cannot reach `decodeURIComponent`. Added a regression test asserting a 30-char Japanese tag (270 encoded) returns 200.

lib/utils/text/mastodonHashtag.ts
```diff
@@ -0,0 +1,27 @@
...
+export const normalizeHashtagParam = (param: string): string | null => {
+  let decoded: string
+  try {
+    decoded = decodeURIComponent(param)
+  } catch {
+    return null
+  }
+  const name = decoded.replace(/^#+/, '')
+  return isMastodonHashtagName(name) ? name : null
+}
```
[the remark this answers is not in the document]
llun: Done in 22e184b7da54397f8a58f46c4e8749498c45a02e. `normalizeHashtagParam` now returns null when the decoded name exceeds `MAX_HASHTAG_NAME_LENGTH` (255) before the alphabet check, so an over-length decoded name can never reach the database even with the larger raw param bound. Added unit tests for the 255 boundary (255 → accepted, 256 → null) and a short-unicode-name-with-long-encoding case.

PR #1197, opened 2026-07-09
minor: persist actors.lastStatusAt for directory active order

llun:
## Summary

Persists a `lastStatusAt` column on the `actors` table so the Mastodon directory `GET /api/v1/directory?order=active` sort and the serializer's `last_status_at` field no longer run a live `MAX(statuses.createdAt) GROUP BY actorId` aggregation on every read. This is the future performance optimization noted for larger instances.

Note on scope: on this branch `order=active` and `order=new` both ordered by `actors.createdAt` (the earlier migration-free discovery polish left them identical, and PR #1195's live-aggregation ordering isn't on this branch). This PR makes `order=active` a real, distinct ordering by the new persisted column — so `active` now differs from `new` for the first time here.

## Changes

- **Migration** (`20260709120000_add_actor_last_status_at.js`): adds a nullable `lastStatusAt` (`timestamptz` on PG / `datetime` on SQLite) column, a `(domain, lastStatusAt)` index to serve the `active` query, and backfills existing rows from `statuses` with a correlated subquery (SQLite + PostgreSQL compatible; actors with no statuses stay `NULL`). Both reference schema dumps (`migrations/schema.sql`, `migrations/schema.sqlite.sql`) regenerated.
- **Maintenance** — atomic inside the existing status transactions:
  - *Create* (notes, announces, polls all funnel through `updateStatusCounters`): guarded set-if-newer keyed on the status's own `createdAt` (not `currentTime`), so backdated/federated/imported inserts stay correct and an out-of-order insert never lowers the value.
  - *Delete* (`deleteStatus`): after the rows are physically deleted, recompute `MAX(createdAt)` per affected actor via a correlated subquery (→ previous max, or `NULL` when none remain). `deleteActorData` removes the actor row, so it needs no maintenance.
  - Includes `Announce`/reblog rows, so `last_status_at` output is identical to the previous aggregation.
- **Reads**: `order=active` orders by `lastStatusAt DESC` (NULLs last via a portable `(lastStatusAt IS NULL)` sort key, `createdAt` tiebreak); `order=new` unchanged. The serializer and every single-actor / follower-list read now read the column, dropping several per-call `statuses` subqueries (including an N+1 in the follower-list path).
- Implements the previously no-op `updateActorLastStatusAt` against the real column.

## Testing

- `lib/database/sql/actorLastStatusAt.test.ts` (new): create sets/advances it, backdated insert doesn't lower it, Announce + Poll advance it, delete-of-newest recomputes to the previous status, delete-of-last clears to `NULL`, `updateActorLastStatusAt` is a guarded set-if-newer, and directory `order=active` vs `order=new` orderings (never-posted actors sort last; `new` ignores activity).
- `app/api/v1/directory/route.test.ts`: `order` pass-through (default `active`, explicit `new`/`active`) and rejection of an unknown `order`.
- Full gate green: `prettier --write .` → `lint` → `build` → `test` (618 files, 5634 tests). Migration verified applying on both local SQLite and local PostgreSQL 17 (backfill runs cleanly on both).

## Notes

- The Vitest suite builds test databases from the schema dumps (it does not run migrations), so the migration's one-time **backfill** isn't exercised by CI; it was verified by hand against a seeded local SQLite and confirmed to apply on a fresh local PostgreSQL 17.


lib/database/sql/status.ts
```diff
@@ -1747,6 +1765,44 @@ export const StatusSQLDatabaseMixin = (
...
+    )) {
+      // Lock the affected actor rows before recomputing (no-op on SQLite, which
+      // serializes writers; matches the pinStatus precedent). This recompute is
+      // an unconditional overwrite, so without the lock a concurrent create — whose
+      // guarded lastStatusAt bump holds the same row lock until it commits — could,
+      // under PostgreSQL READ COMMITTED, let this statement's stale snapshot
+      // clobber the just-committed newer value with an older MAX (or NULL). Taking
+      // the lock first serializes the two and forces the following UPDATE to take a
+      // fresh snapshot that already sees the new status.
+      await trx('actors').whereIn('id', actorIdChunk).select('id').forUpdate()
```
llun, line 1900: **[review round 1 — medium, fixed here]** Row lock added before the recompute `UPDATE`.

The recompute is an unconditional overwrite. Without this lock, a concurrent create — whose guarded `lastStatusAt` bump holds this same actor row lock until it commits — could, under PostgreSQL READ COMMITTED, let this statement's stale snapshot overwrite the just-committed newer `lastStatusAt` with an older `MAX(createdAt)` (or `NULL` when the deleter's snapshot can't see the new row), persisting a wrong value until the actor's next create/delete. Taking the lock first serializes the two operations and forces the following `UPDATE` to run on a fresh snapshot that already sees the new status. `forUpdate()` is a no-op on SQLite (serialized writers), matching the `pinStatus` precedent at status.ts:1388.

llun (commented):
### Sub-agent code-review round 1

Ran a multi-lens adversarial review (correctness/concurrency, cross-dialect SQL, behavioral parity, invariants/tests) with per-finding skeptical verification. 3 findings survived verification; all addressed in d45129b2.

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | medium | Delete-path `recomputeActorsLastStatusAt` overwrote `lastStatusAt` unconditionally, so under PostgreSQL READ COMMITTED a concurrent create could clobber a just-committed newer value with a stale MAX/NULL (a regression vs. the old read-time MAX). | Lock affected actor rows with `SELECT … FOR UPDATE` before the recompute (no-op on SQLite; matches the `pinStatus` precedent) so the two serialize and the recompute takes a fresh snapshot. |
| 2 | low | Migration backfill never exercised against pre-existing statuses (the suite builds from schema dumps, never runs the migration). | Added `actorLastStatusAtMigration.test.ts`: inserts rows, runs `up`, asserts per-actor MAX + NULL-for-never-posted + index + `down`. |
| 3 | low | No test covered the multi-actor delete recompute (cross-author reply subtree via `deleteStatus` without `actorId`). | Added a cross-actor reply-subtree delete test asserting each actor recomputes independently. |

False-positives / uncertain findings from other lenses were dropped after verification.

PR #1198, opened 2026-07-09
fix: remaining small Mastodon-API conformance gaps (batched)

llun:
## Phase 2.9 — remaining small Mastodon-API gaps (batched)

The final Phase 2 PR: twelve independent Mastodon-API conformance fixes, each strict-TDD (failing test → minimal fix → pass).

### Items
1. **filters** — accept the `blur` `filter_action` (Mastodon 4.4) in the domain + Mastodon enums; `FilterEditor` falls back to `warn` for a non-card action.
2. **list accounts** — `GET /lists/:id/accounts?limit=0` returns all members; default limit 40.
3. **poll votes** — coerce string choice indices in JSON vote bodies (Mastodon sends strings).
4. **v2 notifications** — `expand_accounts=partial_avatars` splits full vs partial accounts (`partial_accounts`); unknown value → 422.
5. **notification policy** — export `PUT` (alias of `PATCH`) + CORS.
6. **admin domain rule lists** — cursor pagination (`max_id`/`min_id`/`since_id`) + `Link` headers on domain_blocks and domain_allows; the offset/`X-Total-Count` extension is kept when no cursor is given.
7. **POST domain_blocks** — default severity `silence`; `422` with `existing_domain_block` when re-blocking a domain no more strictly than the existing rule.
8. **reports** — persist `collection_ids[]` (new `reports.collectionIds` column; **migration + both reference schema dumps regenerated**) and, on `forward=true`, fire-and-forget a `SendFlagJob` that federates a `Flag` to the reported actor's origin inbox, **signed by the headless instance actor for reporter anonymity** (skips local targets and domains federation is disabled with).
9. **accounts** — return `404` when unfollowing an unknown account (local `getActorFromId` check) instead of an all-false relationship.
10. **preferences** — report `posting:default:language` as `null` (not `''`) when unset.
11. **v1 notifications** — filter by `account_id` at the DB layer (`sourceActorId`) and emit `group_key` on each notification.
12. **list accounts POST** — `404` when the target is not followed.

### Migration
Item 8 adds `reports.collectionIds text NOT NULL DEFAULT '[]'` (mirrors `statusIds`/`ruleIds`). Both `migrations/schema.sql` (PG) and `migrations/schema.sqlite.sql` (SQLite) were regenerated against fresh local throwaway databases via the documented pipeline; each diff is exactly the new column.

### Testing
Full gate green locally: `yarn lint` 0 errors, `yarn build` pass, `yarn test` **627 files / 5817 tests pass**. Every item ships with tests (new/extended), including the Flag federation job (local-target skip, federation-policy gate, instance-actor signing).


app/api/v1/admin/domain_blocks/route.test.ts
```diff
@@ -189,4 +192,228 @@ describe('/api/v1/admin/domain_blocks', () => {
...
+        })
+      }),
+      { params: Promise.resolve({}) }
+    )
+
+    expect(response.status).toBe(200)
+    expect(mockDatabase.createDomainBlock).toHaveBeenCalled()
+  })
+
+  it('returns 422 for a non-stricter block under a covering wildcard rule', async () => {
```
llun, line 377: **[Round 1 review — test coverage, fixed here]** Adds the missing 422 branch: a non-stricter block under a covering wildcard (`*.covered.test`@suspend + POST `sub.covered.test`@silence → 422). Dropping the `!isDomainBlockStricter(...)` operand previously kept all tests green.

lib/jobs/sendFlagJob.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      reportId: 'report-1',
+      targetActorId: TARGET,
+      statusIds: [],
+      content: ''
+    })
+
+    expect(sendFlag).not.toHaveBeenCalled()
+  })
+
+  it('throws when sendFlag reports failure so the queue retries', async () => {
```
llun, line 110: **[Round 1 review — test coverage, fixed here]** Covers the retry path: `sendFlagJob` re-throws on `sendFlag` `{ ok:false }` so the queue retries. Previously only happy/skip paths were tested (beforeEach always mocked ok:true).

lib/activities/index.test.ts
```diff
@@ -400,6 +401,61 @@ describe('activities', () => {
     })
   })
 
+  describe('sendFlag', () => {
+    it('posts a Flag activity to the target actor inbox', async () => {
```
llun: **[Round 1 review — test coverage, fixed here]** Direct test of the new `sendFlag` builder — asserts the Flag activity shape (type/actor/object/content) and the `{actor}/inbox` fallback. Previously only reached via the fully-mocked job.

lib/components/filters/FilterEditor.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  id: 'f-blur',
+  title: 'Spoilers',
+  context: ['home'],
+  filter_action: 'blur',
+  expires_at: null,
+  keywords: []
+} as unknown as ClientFilter
+
+describe('FilterEditor', () => {
+  it('falls back to the warn card when the filter action has no card (e.g. blur)', () => {
```
llun: **[Round 1 review — test coverage, fixed here]** First test in `lib/components/filters/`: mounts FilterEditor with `filter_action: blur` and asserts the warn card is selected (fallback for a non-card action).

lib/services/notifications/getMastodonNotification.test.ts
```diff
@@ -101,6 +101,80 @@ describe('getMastodonNotification', () => {
...
+        sourceActorId: actor2Id,
+        isRead: false,
+        filtered: false,
+        createdAt: Date.now(),
+        updatedAt: Date.now()
+      })
+      expect(mastodonNotification?.group_key).toBe('ungrouped-gk-2')
+    })
+
+    it('reports ungrouped-<id> for a non-groupable type with a stored internal groupKey', async () => {
```
llun, line 134: **[Round 2 review — MEDIUM correctness, fixed in this commit]** v1 `group_key` surfaced the raw stored `groupKey`, but this app stores internal keys for non-groupable types (`mention:<id>`, `reply:<id>`, collections). Mastodon populates `group_key` only for default-groupable types (like/reblog/follow) → others are `ungrouped-<id>`, which is also what this app's own v2 endpoint returns for the same notification (so v1/v2 disagreed). Gated the raw path on `DEFAULT_GROUPABLE_TYPES`; already-grouped input keeps the grouping-step key (preserving explicit `grouped_types`). Tests: non-groupable-with-stored-key → `ungrouped-<id>`, and grouped input keeps its computed key.

lib/database/sql/admin.test.ts
```diff
@@ -533,3 +534,86 @@ describe('AdminDatabase', () => {
...
+        limit: 1,
+        sinceId: nextPage[0].id
+      })
+      expect(sincePage.map((block) => block.domain)).toEqual(['a.cursor.test'])
+    } finally {
+      await database.destroy()
+    }
+  })
+
+  it('pages domain allows with maxId, minId, and sinceId cursors', async () => {
```
llun, line 579: **[Round 2 review — test coverage, fixed in this commit]** `getDomainAllows` duplicates the block cursor logic but only `maxId` was tested; the `minId` branch orders desc then reverses, so a dropped reverse would silently return the prev page descending. Extended the allows test to assert `maxId`, `minId`, and `sinceId` pages, mirroring the blocks case.

lib/components/filters/FilterEditor.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  // via the API since Mastodon 4.4) opens as warn rather than leaving the
+  // radiogroup with nothing checked.
+  it.each([
+    {
+      description: 'keeps warn selected',
+      filterAction: 'warn',
+      checked: 'warn'
+    },
+    {
+      description: 'keeps hide selected',
```
llun, line 33: **[Round 3 review — test coverage, fixed in this commit]** The guarded action initializer has two branches: preserve a valid card action, and fall back to `warn` for a non-card action. The blur-only test exercised only the fallback. If the `ACTION_CARDS.some(...)` predicate regressed (e.g. compared `card.label` instead of `card.id`), a saved `hide` filter would open with "Hide with a warning" checked and silently downgrade on save — while the blur test stayed green. Made the test table-driven over warn/hide/blur so the preserve branch is pinned too.

app/api/v1/admin/domain_blocks/route.test.ts
```diff
@@ -189,4 +192,282 @@ describe('/api/v1/admin/domain_blocks', () => {
...
+
+    expect(response.status).toBe(422)
+    expect(data.existing_domain_block).toMatchObject({
+      domain: '*.covered.test',
+      severity: 'suspend'
+    })
+    expect(mockDatabase.createDomainBlock).not.toHaveBeenCalled()
+  })
+
+  it('allows an equal-severity block that adds reject_media under a covering wildcard', async () => {
```
llun, line 420: **[Round 4 review — correctness, fixed in this commit]** `isDomainBlockStricter` ranked only severity, ignoring `reject_media`/`reject_reports`. Mastodon's `DomainBlock#stricter_than?` also treats an equal-severity block as stricter when it does not relax those flags. Reachable failure: existing covering wildcard `*.example.com`@silence with `reject_media=false`; admin POSTs `sub.example.com`@silence with `reject_media=true` → we 422'd "already blocked" and never created the stricter block; Mastodon returns 200. Reimplemented the check to mirror `stricter_than?` (suspend always wins; lower severity never; otherwise must not relax either flag). Exact-domain re-blocks still 422 via the domain-equality guard. Added unit rows for add/drop of each flag plus this route test.

lib/activities/index.test.ts
```diff
@@ -400,6 +401,67 @@ describe('activities', () => {
...
+        type: 'Flag',
+        actor: actor1.id,
+        object: [targetId, statusId],
+        content: 'spam report'
+      })
+      expect(result.uri).toBe(`${actor1.id}#reports/report-1`)
+    })
+
+    it.each([
+      { description: 'is ok on a 202 Accepted inbox response', status: 202 },
```
llun, line 451: **[Round 5 review — test coverage, fixed in this commit]** `sendFlag` computes `ok: statusCode === 202`, and `sendFlagJob` throws (queue retry) when `ok` is false — but no test exercised the real 202 comparison (`sendFlagJob.test.ts` fully mocks `@/lib/activities`). Note the review's premise that the stub returns 202 for any `/inbox` path is only true on the **GET** branch; a POST to a per-actor inbox falls through to 404, so a bare `expect(result.ok).toBe(true)` would have failed. Instead the fallback test now drives the inbox response status itself and asserts `ok` for both 202 and a non-202 (200), pinning the mapping in both directions.

app/api/v1/polls/[id]/votes/route.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  it('still rejects non-numeric choices', async () => {
+    const response = await POST(createRequest({ choices: ['abc'] }), {
+      params: Promise.resolve({ id: urlToId(POLL_ID) })
+    })
+    expect(response.status).toBe(422)
+    expect(mockDatabase.recordPollVotes).not.toHaveBeenCalled()
+  })
+
+  it.each([
+    { description: 'an empty string choice', choice: '' },
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.test.ts:100]: **[Round 5 review — correctness, fixed in this commit]** `z.coerce.number().int().nonnegative()` coerces `""` and whitespace to **0** (verified against zod 4.4.3), so a JSON body `{"choices":[""]}` cleared every guard and silently recorded a vote for option index 0, returning 200. The form path already guarded this (`parseFormChoices` maps blank → `NaN` → fails `.int()` → 422), so JSON and urlencoded bodies diverged for the same input. The existing `["abc"]` test missed it because `"abc"` coerces to NaN (rejected) while `""` coerces to 0 (accepted). Fixed by preprocessing blank/whitespace strings to `NaN` before coercion; added `it.each` rows for `""` and `" "` asserting 422 with `recordPollVotes` not called.

app/api/v1/polls/[id]/votes/route.ts
```diff
@@ -23,14 +23,18 @@ const MAX_POLL_CHOICES_PER_VOTE = 20
 
 export const OPTIONS = defaultOptions(CORS_HEADERS)
 
+// Mastodon documents `choices` as an array of STRING indices, so accept numbers
+// and non-blank numeric strings alike. Everything else becomes NaN and 422s:
+// blanket `Number()` coercion maps '', ' ', null, false and [] all to 0, which
+// would silently record a vote for option 0. This mirrors parseFormChoices.
+const PollChoice = z.preprocess((value) => {
+  if (typeof value === 'number') return value
```
Comment (llun) [inline app/api/v1/polls/[id]/votes/route.ts:31]: **[Round 6 review — MEDIUM correctness, fixed in this commit]** My previous blank-string guard was incomplete: it mapped only whitespace/empty **strings** to NaN, so `Number()` still coerced `null`, `false` and `[]` to **0** (and `true` to 1). A JSON body `{"choices":[null]}` therefore cleared every guard, recorded a vote for option index 0 via `recordPollVotes`, federated it, and returned 200 — the exact outcome the previous commit set out to prevent. (Correctly noted: `z.coerce.number()` predates this PR, so this was an incomplete hardening, not a regression.) Now only numbers and non-blank numeric strings are coerced; everything else becomes NaN and 422s. Verified across null/false/true/[]/{}/negative/float/empty-array, with `" 1 "` still accepted to match `parseFormChoices`.

app/api/v2/notifications/route.test.ts
```diff
@@ -213,4 +213,93 @@ describe('GET /api/v2/notifications', () => {
...
+    const response = await GET(request, { params: Promise.resolve({}) })
+    const data = await response.json()
+
+    expect(response.status).toBe(200)
+    // Only the group's most recent account (alice) is rendered in full — and it
+    // keeps the fields a PartialAccountWithAvatar would have stripped, so a
+    // regression that truncated the full array too would fail here.
+    expect(data.accounts).toHaveLength(1)
+    expect(data.accounts[0].id).toBe('other.test:users:alice')
+    expect(data.accounts[0].display_name).toBe('Only in the full shape')
```
llun, line 271: **[Round 6 review — test coverage, fixed in this commit]** The `expand_accounts=partial_avatars` test seeded every account with the marker `display_name: "Only in the full shape"` to distinguish full from truncated entries, but the full-array assertions only checked length and id — so a regression mapping the full `accounts` array through `toPartialAccountWithAvatar` would still pass. Added `expect(data.accounts[0].display_name).toBe("Only in the full shape")` so truncating the full half fails the test.

lib/activities/index.test.ts
```diff
@@ -400,6 +402,81 @@ describe('activities', () => {
...
 
+  describe('sendFlag', () => {
+    it('posts a Flag activity to the inbox resolved from the target person', async () => {
+      if (!actor1) fail('Actor1 is required')
+      const targetId = 'https://somewhere.test/actors/test1'
+      const statusId = 'https://somewhere.test/actors/test1/statuses/1'
+      // Resolve a person whose inbox is deliberately NOT `${targetId}/inbox`,
+      // so this pins the getActorPerson lookup rather than the guessed
+      // fallback (actors on subpath/group deployments have custom inboxes).
+      const resolvedInbox = 'https://somewhere.test/custom-delivery/test1'
```
llun, line 413: **[Round 7 review — test coverage, fixed in this commit]** `sendFlag` posts to `person?.inbox ?? `${targetActorId}/inbox``, but `MockActivityPubPerson` emits `${id}/inbox` — byte-identical to the fallback. Both sendFlag tests asserted that shared string, so a regression removing the `getActorPerson` lookup and always guessing the fallback would have passed both while misdelivering Flag reports to any actor with a non-conventional inbox (group actors, subpath/Pleroma deployments). The success test now resolves a person whose `inbox` is deliberately different from the fallback and asserts the POST targets it, so the resolution branch is genuinely pinned.

PR #1201, opened 2026-07-10
minor: add v1 filters compatibility shim over v2 filter storage

llun:
## Summary

Serves the deprecated-but-still-used Mastodon v1 filters API (`/api/v1/filters`, `/api/v1/filters/:id`) as a pure **view over the existing v2 filter storage**, the same way Mastodon itself shims v1 onto its v2 `CustomFilter`/`CustomFilterKeyword` models. The v1 API predates multi-keyword filters, so each v1 "Filter" row is really **one keyword of a v2 filter**: the v1 `id` is the **keyword id**, `phrase`/`whole_word` come from the keyword, and `context`/`expires_at`/`irreversible` come from the parent filter (`irreversible === (filter_action === 'hide')`). No new tables and no migration — everything reads and writes the existing `filters` / `filter_keywords` tables through the existing `FilterDatabase` methods.

## Endpoints / behavior

- `GET /api/v1/filters` (`read` | `read:filters`) — lists all of the caller's own v2 filters via `getFilterRecordsForActor` (**including expired**, matching Mastodon v1) and flat-maps keywords → one V1 row per keyword (row id = keyword id). Instance-wide server filters are intentionally excluded (unlike the v2 GET).
- `POST /api/v1/filters` (`write:filters`) — creates a v2 filter with exactly one keyword (`title` = `phrase`, `filter_action` = `irreversible ? 'hide' : 'warn'`, keyword `{ keyword: phrase, wholeWord }`); accepts JSON and form-encoded bodies (including repeated `context[]` params). 422 on missing/blank `phrase`, missing/empty `context`, or unparseable `expires_in`.
- `GET /api/v1/filters/:id` (`read` | `read:filters`) — `:id` is the keyword id; 404 if not owned. Returns the V1 view of that keyword.
- `PUT` / `PATCH /api/v1/filters/:id` (`write:filters`) — requires `phrase` and `context`. Updates `phrase`/`whole_word` on the keyword and `phrase`(→title)/`context`/`irreversible`/`expires_in` on the parent. Enforces Mastodon's documented **422 guard** against changing parent attributes (`context`/`irreversible`/`expires_in`/phrase→title) of a filter that has multiple keywords, checked before any write; a sibling-keyword rename collision (`'duplicate'`) also returns 422. The keyword rename and parent update are two sequential non-atomic writes (the DB layer exposes no cross-entity transaction) — documented in a code comment.
- `DELETE /api/v1/filters/:id` (`write:filters`) — deletes the keyword; when it was the filter's last keyword, deletes the parent filter via `deleteFilter` (whose `cascadeDeleteFilter` atomically removes `filter_statuses` + `filter_keywords` + `filters`), otherwise `deleteFilterKeyword`. Returns `{}`.

Entity/serializer changes:

- New `V1Filter` Zod schema + inferred type in `lib/types/mastodon/filter/v1.ts` (re-exported from the `@/lib/types/mastodon` barrel).
- New `getV1Filter(filter, keyword)` serializer in `@/lib/services/mastodon/getMastodonFilter` — pure, DB-free, maps a `Pick<Filter, 'context' | 'expiresAt' | 'filterAction'>` + `FilterKeyword` to the V1 view.
- New v1 input parsers `parseV1FilterCreateInput` / `parseV1FilterUpdateInput` (with tri-state update semantics) beside the v2 ones in `lib/services/filters/parseFilterInput.ts`.
- `parseFilterBody` now also parses `application/x-www-form-urlencoded` bodies via `URLSearchParams` (previously `formData()`-only), which also benefits the existing v2 routes; the JSON and multipart paths are unchanged.

Error bodies use the repo convention (`ERROR_404` = `{ status: 'Not Found' }`, `ERROR_422` = `{ status: 'Unprocessable entity' }` from `@/lib/utils/response`), matching every existing v1/v2 route — not Mastodon's `{ error: … }` shape.

Docs: `docs/features.md` now notes the v1 filters compatibility shim alongside `/api/v2/filters`. `docs/mastodon-api-compatibility.md` needs no change — v1 filters behaves exactly as documented upstream, so it is neither a divergence, an unplanned endpoint, nor an extension.

## Test plan

Ran the full pre-commit gate from the worktree root, in order:

- `yarn run prettier --write .` — no changes.
- `yarn build` — succeeds; `/api/v1/filters` and `/api/v1/filters/[id]` are registered.
- `yarn test` — **633 test files / 5888 tests, all passing.** New coverage: `app/api/v1/filters/route.test.ts` (6 tests) and `app/api/v1/filters/[id]/route.test.ts` (12 tests), plus updated `lib/services/filters/parseFilterInput.test.ts`.

No migration, so `git diff main --stat -- migrations/` is empty and the reference schema dumps are untouched.

## Deviations from the plan

- Path translation only: the plan's code snippets referenced a different worktree path (`.../mystifying-mahavira-53831f`); every path was translated to the actual worktree. No behavioral change.
- Prettier (the mandated `yarn run prettier --write .` gate step) reflowed a few expressions in the new `route.ts`/`route.test.ts` files — e.g. a test `it.each` entry expanded to a multi-line object, and two route expressions collapsed onto single lines. Formatting only; no behavior change.
- The husky pre-commit hook was skipped by git during some task commits because the hook file was not marked executable (an advisory git hint). Harmless: the full prettier → lint → build → test gate was run manually in order before each commit, so the hook's work was already done.

## Out of scope

Found while reading, intentionally not addressed here:

- `lib/utils/getRequestBody.ts` flattens repeated form keys (`Object.fromEntries(new URLSearchParams(...))` keeps only the last value), so any route relying on it can never receive array params like `context[]`. Nothing currently breaks (its consumers use scalar params), but it is a latent trap; the filters stack avoids it via `parseFilterBody`.
- `GET /api/v2/filters` merges read-only instance-wide **server filters** (flagged `server: true`) into the account list. The v1 list intentionally excludes them: v1 rows must be addressable and editable by keyword id, and server-filter keywords are neither owned by the caller nor editable. If server-rule visibility for v1-only clients ever matters, that is a separate PR.
- The v2 keyword-update path (`updateFilter` with `keywords_attributes`) silently skips duplicate-keyword renames inside its savepoint, while the standalone keyword route surfaces `'duplicate'` as 422 — an existing inconsistency, untouched here.


llun (commented):
## Code review — round 1

Sub-agent review of the v1 filters shim (`feat/v1-filters-shim`). Per AGENTS.md every finding this round is posted inline as the source of truth: **1 confirmed** defect and **1 refuted** claim (posted resolved-with-rationale so the reasoning is on the record).

- **[1·confirmed]** `app/api/v1/filters/route.ts:108` — oversized `expires_in` persists a bad row then throws `RangeError` at serialization → HTTP 500.
- **[1·not-a-defect]** `app/api/v1/filters/route.ts:61` — malformed-JSON `catch`→422 branch is uncovered by tests; refuted (not a defect; matches repo-wide convention for defensive error paths).

See inline comments for the full failure scenarios and rationale.

app/api/v1/filters/route.ts
```diff
@@ -0,0 +1,112 @@
...
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_500,
+          responseStatusCode: HTTP_STATUS.INTERNAL_SERVER_ERROR
+        })
+      }
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: getV1Filter(filter, keyword)
```
llun, line 108: **[1·confirmed]** A large `expires_in` is accepted, persisted, then crashes serialization with an unhandled `RangeError`, returning HTTP 500 and leaving an out-of-range filter row committed in the DB.

**Failure scenario:** Client sends `POST /api/v1/filters` with `{phrase:'x', context:['home'], expires_in:'99999999999999'}`. `parseV1FilterCreateInput` → `parseV1FilterBase` → `resolveExpiresAt` computes `now + 99999999999999*1000 ≈ 1.0e17` ms, which is *finite* so it passes validation (not `INVALID_EXPIRES`). `createFilter` (route.ts:82) commits the filter with that out-of-`Date`-range `expiresAt`. Then `getV1Filter(filter, keyword)` at line 108 calls `getISOTimeUTC(1.0e17)`, where `format(new UTCDate(1.0e17))` throws `RangeError: Invalid time value`. That throw is **outside** the `try/catch` (which only wraps `parseFilterBody`, lines 59-68), so Next.js returns HTTP 500 instead of a clean 422 — and a filter with an invalid expiry is already committed.

Verified empirically with the repo's `@date-fns/utc` + `date-fns`: `parseV1FilterCreateInput` returns `expiresAt ≈ 1.00001783679816160e17` (finite), and `getISOTimeUTC(...)` throws `RangeError: Invalid time value`. The same path exists in `PUT /api/v1/filters/[id]` (`[id]/route.ts:194`) for a single-keyword filter, and mirrors a pre-existing latent issue in the v2 filters routes.

**Root cause:** `resolveExpiresAt` in `lib/services/filters/parseFilterInput.ts` (~line 180) does not bound `expires_in` to a valid `Date` range.

**Suggested fix:** In `resolveExpiresAt`, reject expiry timestamps outside the valid `Date` range (e.g. `Math.abs(ms) <= 8.64e15`) by returning the `INVALID_EXPIRES` sentinel → 422, so an oversized `expires_in` yields a clean 422 *before* any DB write instead of a persisted bad row plus an unhandled `RangeError` at serialization.

app/api/v1/filters/route.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const POST = traceApiRoute(
+  'createV1Filter',
+  OAuthGuard(
+    [Scope.enum['write:filters']],
+    async (req: NextRequest, { database, currentActor }) => {
+      let rawBody: unknown
+      try {
+        rawBody = await parseFilterBody(req)
+      } catch {
```
llun, line 61: **[1·not-a-defect]** Claim was: the malformed-JSON `catch`→422 branch (lines 61-68) is never exercised by any test in the new v1 filter routes.

**Refuted — resolved with rationale.** The description of the code is accurate: `parseFilterBody` uses `JSON.parse`, which throws `SyntaxError` on a malformed `application/json` body; the POST (here) and PUT handlers catch it and return 422; and the test helpers only ever send `JSON.stringify(init.body)` (valid JSON), so this branch has no direct coverage. But that is not a defect:

1. Under the repro lens there is no wrong output — the code is correct and returns 422 as intended. No concrete input yields a wrong result or unhandled exception. The failure scenario is explicitly a hypothetical *future* regression, a property true of every untested line, not a present bug.
2. It matches a repo-wide convention: the malformed-body→422 defensive catch is untested everywhere. The analogous `app/api/v1/lists/route.ts` uses `getRequestBody(req).catch(()=>null)`→422 and its `route.test.ts` has zero 422/catch coverage; multiple routes use that pattern and none test the branch. It is a sibling to this route's own `ERROR_500` "unreachable unless storage is inconsistent" branch (line 95).
3. The PR already tests every substantive 422 the plan specifies (missing phrase, missing context, duplicate-sibling keyword, multi-keyword parent guard). The plan's 422 contract is "missing/blank phrase or context," not malformed JSON.

No change requested.

app/api/v1/filters/route.ts
llun: Fixed in a1a1390d. resolveExpiresAt (lib/services/filters/parseFilterInput.ts) now bounds the resolved timestamp to the ECMAScript time-value limit (new MAX_TIMESTAMP_MS = 8_640_000_000_000_000) via an isTimestampInRange guard, applied on both the numeric expires_in path and the expires_at string path (which also folds in the old NaN check). An out-of-range expiry now returns the INVALID_EXPIRES sentinel, so the route responds 422 before any createFilter/updateFilter write. No out-of-range row is committed, and getV1Filter -> getISOTimeUTC no longer reaches date-fns format with an invalid time value, so no more RangeError -> 500. The shared helper backs the v2 filter routes too, so their identical latent path is fixed as well. Regression tests added and verified failing pre-fix / passing post-fix: parseV1FilterCreateInput and parseV1FilterUpdateInput reject an out-of-range expires_in (parseFilterInput.test.ts), and POST /api/v1/filters with a huge expires_in returns 422 and persists no row (route.test.ts).

app/api/v1/filters/route.ts
llun: Confirmed not-a-defect (full rationale above). The malformed-JSON catch -> 422 branch is correct and returns 422 as intended; no concrete input yields wrong output or an unhandled exception, so there is no present bug -- only an uncovered defensive branch, which matches repo-wide convention (e.g. app/api/v1/lists/route.ts uses getRequestBody(req).catch(() => null) -> 422 with zero catch-branch coverage). The PR already tests every substantive 422 the plan specifies. No code change; resolving.

app/api/v1/filters/[id]/route.ts
```diff
@@ -0,0 +1,238 @@
...
+      // Mirror Mastodon's v1 guard: parent-filter attributes (title/phrase,
+      // context, irreversible, expires_in) may only change while the filter
+      // has a single keyword — otherwise a v1 client would silently rewrite
+      // a v2 filter shared by its sibling keywords. The check runs before
+      // any write so a rejected request changes nothing.
+      const parentWouldChange =
+        input.phrase !== filter.title ||
+        contextsDiffer(input.context, filter.context) ||
+        (filterAction !== undefined && filterAction !== filter.filterAction) ||
+        (input.expiresAt !== undefined && input.expiresAt !== filter.expiresAt)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/filters/[id]/route.ts]: Confirmed real defect and fixed in 1eaf54e1. Reproduced empirically: a two-keyword filter (title `Fruits`, keywords `apple`/`banana`), PUT on the `banana` keyword with `phrase: 'banana'` (the keyword's own current phrase, which a v1 client must resend) returned 422 — `input.phrase !== filter.title` was always true for a secondary keyword, so `parentWouldChange` fired and no secondary keyword could ever be updated (not even a `whole_word` toggle).

Fix: I dropped the `input.phrase !== filter.title` term from the guard entirely rather than switching it to `row.keyword.keyword`. The authoritative docs (https://docs.joinmastodon.org/methods/filters/#v1) say the guard rejects changes to `expires_in`, `irreversible`, or `context` for a multi-keyword filter — the keyword phrase is not in that list, because a v1 phrase change only renames that one keyword. Comparing against `row.keyword.keyword` would have fixed the reported symptom but still wrongly 422'd a legitimate secondary-keyword *rename* (e.g. `banana` -> `berry`) on a multi-keyword filter, which Mastodon allows. The guard now checks only the shared parent attributes (context / filter_action / expires_at). Added regression tests `updates a secondary keyword of a multi-keyword filter without touching the shared parent` and `renames a secondary keyword of a multi-keyword filter without changing the parent`, both of which fail on the pre-fix code and pass now.

app/api/v1/filters/[id]/route.ts
```diff
@@ -0,0 +1,238 @@
...
+        filterAction,
+        expiresAt: input.expiresAt
+      })
+      if (!updatedFilter)
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_404,
+          responseStatusCode: HTTP_STATUS.NOT_FOUND
+        })
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/filters/[id]/route.ts]: Confirmed real defect and fixed in 1eaf54e1, matching your suggestion. The parent `updateFilter` (which writes `title: input.phrase`) now runs only when the filter has a single keyword. For a multi-keyword filter the shared parent is left untouched — its title belongs to the v2 client, and the guard above already rejects any real change to its shared attributes. Reproduced empirically first: without this, a v2-created filter titled `Fruits` had its title silently overwritten by the edited keyword's phrase.

On the second point (explicit conditional spreads vs relying on the DB layer's `undefined` handling): I left the direct pass-through because `updateFilter` in `lib/database/sql/filter.ts` already handles this explicitly and by contract — `title: title ?? existing.title`, `context: context ?? existing.context`, `filterAction: filterAction ?? existing.filterAction`, and `expiresAt: expiresAt === undefined ? existing.expiresAt : expiresAt` (so `undefined` keeps the stored value and `null` clears the expiry). That distinction between `undefined` and `null` is load-bearing for the v1 "clear expiry" path, so the direct call is intentional here rather than a reliance on incidental behavior. The single-keyword branch is unchanged in intent otherwise.

app/api/v1/filters/[id]/route.test.ts
```diff
@@ -0,0 +1,458 @@
...
+      actorId: ACTOR1_ID,
+      id
+    })
+    expect(storedKeyword).toMatchObject({
+      keyword: 'new-phrase',
+      wholeWord: true
+    })
+  })
+
+  it('returns 422 when changing parent attributes of a multi-keyword filter and writes nothing', async () => {
```
Comment (llun) [inline app/api/v1/filters/[id]/route.test.ts:207]: **[3·confirmed]** [low/test-coverage] The multi-keyword parent-change guard is only proven to return 422 via its `filter_action`/`irreversible` disjunct; its `context` and `expires_in` disjuncts have no discriminating 422 test.

The route's `parentAttributesWouldChange` guard (`route.ts:143-146`) is a three-way OR over context, filter_action and expires_in, matching Mastodon's documented v1 contract ("error if you attempt to change `expires_in`, `irreversible`, or `context` for a filter with multiple keywords"). But only the filter_action branch is exercised for a 422: this test sends `context: ['home']` identical to the stored context (createFilterFixture default, line 86) and no `expires_in`, so `contextsDiffer(...)` is false and the `input.expiresAt !== filter.expiresAt` term is skipped — the sole trigger is `irreversible: true` → `'hide' !== 'warn'`. No test anywhere in the file sends a *changed context* or an *expires_in* to a multi-keyword filter expecting 422; the post-assertion here checks `{ filterAction: 'warn' }`, not context/expiry. The multi-keyword 200-expecting tests (237, 267, 308) all keep `context: ['home']` and omit expires_in, so they only catch a guard that *over*-fires, never one that *under*-fires on these two attributes.

**Failure:** if a future refactor dropped the `contextsDiffer(...)` term or the `input.expiresAt !== filter.expiresAt` term (or `contextsDiffer` regressed to return false when contexts differ), a v1 PUT could silently rewrite a shared multi-keyword filter's context or expiry and return 200 instead of 422, and the entire suite would still pass.

**Suggested fix:** add two `it.each` rows to the multi-keyword guard test — one PUT changing only context (stored `['home']`, body `['home','public']`, matching phrase, no irreversible) expecting 422, and one changing only `expires_in` expecting 422 — each asserting the stored parent is unchanged.

llun (commented):
Round 3 sub-agent code review — 1 finding.

Round 2 was a clean round (all reviewers, zero findings). Round 3 surfaced one low-severity test-coverage gap, upheld by a majority of 3 independent skeptics and verified against the code and the Mastodon v1 contract in the plan (which documents the guard covering context/irreversible/expires_in for multi-keyword filters). No new correctness, invariant, or hard-project-rule violations were found. Not approving; posting for the author to address.

**1 confirmed** (inline below):
- `app/api/v1/filters/[id]/route.test.ts:207` — the multi-keyword parent-change guard's `context` and `expires_in` disjuncts are never exercised for a 422; only the filter_action/irreversible branch is, so dropping either term would keep the suite green.

**0 refuted.**

app/api/v1/filters/[id]/route.test.ts
Comment (llun) [reply app/api/v1/filters/[id]/route.test.ts:207]: Fixed in 08dd0803. Added two discriminating `it.each` rows to the multi-keyword guard test in `app/api/v1/filters/[id]/route.test.ts` that trip the `contextsDiffer` and `expiresAt` disjuncts in isolation — the prior lone 422 test only tripped the `filter_action` disjunct (context `['home']` identical to stored, no `expires_in`). One row changes only context (`['home']` -> `['home','public']`), the other changes only `expires_in` (3600 on a null-expiry parent); each also carries a genuine phrase rename so a route that dropped the relevant guard term would return 200 and rename the keyword instead of 422 — proving the 422 comes from that specific disjunct, not the phrase change. Each row asserts the keyword was NOT renamed and the parent attribute it tried to change was left intact. Verified by temporarily deleting each guard term and observing exactly the matching new row fail while all others stayed green; the full file (16 tests) passes on the fixed route.

app/api/v1/filters/route.ts
```diff
@@ -0,0 +1,112 @@
...
+        return apiResponse({
```
llun, line 62: **[4·confirmed]** [low/test-coverage] The malformed-body 422 branch is never exercised in either the POST create route (`route.ts:61-67`) or the PUT/PATCH update route (`app/api/v1/filters/[id]/route.ts:103-110`).

`parseFilterBody` does `JSON.parse(text)` for an `application/json` content-type (`lib/services/filters/parseFilterInput.ts:151`), which throws a `SyntaxError` on a malformed body; both routes wrap it in `try { … } catch { → ERROR_422 }`. But every `application/json` request in the tests is built via the `jsonRequest` (`route.test.ts:61-72`) / `requestFor` (`[id]/route.test.ts:62-73`) helpers, which always `JSON.stringify(init.body)` a valid object. The existing 422 tests (out-of-range `expires_in`, missing `phrase`/`context`) all hit the *other* path where `parseV1Filter*Input` returns `null` — never the `JSON.parse`-throws path. A grep of both test files finds no raw-string body.

Consequence: a refactor that narrowed or removed this `try/catch` would let the `SyntaxError` propagate as an HTTP 500 on a garbage request body, and the whole suite would stay green.

Suggested fix: add a discriminating case (an `it.each` row next to the existing missing-phrase / missing-context 422 rows in both `route.test.ts` and `[id]/route.test.ts`) that sends `Content-Type: application/json` with an invalid body such as `'{ not json'` and asserts status `422` (and, for POST, that no filter row was persisted). Cf. `app/api/v1/markers/route.test.ts:221` (`body: 'not json'`) as an in-repo precedent for the malformed-JSON case.

llun (commented):
Round 4 sub-agent code-review loop. This round surfaced one CONFIRMED finding (low-severity test-coverage gap), verified by a majority of independent skeptics and re-checked against the route code and both route test files. Round 2 was already a genuine clean round; the two OPEN gemini threads on the multi-keyword update path were separately re-examined and the current code (`parentAttributesWouldChange` deliberately excludes the keyword's own `phrase`/`whole_word`, and the parent title is only rewritten for single-keyword filters) matches the Mastodon v1 contract — the `updates a secondary keyword…` and `renames a secondary keyword…` tests in `[id]/route.test.ts` pin exactly the behavior those threads worried about. One inline finding below.

app/api/v1/filters/route.ts
llun: Fixed in 0f353af4. Confirmed real: both route test helpers (`jsonRequest` in `route.test.ts`, `requestFor` in `[id]/route.test.ts`) only `JSON.stringify` a valid object, so every existing 422 test hit the field-validation path (`parseV1Filter*Input` returns null) and never the `JSON.parse`-throws path — a refactor narrowing the try/catch would have turned a garbage body into a 500 with the suite still green.

Added a regression test to each co-located suite that sends a raw malformed body (`'{ not json'`) with `Content-Type: application/json` and asserts 422. Because a malformed body is a raw string (not an object the helper can build), each is a standalone `it` with a raw `NextRequest` rather than an `it.each` row. The POST test also asserts the v2 filter count is unchanged (no row persisted); the PUT/PATCH test asserts the addressed keyword is left unchanged. POST suite 7->8 tests, `[id]` suite 16->17 tests, both green (25 total, verified).

app/api/v1/filters/[id]/route.ts
```diff
@@ -0,0 +1,251 @@
...
+      // filter has multiple keywords — otherwise a v1 client would silently
+      // rewrite a v2 filter shared by its sibling keywords
+      // (https://docs.joinmastodon.org/methods/filters/#v1 lists exactly those
+      // three). The keyword's own `phrase`/`whole_word` are not shared, so they
+      // are deliberately excluded here and may always change — including a
+      // secondary keyword whose phrase differs from the parent title. The check
+      // runs before any write so a rejected request changes nothing.
+      const parentAttributesWouldChange =
+        contextsDiffer(input.context, filter.context) ||
+        (filterAction !== undefined && filterAction !== filter.filterAction) ||
```
Comment (llun) [inline app/api/v1/filters/[id]/route.ts]: **[5·confirmed]** [low/mastodon-conformance] The multi-keyword PUT guard compares the tri-state `filterAction` against the two-state v1 `irreversible` projection, so a `blur`-action filter is wrongly rejected with 422 on a plain keyword rename.

`getV1Filter` (`lib/services/mastodon/getMastodonFilter.ts:145`) projects `irreversible = filter.filterAction === 'hide'`, so both `warn` and `blur` filters present to a v1 client as `irreversible: false`. When a v1 client renames a keyword and echoes the filter's own projected value — `irreversible: false` — lines 121-126 derive `filterAction = 'warn'`, and this disjunct evaluates `'warn' !== 'blur'` → `true`. With `isMultiKeyword` true, `parentAttributesWouldChange && isMultiKeyword` fires and the route returns 422.

The Mastodon v1 contract (https://docs.joinmastodon.org/methods/filters/#v1, cited in the plan at pr-3.2.md:1008) only errors when the client *changes* `irreversible`/`context`/`expires_in`. Here the v1-visible `irreversible` stays `false`, so the correct result is 200 with the keyword renamed. `blur` is a valid `FilterAction` (`lib/types/domain/filter.ts:12`) reachable via v2 create, and the multi-keyword branch (line 185) never writes the parent, so relaxing the guard is safe.

Suggested fix: compare the projected boolean rather than the tri-state action — replace the `filterAction !== filter.filterAction` disjunct with something like `input.irreversible !== undefined && input.irreversible !== (filter.filterAction === 'hide')`, so `blur` and `warn` (both `irreversible=false`) are treated as no change to `irreversible`.

app/api/v1/filters/[id]/route.test.ts
```diff
@@ -0,0 +1,553 @@
...
+    })
+    expect(storedFilter).toMatchObject({ filterAction: 'warn' })
+  })
+
+  // Exercises the context and expires_in disjuncts of the multi-keyword guard
+  // in isolation (the test above only trips the filter_action disjunct). Each
+  // body leaves the keyword's own phrase changed so a route that dropped the
+  // relevant guard term would return 200 and rename the keyword — proving the
+  // 422 here comes from that specific disjunct, not from the phrase change.
+  it.each([
```
Comment (llun) [inline app/api/v1/filters/[id]/route.test.ts:242]: **[5·confirmed]** [low/test-coverage] The order-insensitive `.sort()` in `contextsDiffer` (`app/api/v1/filters/[id]/route.ts:55-60`) is never exercised by a discriminating test, so a regression removing it would ship green.

Every multi-keyword guard case sends either a single-element context (`['home']` vs stored `['home']`, where the sort is a no-op) or a length-mismatched context (`['home','public']` vs stored `['home']`, which hits the `a.length !== b.length` early return before the sort). The multi-element `['home','notifications']` at line 171 is on a single-keyword fixture (line 160-163), so it never reaches the guard. No test sends a reordered-but-equal multi-element context on a multi-keyword filter that expects 200.

Failure: a refactor drops `.sort()` and compares element-wise. A v1 client GETs a multi-keyword filter stored with context `['home','public']`, toggles `whole_word`, and PUTs it back with context `['public','home']` (a no-op reorder). The route now computes `contextsDiffer === true`, sees `isMultiKeyword`, and returns 422 — rejecting a legitimate update Mastodon allows. All current tests still pass.

Suggested test: add a row that creates a multi-keyword filter with context `['home','public']`, PUTs `{ phrase: <the same keyword text>, context: ['public','home'], whole_word: true }`, expects 200, and asserts the keyword's `wholeWord` was updated (proving the reordered context is treated as unchanged so the guard does not fire). Confirm the new row fails if `.sort()` is removed from `contextsDiffer` while the rest of the suite stays green.

llun (commented):
**Round 5 code review — PR #1201 (v1 filters shim)**

2 confirmed findings from round 5, each upheld by a majority of independent skeptics. Both are low-severity issues on the multi-keyword v1 filter `PUT` guard in `app/api/v1/filters/[id]/route.ts`; neither corrupts data nor violates a hard project invariant.

- `app/api/v1/filters/[id]/route.ts:145` — the multi-keyword guard rejects a legitimate v1 keyword rename on a `blur`-action filter with a spurious 422.
- `app/api/v1/filters/[id]/route.test.ts:242` — the `contextsDiffer` sort normalization is never exercised by a discriminating test; a regression removing it would ship green.

Round 2 was already a clean round, and no round-5 findings were refuted, so there is nothing to record under a not-a-defect heading this round. Details inline.

app/api/v1/filters/[id]/route.ts
Comment (llun) [reply app/api/v1/filters/[id]/route.ts]: Fixed in a4b45089. The multi-keyword PUT guard no longer compares the derived tri-state filterAction against the stored filterAction; it now compares the two-state v1 projection directly: `input.irreversible !== undefined && input.irreversible !== (filter.filterAction === 'hide')`. So a `blur`-action filter (which projects to the same v1 `irreversible: false` as `warn`) is no longer spuriously 422'd when a v1 client echoes its own `irreversible: false` on a plain keyword rename. Verified across all three actions (hide/warn/blur): only an actual change to the v1-visible `irreversible` now trips the guard, matching the Mastodon v1 contract. `filterAction` is still derived for the single-keyword parent write. Added the regression test `renames a keyword of a blur-action multi-keyword filter when the v1 irreversible projection is unchanged` (`app/api/v1/filters/[id]/route.test.ts:399`), which returns 422 on the old guard and 200 now while leaving the parent `filterAction: 'blur'` untouched. Full `[id]` suite green (19 tests).

app/api/v1/filters/[id]/route.test.ts
Comment (llun) [reply app/api/v1/filters/[id]/route.test.ts:242]: Fixed in a4b45089. Confirmed the gap: the order-insensitive `.sort()` in `contextsDiffer` was never exercised discriminatingly — every prior multi-keyword case used either a single-element context (sort is a no-op) or a length-mismatched context (short-circuits before the sort). Added the regression test `allows a whole_word toggle when the context is reordered but equal on a multi-keyword filter` (`app/api/v1/filters/[id]/route.test.ts:444`): a multi-keyword filter stored with context `['home','public']`, PUT with `['public','home']` plus a `whole_word` toggle, expects 200 and asserts `wholeWord` was updated. Verified it fails (422 vs 200) when `.sort()` is dropped in favor of an element-wise compare, then restored the sort. Full `[id]` suite green (19 tests).

PR #1202, opened 2026-07-10
minor: add user-level domain blocks (/api/v1/domain_blocks)

llun:
## Summary

Implements Mastodon's user-level domain blocks (`https://docs.joinmastodon.org/methods/domain_blocks/`). A signed-in actor can block or unblock an entire remote domain; blocked-domain posts (and boosts) disappear from their timelines, `Relationship.domain_blocking` becomes a real per-viewer lookup, and blocking severs follows in both directions (federating `Undo Follow` through the existing queue path, mirroring `lib/actions/applyBlock.ts`). Backed by a new `actor_domain_blocks` table with an `ActorDomainBlock` database module.

## Endpoints / behavior

- **`GET /api/v1/domain_blocks`** — scopes `read` / `read:blocks`. Returns an array of blocked-domain strings with `Link`-header cursor pagination (limit default **100**, max **200**; cursors are `actor_domain_blocks` row ids).
- **`POST /api/v1/domain_blocks`** — scopes `write` / `write:blocks`. Takes `domain` via JSON or form/query, normalizes it with `normalizeDomain` (rejecting wildcard forms), blocks it idempotently, and severs relationships with that domain (the caller's follows of accounts there and followers from it) via `applyDomainBlock` — a bounded, logged, batch-capped loop that federates `Undo Follow` through `SEND_UNDO_FOLLOW_JOB_NAME` for local actors. Returns `{}`; 422 on missing/invalid/wildcard domain.
- **`DELETE /api/v1/domain_blocks`** — scopes `write` / `write:blocks`. Idempotent unblock; returns `{}` (does not restore severed follows, matching Mastodon).
- **`OPTIONS /api/v1/domain_blocks`** — CORS preflight for the above.
- **`Relationship.domain_blocking`** — now computed from `isDomainBlockedByActor` instead of the hardcoded `false` (`lib/services/accounts/relationship.ts`).
- **Timeline filtering** — `getFilteredStatusPage` loads the viewer's blocked-domain set once per request and drops statuses/boosts from blocked domains as the first per-batch filter, covering home/main, direct API timelines, the home RSC page, and public/tag timelines (`lib/services/timelines/domainBlockFilter.ts` + `getFilteredTimelinePage.ts`).
- **Database module** — new `ActorDomainBlockSQLDatabaseMixin` (`createActorDomainBlock`, `deleteActorDomainBlock`, `isDomainBlockedByActor`, `getActorDomainBlocks`) and a domain-scoped follow lookup `getAcceptedOrRequestedFollowsWithDomain` on `FollowDatabase`.

## Migration & schema dumps

- Migration: `migrations/20260710000000_add_actor_domain_blocks.js` — creates `actor_domain_blocks` (`UNIQUE(actorId, domain)`, index on `(actorId, createdAt)`).
- BOTH reference schema dumps were regenerated in the same PR: `migrations/schema.sql` (PostgreSQL) and `migrations/schema.sqlite.sql` (SQLite). `SELECT count(*) FROM knex_migrations` was verified to equal the `migrations/*.js` file count (136) for each backend, and both dumps reload cleanly.

## Test plan

- `yarn run prettier --write .` — clean (no files rewritten).
- `yarn lint` — **0 errors**, 31 pre-existing warnings.
- `yarn build` — compiled successfully; `/api/v1/domain_blocks` present in the route manifest.
- `yarn test` — **634 test files passed (634); 5899 tests passed (5899).**

Targeted files exercised: `lib/database/sql/actorDomainBlock.test.ts`, `lib/database/sql/follow.test.ts`, `lib/actions/applyDomainBlock.test.ts`, `app/api/v1/domain_blocks/route.test.ts`, `lib/services/accounts/relationship.test.ts`, `lib/services/timelines/domainBlockFilter.test.ts`, `lib/services/timelines/getFilteredTimelinePage.test.ts`, and the tag/hashtag timeline route tests.

## Deviations from the plan

- **Migration count.** The plan text said 131 files → 132 applied; reality is 135 existing files, so the new migration makes 136. The regeneration asserted `knex_migrations` = 136 = `migrations/*.js` count for both SQLite and PG; no count was hardcoded.
- **PostgreSQL schema dump.** Docker is not running on this machine, so the plan's `docker run postgres:17` recipe was ignored in favor of the validated throwaway Homebrew `postgresql@17` cluster helper. Both dumps were reload-verified; `git diff migrations/` shows only rows attributable to the new migration.
- **Migration timestamp** `20260710000000` (from the plan) still sorts after the newest existing file `20260709120000`, so it was kept as-is.
- **Prettier reflows (cosmetic, behavior unchanged).** In `lib/database/sql/actorDomainBlock.ts` the race-path insert chain and the final `.map(fixActorDomainBlockDataDate)` were collapsed to single lines, and in `lib/types/database/operations.ts` the `isDomainBlockedByActor(...)` signature was collapsed to one line. In `lib/database/sql/follow.ts` the `orWhere(...)` builder body was reflowed across multiple lines. `domainBlockFilter.test.ts` had a `note(...)` argument list wrapped across lines.
- **`operations.ts` params placement.** The plan's `~line 1046` anchor for `GetFollowersParams` is stale (now ~1077); `GetAcceptedOrRequestedFollowsWithDomainParams` was inserted immediately after `GetFollowersParams`, preserving intent. Test indentation matched the file's existing 4-space base (the plan snippet showed 2-space).
- **`getRequestBody` catch typing.** The plan snippet `getRequestBody(req).catch(() => ({}))` did not compile under the repo's TypeScript strictness (the untyped `{}` widened `rawBody` so `rawBody.domain` errored during `yarn build`). Changed to `.catch((): Record<string, unknown> => ({}))` — same body-or-query `domain` merge, no behavior change.
- **`relationship.ts` `Promise.all` shape.** The plan's "before" snippet quoted an 8-entry `Promise.all` with no `incomingFollow`; current `main` already has a 9-entry array including `incomingFollow` (the `requested_by` feature). `isDomainBlocking` was added as a 10th entry and the `targetDomain ? database.isDomainBlockedByActor(...) : false` element appended after `getEndorsement`, preserving the plan's helper/comment/field-replacement intent.
- **Commit trailers / gate scope during implementation.** Task commits appended the mandated `Co-Authored-By` trailer that the plan snippets omitted; some tasks ran the gate over the touched test file after the whole-tree prettier/lint/build, while others ran the full `yarn test`.
- **Finishing pass (this agent).** Two pre-existing route test mocks (`app/api/v1/tags/[tag]/route.test.ts`, `app/api/v1/timelines/tag/[hashtag]/route.test.ts`) are consumers of `getFilteredStatusPage` and were missing the new `getActorDomainBlocks` stub, so they threw once the domain filter landed. Added `getActorDomainBlocks: vi.fn(async () => [])` to both (committed as `test:`), which turns the suite fully green.

## Out of scope

- **Notifications from blocked domains are not filtered.** This PR covers relationship + timeline read paths only. Follow-up: extend the notifications query in `lib/database/sql/notification.ts` (or its read service) with the same domain set.
- **Profile / status-detail pages don't apply user domain blocks** — only `getFilteredStatusPage` consumers do. Mastodon behaves similarly for direct profile visits; acceptable, but worth a follow-up decision.
- **`app/api/v1/accounts/[id]/block/route.ts` guards with coarse `write` only** (no granular `write:blocks`), inconsistent with the blocks list route and this new route. Trivial follow-up.
- **Unblocking a domain does not restore severed follows** — matches Mastodon (severed follows are gone for good); documented so nobody "fixes" it later.
- **Port-bearing hosts:** stored blocked domains are normalized (hostname, no port) while `follows.actorHost`/`targetActorHost` and the timeline comparison use `new URL(id).host` (includes a non-default port). No production fediverse domain carries a port, but a dev-only `host:3000` instance would not match — acceptable, noted.
- **`GetMutesParams.limit` is optional (defaults 30) while `GetBlocksParams.limit` is required** — pre-existing inconsistency; `GetActorDomainBlocksParams` deliberately uses optional-`limit` = "no cap", a third variant. A future cleanup could unify these.


app/api/v1/domain_blocks/route.ts
```diff
@@ -0,0 +1,145 @@
...
+  const rawBody = await getRequestBody(req).catch(
+    (): Record<string, unknown> => ({})
+  )
+  const url = new URL(req.url)
+  const parsed = DomainBlockBody.safeParse({
+    domain: rawBody.domain ?? url.searchParams.get('domain') ?? undefined
+  })
+  if (!parsed.success) return null
+
+  const domain = normalizeDomain(parsed.data.domain)
```
llun: **[1·confirmed]** [low/correctness] Stored block domain is port-stripped, but every comparison site keeps the port — so a block against an instance on a non-default port never applies.

`resolveDomain` stores `normalizeDomain(parsed.data.domain)`, and `normalizeDomain` (lib/services/federation/domainRules.ts:68) derives the value from `url.hostname`, which **drops the port**. So POSTing `domain=example.com:8443` stores `example.com`.

Every read path, however, compares against `new URL(actorId).host`, which **keeps the port**:
- timeline filter: `getRelevantStatusDomains` — lib/services/timelines/domainBlockFilter.ts:13 (`new URL(actorId).host` → `example.com:8443`)
- relationship: `getActorDomain` — lib/services/accounts/relationship.ts:19 (`.host`) → `domain_blocking` reports false
- severing: `getAcceptedOrRequestedFollowsWithDomain` filters `follows.actorHost`/`targetActorHost`, which are written via `new URL(actorId).host` (lib/database/sql/follow.ts:114-116), port-included — so no follows are found to sever.

**Failure scenario:** user blocks `example.com:8443`; stored as `example.com`; actor id `https://example.com:8443/users/bob` yields compare-host `example.com:8443`; `blockedDomains.has('example.com:8443')` is false → Bob's posts still show, relationship shows not-domain-blocking, and no follows are severed. For default-port (`:443`) hosts `host === hostname`, so no impact there.

**Suggested fix:** make write and read paths use the same host form. Simplest is to store a port-preserving value (e.g. derive the stored domain from `new URL('https://' + parsed.data.domain).host`) to match the codebase-wide `.host` convention — or, conversely, switch every read site (domainBlockFilter, relationship, and the severing host columns) to `.hostname`. Pick one and apply it consistently.

lib/services/timelines/domainBlockFilter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Domains a status surfaces content from: the author's domain plus, for a
+// boost, the boosted author's domain. Compared against the viewer's
+// user-level domain blocks (`actor_domain_blocks.domain`, which stores
+// normalized hostnames matching `new URL(actorId).host`).
+export const getRelevantStatusDomains = (status: Status): string[] => [
+  ...new Set(
+    getRelevantStatusActorIds(status).flatMap((actorId) => {
+      try {
+        return [new URL(actorId).host]
```
llun, line 13: **[1·confirmed]** [low/blocked-content-leak] The timeline domain filter compares against `new URL(actorId).host` (port-included) while blocks are stored via `normalizeDomain`, which uses `URL.hostname` and strips the port — the two forms diverge for any port-bearing actor id, so blocked content leaks through.

**Failure scenario:** a federation includes an instance on a non-default port; author actor id is `https://dev.example:8443/users/bob`. The blocking user POSTs `domain=dev.example:8443`; `normalizeDomain` (domainRules.ts:67-68) returns `dev.example` (port dropped), so the stored block is `dev.example`. When rendering home/public/tag timelines, `getRelevantStatusDomains` computes `new URL('https://dev.example:8443/users/bob').host` = `dev.example:8443`, and `blockedDomains.has('dev.example:8443')` is false. Bob's statuses and boosts still appear despite the user-level domain block. (Same divergence makes `Relationship.domain_blocking` report false in relationship.ts:19.)

**Suggested fix:** extract the comparison domain with `new URL(actorId).hostname` (or run the host through `normalizeDomain`) here in `getRelevantStatusDomains`, and apply the same treatment to `getActorDomain` in relationship.ts so the stored and compared forms match. (This is the read-side counterpart to the store-side issue flagged on the route.)

lib/services/accounts/relationship.ts
```diff
@@ -12,11 +12,23 @@ interface GetRelationshipParams {
   targetActorId: string
 }
 
+// Actor ids are URLs; a malformed id yields null so the relationship lookup
+// degrades to domain_blocking=false instead of throwing.
+const getActorDomain = (actorId: string): string | null => {
```
llun, line 17: **[1·not-a-defect]** [simplification] A skeptic flagged that this new local `getActorDomain` duplicates the exported `getActorDomain` in `lib/utils/activitypubActor.ts` plus a third inline copy in `domainBlockFilter.ts:13`, so actor-id→host derivation now lives in three places.

**Refuted — posting for the record.** The duplication is factually real, but there is no constructible failing input: both PR-introduced sites compute the comparison host with the identical `new URL(actorId).host` expression and agree for every input today. The finding's own failure scenario is explicitly hypothetical ("someone later edits only one site"), and that risk is already pinned by the documented contract in `lib/types/domain/actorDomainBlock.ts:6-7` that stored `domain` is compared against `new URL(actorId).host`.

More decisively, inlining `new URL(actorId).host` in a try/catch is this repo's established idiom — it appears inline at 15+ non-test sites (block.ts, mute.ts, follow.ts, endorsement.ts, accountNote.ts, ingestCollectionMemberJob.ts, …) with no shared host helper anywhere. The existing exported `getActorDomain` in activitypubActor.ts has **different** semantics (returns the raw actorId on parse failure, for `ActorProfile.domain`), so it cannot be reused as-is for a block-domain comparison. Introducing a brand-new third abstraction for two two-line call sites would be *less* consistent with the surrounding code, not more. Low-severity DRY preference against the repo's deliberate convention — no change warranted.

llun (commented):
Round 1 code review — user-level domain blocks (`/api/v1/domain_blocks`).

2 confirmed findings + 1 finding posted for the record as not-a-defect. Per AGENTS.md every finding is posted as an inline thread so the reasoning stays with the code.

- **[1·confirmed]** Port-stripped store vs port-included compare (`domain_blocks/route.ts` write path + `domainBlockFilter.ts` read path). A block against an actor served on a non-default port silently never applies.
- **[1·not-a-defect]** `getActorDomain` "duplication" in `relationship.ts` — refuted; inline `new URL(actorId).host` is the repo's established idiom and the existing exported helper has different fallback semantics.

All three anchors are inside the diff. See inline threads.

app/api/v1/domain_blocks/route.ts
llun: Fixed in 7b24e9db. Root cause confirmed: `resolveDomain` stored `normalizeDomain(...)`, which derives from `URL.hostname` and drops a non-default port, while all three read paths compare against the port-preserving `new URL(actorId).host` form (timeline filter, relationship `domain_blocking`, and the `follows.actorHost`/`targetActorHost` columns the severing query filters on).

Fix (write-side, single change for both this and the timeline-leak finding): added `normalizeActorHost` to `lib/services/federation/domainRules.ts` — same trim/lowercase/trailing-dot/255-cap validation as `normalizeDomain`, but it reconstructs the host form (hostname plus a non-default port; default `:443` dropped, matching how `.host` renders). `resolveDomain` (shared by POST and DELETE, so create/delete stay consistent) now uses it, keeping the existing wildcard rejection. `normalizeDomain` is left untouched — its ~20 admin/federation callers compare normalized-vs-normalized and must keep dropping ports.

Regression tests: `domainRules.test.ts` table-tests `normalizeActorHost` (non-default port retained, `:443` dropped, trailing-dot-with-port, lowercase, null on empty/unparseable); `route.test.ts` POSTs `dev.example:8443` and asserts `isDomainBlockedByActor({domain:"dev.example:8443"})` is true (fails pre-fix, which stored `dev.example`).

lib/services/timelines/domainBlockFilter.ts
llun: Fixed in 7b24e9db — from the write side, since this is the read-side counterpart of the same single root cause. Rather than change every read site, the store path now records the port-preserving host form via the new `normalizeActorHost` helper, so the stored `domain` matches the `new URL(actorId).host` value that `getRelevantStatusDomains` here (and `getActorDomain` in relationship.ts:19) already compute. This filter is unchanged and correct; only the write path was violating the `.host` contract its own doc comment documents.

Regression test: the route test POSTs `ported.test:8443` and then confirms the follow whose `targetActorHost` is `ported.test:8443` is severed (`severedCount:1`) — proving the stored form now matches the port-bearing read form used across all three read sites. Pre-fix, the stored `ported.test` would not match `ported.test:8443`, so it would fail.

lib/services/accounts/relationship.ts
llun: Declined — this stays a low-severity DRY preference with no constructible failing input. Both PR-introduced sites (relationship.ts:17 null-fallback, domainBlockFilter.ts:13 skip-fallback) derive the comparison host with the identical `new URL(actorId).host` expression and agree for every input today; the only failure scenario is the hypothetical "someone later edits one site," which is already pinned by the documented contract in `lib/types/domain/actorDomainBlock.ts:6-7` (stored `domain` is compared against `new URL(actorId).host` at read sites).

Inlining `new URL(actorId).host` in a try/catch is this repo established idiom — 15+ non-test sites (block.ts, mute.ts, follow.ts, endorsement.ts, accountNote.ts, ingestCollectionMemberJob.ts, …) with no shared host helper anywhere. The exported `getActorDomain` in `lib/utils/activitypubActor.ts` has different semantics (returns the raw actorId on parse failure, for `ActorProfile.domain`), so it cannot be reused for a block-domain comparison — extracting a brand-new third abstraction for two two-line call sites would be less consistent with the surrounding code, not more. No change warranted.

lib/actions/applyDomainBlock.ts
```diff
@@ -0,0 +1,119 @@
...
+    const followsToFederate = (
+      await Promise.all(
+        follows.map(async (follow) => {
+          const followActor = await database.getActorFromId({
+            id: follow.actorId
+          })
+          return followActor?.privateKey ? follow : null
+        })
+      )
+    ).filter((follow): follow is Follow => Boolean(follow))
```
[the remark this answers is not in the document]
llun: Real optimization — adopted in 29776b6c.

`getAcceptedOrRequestedFollowsWithDomain` returns exactly two disjoint kinds of row: the caller's own outbound follows (`follow.actorId === actorId` and `targetActorHost === domain`) and inbound followers on the blocked domain (`targetActorId === actorId` and `follow.actorId` = the remote/other actor). Federation is meant only for the caller's own outbound follows, so `follows.filter((follow) => follow.actorId === actorId)` is the exact set, and it drops the per-follow `getActorFromId` lookup — an N+1 across up to `SEVER_BATCH_SIZE` (100) rows every batch, up to `MAX_SEVER_BATCHES` (10) times.

It's also *more* faithful to the documented intent than the private-key proxy. On a multi-domain server a local follower on a blocked local domain has a signing key, so the old `followActor?.privateKey` check would wrongly federate an Undo Follow "as" that follower; the id comparison correctly excludes it while still federating every outbound follow.

Regression test (`lib/actions/applyDomainBlock.test.ts`) spies on `getActorFromId` and asserts it is not consulted during severing — it fails on the pre-fix code (2 calls) and passes now.

Note the `getActorFromId(...).privateKey` pattern stays as-is in `lib/actions/applyBlock.ts`: that path only ever has ≤2 follows, so there is no N+1 to eliminate there.

app/api/v1/domain_blocks/route.ts
```diff
@@ -0,0 +1,145 @@
...
+// maps null to 422. User-level blocks target one exact domain, so the
+// wildcard forms the admin-level normalizeDomain permits are rejected here.
+const resolveDomain = async (req: NextRequest): Promise<string | null> => {
+  const rawBody = await getRequestBody(req).catch(
+    (): Record<string, unknown> => ({})
+  )
+  const url = new URL(req.url)
+  const parsed = DomainBlockBody.safeParse({
+    domain: rawBody.domain ?? url.searchParams.get('domain') ?? undefined
+  })
```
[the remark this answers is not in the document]
llun: Real defect — confirmed and fixed in 29776b6c.

Verified empirically: a JSON body of literal `null` parses to `null` (`JSON.parse('null')` does not throw, so the route's `.catch` never fires), and `rawBody.domain` then throws `TypeError: Cannot read properties of null (reading 'domain')`, which surfaces as a 500. The route test reproduces exactly this against the pre-fix code.

Rather than guard in this one route, I fixed the root cause in `getRequestBody` (`lib/utils/getRequestBody.ts`). Its return type is `Record<string, unknown>`, but for `application/json` it returned the raw `JSON.parse` result, so it lied for any non-object body. It now normalizes a well-formed non-object JSON value (`null`, numbers, strings, booleans) to `{}`; arrays (also objects) pass through unchanged. That closes the same latent 500 for all 16 `getRequestBody` callers, not just this route — every caller reads `body.<field>` and expects a Record.

Regression tests: `getRequestBody.test.ts` table-tests non-object bodies normalizing to `{}`, and `route.test.ts` POSTs a `null` JSON body and asserts 422 (not 500). Both fail on the pre-fix code with the exact TypeError above and pass now.

lib/services/accounts/relationship.ts
llun, line 17: **[2·not-a-defect]** Candidate: the module-private `getActorDomain` here reuses the same name and signature as the exported `getActorDomain` in `lib/utils/activitypubActor.ts`, with a divergent fallback (this one returns `null`, the util returns the raw `actorId`).

Refuted by a majority of skeptics — not a defect:
- This `getActorDomain` is a module-private `const` (not exported), and `relationship.ts` does not import the util's `getActorDomain`. So there is no shadowing, no name-resolution collision, and no wrong-import at runtime. A shared name across two unrelated files is normal in TypeScript.
- The divergent `null` fallback is intentional and documented by the comment on lines 15-16: a malformed id yields `null` so the `domain_blocking` lookup degrades to `false` instead of throwing. The util's raw-`actorId` fallback serves a different purpose (building profile domains).
- Every current call site uses its intended version; there is no constructible input that produces wrong output or an exception with the code as-is. The alleged harm is entirely hypothetical ("a future maintainer could import/consolidate the wrong one").

A small private helper with deliberately different failure semantics is acceptable here; the suggested rename/extract is optional polish, not a fix for any real bug or invariant violation. No change required.

llun (commented):
Round 2 code-review results for PR #1202 (user-level domain blocks).

After fanning out sub-agent reviewers over the full diff, round 2 surfaced **0 confirmed defects**. The one candidate that reached the reconcile stage was refuted by a majority of independent skeptics and is posted inline below purely to keep the reasoning on the record — it is **not** an actionable finding.

The round-1 write/read normalization fix (`normalizeActorHost`) was re-checked and remains consistent across the write path and every read comparison.

lib/services/accounts/relationship.ts
llun: Declined — refuted, no change. Verified against the code: `relationship.ts` does not import `getActorDomain` from `lib/utils/activitypubActor.ts` (its imports are Database, Mastodon, Actor, FollowStatus, urlToId, getMuteExpiresAt only). The `getActorDomain` at relationship.ts:17 is a module-private `const` returning `string | null`, and the exported one at activitypubActor.ts:7 returns the raw `actorId` and is used only within its own file (line 46). So there is no shadowing, no name-resolution collision, and no wrong-import at runtime — a shared name across two unrelated modules is normal in TypeScript.

The divergent `null` fallback here is intentional and documented by the comment on lines 15-16: a malformed id yields `null` so the `domain_blocking` lookup degrades to `false` instead of throwing (the util needs the raw-`actorId` fallback for building profile domains, a different purpose). Every current call site uses its intended version and there is no constructible input that produces wrong output; the only alleged harm ("a future maintainer could import/consolidate the wrong one") is hypothetical. Low-value maintainability nit, not a present defect or invariant violation.

PR #1203, opened 2026-07-10
minor: add instance legal pages, health endpoint, oEmbed, and v1 notifications policy alias

llun:
## Summary

Adds four small, independent Mastodon-compat surfaces in one PR: the public instance legal-document endpoints (privacy policy + terms of service), an unauthenticated `/health` liveness probe, a public `/api/oembed` provider with status-page discovery, and the deprecated `GET|PUT|PATCH /api/v1/notifications/policy` alias over the shared v2 policy logic. No migration, no new dependencies, and nothing here is called from React components (no `lib/client.ts` change). Two new optional config keys back the legal pages; the v2 notification-policy route is refactored (behavior-neutral) to share a new service module.

## Endpoints / behavior

- `GET /api/v1/instance/privacy_policy` (+ `OPTIONS`) — 200 `{ updated_at: '1970-01-01T00:00:00.000Z', content: '<p>…</p>' }` built from `ACTIVITIES_PRIVACY_POLICY` (HTML-escaped, paragraph-wrapped); **404 when unset/empty** (documented deviation — Mastodon serves a bundled default, this server has none).
- `GET /api/v1/instance/terms_of_service` (+ `OPTIONS`) — 200 Mastodon 4.4 entity `{ effective_date: '1970-01-01', effective: true, content: '<p>…</p>', succeeded_by: null }` from `ACTIVITIES_TERMS_OF_SERVICE`; 404 when unset/empty (matches Mastodon).
- `GET /api/v1/instance/terms_of_service/[date]` (+ `OPTIONS`) — 200 the same entity when the configured document exists and `date === '1970-01-01'`; 404 otherwise.
- `GET /health` — unauthenticated 200 `{"status":"UP"}`, no CORS/`OPTIONS`, no DB/auth (documented deviation — Mastodon replies `text/plain` `OK`; probes should assert on the 200 status, not the body).
- `GET /api/oembed` (+ `OPTIONS`) — public oEmbed provider for this instance's own **public or unlisted** status URLs (canonical `/@user@domain/<id>`, short `/@user/<id>`, percent-encoded, and trusted-host forms), resolved through the existing `resolveStatusFromPath` + `isStatusPubliclyReadable` helpers (local lookup only). Returns a `type: 'rich'` doc whose `html` is a static `<blockquote>` linking to the status page (no per-status embed widget exists). 400 on missing `url`; 404 for foreign hosts, non-status paths, unknown statuses, or non-public statuses.
- Status-page `generateMetadata` now emits the oEmbed discovery `<link rel="alternate" type="application/json+oembed" href="…/api/oembed?url=…">` for every status page.
- `GET|PUT|PATCH /api/v1/notifications/policy` (+ `OPTIONS`) — deprecated alias. Serializes the legacy v1 entity: `filter_*` **booleans** (`true` when the v2 `for_*` value is anything other than `accept`) plus the `summary` object, omitting `for_limited_accounts`. Updates accept `filter_*` booleans (Mastodon maps `update` to both PUT and PATCH) with Mastodon `ActiveModel::Type::Boolean` false-token coercion; 422 on non-scalar values.

Serializer / shared-code changes:

- New `lib/services/notifications/notificationPolicy.ts` — `getNotificationPolicyResponse`, `toV1NotificationPolicy`, and `readNotificationPolicyBody` (JSON/urlencoded/multipart body reader). `app/api/v2/notifications/policy/route.ts` is refactored to consume `getNotificationPolicyResponse` + `readNotificationPolicyBody`; its guards/scopes/`PUT`=`PATCH` and response shape are unchanged.
- New shared `lib/utils/text/escapeHtml.ts`, now used by the legal routes and oEmbed; `app/api/v1/instance/extended_description/route.ts` drops its local copy and imports the shared util (behavior unchanged).
- New optional config keys `privacyPolicy` / `termsOfService` (`z.string().nullish()`) in `lib/config/index.ts`, read from `ACTIVITIES_PRIVACY_POLICY` / `ACTIVITIES_TERMS_OF_SERVICE`.

## Test plan

All commands run from the worktree root against local SQLite (Node 24):

- `yarn run prettier --write .` — clean, no files reformatted.
- `yarn build` — compiled successfully; `/health` and `/api/oembed` register as dynamic routes.
- `yarn test` — **Test Files 637 passed (637), Tests 5893 passed (5893)**.

New/updated co-located tests: `lib/utils/text/escapeHtml.test.ts`, `lib/config/index.test.ts`, `app/api/v1/instance/privacy_policy/route.test.ts`, `app/api/v1/instance/terms_of_service/route.test.ts`, `app/health/route.test.ts`, `app/api/oembed/route.test.ts`, `app/(timeline)/[actor]/[status]/page.test.ts`, `lib/services/notifications/notificationPolicy.test.ts`, `app/api/v1/notifications/policy/route.test.ts`, and the existing `app/api/v2/notifications/policy/route.test.ts` (still green after the refactor).

## Deviations from the plan

- **`terms_of_service/[date]/route.ts` import path.** The plan imported the shared entity builder via a relative parent path (`import { getTermsOfServiceEntity } from '../termsOfService'`), which trips the project's no-`../`-imports lint rule. Changed to the absolute alias `@/app/api/v1/instance/terms_of_service/termsOfService` (the form the lint error prescribes); intent unchanged, lint back to 0 errors.
- **`readNotificationPolicyBody` form parsing.** The plan's snippet read both `application/x-www-form-urlencoded` and `multipart/form-data` via `req.formData()`, which throws (`req.formData is not a function`) on the synthetic string bodies used in the node test env and failed the plan's own "reads a form-encoded body" test. Urlencoded bodies are now parsed via `URLSearchParams` (matching the repo-standard `lib/utils/getRequestBody.ts`); `req.formData()` is retained only for multipart. Behavior is identical in production.
- **v2 policy route refactor was surgical, not a wholesale replace.** The plan's step-5 snippet had drifted from the real `app/api/v2/notifications/policy/route.ts` (which uses `OAuthGuardAnyScope` with two-scope arrays, a `PUT`/`PATCH` binding, and CORS `PUT`). To keep the v2 route behavior-neutral, only the inline `buildPolicyResponse` → `getNotificationPolicyResponse` and the inline body reader → `readNotificationPolicyBody` were swapped, plus the now-unused `import { Database }` removed. All existing v2 route tests still pass unchanged.
- **Docs (this finishing task).** Added feature entries to `docs/features.md`, intentional-divergence entries to `docs/mastodon-api-compatibility.md`, the two legal-page keys to `.env.example`, and the new `/api/oembed` and `/health` route trees to `docs/architecture.md` (the `ACTIVITIES_PRIVACY_POLICY` / `ACTIVITIES_TERMS_OF_SERVICE` rows in `docs/environment-variables.md` landed with the config-keys commit).
- Otherwise none substantive: per-commit `Co-Authored-By` trailers and worktree path translation are the only other differences from the plan snippets, which otherwise matched the real source.

## Out of scope

Found while reading; intentionally left for follow-ups:

- `GET /api/v1/instance/extended_description` returns a hardcoded default description when `serviceDescription` is unset instead of 404/empty — fine per Mastodon (extended_description always 200s), left untouched.
- Neither `app/api/v1/instance/route.ts` nor the v2 instance route advertises the new legal documents (Mastodon 4.4's v2 instance entity exposes `terms_of_service` presence via `api_versions`/links). Wiring the v2 instance payload to reflect configured legal docs would be a separate compat PR.
- The status page `generateMetadata` emits only a bare title — no OpenGraph/Twitter card metadata for status unfurls. Adding `openGraph` fields (author, excerpt, media) would materially improve link previews; separate PR.
- The auth-vs-CORS inconsistency in `app/api/v2/notifications/policy/route.ts` error paths (uses `apiResponse` with `ERROR_500` instead of the newer `apiCorsError` shorthand) was preserved verbatim in the refactor to keep the diff behavior-neutral; a repo-wide sweep to `apiCorsError` is a cleanup candidate.
- `readNotificationPolicyBody`'s JSON-or-form normalization pattern is duplicated ad hoc in other write routes (e.g. several `/api/v1` POST handlers roll their own); a shared `lib/utils/readRequestBody.ts` sweep would deduplicate, but touching those routes here would bloat the PR.
- `getServedDomains` drops wildcard `ACTIVITIES_TRUSTED_HOSTS` entries (e.g. `*.example.com`), so oEmbed requests for status URLs on a wildcard-matched concrete subdomain 404 even though the proxy serves them. Same limitation already applies to the passkey domain chooser; acceptable and documented here, fixable later by matching wildcards in the oEmbed host check.


app/api/v1/notifications/policy/route.ts
```diff
@@ -0,0 +1,134 @@
...
+  ['filter_private_mentions', 'for_private_mentions']
+] as const
+
+// https://docs.joinmastodon.org/methods/notifications/ (deprecated section)
+// The pre-4.3 beta filtered-notifications API. It reads and writes the same
+// stored policy as /api/v2/notifications/policy but serializes filter_*
+// booleans instead of the v2 for_* accept/filter/drop strings.
+export const GET = traceApiRoute(
+  'getNotificationPolicyV1',
+  OAuthGuard([Scope.enum.read], async (req, { currentActor, database }) => {
```
llun: **[1·confirmed]** [medium · correctness] v1 notifications-policy guard rejects granular scopes that the v2 endpoint it aliases accepts

**Defect:** The GET handler uses `OAuthGuard([Scope.enum.read])` and the shared update handler (line 83) uses `OAuthGuard([Scope.enum.write])`. `OAuthGuard` runs matchMode `all`, and `hasGrantedScope` in `lib/services/guards/scopeHierarchy.ts` deliberately does NOT let a granular scope satisfy a coarse requirement (the doc comment there points to `OAuthGuardAnyScope` as the remedy). The v2 route this aliases uses `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']])` for GET and `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']])` for writes, so it accepts both the coarse and the granular scope.

**Failure scenario:** A client authorized with only `read:notifications` (Mastodon's recommended minimal scope for this endpoint, and enough for `GET /api/v2/notifications/policy`) calls `GET /api/v1/notifications/policy`. `hasGrantedScope(['read:notifications'], 'read')` returns `false`, so the request is rejected 401 even though the same token returns 200 on v2. The identical problem hits a `write:notifications`-only token on PUT/PATCH at line 83. The plan's Task 9 intent that the v1 guard scopes "match v2" is therefore not met.

**Suggested fix:** Use `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']])` for the GET handler and `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']])` for `updatePolicyV1`, exactly as `app/api/v2/notifications/policy/route.ts` does.

llun: **[1·confirmed]** [medium · mastodon-conformance] GET guard omits `read:notifications`, rejecting granular-scope tokens Mastodon accepts

**Defect:** `OAuthGuard([Scope.enum.read])` requires the coarse `read` scope (matchMode `all`). Real Mastodon's v1 controller authorizes this endpoint with `:read, :'read:notifications'`, and this repo's own v2 route uses `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']])` — both return 200 for a `read:notifications`-only token.

**Failure scenario:** A client authorized with only the granular `read:notifications` scope (no coarse `read`) calls `GET /api/v1/notifications/policy`. `hasGrantedScope` refuses granular→coarse, so the guard returns 401, diverging from both Mastodon and this repo's v2 route.

**Suggested fix:** Use `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']], ...)` for the GET handler so a granular `read:notifications` token is accepted, matching v2.

app/api/v1/notifications/policy/route.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const data = toV1NotificationPolicy(
+      await getNotificationPolicyResponse(database, currentActor.id)
+    )
+    return apiResponse({ req, allowedMethods: CORS_HEADERS, data })
+  })
+)
+
+const updatePolicyV1 = OAuthGuard(
+  [Scope.enum.write],
```
llun: **[1·confirmed]** [medium · mastodon-conformance] PUT/PATCH guard omits `write:notifications`, rejecting granular-scope tokens Mastodon accepts

**Defect:** `updatePolicyV1 = OAuthGuard([Scope.enum.write])` requires the coarse `write` scope, which a granular `write:notifications` token does not satisfy (`scopeHierarchy` refuses granular→coarse). Mastodon's v1 update authorizes with `:write, :'write:notifications'`, and the v2 route here uses `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']])` — both return 200.

**Failure scenario:** A client authorized with only `write:notifications` calls `PUT` or `PATCH /api/v1/notifications/policy`; the guard returns 401 even though the same token succeeds on `PATCH /api/v2/notifications/policy`.

**Suggested fix:** Use `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']], ...)` for the shared update handler, matching v2.

app/api/v1/notifications/policy/route.ts
```diff
@@ -0,0 +1,134 @@
...
+// Mastodon casts the legacy booleans with ActiveModel::Type::Boolean: a fixed
+// set of false tokens, everything else true.
+const FALSE_TOKENS = new Set(['false', 'f', '0', 'off', ''])
+
+const LegacyBoolean = z
+  .union([z.boolean(), z.number(), z.string()])
+  .transform((value) => {
+    if (typeof value === 'boolean') return value
+    if (typeof value === 'number') return value !== 0
+    return !FALSE_TOKENS.has(value.toLowerCase())
```
llun, line 42: **[1·not-a-defect]** [test-coverage] Claimed: the `LegacyBoolean` transform's case-folding (`toLowerCase`) and the non-`'false'` FALSE_TOKENS (`'f'`,`'0'`,`'off'`,`''`) and numeric branch are never exercised by tests.

**Rationale (refuted):** The coverage observation is factually accurate but describes no current defect. Running the transform over `'False','OFF','f','0','',0,1,'true','yes'` yields the Mastodon-correct value in every case, and all tests pass. The finding's own failure scenario is explicitly conditional on a *hypothetical future edit* ("if a future edit drops `.toLowerCase()`"), i.e. a speculative regression, not a bug in this PR. The delivered test set also matches the approved plan's test contract verbatim, and the string path *is* partly exercised by the form-encoded test (`filter_private_mentions=true&filter_not_followers=false`). A one-line `it.each` hardening test would be cheap but low-value; there is nothing wrong to fix now. No reproducible current defect — not upheld.

lib/services/notifications/notificationPolicy.ts
```diff
@@ -0,0 +1,80 @@
...
+  filter_new_accounts: response.for_new_accounts !== 'accept',
+  filter_private_mentions: response.for_private_mentions !== 'accept',
+  summary: response.summary
+})
+
+// Mastodon accepts the policy update as JSON or form-encoded; normalize both
+// to a plain object for Zod. A request with no (or an unrecognized)
+// content-type is treated as JSON so a bodyless/typeless PATCH still resolves
+// to {} — this mirrors the original inline v2 reader that clients rely on.
+export const readNotificationPolicyBody = async (
```
llun, line 56: **[1·not-a-defect]** [duplication] Claimed: `readNotificationPolicyBody` re-implements the shared `lib/utils/getRequestBody.ts` body parser.

**Rationale (refuted):** Only the single urlencoded branch (line 65) is truly identical. The other two branches diverge deliberately: `getRequestBody` returns `{}` for a missing/unrecognized content-type and only parses JSON when the header is `application/json`, whereas `readNotificationPolicyBody` treats any missing/unrecognized content-type as JSON via `req.json().catch(() => ({}))`, and its multipart branch adds a `.catch(() => null)` + `String()` coercion that `getRequestBody` lacks. The comment on lines 52-55 documents this JSON-default as intentional — it "mirrors the original inline v2 reader that clients rely on." Task 8 is an explicit behavior-preserving refactor, so delegating to `getRequestBody` would change the fallback content-type semantics and regress the exact behavior the refactor must preserve. The author already saw `getRequestBody` (it's referenced in a comment) and deliberately did not delegate. Low-value style nit resting on an intentional, documented choice — not upheld.

llun (commented):
Round 1 sub-agent review. 3 confirmed findings + 2 findings adjudicated as not-a-defect (posted with rationale so the reasoning is on record, per AGENTS.md). All findings anchor inside the diff. The confirmed findings are one root cause: the v1 notifications-policy alias uses coarse-only `OAuthGuard` scopes while the v2 endpoint it aliases uses `OAuthGuardAnyScope` with the granular `read:notifications`/`write:notifications` scopes, so granular-scoped tokens are rejected on v1 but accepted on v2.

app/api/v1/notifications/policy/route.ts
llun: Fixed in 4610f7c4.

`GET` now uses `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']])` and the shared `updatePolicyV1` (PUT/PATCH) uses `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']])`, exactly mirroring `app/api/v2/notifications/policy/route.ts`. Because `OAuthGuardAnyScope` matches ANY listed scope instead of requiring the coarse scope, a token holding only `read:notifications` (GET) or `write:notifications` (PUT/PATCH) is now accepted with 200 rather than rejected 401 — restoring true v1/v2 parity, exactly the suggested fix.

Regression coverage added in `app/api/v1/routeScopeWiring.test.ts`: the v1 policy route is now in both `EXPECTED` (union: read, read:notifications, write, write:notifications) and `MULTI_METHOD` (per-method — GET=read+read:notifications, PUT/PATCH=write+write:notifications). Verified the test FAILS on the pre-fix coarse-only wiring and PASSES on the fix.

llun: Fixed in 4610f7c4.

The `GET` handler now uses `OAuthGuardAnyScope([Scope.enum.read, Scope.enum['read:notifications']])`, so a granular `read:notifications`-only token returns 200 — matching both `GET /api/v2/notifications/policy` and Mastodon's real v1 controller (`:read, :'read:notifications'`). Regression covered by the GET per-method entry added to `app/api/v1/routeScopeWiring.test.ts`.

app/api/v1/notifications/policy/route.ts
llun: Fixed in 4610f7c4.

`updatePolicyV1` (shared by PUT and PATCH) now uses `OAuthGuardAnyScope([Scope.enum.write, Scope.enum['write:notifications']])`, so a granular `write:notifications`-only token is accepted — matching `PATCH /api/v2/notifications/policy` and Mastodon's v1 update (`:write, :'write:notifications'`). Regression covered by the PUT/PATCH per-method entries added to `app/api/v1/routeScopeWiring.test.ts`.

app/api/v1/notifications/policy/route.ts
llun: Declined — not a current defect.

The coverage observation is factually accurate but describes no reproducible bug. Running the `LegacyBoolean` transform over `'False','OFF','f','0','',' ',0,1,'true','yes'` yields the Mastodon-correct result in every case (all false-tokens → accept, everything else → filter), mirroring `ActiveModel::Type::Boolean`, and all tests pass. The finding's own failure scenario is explicitly conditional on a hypothetical future edit ("if a future edit drops `.toLowerCase()`"), i.e. a speculative regression, not a bug in this PR.

The string path is in fact partly exercised by the form-encoded test (`filter_private_mentions=true&filter_not_followers=false`, which invokes `.toLowerCase()` and matches the `'false'` token); only uppercase case-folding, the numeric branch, and the other tokens (`'f'`,`'0'`,`'off'`) are uncovered. The delivered test set matches the approved plan's test contract verbatim. Nothing to fix now.

lib/services/notifications/notificationPolicy.ts
llun: Declined — deliberate, documented divergence, not a defect.

Only the single urlencoded branch is identical to `lib/utils/getRequestBody.ts`. The other branches differ on purpose: `getRequestBody` returns `{}` for a missing/unrecognized content-type and only parses JSON when the header is `application/json`, whereas `readNotificationPolicyBody` treats any missing/unrecognized content-type as JSON via `req.json().catch(() => ({}))`, and its multipart branch adds a `.catch(() => null)` + `String()` coercion that `getRequestBody` lacks. The comment on lines 52-55 documents this JSON-default as intentional — it "mirrors the original inline v2 reader that clients rely on."

Task 8 is an explicit behavior-preserving refactor, so delegating to `getRequestBody` would change the fallback content-type semantics and regress exactly the behavior the refactor must preserve. Low-value style nit resting on an intentional, documented choice — not upheld.

app/api/oembed/route.ts
```diff
@@ -0,0 +1,141 @@
...
+  const displayStatus = getOriginalStatus(status)
+  const author = displayStatus.actor
+  const authorHandle = author
+    ? getMention(author, true)
+    : getMentionFromActorID(displayStatus.actorId, true)
+  const authorName = author?.name || authorHandle
+  const authorUrl = author
+    ? `https://${author.domain}/${getMention(author)}`
+    : displayStatus.actorId
+  const excerpt = htmlToPlainText(displayStatus.text)
```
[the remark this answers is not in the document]
llun: This is a false positive — no runtime crash is possible, so no fallback is needed.

Two independent reasons:

1. `displayStatus` here is `getOriginalStatus(status)`, typed `StatusNote | StatusPoll`. Both schemas define `text: z.string()` (required, non-nullable), so `displayStatus.text` is always a string — a media-only post has `''`, never `null`/`undefined`.
2. Even if it could be nullish, `htmlToPlainText` is declared `(html: string | null | undefined)` and null-guards internally (`sanitizeHtml(html ?? '', …)`), so a `null`/`undefined` argument returns `''` rather than crashing.

Leaving line 115 as-is.

app/api/oembed/route.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const actorSegment = decodePathParam(segments[0])
+  if (!actorSegment.startsWith('@')) return null
+
+  // Single-@ form (/@user/<status>): the proxy serves it on the request host,
+  // so qualify the username with the page URL's host.
+  const actorParam =
+    actorSegment.split('@').length === 3
+      ? actorSegment
+      : `${actorSegment}@${pageUrl.host}`
```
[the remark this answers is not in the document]
llun: False positive — `pageUrl.host` (port included) is the correct qualifier here; switching to `.hostname` would **break** the single-`@` oEmbed form in the standard local-dev configuration. Leaving line 70 as-is.

**Why the port must be preserved.** A local actor's `domain` is stored verbatim as `config.host` (the raw `ACTIVITIES_HOST`), port included:

- `lib/services/accounts/registerAccount.ts` → `const { host: domain } = config` → `createAccount({ domain, … })`; `scripts/mock/createMockUser.ts` does the same.
- `lib/config/host.ts` reads `host: process.env.ACTIVITIES_HOST || ''` with no port stripping, and `buildBaseURL` keeps the host verbatim.

So with the standard local-dev `ACTIVITIES_HOST=localhost:3000` (per CLAUDE.md, `ACTIVITIES_HOST` matches the `yarn dev --port`), the actor row's `domain` is `localhost:3000`, and its id is `http://localhost:3000/users/test1`.

The single-`@` qualifier builds `@user@<domain>`, which `resolveStatusFromPath` splits straight back into `[username, domain]` and passes to `database.getActorFromUsername({ username, domain })` — an **exact** match on the stored `domain` string (`lib/database/sql/actor.ts` `.where('domain', domain)`). So:

- `.host` → `@test1@localhost:3000` → `getActorFromUsername({ domain: 'localhost:3000' })` → **matches** ✅
- `.hostname` → `@test1@localhost` → `getActorFromUsername({ domain: 'localhost' })` → **no such actor → 404** ❌

In production (`ACTIVITIES_HOST=llun.dev`, standard 443) `.host` and `.hostname` are identical, so `.host` is strictly the safer choice.

**On the `servedDomains` gate** (line 57 uses `pageUrl.hostname`): that is deliberate and not a contradiction. `getServedDomains` returns bare hostnames (via `toHostname` → `url.hostname`), so the gate is a coarse "is this a host we serve" allowlist that is intentionally port-insensitive. The actor qualification, by contrast, is an exact DB key and must preserve the port.

**Verified empirically** with a throwaway vitest exercising the real `GET /api/oembed` route with `getConfig().host = 'localhost:3000'` and a mock `getActorFromUsername` that matches only `domain === 'localhost:3000'` (mirroring the real store): the current `.host` code returns **200** and calls `getActorFromUsername({ username: 'test1', domain: 'localhost:3000' })`; the proposed `.hostname` would pass `domain: 'localhost'` → null → 404. Test deleted after confirming.

This is the same conclusion PR #1202's review reached: actor-host comparisons in this multi-domain repo must preserve a non-default port because actor ids/domains are stored and compared with the port. Not changing line 70.

app/api/v1/notifications/policy/route.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      })
+      expect(data.filter_not_following).toBe(true)
+      expect(data.summary).toEqual({
+        pending_requests_count: 1,
+        pending_notifications_count: 3
+      })
+    }
+  )
+
+  it('accepts form-encoded legacy booleans', async () => {
```
llun, line 115: **[2·confirmed]** low · test-coverage

The new `LegacyBoolean` transform's `FALSE_TOKENS = new Set(['false','f','0','off',''])` (route.ts:35) maps anything **not** in that set to `true` (route.ts:42, `return !FALSE_TOKENS.has(value.toLowerCase())`). But the only inputs this suite exercises are boolean `true`/`false` (the PUT/PATCH `it.each`) and the string tokens `'true'`/`'false'` here on line 121. The remaining four false tokens (`'f'`, `'0'`, `'off'`, `''`) and the numeric / non-`'true'` truthy-string branches are never parsed.

Because of that, a regression that 'simplified' `FALSE_TOKENS` down to just `['false']` would keep **every** existing test in this file green, yet a real form request like `filter_not_following=0` (or `off`, or an empty `filter_not_following=` from an unchecked checkbox) would coerce to `true` and be stored as `for_not_following: 'filter'` instead of `'accept'` — silently flipping the user's notification policy.

Suggested fix: add a table-driven `it.each` over the `LegacyBoolean` input/expected variants — `'0'`,`'off'`,`'f'`,`''` (and numeric `0`) → `false`/`'accept'`; `'1'`,`'yes'`, an arbitrary string (and numeric `1`) → `true`/`'filter'` — asserting the resulting `for_*` value passed to `updateNotificationPolicy`, so any change to the token set is caught.

lib/services/notifications/notificationPolicy.ts
llun, line 56: **[2·not-a-defect]** simplification (refuted — posted for the record)

A reviewer flagged that `readNotificationPolicyBody` reimplements `getRequestBody`'s form-body parsing instead of delegating to it. On adjudication this is **not** a defect:

- There is no concrete failing input. The finding's own 'failure scenario' is hypothetical future drift ('if `getRequestBody`'s parsing is later changed…'), not a current wrong output or exception.
- The two parsers are **not** equivalent, by design. This reader deliberately falls through to `req.json().catch(() => ({}))` for a bodyless/typeless or unrecognized content-type (line 79), and the comment on lines 52–55 documents that this mirrors the original inline v2 reader clients rely on. `getRequestBody` instead returns `{}` for any unrecognized/typeless content-type. Its multipart branch also has no `.catch`, whereas this one wraps `req.formData().catch(() => null)` and resolves to `{}` on malformed input.
- The suggested refactor (route non-JSON content-types through `getRequestBody`) would regress both behaviors: a present-but-unrecognized content-type would return `{}` instead of the documented JSON parse, and malformed multipart would throw instead of resolving to `{}`.

The function correctly diverges on purpose; the observation is style/maintainability and the proposed change is behavior-changing. No action needed.

llun (commented):
Round 2 code-review of PR #1203 (head 4610f7c4). One confirmed finding and one refuted finding (posted for the record). Both anchor inside newly-added files.

- 1 confirmed: `[2·confirmed]` low/test-coverage on the LegacyBoolean FALSE_TOKENS set.
- 1 refuted: `[2·not-a-defect]` simplification claim on `readNotificationPolicyBody` — the divergence from `getRequestBody` is intentional and documented; posted with rationale.

No new correctness or invariant violations. Not approving.

lib/services/notifications/notificationPolicy.ts
llun: Refuted — no code change. `readNotificationPolicyBody` is a deliberate, documented divergence from `getRequestBody`, not a naive copy.

- No concrete failing input: the finding's own failure scenario is hypothetical future drift in `getRequestBody`, not a current wrong output or exception.
- The parsers are not equivalent by design. `readNotificationPolicyBody` falls through to `req.json().catch(() => ({}))` for a bodyless/typeless or unrecognized content-type (documented in the comment at lines 52-55 as mirroring the original inline v2 reader clients rely on), whereas `getRequestBody` returns `{}` for any unrecognized/typeless type. Its multipart branch also wraps `req.formData().catch(() => null)` and resolves to `{}` on malformed input, while `getRequestBody`'s multipart branch has no catch.
- The suggested refactor (route non-JSON content-types through `getRequestBody`) would regress both behaviors: a present-but-unrecognized content-type would return `{}` instead of the documented JSON parse, and malformed multipart would throw instead of resolving to `{}`.

The function correctly diverges on purpose; the observation is style/maintainability and the proposed change is behavior-changing. Resolving.

app/api/v1/notifications/policy/route.test.ts
llun: Fixed in e2546184.

Added a table-driven `it.each` to `app/api/v1/notifications/policy/route.test.ts` that drives the real `PUT` handler through `LegacyBoolean` for every branch and asserts the `for_not_following` value passed to `updateNotificationPolicy`:

- Every FALSE_TOKEN — form `'false'`, `'f'`, `'0'`, `'off'`, empty string, plus uppercase `'FALSE'` (exercising `.toLowerCase()`) — maps to `for_not_following: 'accept'`.
- Truthy string tokens `'true'`, `'1'`, `'on'`, `'yes'` map to `'filter'`.
- JSON numeric coercion `0` maps to `'accept'`; `1`/`2` map to `'filter'`.

Confirmed the guard bites: temporarily shrinking `FALSE_TOKENS` to `['false']` fails exactly the four previously-uncovered tokens (`'f'`, `'0'`, `'off'`, and the empty string), then reverted. Resolving.

app/api/oembed/route.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  // For a boost page, embed the boosted status — matching what the page shows.
+  const displayStatus = getOriginalStatus(status)
+  const author = displayStatus.actor
+  const authorHandle = author
+    ? getMention(author, true)
+    : getMentionFromActorID(displayStatus.actorId, true)
+  const authorName = author?.name || authorHandle
+  const authorUrl = author
```
llun, line 112: **[3·not-a-defect]** [reuse] Claim: `authorUrl` reinlines the body of the existing `getActorURL` helper instead of calling `getActorURL(author)`.

**Refuted (majority of skeptics).** This is a DRY nit, not a defect, and following the repo's prevailing convention.

- **No behavioral divergence.** At runtime `https://${author.domain}/${getMention(author)}` (default `withDomain=false`) is byte-for-byte identical to `getActorURL(author)`, which is itself `https://${actor.domain}/${getMention(actor, false)}`. There is no actor input for which the two produce different `author_url`. The finding's own failure scenario is conditional on a hypothetical future edit to `getActorURL` that has not happened.
- **Inlining is the de-facto convention here.** The exact `https://${actor.domain}/${getMention(actor)}` construction appears inline in `createPoll.ts:113`, `createNote.ts:421`, `importFitnessFilesJob.ts:217`, and `importStravaActivityJob.ts:417`, plus a dozen `https://${domain}/@${username}` variants across `lib/` and `app/`. `getActorURL` has exactly one non-test caller in the whole repo (`mentionNotification.ts:147`), so it is the outlier, not the standard.
- **The suggested fix doesn't typecheck and is invasive.** `author` here is `ActorProfile` while `getActorURL` takes `Actor`. `Actor extends ActorProfile`, so `ActorProfile` is the wider base and is not assignable to `Actor`; adopting the helper would force widening a shared helper's signature and touching its lone caller purely to dedup a one-line expression.

No bug, contradicts the repo's prevailing style, and requests a shared-signature refactor for cosmetics — not worth changing.

llun (commented):
Round 3 sub-agent code review of PR #1203.

**Summary:** 0 confirmed findings this round. 1 candidate was raised and refuted by a majority of independent reviewers; it is posted inline below with its rationale so the reasoning is on the record. No approval/merge action is implied by this review.

All candidate anchors fall inside the diff; there are no findings outside the diff.

app/api/oembed/route.ts
llun: Refuted — no code change. `authorUrl` inlining `https://${author.domain}/${getMention(author)}` is not a defect.

- **No behavioral divergence.** With default `withDomain=false`, `getMention(author)` → `@${author.username}`, so the inline is `https://${author.domain}/@${author.username}`. `getActorURL(author)` (actor.ts:68) is `https://${actor.domain}/${getMention(actor, false)}` — the identical string. Verified against the source: there is no actor input for which the two diverge. The finding's failure scenario is conditional on a hypothetical future edit to `getActorURL` that has not happened.
- **Inlining is the de-facto convention.** The same `https://${actor.domain}/${getMention(actor)}` construction appears inline in createPoll.ts:113, createNote.ts:421, importFitnessFilesJob.ts:217, importStravaActivityJob.ts:417. `getActorURL` has exactly one non-test caller in the whole repo (mentionNotification.ts:147), so it is the outlier, not the standard.
- **The suggested fix does not typecheck.** `author` is an `ActorProfile`; `getActorURL` takes `Actor`, and `Actor extends ActorProfile` (actor.ts:43), so the wider `ActorProfile` base is not assignable to `Actor`. Adopting the helper would force widening a shared helper's signature and touching its lone caller purely to dedup a one-line expression.

No bug, follows the repo's prevailing style, and the proposed change is an invasive shared-signature refactor for cosmetics. Resolving.

PR #1204, opened 2026-07-10
minor: add tags feature/unfeature endpoints and the Mastodon Profile entity

llun:
## Summary

Brings two Mastodon 4.4/4.6 client-API surfaces to activities.next. **Part A** adds `POST /api/v1/tags/:tag/feature` and `POST /api/v1/tags/:tag/unfeature` (scope `write:accounts`, both returning the `Tag` entity), reusing the existing featured-tags create-or-return-existing / idempotent-delete semantics through a new shared service, and adds the optional `featuring` boolean to the `Tag` entity for these routes plus `GET /api/v1/tags/:tag`. **Part B** reshapes `GET`/`PATCH /api/v1/profile` — which previously returned `CredentialAccount` — to return the Mastodon 4.6 **Profile** entity, and teaches the PATCH to accept and persist the new appearance params. No migration: all profile preferences live in the existing `actors.settings` JSON column.

## Endpoints / behavior

- **`POST /api/v1/tags/:tag/feature`** (new) — scope `write` / `write:accounts`; idempotent create-or-return-existing via the new `featureTag` service; returns the `Tag` entity with `featuring: true`; returns `422` (`You can only feature up to N hashtags`) when the per-account cap is reached.
- **`POST /api/v1/tags/:tag/unfeature`** (new) — scope `write` / `write:accounts`; idempotent delete-by-name (no-op when not featured); returns the `Tag` entity with `featuring: false`.
- **`GET /api/v1/tags/:tag`** (changed) — now surfaces `featuring` alongside the existing `following`/`history`, derived from `getFeaturedTagByName` (both default to `false` for anonymous requests).
- **`GET /api/v1/profile`** (changed) — now returns the Mastodon 4.6 `Profile` entity via the new `buildProfile` service (`getProfileHandler`); additionally accepts the documented `profile` scope alongside `read`/`read:accounts`.
- **`PATCH /api/v1/profile`** (changed) — now returns the `Profile` entity and accepts/persists `avatar_description`, `header_description`, `show_media`, `show_media_replies`, `show_featured` (plus the pre-existing `indexable`/`hide_collections`/`attribution_domains`) into `actors.settings`.
- **`PATCH /api/v1/accounts/update_credentials`** (unchanged behavior) — still returns `CredentialAccount`; `updateCredentialsHandler` gained an opt-in `responseEntity: 'profile'` flag that only the profile route passes.

Entity / serializer changes:
- `Tag` (`lib/types/mastodon/tag.ts`) gains an optional `featuring` boolean; `getMastodonTag(name, following, history = [], featuring?)` emits it only when explicitly provided.
- New `Profile` Zod entity (`lib/types/mastodon/account/profile.ts`, re-exported as `Mastodon.Profile`) and `buildProfile` service (`lib/services/accounts/profile.ts`), the single construction point for the Profile entity (nullable `avatar`/`header`, non-null empty-string `avatar_description`/`header_description`, `hide_collections ?? null`, `indexable ?? false`, `show_media`/`show_media_replies`/`show_featured ?? true`, `attribution_domains ?? []`).
- New shared `featureTag`/`unfeatureTag` service (`lib/services/mastodon/featureTag.ts`); `POST /api/v1/featured_tags` now delegates to it.
- `ActorSettings` / `UpdateActorParams` gain `avatarDescription`/`headerDescription`/`showMedia`/`showMediaReplies`/`showFeatured`; `updateActor` merges them into the settings JSON with `undefined`-guarded spreads so explicit `false`/`''` values persist.

## Test plan

All commands run from the worktree root on Node 24 with yarn:

- `yarn run prettier --write .` — clean (no files rewritten).
- `yarn build` — succeeded (this repo's TypeScript gate).
- `yarn test` — **635 test files passed, 5878 tests passed.**

New/updated tests: `getMastodonTag`, `featureTag` service (real in-memory SQLite), tags `feature`/`unfeature` route tests, `GET /api/v1/tags/:tag` featuring assertion, `buildProfile`, `GET`/`PATCH /api/v1/profile` Profile-shape + appearance-param round-trips, and an `updateActor` → `getActorSettings` settings round-trip (including explicit `false` flags).

## Deviations from the plan

- **`getMastodonTag` signature moved on.** The plan assumed a 2-arg `(name, following)` serializer and prescribed adding `featuring?` as the 3rd param. The live serializer is `(name, following, history: TagHistory[] = [])` with call sites already passing `history`, so `featuring?` was added as the **4th** param `(name, following, history = [], featuring?)`; all feature/unfeature/GET call sites pass `getMastodonTag(tag, following, [], true/false)` and `getMastodonTag(tag, following, history, featuring)`. Every existing call site compiles unchanged.
- **Hashtag path-param validation uses the shared Unicode validator.** Instead of copying the plan's ASCII-only regex, the feature/unfeature routes import `normalizeHashtagParam` + `MAX_ENCODED_HASHTAG_PARAM_LENGTH` from `@/lib/utils/text/mastodonHashtag` (the Phase 2 shared validator that sibling follow/unfollow/GET tag routes already use), so they accept valid Unicode tags and reject inputs like `nope nope` with `400`.
- **`DELETE /api/v1/featured_tags/:id` was intentionally not routed through the new service** — it keeps its own by-id owner-scoped delete because its 404-on-not-owned semantics differ from `unfeatureTag`'s name-based no-op. The route-level HTTP-contract comment above `POST /api/v1/featured_tags` was kept rather than trimmed, as it documents still-relevant HTTP-status behavior.
- **`ActorSettings`/`UpdateActorParams`/`updateActor` `attribution_domains` was already present** (landed with Phase 2 PR 2.5), so only the five genuinely-new appearance fields were added; `attribution_domains` was not re-added (kept its existing `.max(100)`), and the shared `UpdateCredentialsRequest` already carried `indexable`/`hide_collections`/`attribution_domains`.
- **`buildProfile` no longer hardcodes `hide_collections: null` / `indexable: false`.** Because `update_credentials` (via 2.5) persists those flags, `buildProfile` now reads `settings?.hideCollections ?? null` / `settings?.indexable ?? false` so the GET/PATCH Profile echo is accurate; its unit test asserts the pass-through. As a consequence, `PATCH /api/v1/profile` also accepts `hide_collections`/`indexable` (the overlap the plan flagged is resolved now that 2.5 landed).
- **`updateCredentialsHandler` gained an opt-in `responseEntity` option** (`'credential_account'` default | `'profile'`) so `PATCH /api/v1/accounts/update_credentials` stays byte-identical (still `CredentialAccount`, its 24-test suite unchanged) while `PATCH /api/v1/profile` opts into the Profile entity.
- **The `Profile` entity test literal keeps the plan's verbatim `Account` object** (omitting the now-required `uri`/`roles`); Vitest transpiles tests without type-checking and `buildProfile` never reads those fields, so it runs and passes. Not reused in type-checked code.
- **Docs:** `docs/features.md` gained a **Profile entity** entry and expanded the **Featured tags API** entry to cover the feature/unfeature aliases; `docs/mastodon-api-compatibility.md` gained one divergence note (activities.next always emits `following`/`featuring` on `GET /api/v1/tags/:name`, defaulting to `false` for anonymous tokens, and `featuring` appears on the new feature/unfeature responses). `docs/architecture.md` enumerates DB tables, not individual routes, so it needed no change.

## Out of scope

Copied from the plan's "Out of scope, found while reading":

- `updateCredentialsHandler` does not run `localizeAccount` before `buildCredentialAccount`, while `getCredentialAccountHandler` does — on a multi-domain deployment a PATCH `update_credentials` response can render `acct` inconsistently with the GET. (Profile has no `acct`, so this PR sidesteps it.)
- `POST /api/v1/tags/:tag/follow` and `/unfollow` responses omit `featuring` (Mastodon 4.4 includes it for authorized tokens); `app/api/v1/featured_tags/suggestions/route.ts` hardcodes `following: false` without consulting `isFollowingTag`. Both are one-line follow-ups after the serializer change.
- The old tags path-param regex `/^[a-zA-Z0-9_]*[a-zA-Z_][a-zA-Z0-9_]*$/` (route + follow + unfollow) rejects unicode hashtags that `featured_tags`' `/^[\p{L}\p{N}_]+$/u` accepts — the new feature/unfeature routes already use the shared Unicode validator; migrating the remaining siblings is Phase 2 PR 2.7/2.8's shared-validator work.
- `GET /api/v1/tags/:tag` returns `following: false` (and now `featuring: false`) for anonymous requests instead of omitting the optional fields as Mastodon does for unauthorized tokens — pre-existing behavior, kept for consistency (and now documented in the compatibility reference).
- `app/api/v1/trends/tags/route.ts` issues one `isFollowingTag` query per trending tag inside `Promise.all` — a batched lookup would be nicer, unrelated to this PR.


lib/services/accounts/profile.ts
```diff
@@ -0,0 +1,41 @@
...
+  Mastodon.Profile.parse({
+    id: account.id,
+    display_name: account.display_name,
+    note: account.source.note,
+    fields: account.source.fields,
+    avatar: settings?.iconUrl ?? null,
+    avatar_static: settings?.iconUrl ?? null,
+    avatar_description: settings?.avatarDescription ?? '',
+    header: settings?.headerImageUrl ?? null,
+    header_static: settings?.headerImageUrl ?? null,
```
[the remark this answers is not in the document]
llun: Thanks — but this is a no-op here rather than a real fix, so I am leaving the `??`.

`ActorSettings.iconUrl` / `headerImageUrl` are typed `string | undefined`, and an empty string is never a reachable stored value:

- Clearing an image deletes the key rather than writing `""` (`lib/database/sql/actor.ts:889-892`, `applyExplicitClears` → `delete target.iconUrl`), so it becomes `undefined`.
- The media-upload path only ever assigns a real URL and is guarded with `iconUrl !== undefined` before persisting (`lib/services/accounts/updateCredentialsHandler.ts:236,304`).
- Every settings merge guards on truthiness: `...(settings.iconUrl ? { iconUrl } : null)` (`lib/database/sql/account.ts:477`, `lib/database/sql/actor.ts:666,845`), so a falsy `""` is dropped, not stored.

Since the only "unset" representation is `undefined`, `settings?.iconUrl ?? null` already returns `null` for every unset case; `|| null` would be behaviorally identical. The `??` is also the intentional choice documented in the file header (avatar/header are `null` when unset, mirroring the Account entity’s `?? ""`).

app/api/v1/tags/[tag]/feature/route.ts
```diff
@@ -0,0 +1,97 @@
...
+          responseStatusCode: 400
+        })
+      }
+
+      const result = await featureTag({
+        database,
+        actorId: currentActor.id,
+        name: tag
+      })
+      if (result.status === 'limit_reached') {
```
Comment (llun) [inline app/api/v1/tags/[tag]/feature/route.ts:71]: **[2·confirmed]** [low/test-coverage] The `limit_reached` → 422 branch (lines 71–80) is not exercised by any route-level test. `app/api/v1/tags/[tag]/feature/route.test.ts` covers only 200 (feature), 200 (idempotent), 400 (invalid name) and 401 (unauth); none of them first features `FEATURED_TAGS_LIMIT` (10) tags for the actor before POSTing an 11th, so if this HTTP mapping regressed — branch deleted → 200 with a Tag entity, or wrong status/message — the suite would stay green. The shared service test `lib/services/mastodon/featureTag.test.ts` asserts the service returns `{ status: 'limit_reached' }` but never verifies that this route maps it to a 422 with the cap-message body.

Suggested fix: add a case to `app/api/v1/tags/[tag]/feature/route.test.ts` that seeds `FEATURED_TAGS_LIMIT` distinct featured tags for the seeded actor (e.g. via `database.createFeaturedTag`), then POSTs one more through the route and asserts `response.status === 422` and the returned error message. `app/api/v1/featured_tags/route.test.ts` already does exactly this for the sibling endpoint, so it is the established repo convention.

llun (commented):
Round 2 sub-agent code review — 2 confirmed findings, 0 refuted. Round 1 was a clean round (8 reviewers, zero findings); these two surfaced on the re-sweep.

One finding is anchored inline below. The other cannot be anchored because its line is outside this PR's diff — the file itself was not modified; the PR made the comment stale indirectly — so it is recorded here:

### Findings outside the diff

**[2·confirmed]** [low/stale-comment] `lib/services/accounts/credentialsHandler.ts:13` — `getCredentialAccountHandler`'s doc comment still names `GET /api/v1/profile` as a consumer ("used by both GET /api/v1/accounts/verify_credentials and GET /api/v1/profile") and justifies the `corsHeaders` parameter with "The CORS allow-list differs per route (verify_credentials = OPTIONS,GET; profile = OPTIONS,GET,PATCH), so it is passed in." This PR rerouted profile to `getProfileHandler` — `app/api/v1/profile/route.ts` now imports `getProfileHandler` and uses it for `GET` (line 19), and the `getCredentialAccountHandler` import was removed. The only remaining caller is `app/api/v1/accounts/verify_credentials/route.ts`, whose `CORS_HEADERS = [OPTIONS, GET]`. So both comment claims are now false: profile no longer shares this handler, and `corsHeaders` is only ever `[OPTIONS, GET]`. Suggested fix: update the comment to name the single remaining consumer (e.g. "Read handler for the current actor's CredentialAccount, used by GET /api/v1/accounts/verify_credentials.") and drop/rephrase the per-route CORS justification.

This is a COMMENT review — not approving or requesting changes.

app/api/v1/tags/[tag]/feature/route.ts
Comment (llun) [reply app/api/v1/tags/[tag]/feature/route.ts:71]: Fixed in 3da6a656. Added a route-level test to `app/api/v1/tags/[tag]/feature/route.test.ts` — `returns 422 once the per-account featured-tags limit is reached` — that seeds a fresh actor (Actor2) with all `FEATURED_TAGS_LIMIT` (10) tags through the route itself, then POSTs an 11th and asserts `response.status === 422` and the exact body `{ error: 'You can only feature up to 10 hashtags' }`. This exercises the previously-untested `limit_reached` → 422 branch at route.ts:71-80: deleting that branch would fall through to `isFollowingTag`/`getMastodonTag` and return 200 with a Tag entity, so the new assertion is a real guard. File went from 4 to 5 tests, all passing.

lib/services/accounts/profile.ts
```diff
@@ -0,0 +1,41 @@
...
+    header_description: settings?.headerDescription ?? '',
+    locked: account.locked,
+    bot: account.bot,
+    hide_collections: settings?.hideCollections ?? null,
```
llun, line 34: **[3·not-a-defect]** _(refuted — posted for the record, no change requested)_

**Claim:** `buildProfile` re-derives `hide_collections`, `indexable`, and `attribution_domains` from `settings` even though the `account` argument already carries those exact values (computed with identical default expressions in `getMastodonActor`), and it "inconsistently" reuses `account.discoverable` for the one field it doesn't re-derive.

**Why this is not a defect:**

1. **No runtime bug — the two paths are provably identical.** `buildProfile` (lines 34/36/40) and `getMastodonActor` (`actor.ts` ~296/297/310) apply the *same* default expressions (`settings.hideCollections ?? null`, `settings.indexable ?? false`, `settings.attributionDomains ?? []`) to the *same* actor settings row. Given identical input and identical fallbacks, the output is byte-identical today. The finding's own failure scenario is explicitly conditional on a hypothetical *future* change ("if a later change adjusts any of those defaults") — speculative divergence, not a present defect.

2. **The `discoverable` "inconsistency" is necessity, not oversight.** `getMastodonActor` computes `discoverable` as `settings.discoverable ?? !isLocalHeadlessSigner`, and `buildProfile` has no access to `isLocalHeadlessSigner`, so it *must* read `account.discoverable`. The other three fields have trivial, reproducible `?? false/null/[]` defaults and are computed inline — a principled distinction, not an accident.

3. **Not actually a simplification.** `buildProfile` is inherently a settings-based builder: it already needs `settings` for ~10 fields with no `account` equivalent (avatar/header descriptions, `show_media`, `show_media_replies`, `show_featured`). Switching three fields to `account.*` removes no dependency and no line — it's a lateral move that keeps a *less* coherent mental model.

Majority of skeptics: not a defect. Low-value stylistic nit; left as-is.

llun (commented):
## Round 3 code review — activities.next PR #1204

Round 3 fanned out independent skeptics over the full diff. **No confirmed defects** — the PR is clean on this pass (tags feature/unfeature endpoints, `featuring` on the Tag entity, and the Mastodon 4.6 Profile entity all check out against the plan, and the `hide_collections`/`indexable`/`attribution_domains` acceptance in `UpdateCredentialsRequest` is the intended, scope-guarded Phase 2 PR 2.5 behavior).

One finding was raised and **refuted** by a majority of independent skeptics. It is posted inline below for the record with the reasoning, so the decision is auditable — it is **not** an action item.

lib/services/accounts/profile.ts
llun: Refuted, no code change. Adjudicated on the code at `lib/services/accounts/profile.ts:34-40`: the finding is accurate that `hide_collections`/`indexable`/`attribution_domains` are re-derived from `settings` while `discoverable` (line 35) reuses `account`, but this is not a defect. The two paths are provably identical today — `buildProfile` and `getMastodonActor` apply the same default expressions (`hideCollections ?? null`, `indexable ?? false`, `attributionDomains ?? []`) to the same actor settings row, so output is byte-identical. The `discoverable` "inconsistency" is necessity, not oversight: `getMastodonActor` computes `discoverable` as `settings.discoverable ?? !isLocalHeadlessSigner`, and `buildProfile` has no access to `isLocalHeadlessSigner`, so it must read `account.discoverable`; the other three have trivial reproducible defaults. `buildProfile` is inherently a settings-based builder (it already needs `settings` for ~10 fields with no account equivalent), so switching these three to `account.*` removes no dependency — a lateral move, not a simplification. The failure scenario is purely hypothetical future divergence with no runtime impact. Low-value stylistic nit; left as-is.

PR #1205, opened 2026-07-10
minor: align collections API with the final Mastodon 4.6 spec

llun:
## Summary

The collections stack shipped against a pre-final draft of the Mastodon 4.6 Collections API. This PR aligns the routes with the finalized spec: the dual request vocabulary (`name`/`tag_name`/`discoverable`/`sensitive`/`account_ids` — spec params win — alongside the pre-final `title`/`topic`/`visibility` extensions), the spec response entities (`WrappedCollection`, `CollectionWithAccounts`, `WrappedCollectionItem` with stable item ids), anonymous reads of discoverable/link-shareable collections, the new `GET /api/v1/accounts/:id/collections` listing, cross-owner `in_collections`, and item-id-addressed remove/revoke. Every existing activities.next behavior (consent gate, public feed, owner list, bulk `account_ids` mutations, account-id addressing) is preserved as a documented extension.

## Endpoints / behavior

- `POST /api/v1/collections` — now accepts the dual vocabulary (`name`/`tag_name`/`discoverable`/`sensitive`/`account_ids[]` plus legacy `title`/`topic`/`visibility`/`feed_enabled`); `sensitive`/`discoverable` are strict `z.boolean()`, `account_ids[]` capped at 100; returns `WrappedCollection` (`{ collection }`).
- `GET /api/v1/collections/:id` — now uses an optional OAuth guard so anonymous callers can read `public`/`unlisted` collections; 404s only for missing or non-owner `private`; returns `CollectionWithAccounts` (`{ accounts, collection }`).
- `PATCH /api/v1/collections/:id` — dual vocabulary; passes `sensitive` through; returns `WrappedCollection`.
- `DELETE /api/v1/collections/:id` — unchanged (`{}`).
- `GET /api/v1/accounts/:id/collections` — **new** route (optional auth): owner sees all their collections, strangers see `visibility='public'` only; `{ collections: [...] }` with `limit`/`offset` (default 40, max 80) + offset Link headers.
- `GET /api/v1/accounts/:id/in_collections` — rewritten to return `{ collections: [...] }` (was a bare array) with offset Link headers; now cross-owner via `getCollectionsFeaturingAccount` (viewer's own collections in any state UNION other owners' public collections with an approved membership).
- `POST /api/v1/collections/:id/items` — spec `{ account_id }` returns `WrappedCollectionItem` (`{ collection_item }`); bulk `{ account_ids[] }` extension returns `{}`; both share `addMembersToCollection`.
- `DELETE /api/v1/collections/:id/items/:item_id` — **new** item-id-addressed removal (owner-scoped, 404 for unknown item or non-owner). The dynamic segment was renamed `[account_id]` → `[item_id]`; the bulk `account_ids[]` body DELETE on the parent route is kept.
- `POST /api/v1/collections/:id/items/:item_id/approve` and `.../revoke` — resolve the segment item-id-first with an account-id fallback (the extension form the first-party UI uses); 404 for a nonexistent membership, 403 for a non-self membership.

Entity / serializer changes:

- New `MastodonCollectionEntity` (merged wire entity: spec fields `account_id`/`uri`/`url`/`name`/`description`/`local`/`sensitive`/`discoverable`/`tag`/`item_count`/`items`/`created_at`/`updated_at` PLUS kept legacy `title`/`topic`/`language`/`visibility`/`feed_enabled`/`size`), `CollectionItemEntity`, `CollectionTag`, `WrappedCollection`, `WrappedCollectionItem`, `CollectionWithAccounts`.
- New serializer service `lib/services/collections/serializers.ts` centralizes the consent-state mapping (`pending→pending`, `approved→accepted`, `revoked→revoked`; `rejected` is never emitted) and the `discoverable = visibility==='public'` mapping, and batches item embedding/counting via `getCollectionItems`/`countCollectionItems` (public projection embeds/counts approved-only items so pending/revoked never leak).
- New DB methods on `CollectionDatabase`: `getCollectionItems`, `getCollectionItem`, `getCollectionItemByAccount`, `countCollectionItems`, `getAccountCollections`, `getCollectionsFeaturingAccount`, `removeCollectionItemById`; removed the now-unused `getCollectionsWithAccount`. Domain `Collection` gained a required `sensitive: boolean`.
- `lib/client.ts` `createCollection`/`updateCollection` now unwrap `.collection` from the `WrappedCollection` responses (component callers unchanged — the merged entity is a superset of the legacy `CollectionEntity`).

## Migration & schema dumps

- Migration `migrations/20260710000000_add_collections_sensitive.js` adds a `sensitive` boolean column (NOT NULL, DEFAULT false, idempotent hasTable/hasColumn guards) to `collections`. It is migration 136 and sorts last.
- BOTH reference schema dumps were regenerated: `migrations/schema.sql` (PostgreSQL) now has `sensitive boolean DEFAULT false NOT NULL` and `migrations/schema.sqlite.sql` (SQLite) now ends the `collections` CREATE with `` `sensitive` boolean not null default '0'``. Each diff contains only that one column.

## Test plan

Full pre-commit gate run from the worktree root, in order:

- `yarn run prettier --write .` — clean, no files rewritten.
- `yarn build` — compiled successfully in ~14s; all routes registered (including `/api/v1/accounts/[id]/collections` and the renamed `.../items/[item_id]` + `/approve` + `/revoke`).
- `yarn test` — 638 test files passed, 5899 tests passed (0 failures).

New coverage includes dual-vocabulary create, anonymous discoverable read + private-read 404, WrappedCollectionItem→DELETE round-trip, revoke-by-item-id 403 for non-self, `accounts/:id/collections` visibility split, and cross-owner `in_collections`.

## Deviations from the plan

- Task 1: schema-only migration with no test to write first, so the plan's TDD red-step did not apply (`redStepObserved` false); correctness was verified by asserting the regenerated schema diffs contain only the new column and running the existing 24-test collection suite against the regenerated SQLite dump. Used the validated Homebrew-postgresql@17 regen script instead of the plan's `docker run postgres:17` recipe (Docker is not running on this machine); it reproduces both committed dumps byte-for-byte at `main`. Migration counts differ from the stale plan text: the repo has 135 existing migrations, so this is migration 136 (not 132); confirmed `knex_migrations` == `ls migrations/*.js | wc -l` == 136 on both backends.
- Task 2: prettier reflowed the `it.each([...])` table literal in `collection.test.ts` and a single-line `for` loop in `getCollectionItems` into multiple lines — pure auto-format, logic matches the plan verbatim. No other deviations.
- Task 3: prettier reflowed a couple of single-line annotations and a long `it.each` callback onto multiple lines; content identical, formatting only.
- Task 4: import ordering in the rewritten `app/api/v1/collections/route.ts` differs from the plan snippet because prettier's import-sort plugin mandates alphabetical `@/lib/...` ordering (same imports, functionally identical). Added the standard `Co-Authored-By: Claude Opus 4.8` commit trailer per the environment's git-commit guidance.
- Task 5: the task's git-add list named only `route.ts`/`route.test.ts`/`lib/client.ts`, but `lib/client.test.ts` also had to be updated to keep `yarn test` green — Task 4's `createCollection` `.collection` unwrap had left the createCollection mock as a bare entity (pre-existing red on the branch); wrapped both collection-helper mocks in `{ collection: ... }` and added a return-value assertion. Prettier reformatted two spots relative to the plan snippets; content unchanged.
- Task 6: the plan's DELETE test snippet used `import { POST } from '../route'`, which trips the project's no-`../`-imports lint rule; changed it to the absolute `@/app/...` path, preserving intent (the only lint error, resolved). Copied the sibling collection tests' shared 5-mock preamble verbatim (the plan called it "the six vi.mock blocks"). Husky pre-commit hook is non-executable in this environment, so it was skipped; the prettier/lint/build/test gate was run manually and is green.
- Task 7: the plan referenced a "six-mock preamble" but every sibling collection/account route test of this shape uses a 5-mock preamble, so the new tests match the real siblings (5 mocks). In the `in_collections` red step only the two data-shape cases failed pre-implementation; the anonymous-rejection case already passed because the pre-existing route also used `OAuthGuardAnyScope`. Prettier reflowed one ternary. The task's file list contained no docs, so `docs/mastodon-api-compatibility.md`/`docs/features.md` were left for this PR-level ship step (done here).
- Task 8 (ship): docs were updated here (`docs/features.md` collections bullet + `docs/mastodon-api-compatibility.md` collections extension entry, adding the new `/api/v1/accounts/:id/collections` route and the final-4.6-alignment note).

## Out of scope

- `z.coerce.boolean()` on `feed_enabled` (kept as-is for compatibility) coerces the string `'false'` to `true`; the new `sensitive`/`discoverable` params deliberately use strict `z.boolean()`. Migrating `feed_enabled`/`visibility` off coercion is a separate cleanup.
- The owner items listing (`GET /api/v1/collections/:id/items`) still returns bare `Mastodon.Account[]` without item ids; extending it to return `CollectionItem`-shaped rows would let `lib/client.ts` drop the account-id extension eventually.
- `app/(timeline)/lists/page.tsx`, `collections/[id]/page.tsx`, and `collections/[id]/edit/page.tsx` build the legacy entity server-side via `getMastodonCollection`; migrating those pages (and `CollectionEditor`/`CollectionDetail`) to the merged serializer would remove the last legacy-entity producer.
- The public collection page (`/collections/[id]`) gates on `visibility !== 'private'` while the feed route additionally requires `publicFeed`; the entity's `url` field now always points at that page even when the feed is disabled — consider surfacing `feed_enabled` there.
- `getPublicCollectionTimeline` treats `unlisted` as feed-readable (matches this PR's GET :id gate) but the `discoverable` flag is not consulted anywhere in the feed path — fine today, worth a comment if feeds ever get a discoverability surface.
- Inbound collection federation (consuming remote `FeaturedCollection`) remains planned; `local` is therefore hard-coded `true` in the serializer — revisit when remote collections are stored.


app/api/v1/collections/[id]/items/[item_id]/approve/route.ts
```diff
@@ -8,29 +9,40 @@ import {
...
+      const item = await resolveCollectionItem(database, id, item_id)
+      if (!item) {
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_404,
+          responseStatusCode: 404
+        })
+      }
+      if (item.targetActorId !== currentActor.id) {
```
Comment (llun) [inline app/api/v1/collections/[id]/items/[item_id]/approve/route.ts]: **[1·confirmed]** medium · authorization

`resolveCollectionItem` (line 36) is scoped only by `collectionId` — it calls `getCollectionItem`/`getCollectionItemByAccount`, neither of which filters by owner or visibility — and this route is reachable by ANY token holding `write:collections`. When the resolved item belongs to someone else, line 45 returns a **403**, but an unknown item returns **404** (line 42): a distinguishable response that leaks collection membership and consent state to any authenticated non-owner.

**Failure scenario:** Attacker A (not the owner, not a member) knows a public/unlisted collection X — its id is exposed in its public URL `https://host/collections/{id}` — and a candidate account V. A sends `POST /api/v1/collections/X/items/<V-account-id>/approve`. If V is a member of X in ANY state (including `pending`/`revoked`, which the public projection deliberately hides), the resolved `item.targetActorId` (=V) ≠ A yields **403**; if V is not a member, it yields **404**. A thus learns V's un-consented (pending) or revoked membership, defeating the consent-hiding design. For a private collection whose UUID A knows, this also defeats the existence-hiding that `GET /collections/:id` enforces (404 for private to non-owners).

**Suggested fix:** When `item.targetActorId !== currentActor.id`, return `ERROR_404` (not `ERROR_403`) so "not your item" is indistinguishable from "no such item", closing the membership/consent-state oracle and matching the existence-hiding used by the GET detail route. Apply the same change in the revoke route.

app/api/v1/collections/[id]/items/[item_id]/revoke/route.ts
```diff
@@ -8,29 +9,39 @@ import {
...
+      const item = await resolveCollectionItem(database, id, item_id)
+      if (!item) {
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_404,
+          responseStatusCode: 404
+        })
+      }
+      if (item.targetActorId !== currentActor.id) {
```
Comment (llun) [inline app/api/v1/collections/[id]/items/[item_id]/revoke/route.ts]: **[1·confirmed]** medium · authorization

Same root cause as the sibling approve route: `resolveCollectionItem` (line 35) is not owner- or visibility-scoped, and the endpoint is reachable by any `write:collections` token. Line 44 returns **403** when the target item belongs to someone else, but an unknown item returns **404** (line 40).

**Failure scenario:** `POST /api/v1/collections/X/items/<victim-account-id>/revoke` returns **403** when the victim is a member of X (any state) and **404** otherwise, so a stranger who knows the collection id can enumerate whether arbitrary accounts are `pending`/`revoked`/`approved` members — information the approved-only public projection and the private-collection 404 are designed to hide.

**Suggested fix:** Return `ERROR_404` instead of `ERROR_403` when `item.targetActorId !== currentActor.id`, so the response does not distinguish "someone else's item" from "unknown item".

app/api/v1/collections/[id]/route.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      method: 'PATCH',
+      headers: {
+        'content-type': 'application/json',
+        origin: 'https://llun.test'
+      },
+      body: JSON.stringify(body)
+    })
+  const context = (id: string) => ({ params: Promise.resolve({ id }) })
+
+  it('serves a discoverable collection anonymously as CollectionWithAccounts', async () => {
```
Comment (llun) [inline app/api/v1/collections/[id]/route.test.ts:141]: **[1·confirmed]** low · test-coverage

No test covers the headline "public reads are broader than discoverable" behavior: a GET of an **unlisted** collection served to an anonymous/stranger viewer (200, public projection). The `beforeAll` setup only creates `public` and `private` collections, so the load-bearing unlisted-vs-private distinction is untested.

**Failure scenario:** The GET handler (`route.ts:53`) 404s a non-owner only when `collection.visibility === 'private'`; unlisted is intentionally served. If a refactor narrowed this to a discoverable-only guard (e.g. `!isOwner && collection.visibility !== 'public'`), unlisted (link-shareable) collections would incorrectly 404 for anonymous/stranger viewers, breaking link sharing — and the entire suite would still pass, because no test reads an unlisted collection as a non-owner. (The listing endpoint's unlisted-hidden case IS covered in `accounts/[id]/collections/route.test.ts`, but the single-collection unlisted-visible case is not.)

**Suggested fix:** Add an unlisted collection to the `beforeAll` setup and a GET test asserting an anonymous/stranger viewer receives 200 with the public (approved-only) projection for it, alongside the existing private→404 tests.

app/api/v1/accounts/[id]/in_collections/route.ts
```diff
@@ -1,42 +1,94 @@
...
   'getAccountInCollections',
   OAuthGuardAnyScope<Params>(
     [Scope.enum.read, Scope.enum['read:collections']],
     async (req, { database, currentActor, params }) => {
       const { id } = await params
-      const collections = await database.getCollectionsWithAccount({
-        actorId: currentActor.id,
-        targetActorId: idToUrl(id)
+      const url = new URL(req.url)
+      const parsedLimit = parseInt(
```
Comment (llun) [inline app/api/v1/accounts/[id]/in_collections/route.ts]: **[1·confirmed]** low · duplication

The two new account-collections listing routes copy-paste identical limit/offset parsing and offset Link-header construction instead of sharing a helper, and both reinvent the existing `clampedLimit`/`clampedOffset` utilities.

**Failure scenario:** Lines 34-46 here re-inline verbatim the exact limit/offset parse block that the sibling `accounts/[id]/collections/route.ts` already extracted into a local `parsePaging` helper; the `DEFAULT_LIMIT=40` / `MAX_LIMIT=80` constants and the `buildLink`/`links` Link-header assembly are also duplicated byte-for-byte except the URL path. A future change to paging behavior — bumping `MAX_LIMIT`, fixing the `parseInt`/`Number.isSafeInteger` clamp semantics, or altering the Link-header format — must be applied in both files; touching only one silently makes `/accounts/:id/collections` and `/in_collections` paginate differently. The repo already ships `lib/utils/clampedLimit.ts` (`clampedLimit`/`clampedOffset`), used by ~15 other routes for exactly this query-param clamping, so the hand-rolled parsers are also a reuse gap.

**Suggested fix:** Parse limit/offset with the existing `clampedLimit(80, 40)` / `clampedOffset()` Zod helpers (matching `directory/route.ts` and the other listing routes), and hoist the shared offset Link-header builder into a small helper that both account-collections routes call, so the paging contract lives in one place.

llun (commented):
## Round 1 code review — collections Mastodon 4.6 alignment

Automated sub-agent review of the full diff. 4 findings, all **confirmed** (each upheld by a majority of 3 independent skeptics). Posted as inline comments below.

- 2x **medium/authorization** — approve & revoke item routes leak collection membership / consent state via a distinguishable 403-vs-404 response to any authenticated non-owner (IDOR/oracle).
- 1x **low/test-coverage** — no test reads an *unlisted* collection as an anonymous/stranger viewer, leaving the load-bearing unlisted-vs-private GET distinction unguarded.
- 1x **low/duplication** — the two account-collections listing routes duplicate limit/offset parsing + Link-header building instead of reusing `clampedLimit`/`clampedOffset`.

No findings were refuted this round. All 4 anchor to lines inside the diff.

lib/services/collections/serializers.ts
```diff
@@ -0,0 +1,151 @@
...
+  projection: 'owner' | 'public'
+): Promise<MastodonCollectionEntity[]> => {
+  if (collections.length === 0) return []
+  const collectionIds = collections.map((collection) => collection.id)
+  const approvedOnly = projection === 'public'
+  const [itemsMap, totalCounts, approvedCounts] = await Promise.all([
+    database.getCollectionItems({ collectionIds, approvedOnly }),
+    database.countCollectionItems({ collectionIds }),
+    database.countCollectionItems({ collectionIds, approvedOnly: true })
+  ])
```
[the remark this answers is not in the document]
llun: Confirmed and fixed in 04062369. In the public projection `getCollectionEntities` reports `approvedCounts` for both `item_count` and `size`, so the all-states `totalCounts` map was indeed unused there. It now resolves `{}` instead of issuing `countCollectionItems({ collectionIds })` when `approvedOnly` is true, saving one DB round-trip per public listing. Added a `getCollectionEntities` regression test that asserts the public projection issues exactly one (approved-only) count query while the owner projection still issues both — it fails against the previous double-query code and passes after the fix.

app/api/v1/collections/[id]/items/[item_id]/approve/route.ts
Comment (llun) [reply app/api/v1/collections/[id]/items/[item_id]/approve/route.ts]: Fixed in 73c4572c. The approve route now returns `ERROR_404` (404) whenever `!item || item.targetActorId !== currentActor.id`, so acting on someone else's item and acting on an unknown item are indistinguishable — closing the 403/404 membership/consent-state oracle and matching the GET detail route's private existence-hiding. Verified against the current head (`app/api/v1/collections/[id]/items/[item_id]/approve/route.ts` lines 35-42). `approve/route.test.ts` now asserts the non-owned-item case returns 404.

app/api/v1/collections/[id]/items/[item_id]/revoke/route.ts
Comment (llun) [reply app/api/v1/collections/[id]/items/[item_id]/revoke/route.ts]: Fixed in 73c4572c, same as the sibling approve route. The revoke route now returns `ERROR_404` (404) when `!item || item.targetActorId !== currentActor.id` (`app/api/v1/collections/[id]/items/[item_id]/revoke/route.ts` lines 35-42), so a stranger can no longer distinguish "someone else's item" (previously 403) from "unknown item" (404) and thus cannot enumerate pending/revoked/approved membership. Covered by the revoke route test.

app/api/v1/collections/[id]/route.test.ts
Comment (llun) [reply app/api/v1/collections/[id]/route.test.ts:141]: Fixed in 73c4572c. The GET detail test now creates an `unlisted` collection in `beforeAll` and adds two tests — "serves an unlisted collection to an anonymous viewer in the approved-only projection" and "serves an unlisted collection to a signed-in stranger" — both asserting 200 with the approved-only public projection, so the load-bearing unlisted-vs-private distinction is covered alongside the existing private→404 cases (`app/api/v1/collections/[id]/route.test.ts`).

app/api/v1/accounts/[id]/in_collections/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/in_collections/route.ts]: Fixed in 73c4572c. The copy-pasted limit/offset parsing and offset Link-header assembly are gone: both `/accounts/:id/collections` and `/accounts/:id/in_collections` now call `parseAccountCollectionsPaging` (`lib/services/collections/accountCollectionsPaging.ts`, built on the shared `clampedLimit`/`clampedOffset` helpers used by the other listing routes) and `buildOffsetPaginationLinkHeader` (`lib/utils/paginationLinkHeader.ts`). The paging + Link-header contract now lives in one place, so the two routes can no longer drift.

app/api/v1/collections/[id]/items/[item_id]/approve/route.ts
```diff
@@ -1,41 +1,43 @@
...
+// added as `pending` and only appear publicly once they approve. Mastodon 4.6
+// addresses the membership by its CollectionItem id; the segment also accepts
+// the member's Account id as an activities.next extension (the first-party UI
+// uses that form). The resolved membership must belong to the authenticated
+// caller. Any token with write:collections can reach this route, so acting on
+// someone else's membership answers 404 — indistinguishable from an unknown
+// item — rather than 403: a 403/404 split would let a stranger who knows the
+// collection id probe whether an arbitrary account is a pending/revoked/approved
+// member, defeating the consent-hiding and private-collection existence-hiding.
 export const POST = traceApiRoute(
```
Comment (llun) [inline app/api/v1/collections/[id]/items/[item_id]/approve/route.ts:28]: **[2·not-a-defect]** simplification — the approve and revoke consent-route POST handlers are near-duplicates; suggestion was to extract a shared handler factory.

**Refuted** (majority of skeptics). The structural claim is accurate: `approve/route.ts` and `revoke/route.ts` differ only in the doc comment, the `traceApiRoute` name, and the `'approved'`/`'revoked'` state literal (already isolated in `setOwnCollectionMembershipState({ state })`). But it is not a defect:

- **No correctness or security impact.** Every input produces the right output in both handlers. The finding's own "failure scenario" is a hypothetical future-maintenance drift risk, not a reproducible wrong behavior — nothing breaks today.
- **Thin per-verb `route.ts` files are the established convention here.** The Next.js App Router structurally requires a `route.ts` per segment, and every Mastodon action pair in this repo (bookmark/unbookmark, pin/unpin, favourite/unfavourite, mute/unmute, reblog/unreblog) ships its own duplicated boilerplate with no shared factory — some have even organically diverged. A `lib/services/collections/` handler factory would be the only action route built that way, contradicting the uniform convention.
- **The two doc comments are deliberately different prose** (opt-in / public-projection vs opt-out / retained-in-private-feed), and keeping them per-file is a benefit. The genuinely reusable logic was already extracted (`resolveCollectionItem`, `addMembersToCollection`).

A low-value DRY nit on two ~30-line siblings whose sole behavioral difference is one already-isolated literal. Per default-false, refuted; filed for the record only.

llun (commented):
## Round 2 code review — collections Mastodon 4.6 alignment

Automated sub-agent review, round 2. The skeptic panel upheld **0** findings this round. **1** finding was refuted by a majority of independent skeptics and is posted inline below with its rationale, on the record only — no change is requested.

For context, round 1's 4 confirmed findings (approve/revoke 403→404 membership oracle, unlisted-GET test coverage, account-collections paging duplication) landed in `73c4572c`, and gemini's `serializers.ts` double-count-query finding landed in `04062369`. All of those threads are already replied-to and resolved.

app/api/v1/collections/[id]/items/[item_id]/approve/route.ts
Comment (llun) [reply app/api/v1/collections/[id]/items/[item_id]/approve/route.ts:28]: Round 2 closeout — no code change; refutation stands.

Verified both files at head: `approve/route.ts` and `revoke/route.ts` are identical apart from the doc comment, the `traceApiRoute` name (`approveCollectionMembership` vs `revokeCollectionMembership`), and the single `'approved'`/`'revoked'` state literal already isolated inside `setOwnCollectionMembershipState({ state })`. Both handlers are correct — every input produces the right output, and the round-1 membership-hiding fix (404 when `!item || item.targetActorId !== currentActor.id`, lines 35-42) landed on both.

No correctness or security impact; the "future drift" scenario is hypothetical, not a reproducible defect. Thin per-verb `route.ts` files are the established convention here — every Mastodon action pair (bookmark/unbookmark, pin/unpin, favourite/unfavourite, mute/unmute, reblog/unreblog) ships its own boilerplate with no shared factory, and the App Router requires a `route.ts` per segment. The genuinely reusable logic (`resolveCollectionItem`, `addMembersToCollection`) was already extracted, and the two doc comments are deliberately different prose. Declining per default-false; resolving.

PR #1218, opened 2026-07-14
fix: keep OAuth app registrations so cached client_ids stay valid

llun:
## Summary

Logging into this instance from **Phanpy** failed with `invalid_client` / **`client_id is required`** — even though `client_id` was plainly present in the request. The login popup ended up on the home timeline and the client never received a code.

From the reported HAR:

```
307  GET /oauth/authorize?client_id=471W4bGb…&code_challenge=…&redirect_uri=https://phanpy.social&…
302  GET /api/auth/oauth2/authorize?response_type=code&client_id=471W4bGb…   ← client_id IS here
302  GET /api/auth/error?error=invalid_client&error_description=client_id+is+required
200  GET /?error=invalid_client&error_description=client_id+is+required      ← home timeline
```

### Root cause

`client_id` was never missing — **the client row had been deleted**. Better Auth's authorize endpoint uses the *same* message for two different checks (`@better-auth/oauth-provider`):

```js
if (!query.client_id) → invalid_client, "client_id is required"
if (!client)          → invalid_client, "client_id is required"   ← this one
```

So `client_id is required` actually means **"client not found"**. Two behaviours combined to produce it:

1. **Mastodon-API clients cache the registration forever.** Phanpy stores the `client_id`/`client_secret` from `POST /api/v1/apps` in `localStorage` (`credentialApplications[instanceURL]`) and only re-registers when that entry is **missing**.
2. **This server deleted registrations after 24h.** `garbageCollectStaleAppRegistrations` in `createApplication.ts` ran on every `/api/v1/apps` POST and deleted any registration older than 24h with no access token, refresh token, or consent — i.e. **any login that was started but never completed**. It was the only `DELETE FROM oauthClient` in the codebase.

So: start a login and abandon it → 24h later the server deletes the client → the client still has the `client_id` cached, never re-registers, and is **permanently wedged**. A time-based cleanup can't fix this — any finite TTL eventually deletes a live cached client.

## Changes

- **`app/api/v1/apps/createApplication.ts`** — remove the collector. Registrations are now durable, exactly as they are in Mastodon (which never expires application records). Abuse stays bounded by the existing per-source registration rate limit (5 per 10 min), which is untouched.
- **`app/(nosidebar)/oauth/authorize/page.tsx`** — validate the client and its `redirect_uri` **before** delegating to Better Auth. An unknown `client_id` now returns `404` instead of being forwarded into Better Auth's `/api/auth/error` → `/` bounce, which rendered a failed login as "nothing happened". Per RFC 6749 §4.1.2.1 an invalid `client_id`/`redirect_uri` must be reported to the user, not redirected to the requested `redirect_uri`.
- **`AGENTS.md`** — pin both invariants so they don't get re-introduced as "cleanup".

Net **−108 lines**. No migration; no schema change.

## Verification

Reproduced and verified end-to-end against the `activities.local` docker-compose stack (containers updated to latest: PostgreSQL 18.4, Redis 8, Traefik v3, Node 24), driving **real Phanpy** at `phanpy.local`.

**Before the fix** — register an app, backdate it past 24h, let another registration trigger the collector, then authorize with the cached `client_id`. Reproduces the HAR byte-for-byte:

```
307 → /api/auth/oauth2/authorize?…client_id=…
302 → /api/auth/error?error=invalid_client&error_description=client_id+is+required
200 → /
```

**After the fix** — same scenario:

```
307 → /api/auth/oauth2/authorize?…client_id=…
302 → /oauth/authorize?…&exp=…&sig=…     ← signed consent query
200 → consent page
```

Full browser run: Phanpy registered an app, cached it in `localStorage`, the registration was aged to 30h and survived a subsequent registration (previously deleted), sign-in resumed the OAuth flow, the consent screen rendered, **Approve** completed the PKCE code exchange, and Phanpy landed logged-in on its Home timeline. Confirmed in the DB — the aged registration now has an issued access token.

An unregistered `client_id` returns `404` instead of the redirect chain to `/`.

**Tests:** `prettier` ✓ · `lint` (0 errors) ✓ · `build` ✓ · `test` **6133 passed** ✓

New regression coverage:
- `createApplication.test.ts` — an old, unused registration (both the rate-limited-source and anonymous shapes) survives a later registration.
- `page.test.tsx` — an unregistered `client_id` returns not-found and is **not** forwarded to Better Auth, with or without a session; an unregistered `redirect_uri` likewise; a registered client still delegates normally.

## Note for deployment

This stops *future* breakage, but it cannot resurrect already-deleted rows. Any client still holding a `client_id` that was already collected (likely including your Phanpy for this instance) must drop its cached credentials once — in Phanpy, clear site data for the instance, or delete the `credentialApplications` key from `localStorage` — after which it re-registers and logs in normally.


app/(nosidebar)/oauth/authorize/page.tsx
```diff
@@ -44,6 +44,31 @@ const Page: FC<Props> = async ({ searchParams }) => {
...
+  // nothing.
+  const client = await database.getClientFromId({ clientId: params.client_id })
+  if (!client) {
+    return notFound()
+  }
+
+  // Validate redirect_uri against registered URIs to prevent open redirect
+  if (
+    params.redirect_uri &&
+    !client.redirectUris.includes(params.redirect_uri)
```
llun: **Regression: this exact-match check is stricter than the Better Auth delegate it now fronts.**

`@better-auth/oauth-provider` (`dist/index.mjs:3864-3871`) matches a redirect_uri as an *exact string* **OR**, when the registered host is a loopback IP, on protocol + hostname + pathname + search — deliberately **ignoring the port** (RFC 8252 §7.3: native apps bind an ephemeral callback port). `createApplication.ts` explicitly advertises loopback support at registration time.

Before this PR the page's exact-match check only ran on the *consent-render* hop, so the first hop was delegated and Better Auth's tolerant matcher was authoritative. Moving the check ahead of the delegation makes the page the sole gate — with stricter semantics.

**Failure scenario:** a native client registers `http://127.0.0.1:8080/callback` and consents once. It restarts, finds :8080 taken, binds an ephemeral port, and requests `redirect_uri=http://127.0.0.1:51234/callback`.
- Before: delegated → Better Auth loopback-matches (port ignored) → stored consent covers the scopes → code issued. **Login works.**
- After: `notFound()` → **404**, Better Auth never reached.

Ironic given this PR exists to stop breaking clients that cache their registration. Fix: mirror Better Auth's matcher — `isLoopbackIP` is a public export of `@better-auth/core/utils/host` (same precedent as the `SafeUrlSchema` import already in `createApplication.ts`).

app/api/v1/apps/createApplication.test.ts
```diff
@@ -616,220 +609,67 @@ describe('createApplication', () => {
...
-        updatedAt: staleCreatedAt
-      },
-      {
-        id: 'stale-anonymous-client-id',
-        clientId: 'stale-anonymous-client',
+  ])(
+    'it keeps an old unused app registration $description when another app registers',
+    async ({ clientId, referenceId, metadata }) => {
+      const staleCreatedAt = new Date('2026-05-10T00:00:00.000Z')
+      const now = new Date('2026-05-12T12:00:00.000Z')
```
llun: **This regression pin is half-dead — the `registered anonymously` case PASSES against the old buggy code.**

Verified empirically: restored `main`'s `createApplication.ts` (collector intact) and ran this file.

```
× it keeps an old unused app registration 'registered from a rate-limited source' … ← fails (good)
✓ it keeps an old unused app registration 'registered anonymously' …               ← PASSES on buggy code
```

**Cause:** both table cases share this single `now`. The old collector was throttled by module-level state (`shouldRunAppRegistrationGc`, 1h interval). Case 1 consumes the GC token at 12:00; case 2 then passes the *identical* `now`, so the collector never runs and the anonymous row survives **for the wrong reason**.

This is the half that matters: the anonymous shape (`referenceId = ''`, `metadata = {"registeredUnauthenticated":true}`) is exactly what Phanpy/Elk produce, since they register without a rate-limit key. A reintroduced collector that reaped only anonymous rows would keep this suite green.

Fix: move `now` into the table with ≥1h separation so neither case can be shadowed by the throttle.

app/api/v1/apps/createApplication.ts
```diff
@@ -205,9 +122,16 @@ export const createApplication = async (
...
+        // intentionally not a cross-database atomic hard cap.
+        //
+        // Registrations are deliberately NOT garbage-collected. Mastodon-API
+        // clients (Phanpy, Elk, Tusky, …) persist the client_id/client_secret
+        // they get from here indefinitely and only re-register when their
+        // stored copy is missing, so deleting a registration permanently wedges
+        // any client still holding it: it keeps presenting a client_id this
+        // server no longer knows and has no way to discover it must register
+        // again. Mastodon never expires application records either. Abuse is
+        // bounded by the rate limit above, not by deleting live clients.
```
llun: **This mitigation claim is not true in the default configuration** (and it is repeated in `AGENTS.md`).

`getAppRegistrationKey` (`route.ts:57`) returns `undefined` when `getTrustProxyIpHeadersConfig()` is false — and that config is `ACTIVITIES_TRUST_PROXY_IP_HEADERS === 'true'`, i.e. **opt-in, default off**. With no key, `getRegistrationReference` → `null` → `isAppRegistrationRateLimited` returns `false` on its first line. **The throttle is entirely disabled on a default deployment**, so "abuse is bounded by the rate limit above" is false there.

The unbounded-insert vector is pre-existing (the throttle was already off by default, and the collector was never a flood defense — rows only became eligible after 24h). Removing the collector is still correct. But the comment and `AGENTS.md` must not assert a bound that isn't there by default, or future readers and agents will trust it.

llun (commented):
Sub-agent review round 1 — three findings, each verified against the code and by running the tests.

app/(nosidebar)/oauth/authorize/page.tsx
llun: Fixed in 3396c1dc4 — confirmed and addressed.

Verified the matcher in `@better-auth/oauth-provider/dist/index.mjs:3864-3871`: exact match **or** `isLoopbackIP(registered.hostname)` with equal protocol/hostname/pathname/search — port excluded.

Extracted `lib/services/oauth/matchRedirectUri.ts`, which reproduces that rule exactly (importing `isLoopbackIP` from `@better-auth/core/utils/host`, the same public-export precedent as the existing `SafeUrlSchema` import), and the page now uses it — so this gate can never be stricter than the delegate it fronts. It also fixes the *pre-existing* 404 such a client hit on its first-ever consent, since the old exact-match check governed the consent-render hop too.

Also dropped the vestigial `params.redirect_uri &&` guard: `redirect_uri` is required by the Zod schema but `z.string()` accepts `''`, so a blank value used to skip the page check and bounce through Better Auth's error page — the exact confusing UX this PR is fixing. It now 404s here.

Covered by 15 table-driven cases in `matchRedirectUri.test.ts` (loopback port-insensitivity, IPv6 `[::1]`, path/scheme/query mismatches, `localhost` being exact-only since it isn't a loopback *IP*, non-loopback ports NOT ignored) plus a page-level test that a loopback redirect differing only by port still delegates.

app/api/v1/apps/createApplication.test.ts
llun: Fixed in 3396c1dc4 — confirmed by reproducing it exactly as described.

Against `main`'s `createApplication.ts`:
```
× 'registered from a rate-limited source' … ← failed (good)
✓ 'registered anonymously' …               ← PASSED on buggy code
```
Root cause was exactly as you diagnosed: one shared `now` plus the collector's 1h per-process throttle, so case 1 consumed the token and the collector never ran for case 2.

Each case now carries its own `now`, 2h apart. Re-verified both directions:
- vs `main`'s source → **both cases fail** (`2 failed`)
- vs this branch → **both pass**

The pin is now load-bearing on the anonymous shape, which is the one that actually matters — that's what Phanpy/Elk produce.

app/api/v1/apps/createApplication.ts
llun: Fixed in 3396c1dc4 — you're right, and the claim was load-bearing precisely because it would be trusted.

Confirmed the chain: `getTrustProxyIpHeadersConfig()` is `ACTIVITIES_TRUST_PROXY_IP_HEADERS === 'true'` (opt-in, default off) → `getAppRegistrationKey` returns `undefined` → `getRegistrationReference` → `null` → `isAppRegistrationRateLimited` returns `false` immediately. No per-source limit on a default deployment.

Both the code comment and `AGENTS.md` now state the real trade-off instead of asserting a bound that isn't there: abandoned registrations accumulate and nothing bounds them by default; that is accepted because the old collector was never a flood defence either (rows only became eligible 24h *after* they were written), so it bought no protection while silently breaking real clients. Both now say any future bound must **reject writes**, never delete registrations.

lib/services/oauth/matchRedirectUri.ts
```diff
@@ -0,0 +1,51 @@
...
+// only. That is Better Auth's behaviour, and matching it is the point.
+export const matchesRegisteredRedirectUri = (
+  registeredUris: string[],
+  requestedUri: string
+): boolean => {
+  // Guard the empty string: `redirect_uri` is a required query param, but
+  // `z.string()` accepts '', and Better Auth rejects a blank redirect_uri too.
+  if (!requestedUri) return false
+
+  return registeredUris.some((registeredUri) => {
```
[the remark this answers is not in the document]
llun: Thanks — but this is a false positive for this code, and I'm declining it.

`registeredUris` is typed `string[]` (non-nullable), and its only caller passes `client.redirectUris`, which comes from `parseClientRow` → `Client.parse(...)` where the schema is `redirectUris: z.string().url().array().min(1)` (`lib/types/oauth2/client.ts:10`). So `getClientFromId` either returns a **non-null, non-empty** array of valid URLs or throws before this function is ever reached — a null/undefined `registeredUris` is unreachable at both the type and runtime level.

Adopting the suggestion would require widening the parameter to `string[] | null | undefined`, which *weakens* the contract and invites callers to pass null into what is a required OAuth field.

Also flagging for anyone reading: the suggested snippet replaces the body with a plain `registeredUris?.some(uri => uri === requestedUri)`, which drops the RFC 8252 §7.3 loopback (port-insensitive) matching this function exists to provide — applying it verbatim would reintroduce the exact bug the previous review round fixed. The `[]`-returns-`false` case is already covered by a test (`no registered URIs at all`).

PR #1226, opened 2026-07-16
fix: show remote actor profile data and posts in the Mastodon API

llun:
## Summary

Opening a remote user's profile (e.g. `@mttvll@mastodon.social`) in a Mastodon client (official iOS app, Ivory, etc.) showed **0 followers, 0 following, a wrong post count, the bare username instead of the display name, no bio, a wrong "Request to follow" (locked) button, and an empty Activity tab** — while the same profile looked fine on mastodon.social (5,370 followers / 519 following / 641 posts for this user).

Root causes:

1. **Account counts came only from local counters.** `getMastodonActors` builds `followers_count` / `following_count` / `statuses_count` from the `counters` table, which for a remote actor only tracks relationships/statuses this instance has seen — nothing ever fetched the totals the remote server advertises on its `followers` / `following` / `outbox` collections. (The web UI profile page fetches these live per view, but never persisted them, and the Mastodon API never fetched them at all.)
2. **`recordActorIfNeeded` dropped most profile data.** It persisted neither `name`, `summary`, header image, profile metadata fields, nor `manuallyApprovesFollowers` — so `locked` fell back to `true` ("Request to follow" for every remote account, even unlocked ones) and the display name fell back to the username.
3. **`GET /api/v1/accounts/:id/statuses` served only locally-stored statuses**, which is empty for a remote actor nobody on the instance follows.
4. Existing actors were never refreshed by the client-visible flows: lookup/search only *record missing* actors, so stale rows stayed stale.

### Changes

- **`recordActorIfNeeded`** now persists the full remote profile (name, bio, avatar, header, `manuallyApprovesFollowers`, `PropertyValue` metadata fields) on both the create and 3-day-refresh paths, and syncs the remote-advertised collection totals into the actor's counter rows via the new `getActorCollectionCounts` helper (3 parallel collection-root fetches, best-effort). It also refreshes a *fresh* actor whose counters were never synced, so actors recorded before this fix heal on the next view instead of waiting out the refresh interval.
- **`GET /api/v1/accounts/:id`** refreshes a known, stale remote actor before serialization (authenticated viewers only, existing actors only — anonymous requests can't trigger remote fetches for arbitrary ids). Failures fall back to the stored profile.
- **`GET /api/v1/accounts/:id/statuses`** falls back to fetching the actor's recent public posts live from their outbox (new `getRemoteActorStatuses` service) when the local store can't fill the first page for an authenticated viewer. Statuses are display-only (not persisted), filtered to public/unlisted, respect `exclude_replies` / `exclude_reblogs` / `only_media`, and drop reblogs whose original author is unknown locally (the serializer can't render those). A live page carries no `Link` pagination headers since remote ids can't cursor the local store.
- **Cursor decoding fix:** the statuses endpoint passed the client-echoed `max_id`/`min_id`/`since_id` (urlToId-encoded) straight to the database, which silently ignores unknown ids — so page 2 of a profile repeated page 1. Cursors are now decoded with `safeIdToUrl` (undecodable → 400), matching the timeline endpoints.
- **`getProfileData`** (web UI remote profile) persists the counts it already fetches, so the web view and the Mastodon API stay consistent.
- **DB:** new `setActorCounters` / `hasActorCounters` on the actor mixin; `CreateActorParams` gains `manuallyApprovesFollowers` and `fields`. No migrations — the `counters` table already exists, so both schema dumps are untouched.
- **Docs:** documented the live-fetch divergence from stock Mastodon in `docs/mastodon-api-compatibility.md`.

### Verification

Ran locally with SQLite + mock user (`scripts/mock/createMockUser.ts`) against the real fediverse:

`GET /api/v1/accounts/lookup?acct=mttvll@mastodon.social&resolve=true` (before → after):

| field | before | after |
| --- | --- | --- |
| `display_name` | `""` (client falls back to `mttvll`) | `Matteo` |
| `note` | `""` | real bio |
| `locked` | `true` ("Request to follow") | `false` |
| `followers_count` | 0 | 5370 |
| `following_count` | 0 | 519 |
| `statuses_count` | local-only count | 641 |
| `fields` | `[]` | 2 profile metadata fields |

`GET /api/v1/accounts/:id/statuses?exclude_replies=true` returned the user's actual recent public posts (including media attachments and a reblog) instead of `[]`, with no `Link` header on the live page. The web profile page (`/@mttvll@mastodon.social`) was verified in a real browser and shows the same counts and posts.

## Checklist

- [x] If migrations changed: BOTH `migrations/schema.sql` and `migrations/schema.sqlite.sql` are regenerated in this PR (no migrations changed)

https://claude.ai/code/session_011NQyx4bZjX2GFGRRw5Kaht

---
_Generated by [Claude Code](https://claude.ai/code/session_011NQyx4bZjX2GFGRRw5Kaht)_

### DIFF
--- a/app/(timeline)/[actor]/getProfileData.ts
+++ b/app/(timeline)/[actor]/getProfileData.ts
@@ -147,6 +147,29 @@ export const getProfileData = async (
     getActorFollowers({ person, ...signingParams })
   ])
 
+  // Persist the freshly-fetched collection sizes for known actors so the
+  // Mastodon API (which reads the counter rows) serves the same counts this
+  // page displays. The collection helpers report 0 for both "empty" and
+  // "fetch failed", so only positive values are trusted here — null preserves
+  // the stored counter. Best-effort — the page renders from the live values
+  // either way.
+  if (persistedActor) {
+    await database
+      .setActorCounters({
+        actorId: person.id,
+        followersCount: actorFollowersResponse.followerCount || null,
+        followingCount: actorFollowingResponse.followingCount || null,
+        statusCount: actorPostsResponse.statusesCount || null
+      })
+      .catch((error) => {
+        logger.warn({
+          message: 'Failed to persist remote actor collection counts',
+          actorId: person.id,
+          error: error instanceof Error ? error.message : String(error)
+        })
+      })
+  }
+
   return {
     ...actorPostsResponse,
     person,
--- a/app/(timeline)/[actor]/getProfileData.ts
+++ b/app/(timeline)/[actor]/getProfileData.ts
@@ -1,14 +1,13 @@
-import { getActorFollowers } from '@/lib/activities/getActorFollowers'
-import { getActorFollowing } from '@/lib/activities/getActorFollowing'
+import { getPersistableProfile } from '@/lib/actions/utils'
+import { getActorCollectionCounts } from '@/lib/activities/getActorCollectionCounts'
 import { getActorPerson } from '@/lib/activities/getActorPerson'
 import { getActorPosts } from '@/lib/activities/getActorPosts'
 import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
 import { Database } from '@/lib/database/types'
-import { getFederationSigningActor } from '@/lib/services/federation/getFederationSigningActor'
+import { getFederationSigningActorSafe } from '@/lib/services/federation/getFederationSigningActor'
 import { Actor } from '@/lib/types/activitypub'
 import { Attachment } from '@/lib/types/domain/attachment'
 import { Status } from '@/lib/types/domain/status'
-import { getActorImageUrl } from '@/lib/utils/activitypubActor'
 import { getPersonFromActor } from '@/lib/utils/getPersonFromActor'
 import { logger } from '@/lib/utils/logger'
 
@@ -88,27 +87,14 @@ export const getProfileData = async (
   // resolvable). The instance actor always exists, always has a private key,
   // and is served at a publicly resolvable URL so the remote can fetch its key
   // and verify the signature. This is the same headless signer used by the
-  // federation jobs and relay/follow flows (getFederationSigningActor); without
-  // it, secure-mode remote profiles 404.
+  // federation jobs and relay/follow flows; without it, secure-mode remote
+  // profiles 404. Resolution is best-effort and degrades to an unsigned fetch.
   //
   // WebFinger discovery and signer resolution are independent, so resolve them
   // concurrently to avoid stacking their latencies on the profile render.
-  // Signer resolution is best-effort: a missing/failed instance actor must not
-  // turn a clean 404 (unknown actor) into a 500, so a failure degrades to an
-  // unsigned fetch via the `signingActor`-less branch below.
   const [actorId, signingActor] = await Promise.all([
     getWebfingerSelf({ account: actorHandle.slice(1) }),
-    getFederationSigningActor(database).catch((error) => {
-      // Degrade to an unsigned fetch, but surface the failure so a persistently
-      // broken signer (which would silently 404 every secure-mode profile)
-      // remains diagnosable rather than vanishing.
-      logger.warn({
-        message:
-          'Failed to resolve federation signing actor for remote profile fetch; falling back to an unsigned request',
-        error: error instanceof Error ? error.message : String(error)
-      })
-      return undefined
-    })
+    getFederationSigningActorSafe(database, 'for remote profile fetch')
   ])
   if (!actorId) return null
 
@@ -117,48 +103,40 @@ export const getProfileData = async (
   if (!person) return null
 
   if (persistedActor) {
+    // Same field set recordActorIfNeeded persists, so the web profile page
+    // and the Mastodon API refresh paths write consistent snapshots (including
+    // metadata fields and the locked state).
     await database.updateActor({
       actorId: person.id,
-      name: person.name,
-      summary: person.summary || '',
-      iconUrl: getActorImageUrl(person.icon),
-      headerImageUrl: getActorImageUrl(person.image),
-      publicKey: person.publicKey.publicKeyPem,
-      followersUrl: person.followers,
-      inboxUrl: person.inbox,
-      sharedInboxUrl: person.endpoints?.sharedInbox || ''
+      ...getPersistableProfile(person)
     })
   }
 
-  const [
-    actorPostsResponse,
-    attachments,
-    actorFollowingResponse,
-    actorFollowersResponse
-  ] = await Promise.all([
-    getActorPosts({
-      database,
-      person,
-      pageUrl: options.statusPageUrl,
-      ...signingParams
-    }),
-    database.getAttachmentsForActor({ actorId: person.id }),
-    getActorFollowing({ person, ...signingParams }),
-    getActorFollowers({ person, ...signingParams })
-  ])
+  const [actorPostsResponse, attachments, collectionCounts] = await Promise.all(
+    [
+      getActorPosts({
+        database,
+        person,
+        pageUrl: options.statusPageUrl,
+        ...signingParams
+      }),
+      database.getAttachmentsForActor({ actorId: person.id }),
+      getActorCollectionCounts({ person, ...signingParams })
+    ]
+  )
 
   // Persist the freshly-fetched collection sizes for known actors so the
   // Mastodon API (which reads the counter rows) serves the same counts this
-  // page displays. The collection helpers report 0 for both "empty" and
-  // "fetch failed", so only positive values are trusted here — null preserves
-  // the stored counter. Best-effort — the page renders from the live values
-  // either way.
+  // page displays. getActorCollectionCounts distinguishes a fetch failure
+  // (null, preserves the stored counter) from a real zero. getActorPosts
+  // reports 0 for both, so only positive status counts are trusted here.
+  // Best-effort — the page renders from the live values either way.
   if (persistedActor) {
     await database
       .setActorCounters({
         actorId: person.id,
-        followersCount: actorFollowersResponse.followerCount || null,
-        followingCount: actorFollowingResponse.followingCount || null,
+        followersCount: collectionCounts.followersCount,
+        followingCount: collectionCounts.followingCount,
         statusCount: actorPostsResponse.statusesCount || null
       })
       .catch((error) => {
@@ -178,8 +156,8 @@ export const getProfileData = async (
       prevPageUrl: actorPostsResponse.prevPageUrl ?? null
     },
     attachments,
-    followingCount: actorFollowingResponse.followingCount,
-    followersCount: actorFollowersResponse.followerCount,
+    followingCount: collectionCounts.followingCount ?? 0,
+    followersCount: collectionCounts.followersCount ?? 0,
     isInternalAccount: false,
     hasFitnessData: false
   }
--- a/app/api/v1/accounts/[id]/remove_from_followers/route.ts
+++ b/app/api/v1/accounts/[id]/remove_from_followers/route.ts
@@ -45,14 +45,15 @@ export const POST = traceApiRoute(
       })
 
       if (follow && follow.status === FollowStatus.enum.Accepted) {
+        // updateFollowStatus decrements both actors' follow counters inside
+        // its transaction. Do NOT recount them from the local follows table
+        // here: for a remote target that recount would overwrite the
+        // remote-advertised totalFollowing synced from their server with a
+        // local-only value.
         await database.updateFollowStatus({
           followId: follow.id,
           status: FollowStatus.enum.Undo
         })
-        await Promise.all([
-          database.updateActorFollowersCount(currentActor.id),
-          database.updateActorFollowingCount(targetActorId)
-        ])
 
         // Federate the removal to remote followers so they learn they were
         // removed (Mastodon sends a Reject for the original Follow).
--- a/app/api/v1/accounts/[id]/route.ts
+++ b/app/api/v1/accounts/[id]/route.ts
@@ -1,3 +1,4 @@
+import { recordActorIfNeeded } from '@/lib/actions/utils'
 import { localizeAccount } from '@/lib/services/accounts/localizeAccount'
 import {
   OptionalOAuthGuard,
@@ -6,6 +7,7 @@ import {
 import { headerHost } from '@/lib/services/guards/headerHost'
 import { Scope } from '@/lib/types/database/operations'
 import { HttpMethod } from '@/lib/utils/http-headers'
+import { logger } from '@/lib/utils/logger'
 import { apiCorsError, apiResponse, defaultOptions } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
 import { idToUrl } from '@/lib/utils/urlToId'
@@ -29,10 +31,33 @@ export const GET = traceApiRoute(
   OptionalOAuthGuard<Params>(
     [Scope.enum.read, Scope.enum['read:accounts']],
     async (req, context) => {
-      const { database, params } = context
+      const { database, currentActor, params } = context
       const encodedAccountId = (await params).id
       if (!encodedAccountId) return apiCorsError(req, CORS_HEADERS, 400)
       const id = idToUrl(encodedAccountId)
+
+      // Opening a profile is the moment a client needs current remote data
+      // (display name, bio, lock state, follower/following/status counts), so
+      // refresh known remote actors here. recordActorIfNeeded is a no-op for
+      // recently-synced actors, only refreshing stale ones, and any failure
+      // falls back to the stored profile. Gated on an authenticated viewer and
+      // an already-known actor so anonymous requests never trigger remote
+      // fetches for arbitrary ids.
+      if (currentActor) {
+        const persistedActor = await database.getActorFromId({ id })
+        if (persistedActor && !persistedActor.privateKey) {
+          await recordActorIfNeeded({ actorId: id, database }).catch(
+            (error) => {
+              logger.warn({
+                message: 'Failed to refresh remote actor for account view',
+                actorId: id,
+                error: error instanceof Error ? error.message : String(error)
+              })
+            }
+          )
+        }
+      }
+
       const actor = await database.getMastodonActorFromId({ id })
       if (!actor) return apiCorsError(req, CORS_HEADERS, 404)
       return apiResponse({
--- a/app/api/v1/accounts/[id]/route.ts
+++ b/app/api/v1/accounts/[id]/route.ts
@@ -44,8 +44,11 @@ export const GET = traceApiRoute(
       // an already-known actor so anonymous requests never trigger remote
       // fetches for arbitrary ids.
       if (currentActor) {
-        const persistedActor = await database.getActorFromId({ id })
-        if (persistedActor && !persistedActor.privateKey) {
+        const [persistedActor, isInternalActor] = await Promise.all([
+          database.getActorFromId({ id }),
+          database.isInternalActor({ actorId: id })
+        ])
+        if (persistedActor && !isInternalActor) {
           await recordActorIfNeeded({ actorId: id, database }).catch(
             (error) => {
               logger.warn({
--- a/app/api/v1/accounts/[id]/route.ts
+++ b/app/api/v1/accounts/[id]/route.ts
@@ -44,11 +44,15 @@ export const GET = traceApiRoute(
       // an already-known actor so anonymous requests never trigger remote
       // fetches for arbitrary ids.
       if (currentActor) {
-        const [persistedActor, isInternalActor] = await Promise.all([
-          database.getActorFromId({ id }),
-          database.isInternalActor({ actorId: id })
-        ])
-        if (persistedActor && !isInternalActor) {
+        const persistedActor = await database.getActorFromId({ id })
+        // Internal actors (account-backed users and the headless signer, which
+        // always carries a private key) are never refreshed from the network.
+        // The loaded row already answers this, so no isInternalActor query.
+        if (
+          persistedActor &&
+          !persistedActor.account &&
+          !persistedActor.privateKey
+        ) {
           await recordActorIfNeeded({ actorId: id, database }).catch(
             (error) => {
               logger.warn({
--- a/app/api/v1/accounts/[id]/statuses/route.ts
+++ b/app/api/v1/accounts/[id]/statuses/route.ts
@@ -94,10 +94,10 @@ export const GET = traceApiRoute(
       // stores raw status URLs and silently ignores an unknown cursor, which
       // would serve the same first page over and over. An undecodable cursor
       // is a deliberate 400, using the same decodeCursor rule as the timeline
-      // endpoints.
-      const maxId = decodeCursor(encodedMaxId)
-      const minId = decodeCursor(encodedMinId)
-      const sinceId = decodeCursor(encodedSinceId)
+      // endpoints; an empty cursor param means "no cursor" there too.
+      const maxId = decodeCursor(encodedMaxId || undefined)
+      const minId = decodeCursor(encodedMinId || undefined)
+      const sinceId = decodeCursor(encodedSinceId || undefined)
       if (maxId === null || minId === null || sinceId === null) {
         return apiResponse({
           req,
@@ -238,19 +238,21 @@ export const GET = traceApiRoute(
       // them — the outbox only exposes public/unlisted posts, so a follower's
       // locally-federated followers-only statuses must survive. On id
       // collisions keep the local copy: it carries viewer interaction state
-      // (favourited/bookmarked/own reblog).
+      // (favourited/bookmarked/own reblog). Every readable local status is
+      // kept (the live-fetch gate guarantees there are fewer than `limit` of
+      // them) and the newest live statuses top up the remaining slots — a
+      // live page carries no pagination links, so a follower-only local post
+      // pushed off the page would otherwise become unreachable.
       const localStatusIds = new Set(
         readableStatuses.map((status) => status.id)
       )
       const visibleStatuses = servedLiveStatuses
         ? [
             ...readableStatuses,
-            ...liveRemoteStatuses.filter(
-              (status) => !localStatusIds.has(status.id)
-            )
-          ]
-            .sort((a, b) => b.createdAt - a.createdAt)
-            .slice(0, limit)
+            ...liveRemoteStatuses
+              .filter((status) => !localStatusIds.has(status.id))
+              .slice(0, Math.max(0, limit - readableStatuses.length))
+          ].sort((a, b) => b.createdAt - a.createdAt)
         : readableStatuses.slice(0, limit)
       const visibleStatusIds = visibleStatuses.map((status) => status.id)
       const pinnedStatusIds =
--- a/app/api/v1/accounts/[id]/statuses/route.ts
+++ b/app/api/v1/accounts/[id]/statuses/route.ts
@@ -8,6 +8,7 @@ import { headerHost } from '@/lib/services/guards/headerHost'
 import { getMastodonStatuses } from '@/lib/services/mastodon/getMastodonStatus'
 import { getRemoteActorStatuses } from '@/lib/services/mastodon/getRemoteActorStatuses'
 import { canActorReadStatus } from '@/lib/services/statusAccess'
+import { decodeCursor } from '@/lib/services/timelines/request'
 import { Scope } from '@/lib/types/database/operations'
 import { FollowStatus } from '@/lib/types/domain/follow'
 import { type Status, StatusType } from '@/lib/types/domain/status'
@@ -20,7 +21,7 @@ import {
   defaultOptions
 } from '@/lib/utils/response'
 import { traceApiRoute } from '@/lib/utils/traceApiRoute'
-import { idToUrl, safeIdToUrl } from '@/lib/utils/urlToId'
+import { idToUrl } from '@/lib/utils/urlToId'
 
 const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
 const MAX_STATUS_SCAN_BATCHES = 10
@@ -92,10 +93,11 @@ export const GET = traceApiRoute(
       // status URLs), so decode the cursors before querying — the database
       // stores raw status URLs and silently ignores an unknown cursor, which
       // would serve the same first page over and over. An undecodable cursor
-      // is a deliberate 400, matching the timeline endpoints.
-      const maxId = encodedMaxId ? safeIdToUrl(encodedMaxId) : undefined
-      const minId = encodedMinId ? safeIdToUrl(encodedMinId) : undefined
-      const sinceId = encodedSinceId ? safeIdToUrl(encodedSinceId) : undefined
+      // is a deliberate 400, using the same decodeCursor rule as the timeline
+      // endpoints.
+      const maxId = decodeCursor(encodedMaxId)
+      const minId = decodeCursor(encodedMinId)
+      const sinceId = decodeCursor(encodedSinceId)
       if (maxId === null || minId === null || sinceId === null) {
         return apiResponse({
           req,
@@ -216,10 +218,10 @@ export const GET = traceApiRoute(
       if (
         currentActor &&
         isFirstPage &&
-        !actor.privateKey &&
         readableStatuses.length < limit &&
         parsedParams.pinned !== true &&
-        !parsedParams.tagged
+        !parsedParams.tagged &&
+        !(await database.isInternalActor({ actorId: id }))
       ) {
         liveRemoteStatuses = await getRemoteActorStatuses({
           database,
@@ -232,8 +234,23 @@ export const GET = traceApiRoute(
       }
       const servedLiveStatuses = liveRemoteStatuses.length > 0
 
+      // Merge live statuses with the locally-stored ones instead of replacing
+      // them — the outbox only exposes public/unlisted posts, so a follower's
+      // locally-federated followers-only statuses must survive. On id
+      // collisions keep the local copy: it carries viewer interaction state
+      // (favourited/bookmarked/own reblog).
+      const localStatusIds = new Set(
+        readableStatuses.map((status) => status.id)
+      )
       const visibleStatuses = servedLiveStatuses
-        ? liveRemoteStatuses
+        ? [
+            ...readableStatuses,
+            ...liveRemoteStatuses.filter(
+              (status) => !localStatusIds.has(status.id)
+            )
+          ]
+            .sort((a, b) => b.createdAt - a.createdAt)
+            .slice(0, limit)
         : readableStatuses.slice(0, limit)
       const visibleStatusIds = visibleStatuses.map((status) => status.id)
       const pinnedStatusIds =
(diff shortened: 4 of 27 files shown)


lib/actions/utils.ts
```diff
@@ -58,46 +114,48 @@ export const recordActorIfNeeded = async ({
...
-      publicKey: person.publicKey.publicKeyPem || ''
-    })
-    return actor ?? undefined
+  const isStale =
+    currentTime - existingActor.updatedAt > REMOTE_ACTOR_REFRESH_INTERVAL_MS
+  // Also refresh when the actor's collection counters were never synced —
+  // remote actors recorded before counter syncing existed would otherwise
+  // keep showing zero followers/following until the next stale refresh.
+  const needsCounterSync =
+    !isStale && !(await database.hasActorCounters({ actorId }))
```
llun: **[sub-agent review — CONFIRMED]** `hasActorCounters` uses counter-**row existence** as the "remote counts were synced" signal, but local accumulation creates the same rows: an accepted follow creates `total-followers:{actorId}` / `total-following:{actorId}` (via `increaseCounterValue`'s `ensureCounterRow`), and a federated note creates `total-status:{actorId}`. A legacy remote actor with a mutual follow and one stored note has all three rows with local-only values (e.g. 1/1/1), so the backfill this PR promises is skipped for exactly the most-viewed actors, and they keep serving `followers_count: 1` instead of the remote total until the 3-day stale refresh.

Fix: record the sync with an explicit marker row (`remote-counts-synced:{actorId}` in the counters KV table — no migration needed) written by `setActorCounters`, and make `hasActorCounters` check that marker instead of the three value rows.

---

lib/actions/utils.ts
```diff
@@ -58,46 +114,48 @@ export const recordActorIfNeeded = async ({
...
-  return existingActor
+
+  const resolvedSigningActor = await getResolvedSigningActor()
+  const person = await getActorPerson({
+    actorId,
+    signingActor: resolvedSigningActor
+  })
+  // A counter-only sync must not degrade the previous behavior of returning
+  // the stored actor when the remote fetch fails.
+  if (!person) return isStale ? undefined : existingActor
```
llun: **[sub-agent review — CONFIRMED]** When the counter-only sync path fails to fetch the person (dead/410 instance, secure-mode rejection), this returns the stored actor **without writing any counter rows**, so `hasActorCounters` stays false and every subsequent `recordActorIfNeeded` call — which runs on the inbox hot path for every inbound note/announce/poll referencing the actor — re-attempts the blocking remote fetch (up to the 10s request timeout, with retries) forever. Before this PR the fresh-actor path had zero network I/O.

Fix: on a failed person fetch during a counter-only sync, still stamp the sync marker (`setActorCounters({ actorId })`) so the retry is bounded by the 3-day stale-refresh window instead of firing per-activity.

---

lib/services/mastodon/getRemoteActorStatuses.ts
```diff
@@ -0,0 +1,105 @@
...
+        // response.
+        const visibility = getVisibility(status.to, status.cc)
+        if (visibility !== 'public' && visibility !== 'unlisted') return false
+
+        if (status.type === StatusType.enum.Announce) {
+          if (excludeReblogs) return false
+          return knownAuthorIds.has(status.originalStatus.actorId)
+        }
+        if (excludeReplies && status.reply) return false
+        if (onlyMedia && status.attachments.length === 0) return false
```
llun, line 103: **[sub-agent review — CONFIRMED]** The Announce branch returns before the `onlyMedia` check ever runs, so live-served pages include text-only boosts when `only_media=true` — the media tab. The local DB path (`getActorStatuses` with `onlyMedia`) excludes announce rows entirely, and Mastodon's media tab never shows boosts, so the live path diverges from both.

Fix: drop announces when `onlyMedia` is set (`if (excludeReblogs || onlyMedia) return false`), and skip the `announceAuthorIds` DB lookup in that case too since its result can never be consulted.

---

lib/database/sql/actor.ts
```diff
@@ -995,6 +1004,45 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
...
+  // server advertises (followers/following/outbox totalItems). A null or
+  // undefined value keeps the locally-accumulated counter but still creates
+  // the row, so hasActorCounters reports the actor as synced even when the
+  // remote hides a collection's total.
+  async setActorCounters({
+    actorId,
+    followersCount,
+    followingCount,
+    statusCount
+  }: SetActorCountersParams) {
```
llun, line 991: **[sub-agent review — CONFIRMED]** Remote-synced counters written here are silently clobbered by the local recount in `updateActorFollowersCount`/`updateActorFollowingCount`, which `POST /api/v1/accounts/:id/remove_from_followers` still calls with a possibly-remote `targetActorId`: remote actor R (following 500 accounts, `total-following:R = 500` synced from R's server) follows local user A; A removes R from followers → `updateActorFollowingCount(R)` recounts local `follows` rows (0 after the Undo) and overwrites the counter with 0. The account then reports `following_count: 0` until R's next stale refresh.

Those two recount calls are also redundant for local actors — `updateFollowStatus` already decrements both actors' counters inside its transaction on the Accepted→Undo transition (lib/database/sql/follow.ts). Fix: drop the recount calls from the remove_from_followers route.

---

app/api/v1/accounts/[id]/statuses/route.ts
```diff
@@ -82,11 +83,28 @@ export const GET = traceApiRoute(
...
       } = parsedParams
 
+      // Clients echo back the opaque ids this endpoint emits (urlToId-encoded
+      // status URLs), so decode the cursors before querying — the database
+      // stores raw status URLs and silently ignores an unknown cursor, which
+      // would serve the same first page over and over. An undecodable cursor
+      // is a deliberate 400, matching the timeline endpoints.
+      const maxId = encodedMaxId ? safeIdToUrl(encodedMaxId) : undefined
+      const minId = encodedMinId ? safeIdToUrl(encodedMinId) : undefined
+      const sinceId = encodedSinceId ? safeIdToUrl(encodedSinceId) : undefined
```
Comment (llun) [inline app/api/v1/accounts/[id]/statuses/route.ts]: **[sub-agent review — CONFIRMED]** Two issues with the inline cursor decoding:

1. It re-implements the `decodeCursor` convention (`undefined` = absent, `null` = undecodable → 400) that `lib/services/timelines/request.ts` already owns — the comment even says it matches the timeline endpoints. Export and reuse `decodeCursor` so the rule can't drift.
2. A **raw status URL** cursor (`max_id=https://remote.example/users/x/statuses/123`) — the only cursor form that previously worked against the DB's raw-URL ids — is now mangled by `idToUrl`'s scheme handling (`https://…` → `https:////…`), which still parses as a URL, so it is neither matched nor 400'd: the cursor is silently ignored and the client gets the first page forever. `safeIdToUrl` should pass through values that already start with `http(s)://` (encoded ids never do, so there's no ambiguity).

---

app/(timeline)/[actor]/getProfileData.ts
```diff
@@ -147,6 +147,29 @@ export const getProfileData = async (
...
   ])
 
+  // Persist the freshly-fetched collection sizes for known actors so the
+  // Mastodon API (which reads the counter rows) serves the same counts this
+  // page displays. The collection helpers report 0 for both "empty" and
+  // "fetch failed", so only positive values are trusted here — null preserves
+  // the stored counter. Best-effort — the page renders from the live values
+  // either way.
+  if (persistedActor) {
+    await database
```
Comment (llun) [inline app/(timeline)/[actor]/getProfileData.ts:135]: **[sub-agent review — CONFIRMED]** The web profile page's refresh path is now the second, divergent writer of remote-actor data: the `updateActor` call above persists neither `manuallyApprovesFollowers`, `fields`, nor `type` (which `recordActorIfNeeded` now persists via `getPersistableProfile`), so which lock state / metadata fields Mastodon clients see depends on which surface refreshed the actor last. This counter-persist block likewise duplicates the trust rule with `|| null` semantics that differ from the collection-counts helper (a genuine `totalItems: 0` can never persist through this path).

Fix: export `getPersistableProfile` and use it in the `updateActor` call here, and fetch counts via `getActorCollectionCounts` (which distinguishes failure `null` from a real 0) instead of the page-oriented `getActorFollowers`/`getActorFollowing` helpers, which also skips their unused first-page fetches.

---

lib/activities/getActorCollectionCounts.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!url) return null
+
+  try {
+    const response = await request({
+      url,
+      headers: activityPubRequestHeaders({ url, signingActor })
+    })
+    if (response.statusCode !== 200) return null
+
+    const collection = JSON.parse(response.body) as OrderedCollection
```
llun: **[sub-agent review — CONFIRMED (duplication) / PLAUSIBLE (compaction)]** Two notes on this fetch:

1. `getCollectionTotalItems` re-implements the collection-root fetch (`request` + `activityPubRequestHeaders` + status check + `JSON.parse as OrderedCollection`) that `getActorCollections` already performs — two parallel implementations of the same AP fetch will drift (this one already lacks that helper's tracing; that one lacks this one's totalItems validation). Extract a shared `fetchCollectionRoot` helper both consume.
2. Convention note: this is a new untrusted-remote-AP parse path without `compactActivityPub`. The AGENTS.md compaction rule's bolded scope is notes/activities/actors (collections aren't listed), and the pre-existing `getActorCollections` parses collection roots the same uncompacted way, so this follows accepted precedent — but a remote emitting non-compact JSON-LD (`as:totalItems`, `{"@value": n}`) yields `null` counts and the counters keep their local values. Graceful degradation, but worth a maintainer decision on whether collection fetches should compact too (it would add a jsonld run per collection root on the inbox path).

---

app/api/v1/accounts/[id]/route.ts
```diff
@@ -29,10 +31,33 @@ export const GET = traceApiRoute(
...
+      // (display name, bio, lock state, follower/following/status counts), so
+      // refresh known remote actors here. recordActorIfNeeded is a no-op for
+      // recently-synced actors, only refreshing stale ones, and any failure
+      // falls back to the stored profile. Gated on an authenticated viewer and
+      // an already-known actor so anonymous requests never trigger remote
+      // fetches for arbitrary ids.
+      if (currentActor) {
+        const persistedActor = await database.getActorFromId({ id })
+        if (persistedActor && !persistedActor.privateKey) {
+          await recordActorIfNeeded({ actorId: id, database }).catch(
```
Comment (llun) [inline app/api/v1/accounts/[id]/route.ts:56]: **[sub-agent review — PLAUSIBLE, acknowledged trade-off]** Awaiting `recordActorIfNeeded` inline means an authenticated profile open of a **stale** remote actor whose instance is down blocks on the person fetch's full timeout budget before serving the (unchanged) stored data, and because a failed refresh doesn't bump `updatedAt`, every subsequent view of that actor pays the timeout again — there is no negative caching.

The synchronous await is deliberate (the first profile open must show correct counts — that's the bug this PR fixes) and the cost is bounded to actors >3 days stale on unreachable servers, but a follow-up could stamp a short-lived "refresh attempted" marker on failure so dead instances don't add latency to every view. Also worth noting the sibling gate here duplicates `!privateKey` as the "is remote" test — `database.isInternalActor` is the named abstraction for this.

---

lib/actions/utils.ts
```diff
@@ -58,46 +114,48 @@ export const recordActorIfNeeded = async ({
...
       domain: new URL(person.id).hostname,
-      followersUrl: person.followers ?? '',
-      inboxUrl: person.inbox,
-      sharedInboxUrl: person.endpoints?.sharedInbox ?? person.inbox,
-      ...(iconUrl ? { iconUrl } : {}),
-      publicKey: person.publicKey.publicKeyPem || '',
+      ...getPersistableProfile(person),
       createdAt: new Date(person.published ?? Date.now()).getTime()
     })
+    await syncActorCollectionCounts(database, person, resolvedSigningActor)
```
llun, line 133: **[sub-agent review — PLAUSIBLE, acknowledged trade-off]** The create path now awaits three extra collection-root fetches before returning, adding synchronous latency to every caller that records a new remote actor — including `POST /api/v1/accounts/:id/block` (blocking a not-yet-known harasser now also waits on their server's followers/following/outbox endpoints), handle search/lookup, and inbox `createFollower` processing.

The fetches run in parallel (≈1 extra RTT typical, bounded by the 10s request timeout) and this mirrors what Mastodon's ProcessAccountService does on discovery, so first-render correctness was chosen over latency here. If it bites in practice, the sync could move to a fire-and-forget/job for the non-user-facing callers (inbox jobs, block) while staying synchronous for lookup/search.

---

llun (commented):
Sub-agent code-review round 1 (8 finder angles: line-scan, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions — findings verified before posting). CONFIRMED items are being fixed in the next commit on this branch; PLAUSIBLE items are documented trade-offs with proposed follow-ups. Each thread will be replied to and resolved as it's addressed.

---

lib/services/mastodon/getRemoteActorStatuses.ts
```diff
@@ -0,0 +1,105 @@
...
+    // The Mastodon serializer resolves a reblog's author from the database, so
+    // an Announce whose original author is unknown locally can't be rendered.
+    // Keep only announces with locally-known original authors.
+    const announceAuthorIds = [
+      ...new Set(
+        statuses
+          .filter((status) => status.type === StatusType.enum.Announce)
+          .map((status) => status.originalStatus.actorId)
+      )
+    ]
```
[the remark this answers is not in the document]
llun: Fixed in b3f7736 — the announce-author collection now filters on `status.originalStatus?.actorId` before mapping, and the whole lookup is skipped when the filter would drop every announce anyway (`exclude_reblogs`/`only_media`). Added a regression case (`drops malformed reblogs with a missing original status`) to the service tests.

---

lib/services/mastodon/getRemoteActorStatuses.ts
```diff
@@ -0,0 +1,105 @@
...
+        // The outbox should only expose public/unlisted posts; enforce that
+        // here so a misbehaving remote can't slip restricted posts into the
+        // response.
+        const visibility = getVisibility(status.to, status.cc)
+        if (visibility !== 'public' && visibility !== 'unlisted') return false
+
+        if (status.type === StatusType.enum.Announce) {
+          if (excludeReblogs) return false
+          return knownAuthorIds.has(status.originalStatus.actorId)
+        }
```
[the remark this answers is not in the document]
llun: Fixed in b3f7736 — the visibility filter's Announce branch now guards with `Boolean(status.originalStatus?.actorId && …)` exactly as suggested, so a malformed announce is dropped instead of aborting the whole live fetch.

---

app/api/v1/accounts/[id]/statuses/route.ts
```diff
@@ -187,7 +205,36 @@ export const GET = traceApiRoute(
...
+          excludeReplies: parsedParams.exclude_replies === true,
+          excludeReblogs: parsedParams.exclude_reblogs === true,
+          onlyMedia: parsedParams.only_media === true
+        })
+      }
+      const servedLiveStatuses = liveRemoteStatuses.length > 0
+
+      const visibleStatuses = servedLiveStatuses
+        ? liveRemoteStatuses
+        : readableStatuses.slice(0, limit)
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts:261]: Fixed in b3f7736 with one refinement over the suggestion: the merge prefers the **local** copy on id collisions (rather than the live one) because the DB-backed status carries the viewer's interaction state (favourited/bookmarked/own-reblog), which the ephemeral outbox copy would reset to false. Merged, deduped by id, sorted by createdAt descending, sliced to the limit — covered by the new `merges live statuses with locally-stored ones, preferring the local copy on id collisions` route test.

---

lib/actions/utils.ts
```diff
@@ -58,46 +114,48 @@ export const recordActorIfNeeded = async ({
...
-      signingActor: await getResolvedSigningActor()
+      signingActor: resolvedSigningActor
     })
     if (!person) return
-    const iconUrl = getActorImageUrl(person.icon)
     const actor = await database.createActor({
       actorId,
-      type: person.type,
       username: person.preferredUsername,
       domain: new URL(person.id).hostname,
```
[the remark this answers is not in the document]
llun: Fixed in b3f7736 — `domain: new URL(person.id).host` now, matching `getActorDomain`'s existing `url.host` convention so port-bearing instances keep the port in the stored domain. (This line predates the PR — the refactor moved it — but the fix is right regardless.)

---

app/(timeline)/[actor]/getProfileData.ts
```diff
@@ -147,6 +147,29 @@ export const getProfileData = async (
...
+      .setActorCounters({
+        actorId: person.id,
+        followersCount: actorFollowersResponse.followerCount || null,
+        followingCount: actorFollowingResponse.followingCount || null,
+        statusCount: actorPostsResponse.statusesCount || null
```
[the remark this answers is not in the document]
Comment (llun) [reply app/(timeline)/[actor]/getProfileData.ts]: Addressed in b3f7736, but not with `?? null`: the premise doesn't hold here — `getActorFollowers`/`getActorFollowing`/`getActorPosts` return the **number 0** on fetch failure (`value?.totalItems ?? 0`), never `undefined`/`null`, so `?? null` would persist failure-zeros and clobber previously-synced counts whenever the remote briefly errors. Instead the page now uses `getActorCollectionCounts`, which genuinely distinguishes failure (`null`, preserves the stored counter) from an advertised `totalItems: 0` (persists as 0) for followers/following. The status count still comes from `getActorPosts` (already fetched for the page) which keeps the 0-is-ambiguous limitation — only positive values are trusted there, and the authoritative `recordActorIfNeeded` sync path persists genuine zeros on refresh.

---

lib/actions/utils.test.ts
```diff
@@ -6,11 +6,39 @@ import { getTestSQLDatabase } from '@/lib/database/testUtils'
...
+  following: `${actorId}/following`,
+  inbox: `${actorId}/inbox`,
+  outbox: `${actorId}/outbox`,
+  publicKey: {
+    id: `${actorId}#main-key`,
+    owner: actorId,
+    publicKeyPem: 'public-key'
+  },
+  endpoints: {
+    sharedInbox: `https://${new URL(actorId).hostname}/inbox`
```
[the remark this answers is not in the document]
llun: Applied in b3f7736 — the test helper now derives the mock sharedInbox from `new URL(actorId).origin` so it follows the actor id's protocol/port.

---

lib/actions/utils.ts
llun: Fixed in b3f7736 — `setActorCounters` now stamps a `remote-counts-synced:{actorId}` marker row (epoch seconds) on every sync, `hasActorCounters` checks only that marker, and the marker is cleaned up with the other counters on actor deletion. Legacy actors with locally-created counter rows now sync on first view; covered by updated `setActorCounters`/`hasActorCounters` DB tests and the `syncs profile and counters for a fresh actor whose counters were never synced` case.

---

lib/actions/utils.ts
llun: Fixed in b3f7736 — when the counter-only sync path can't fetch the person, `setActorCounters({ actorId })` now stamps the sync marker (values untouched) before returning the stored actor, so an unreachable actor retries on the 3-day stale-refresh cadence instead of on every inbox activity. Covered by the `returns the stored actor when a counter-only sync cannot fetch the person` test plus a follow-up assertion that the next call skips the network entirely.

---

lib/services/mastodon/getRemoteActorStatuses.ts
llun: Fixed in b3f7736 — announces are dropped when `onlyMedia` is set (`if (excludeReblogs || onlyMedia) return false`) and the `announceAuthorIds` lookup is skipped in both cases. Regression case `drops reblogs on media-only pages` added to the service tests.

---

lib/database/sql/actor.ts
llun: Fixed in b3f7736 — the `remove_from_followers` route no longer calls `updateActorFollowersCount`/`updateActorFollowingCount`; `updateFollowStatus`'s transactional decrement is the single writer for the Accepted→Undo transition, so a remote target's synced `totalFollowing` is decremented by one instead of being overwritten with a local-only recount.

---

app/api/v1/accounts/[id]/statuses/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts]: Fixed in b3f7736 — `decodeCursor` is now exported from `lib/services/timelines/request.ts` and reused here, and `safeIdToUrl` passes raw `http(s)://` values through unchanged (with a URL-validity check) so raw-URL cursors keep working; covered by a new `passes a raw status URL through unchanged` case in the urlToId tests. The same missing-decode pattern in `statuses/[id]/reblogged_by` (bare `idToUrl` mapping undecodable cursors to `''`) is pre-existing and untouched by this PR — noted for a follow-up.

---

app/(timeline)/[actor]/getProfileData.ts
Comment (llun) [reply app/(timeline)/[actor]/getProfileData.ts:135]: Fixed in b3f7736 — `getPersistableProfile` is exported and used by the profile page's `updateActor` call (both refresh paths now persist the same snapshot including `manuallyApprovesFollowers`, `fields`, and `type`), and the page fetches counts via `getActorCollectionCounts` (null-on-failure vs real-zero semantics) instead of `getActorFollowers`/`getActorFollowing`, dropping their unused first-page fetches. The status count keeps the positive-only trust rule since `getActorPosts` still conflates 0 with failure.

---

lib/activities/getActorCollectionCounts.ts
llun: Point 1 fixed in b3f7736 — the collection-root fetch is extracted as `fetchCollectionRoot` in `getActorCollections.ts` and both `getActorCollections` and `getActorCollectionCounts` consume it (error semantics stay at the callers: the full fetch keeps its span/exception handling, the counts helper keeps null-on-failure).

Point 2 (compaction) intentionally not changed: collection roots follow the existing `getActorCollections` precedent (uncompacted), the AGENTS.md rule's bolded scope enumerates notes/activities/actors, running `jsonld` per collection root ×3 on the inbox path would be a meaningful cost, and the failure mode is graceful (non-compact `totalItems` → null → counters keep their current value). Can revisit if a real-world implementation ships non-compact collection roots.

---

app/api/v1/accounts/[id]/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/route.ts:56]: Acknowledged as a deliberate trade-off — the synchronous refresh is the point of the change (first profile open must serve real counts), and the cost is bounded to authenticated views of actors >3 days stale on unreachable servers. The `isInternalActor` note was applied in b3f7736 (both this route and the statuses route now use `database.isInternalActor` instead of bare `privateKey` checks). Negative caching for failed refreshes (a short-lived "attempted" marker so dead instances don't add timeout latency per view) is left as a follow-up since it changes retry semantics for every `recordActorIfNeeded` caller.

---

lib/actions/utils.ts
llun: Acknowledged as a deliberate trade-off: the three root fetches run in parallel (≈1 extra RTT typical, bounded by the request timeout), the pattern matches Mastodon's own account-discovery behavior, and first-render correctness is the goal of the PR. If it proves costly in practice, the follow-up is to make the sync fire-and-forget (or a job) for non-user-facing callers (inbox jobs, block) while keeping it synchronous for lookup/search. Leaving as-is for this PR.

---

app/api/v1/accounts/[id]/statuses/route.ts
```diff
@@ -82,11 +84,29 @@ export const GET = traceApiRoute(
...
+        since_id: encodedSinceId
       } = parsedParams
 
+      // Clients echo back the opaque ids this endpoint emits (urlToId-encoded
+      // status URLs), so decode the cursors before querying — the database
+      // stores raw status URLs and silently ignores an unknown cursor, which
+      // would serve the same first page over and over. An undecodable cursor
+      // is a deliberate 400, using the same decodeCursor rule as the timeline
+      // endpoints.
+      const maxId = decodeCursor(encodedMaxId)
```
Comment (llun) [inline app/api/v1/accounts/[id]/statuses/route.ts]: **[sub-agent review round 2 — CONFIRMED]** Switching to `decodeCursor(encodedMaxId)` regressed blank cursors: `?max_id=` (empty param, which some clients emit for an unset cursor variable) now decodes `''` → `null` → 400, where it previously served the first page. The timeline endpoints avoid this by coercing `''` to `undefined` before decoding (`parseTimelineQuery`: "a blank cursor means 'no cursor'"). Fix: `decodeCursor(encodedMaxId || undefined)` for all three cursors, plus a regression test.

---

app/api/v1/accounts/[id]/statuses/route.ts
```diff
@@ -187,7 +207,51 @@ export const GET = traceApiRoute(
...
+      )
+      const visibleStatuses = servedLiveStatuses
+        ? [
+            ...readableStatuses,
+            ...liveRemoteStatuses.filter(
+              (status) => !localStatusIds.has(status.id)
+            )
+          ]
+            .sort((a, b) => b.createdAt - a.createdAt)
+            .slice(0, limit)
```
Comment (llun) [inline app/api/v1/accounts/[id]/statuses/route.ts]: **[sub-agent review round 2 — CONFIRMED]** The merged sort+`slice(0, limit)` can still push an *older* local followers-only post off the page when the outbox first page has ≥ limit newer public posts — and since live pages carry no Link headers, that post becomes unreachable, defeating the merge's stated purpose. Fix: keep every readable local status (the live-fetch gate guarantees there are fewer than `limit` of them) and top up only the remaining slots with the newest live statuses, then sort for display.

---

lib/actions/utils.ts
```diff
@@ -58,46 +115,55 @@ export const recordActorIfNeeded = async ({
...
+    signingActor: resolvedSigningActor
+  })
+  if (!person) {
+    if (isStale) return undefined
+    // A counter-only sync must not degrade the previous behavior of returning
+    // the stored actor when the remote fetch fails. Mark the counters synced
+    // (rows created, values preserved) so an unreachable actor doesn't
+    // re-trigger a blocking remote fetch on every subsequent call — the
+    // 3-day stale refresh remains the retry path.
+    await database.setActorCounters({ actorId })
```
llun: **[sub-agent review round 2 — two findings]**

1. **CONFIRMED:** the `await database.setActorCounters({ actorId })` on the failure path is unguarded, so a transient DB error (e.g. SQLITE_BUSY) makes `recordActorIfNeeded` reject on exactly the path whose contract is "return the stored actor when the remote fetch fails" — inbox job callers don't wrap this in `.catch`. Fix: make the marker write best-effort (try/catch + warn).
2. **PLAUSIBLE, accepted trade-off:** stamping the marker on *any* person-fetch failure freezes never-synced counters for up to the 3-day stale window even when the outage was momentary. The alternative (no stamp) is an unbounded per-activity fetch storm against dead instances — strictly worse for inbox throughput and fediverse politeness. The freeze only affects pre-deploy actors whose first post-deploy touch coincides with a remote outage, and heals early via any web profile view (which always fetches live and persists counts). Keeping the bounded behavior.

---

lib/types/database/operations.ts
```diff
@@ -81,6 +84,16 @@ export type GetLocalActorsParams = {
   // database layer to preserve the method's historical behavior.
   local?: boolean
 }
+export type SetActorCountersParams = {
+  actorId: string
+  // null/undefined preserves the locally-accumulated value for that counter
```
llun, line 89: **[sub-agent review round 2 — CONFIRMED]** This contract comment describes the removed round-1 semantics ("while still creating the row, so hasActorCounters treats the actor as synced") — after the marker-row redesign, null fields create no rows and the sync signal is the separate `remote-counts-synced:{actorId}` marker. A future implementer following this doc would rebuild the row-existence heuristic and reintroduce the local-accumulation false-positive. Fix: update the doc to describe the marker semantics.

---

app/api/v1/accounts/[id]/remove_from_followers/route.ts
```diff
@@ -45,14 +45,15 @@ export const POST = traceApiRoute(
       })
 
       if (follow && follow.status === FollowStatus.enum.Accepted) {
+        // updateFollowStatus decrements both actors' follow counters inside
+        // its transaction. Do NOT recount them from the local follows table
+        // here: for a remote target that recount would overwrite the
+        // remote-advertised totalFollowing synced from their server with a
+        // local-only value.
```
Comment (llun) [inline app/api/v1/accounts/[id]/remove_from_followers/route.ts:52]: **[sub-agent review round 2 — CONFIRMED]** Dropping these recount calls orphaned `updateActorFollowersCount`/`updateActorFollowingCount` entirely — zero callers remain, and this route's own comment documents why the recount-from-local-follows semantics is wrong for remote actors. Leaving the counter-clobbering helpers exported invites a future contributor to call them on a remote actor and reintroduce the bug. Fix: delete both methods from the Database interface and the SQL mixin.

---

llun (commented):
Sub-agent code-review round 2 (line-scan + cross-file tracer over the round-1 fix diff). CONFIRMED items are fixed in the next commit; each thread gets a reply and resolution as addressed.

---

app/api/v1/accounts/[id]/statuses/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts]: Fixed in e5340ea — all three cursors now decode via `decodeCursor(encoded || undefined)` so a blank param means "no cursor", with the `treats an empty cursor param as no cursor instead of rejecting it` regression test.

---

app/api/v1/accounts/[id]/statuses/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts]: Fixed in e5340ea — every readable local status is kept and only the remaining `limit - locals` slots are topped up with the newest live statuses before the display sort, so a link-less live page can never hide a local post. Covered by the new `keeps every local status on a live page instead of slicing it away` route test (limit=2, older local post + two newer live posts).

---

lib/actions/utils.ts
llun: Point 1 fixed in e5340ea — the marker write is wrapped in try/catch with a warn log, so a transient DB error degrades to the stored-actor return instead of rejecting. Point 2 stands as the accepted trade-off documented in the thread (bounded retries beat per-activity fetch storms; heals via the 3-day refresh or any web profile view).

---

lib/types/database/operations.ts
llun: Fixed in e5340ea — the `SetActorCountersParams` doc now describes the marker-row semantics: null preserves the local value, every call stamps `remote-counts-synced`, and counter-row existence explicitly can't carry the sync signal.

---

app/api/v1/accounts/[id]/remove_from_followers/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/remove_from_followers/route.ts:52]: Fixed in e5340ea — `updateActorFollowersCount`/`updateActorFollowingCount` are deleted from both the Database interface and the SQL mixin (zero callers remained).

---

llun:
/gemini review

---

app/api/v1/accounts/[id]/route.ts
```diff
@@ -29,10 +31,36 @@ export const GET = traceApiRoute(
...
+            (error) => {
+              logger.warn({
+                message: 'Failed to refresh remote actor for account view',
+                actorId: id,
+                error: error instanceof Error ? error.message : String(error)
+              })
+            }
+          )
+        }
+      }
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/accounts/[id]/route.ts:66]: Applied in 043408e — the refresh gate now checks the already-loaded row (`!persistedActor.account && !persistedActor.privateKey`) instead of an extra `isInternalActor` query. Equivalent semantics for this purpose: account-backed users have `account`, and the headless signer always carries a private key.

---

app/api/v1/accounts/[id]/statuses/route.ts
```diff
@@ -187,7 +207,53 @@ export const GET = traceApiRoute(
...
+      let liveRemoteStatuses: Status[] = []
+      const isFirstPage = !maxId && !minId && !sinceId
+      if (
+        currentActor &&
+        isFirstPage &&
+        readableStatuses.length < limit &&
+        parsedParams.pinned !== true &&
+        !parsedParams.tagged &&
+        !(await database.isInternalActor({ actorId: id }))
+      ) {
```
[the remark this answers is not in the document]
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts:230]: Applied in 043408e — the live-fetch gate reuses the `actor` row loaded at the top of the handler (`!actor.account && !actor.privateKey`) and drops the `isInternalActor` query.

---

lib/activities/getActorCollections.ts
```diff
@@ -35,6 +35,31 @@ export const isCollectionPageUrl = (pageUrl: string, collectionUrl: string) => {
...
+    url,
+    headers: activityPubRequestHeaders({ url, signingActor })
+  })
+  if (response.statusCode !== 200) {
+    return { statusCode: response.statusCode, collection: null }
+  }
+  return {
+    statusCode: response.statusCode,
+    collection: JSON.parse(response.body) as OrderedCollection
+  }
```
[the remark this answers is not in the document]
llun: Applied in 043408e — `fetchCollectionRoot` now catches a `JSON.parse` failure on a 200 response and returns a null collection (same handling as a non-200), so a garbage or empty body degrades gracefully for both `getActorCollections` and `getActorCollectionCounts`.

---

llun:
/gemini review

---

lib/actions/utils.ts
```diff
@@ -28,6 +35,56 @@ export const assertActorCanFederate = async ({
...
+    ...(person.summary ? { summary: person.summary } : {}),
+    ...(iconUrl ? { iconUrl } : {}),
+    ...(headerImageUrl ? { headerImageUrl } : {}),
+    // ActivityStreams treats an absent flag as "does not require approval".
+    manuallyApprovesFollowers: person.manuallyApprovesFollowers ?? false,
+    fields: getActorProfileFields(person),
+    followersUrl: person.followers ?? '',
+    inboxUrl: person.inbox,
+    sharedInboxUrl: person.endpoints?.sharedInbox ?? person.inbox,
+    publicKey: person.publicKey.publicKeyPem || ''
```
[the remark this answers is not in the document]
llun: No change needed: `person` here is never a raw payload — every caller gets it from `getActorPerson`, which runs the document through `compactActivityPub` and then `Actor.safeParse`, and the `APActor` Zod schema declares `publicKey` (with `publicKeyPem`) as **required** (`lib/types/activitypub/actor.ts`). An actor document without a public key fails validation and `getActorPerson` returns null before this line can run, so the optional chain would guard an impossible state. (The expression is also unchanged pre-existing behavior — this PR only moved it into `getPersistableProfile`.)

---

lib/activities/getActorCollectionCounts.ts
```diff
@@ -0,0 +1,67 @@
...
+  } catch (error) {
+    logger.warn({
+      message: 'Failed to fetch actor collection total items',
+      actorId: person.id,
+      field,
+      error: error instanceof Error ? error.message : String(error)
+    })
+    return null
+  }
+}
```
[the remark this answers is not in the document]
llun: No change needed: this helper only ever receives a `person` produced by `getActorPerson`, i.e. post-`compactActivityPub` (which canonicalizes id references to strings) and post-`Actor.safeParse` — and the schema types `followers`/`following` as `z.string().url().optional()` and `outbox` as `z.string().url()`. An inline collection object would fail actor validation upstream and never reach here, so the `as any` normalization would be dead code. Per the repo's ActivityPub conventions (AGENTS.md), wire-shape tolerance lives at the compaction/schema boundary, not scattered through typed consumers. If inline collection objects on actors ever need supporting, the right place is the `APActor` schema (accepting them and normalizing to the id string), not this consumer.

---

lib/actions/utils.ts
```diff
@@ -28,6 +35,56 @@ export const assertActorCanFederate = async ({
...
+  return {
+    type: person.type,
+    ...(person.name ? { name: person.name } : {}),
+    ...(person.summary ? { summary: person.summary } : {}),
+    ...(iconUrl ? { iconUrl } : {}),
+    ...(headerImageUrl ? { headerImageUrl } : {}),
+    // ActivityStreams treats an absent flag as "does not require approval".
+    manuallyApprovesFollowers: person.manuallyApprovesFollowers ?? false,
+    fields: getActorProfileFields(person),
+    followersUrl: person.followers ?? '',
```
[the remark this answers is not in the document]
llun: No change needed: `person.followers` is typed `string | undefined` after `compactActivityPub` + `Actor.safeParse` (the schema is `z.string().url().optional()`), so the object case can't occur here, and `[object Object]` can't be stored. The `?? ''` fallback is pre-existing behavior this PR only relocated, and switching to `|| null` would break the contract — `CreateActorParams.followersUrl` is a required `string`, and `updateActor` treats falsy as "no change", which is the intended handling for an actor that omits the field.

---

PR #1228, opened 2026-07-16
fix: refresh known remote actors on lookup and search account paths

llun:
## Summary

Follow-up to #1226. With that fix deployed, a remote user's posts show up in Mastodon apps, but the **profile header can still show the old zeroed stats** (0 followers / 0 following / wrong post count, stale display name and lock state). The tell in the report's screenshot: the post bylines show the fresh display name while the header shows the stale one — and the app is on the **search tab**.

Root cause: #1226 added the known-remote-actor refresh only to `GET /api/v1/accounts/:id`. Mastodon apps commonly build the profile header from the **search/lookup** response instead — and those paths only *record missing* actors; an actor that already exists locally is served straight from storage and never refreshed. Freshness was endpoint-dependent (this exact gap was flagged and deferred in #1226's review round 1).

### Changes

- **New `refreshKnownRemoteActor` service** (`lib/services/actors/refreshRemoteActor.ts`): wraps `recordActorIfNeeded` with the internal-actor guard (account-backed users and the headless signer never refresh) and best-effort error handling (a failed refresh falls back to the stored actor). `GET /api/v1/accounts/:id` now uses it instead of its inline block.
- **Refresh guardrails** (from review round 1): a per-actor **failure cooldown** (5 min) so an actor whose home server is down stops re-firing a blocking outbound fetch on every request; **single-flight dedup** so concurrent viewers of the same stale actor share one refresh; and a **5s wait budget** — a slow remote only delays the response briefly while the refresh finishes in the background and persists for the next request (stale-while-revalidate).
- **Shared `recordRemoteActorBestEffort` helper** (replaces v2 search's local wrapper): recording an unknown actor degrades to "no exact match" on failure instead of 500ing the whole request (federation-blocked domains, network errors).
- **`GET /api/v1/accounts/lookup`** refreshes a known remote actor before serialization for authenticated viewers (web session, or bearer with `read`/`read:accounts`). Auth is resolved lazily so lookups that never need a remote fetch skip the auth work, and internal actors skip even the auth check. A **presented bearer is validated up front** — an invalid/expired token gets `401` on every branch (matching the `OptionalOAuthGuard` routes) instead of silently serving stale data with the refresh skipped; credential-less lookups still serve stored data with no remote fetch.
- **`GET /api/v2/search`** (`resolveAccountId`): both the account-URL and handle branches now refresh a known remote actor instead of returning it untouched; unknown actors are webfinger-resolved and recorded as before. `resolve=true` already requires an authenticated viewer at the route boundary, and the refresh is signed by the instance actor like the rest of search resolution.
- **`GET /api/v1/accounts/search`**: the exact-handle resolution now runs before the index search — a known actor is refreshed (in parallel with the index search), an unknown one is webfinger-resolved best-effort — and the result is passed as `exactActorIds` in a single search call (previously: search, scan results for the exact acct, re-search on a miss). The route is always authenticated (`OAuthGuardAnyScope`).
- Docs: the divergence entry in `docs/mastodon-api-compatibility.md` lists every refreshing path, the guardrails, and lookup's strict bearer validation.

### Why this fixes the report

The reported instance's stored actor was verified healed (correct name/counts/lock) after an `accounts/:id` touch — but the app kept rendering the stale header because the search flow it used serves accounts without refreshing. With this change, every path a client uses to open a profile serves the same refreshed data, so the first search/lookup after deploy heals and returns the real stats.

### Verification

Simulated the production bug state in a local SQLite dev instance (existing `mttvll@mastodon.social` actor with a fresh `updatedAt`, no counter rows, no sync marker, cleared display name — the state a pre-#1226 actor is in), then:

- `GET /api/v1/accounts/lookup?acct=mttvll@mastodon.social` **with a session** → triggers the refresh against the real mastodon.social and returns `display_name: "Matteo"`, `locked: false`, `followers_count: 5370`, `following_count: 519`, `statuses_count: 641`, and persists the counters + sync marker.
- `GET /api/v2/search?q=mttvll@mastodon.social&resolve=true&type=accounts` **with a session** → same refreshed stats in the account result.
- The same lookup **anonymous** → serves stored data without any remote fetch.

Unit/route coverage: service tests for the refresh (internal-actor guard, fallback, cooldown, backoff expiry, single-flight, wait budget) and best-effort recording; lookup tests (session/bearer refresh, anonymous no-refresh, internal-actor skip, refresh-failure fallback, invalid-bearer 401 for stored actors); v1 accounts/search tests (resolve-before-search flow, known-handle refresh, bare-query and local-actor gating, record-failure fallback); v2 search tests (URL/handle-branch refresh signed by the instance actor, refresh-failure fallback, local-actor skip).

> **Branch note:** this branch was restarted from `main` after #1226 merged (same branch name, per workflow). The commit list shows #1226's pre-squash commits absorbed via an `-s ours` merge — they contribute **nothing** to the diff; the change vs `main` is exactly the files listed above.

## Checklist

- [x] If migrations changed: BOTH `migrations/schema.sql` and `migrations/schema.sqlite.sql` are regenerated in this PR (no migrations changed)

https://claude.ai/code/session_011NQyx4bZjX2GFGRRw5Kaht

### DIFF
--- a/app/api/v1/accounts/lookup/route.ts
+++ b/app/api/v1/accounts/lookup/route.ts
@@ -5,6 +5,7 @@ import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { localizeAccount } from '@/lib/services/accounts/localizeAccount'
+import { refreshKnownRemoteActor } from '@/lib/services/actors/refreshRemoteActor'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import {
   OptionalOAuthGuard,
@@ -91,19 +92,49 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
 
   const { username, domain } = handle
   let actor = await database.getActorFromUsername({ username, domain })
-  if (!actor && resolve && domain !== config.host) {
-    const hasBearerAuthorization = isBearerAuthorizationHeader(
-      req.headers.get('Authorization')
-    )
-    const session = hasBearerAuthorization ? null : await getServerAuthSession()
-    let canResolveRemote = Boolean(session?.user?.email)
-    if (!canResolveRemote && hasBearerAuthorization) {
-      const bearerAuth = await authorizeBearerRemoteLookup(req)
-      if (!bearerAuth.authorized) return bearerAuth.response
-      canResolveRemote = true
-    }
 
-    if (!canResolveRemote) {
+  // Remote fetches (resolving an unknown handle, or refreshing a known remote
+  // actor) require an authenticated viewer: the web session, or a bearer token
+  // carrying a read scope. Resolved lazily (and once) so anonymous lookups of
+  // stored actors stay free of auth work. An invalid bearer keeps its guard
+  // response so the resolve branch can surface it exactly as before.
+  const hasBearerAuthorization = isBearerAuthorizationHeader(
+    req.headers.get('Authorization')
+  )
+  type RemoteFetchAuth =
+    { authorized: true } | { authorized: false; bearerResponse?: Response }
+  let cachedRemoteFetchAuth: RemoteFetchAuth | null = null
+  const getRemoteFetchAuth = async (): Promise<RemoteFetchAuth> => {
+    if (cachedRemoteFetchAuth) return cachedRemoteFetchAuth
+    let auth: RemoteFetchAuth
+    if (!hasBearerAuthorization) {
+      const session = await getServerAuthSession()
+      auth = { authorized: Boolean(session?.user?.email) }
+    } else {
+      const bearerAuth = await authorizeBearerRemoteLookup(req)
+      auth = bearerAuth.authorized
+        ? { authorized: true }
+        : { authorized: false, bearerResponse: bearerAuth.response }
+    }
+    cachedRemoteFetchAuth = auth
+    return auth
+  }
+
+  if (actor) {
+    // Profile headers are commonly built from this endpoint's response, so a
+    // known remote actor is refreshed (stale profile + counter sync) before
+    // serialization — otherwise clients keep seeing zeroed counts until some
+    // other endpoint happens to refresh the actor. No-op for recently-synced
+    // actors; failures fall back to the stored actor. Internal actors skip
+    // even the auth check — nothing to refresh.
+    const isInternal = Boolean(actor.account || actor.privateKey)
+    if (!isInternal && (await getRemoteFetchAuth()).authorized) {
+      actor = await refreshKnownRemoteActor({ database, actor })
+    }
+  } else if (resolve && domain !== config.host) {
+    const auth = await getRemoteFetchAuth()
+    if (!auth.authorized) {
+      if (auth.bearerResponse) return auth.bearerResponse
       return apiResponse({
         req,
         allowedMethods: CORS_HEADERS,
--- a/app/api/v1/accounts/lookup/route.ts
+++ b/app/api/v1/accounts/lookup/route.ts
@@ -1,11 +1,13 @@
 import { NextRequest } from 'next/server'
 
-import { recordActorIfNeeded } from '@/lib/actions/utils'
 import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { localizeAccount } from '@/lib/services/accounts/localizeAccount'
-import { refreshKnownRemoteActor } from '@/lib/services/actors/refreshRemoteActor'
+import {
+  recordRemoteActorBestEffort,
+  refreshKnownRemoteActor
+} from '@/lib/services/actors/refreshRemoteActor'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import {
   OptionalOAuthGuard,
@@ -91,13 +93,11 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
     })
 
   const { username, domain } = handle
-  let actor = await database.getActorFromUsername({ username, domain })
 
   // Remote fetches (resolving an unknown handle, or refreshing a known remote
   // actor) require an authenticated viewer: the web session, or a bearer token
-  // carrying a read scope. Resolved lazily (and once) so anonymous lookups of
-  // stored actors stay free of auth work. An invalid bearer keeps its guard
-  // response so the resolve branch can surface it exactly as before.
+  // carrying a read scope. Resolved lazily (and once) so lookups that never
+  // need a remote fetch skip the auth work.
   const hasBearerAuthorization = isBearerAuthorizationHeader(
     req.headers.get('Authorization')
   )
@@ -120,6 +120,17 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
     return auth
   }
 
+  // A presented bearer is validated up front, matching the guarded routes
+  // (OptionalOAuthGuard rejects an invalid token instead of downgrading it to
+  // anonymous): an expired-token client gets a 401 re-auth signal here rather
+  // than a silently-stale profile with the refresh skipped.
+  if (hasBearerAuthorization) {
+    const auth = await getRemoteFetchAuth()
+    if (!auth.authorized && auth.bearerResponse) return auth.bearerResponse
+  }
+
+  let actor = await database.getActorFromUsername({ username, domain })
+
   if (actor) {
     // Profile headers are commonly built from this endpoint's response, so a
     // known remote actor is refreshed (stale profile + counter sync) before
@@ -132,9 +143,10 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
       actor = await refreshKnownRemoteActor({ database, actor })
     }
   } else if (resolve && domain !== config.host) {
-    const auth = await getRemoteFetchAuth()
-    if (!auth.authorized) {
-      if (auth.bearerResponse) return auth.bearerResponse
+    // An invalid bearer already returned above, so an unauthorized viewer
+    // here is credential-less and gets the same 404 an unresolvable handle
+    // would.
+    if (!(await getRemoteFetchAuth()).authorized) {
       return apiResponse({
         req,
         allowedMethods: CORS_HEADERS,
@@ -145,7 +157,7 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
 
     const actorId = await getWebfingerSelf({ account: `${username}@${domain}` })
     actor = actorId
-      ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
+      ? await recordRemoteActorBestEffort({ actorId, database })
       : null
   }
 
--- a/app/api/v1/accounts/search/route.test.ts
+++ b/app/api/v1/accounts/search/route.test.ts
@@ -1,5 +1,7 @@
 import { NextRequest } from 'next/server'
 
+import { resetRefreshRemoteActorStateForTesting } from '@/lib/services/actors/refreshRemoteActor'
+
 import { GET } from './route'
 
 const mockSearchAccountIds = vi.fn()
@@ -69,6 +71,7 @@ const context = { params: Promise.resolve({}) }
 describe('GET /api/v1/accounts/search', () => {
   beforeEach(() => {
     vi.clearAllMocks()
+    resetRefreshRemoteActorStateForTesting()
     mockGetServerSession.mockResolvedValue(null)
     mockStoredToken.mockResolvedValue({
       expiresAt: new Date(Date.now() + 60_000),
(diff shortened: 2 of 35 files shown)


lib/services/actors/refreshRemoteActor.ts
```diff
@@ -0,0 +1,41 @@
...
+}: {
+  database: Database
+  actor: Actor
+  signingActor?: Actor
+}): Promise<Actor> => {
+  if (actor.account || actor.privateKey) return actor
+
+  try {
+    return (
+      (await recordActorIfNeeded({
```
llun: **[review-loop finding — major, perf/security]** No negative caching or request dedup around the refresh. `recordActorIfNeeded`'s *stale*-path failure (`lib/actions/utils.ts`) returns `undefined` without stamping the sync marker or bumping `updatedAt`, so for an actor whose home instance is down **every** authenticated lookup/search re-fires the blocking outbound fetch (up to 2 attempts × 10s timeout each). The check is also read-then-fetch with no claim, so N concurrent viewers of the same stale actor fire N independent refreshes (up to 4 outbound requests each). Pre-PR this mechanism had one caller (`accounts/:id`); this PR wires it into four hot account-serving routes, which materially amplifies both problems — a `read`-scoped client hammering `lookup?acct=x@dead-instance` holds open outbound sockets with zero backoff.

Fix (this service, so all four routes inherit it): a per-actor failure cooldown (5 min) so a failed refresh backs off instead of retrying per request, and single-flight dedup so concurrent requests share one in-flight refresh.

---

app/api/v1/accounts/lookup/route.ts
```diff
@@ -91,19 +92,49 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
...
-    if (!canResolveRemote) {
+  if (actor) {
+    // Profile headers are commonly built from this endpoint's response, so a
+    // known remote actor is refreshed (stale profile + counter sync) before
+    // serialization — otherwise clients keep seeing zeroed counts until some
+    // other endpoint happens to refresh the actor. No-op for recently-synced
+    // actors; failures fall back to the stored actor. Internal actors skip
+    // even the auth check — nothing to refresh.
+    const isInternal = Boolean(actor.account || actor.privateKey)
+    if (!isInternal && (await getRemoteFetchAuth()).authorized) {
```
llun, line 142: **[review-loop finding — major, latency]** This refresh is synchronous on the hottest branch of the endpoint with no overall time budget: opening a profile of a stale actor on a slow/blackholed instance blocks the lookup response for up to ~20s (person fetch retries) + the collection-count fetches — pre-PR this path served straight from the DB, and Mastodon apps will trip their client timeouts and render the profile as broken. `resolve=false` lookups are also affected, which deviates from Mastodon's serve-from-storage lookup semantics more than intended.

Fix: give `refreshKnownRemoteActor` a wait budget (5s) — if the refresh hasn't finished by then, serve the stored actor immediately while the refresh completes in the background and persists for the next request (stale-while-revalidate). Combined with the failure cooldown this bounds the worst case for all four routes.

---

app/api/v1/accounts/lookup/route.ts
```diff
@@ -91,19 +92,49 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
...
-      req.headers.get('Authorization')
-    )
-    const session = hasBearerAuthorization ? null : await getServerAuthSession()
-    let canResolveRemote = Boolean(session?.user?.email)
-    if (!canResolveRemote && hasBearerAuthorization) {
+
+  // Remote fetches (resolving an unknown handle, or refreshing a known remote
+  // actor) require an authenticated viewer: the web session, or a bearer token
+  // carrying a read scope. Resolved lazily (and once) so anonymous lookups of
+  // stored actors stay free of auth work. An invalid bearer keeps its guard
```
llun: **[review-loop finding — minor, stale comment]** "anonymous lookups of stored actors stay free of auth work" is only true for internal actors: for any stored *remote* actor the `if (actor)` branch calls `getRemoteFetchAuth()` unconditionally, which for cookie-less anonymous requests means a session resolution per request on a path that previously did none. Small overhead, but the comment overstates the guarantee — reword it to match the actual behavior.

---

app/api/v1/accounts/search/route.ts
```diff
@@ -80,36 +81,41 @@ export const GET = traceApiRoute(
...
+          exactActor = await refreshKnownRemoteActor({
+            database,
+            actor: exactActor
+          })
+        } else {
+          const actorId = await getWebfingerSelf({
+            account: `${handle.username}@${handle.domain}`
           })
+          exactActor = actorId
+            ? ((await recordActorIfNeeded({ actorId, database })) ?? null)
```
llun: **[review-loop finding — minor, API compat]** Two issues in the unknown-handle branch of the restructure:

1. **Blocked-domain 500 (new):** exact-handle detection moved from the old case-insensitive scan of search results to `getActorFromUsername`, which is case-sensitive on username — so a case-mismatched query for a *stored* actor (e.g. `gargron@…` stored as `Gargron`) now takes this webfinger+record path, and if the domain is federation-blocked, `recordActorIfNeeded` throws `BlockedFederationDomainError` with no catch → the whole search 500s where pre-PR it returned 200 with results. Any other `recordActorIfNeeded` rejection (network, DB) 500s the same way. Fix: wrap the recording best-effort (log + degrade to no exact match) — same pattern v2 search already uses.
2. **Serialization ordering (perf):** the known-actor refresh is awaited before `searchAccountIds` even though `exactActor.id` is already in hand — only the final `getMastodonActorsFromIds` read needs the refreshed row. Run the refresh alongside the index search (v2 search already runs `resolveAccountId` in `Promise.all` with its index search).

Residual (accepted): the case-mismatch path costs one webfinger round-trip that the old flow avoided; it's output-equivalent (webfinger returns the canonical id, which refreshes the stored actor) and only fires on `resolve=true` exact-handle queries.

---

app/api/v2/search/route.test.ts
```diff
@@ -873,7 +873,7 @@ describe('GET /api/v2/search', () => {
     ])
   })
 
-  it('uses existing canonical actors for Mastodon account profile URLs', async () => {
+  it('refreshes existing canonical actors for Mastodon account profile URLs instead of re-recording them', async () => {
```
llun, line 879: **[review-loop finding — minor, test coverage]** Three gaps in the v2 search refresh coverage:

1. This renamed test can't actually distinguish the new refresh path from the old re-record path — both funnel into `recordActorIfNeeded` with identical arguments, so its assertions would still pass if the `persistedActor ? refresh : record` branch were deleted. A discriminating test: make `recordActorIfNeeded` **reject** for a stored actor and assert the account is *still returned* (refresh falls back to the stored actor; the old record path returned null and dropped the account).
2. The URL branch for a *directly* stored actor id (query is the actor id itself, no canonicalization) is traversed by pre-existing tests but never asserted — dropping the refresh or the `signingActor` there breaks nothing.
3. No route-level test that a stored **local** actor resolved by handle skips `recordActorIfNeeded` — the search routes rely solely on the service's unit-tested internal-actor guard.

---

app/api/v1/accounts/search/route.test.ts
```diff
@@ -110,16 +110,14 @@ describe('GET /api/v1/accounts/search', () => {
     })
   })
 
-  it('resolves a remote handle only after indexed search misses', async () => {
+  it('resolves an unknown remote handle via webfinger before the indexed search', async () => {
```
llun, line 116: **[review-loop finding — minor, test coverage]** Two gaps in the v1 accounts/search coverage of the new resolve-before-search flow:

1. No test for `resolve=true` with a bare non-handle query (e.g. `q=runner`): `parseAccountHandle` returns null so both the `getActorFromUsername` lookup and webfinger must be skipped, but nothing pins that gating (existing non-handle tests either omit `resolve` or use `offset=1`).
2. No route-level test that a stored **local** exact-handle match skips the refresh (`recordActorIfNeeded` not called) while still landing in `exactActorIds` — a regression that bypassed `refreshKnownRemoteActor`'s internal-actor guard would fire network fetches for local actors unnoticed.

---

llun (commented):
Sub-agent review loop, round 1: six reviewers (correctness, security, perf/concurrency, tests, Mastodon API compat, repo conventions) over the full diff. Correctness and conventions came back clean; the findings that survived verification are posted inline. The dominant theme — the refresh has no negative caching, no dedup, and no time budget, which four hot routes now amplify — is being addressed centrally in `refreshKnownRemoteActor` (failure cooldown + single-flight + wait budget) together with Gemini's two findings; fix commit incoming.

---

app/api/v1/accounts/lookup/route.ts
```diff
@@ -91,19 +92,49 @@ export const GET = traceApiRoute('lookupAccount', async (req: NextRequest) => {
...
       const bearerAuth = await authorizeBearerRemoteLookup(req)
-      if (!bearerAuth.authorized) return bearerAuth.response
-      canResolveRemote = true
+      auth = bearerAuth.authorized
+        ? { authorized: true }
+        : { authorized: false, bearerResponse: bearerAuth.response }
     }
+    cachedRemoteFetchAuth = auth
+    return auth
+  }
```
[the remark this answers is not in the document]
llun: Implemented in 37dd106 — a presented bearer is now validated up front and an invalid one gets its guard `401` on every branch, with a test (`rejects an invalid bearer token even when the actor is stored locally`) pinning that the token is checked before any lookup work.

One framing correction for the record: this wasn't a new leak or bypass — the pushed code's status codes were byte-identical to pre-PR `main` (auth was never consulted for stored actors), and cache-state probing needs no token at all on this endpoint (`200` vs `404` anonymously is Mastodon's documented public-lookup behavior, so rejecting invalid bearers doesn't remove that signal). The reasons to accept anyway: every `OptionalOAuthGuard` route in this codebase already rejects invalid bearers rather than downgrading them to anonymous, so lookup was the odd one out — and more importantly, an expired-token client would have silently received stale data with the refresh skipped forever, which is the exact bug this PR fixes. The 401 gives it a re-auth signal.

---

app/api/v1/accounts/search/route.ts
```diff
@@ -80,36 +81,41 @@ export const GET = traceApiRoute(
...
+            : null
```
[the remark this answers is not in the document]
llun: Implemented in 37dd106, with two adjustments to the suggestion: the wrapper is a shared `recordRemoteActorBestEffort` in `lib/services/actors/refreshRemoteActor.ts` (it also replaces v2 search's local `recordResolvedActorIfNeeded` and guards the lookup resolve branch, so all three record sites degrade the same way), and it logs via the project `logger`, not `console.warn`. Regression test added: `falls back to the indexed results when recording a webfinger-resolved actor fails`. This also closes a real client-visible case the review loop found independently: a federation-blocked domain makes `recordActorIfNeeded` throw, which previously 500'd the whole search.

---

lib/services/actors/refreshRemoteActor.ts
llun: Fixed in 37dd106. `refreshKnownRemoteActor` now keeps module-level per-actor state: a failed refresh (rejection **or** `recordActorIfNeeded` returning `undefined`, which is how the stale-path fetch failure surfaces) records a 5-minute cooldown during which further calls serve the stored actor without any outbound work; a success clears the cooldown; concurrent callers share one in-flight refresh promise (single-flight), so N simultaneous viewers cost one fetch. The failure map sweeps expired entries opportunistically so it can't grow unbounded across many dead remotes. Covered by `backs off after a failed refresh instead of retrying every request`, `treats an empty refresh result as a failure for the backoff`, `retries a failed refresh once the cooldown expires`, `clears the failure backoff after a successful refresh`, and `shares a single in-flight refresh across concurrent requests`. Suites that exercise the refresh reset this state in `beforeEach` via `resetRefreshRemoteActorStateForTesting`.

---

app/api/v1/accounts/lookup/route.ts
llun: Fixed in 37dd106. `refreshKnownRemoteActor` races the refresh against a 5s wait budget: past the budget the stored actor is served immediately while the refresh keeps running, and its result persists for the next request (stale-while-revalidate). With the failure cooldown and single-flight this bounds every route's worst case to ~5s once per actor per cooldown window, instead of ~20s on every request. Covered by `serves the stored actor when the refresh exceeds the wait budget` (fake timers). Keeping the refresh on `resolve=false` lookups is intentional — serving current stats on the path apps actually use to open profiles is the point of the PR — the budget just caps what that can cost.

---

app/api/v1/accounts/lookup/route.ts
llun: Fixed in 37dd106 — the comment now says lazy auth means "lookups that never need a remote fetch skip the auth work," which is the actual guarantee (internal actors skip it entirely; stored remote actors do resolve the session lazily). The early bearer validation added for the other thread doesn't change the anonymous path: no Authorization header means no up-front work.

---

app/api/v1/accounts/search/route.ts
llun: Both fixed in 37dd106. (1) The webfinger-resolved recording goes through the shared `recordRemoteActorBestEffort` (also adopted by v2 search and the lookup resolve branch), so a blocked domain or network failure degrades to no exact match instead of a 500 — test: `falls back to the indexed results when recording a webfinger-resolved actor fails`. (2) The known-actor refresh now runs in `Promise.all` with `searchAccountIds`, since only the final serialization read needs the refreshed row. The case-mismatch webfinger round-trip remains as the accepted residual.

---

app/api/v2/search/route.test.ts
llun: All three added in 37dd106: `still returns a stored account when its refresh fails during URL resolution` (discriminating test — covers the directly-stored actor-id branch too, asserting the refresh is signed by the instance actor and its failure falls back to the stored account instead of dropping it) and `does not refresh stored local actors resolved by handle` (route-level internal-actor gating).

---

app/api/v1/accounts/search/route.test.ts
llun: Both added in 37dd106: `skips exact-handle resolution for bare non-handle queries` (asserts neither `getActorFromUsername` nor webfinger fire for `resolve=true&q=runner`) and `does not refresh a stored local actor but still pins it as the exact match` (no `recordActorIfNeeded`, `exactActorIds` carries the local actor's id).

---

llun:
/gemini review

---

lib/services/actors/refreshRemoteActor.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const recordRefreshFailure = (actorId: string) => {
+  if (failedRefreshesAt.size >= FAILURE_SWEEP_SIZE) {
+    const cutoff = Date.now() - FAILURE_COOLDOWN_MS
+    for (const [id, failedAt] of failedRefreshesAt) {
+      if (failedAt < cutoff) failedRefreshesAt.delete(id)
+    }
+  }
+  failedRefreshesAt.set(actorId, Date.now())
+}
```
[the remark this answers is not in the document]
llun: Implemented in e144fde — the sweep now runs at most once per minute (`FAILURE_SWEEP_INTERVAL_MS`), and `resetRefreshRemoteActorStateForTesting` clears the throttle timestamp for test isolation, as suggested.

---

lib/services/actors/refreshRemoteActor.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    vi.advanceTimersByTime(FAILURE_COOLDOWN_MS + 1)
+    await refreshKnownRemoteActor({
+      database: mockDatabase,
+      actor: remoteActor
+    })
+
+    expect(recordActorIfNeeded).toHaveBeenCalledTimes(2)
+  })
+
+  it('clears the failure backoff after a successful refresh', async () => {
```
llun: **[review-loop finding — minor, vacuous test name]** This test passes even if the success-path `failedRefreshesAt.delete(actorId)` is removed: by construction (single-flight + the cooldown gate) a refresh can only succeed after the failure entry has already expired, at which point the entry never gates anything again — the delete is pure map hygiene with no behavior observable through the public API. What the test *does* usefully pin is that a **settled** refresh promise is removed from the in-flight map (the third call only re-invokes `recordActorIfNeeded` because `.finally` cleaned it up), so it should be renamed to claim that, and the service's delete-on-success line deserves a comment saying it's hygiene, not behavior.

---

llun (commented):
Sub-agent review loop, round 2 (over the guardrails delta 37dd106+e144fde): three reviewers — service concurrency, route integration, conventions/docs. Route integration and conventions/docs came back clean (memoized single token validation, CORS'd 401, Promise.all ordering, docs claims all verified). Service concurrency found no defects in the guardrail logic itself — races, timer leaks, unhandled rejections, sweep correctness, fake-timer validity, and cross-route module sharing all check out — with one minor test-naming finding posted inline. Fix incoming.

---

lib/services/actors/refreshRemoteActor.test.ts
llun: Fixed in d2d2f12 — renamed to `starts a new refresh for the next request once a completed one settles` with an assertion comment explaining the third call pins in-flight map cleanup, and the service's delete-on-success now carries a "hygiene, not behavior" comment.

---

llun:
/gemini review

---

llun:
Sub-agent review loop, round 3 (final delta: the sweep throttle and the test rename): **clean, no findings**. Throttle verified (first sweep fires immediately, never deletes fresh entries, no interaction with cooldown/single-flight, test reset clears all module state), renamed test verified to pin exactly what it asserts, docs unaffected. All 80 tests across the five touched suites pass. Loop summary: round 1 — six reviewers, six findings (all fixed in 37dd106); round 2 — three reviewers, one finding (fixed in d2d2f12); round 3 — clean. All review threads resolved.

---

PR #1236, opened 2026-07-17
minor: quote data model + inbound quote federation (Epic 4.1a)

llun:
## Epic 4.1 — Quote posts (Mastodon 4.5), PR 4.1a of 4

First PR of the quote-posts epic. After this, quotes arriving from Mastodon / Misskey / Fedibird render correctly in every timeline for API clients. **No local quoting yet** — that's 4.1b.

Design is per the Phase 4 plan's binding decisions: `status_quotes` satellite table (5 persisted states; viewer-relative states computed at serialization), `quoteApprovalPolicy` in the per-status content JSON (no column), full FEP-044f vocabulary aliased in `CANONICAL_CONTEXT`.

### What's in this PR
- **Migration + both dumps.** `status_quotes` (PK on `statusId`, no FK — the edge can be created before the quoting status row exists, matching likes/recipients). Regenerated `migrations/schema.sql` (PG) and `migrations/schema.sqlite.sql` (SQLite) from a full migration run of each backend.
- **JSON-LD compaction aliases** (the load-bearing step). `quote`, `quoteUrl`, `quoteUri`, `_misskey_quote`, `quoteAuthorization`, `QuoteRequest`/`QuoteAuthorization` types, and `interactionPolicy.canQuote.{automaticApproval,manualApproval}` — with survives-compaction regression tests, including the single-element array-collapse guard (#1119/#1120 class) and a sender-defined-IRI variant (not the blank-node accident).
- **Liberal AP schema fields** on `BaseContent` + `getQuoteTargetId` (precedence `quote` → `quoteUrl` → `quoteUri` → `_misskey_quote`).
- **Domain model.** `QuoteState`, `QuoteApprovalPolicy`, `StatusQuote`; `quote` edge + `quoteApprovalPolicy` on `StatusNote` (`StatusPoll` inherits; `StatusAnnounce` correctly excluded).
- **DB module** `statusQuote.ts` — `createStatusQuote` (upsert), `getStatusQuote`, `updateStatusQuoteState` (one-way state machine), `getQuotingStatusIds` (keyset pagination). Plus status hydration: batch quote-edge reads in `getStatusesByIds` and `quoteApprovalPolicy` read from the content blob.
- **Inbound ingest.** `verifyRemoteQuote` implements the FEP receiver rules — same-author shortcut, instance-signed stamp fetch + three-field match, degrade to `pending` on any failure — wired into `createNoteJob`.
- **Mastodon entities + serializer.** `Quote` / `ShallowQuote` / `QuoteApproval` on `BaseStatus`; depth-limited serialization (full at depth 0, shallow at depth ≥ 1), viewer-relative `deleted`/`unauthorized` downgrades, and a batched `quotedStatusCache` (no N+1).

### Deviations from the plan (mechanical drift; code at HEAD wins)
- The plan's Decision-3 context block put `@type: @id` on the legacy compat aliases (`quoteUrl`/`quoteUri`/`_misskey_quote`), but its own RED test emits those as plain URL-string literals — inconsistent. Fedibird/Misskey define these terms **without** `@type: @id`, so an `@type: @id` mapping drops the literal during compaction. Reconciled: compat aliases are plain term→IRI; `@type: @id` is kept only on the FEP-044f id references (`quote`, `quoteAuthorization`). Verified by the regression test.
- Migration uses `table.timestamp(col, { useTz: true })` (the current repo convention → `timestamp with time zone` in PG, `datetime` in SQLite) instead of the plan sketch's `table.datetime`, matching the existing satellite tables.
- `quote_approval.current_user` for the `followers` policy returns `unknown` until `canQuoteStatus` lands in 4.1b/4.1c (the plan shares that service for the accurate follower check).

### Test evidence
Local gate green on Node 24: `prettier --write .`, `yarn lint` (0 errors), `yarn build`, `yarn test` (664 files, 6307 tests passing). New coverage: compaction alias regressions, `getQuoteTargetId`, the `statusQuote` module (upsert, full transition matrix, pagination, hydration), `verifyRemoteQuote` (all FEP branches incl. stamp field mismatch), `createNoteJob` quote ingest (self-quote/Misskey/Fedibird/duplicate), and the serializer state × viewer matrix + shallow-recursion + no-N+1 batch.

### DIFF
--- a/lib/activities/jsonld/index.test.ts
+++ b/lib/activities/jsonld/index.test.ts
@@ -196,6 +196,106 @@ describe('compactActivityPub', () => {
     expect(tags[0].name).toBe('#fediverse')
   })
 
+  it('keeps quote extension terms when the sender defines them with full IRIs', async () => {
+    // Mastodon 4.5 / Fedibird / Misskey each carry the quote target under their
+    // own vocabulary and *define* those terms in the outbound @context. Without
+    // matching aliases in CANONICAL_CONTEXT they compact to CURIEs/full IRIs and
+    // vanish from the parsed note.
+    const result = asRecord(
+      await compactActivityPub({
+        '@context': [
+          ACTIVITY_STREAMS_CONTEXT_URL,
+          {
+            quote: { '@id': 'https://w3id.org/fep/044f#quote', '@type': '@id' },
+            quoteUri: 'http://fedibird.com/ns#quoteUri',
+            _misskey_quote: 'https://misskey-hub.net/ns#_misskey_quote',
+            quoteAuthorization: {
+              '@id': 'https://w3id.org/fep/044f#quoteAuthorization',
+              '@type': '@id'
+            }
+          }
+        ],
+        id: 'https://remote.example/notes/2',
+        type: 'Note',
+        attributedTo: 'https://remote.example/users/alice',
+        published: '2026-01-01T00:00:00Z',
+        quote: 'https://llun.test/users/me/statuses/1',
+        quoteUri: 'https://llun.test/users/me/statuses/1',
+        _misskey_quote: 'https://llun.test/users/me/statuses/1',
+        quoteAuthorization: 'https://llun.test/users/me/quote_authorizations/9'
+      })
+    )
+
+    expect(result.quote).toBe('https://llun.test/users/me/statuses/1')
+    expect(result.quoteUri).toBe('https://llun.test/users/me/statuses/1')
+    expect(result._misskey_quote).toBe('https://llun.test/users/me/statuses/1')
+    expect(result.quoteAuthorization).toBe(
+      'https://llun.test/users/me/quote_authorizations/9'
+    )
+  })
+
+  // The blank-node accident (`_:QuoteRequest`) only recovers a bare term when
+  // the sender does NOT define the type. Mastodon DOES define these terms, so
+  // assert the sender-defined-IRI path survives via a real alias.
+  it.each([
+    { type: 'QuoteRequest', iri: 'https://w3id.org/fep/044f#QuoteRequest' },
+    {
+      type: 'QuoteAuthorization',
+      iri: 'https://w3id.org/fep/044f#QuoteAuthorization'
+    }
+  ])(
+    'keeps a $type type as a bare term when the sender defines it with a full IRI',
+    async ({ type, iri }) => {
+      const result = asRecord(
+        await compactActivityPub({
+          '@context': [ACTIVITY_STREAMS_CONTEXT_URL, { [type]: iri }],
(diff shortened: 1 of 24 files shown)


llun:
### Sub-agent code-review — round 1

Fanned out 5 dimension reviewers (correctness, security, invariants, tests, JSON-LD/schema) over the diff; each finding was adversarially verified by an independent skeptic. Result: **7 confirmed** (1 high, 6 low), the rest false positives (verified as correct: the `as:quoteUrl`/`manualApproval` aliases match the real Fedibird/Misskey conventions and are load-bearing; the no-N+1 spy test lives in `getMastodonStatus.test.ts`).

Addressed in `0c04d4ed`:

- **[high · security] Forged quote authorization.** `verifyRemoteQuote` validated the stamp only on self-asserted fields, so a quoter could serve a fake `QuoteAuthorization` (naming the victim in `attributedTo`) from their own host and get the quote marked `accepted`. Now the stamp must be hosted under the quoted author's authority — both the fetched URL and the stamp's declared `id` must share the author's host — in addition to the three-field match.
- **[low · consistency] Shallow quote id exposure.** A depth-≥1 `ShallowQuote` now withholds `quoted_status_id` (null) for non-accepted edges, matching the depth-0 contract.
- **[low · tests] Coverage gaps closed:** foreign-authority/forged-stamp + invalid-body rejection; the full 25-pair state-transition matrix; `getQuotingStatusIds` `sinceId` / unfiltered / missing-cursor cases; authenticated-viewer quote serialization (single + batch); the public-policy authenticated-non-author `quote_approval` row.

Gate re-run green: prettier, lint (0 errors), build, test (664 files, 6328 passing).

llun:
### Sub-agent code-review — converged (rounds 2–4)

- **Round 2** (from round-1's full result): applied the deleted/unauthorized viewer downgrade identically at depth 0 and depth ≥ 1 (a nested/shallow quote no longer leaks a deleted/unreadable status id); corrected the QuoteState docs so they don't claim the block/mute-relative states are computed yet.
- **Round 3** (convergence): batched quote-edge hydration in `getActorStatuses`/`getStatusReplies` (removing a per-status N+1); `getQuotedStatus` now passes `currentActorId` on the single-status path so an embedded quote reflects the viewer's favourited/bookmarked/reblogged; `BaseContent.quote` tolerates an id-less object instead of dropping the whole note (liberal-inbound mandate). Deferred to 4.1b (documented): persisting a remote status's inbound `interactionPolicy` — policy semantics belong to the outbound epic and are inert here.
- **Round 4** (final): clean — zero findings.

gemini is sunset and cubic/kilo did not post, so the sub-agent fan-out was the review mechanism. All CI checks green on `5a314de`. Enabling squash auto-merge.

PR #1239, opened 2026-07-17
minor: outbound quoting + QuoteRequest/Accept/Reject handshake (Epic 4.1b)

llun:
## Epic 4.1 — Quote posts (Mastodon 4.5), PR 4.1b of 4

Builds on 4.1a (merged). Local users can now **quote** statuses, and the full FEP-044f handshake + revocation federate. Base: current `main` (includes 4.1a).

### What's in this PR
- **`POST /api/v1/statuses`**: `quoted_status_id` + `quote_approval_policy` (JSON and form-encoded). The route resolves `idToUrl`, returns **404** when the target is invisible to the caller, and **422** when the quoted author's policy denies the quote.
- **`canQuoteStatus`**: pure verdict (`automatic | denied`) from `quoteApprovalPolicy` + block relations + accepted-follow — shared by the create route and the inbound QuoteRequest handler.
- **Create path** (`createNoteFromUserInput`): persists the policy into the blob; a **local** quoted status → `accepted` edge + a `quote` notification (same-authority shortcut); a **remote** quoted status → `pending` edge + a `SendQuoteRequest`.
- **`quote` notification type**: wired through the Mastodon serializer, push payloads, the notifications UI config, and the email/push preference maps.
- **Jobs + senders**: `SendQuoteRequest` / `SendQuoteAccept` / `SendQuoteReject`, plus a shared-inbox `HandleQuoteRequest` fallback.
- **Inbound handshake** (per-user inbox): the `QuoteRequest` type (previously swallowed by `GracefullyAcceptedActivity`) is now handled — evaluate `canQuoteStatus`, then enqueue `Accept` (+ a hosted stamp) or `Reject`. Inbound `Accept`/`Reject` are matched against our outbound QuoteRequests (by `quoteRequestId`) and, on Accept, re-federate the note carrying `quoteAuthorization`; the follow handshake path is untouched (falls through on no match). Shared-inbox `getJobMessage` routes `QuoteRequest`.
- **Stamp hosting**: `GET /users/:name/quote_authorizations/:id` serves the `QuoteAuthorization` while the edge is `accepted`, **404** once revoked.
- **Revocation**: a `Delete` of a stamp (host-authority verified) marks the edge `revoked`; local status deletion sweeps edges quoting it to `deleted` and removes the deleted status's own edge.
- **Outbound emission**: `getNoteFromStatus` writes `quote`/`quoteUrl`/`quoteUri`/`_misskey_quote`, `quoteAuthorization` (when accepted), and `interactionPolicy.canQuote.{automaticApproval,manualApproval}`, under a shared quote `@context`.
- **Race fix**: `createNoteJob`'s edge upsert is now non-regressing (one-way state machine), so an `accepted` edge from our Accept is never downgraded by a later stampless `Create Note`.

New DB operations (no migration — reuse `status_quotes`): `getStatusQuoteByQuoteRequestId`, `getStatusQuoteByAuthorizationUri`, `markQuotesDeletedByQuotedStatusId`.

### Notes / deviations
- Docs (`docs/mastodon-api-compatibility.md`, `docs/features.md`) are updated in the landing PR (4.1c), per the plan.
- Persisting a *remote* status's inbound `interactionPolicy` as `quoteApprovalPolicy` is deferred (policy is enforced for local statuses only; a remote status's advertised policy is inert here). Noted for 4.1c.

### Test evidence
Local gate green on Node 24: prettier, lint (0 errors), build, test (**668 files, 6382 tests**). New coverage: `canQuoteStatus` matrix; the new `status_quotes` lookups + delete sweep; `handleQuoteRequest`/`handleQuoteResponse` (accept/reject/no-match/forgery-guard); `getNoteFromStatus` emission; create-path local quote + policy persistence; route 404/422/success; `createNoteJob` non-regression; and the revocation host-authority check.


llun:
### Sub-agent code-review — round 1

Fanned out security/correctness/invariants+tests reviewers over the diff; each finding adversarially verified. **4 confirmed** (1 high, 1 medium, 2 low), rest false positives. Addressed in `1d833101`:

- **[high · security] Forged quote handshake.** `handleQuoteResponse` settled an inbound Accept/Reject by `quoteRequestId` alone (a publicly-derivable id), so any validly-signed federated actor could force our pending outbound quote to `accepted` (injecting an attacker-controlled `quoteAuthorization`) or grief it via `Reject`. Now the responder must be the quoted status's **author** (exact match, host fallback), and only a same-authority stamp is stored.
- **[medium · security] Non-public status defaulted to public quoting.** A followers-only/direct post with no explicit policy defaulted to `public`, so `canQuoteStatus` allowed anyone to quote it. New `getEffectiveQuoteApprovalPolicy` defaults non-public statuses to `nobody` (used by the enforcement point and both advertising paths), and `handleQuoteRequest` now gates on `canActorReadStatus` — symmetric with the create route.
- **[low · tests] Shared-inbox `QuoteRequest` routing** now has `getJobMessage` tests (happy path + spoofed-actor reject).
- **[low] Stamp-route params type** (`OnlyLocalUserGuardHandle & { id }`): kept consistent with the sibling `statuses/[id]` route's established pattern; it type-checks and only backs the `.params` cast. Declined for consistency.
- **[chore] Prettier**: reformatted a test file the CI `prettier:check` gate flagged.

Gate re-run green: `prettier:check`, lint (0 errors), build, test (668 files, 6388 passing).

llun:
### Sub-agent code-review — round 2 (convergence)

Re-audited the handshake authorization after the round-1 fixes. One confirmed **[high]** left on the *request* side (the mirror of round-1's response-side fix): `handleQuoteRequest` only verified the `instrument`'s authorship for the embedded-object form. A remote actor could send a QuoteRequest quoting our public status with a **bare-id `instrument` pointing at a different local user's status**, and `createStatusQuote` (upsert by `statusId`, no FK) would forge/clobber an accepted "Carol quoted Bob" edge and host a real stamp for it.

Fixed in `d3791969`: the instrument must belong to the HTTP-sig-verified requester (embedded → `attributedTo` == actor; bare id → same host as actor) **and** must not be on our own host (an inbound QuoteRequest's instrument is always remote). Tests cover foreign-host and local-host rejection. Gate green (prettier:check, lint, build, test 6390).

llun:
## Round 3 — final handshake-authority audit addressed (commit `a350fe8d`)

The final adversarial audit surfaced two real, remaining authorization defects in the FEP-044f handshake. Both are now fixed with regression tests (each verified failing on the pre-fix tree first):

### 1. HIGH — embedded-instrument authority bypass (`lib/actions/handleQuoteRequest.ts`)
The round-2 fix bound the quoting note to the requester's host only on the **bare-id** `instrument` path. The **embedded-object** path checked only `attributedTo === actor` — but `attributedTo` is an attacker-controlled field of the attacker-supplied embedded object. A remote actor could send `instrument: { id: "https://victim.example/…/statuses/999", attributedTo: <self> }` and forge/clobber a third party's quote edge (and get us to issue a hosted QuoteAuthorization stamp for a note the victim never made).
**Fix:** the instrument id must **always** be hosted under the requester's authority (`sameHost(instrumentId, request.actor)`), with the embedded `attributedTo` checked as an additional guard.

### 2. MEDIUM — unauthenticated inbound stamp persistence + ambiguous lookup (`lib/jobs/createNoteJob.ts`, `lib/database/sql/statusQuote.ts`)
`createNoteJob` persisted an inbound note's `quoteAuthorization` verbatim, with no authority check (the sibling `handleQuoteResponse` already guards with `sameHost`). Combined with the non-unique `authorizationUri` index + `.first()`, a forged note could shadow a legitimate accepted stamp (404 a valid authorization; misdirect revocation).
**Fix:** a stamp uri is stored only when the quote **verified as accepted** AND the stamp is same-authority with the quoted status; `getStatusQuoteByAuthorizationUri` now prefers the accepted edge and orders deterministically.

Also reconciled the rebase onto main's moderation-state feature (#1240).

Gate green on `a350fe8d`: prettier ✓, lint 0 errors ✓, build ✓, **6435 tests passed** ✓.

llun:
## Round 4 — instrument authorship (not just host) addressed (commit `3281c48b`)

The round-4 audit found the round-3 `sameHost` binding was still insufficient: a host serves many actors, so on a **multi-user instance** a verified `mallory@shared.example` could name co-resident `alice@shared.example`'s note as the QuoteRequest `instrument` — `sameHost` passes, and we'd forge an accepted edge + hosted stamp for alice's note and clobber alice's real edge. HTTP signatures prove who **sent** the request, not who **authored** the instrument note.

**Fix:**
- New `lib/services/quotes/verifyQuoteInstrument`: dereferences the canonical instrument note (instance-signed + compacted) and requires (a) the fetched note id == the requested id (no redirect/alias substitution), (b) `attributedTo` == the requester, and (c) it actually quotes the target status. Any fetch/parse failure ⇒ not verified (never auto-approved).
- `handleQuoteRequest` keeps the cheap `sameHost`/own-host pre-filter as a fast reject, then gates on `verifyQuoteInstrument` as the authoritative authorship + intent check. The self-attested embedded `attributedTo` is no longer trusted.
- The per-user inbox now dispatches `QuoteRequest` to `HANDLE_QUOTE_REQUEST_JOB` (like the shared inbox), so the authorship-verifying fetch runs in the worker instead of blocking the inbox response.

Gate green on `3281c48b`: prettier ✓, lint 0 errors ✓, build ✓, **6442 tests passed** ✓ (+7).

llun:
## Round 5 — revocation/response authorized by author, not host (commit `f457485c`)

The round-5 audit found the round-4 flaw class (host ≠ authorship on a multi-user instance) still present on the **revocation** path. Because a stamp URI is public and computable, any co-resident of the quoted author could `Delete` it and permanently revoke another user's authorized quote (`revoked` is terminal ⇒ irreversible).

**Fix:**
- `deleteObjectJob`: resolve the quoted status's author from `edge.quotedStatusId` (`getOriginalStatus` + `normalizeActorId`) and require the Delete sender to match it **exactly**; fail closed if unresolvable. The host-only check is gone.
- `handleQuoteResponse`: hardened the sibling accept/reject path the same way — exact author match, no same-host fallback (which previously let a co-resident settle our outbound quote when the quoted status wasn't stored locally).

Both are safe: the outbound-quote flow stores the quoted status locally ( resolves it before creating the edge), so the exact author is always resolvable in practice.

Gate green on `f457485c`: prettier ✓, lint 0 errors ✓, build ✓, **6444 tests passed** ✓.

PR #1240, opened 2026-07-17
minor: add moderation state model and enforcement hooks

llun:
## Epic 4.2a — Moderation data model + enforcement plumbing

First PR of the Admin moderation API epic (Phase 4). Ships the moderation **data model** and the **enforcement hooks**, with states settable only from the DB mixin (tests) — **no HTTP endpoints yet**. Every new column defaults `NULL`, so production behavior is unchanged until later PRs (4.2b/4.2c) expose the switches.

### Data model (1 migration, BOTH reference dumps regenerated)
- `actors.suspendedAt` / `silencedAt` / `sensitizedAt` — on `actors` because suspend/silence/sensitize apply to **remote actors too**
- `accounts.disabledAt` / `approvedAt` — login concepts; `approvedAt` backfilled to `createdAt` for existing rows
- `reports.assignedActorId` / `actionTakenAt` / `actionTakenByActorId` — report workflow columns (consumed in 4.2c)
- new append-only `moderation_actions` audit table

Both `migrations/schema.sql` (PostgreSQL) and `migrations/schema.sqlite.sql` (SQLite) were regenerated by running all 139 migrations against a fresh local PG 17 and SQLite; `SELECT count(*) FROM knex_migrations` = 139 files on both.

### Enforcement hooks (Decision 3 matrix)
| State | Hook | Effect |
|---|---|---|
| suspended (local) — API | `OAuthGuard` (bearer + cookie + `OAuthAppGuard`) | 403 |
| suspended (local) — AP | `OnlyLocalUserGuard` | 410 Gone on actor/outbox/followers/following/inbox/status |
| suspended (remote) — inbound | shared + per-user inbox | swallow (202, no queue / no side effects) |
| suspended+silenced — timelines | `filterModeratedStatuses` in `getFilteredStatusPage` | drop suspended everywhere; drop silenced on public surfaces |
| disabled — sign-in | `canCreateSessionForAccount` helper | reject session creation |
| disabled — API | `OAuthGuard` | 403 |
| sensitized — creation | `createNoteFromUserInput` / `createPollFromUserInput` | force `sensitive: true` |
| serializer flags | `getMastodonAccountFromSQLActor` | emit `suspended` / `limited`; directory excludes both |

Silence never blocks API access — only public-surface visibility.

### New moderation DB mixin (`lib/database/sql/moderation.ts`)
`setActorSuspended` / `setActorSilenced` / `setActorSensitized`, `setAccountDisabled`, `approveAccount` (idempotent), `rejectPendingAccount` (deletes a pending account + its actors), `getModerationStatesForActors` (single batch query), `createModerationAction`, `deleteAllAccountSessions`.

### Plan deviations (reconciled with code at HEAD)
- **`approvedAt` is set in `createAccount`, not `registerAccount`** — `createAccount` is the single insertion point for local accounts, so approving there uniformly covers registration, the mock-user script, and stub seeding, and guarantees "every account is approved today" with no second write. Functionally identical to the plan's Decision 4.
- **`surface: 'public'` applied to all genuine public surfaces**, including the web hashtag page and the anonymous landing feed (not only the three API routes named in the plan), so a silenced author is hidden consistently across API and web. Suspended authors were already dropped everywhere via the default.
- Added `GONE` (410) to the shared response utility (`HTTP_STATUS`/`codeMap`/`REASON_PHRASE`) so `OnlyLocalUserGuard` can return it.

### Tests / gate
Strict TDD (RED → GREEN) for every task. Full pre-commit gate green: `prettier` → `lint` (0 errors) → `build` → `test` (**667 files, 6384 tests passing**).


llun:
### Sub-agent code-review — round 1 (addressed in 851da49b)

Two review agents (correctness + invariants/tests) reviewed the full diff. Findings and resolutions:

- **HIGH — moderation state not enforced on the cookie/session auth path.** `getActorsForAccount` (used by `getActorFromSession`) built its `Actor` directly and omitted `suspendedAt`/`silencedAt`/`sensitizedAt`, so a suspended local actor was blocked only on the bearer path (which uses `getActor`), and a sensitized actor's web-composed notes/polls (`POST /accounts/outbox` via `AuthenticatedGuard`) were not forced sensitive. **Fixed** by threading the three columns through `getActorsForAccount`'s `Actor.parse`. Regressions added: `getActorsForAccount` surfaces the state; OAuthGuard cookie session with a suspended actor → 403.
- **MEDIUM — federated public timeline leaked silenced authors.** The `getFilteredTimelinePage` wrapper never forwarded the new `surface` param, so the `[timeline]` route's `FEDERATED_PUBLIC` feed used the `'following'` default. **Fixed** by forwarding `surface` and passing `'public'` for `federated-public`. Regression added.
- **Coverage gaps — fixed:** `OAuthAppGuard` suspended-actor → 403; `rejectPendingAccount` non-existent-account → false, plus `account_providers`/`sessions` cascade assertions.

Verified non-issues (agent-confirmed): `rejectPendingAccount` FK-safety on PostgreSQL; inbox suspended-sender drop placement + id-form match; timeline filter Announce handling + pagination; app-token handling (no-actor tokens skip the block); migration + both dumps consistent. `approvedAt` gate is a deliberate no-op today ("no approval-required mode yet"), and `rejectPendingAccount` is correspondingly dormant until such a mode creates `approvedAt = null` rows — by design.

Full pre-commit gate green after the fixes: prettier → lint (0 errors) → build → test (**667 files, 6389 tests**).

llun:
### Sub-agent code-review — round 2 (clean)

A focused round-2 agent re-reviewed the round-1 fix delta (commit 851da49b) and traced each enforcement boundary end-to-end. **Verdict: both fixes correct and complete, no remaining defects.**

- `getActorsForAccount` now loads (SELECT *) and threads `suspendedAt`/`silencedAt`/`sensitizedAt` with the same time conversion as `getActor`; `getActorFromSession` resolves `currentActor` exclusively via this path, so the OAuthGuard cookie branch now blocks suspended actors and the sensitized-forces-sensitive rule fires for web-composed statuses. No other authenticated entry point drops the columns.
- The `[timeline]` route's `surface` mapping is correct and complete: `FEDERATED_PUBLIC` → `'public'`, everything else served here → `'following'`; `LOCAL_PUBLIC` is 404'd and the other public surfaces (`/timelines/public`, tag routes) pass `surface: 'public'` on their own direct `getFilteredStatusPage` calls.
- The three new regression tests genuinely exercise the fixed paths (verified non-tautological).

A clarifying comment about the `LOCAL_PUBLIC` 404 assumption was added in 678abcdd. A round-2 note that fitness/Strava **import** jobs call `createNote` directly (bypassing the sensitized rule) is a pre-existing, out-of-scope path — Decision 3 scopes forced-sensitive to the user-input services (`createNoteFromUserInput`/`createPollFromUserInput`), and a suspended actor cannot trigger imports anyway. Left as-is.

Round 2 surfaced no actionable issues, so the sub-agent review loop is complete for this PR.

PR #1253, opened 2026-07-18
minor: quote-compose action + embedded quote card (Epic 4.1e)

llun:
## Epic 4.1e — Quote-compose action + embedded quote card

Completes the two pieces deferred from 4.1d (#1251), finishing the Mastodon 4.5 quote-posts UI. **Browser-verified end to end** (screenshots below).

### Quote-compose action
- A **"Quote post"** item in the status menu (`PostMenu`), shown when an `onQuote` handler is provided, opens the composer with the quoted status pinned in a new `QuotedPreview`. Posting sends `quoted_status_id` through `createNote` (the outbox path authorizes it via `resolveQuoteForCreate`, shipped in 4.1d).
- Wiring: new `quoteAction`/`quoteStatus` in the timeline `statusActionReducer`; `onQuote` drilled `Posts → Post → Actions → PostMenu`; `MainPageTimeline` feeds `quotedStatus` to the `PostBox`.

### Embedded quote card
- `Post` renders `status.quote` as a `QuoteCard`. For an **accepted** edge it lazily fetches the quoted status via a new `getStatusById` client function — the status route **404s when the viewer may not see it**, so unreadable quotes fall back to an "unavailable" tombstone rather than leaking content. `pending`/`rejected`/`revoked`/`deleted` render state-specific tombstones.
- **Depth bounded at 1**: the fetched quoted status is a compact plain-text preview (`htmlToPlainText`, so no `dangerouslySetInnerHTML`), never a nested card. Relative time uses the server-provided `currentTime` (no `Date.now()` in client render).
- **No hot-path DB change**: the card reads the already-hydrated domain edge and fetches content client-side through the existing, readability-enforcing route — avoiding a recursion-sensitive change to the timeline hydration path.

### Verification
- Gate green: `prettier:check` ✓, `lint` 0 errors ✓, `build` ✓, **6544 tests** ✓ (new: reducer `quoteAction`, PostMenu Quote item, PostBox quote passthrough + `QuotedPreview`, `QuoteCard` accepted/unavailable/tombstone matrix).
- **Browser-verified** on the local SQLite stack: opened a post's menu → "Quote post" → composer showed the "Quoting" preview → posted → the new status rendered an embedded quote card with the quoted author, timestamp, and content.

With this, Epic 4.1 (quote posts) is feature-complete: compose a quote, choose/change who can quote, revoke, federate (FEP-044f), and see rendered quote cards.

### DIFF
--- a/app/(timeline)/MainPageTimeline.tsx
+++ b/app/(timeline)/MainPageTimeline.tsx
@@ -16,7 +16,12 @@ import { ActorProfile } from '@/lib/types/domain/actor'
 import { EditableStatus, Status } from '@/lib/types/domain/status'
 import { cn } from '@/lib/utils'
 
-import { clearAction, editAction, statusActionReducer } from './reducer'
+import {
+  clearAction,
+  editAction,
+  quoteAction,
+  statusActionReducer
+} from './reducer'
 
 interface MainPageTimelineProps {
   host: string
@@ -57,6 +62,11 @@ export const MainPageTimeline: FC<MainPageTimelineProps> = ({
 
   const onEdit = (status: EditableStatus) => {
     dispatchStatusAction(editAction(status))
+    window.scrollTo({ top: 0 })
+  }
+
+  const onQuote = (status: Status) => {
+    dispatchStatusAction(quoteAction(status))
     window.scrollTo({ top: 0 })
   }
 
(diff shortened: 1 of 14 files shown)


llun:
## Review round 1 — two confirmed defects fixed (commit `233274c8`)

The adversarial review surfaced two real (low-severity) defects, both fixed with regression tests:

1. **QuoteCard stuck on loading** — the lazy `getStatusById` fetch chained only `.then`, so a *rejected* fetch (offline, aborted, malformed body) left the card on "Loading quoted post…" forever (the tombstone only covered a resolved 404). Added a `.catch` → falls back to the unavailable tombstone.
2. **Poll + quote leak** — composing a poll while a quote was active silently dropped the quote (`createPoll` can't carry one) AND the poll branch never cleared parent state, so `quoteStatus` leaked onto the next post. Made poll and quote mutually exclusive: the poll toggle is disabled while quoting, and an already-open poll closes when a quote starts, so the poll branch can never run with a quote.

Also: the earlier `All Tests` red was the known-flaky `FitnessHeatmapView` map-effect test (unrelated to this diff; passes locally 44/44) — cleared on re-run.

Gate green on `233274c8`: prettier ✓, lint 0 errors ✓, build ✓, **6546 tests** ✓.

PR #1264, opened 2026-07-21
feat: merge post-box quote policy into the visibility dropdown

llun:
## Summary

Updates the post box to match the design-system Composer spec: the composer's two separate toolbar controls — **visibility** and **who-can-quote** — are merged into a **single dropdown**. The trigger shows the visibility, and the menu adds a **"Who can quote"** section (Anyone / Followers / No one) below a separator. The trigger surfaces the active quote-policy icon (in `text-primary`) once it differs from the default "Anyone".

Before: two toolbar dropdowns (`Public ⌄` and `Anyone can quote ⌄`). After: one (`Public ⌄`), matching the attached design.

## Changes

- **`visibility-selector.tsx`** — now accepts optional `quotePolicy` / `onQuotePolicyChange` props. When both are provided it renders the "Who can quote" section and the trigger's quote icon; when omitted it behaves exactly as before, so the standalone uses (`FitnessImport`, `StravaSettingsForm`, `StravaArchiveImportSection`) are unaffected. The followers visibility label is renamed `Followers only` → `Followers` per the design.
- **`post-box.tsx`** — replaces the `VisibilitySelector` + `QuoteApprovalPolicySelector` pair with the single merged `VisibilitySelector`.
- **Removed** the now-unused `quote-approval-policy-selector.tsx` (+ its test). The "Change who can quote" submenu on *existing* posts (`post-menu.tsx`) has its own inline options and is untouched.
- Added `visibility-selector.test.tsx` covering both the standalone visibility path and the merged quote path.

Net: 193 insertions, 168 deletions across 5 files.

## Design reference

Design-system `ui_kits/web/Composer.jsx` in the "Activities.next Design System" project (translated from the kit's raw `hsl()` styles to the app's shadcn/Tailwind tokens).

## Verification

Definition-of-Done gate, all green:

- `yarn run prettier --check` ✓
- `yarn lint` — 0 errors ✓
- `yarn build` ✓
- `yarn test` — post-box + visibility-selector + post-menu suites, 34 passed ✓

Browser-verified with a local SQLite mock user in **light and dark mode**:

- Composer shows a single merged dropdown (was two).
- Open menu matches the design: Public / Unlisted / Followers / Direct with descriptions, a separator, then the **WHO CAN QUOTE** section.
- Selecting "No one" updates the policy and the trigger shows the orange no-quote icon; the trigger's accessible name reads *"Set visibility and who can quote, current: Public, No one can quote"*.
- No console errors.


lib/components/post-box/visibility-selector.tsx
```diff
@@ -75,12 +117,15 @@ export const VisibilitySelector: FC<Props> = ({
...
           disabled={disabled}
-          title={currentOption.label}
-          aria-label={`Set visibility, current: ${currentOption.label}`}
+          title={triggerTitle}
+          aria-label={triggerLabel}
           className="gap-1.5 px-2.5 text-xs font-medium text-muted-foreground hover:text-foreground"
         >
           <CurrentIcon className="size-4" />
           <span>{currentOption.label}</span>
+          {showQuotePolicy && QuoteIcon && quotePolicy !== 'public' ? (
```
llun, line 129: **[test gap]** This `quotePolicy !== 'public'` trigger-icon branch is the key new piece of logic the refactor adds, and it has no test. A future edit flipping or dropping the guard would go uncaught. Add a test asserting the trigger gains one extra `<svg>` for a non-public policy and none for `public`.

lib/components/post-box/visibility-selector.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    expect(
+      screen.getByRole('button', {
+        name: /set visibility and who can quote, current: public, followers can quote/i
+      })
+    ).toBeInTheDocument()
+  })
+
+  it('invokes onQuotePolicyChange with the picked policy', async () => {
```
llun: **[test gap]** Quote-policy selection is only exercised for `nobody`. `followers` is the interesting case: the merge makes its label collide with the visibility "Followers", so a scoped test (within the "Who can quote" group) proving the quote "Followers" invokes `onQuotePolicyChange('followers')` would catch a mis-wire the current suite cannot. Suggest folding all three (`public`/`followers`/`nobody`) into a table-driven `it.each`, scoped to the quote group.

lib/components/post-box/visibility-selector.tsx
```diff
@@ -122,6 +167,50 @@ export const VisibilitySelector: FC<Props> = ({
             </DropdownMenuItem>
           )
         })}
+
+        {showQuotePolicy ? (
+          <>
+            <DropdownMenuSeparator />
+            <DropdownMenuLabel className="px-2 py-1 text-[11px] font-medium tracking-wide text-muted-foreground uppercase">
```
llun: **[a11y]** The merged menu now contains two independent single-select sets — 4 visibility `menuitemradio`s and 3 quote-policy `menuitemradio`s — as flat siblings under one `role="menu"`, with no `role="group"` wrapper. Assistive tech announces a single radio set with two simultaneously-checked items, and the renamed "Followers" visibility label collides by accessible name with this quote "Followers", distinguished only by the visually-adjacent "Who can quote" label, which is not programmatically associated. Wrap each set in a labeled `DropdownMenuGroup` (`aria-label="Visibility"` / `aria-labelledby` → this label id) so the two radio sets are distinct, named groups. Flagged independently by two reviewers.

llun: Fixed in e5d9bad4. Wrapped the visibility options in `<DropdownMenuGroup aria-label="Visibility">` and the quote options in `<DropdownMenuGroup aria-labelledby={quoteLabelId}>`, with the "Who can quote" `DropdownMenuLabel` carrying that id. Each set is now a distinct, named `role="group"`, so AT reads two independent single-select groups and the duplicate "Followers" name is disambiguated by group.

lib/components/post-box/visibility-selector.tsx
llun: Fixed in e5d9bad4. Added a test asserting the trigger has one extra `<svg>` for a non-public policy (`nobody`) versus `public`.

lib/components/post-box/visibility-selector.test.tsx
llun: Fixed in e5d9bad4. Replaced the single `nobody` case with a table-driven `it.each` over `public`/`followers`/`nobody`, resolving each option via `within(getByRole('group', { name: /who can quote/i }))` so the quote "Followers" is scoped away from the visibility "Followers".

PR #1265, opened 2026-07-21
fix: align inline fitness post card with design system

llun:
## Summary

Restyles the fitness activity card rendered inside a timeline post ([`lib/components/posts/post.tsx`](https://github.com/llun/activities.next/blob/claude/fitness-status-design-1c021f/lib/components/posts/post.tsx)) to match the design system's `FitnessChip` spec (`ui_kits/web/FitnessChip.jsx` in the "Activities.next Design System" project). This is the inline card in the post/feed — **not** the completed-fitness detail page (`FitnessStatusDetail.tsx`, unchanged).

### Changes
- **Card**: `rounded-lg border` on `bg-background` (was a `bg-muted/30` gray tint) — a theme-aware white that lifts off the framed timeline card (`bg-card`) and stays correct in dark mode.
- **Header row**: orange `Activity` icon (`text-primary`), a **monospace** muted file-name link, and a right-aligned **rounded pill** for the file type. Removes the old "Fitness" word label and the `uppercase` type text.
- **Stats**: a labeled **4-up grid** (Distance / Duration / Pace|Speed / Elevation — label over value) instead of the old inline "Distance: **value**" spans. The pace/speed label stays derived from `getFitnessPaceOrSpeed`.
- **Device**: moved to its own `Via: <device>` footer line (still via `BrandedDeviceLink`).

No activity-title row: the design's `fitness.title` will be handled in the design; `StatusFitnessFile` has no title field and `fitness.description` is left untouched by the chip.

The processing, failed/retry, and "View on Strava" source-link branches are unchanged.

## Testing
- `yarn run prettier --write .`, `yarn lint` (0 errors), `yarn build` (exit 0), `yarn test` (**688 files / 6572 tests passed**) — all green.
- Updated `post.test.tsx` to assert the new labeled stat grid, type pill, and device link, and that the old inline "Distance:" / "Fitness" markup is gone.
- **Browser-verified** locally (SQLite mock data) in **light and dark mode** across three variants: full card with a branded device, a card with a Strava source link, and a card with a plain (unknown-brand) device.

## Notes
- Number formatting is unchanged (existing formatters): Duration renders `H:MM:SS`, and the metric is `Pace` (run/walk/hike/swim) or `Avg speed` otherwise — so it may differ from a design mock that hardcodes `2h 38m` / `312 m/h`. Out of scope for this restyle.

### DIFF
--- a/lib/components/posts/post.tsx
+++ b/lib/components/posts/post.tsx
@@ -200,6 +200,9 @@ export const Post: FC<PostProps> = (props) => {
         <div className="mt-2 max-w-full rounded-lg border bg-background p-3 text-xs">
           <div className="flex max-w-full items-center gap-2">
             <Activity className="size-4 shrink-0 text-primary" />
+            {/* The visible "Fitness" label was dropped in the redesign; keep the
+                activity semantic for screen readers. */}
+            <span className="sr-only">Fitness activity</span>
             <a
               href={fitnessFile.url}
               target="_blank"
--- a/lib/components/posts/post.tsx
+++ b/lib/components/posts/post.tsx
@@ -139,6 +139,22 @@ export const Post: FC<PostProps> = (props) => {
     activityType: fitnessFile?.activityType
   })
   const fitnessSourceUrl = normalizeFitnessSourceUrl(fitnessFile?.sourceUrl)
+  const fitnessDeviceLabel = getDeviceDisplayLabel(
+    fitnessFile?.deviceName,
+    fitnessFile?.deviceManufacturer
+  )
+  // Labeled stat cells for the 4-up grid, in the design's order, dropping any
+  // metric the file doesn't provide.
+  const fitnessStats = [
+    { label: 'Distance', value: fitnessDistance },
+    { label: 'Duration', value: fitnessDuration },
+    fitnessPaceOrSpeed
+      ? { label: fitnessPaceOrSpeed.label, value: fitnessPaceOrSpeed.value }
+      : null,
+    { label: 'Elevation', value: fitnessElevation }
+  ].filter((stat): stat is { label: string; value: string } =>
+    Boolean(stat?.value)
+  )
   const isOwner =
     Boolean(actualStatus.isLocalActor) &&
     props.currentActor?.id === actualStatus.actorId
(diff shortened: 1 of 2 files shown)


lib/components/posts/post.tsx
```diff
@@ -181,22 +197,19 @@ export const Post: FC<PostProps> = (props) => {
...
       </TranslateContent>
       {fitnessFile ? (
-        <div className="mt-2 max-w-full rounded-md border bg-muted/30 px-3 py-2 text-xs">
+        <div className="mt-2 max-w-full rounded-lg border bg-background p-3 text-xs">
           <div className="flex max-w-full items-center gap-2">
-            <Activity className="size-3.5 shrink-0 text-muted-foreground" />
-            <span className="shrink-0 font-medium text-muted-foreground">
-              Fitness
-            </span>
+            <Activity className="size-4 shrink-0 text-primary" />
```
llun, line 202: **a11y:** The redesign drops the visible "Fitness" text label, so the activity semantic now rests on the `Activity` icon alone, which has no accessible name. A screen-reader user hears only the filename link and the type pill, with no cue that this is a recorded fitness activity. Suggest an `sr-only` "Fitness activity" label to restore that cue without touching the visuals.

lib/components/posts/post.tsx
```diff
@@ -225,48 +238,28 @@ export const Post: FC<PostProps> = (props) => {
...
-              ) ? (
-                <span className="text-muted-foreground">
-                  Via:{' '}
-                  <BrandedDeviceLink
-                    deviceName={fitnessFile.deviceName}
-                    deviceManufacturer={fitnessFile.deviceManufacturer}
-                  />
-                </span>
-              ) : null}
+          {isFitnessCompleted && fitnessStats.length > 0 ? (
```
llun, line 244: **test coverage:** This new `.filter` empty-metric drop + `fitnessStats.length > 0` grid guard is the core new logic and is untested. Missing cases: (1) partial metrics — some cells dropped, others still render; (2) zero metrics — no grid renders at all, while the (now-decoupled) `Via:` device footer still shows; (3) the dynamic `Pace` vs `Avg speed` label and the `Duration` cell are not asserted in the card. Adding these pins the new logic against regressions.

lib/components/posts/post.tsx
llun: Fixed in 727198d5 — added an `sr-only` "Fitness activity" span right after the `Activity` icon, so screen readers announce the card as an activity again. Covered by a `getByText('Fitness activity')` assertion.

lib/components/posts/post.tsx
llun: Fixed in 727198d5. Added tests covering: partial metrics (only present cells render, missing labels absent), zero metrics (stat grid omitted while the `Via:` device footer still shows), the dynamic `Pace` vs `Avg speed` label (run vs ride), and the `Duration` cell. Post-file suite is now 35 tests, all green.

PR #1266, opened 2026-07-21
fix: resolve inbound remote quotes with valid FEP-044f stamps (Epic 4.1f)

llun:
## Summary

Fixes quote posts that rendered as a **"Quote pending approval"** tombstone instead of the embedded quote card shown in the design. When such a quote is also boosted, it shows up as **"Boosted by …"** over a broken/pending quote — which is the "quote status that shows Boosted by at the top and different from the design" in the report.

## Root cause

`verifyRemoteQuote` returns `pending` whenever the quoted post is **not already stored locally** — it needs the quoted author's id to validate the FEP-044f `quoteAuthorization` stamp, and that id was only read from a locally-stored quoted status (`lib/services/quotes/verifyRemoteQuote.ts:104` bails before fetching the stamp). Instances rarely already store the quoted post (e.g. someone quoting a `@Techmeme` post you don't follow), so virtually **every** inbound remote quote was stuck `pending` and rendered as a tombstone.

Confirmed against the live instance: `paul@tapbots.social`'s quote of a Techmeme post carries a **valid** `quoteAuthorization` stamp hosted on `techhub.social`, and Techmeme's policy is `canQuote.automaticApproval: Public` — yet llun.dev rendered it as `pending` because Techmeme's post isn't stored locally.

## Fix

`createNoteJob` now fetches the quoted note (instance-signed, mirroring the boost path in `createAnnounceJob`) and stores it when the quoting note carries a `quoteAuthorization` stamp. The quoted author is then knowable, the stamp verifies, the edge becomes `accepted`, and the quote card can load the quoted content.

**Security invariant preserved:** fetching only makes the author *knowable*; the FEP-044f three-field stamp check (`attributedTo`/`interactingObject`/`interactionTarget` + same-authority) still gates approval. A forged or mismatched stamp stays `pending` and its uri is not persisted (covered by a new test). The fetch is gated on a stamp being present, so stamp-less Fedibird/Misskey quotes keep their existing `pending` behavior and we don't fetch on every quote.

## Scope note

The **"Boosted by …"** header itself is a genuine boost of a quote (e.g. `mishari@mastodon.in.th` boosted paul's quote), which correctly matches the design's boost rows — so it is intentionally unchanged. This PR fixes the quote *content* not rendering, which is the actual divergence from the design.

## Testing

- New integration tests in `lib/jobs/createNoteJob.test.ts`:
  - stamped remote quote whose target isn't stored locally → fetched + stored + `accepted`;
  - unreachable quoted note → graceful `pending` (no crash, not stored);
  - forged stamp after fetching the note → still `pending`, stamp uri not persisted.
- `prettier --write .` → `lint` (0 errors) → `build` → `test` (**6574 tests**) all green.
- Verified separately in a local browser that an `accepted` quote renders exactly like the design's card.

### DIFF
--- a/docs/mastodon-api-compatibility.md
+++ b/docs/mastodon-api-compatibility.md
@@ -154,7 +154,12 @@ product or security decision, not a gap to be closed.
   non-author viewer sees `automatic` when they are an accepted follower of the
   author and `denied` otherwise (an anonymous viewer still sees `unknown`); the
   verdict is resolved in one batched follow query per page, so it adds no N+1.
-  Legacy Fedibird (`quoteUri`)
+  An inbound quote that arrives with a valid `quoteAuthorization` stamp is
+  accepted even when the quoted post is not already stored locally: the quoted
+  note is fetched (instance-signed, like the boost path) so the stamp can be
+  verified against its author and the quote card can embed the content. Fetching
+  only makes the author knowable — the stamp's three-field match against that
+  author is still what grants approval. Legacy Fedibird (`quoteUri`)
   and Misskey (`_misskey_quote`) quotes carry no stamp, so they are stored and
   rendered as unapproved (`pending`) rather than as embedded quotes, matching
   Mastodon 4.5's treatment of stamp-less quotes. Revoking approval fans the stamp
--- a/lib/jobs/createNoteJob.test.ts
+++ b/lib/jobs/createNoteJob.test.ts
@@ -3,10 +3,15 @@ import fetchMock, { enableFetchMocks } from 'jest-fetch-mock'
 import { getTestSQLDatabase } from '@/lib/database/testUtils'
 import { createNoteJob } from '@/lib/jobs/createNoteJob'
 import { CREATE_NOTE_JOB_NAME } from '@/lib/jobs/names'
+import {
+  buildQuoteAuthorizationObject,
+  buildQuoteAuthorizationUri
+} from '@/lib/services/quotes/quoteAuthorization'
 import { mockRequests } from '@/lib/stub/activities'
 import { seedDatabase } from '@/lib/stub/database'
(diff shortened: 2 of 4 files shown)


llun:
### Sub-agent code review — round 1 (addressed in 18ed7688)

Fanned out two reviewers (security/correctness + conventions/tests) over the diff. Findings and resolutions:

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | med-high | **Unbounded recursion / DoS**: the new recursive `createNoteJob` for the quoted note had no depth bound — a chain of distinct stamped notes (A→B→C…) could drive N nested fetches per inbound note. | Bounded to a **single hop** via an internal `skipQuoteResolution` flag on `JobMessage`, set on the recursive call. New test `bounds quote resolution to a single hop` (verified it fails without the flag). |
| 2 | med | **Orphaned note on throw**: the recursive call ran outside try/catch; a blocked quoted-author domain (`assertActorCanFederate`) or a store error would throw *after* the quoting note was inserted, permanently orphaning it (no edge/timeline/tags) — violating the "never drops the note" invariant. | Wrapped the fetch/recursion in **try/catch** → degrades to `pending` and logs a warning. New test `does not orphan the quoting note when the quoted author domain is blocked`. |
| 3 | low | **Test gap**: no case proving a stamp-less, not-locally-stored quote does **not** fetch. | Added `does not fetch a not-locally-stored quote target when there is no stamp`. |

Confirmed safe (no change needed): trust model (the FEP-044f three-field stamp check still gates approval after the fetch — forged stamp stays `pending`, covered by a test), SSRF posture (same instance-signed `getNote` guard as the boost path), id-mismatch degradation, and timeline side effects (consistent with the boost path).

Full gate green after the fix: prettier → lint (0 errors) → build → **6577 tests**.

PR #1283, opened 2026-07-22
fix: unify status post actions across every page via one shared component

llun:
## Problem

A status post showed **different controls depending on the page**. The reported case: a post on a **profile page** was missing the **Quote** action the home timeline has. It turned out to be systemic:

| Surface | Reply | Quote | Edit own |
|---|---|---|---|
| Home timeline | ✅ | ✅ | ✅ |
| Profile | ✅ | ❌ | ❌ |
| Lists / Favourites / Hashtags / Bookmarks / Collections | ❌ dead button | ❌ | ❌ |
| Search | ❌ dead button | ❌ | ❌ |
| Status detail | ✅ | ❌ | ❌ |

The components were **already shared** (`Posts` → `Post` → `Actions` → `PostMenu`); the divergence was that **each page wired a different subset of action callbacks**, and the shared components silently hid any action whose callback was missing (e.g. the menu only rendered "Quote post" `if (onQuote)`). Reply/quote/edit need a composer, and that composer + its state lived only in `MainPageTimeline`, so every other page under- or un-wired them — the **Reply button was a no-op on six feeds**.

## Fix

Move the reply/quote/edit interaction **into the shared feed layer** so the action set is owned by `Posts`, not by pages — pages can no longer drift because there is nothing per-page to wire.

- **New shared pieces**
  - `useInlineComposer` — tiny hook holding the active `{ status, mode }`.
  - `InlineStatusComposer` — one inline composer rendered beneath the post: `StatusReplyBox` for reply, `PostBox` for quote/edit. Reply/quote/edit all open **inline** (chosen over a modal).
- **`Posts`** now supplies `onReply` / `onQuote` / `onEdit` itself whenever `currentActor` + `showActions` are present, and renders the composer under the active post. Pages pass only optional **data-sync** callbacks: `onStatusCreated`, `onPostUpdated`, `onPostDeleted`, `onLikeChanged`, `onBookmarkChanged`.
- **Status detail** (`StatusBox`) adopts the same hook + composer for its primary post.
- **Home timeline** keeps its top composer for brand-new posts; per-status quote/edit now happen inline like everywhere else, so the page-local `statusActionReducer` is removed.
- `isMediaUploadEnabled` is threaded to every feed so the composer can attach media on all surfaces, not just the home timeline.

Out of scope by design: the bespoke `FitnessStatusDetail`, the `StatusNotification` snippet, and read-only / logged-out surfaces (`showActions={false}`).

## Docs

- **AGENTS.md** → new **Status Posts & Actions** section + a step in the "Adding a page in the `(timeline)` group" recipe.
- **CLAUDE.md** → key-reminder bullet.
- **REVIEW.md** → checklist item.

These mandate the single component + centralized action set so future pages cannot reintroduce the drift.

## Tests

- `useInlineComposer.test.ts` — open / replace / close state machine.
- `inline-status-composer.test.tsx` — mode → component mapping + created/updated bubbling and close.
- `posts.test.tsx` — a post action opens the shared inline composer for the active post.
- Updated `ActorTimelines.test.tsx` for the renamed callback.
- Full gate green: `prettier --write .` clean, `lint` 0 errors, `build` ✓, `test` **689 files / 6596 tests pass**.

## Manual verification (local SQLite + mock data)

- Home-timeline own-post ⋯ menu: Quote / Edit / Change visibility / Change who can quote / Copy link / Delete. ✅
- **Profile-page** own-post ⋯ menu now shows **Quote + Edit** (previously absent) — the reported bug. ✅
- "Quote post" opens the inline quote composer directly beneath the post. ✅
- New-post composer create flow works. ✅


llun:
### Sub-agent review — round 1

Fanned out three reviewers (correctness, project-conventions, test-coverage) over the full diff. **No blockers or majors on conventions**; one reachable correctness defect and several quality/coverage items. All addressed in `b37bbbdf`:

| # | Finding | Severity | Resolution |
|---|---------|----------|------------|
| 1 | **Duplicate composer render** — the active composer was matched by the *unwrapped* status id, so two rows resolving to the same original (a boost + its original, or two boosts of the same post) both rendered a composer. Reachable on the home timeline (dedupe only drops a boost when the original is independently present). | Major | Anchor the composer on the **wrapper row id** (unique per row), passed from the per-row map. Added a regression test asserting two boosts of one original render a single composer. |
| 2 | Six secondary feeds (lists, favourites, bookmarks, hashtags, collections, search) enabled inline **Edit** but didn't wire `onPostUpdated`, so an edit persisted but showed stale text until reload. | Minor | Wired `onPostUpdated` (map-replace by id) on all six. |
| 3 | `AGENTS.md` "must not pass `onReply`/`onQuote`/`onEdit`" wording could be misread against `StatusBox`, which legitimately wires the shared composer directly. | Minor (docs) | Added a one-line carve-out for the detail surface. |
| 4 | Redundant `status as EditableStatus` cast (types already equal). | Nit | Removed. |
| 5 | Coverage gaps: `handlePostUpdated`, the detail-vs-comment gate, boost anchoring, the not-actionable gate, and the callback-omitted close path. | — | Added tests in `useInlineComposer.test.ts`, `posts.test.tsx`, `inline-status-composer.test.tsx`, `ActorTimelines.test.tsx`, and `StatusBox.test.tsx`. |

Gate after fixes: `prettier` clean · `lint` 0 errors · `build` ✓ · **6603 tests pass**.

llun:
Merged latest `main` in (no conflicts). Since #1279 landed the `/explore` posts feed with the old per-page `onReplyCreated` wiring right as this PR removes it, CI's merge build failed — migrated `ExplorePageClient` to the shared contract (reply/quote/edit now come from `Posts`; the composer self-closes, and `onPostUpdated` reflects inline edits in the trends list). `/explore` is now covered by the same guarantee as every other feed. Gate green locally: prettier · lint · build · **6618 tests**.

llun:
### Sub-agent review — round 2 (convergence)

Two agents (delta-correctness + contract-uniformity) re-swept the changes since round 1 (anchor fix, `onPostUpdated` across all feeds, the main merge, and the `/explore` migration). **No blockers or majors** — the anchor fix is confirmed regression-free, all 12 `<Posts>` surfaces verified consistent, and the docs match the final code.

Applied the two optional nits they raised:
- `ActorTimelines` / `FavoritesTimeline` inline-edit sync is now Announce-aware, matching their delete handlers (a boosted row whose original was edited also refreshes).
- Corrected an AGENTS.md read-only example (`showReadOnlyStats` is used by the logged-out profile too, not only the landing feed).

One **pre-existing, out-of-scope** display gap was noted (logged-out hashtag/collection feeds don't pass `showReadOnlyStats`) — tracked separately, not touched here.

Round 2 surfaced nothing requiring a further round, so the loop has converged. Gate: prettier · lint 0 errors · build ✓ · **6618 tests**.

PR #1293, opened 2026-07-22
fix: compute fitness average speed from moving time to match Strava

llun:
## Problem

Fitness activity **average speed** computes lower here than on Strava. For [this ride](https://www.strava.com/activities/19419786976) Strava shows **25.7 km/h** but the service showed **24.4 km/h**.

Root cause (confirmed with the user's TCX): average speed was computed as `distance ÷ elapsed time`, where elapsed time is the TCX lap's `TotalTimeSeconds` (**including stops**). This ride begins with a **3:45 (225 s) stationary stop**, so:

| | Time used | Result |
|---|---|---|
| Service (before) | elapsed `4614 s` | 31.33 km ÷ 1.2817 h = **24.45 km/h** |
| Strava | moving `≈4389 s` | 31.33 km ÷ 1.2192 h = **25.7 km/h** |

Strava measures average pace/speed over **moving time** (stops excluded), not elapsed time.

## Fix

- **`getMovingTimeSeconds`** in `parseFitnessFile.ts` sums per-segment time, excluding segments at or below a `0.5 m/s` stopped threshold. Segment speed prefers the per-point device speed (TCX `ns3:Speed`, GPX/FIT speed), falling back to GPS `haversine/Δt`; unmeasurable segments count as moving so missing data never shrinks moving time below elapsed.
- New nullable **`fitness_files.movingTimeSeconds`** column (migration + both schema dumps regenerated), threaded through the parse → store → serialize → render path.
- **`getFitnessPaceOrSpeed`** now uses moving time for pace/speed when present, falling back to elapsed time otherwise. The displayed **Duration stays elapsed time** — exactly how Strava presents both.
- **`scripts/fitness/backfillFitnessMovingTime.ts`** recomputes the value for already-imported activities (idempotent; `--dry-run`, `--force`). New imports compute it automatically during processing.

## Verification

End-to-end through the real parser on the attached TCX:

```
elapsed (s)     : 4614
movingTime (s)  : 4374
BEFORE (elapsed): 24.4 km/h   <- old service value
AFTER  (moving) : 25.8 km/h   <- Strava says 25.7 km/h
```

(25.8 vs 25.7 is within display rounding; the 1.3 km/h gap is closed.)

## Tests

- `getFitnessPaceOrSpeed`: moving-time speed/pace + elapsed fallback
- `parseFitnessFile`: moving time excludes a stopped segment
- DB round-trip: `movingTimeSeconds` persisted separately from elapsed duration
- `backfillFitnessMovingTime` core + script arg parsing

`prettier` / `lint` / `build` / full `test` suite all green (the one full-suite "error" is the pre-existing worker heap-OOM the repo shards CI to avoid, not a failure).

## Notes for reviewers

- Migration `20260722000000_add_fitness_file_moving_time.js` adds a nullable `float` column; both `migrations/schema.sql` and `migrations/schema.sqlite.sql` regenerated.
- To fix existing posts on a deployment: `NODE_ENV=production ./scripts/fitness/backfillFitnessMovingTime.ts --actor-id https://<host>/users/<username>`.


llun:
### Sub-agent review — round 1

Fanned out three independent reviewers over the diff (correctness, project-invariant compliance, test coverage).

**Correctness** — no blocker/major bugs. Verified end-to-end: km/h↔m/s conversions (all parsers normalize `speed` to km/h; the threshold is m/s), the moving-time accumulation loop (Δt≤0 guard, `measuredAny`/`≤0 → undefined` fallback), the `Math.min(moving, elapsed)` clamp, the `getFitnessPaceOrSpeed` elapsed fallback, persistence (`numberFields` + `?? null`), and the backfill (candidate filter, dry-run, per-file error isolation, stable pagination). Non-actionable notes only (PG `real` rounding makes the `--force` idempotency check occasionally do a redundant write; auto-pause gaps can over-count moving time, which only ever *understates* the speed gain — matches the documented "close enough to Strava" intent).

**Project invariants** — fully compliant: migration + both schema dumps regenerated with a `float`/`real` column matching its siblings; no `console.*` outside scripts; no `fetch()` in components; `apiResponse`/`safeParse` untouched; domain zod + serializer + client type + route mapping all threaded consistently; the `post.tsx` change stays inside the shared `Post` renderer; docs updated.

**Test coverage** — flagged real gaps in the core derivation, now closed in c6827982:
- GPX with **no per-point speed** → GPS-derived moving time (haversine fallback)
- **FIT** records with device speed (excludes a stop)
- TCX with **no timestamps** → `movingTimeSeconds` undefined (elapsed fallback)
- TCX with **no stops** → moving == elapsed
- TCX **clamp** (lap total < trackpoint span) → moving ≤ elapsed
- serializer emits `movingTimeSeconds` only when present

This confirms the fix works for **.fit, .gpx, and .tcx** imports (device-speed path and GPS-fallback path), with graceful fallback to elapsed time when a file lacks the data to derive moving time.

PR #1295, opened 2026-07-23
none: fix jsdom focus-scope recursion crashing CI test shard 4 (the "OOM")

llun:
## Summary

CI **Test Shard 4 of 4** has been failing on every run (including on `main` / `version-bump/main`). The failures were being treated as a **heap OOM** — three prior changes raised the Node heap ceiling and split the suite into shards ([#1287](https://github.com/llun/activities.next/pull/1287), [#1289](https://github.com/llun/activities.next/pull/1289), [#1290](https://github.com/llun/activities.next/pull/1290)) — but the crash kept moving with the shard that happened to contain one test file.

**It was never heap growth.** It is a Radix UI `FocusScope` ↔ jsdom **infinite focus recursion**:

- jsdom dispatches `focus` / `blur` / `focusin` / `focusout` **synchronously** from `HTMLElement.focus()`.
- Radix's `FocusScope` (used by `DropdownMenu`, `Dialog`, …) answers a `focusout` that lands outside its container by calling `.focus()` again to pull focus back — which synchronously fires another `focusout` and **re-enters the same handler**.
- When the post `⋯` menu closes while its **Delete / Report / Mute / Block** confirmation dialog opens (`post-menu.test.tsx` → *"confirms before deleting…"*), two focus scopes hand focus back and forth forever, throwing `RangeError: Maximum call stack size exceeded`.
- The thrown errors piled up — **297,242 errors** and a **~76-minute** run in one CI shard — until the worker died. That surfaced as *"JavaScript heap out of memory."*

`post-menu.test.tsx` deterministically lands in shard 4 (Vitest shards by file), so **shard 4 failed every time** while 1–3 stayed green. A real browser dispatches these events asynchronously and has focus-fixup guards, so this is purely a jsdom artifact.

## Fix

Bound the **synchronous focus re-entry depth** in `vitest.setup.ts` (jsdom only) so the loop unwinds instead of overflowing the stack. Genuine focus flows never nest `.focus()` anywhere near this deep, so top-level `focus()` and `document.activeElement` behave exactly as before. This is centralized in the shared setup, so it also protects the other menu→dialog tests (mute / block / report) and any future ones.

Only `vitest.setup.ts` changes — no shipped code, so this is a `none:` (no-release) change. The prior heap-raise + 4-way sharding are left in place (harmless, and still good for wall-clock parallelism).

## Verification

| Check | Before | After |
|---|---|---|
| `post-menu.test.tsx` in isolation | 155s then worker crash, 25,689 errors | **2.2s, 10/10 pass, 0 errors** |
| CI shard 4/4 locally | crash, `Maximum call stack size exceeded` | completes, **0 stack overflows** |
| Full single-process suite (`--max-old-space-size=8192`) | intermittent OOM | completes, **0 "Maximum call stack" errors** |

- `yarn lint` — 0 errors
- `yarn prettier:check` — clean
- `yarn build` — passes

The only failures in the full local single-process run were `Test timed out in 30000ms` on node-env SQLite DB tests (a *different* set each run), which never install the jsdom focus guard and pass in <2s in isolation — local CPU contention from running 689 files single-machine, not regressions. Each CI shard runs ~173 files on its own runner where those tests complete in <1s.


llun:
### Sub-agent code-review round 1

Fanned out two review sub-agents (correctness/regression-risk + conventions/test-coverage) over the diff. Both concluded the change is **sound with no blocking issues**. Actionable items addressed in `8732920b`:

- **SVGElement coverage** — jsdom gives `HTMLElement` and `SVGElement` their own separate `focus()`, so a recursion routed through an SVG element wouldn't have been bounded. Factored the wrapper into a shared `installFocusDepthGuard()` and applied it to both prototypes. (Not a trigger for the observed bug — Radix `FocusScope` focuses HTML `<div>`s — but belt-and-suspenders.)
- **Documentation Maintenance** — added the new global install to `AGENTS.md`'s "`vitest.setup.ts` installs global mocks…" enumeration, and added a cross-reference in `ci.yml` so its "heap OOM" comment doesn't lead future readers to over-attribute the crash to heap growth.

Verified against the current tree (with `@radix-ui/react-dialog` 1.1.20): the previously-crashing `lib/components/posts/actions/post-menu.test.tsx` passes 10/10 in ~2.4s (was 155s + worker crash). `yarn lint`, `yarn prettier:check`, and `yarn build` are green. Confirmed no dedicated regression test is warranted — `post-menu.test.tsx` was red and is now green, so it *is* the regression coverage.

llun:
### Added commit — review-loop docs (`5705a2ca`)

Per follow-up: `gemini-code-assist` has been removed from the repo, so the docs no longer tell agents to loop in an external review bot. Updated `AGENTS.md` + `CLAUDE.md` so the **sub-agent fan-out is the whole review process** — renamed the section to "Code Review Loop (Sub-Agents)", replaced the "Loop in the other review bots" step with a note that no external bot runs and `/gemini review` (or any bot trigger) must not be posted/awaited, and dropped the "plus any external bots" done-when clause (kept a conditional note for re-adding a bot loop if one is reintroduced). Docs-only, `none:`.

(Bundled here since this PR already touches `AGENTS.md` and the change surfaced from this PR's own review loop — happy to split into its own PR if you'd prefer it separate.)

PR #1297, opened 2026-07-23
minor: add database-backed admin server settings

llun:
## Summary

Moves instance server settings out of `ACTIVITIES_*` environment variables and hardcoded constants into a database-backed settings layer that admins can edit from the UI, resolved by **`env → database → default`** precedence: an environment variable always wins and locks the field with a "Set by environment" badge, so removing the variable hands control back to the admin form.

Implements the `ui_kits/web/admin-settings.html` design (Activities.next Design System).

### What's included

- **New `server_settings` table** + `ServerSettingDatabase` operations (both reference schema dumps regenerated).
- **Registry + resolver** (`lib/config/serverSettings.ts` + `lib/services/serverSettings/`): typed settings with per-field lock metadata, atomic validated writes, and a per-database TTL cache so hot paths (inbox verification, outbound requests) don't pay a DB read per call.
- **Admin API** `GET`/`PATCH /api/v1/admin/server_settings` + `lib/client.ts` helpers.
- **Three new admin tabs** in a "Settings" group of the admin dropdown (before System):
  - **Instance** — name, description, contact email, languages; registrations open + allowed emails.
  - **Posts & media** — post size, media per post, poll limits, upload size limit, read-only storage backend.
  - **Network** — outbound request timeout, retries, response-size cap.
- **Federation policy block** on the existing Federation tab (mode, allowed actor domains; trusted media domains shown read-only since they feed the Edge-runtime CSP).
- **Consumers rewired to the resolver**: `/api/v1` + `/api/v2/instance`, registration (signup/signin/landing/status/heatmap pages, `registerAccount`, accounts + email-confirmation APIs, `getAccountFromSession`), federation domain policy, outbound request tuning, and **new** server-side status-length + poll-limit enforcement on the create/edit status APIs.

### Precedence & notable decisions

- `env → database → default`. Env-pinned fields are read-only in the UI.
- Poll `max_expiration` default is aligned to the enforced constant (`2678400`s) so advertising and enforcement agree at the default; poll schemas keep only a structural floor + absolute safety ceilings so an admin can raise limits above the Mastodon defaults.
- The unconfigured outbound-request timeout unifies to the documented config-schema default (`4000` ms) instead of the request wrapper's ad-hoc `10000` ms.
- Maintenance mode is intentionally out of scope (deferred; enforcement touches auth + federation).

## Commits

16 small, independently-green commits (data layer → registry/resolver → admin API → shared UI → grouped nav → one commit per tab → federation block → one consumer subsystem per commit → docs). See the commit list.

## Verification

Full local browser walk-through as an admin (SQLite): all four surfaces render correctly, and editing **Instance name → "My Test Instance"** persisted to `server_settings` and `GET /api/v2/instance` echoed `"title": "My Test Instance"` — confirming the admin UI → PATCH → DB → resolver → instance API chain end to end. Screenshots of the Instance, Posts & media, Network, and Federation-policy surfaces are attached in the review thread.

- `yarn run prettier --write .` → clean
- `yarn lint` → 0 errors
- `yarn build` → compiles
- `yarn test` → **702 files, 6701 tests passing**
- Both schema dumps regenerated (SQLite drift-check clean).

## Follow-ups (deliberately out of this PR)

These are already advertised via the instance API and enforced server-side / via env, so they're refinements rather than gaps:

- Make the built-in web composer's character counter reflect a non-default post-size limit (needs instance-config threading into the composer).
- Read the resolved upload-size cap in the media upload path (currently a synchronous Zod refine reading env config; the value is advertised via the instance API and the env-configured case already works).
- Maintenance mode (store + enforce).
- DB-backing trusted media domains (blocked on Edge-runtime CSP reads).

### DIFF
--- a/app/(nosidebar)/auth/signin/page.tsx
+++ b/app/(nosidebar)/auth/signin/page.tsx
@@ -15,6 +15,7 @@ import {
 import { getBaseURL, getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { getResolvedServerSettings } from '@/lib/services/serverSettings'
 import { getActorFromSession } from '@/lib/utils/getActorFromSession'
 import { Booleanish } from '@/lib/utils/zodBooleanish'
 
@@ -74,7 +75,10 @@ const Page: FC<Props> = async ({ searchParams }) => {
     return redirect(actor ? target : '/')
   }
 
-  const { auth, serviceName, registrationOpen } = getConfig()
+  const { auth, serviceName } = getConfig()
+  const {
+    registrations: { open: registrationOpen }
+  } = await getResolvedServerSettings(database)
   const credentialEnabled = auth?.enableCredential !== false
 
   // Use an absolute URL on the configured host (ACTIVITIES_HOST) so the logo
--- a/app/(nosidebar)/auth/signup/page.tsx
+++ b/app/(nosidebar)/auth/signup/page.tsx
@@ -14,9 +14,9 @@ import {
 } from '@/lib/components/ui/card'
 import { Input } from '@/lib/components/ui/input'
 import { Label } from '@/lib/components/ui/label'
-import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { getResolvedServerSettings } from '@/lib/services/serverSettings'
 
 export const dynamic = 'force-dynamic'
 export const metadata: Metadata = {
@@ -29,7 +29,7 @@ const Page: FC = async () => {
 
   // When the server has closed registration there is no sign-up to show; send
   // visitors to the landing, whose auth card explains registration is closed.
-  if (!getConfig().registrationOpen) {
+  if (!(await getResolvedServerSettings(database)).registrations.open) {
     return redirect('/')
   }
 
--- a/app/(timeline)/PublicTopBar.tsx
+++ b/app/(timeline)/PublicTopBar.tsx
@@ -1,16 +1,18 @@
 import Link from 'next/link'
-import { FC } from 'react'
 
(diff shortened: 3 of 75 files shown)


llun:
### Sub-agent code review — round 1 (addressed)

Fanned out three specialized reviewers (correctness/invariants, silent-failure/hot-path safety, test coverage) over the full diff. Findings and how they were addressed, in commits `fix: harden the server settings resolver per review` and `test: cover server settings resilience, parity, and limit enforcement`:

- **DB read failure silently reverted admin policy to permissive defaults, cached for the full TTL, unlogged.** Now logs, serves the last-known-good cached view (even if expired) rather than reverting a stored closed-registration/allowlist to the permissive default, and only falls back to env+defaults (cached briefly) when there is no prior view.
- **`updateServerSettings` claimed "atomic" but wrote non-transactionally and invalidated the cache outside `finally`.** Added a transactional batch write (`setServerSettings`) and moved invalidation into `finally`.
- **Env-parser parity bug:** `readEnvNumber` used `Number()` where `getConfig` uses `parseInt(,10)` (diverged on `5000ms`/`5e3`/`500.9`); `readEnvFederationMode` coerced an invalid value to the permissive `open`. Both fixed to match `getConfig` / fail safe.
- **Outbound request timeout default** kept at 10000ms (the request wrapper's prior no-env behavior) instead of dropping the common no-env deployment to 4000ms.
- **Cache TTL** shortened to 15s to bound cross-instance propagation of a security-tightening change; documented.
- **Coverage gaps** filled: registry env-parser parity, resolver cache-TTL-expiry + DB-failure-fallback + serve-last-known-good, the transactional batch, status create/edit character- and poll-limit enforcement at the route level, and the form hook's save-error + baseline-adoption paths.

Correctness reviewer confirmed the resolver precedence/cache, `domainPolicy` async refactor, instance-API wiring, status-limit off-by-one checks, and AGENTS/REVIEW invariants are correct.

Full suite: 703 files / 6717 tests passing.

PR #1299, opened 2026-07-24
fix: make the configured post, poll and upload limits take effect everywhere, and unbreak poll creation

llun:
Closes the two documented follow-ups from #1297 (database-backed admin server
settings) — and, in the course of doing so, three rounds of sub-agent review
turned up a set of related defects, including one complete outage. The through
line is the same throughout: **an admin-editable setting should actually take
effect, and the client should agree with the server about it.**

## What #1297 left unfinished

**1. The composer's character counter ignored `posts.maxCharacters`.**
`post-box.tsx` hardcoded `MAX_STATUS_LENGTH = 500`. Raising the admin post-size
setting changed what the instance API advertised to Mastodon clients but not what
the web composer counted down from or let you submit; lowering it let you build a
draft the server would reject.

The resolved values are now published once from the `(timeline)` layout through a
client `InstanceLimitsProvider` and read with `useInstanceLimits()`. That avoids
threading a prop through every surface that renders the shared inline composer,
and it keeps the settings registry — which carries `process.env` closures — out of
the client bundle: the 500 fallback moved to the dependency-free
`lib/services/mastodon/constants` module the registry already imports its sibling
defaults from.

**2. `media.maxFileSize` was not enforced.** The cap lived in a **synchronous**
Zod `.refine` reading `getConfig().mediaStorage?.maxFileSize ?? MAX_FILE_SIZE`,
which got the database-backed setting wrong in both directions: a cap an admin
*lowered* (no env var) was never applied, and one they *raised* was rejected
anyway while `/api/v[12]/instance` advertised the higher number. Resolving it
needs a database read, which a refine cannot do, so the check moved into the async
handlers — **all** of them, not just the two routes named in the follow-up, since
removing the refine would otherwise have silently dropped enforcement elsewhere:

| Path | Checked |
| --- | --- |
| `POST /api/v1/media`, `POST /api/v2/media` | file + optional thumbnail (shared `handleSyncMediaUpload`) |
| `PUT`/`PATCH /api/v1/media/:id` | replacement thumbnail |
| `POST /api/v1/medias/presigned` | declared byte size |
| `PATCH /api/v1/accounts/update_credentials` | avatar / header |
| `POST /api/v1/admin/custom_emojis` | emoji image |

Response shape is unchanged — all of these already answered `422`.

## What review found

**The premise of (1) was false.** `createNote`/`createPoll` post to
`/api/v1/accounts/outbox`, whose schema is a bare `message: z.string()` and which
never called `validateStatusContentLimits`. Only `/api/v1/statuses` enforced the
limits, so of the composer's two buttons only **Update** was covered — a create
could exceed `posts.maxCharacters` unconditionally, from the browser or from curl.
Now enforced there too, for notes and polls.

**Poll creation was broken outright — 400 for every duration.** Two stacked bugs,
found only by driving a real browser:

- `CreatePollRequest` validated `durationInSeconds` against
  `Object.keys(...).map(parseInt)`. `Array#map` passes the index as the radix, so
  the list evaluated to `[300, NaN, NaN, 7, NaN, NaN, 17, 298]`.
- More seriously, it imported that table from `poll-choices.tsx`, a **`'use
  client'` module**. Imported from server code that resolves to a *client
  reference*, not the object, so `Object.keys()` returned `[]` and nothing
  validated. **No unit test can see this** — Vitest has no RSC boundary, so the
  import returns the real module there and the suite passes either way.

Durations moved to `lib/services/statuses/pollDurations.ts`, and
`lib/clientModuleBoundary.test.ts` now guards the bug class across `app/api/`,
`lib/services|actions|jobs|database|config` (types exempt — they're erased).
Verified all eight durations return 200 against a local instance.

**Enforcing poll limits was itself a regression** until the editor was bound to
them: it offered a 5th choice against a default of 4, no cap on option length
against a default of 50, and every duration regardless of the configured range —
all of which previously succeeded, because nothing validated them. The editor now
reads the resolved `polls.*` and can only build what the endpoint accepts.

**Poll drafts leaked between composers.** `DEFAULT_CHOICES`/`DEFAULT_STATE` were
module-level objects, and the poll editor's uncontrolled inputs write straight
into them. Posting a poll with options "SECRET-A"/"SECRET-B" left them pre-filled
in the next poll opened *anywhere on the page*, one click from being republished.
Replaced with state factories.

**The upload cap could exceed what the storage driver serves.**
`S3StorageFile`'s `getFile` bounds a read-back object by the env/200 MiB value the
database cannot influence; before this PR the upload refine used the same
expression so they could never disagree. Raising the cap above 200 MiB would have
stored media that 500s on every read. `media.maxFileSize` is now bounded by that
ceiling, following the precedent of `posts.maxMediaAttachments`.

**Smaller, all from review:** stale `allowPost` when a limit changed under an open
draft (silently dead Post button); the poll branch not clearing the draft, so the
re-armed button posted a duplicate plain note; `lib/client.ts` discarding the
server's `{ error }` so every new rejection read "Fail to create a new note";
`update_credentials` writing the avatar to storage before 422-ing on the header;
`ImageUploadField` and the composer's media picker checking the wrong size (and
the pre-resize file); the duration effect claiming "closest" while taking the
shortest; multi-select rejections overwriting each other.

## Documentation

`AGENTS.md`, `REVIEW.md`, and `CLAUDE.md` gain the two conventions this PR
establishes — **Server/Client Module Boundary** and **Instance Limits in Client
Components** — and the enforcement list names the new guard test.
`docs/environment-variables.md` documents what is enforced where, including that
`posts.maxMediaAttachments` is advertised but **not** enforced by either create
endpoint.

## Testing

`yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test`, in that
order, all green: **708 test files / 6794 tests**.

New coverage includes the instance-limits provider and its fallbacks; the
composer counter across default/raised/lowered limits including the destructive
styling, the edit path, and the attachment call sites (mutation testing showed a
hardcoded 500 in those survived the original suite); the reply box counter; the
poll editor's option/length/duration bounds; the poll-draft leak; the outbox
limit enforcement; every poll duration the picker offers; the resolved upload cap
at each of the five upload paths and at the boundary; and the client/server module
boundary guard (verified to fail when the import is pointed back at the client
module).

## Browser verification

Local SQLite instance driven through the real admin UI:

1. No stored setting → counter reads **500**.
2. Post size **1200** → counter reads 1200; a 1052-character status posts
   end-to-end (previously `-552` and Post disabled). The inline **quote** composer
   reports the same limit, so the shared `InlineStatusComposer` → `PostBox` path
   picks it up.
3. Post size **200**, 207-character draft → **-7** in `text-destructive`, Post
   disabled.
4. Poll editor stops at **4 choices** with `maxlength=50`; all eight durations
   `POST` 200; posting a poll clears the draft and the next poll opens with
   **empty** choices.

Screenshots for 2 and 3 are in the review thread.

## Notes for the reviewer

- Prefix is `fix:`: no new endpoint, page, or config key — every setting here
  shipped in #1297. Worth a release note, though: an instance with an
  admin-lowered `media.maxFileSize` will start rejecting uploads it previously
  accepted, and the admin form now refuses values above 200 MiB.
- `.env.example` was not updated — its `ACTIVITIES_MEDIA_STORAGE_MAX_FILE_SIZE`
  line states the default, which has not changed. Worth a second look if you'd
  like the pin/bound behaviour mentioned there too.
- Known gap, deliberately not fixed here: a poll draft already over a *lowered*
  `polls.maxOptions` is not reconciled — the endpoint rejects it with a specific,
  actionable message rather than the editor pre-empting it.
- `lib/clientModuleBoundary.test.ts` checks direct imports only. A client module
  reached through a non-client barrel would slip past it; catching that needs
  real module-graph resolution, and the limitation is documented in the file.


llun:

Four sub-agents reviewed the full diff in parallel (server-side upload enforcement; the client composer/context; project invariants per `AGENTS.md` + `REVIEW.md`; test quality, including mutation testing). Findings and what changed:

### Confirmed defects — fixed

**1. The premise of the composer change was wrong: the create path was never enforced.** (found independently by two agents)
`createNote`/`createPoll` in `lib/client.ts` POST to `/api/v1/accounts/outbox`, whose schema is a bare `message: z.string()` and which never called `validateStatusContentLimits`. Only `POST /api/v1/statuses` and `PUT /api/v1/statuses/:id` enforced the resolved limits — so of the composer's two buttons, only **Update** was covered. A create could exceed `posts.maxCharacters` unconditionally, from the browser or from curl. The comments (and the original PR description) claiming "server-side enforcement already exists, this is client UX only" were false.
→ `240f0180` applies the check on that route for both the note and poll branches, and corrects the comments.

**2. The upload cap could exceed what the storage driver will serve.**
`S3FileStorage.getFile` bounds the object it buffers back out by `mediaStorage.maxFileSize ?? MAX_FILE_SIZE` — env or 200 MB, which the database cannot influence. Before this PR the upload refine used that same expression, so the two could never disagree; afterwards an admin could raise the cap above 200 MB, have the upload stored, and have every subsequent read throw `StreamByteLimitError` → a permanent 500 on that media. Object storage without a CDN hostname only.
→ `68396b82` bounds the setting rather than the read: `media.maxFileSize` validates at most `MAX_FILE_SIZE`, following the precedent of `posts.maxMediaAttachments`. Lowering the cap is unaffected; an env-pinned value still applies because the storage driver reads the same variable.

**3. `StatusReplyBox` had no counter or length check at all.** With #1 fixed, an over-long reply would have failed only on submit — and two composers on one screen would have disagreed about the limit.
→ `240f0180` gives it the same counter and resolved limit.

**4. Stale `allowPost` when the limit changes under an open draft.** `allowPost` was only recomputed by the draft-editing handlers, so a layout re-render carrying a new limit left the submit button stale — enabled for a draft the server now rejects, whose click hit the guard in `onPost` and silently did nothing (no message, no state change).
→ `240f0180` recomputes it in both composers.

**5. `update_credentials` orphaned the avatar on an invalid header.** Validating and saving in one pass wrote the avatar to storage (and against the quota) before 422-ing on the header, with the actor never updated. Pre-existing shape, but the new check preserved it.
→ `68396b82` validates both files before saving either.

**6. `ImageUploadField` still hardcoded 200 MB**, telling the user "Maximum size is 200MB" no matter the configured cap — so a lowered cap surfaced as a generic upload failure. I had flagged this as out of scope; on reflection it is the same bug this PR exists to fix, one field over.
→ `68396b82` reads the resolved value via `maxMediaFileSize` on the same context.

**7. Coverage gaps proven by mutation testing.** Hardcoding `500` back into `isEditSubmittable` and into the four non-text `hasNewPostContent` call sites left the entire suite green, as did reverting the over-limit `text-destructive` styling. The presigned schema's new `nonnegative()` was untested.
→ `68396b82` adds tests that fail under each of those mutations.

**8. Convention:** the v2 media size cases were two near-identical `it` blocks where `AGENTS.md` → Testing Guidelines requires a table-driven `it.each` — inconsistent with the sibling v1 file in the same PR. Stale `{ size, type }` example in a `FileSchema` comment. → fixed in `68396b82`.

### Found while fixing — separate commit

`04b97a3d`: `CreatePollRequest` validated `durationInSeconds` with `Object.keys(...).map(parseInt)`, which passes the array index as the radix. The list evaluated to `[300, NaN, NaN, 7, NaN, NaN, 17, 298]`, so **every poll duration the composer offers except "5 minutes" was rejected with a 400.** Pre-existing and unrelated to the two follow-ups, so it is isolated in its own commit with a regression test over every duration the picker offers.

### Raised and deliberately not acted on

- **Settings-read failure is fail-open for up to 2s.** When the database read fails and there is no cached view, the resolver serves env+defaults, so an admin-lowered cap is briefly not applied. This is the resolver's documented, codebase-wide trade-off (#1297), not something for this PR to special-case — and the database being down means the subsequent `createMedia` fails anyway.
- **The new stored-setting route tests do not neutralise `ACTIVITIES_MEDIA_STORAGE_MAX_FILE_SIZE`.** An env value would win and lock the field. Vitest does not load `.env.local`, so CI is unaffected, and a developer with it exported sees the tests *fail loudly* rather than pass on the fallback path — there is no false-green risk. I added the save/delete/restore guard to `lib/services/medias/uploadSizeLimit.test.ts` (the test that is specifically about resolution, mirroring `lib/services/serverSettings/index.test.ts`) and left the route tests alone rather than duplicating it five times.
- **"Links always count as 23"** in the admin help text, while both the composer and `validateStatusContentLimits` count raw characters. Pre-existing and a server-side semantics question (Mastodon normalises URL length); out of scope here.

Gate re-run after all of the above: prettier → lint → build → test, **706 test files / 6777 tests green**.

llun:
## Sub-agent code review — rounds 2 and 3

Two more rounds, two sub-agents each, reviewing the previous round's fixes rather than re-deriving settled findings. Both rounds found real regressions introduced by the round before, which is the point of running them.

### Round 2 — the round-1 fix was a regression

**1. Enforcing poll limits on the outbox route broke poll creation that previously worked.** (both agents, independently) The composer's poll editor was driven by hardcoded constants that don't match the registry defaults: it offered a 5th choice against `polls.maxOptions: 4`, no cap on option length against 50, and every duration regardless of the configured range. All three succeeded before, because the route validated none of it — so round 1 turned working flows into 422s.
→ `6fa91fc1` binds the editor to the resolved `polls.*` values, so it can only build what the endpoint accepts.

**2. The new `allowPost` effect re-armed Post after a poll create, with the question still in the textarea** — a second click posted it again as a plain note. The poll branch never cleared the draft; keying the effect on `isPosting` made that reachable. Confirmed by mutation testing (the probe passes on the branch, fails against `main`).
→ Fixed by clearing the draft like the note branch does.

**3. Every rejection this PR introduces surfaced as a generic message.** `lib/client.ts` discarded the server's `{ error }`, so an admin-configured limit read as "Fail to create a new note". This matters more than pre-PR because a tab holds the limit its layout resolved and can legitimately disagree with the server.
→ `throwApiError` surfaces the specific message; `StatusReplyBox` shows it inline instead of an `alert`.

**4.** `ImageUploadField` checked the **pre-resize** size, so a lowered cap rejected photos that resizing would have brought under it. The composer's media picker had no size check at all. Both fixed; the layout comment that claimed otherwise corrected.

**5.** Doc claims corrected: `media.maxFileSize` is not the only bounded setting, and `posts.maxMediaAttachments` is advertised but not enforced.

### Round 3 — the round-2 fix had its own problems

**1. Poll drafts leaked between composers.** `DEFAULT_CHOICES`/`DEFAULT_STATE` are module-level objects and the poll editor's inputs are uncontrolled, writing straight into them. Post a poll with options "SECRET-A"/"SECRET-B", open a poll anywhere else on the page, and they're pre-filled — one click from republication. Pre-existing, but it's the one part of the draft round 2's "clear after posting" fix didn't reach.
→ `cdcc0f8a` replaces them with state factories, with a regression test. Verified in a browser: reopening a poll after posting now shows empty choices.

**2. The `'use client'` fix put the shared duration table in `lib/components/post-box/poll-durations.ts`** — leaving an API route as the only importer of `lib/components` anywhere under `app/api/`, re-creating one directory over the exact adjacency that caused the outage.
→ Moved to `lib/services/statuses/pollDurations.ts`. The guard test moved with it and widened: it now covers `lib/services|actions|jobs|database|config` too, resolves relative imports, and exempts `import type`. Verified it still fails when the import is pointed back at the client module.

**3. AGENTS.md requires new conventions to be documented, and rounds 1–2 didn't.** Two were established without prose rules: the client-module boundary and the instance-limits context.
→ Both now have sections in `AGENTS.md`, entries in `REVIEW.md`, and reminders in `CLAUDE.md`; the enforcement list in `AGENTS.md` names the new guard.

**4.** The out-of-range duration effect claimed to pick "the closest offered one" but took `durations[0]` — with a range bounded from above, a 1-day selection silently became 5 minutes. Now picks the nearest, tested in both directions.

**5.** The media picker called `setWarningMsg` once per rejected file, so a multi-select showed only the last name — reading as "the others uploaded". Rejections are aggregated.

**6.** More doc corrections: not every 422 names the limit (only the status routes do); the client *can* still offer something the endpoint rejects (the duration picker deliberately degrades when a configured range excludes every labelled duration); the `posts.maxCharacters` bound is a sanity bound, not stack-honouring; MB vs MiB made consistent.

**7.** The `parseInt` comment claimed the composer "could only create 5-minute polls" — in production the client-module import made the list empty, so no duration validated at all. Corrected in source and called out here because that commit message is otherwise misleading in `main`'s history after squash.

### Raised and deliberately not acted on

- **A poll draft already over a *lowered* `polls.maxOptions` isn't reconciled.** The editor hides "Add choice" but keeps the existing over-limit choices and an enabled Post button. Narrow race (admin lowers the limit under an open draft), and thanks to the error-surfacing fix the 422 now says exactly what's wrong — building client-side poll validation for it isn't worth the surface area. Noted in the PR description.
- **Upload 422s return the generic `Unprocessable entity`** rather than naming the byte limit. Changing six endpoints' error bodies is more churn than it's worth here, and `uploadAttachment` discards the body anyway. The doc now states this accurately instead of overclaiming.
- **`lib/components/post-box/upload-media-button.tsx` imports the logger** (REVIEW.md forbids logging from client code) and has no `'use client'` directive despite now calling a hook. Both pre-existing and safe today (it's only imported from client components); out of scope.
- **A lint rule instead of the guard test.** ESLint can express "nothing under `app/api/` imports `@/lib/components`", but that's a proxy for the directory, not the `'use client'` directive — it would miss a client module elsewhere and false-positive on legitimately shared modules. ESLint can't read an imported file's first line without a custom plugin. Keeping the test, which checks the real invariant. Worth revisiting if someone writes the plugin.

### Gate

prettier → lint → build → test, green after every round. Now **708 test files / 6791 tests**.

The PR title and description were materially stale after three rounds (they still described the reply box as having no counter, called the change "client UX only", and listed `ImageUploadField` as out of scope — all since contradicted by the diff). Both have been rewritten.

llun:
## Sub-agent code review — rounds 4–7 (loop converged)

### Round 4

**1. The guard test I added in round 3 had a hole in the exact bug it exists to prevent.** It matched only `import … from`, so the same outage could land as a re-export — `export { SecondsToDurationText } from '@/lib/components/post-box/poll-choices'` inside a server module resolves through the same client reference, empties the table, and 400s every poll, with the test green. The form is idiomatic here (four `export … from` lines already sit inside the scanned roots). It also false-positived on inline `import { type X }`.

**2. `POST /api/v1/accounts/outbox` accepted polls with blank options and fewer than two**, while `POST /api/v1/statuses` rejected both. Since this PR is the change that made the two endpoints agree on the admin-configured bounds, the structural floor should agree too — `CreatePollRequest` now carries the same constraints as `PollSchema`. The composer already refused blank choices client-side, so no working flow changes.

**3. Poll choice keys came from `Math.round(Math.random() * 1000)`** — a ~1/1000 collision that reconciles two uncontrolled inputs onto one DOM node and silently drops an option. Harmless when the default pair was minted once per module load; this PR mints it on every reset. Now monotonic.

**4. Doc/comment accuracy**, all overclaims from my own earlier rounds: `update_credentials` returns `Invalid image file`, not the generic `Unprocessable entity`; the poll-expiry message names no bound; the `useInstanceLimits()` consumer list omitted the reply box in `AGENTS.md` and `CLAUDE.md` while `docs/` listed it; `CLAUDE.md` forbade importing `lib/config/serverSettings` into a Client Component without the "for its value exports" qualifier, outlawing the `import type` four admin forms already do.

### Round 5

**The regex could still be defeated, in the direction that matters.** Its clause has to span newlines, and since this project writes no semicolons a `from`-less `export type { … }` block merges with the *next* statement into one match — whose clause starts with `type`, so the exemption fires and a genuine value import is never reported. Replaced with a `ts.createSourceFile` parse. Exercised against value import, `export … from`, `export * from`, `import type`, inline `{ type X }`, `export type … from`, and the merge case: each behaves as intended.

### Round 6 — clean, with one note

Verified the AST rewrite and found one residual blind spot: it always parsed as `ScriptKind.TSX`, and a `.ts` file containing a generic arrow (`<T>(x: T) => x`) misparses under TSX — the arrow becomes JSX and every statement after it is dropped, across 32 files. Nothing escaped today (imports precede that syntax everywhere), but a guard for an invisible bug should not itself have a blind spot. ScriptKind now follows the extension; verified an import appended to the bottom of `lib/config/serverSettings.ts` is now caught and was not before.

### Round 7 — clean

No findings. Independently re-verified the two newest commits (catching both a bottom-of-file value import and a re-export while type-only forms stay exempt) and did a fresh whole-diff pass focused on what an operator upgrading an existing instance experiences.

### Upgrade impact, for the record

- An instance with an admin-lowered `media.maxFileSize` **starts rejecting uploads it previously accepted** — that is the bug being fixed, but it is a visible behaviour change worth a release note.
- A stored `media.maxFileSize` above 200 MiB is now ignored in favour of the default. Those uploads could never be read back out of object storage anyway.
- The newly-enforced limits on `POST /api/v1/accounts/outbox` only tighten an internal endpoint — and its poll path was returning 400 for *every* duration before this PR, so nothing that worked stops working.

### Loop summary

7 rounds, 15 sub-agents, 10 commits. Rounds 1–3 each found a real regression introduced by the round before, which is the argument for running the loop rather than declaring victory after one pass. Two of the most serious findings — poll creation being broken outright, and poll drafts leaking between composers — were only reachable by driving a real browser: the `'use client'` import returns the real module under Vitest, so no unit test could have seen it.

Final gate: prettier → lint → build → test, green. **708 test files / 6794 tests.**

PR #1300, opened 2026-07-24
fix: align the admin sub-nav dropdown with the design system

llun:
## Summary

The admin sub-nav dropdown split its three server-settings tabs off behind a `DropdownMenuSeparator` + a **SETTINGS** heading. Every entry in that menu is an admin setting, so carving out a "Settings" group only implied the other tabs were something else — and the design system's sub-nav has no group headings or separators at all (`ui_kits/web/SettingsKit.jsx` → `NavDropdown` renders one flat run of items; `ui_kits/web/AdminPage.jsx` → `ADMIN_TABS` is a flat array).

The grouping was introduced with the DB-backed server settings in #1297; `AGENTS.md` has always documented the tab shape as `{ name, url, icon }`, so this puts the code back in line with both the docs and the design.

## What changed

**`lib/components/section-nav-dropdown.tsx`**
- Removed the optional `group` field from `SectionNavTab` and the separator/label rendering it drove. The menu is now one flat run of `DropdownMenuItem` links. `admin` was the only caller that used it.

**`app/(timeline)/admin/layout.tsx`** — reordered to the design's `ADMIN_TABS` and adopted its labels and icons:

| | Before | After |
|---|---|---|
| Order | Overview, Accounts, Reports, Hashtags, Filters, Rules, Announcements, Federation, Relays, Custom emojis, **─ SETTINGS ─**, Instance, Posts & media, Network, System | Overview, Accounts, Reports, Server rules, Hashtags, Announcements, Filters, Custom emojis, Federation, Relays, Posts & media, Network, Instance, System |
| Rules | `Rules` | `Server rules` — matches the design *and* the page's own `PageHeader` title |
| Accounts icon | `Users` (two people) | `User` (single person) |
| Reports icon | `Flag` | `AlertTriangle` |
| Overview icon | `BarChart3` | `BarChart` |
| Network icon | `Network` | `Link` (chain) |
| Instance icon | `Server` | `Settings` (gear) |
| System icon | `Settings` | `Server` (freed by Instance) |

The design has no entry for Hashtags, Custom emojis, Relays or System, so each sits next to the tab it belongs with while the design's own relative order is preserved verbatim. Its Trends and Invites entries have no page in the app yet and were not added.

## Tests

`lib/components/section-nav-dropdown.test.tsx` — the group-heading test is replaced by a regression guard asserting the open menu contains no `separator` role, exactly one `menuitem` per tab, and no text beyond the tab names.

## Verification

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (708 files, 6794 tests) all green.
- Verified in a browser against a local SQLite instance signed in as an admin: `/admin` opens a flat 14-item menu with no separator or heading, and `/admin/instance` resolves the active tab correctly (gear + "Instance" in the trigger).

## Regression guard

The flat menu is enforced by `renders the menu as one flat run even for tabs carrying a group` in `lib/components/section-nav-dropdown.test.tsx`. It feeds the menu tabs that carry a `group` — cast with `as SectionNavTab[]`, since the field no longer exists on the type — and asserts no `separator` role and no `Settings` heading. The cast is deliberate: a test using only well-typed tabs passes against the grouping implementation too, and so guards nothing.

To confirm the guard actually bites:

```bash
git checkout main -- lib/components/section-nav-dropdown.tsx
yarn vitest run lib/components/section-nav-dropdown.test.tsx   # fails on the separator assertion
git checkout HEAD -- lib/components/section-nav-dropdown.tsx   # restore
```

Note: the `05e3094` commit body says "verified it fails against `HEAD~1`", which was accurate when written but went stale once that commit landed. Use `main` as above.


lib/components/section-nav-dropdown.test.tsx
```diff
@@ -75,9 +75,18 @@ describe('SectionNavDropdown', () => {
...
+  // The design system's sub-nav is one flat run of links. A `group` field once
+  // split admin's server-settings tabs off behind a separator + heading, which
+  // read as though the other tabs weren't settings.
+  //
+  // Feed the menu the exact shape that used to trigger that split — tabs
+  // carrying a `group` — and assert it still renders flat. The cast is
+  // deliberate and load-bearing: `group` is no longer part of `SectionNavTab`,
+  // and a test using only well-typed tabs would pass against the grouping
+  // implementation too, guarding nothing.
+  it('renders the menu as one flat run even for tabs carrying a group', async () => {
```
llun, line 111: **The flat-menu guard did not guard anything.** As first written this test rendered the module-level `tabs` fixture (the fitness tabs at lines 18-23), none of which ever carried a `group`. Against the previous implementation `startsGroup` was therefore always `false`, so it emitted a `<Fragment>` (zero DOM) around the identical `DropdownMenuItem` — the DOM was byte-identical under both implementations and every assertion held either way.

Concrete consequence: the deleted `renders a group heading before a run of grouped tabs` test was the only coverage of grouped rendering, and nothing replaced it. A future PR re-adding `group?: string` plus separator/label rendering would have been caught by no test at all.

Secondary: `expect(menu.textContent).toBe(...)` was exact whole-menu string equality, so it would also fail if anyone added legitimate `sr-only` text to a menu item — a failure mode unrelated to flatness.

app/(timeline)/admin/layout.tsx
```diff
@@ -1,20 +1,20 @@
 'use client'
 
 import {
-  BarChart3,
+  AlertTriangle,
```
llun, line 4: **Durability note (non-blocking, no change made).** Verified against the installed `lucide-react@1.25.0`: all 14 icon names resolve and render. Two of the new ones are back-compat aliases lucide keeps pointing at renamed icons — `AlertTriangle` → `TriangleAlert` and `BarChart` → `ChartNoAxesColumnIncreasing` (`Filter`, pre-existing here, is likewise an alias of `Funnel`). The net effect of this diff is that `Flag` (a canonical name) becomes `AlertTriangle` (an alias), so the file gains one deprecated-alias dependency. Nothing breaks today; the risk is a future lucide major dropping the aliases.

Worth stating explicitly for the record: `BarChart` is not a cosmetic no-op swap for `BarChart3`. `BarChart` → `ChartNoAxesColumnIncreasing` has no axis; `BarChart3` → `ChartColumn` does. The axis-less form is what the design specifies.

llun (commented):
Sub-agent review round 1, per AGENTS.md → Code Review Loop. One substantive finding on the test plus one durability note on icon naming; both addressed below.

Categories that came back clean: no leftover `group` references at any of the four `SectionNavDropdown` call sites (settings, fitness, account, admin — `FitnessStatusDetail` has its own local `SectionTab` type and never had the field); no unused or missing imports in any of the three files, and `DropdownMenuLabel`/`DropdownMenuSeparator` are still consumed by four other components so no dead exports are left behind; admin routes and nav entries are an exact 1:1 (13 subdirectories + the root page = 14 routes, 14 tabs, no orphan page and no dead link, with `accounts/[id]`, `reports/[id]` and `tags/[tag]` resolving to their parent tab via the prefix rule); active-tab resolution is order-independent (it filters to prefix matches then takes the longest URL, and `.sort()` runs on the `.filter()` copy so the module-level `tabs` const is not mutated), so the reorder cannot break it; and no Markdown or `docs/` file is made stale — nothing ever documented the `group` field or a "Settings" group, AGENTS.md:184 already described tabs as `{ name, url, icon }`, and no doc names a "Rules" tab.

lib/components/section-nav-dropdown.test.tsx
llun: Addressed in 05e3094.

The test now feeds the menu the exact shape that used to trigger the split — tabs carrying a `group`, cast in with `as SectionNavTab[]` since the field is gone from the type — and asserts the menu stays flat: no `separator` role, no `Settings` heading text, one `menuitem` per tab.

Proved the guard bites rather than assuming it: with the test at its new form and the component checked out at `HEAD~1` (the grouping implementation), it **fails** on `expect(within(menu).queryByRole("separator")).not.toBeInTheDocument()` — `data-slot="dropdown-menu-separator"` is present. Restored to the flat implementation, all 5 pass. That is the property the old version could not distinguish.

The cast is called out in a comment as deliberate and load-bearing, precisely so nobody "fixes" it back into well-typed tabs and silently re-neuters the test. Also dropped the exact whole-menu `textContent` equality per the secondary point.

app/(timeline)/admin/layout.tsx
llun: Acknowledged, keeping the alias names — deliberately, not by oversight.

`AlertTriangle` is already the name this repo uses in `lib/components/error-page.tsx` and `app/(timeline)/settings/featured-hashtags/FeaturedTagsEditor.tsx`, so it is not a new alias dependency for the codebase, only for this file. Switching just these two to canonical names while `Filter` stays an alias here (and `AlertTriangle` stays one elsewhere) would trade a theoretical future break for actual present inconsistency. If lucide ever drops the back-compat aliases it is a repo-wide migration, better done in one pass than piecemeal.

Confirming the second half explicitly: the axis-less `BarChart` is intended. The design specifies three plain bars with no axis, and the local browser check shows exactly that in the trigger and the Overview row.

llun:
## Sub-agent review round 2

Reviewed the round-1 fix (05e3094) and the full diff again. **No actionable code findings.** One inaccuracy in the permanent record, corrected below.

### The round-1 fix is genuinely a guard, verified independently

Round 2 re-ran the falsification rather than trusting my claim: component checked out at `main`, test kept at HEAD →

```
× renders the menu as one flat run even for tabs carrying a group
  expected document not to contain element, found <div … role="separator" …/>
Tests  1 failed | 4 passed (5)
```

Restored to HEAD → 5 passed. The tautology is fixed.

### The `as SectionNavTab[]` cast does not mask anything

Confirmed with an isolated `tsc --strict` run: the fixture with its extra `group` field compiles clean, while a deliberately wrong shape (`url: 42`) is still rejected with `TS2352 … neither type sufficiently overlaps`. It suppresses only object-literal excess-property freshness — a widening cast, not the type-erasing `as unknown as` escape hatch — so a genuine mismatch would still be caught. The "CI does not typecheck tests" caveat is therefore moot here.

A full `tsc --noEmit --ignoreDeprecations 6.0` reports zero errors in all three changed files. That also structurally re-proves round 1's "no leftover `group`" result: an excess `group` in any `SectionNavTab[]` literal would now be a compile error, and there is none anywhere.

### Assertion precision, for the record

`queryByRole("separator")` and `queryByText("Settings")` are each independently load-bearing — the second catches a heading re-added *without* a separator, and is correctly keyed since the fixture's `group` value is exactly `"Settings"`. The `toHaveLength(3)` assertion passes under both implementations (`DropdownMenuLabel` renders a plain div with no `menuitem` role), so it is a completeness check that grouped tabs are not dropped, **not** a flatness guard. Worth being precise about which assertion does the work.

Tests 1-4 and the shared fixture are byte-identical to `main`; the three consumer layout suites (settings, fitness, account) plus this file pass together, 26 tests.

### Correction: the 05e3094 commit message is wrong about `HEAD~1`

The body says *"Verified it fails against HEAD~1 on the `separator` role."* That was true when written — at that moment `HEAD` was f87d5dd7 and `HEAD~1` was `main`. Once the test-fix commit landed on top it went stale: `HEAD~1` is now the fix commit, which already contains the flat component, so the test **passes** there.

The correct reproduction is against `main`:

```bash
git checkout main -- lib/components/section-nav-dropdown.tsx
yarn vitest run lib/components/section-nav-dropdown.test.tsx   # fails on the separator assertion
git checkout HEAD -- lib/components/section-nav-dropdown.tsx   # restore
```

Left the commit message alone rather than force-pushing a branch that is already open for review; this comment and the PR description carry the accurate command instead. Flagging it explicitly because a reader who followed the stale instruction would see the test pass and conclude the guard is a tautology — exactly the failure round 1 existed to catch.

### Round 2 verdict

Nothing further to address in the code. Per AGENTS.md the loop stops when a full round surfaces no new actionable issues; a round 3 would review an unchanged diff.

llun:
## Sub-agent review round 3 — chrome restyle

Two blocking findings, both confirmed and fixed in 0a17378; one finding declined with numbers.

### Fixed — focus indicator vanished on the active row (WCAG 2.4.7)

Real, and self-inflicted by the previous commit. Pinning `focus:bg-primary/10 focus:text-primary` so the current row keeps its orange on hover also made focused and resting render **identically**: the shared item base sets `outline-hidden`, the only global outline rule sets `outline-color` with no style or width, nothing in the repo styles `data-highlighted`, and every row is now `font-medium` so even the old weight difference is gone. A keyboard user arrowing down the menu watched the highlight travel and then disappear on exactly one row, unable to tell "highlighted" from "menu not focused" before pressing Enter.

Worth noting I had measured this myself earlier and mis-read it as a success — I recorded that focused and resting were identical and called it correct hover behaviour, without asking what it meant for keyboard focus.

Fix: a 2px `ring-primary/50` on the active row only; inactive rows keep the shared grey focus wash and need nothing. Verified the focused active row now reports a 2px ring where resting reports `box-shadow: none`. Guarded by a new test — jsdom has no Tailwind, so it asserts the utility is present rather than a computed style.

### Fixed — cross-`cn()` override, and the `text-current` no-op

Correct that `focus:bg-primary/10` was beating `focus:bg-accent` only on Tailwind emit order: split across the item and its child `Link`, Radix's `Slot` string-joins the two class attributes rather than tailwind-merging them. True today, not a property worth depending on — the demonstration that `shadow-lg` happens to be emitted *before* `shadow-md` makes the hazard concrete. All active-state classes now live in the same `cn()` call as the item base.

Also took the `text-current` point. Both forms opt the icon out of the shared rule that greys un-coloured menu icons (it substring-tests the class attribute for `text-`), but `text-current` is a semantic no-op that reads as dead code and invites deletion — which would silently grey every icon again with no test failing. The colour is now named outright.

### Fixed — the un-updated mirror

`FitnessStatusDetail`'s `SectionNav` carries a comment saying it mirrors `SectionNavDropdown`. It has been given the same chrome, ring and icon colours; otherwise the activity detail page would have been the one section dropdown left on the old styling, which is exactly the drift that comment promises against.

### Declined, with numbers — the active row's contrast

Independently recomputed and the arithmetic holds: `--primary` `hsl(24 95% 46%)` on its own 10% wash over white is **3.12:1**, under AA's 4.5:1 for 14px/500 text. Dark theme is 4.58:1, passing by 0.08. And `app/globals.contrast.test.ts` genuinely does not cover this — it only pairs `--muted-foreground` against opaque surface tokens, never compositing an alpha wash and never testing `--primary` as a foreground.

Not fixing it here, because it is not this menu's problem: `text-primary` on **plain white** is 3.52:1, so the pairing fails anywhere it appears in the app, including the sidebar's active item. The suggested light-theme-only `hsl(24 95% 36%)` does clear it at 4.78:1 — but that is a design-token change that deviates from the orange this design system specifies, across every surface, and it deserves a deliberate decision rather than being smuggled into a PR about a dropdown. Flagging it as its own piece of work.

One correction to the finding's framing: the commit did not create the failure, and the pre-change state was not passing either — the high-contrast grey only appeared on hover/focus, so the row was already 3.12:1 at rest, which is how it renders almost all of the time.

Also fixed the misleading comment that called the orange wash "a non-color-dependent cue". It is a second colour cue, and a faint one at 1.13:1 against the popover; `aria-current` is the actual non-visual cue.

PR #1301, opened 2026-07-24
fix: match the design system for the admin post size field

llun:
## Summary

Admin → **Posts & media** → **Post size** rendered a bare number input, but the design system specifies a **preset dropdown** with a `Custom…` escape hatch. This aligns the app with the kit.

Source of truth: `ui_kits/web/AdminSettingsKit.jsx` → `PostsSection` in the "Activities.next Design System" project.

| | Before | After |
| --- | --- | --- |
| Control | number input + `characters` suffix | select: `500 characters — Mastodon default` / `1,000 characters` / `5,000 characters` / `Custom…` |
| Custom value | always typed directly | number input revealed beside the select only when `Custom…` is picked |
| Layout | single input | `grid gap-3 sm:grid-cols-[1fr_auto]`, per the kit |

## Behaviour

- Picking a preset writes that value immediately; the help line (`New posts and edits are capped at N characters.`) tracks it.
- The mode is derived from the stored value, so a value that is **not** a preset (e.g. `750`, or anything set through the API) opens on `Custom…` with the input pre-filled — nothing becomes uneditable and no stored value is silently rewritten.
- `Custom…` is sticky once chosen, so typing a number that happens to equal a preset does not snap the select back mid-edit.
- The revealed input gets `aria-label="Custom post size"` because the field label points at the select; `NumberField` grew an optional `ariaLabel` prop for that.

## Verification

Local dev server (SQLite `dev.sqlite3`, `localhost:3042`), signed in as an admin:

- Default state shows `500 characters — Mastodon default`.
- Selecting `Custom…` reveals the number input.
- Saving `750` persists (`server_settings.posts.maxCharacters = 750`) and is advertised as `configuration.statuses.max_characters = 750` on `/api/v2/instance`.
- Reloading returns to `Custom…` with `750` pre-filled.

Gate: `prettier --write .` → `lint` (0 errors) → `build` (exit 0) → `test` (708 files / 6798 tests passed).

## Note, not changed here

**Media per post** in the same section also differs from the kit — the design shows a select (`4 / 8 / 16 attachments`), the app has a number input. Left alone deliberately: the kit's fixed list has no `Custom…` option, so adopting it as-is would make the app's supported 1–20 range (default 20) unreachable. Happy to follow up if you want it converted with the same preset + `Custom…` pattern.


lib/components/admin/settings/PostsMediaSettingsForm.tsx
```diff
@@ -78,6 +91,23 @@ export const PostsMediaSettingsForm: FC<PostsMediaSettingsFormProps> = ({
   const maxCharacters = values['posts.maxCharacters'] as number
   const uploadBytes = values['media.maxFileSize'] as number
 
+  const isPresetPostSize = POST_SIZE_OPTIONS.some(
+    (option) => option.value === String(maxCharacters)
+  )
+  const postSizeMode =
+    customPostSizeSelected || !isPresetPostSize
+      ? CUSTOM_POST_SIZE
+      : String(maxCharacters)
```
llun: **Bug: the custom input unmounts mid-typing on the implicit-custom path.**

`customPostSizeSelected` is only set when the admin *clicks* `Custom…`. When custom mode is reached implicitly — the stored value is not a preset, e.g. `750` — the flag stays `false`, so the mode is recomputed from `maxCharacters` on every keystroke. Any transient prefix that equals a preset flips the select back to that preset and unmounts the input.

Repro: stored value `750`, clear the field and type toward `5000` one digit at a time:

```
5 → custom   50 → custom   500 → select flips to "500 characters — Mastodon default", input gone
```

The admin loses focus and is stranded at 500; to reach 5000 they have to re-pick `Custom…` and retype. The same happens to anyone editing an out-of-preset value downward through 500.

The design kit avoids this by keeping the mode as its own state (`postSizeMode`) rather than deriving it from the value. Deriving it here was the deviation. Fix: make the mode real state, initialised once from whether the stored value is a preset and mutated only by the select — with one reconciliation in the preset direction, since `useServerSettingsForm.saveSection` adopts server-resolved values and can change the value out from under a selected preset.

llun: Fixed in 0692fed84 — and confirmed: a second reviewer hit the same repro independently.

The mode is now its own state (`selectedPostSizeMode`), initialised once from whether the stored value is a preset and mutated only by the select, which is what the kit does. The derivation that ran on every keystroke is gone, so the custom input never unmounts mid-edit. One reconciliation remains in the preset direction only — a picked preset falls back to Custom when the value no longer matches it — because `useServerSettingsForm.saveSection` adopts the server-resolved value; that branch can only run while the number input is hidden, so it can't interfere with typing.

Two regression tests added, both of which fail against the previous commit:
- `keeps the custom input mounted while typing past a preset value` — starts from a stored `750` and walks `5 → 50 → 500 → 5000`, asserting the select stays on `custom` at every step.
- `switches back from Custom to a preset` — covers the reverse branch, which was previously a no-op in the existing preset test.

Also verified in the browser against the real dev server with `750` stored: `typed "5"/"50"/"500"/"5000" -> select=custom customInputPresent=true` at every step.

One related nit deliberately skipped: the custom input still has no `max`, even though the server schema caps `posts.maxCharacters` at 100000. `lib/config/serverSettings.ts` is server-only (env closures), so a client `max` would have to hardcode the bound, and `NumberField` doesn't clamp anyway — it forwards native hints. The server rejects out-of-range values regardless, so this stays as-is rather than duplicating the bound.

lib/components/admin/settings/PostsMediaSettingsForm.tsx
```diff
@@ -78,6 +99,19 @@ export const PostsMediaSettingsForm: FC<PostsMediaSettingsFormProps> = ({
   const maxCharacters = values['posts.maxCharacters'] as number
   const uploadBytes = values['media.maxFileSize'] as number
 
+  // A preset stays picked only while the value still matches it: saving adopts
+  // the server-resolved value, which can land somewhere else entirely.
+  const postSizeMode =
+    selectedPostSizeMode === CUSTOM_POST_SIZE ||
+    selectedPostSizeMode === String(maxCharacters)
+      ? selectedPostSizeMode
+      : CUSTOM_POST_SIZE
```
llun: **Round 2: the fallback to Custom is render-only, so a stale preset can reawaken.**

This expression corrects the mode for the current render but never writes back into `selectedPostSizeMode`, which is only ever set by the select. So an orphaned preset stays in state:

1. Pick `1,000 characters` — `selectedPostSizeMode='1000'`, value `1000`.
2. Save. `useServerSettingsForm.saveSection` adopts the server-resolved value, which can differ from what was sent (a concurrent edit by another admin). Say it lands on `500`.
3. The fallback fires: select shows `Custom…`, input reappears with `500`. Correct so far — but `selectedPostSizeMode` is still `'1000'`.
4. The admin types `1000` back. Now `selectedPostSizeMode === String(maxCharacters)`, the fallback stops applying, the select jumps to `1,000 characters` and the input unmounts **mid-edit** — the round-1 symptom, reached through the save path.

`custom` is properly sticky (first ternary arm always wins); a stale preset string is not. That asymmetry is the gap.

llun: Confirmed and fixed in 00cabd3df. I traced the scenario against the code before acting on it and it holds: `selectedPostSizeMode` was only ever written by the select, so the fallback corrected the render but left the orphaned preset in state, free to match again later.

An effect now commits the fallback whenever the value changes:

```tsx
useEffect(() => {
  setSelectedPostSizeMode((mode) =>
    mode === CUSTOM_POST_SIZE || mode === String(maxCharacters)
      ? mode
      : CUSTOM_POST_SIZE
  )
}, [maxCharacters])
```

The `custom` arm short-circuits, so it is a no-op during an edit and cannot reintroduce the round-1 bug. The render-time expression stays as well — it keeps the very first render after a value change honest, before the effect runs — and the effect makes the correction stick. Same idiom as `NumberField`'s existing re-sync effect in this directory.

Regression test added: `does not reselect a stale preset after a save resolves elsewhere` mocks a save resolving to `500` after `1000` was submitted, asserts the select falls back to Custom, then types `1000` back and asserts it stays on Custom with the input mounted. It fails against 0692fed84 and passes now.

Re-verified in the browser against the dev server with `750` stored — `select=custom customPresent=true` across `'' → 5 → 50 → 500 → 5000 → 1000`. Full gate green: lint 0 errors, build exit 0, 6801 tests passing (the fitness `RegionMapKit` flake from the previous run did not recur).

llun:
Merged current `main` into the branch: #1302 landed in the same three files (it removed the `storageBackend` prop and added blur-clamping to `NumberField`). Both components auto-merged; the only conflict was `renderForm`'s signature in the test file, resolved to keep both changes — no `storageBackend` prop, plus the `settings` override this PR added. Full suite green after the merge: 710 files / 6816 tests.

Two follow-up candidates deliberately left out of this PR:

1. **Media per post** still diverges from the kit (design: a `4 / 8 / 16` select; app: a number input). The kit's list has no `Custom…`, so adopting it as-is would make the app's 1–20 range unreachable.
2. **The custom post size field has no `max`**, while the server schema caps `posts.maxCharacters` at 100000. This was only cosmetic before, but #1302 just added clamp-on-blur to `NumberField` precisely so an out-of-range value gets clamped instead of returning an opaque 422 — so wiring a ceiling in would now actually do something. It needs the bound to live in a dependency-free module a Client Component may import, exactly as #1302 did with `MAX_CONFIGURABLE_FILE_SIZE`; `lib/config/serverSettings.ts` carries `process.env` closures and can't be imported for values.

PR #1304, opened 2026-07-24
minor: match the design system on the admin Posts & media page

llun:
## Summary

The **Media** section and the environment configuration block on **Admin → Posts & media** had drifted from the design system (`ui_kits/web/AdminSettingsKit.jsx` in [Activities.next Design System](https://claude.ai/design/p/a61009c5-7207-4e18-8d2b-a0cace8d672b?file=ui_kits%2Fweb%2Findex.html)). This brings both in line.

### Media — the storage backend is now reported

`Media` gains a read-only **Storage backend** field carrying the amber `Set by environment` badge and its own help line ("Read from the environment at boot — change it with the builder below, not here."). The backend is infrastructure read once at boot, so the admin UI *describes* it rather than offering to edit it.

`describeMediaStorageBackend` derives the one-line summary on the server (`S3 — media.example.social (eu-central-1)`, `Local filesystem — ./uploads`, or `Not configured — media uploads are unavailable`), so storage paths and endpoints never enter the client bundle. A custom endpoint is folded into the detail, which is what distinguishes R2/MinIO/Spaces from AWS.

The badge is rendered through a new `EnvLockLabel` rather than `SettingsField`'s `locked` prop: `locked` swaps the help for the "pinned by `<var>`" line, which is the right story for a field an admin *could* otherwise edit and the wrong one for a value that is only ever reported.

### New "Configure environment" section

A builder for the two areas that stay environment-only — **media storage** (filesystem or S3) and the **fitness map provider** (OpenStreetMap, Mapbox, Apple Maps). Pick an area, fill in the values, copy the assembled `.env` block, paste, restart.

It is deliberately inert: nothing typed into it is submitted, stored, or sent anywhere. An amber notice says so plainly, because the section sits directly under three sections that *do* save — "The **Update** button above does not save any of this — the server reads {storage,fitness maps} from the environment at boot, never from the database."

Details worth calling out:

- **Required variables always appear** in the block, carrying their placeholder as a visible to-do; **optional ones join only once filled**, so a copied block never contains a fake credential.
- **Secrets are masked** in the preview (12 bullets regardless of length) but copied verbatim.
- Every area stays mounted behind the native `hidden` attribute, so half-typed values survive an area switch while staying out of the tab order and the accessibility tree.
- The preview is a fixed dark terminal surface in both themes; in dark mode it steps *below* the card and keeps a hairline ring, since plain `neutral-900` *is* the card colour there.

### Real variable names, not the mock's

The design mock invents `ACTIVITIES_OBJECT_STORAGE_*`. The builder emits what this server actually reads:

- Storage: `ACTIVITIES_MEDIA_STORAGE_TYPE` / `_PATH` / `_BUCKET` / `_REGION` / `_ENDPOINT` / `_HOSTNAME`.
- **S3 credentials are not `ACTIVITIES_*` at all** — `createStorageS3Client` leaves them to the AWS SDK's default chain — so the builder emits `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`, both optional for IAM-role hosts. This is now documented too.
- Maps: `ACTIVITIES_FITNESS_MAP_PROVIDER` plus the Mapbox token or the three Apple Maps values. The Apple private-key placeholder is the single-line `\n`-escaped PEM `getMapProviderConfig` actually expands, not the mock's bare base64.

The name constants live in `lib/config/environmentTemplates.ts` because environment-variable name constants must stay inside `lib/config` (`AGENTS.md`, enforced by `envAccess.test.ts`). That module reads nothing from `process.env`, so the Client Component can import it safely.

## Deliberate deviation from the mock

The upload-size help keeps its `Up to 1,024 MB.` sentence, which the mock does not have — the mock has no configurable ceiling, and that sentence landed in #1302 for a reason. Everything else in these two sections matches.

## Tests

- `lib/services/medias/storageBackendSummary.test.ts` — new; every backend shape including the endpoint case.
- `lib/components/admin/settings/EnvBlockBuilder.test.tsx` — new; notice wording per area, field swapping per choice, optional-variable omission, secret masking, copy carrying the real secret, and state surviving an area switch.
- `PostsMediaSettingsForm.test.tsx` — the old `does not offer the storage backend` case is replaced by three that assert it is reported read-only with the badge and its own help, and that the builder renders.

Full gate green: `prettier` → `lint` (0 errors) → `build` → `test` (711 files, 6,822 tests).

## Verification

Verified in a real browser against a local SQLite instance with `ACTIVITIES_MEDIA_STORAGE_TYPE=s3` set, in both light and dark mode: the read-only backend line, the notice wording changing with the area, S3 and filesystem field sets, all three map providers, the optional-endpoint and secret-masking behaviour, and the copy button.


llun (commented):
## Sub-agent review round 1

Five reviewers (correctness, project invariants, security/privacy, a11y + design fidelity, tests) over the whole diff, then every finding put to three independent skeptics with distinct lenses (does it reproduce / already handled / does the cited rule exist), majority-refute to kill.

**19 raised → 3 survived.** The three are inline below and are all fixed in 6eeb531f8, each falsified against the mutation that previously slipped through.

### Refuted, for the record (16)

Killed 3-0 or 2-1, with the reason:

| Claim | Why it died |
|---|---|
| `autoComplete="off"` ignored on `type=password`, secret lands in the browser vault | Not a defect of this PR: no rule, and the alternative (`type=text`) is strictly worse. The section is admin-only and inert. |
| Copy emits placeholders as real values (×2) | Deliberate, documented template convention — the block is a template, and required fields advertise what still needs filling. No rule governs it. |
| `ACTIVITIES_MEDIA_STORAGE_ENDPOINT` placeholder is an AWS URL (×2) | Not a defect, but the reviewers were right that it was self-contradictory — changed anyway. |
| Secret mask reveals length under 12 chars (×2) | Low, no rule — but the test comment claimed otherwise, so the mask is now fixed-width. |
| Read-only Storage backend renders a `<label>` bound to no control (×3) | Mechanically true, but no such rule exists in AGENTS.md/REVIEW.md, the pattern predates this PR in `SettingsField`, and the caption is adjacent to its text. |
| Copy button has no `focus-visible` style / ring is ~2.2:1 on the dark block | No focus-contrast rule in the repo; `app/globals.contrast.test.ts` is explicitly scoped to muted text tokens. |
| Scrollable `.env` block not keyboard reachable (no `tabIndex`) | No rule; and the same content is reachable via the Copy button. |
| `=` separator fails AA (3.79:1 / 4.18:1) | Punctuation between two AA-passing runs; not covered by the enforced contrast floor. |
| Selector values are hardcoded with no test tying them to the resolvers | The resolvers accept exactly these literals; a test would restate the constant. |
| `ACTIVITIES_MEDIA_STORAGE_ENDPOINT` placeholder reaching the clipboard | Cannot happen — the field is `optional`, and `lines` skips optional fields with no value. |
| Only one `storageBackend` shape is rendered | Partly fair; the no-detail branch now has a test, the shapes themselves are covered by `storageBackendSummary.test.ts`. |

Nothing was found against the module-boundary, env-access, env-documentation, client-fetch, logging, or page-chrome invariants, and no path was found by which a typed credential leaves the browser.

lib/components/admin/settings/EnvBlockBuilder.tsx
```diff
@@ -0,0 +1,245 @@
...
+        onClick={() => copy(block)}
+        className={`absolute top-3 right-3 rounded-md border px-2.5 py-1 text-[11px] font-medium transition-colors ${
+          copied
+            ? 'border-green-400 text-green-300'
+            : 'border-white/20 text-white/75 hover:bg-white/10'
+        }`}
+      >
+        {copied ? 'Copied' : 'Copy .env block'}
+      </button>
+      <pre className="overflow-x-auto font-mono text-[12.5px] leading-6 text-neutral-200">
```
llun, line 114: **[medium / a11y-design] Copy button permanently hides the end of the block's first line at narrow widths.**

The button is `absolute top-3 right-3` over the `<pre>`, which reserves no right padding. Measured live at a 375px viewport: the pre's content box is 246px and the button is 107.4px, so it occludes the rightmost 103.4px of line 1.

Because a `pre`'s scroll extent is set by its *longest* line, that occlusion is unrecoverable whenever line 1 *is* the longest — 17px of available scroll against 103.4px of occlusion for Fitness maps → OpenStreetMap (a single-line block), and 48px vs 50.7px for Media storage → Local filesystem. Desktop is unaffected (860px pre), which is why it reviews clean on a laptop. The design spec has the same overlay, but it is a desktop-only showcase.

lib/components/admin/settings/EnvBlockBuilder.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    ).toHaveTextContent(
+      'the server reads storage from the environment at boot, never from the database'
+    )
+  })
+
+  it('starts on media storage with the S3 block', () => {
+    render(<EnvBlockBuilder />)
+    const area = activeArea(STORAGE_AREA)
+    expect(area.getByLabelText('Storage type')).toHaveValue('s3')
+    expect(area.getAllByText('ACTIVITIES_MEDIA_STORAGE_TYPE')).not.toHaveLength(
```
llun: **[medium / tests] Both “…block” tests pass with the generated block deleted outright.**

`expect(area.getAllByText("ACTIVITIES_MEDIA_STORAGE_TYPE")).not.toHaveLength(0)` here, and `.toBeGreaterThan(0)` in `switches to the fitness map provider block`, cannot fail: `getAllByText` already throws on zero matches, and that name is independently rendered *outside* the block by the selector’s help line (`Sets <EnvVarName name={area.selectorName} />`).

Verified by mutation — replacing `<EnvBlockPreview lines={lines} />` with `{false && …}` left both green, and restricting the preview to the storage area left **all** tests green, so the fitness-maps block (the deliverable of that area) had no coverage at all. Contrast the exact-count assertions in `leaves an optional variable out of the block…`, which do pin block membership.

lib/components/admin/settings/PostsMediaSettingsForm.test.tsx
```diff
@@ -60,11 +71,33 @@ describe('PostsMediaSettingsForm', () => {
...
 
-  it('does not offer the storage backend, which is environment-only', () => {
+  // The backend is infrastructure read from the environment at boot, so it is
+  // reported with an env-lock badge rather than offered as an editable control.
+  it('reports the storage backend read-only', () => {
     renderForm()
+    expect(screen.getByText('Storage backend')).toBeInTheDocument()
+    expect(screen.getByText('S3 — media.example.social')).toBeInTheDocument()
+    expect(screen.getByText('(eu-central-1)')).toBeInTheDocument()
     expect(screen.queryByLabelText('Storage backend')).not.toBeInTheDocument()
```
llun: **[medium / tests] The “not editable” assertion cannot fail, leaving the PR’s central claim unguarded.**

`queryByLabelText("Storage backend")` returns null for two reasons unrelated to whether a control exists: the field passes no `htmlFor`, and the new `EnvLockLabel` nests the badge inside the label so RTL computes the accessible name as `"Storage backendSet by environment"`.

Verified by mutation: inserting an editable `<Select>` into that field left all 8 tests green, with *and* without a matching `htmlFor`/`id`. The same assertion does trip on `main`, so this PR turned a working guard into a tautology. It needs to assert on controls, not on label association.

lib/components/admin/settings/EnvBlockBuilder.tsx
llun: Fixed in 6eeb531f8: the `<pre>` now carries `pr-28` (112px > the 103.4px occlusion), reserving the button’s width *inside* the scroll area so scrolling clears it.

Re-measured live at 375px, gap between the longest line’s right edge and the button’s left edge at max scroll:

| area / choice | before | after |
|---|---|---|
| maps → osm | −86px (unreachable) | **+26.1px** |
| maps → mapbox | +5.3px | **+146.6px** |
| maps → apple | ok | **+455.1px** |
| storage → fs | −2.7px (unreachable) | **+56.1px** |
| storage → s3 | ok | **+154.1px** |

Kept the overlay rather than moving the button to its own row, so the design-system construction is preserved; on desktop the reserve is invisible.

lib/components/admin/settings/EnvBlockBuilder.test.tsx
llun: Fixed in 6eeb531f8. Both assertions now pin exact help-plus-block counts and the emitted value:

- `starts on media storage with the S3 block` → `getAllByText("ACTIVITIES_MEDIA_STORAGE_TYPE")` has length **2**, plus `getByText("s3")`.
- `switches to the fitness map provider block` → same shape for `ACTIVITIES_FITNESS_MAP_PROVIDER`, plus `getByText("mapbox")`.
- `drops the credential fields for keyless OpenStreetMap` gained the equivalent pair, since the provider line *is* the whole block for `osm`.

Falsified against both mutations you used. `{false && <EnvBlockPreview …/>}` now fails 7 of 10; restricting the preview to the storage area now fails the 2 maps tests (it previously failed none).

lib/components/admin/settings/PostsMediaSettingsForm.test.tsx
llun: Fixed in 6eeb531f8. Replaced with a control-based assertion scoped to the Media section:

```tsx
const mediaSection = screen.getByRole("heading", { name: "Media" }).closest("section")
const controls = mediaSection.querySelectorAll("input, select, textarea")
expect(controls).toHaveLength(1)
expect(controls[0]).toHaveAttribute("id", "media-max-file-size")
```

The upload-size input is the section’s only control, so a storage-backend control of *any* kind — labelled or not — fails this. Falsified against your mutation: re-adding the editable `<Select>` now fails this test.

Also added `omits the parenthesised detail when the backend has none`, covering the other render branch.

lib/components/admin/settings/EnvBlockBuilder.tsx
```diff
@@ -0,0 +1,252 @@
...
+      >
+        {copied ? 'Copied' : 'Copy .env block'}
+      </button>
+      {/* pr-28 reserves the overlaid button's width. Without it the button
+          covers the tail of the first line, and since the scroll extent is set
+          by the longest line, that tail is unreachable at narrow widths
+          whenever line 1 is the longest — which is every single-line block
+          (`osm`) and the filesystem one. The reserve is inside the scroll area,
+          so scrolling now clears the button; on desktop nothing is near it. */}
+      <pre className="overflow-x-auto pr-28 font-mono text-[12.5px] leading-6 text-neutral-200">
```
llun: ## Round 2 — this fix does not work

`pr-28` is inert. A `pre`’s inline scroll extent is `max(clientWidth, longest line)`: end-side padding lives *inside* the client box and is never appended after overflowing content, so the reserve changes neither `scrollWidth` nor what the button covers. Measured in Chromium at 375px, `padding-right: 112px` and `0px` give identical geometry.

**My round-1 verification was invalid.** I compared the button against each line’s `<div>` box — and a block div’s box shrinks with the padding, so it moved left and reported a gap the glyphs never had. Measuring the text with a `Range` shows the truth.

So the two cases round 1 reported (`maps → osm`, `storage → fs`) were byte-for-byte unchanged.

Round 2 raised 11 findings; these two (the same root cause, from different dimensions) are the only survivors.

llun: Fixed structurally in 2017414d8: the button leaves the overlay and sits in normal flow above the block. It cannot overlap any line at any width, and no arithmetic is involved — which matters, since arithmetic is what got this wrong the first time.

Re-measured at 375px with a `Range` over the text (not the line `<div>`), scrolled to the end:

| area / choice | button clear of block | maxScroll | last glyph inside the block |
|---|---|---|---|
| maps → osm | ✅ | 17px | 0.1px |
| maps → mapbox | ✅ | 138px | 0.2px |
| maps → apple | ✅ | 446px | 0.1px |
| storage → fs | ✅ | 48px | 0.0px |
| storage → s3 | ✅ | 145px | 0.2px |

Every line now reaches its end on screen. Desktop is unchanged apart from the button moving up one row.

Also added a structural test, since jsdom cannot see layout: the button must not live inside the `pre`, must not be absolutely positioned, and must precede the block in document order.

llun:
## Sub-agent review loop — closed after 3 rounds

| round | raised | survived verification | outcome |
|---|---|---|---|
| 1 | 19 | 3 | all fixed in 6eeb531f8 |
| 2 | 11 | 2 (one root cause) | round-1 layout fix was inert → replaced structurally in 2017414d8 |
| 3 | 3 | **0** | clean round; loop closed |

110 agents total. Round 3 is the clean round AGENTS.md asks for, so the loop is done. Every thread is replied-to and resolved.

### The one that mattered

Round 1 found the overlaid “Copy .env block” button hiding the tail of the block’s first line at phone widths. I fixed it with a `pr-28` padding reserve and “verified” it by measuring the button against each line’s `<div>` box — but a block div’s box *shrinks with the padding*, so it reported a gap the glyphs never had. Round 2 caught that: a `pre`’s inline scroll extent is `max(clientWidth, longest line)`, end-side padding is never appended after overflowing content, and six independent Chromium measurements put `padding-right: 112px` and `0px` at identical geometry.

The real fix takes the button out of the overlay entirely — it now sits in normal flow above the block, so it cannot overlap any line at any width and no arithmetic is involved. Re-measured with a `Range` over the text at 375px: the button is vertically clear of the block and every line’s last glyph lands within 0.2px of the right edge at max scroll, for all five (area, choice) combinations. A structural test guards it, since jsdom cannot see layout.

### Round 3’s refuted findings, for the record

- *Storage builder omits the documented `object` backend* — `object` and `s3` hit the same branch in `getMediaStorageConfig` and the same driver, so offering both would present a distinction that does not exist.
- *`dark:bg-neutral-950` makes the block the same colour as its section* — measured true (both #0a0a0a), but that is the app’s dark surface language and the ring draws the edge. It did make a code comment of mine inaccurate, corrected in 899aa0a01.
- *Light-theme scrollbar renders light-on-black inside the dark block* — the 15px classic scrollbar the finding measured does not exist in this environment, and the residue is native-chrome tint with no rule behind it.

Full gate green on every round: `prettier` → `lint` (0 errors) → `build` → `test` (711 files, 6,825 tests).

PR #1311, opened 2026-07-24
chore: remove unused dependencies

llun:
Removes 17 dependencies from `package.json` that no code imports. Purely subtractive: no new code, no behaviour change, no new tests.

> **Review note:** this PR originally removed 18. Code review caught that `kysely` is a declared `peerDependency` of `@better-auth/core` — a *direct* dependency of this workspace — so dropping it introduced a new `YN0002` install warning. It has been restored, at a cost of **zero** packages (`kysely@0.29.4` is in the tree regardless as a hard dep of `better-auth`), and moved into the "must NOT be removed" table below.

## Measured before / after

| Metric | Before | After | Delta |
|---|---|---|---|
| `yarn.lock` resolutions | 1136 | 821 | **−315 packages** |
| `node_modules` size | 861M | 728M | **−133M** |
| `yarn install` peer warnings (`YN0060`) | 6 | 2 | −4 |
| Unmet-peer warnings (`YN0002`) | 0 | 0 | unchanged |
| `yarn lint` | 0 errors, 31 warnings | 0 errors, 31 warnings | **unchanged** |
| `yarn test` | 712 files / 6835 tests | 712 files / 6835 tests | **unchanged** |

Lint output being byte-for-byte identical is the proof that the removed ESLint packages contributed no rules. The 4 peer warnings that disappeared were all emitted by the legacy `eslint-config-standard` chain and `eslint-config-next`; the 2 that remain (`better-call`, `better-sqlite3`) are deliberate exact pins.

## What was removed, and the evidence for each

Every package below returns **zero hits** for an import-specifier search across `app/`, `lib/`, `scripts/`, `migrations/`, `.github/`, and every root-level config file:

```bash
grep -rnE "(from|import|require\()[ (]*['\"]<package>(/[^'\"]*)?['\"]" <paths>
```

**Markdown stack (4)** — `react-markdown`, `rehype-sanitize`, `remark-breaks`, `remark-gfm`
`lib/components/post-box/post-box.tsx` no longer imports any of them; only the test still mocked them. Those four `vi.mock()` blocks are deleted here too (Vitest does not require a mocked module to be imported, so they were inert). The test file passes with the same 31 tests before and after.

**Runtime (5)**
- `framer-motion` — no references of any kind.
- `@aws-sdk/s3-presigned-post` — no references. Presigned URLs come from `getSignedUrl` in `@aws-sdk/s3-request-presigner`, which stays.
- `content-type` — no imports. A bare text search hits hundreds of times, but every one is an HTTP header string (`headers.get('content-type')`), not a module import.
- `@simplewebauthn/browser`, `@simplewebauthn/server` — both are regular `dependencies` (not peers) of `@better-auth/passkey`. Verified both still resolve from `node_modules/@better-auth/passkey` after removal, so the passkey plugin is unaffected.

**Tooling (8)** — `@eslint/eslintrc`, `eslint-config-next`, `eslint-config-prettier`, `eslint-config-standard`, `eslint-plugin-import`, `eslint-plugin-n`, `eslint-plugin-promise`, `vitest-fetch-mock`
`eslint.config.mjs` is a flat config importing exactly six packages — `@eslint/js`, `@next/eslint-plugin-next`, `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `eslint-plugin-unused-imports`, `globals` — none of which are touched here. Everything else in the ESLint family was declared but wired to nothing. `vitest-fetch-mock` is redundant with `jest-fetch-mock`, which is what `vitest.setup.ts` actually imports to install the global `fetchMock`.

## This does NOT shrink the runtime Docker image

Worth stating plainly so nobody measures the wrong thing: Next bundles application code into `.next/server` chunks and only copies native/external packages into the standalone output, so the shipped image is essentially unchanged. The real wins are **install time, CI time, lockfile size, and ~315 fewer packages in the Dependabot/audit surface**.

## Packages that must NOT be removed (verified — please don't "helpfully" extend this cleanup)

These look unused by a naive import scan but are load-bearing. Recording the analysis here so a future reader doesn't repeat it:

| Package | Why it stays despite zero imports in our source |
|---|---|
| `nanostores` | Declared `peerDependency` of `@better-auth/passkey`. Yarn does not auto-install peers, so the workspace must declare it. |
| `better-call` | Same — peer of `@better-auth/passkey` and `@better-auth/oauth-provider`. The exact pin is deliberate. |
| `@better-auth/utils` | Same. Exact pin is deliberate. |
| `@better-fetch/fetch` | Same. Exact pin is deliberate. |
| `kysely` | Peer of `@better-auth/core` (`^0.28.5 \|\| ^0.29.0`), in the same `peerDependencies` block as `nanostores` and `better-call`. It resolves today only because `better-auth` hard-depends on it and the node-modules linker hoists it — not because of a declared edge. Removing the declaration emits `YN0002` and would break under `nodeLinker: pnpm`/PnP. Costs nothing to keep. |
| `@next/env` | Imported by `scripts/maintenance/fixAttachmentUrls.ts` and `scripts/maintenance/manageAdminRole.ts` (`loadEnvConfig`). An import scan misses it because `scripts/` is lint-ignored. |
| `@aws-sdk/client-lambda`, `@aws-sdk/util-utf8-node` | Both imported by `lib/services/email/lambda.ts`. |
| `postcss` (devDep) + the `postcss` entry in `resolutions` | Load-bearing. `next@16` hard-pins `postcss` to an exact `8.4.31`; only the resolution lifts the whole tree onto a patched 8.5.x. Removing it reopens a previously-fixed Dependabot alert. |
| `@radix-ui/react-slot` pinned to exact `1.3.0` | Deliberate. 1.3.1 shipped a module-scope `React.createContext` without its `use client` directive, breaking the RSC build of every page rendering `lib/components/ui/button.tsx`. Upstream fixed it in 1.3.2 — bumping to `^1.3.2` is fine, but do not float it to `^1.3.1`. |

## Follow-up candidates (deliberately out of scope — each needs its own PR)

1. **`react-modal` → `@radix-ui/react-dialog`** (2 call sites; `@radix-ui/react-dialog` is already a dep and is the design-system standard). It's a UI change, so it needs real-browser verification and screenshots.
2. **`lodash`** — 14 import sites using 8 members. `memoize`/`noop`/`identity`/`trim` inline trivially, but `_.chain` and `_.curryRight` in `lib/utils/text/processStatusText.ts` and `lib/components/posts/post.tsx` need a genuine rewrite. A refactor, not a dependency swap.
3. **`@aws-sdk/util-utf8-node`** — deprecated upstream, one call site (`fromUtf8` in `lib/services/email/lambda.ts`) that `Buffer.from(value, 'utf8')` covers. Touches the Lambda email backend, which has no local test harness.
4. **`@vitest/coverage-v8`** — no CI job or `vitest.config.ts` setting uses coverage. Team preference, not a technical finding.

## Test plan

Full gate run in order on the combined change, all green:

- `yarn run prettier --write .` — no files rewritten
- `yarn lint` — 0 errors, 31 warnings (identical to `main`; the 31 are pre-existing `no-explicit-any` warnings in `vitest.d.ts`)
- `yarn build` — compiles
- `yarn test` — 712 files / 6835 tests passed
- `yarn install --immutable` — exits 0 with no new warnings versus `main`, so CI's install mode is satisfied
- `require.resolve` checks confirm `@simplewebauthn/server` and `@simplewebauthn/browser` still resolve transitively from `node_modules/@better-auth/passkey`, where both are regular (non-peer) dependencies
- Client-side passkey resolution verified: `@better-auth/passkey/dist/client.mjs` imports `@simplewebauthn/browser` from inside its own package, so it never relied on top-level hoisting
- All CI checks green, including `Build`, `CodeQL`, all four test shards, `Schema Dump Sync`, and the aggregate `CI Success`

`package.json` `version` is untouched.


llun:

Three reviewers ran in parallel over the full diff: (1) adversarial dependency-removal safety, (2) test-file change + project conventions, (3) CI/build/lockfile/Docker risk.

### Finding 1 — `kysely` is a declared peer of `@better-auth/core` (**fixed** in f08ec73)

**Severity: medium.** Two reviewers independently found this.

`@better-auth/core@1.6.25` is a **direct** dependency of this workspace (and is imported at `app/api/v1/apps/createApplication.ts:10` and `lib/services/oauth/matchRedirectUri.ts:5`). It declares `kysely` as a non-optional `peerDependency`:

```
"kysely": "^0.28.5 || ^0.29.0"
```

That entry sits in the **same `peerDependencies` block** as `nanostores` and `better-call` — both of which this PR's own "must NOT be removed" table preserves for exactly this reason. The original analysis reasoned about `better-auth`'s *hard* dependency on kysely, which is a different package and does not address the peer contract.

Effect of removing it:

```
YN0002: activities.next@workspace:. doesn't provide kysely (pcd4340), requested by @better-auth/core.
```

— a new install warning not present on `main`. It did not break anything today (`yarn install --immutable` still exits 0, and `kysely@0.29.4` stays hoisted via `better-auth`), but resolution then depended on hoisting rather than a declared edge, which breaks under `nodeLinker: pnpm`/PnP or if `better-auth` ever drops its direct kysely dep.

**Resolution:** restored as a `dependency` and moved into the "must NOT be removed" table. It costs **zero** packages — the lockfile is still at 821 resolutions — so the −315 result is unchanged. The PR now removes 17 packages, not 18. Re-gated: lint 0 errors/31 warnings, build passes, 712 files / 6835 tests pass, `yarn install --immutable` clean with no YN0002.

### Everything else: verified clean

Both reviewers tried hard to break the remaining 17 and could not:

- **Non-obvious import forms** — repo-wide sweep covering `proxy.ts`, `instrumentation.ts`, `next.config.ts`, `knexfile.js`, `vitest.config.ts`, `vitest.setup.ts`, `vitest-shims/`, `migrations/`, `.github/`, `Dockerfile`, `.husky/`, `postcss.config.mjs`, `scripts/`: zero hits. The only two dynamic imports in the repo take local route-path variables, not package names.
- **`@simplewebauthn/browser` on the client** — the sharpest risk, since passkeys are live across ~20 files. Confirmed `@better-auth/passkey/dist/client.mjs` imports it from *inside its own package*, so it never relied on top-level hoisting. The project's own passkey UI uses raw `window.PublicKeyCredential`.
- **`eslint-config-next` / `next lint`** — `next lint` was removed in Next 16 and is invoked nowhere; `next.config.ts` has no `eslint` key, so there is no build-time lint step being silently skipped.
- **`framer-motion`** — no `motion` / `motion/react` alias imports either. Animation comes from `tw-animate-css`.
- **Type-only usage** — no `import type` from any removed package; none of the 14 remaining `@types/*` entries corresponds to a removed runtime package.
- **`yarn.lock`** — pure removal, **no version churn**. Every apparent version change is duplicate-version elimination where the surviving version already existed on `main` (e.g. `@typescript-eslint/*` 8.56.0 dropped, 8.65.0 kept). `__metadata` unchanged.
- **`package.json`** — only the intended entries changed; `version`, the load-bearing `postcss` resolution, `dependenciesMeta.better-sqlite3.built:false`, `packageManager`, `engines`, `scripts`, and `lint-staged` all intact.
- **Docker** — no removed package referenced; the "no runtime image shrink" claim is confirmed (Next standalone is `@vercel/nft`-traced, and none of these were ever traced in).
- **Docs** — every tracked `*.md` grepped for all 18 names: zero hits, so nothing is made stale.

### Unrelated pre-existing issue (not addressed here)

One reviewer hit a single failure at `app/(timeline)/[actor]/ActorTimelines.test.tsx:320` on one full-suite run. It is a **pre-existing flake unrelated to this PR** — the file is untouched by the diff, it passes in isolation (17/17) and on re-run, and the assertion is a genuine race (`getAllByText` fires immediately after a `waitFor` on the mock call rather than on the re-render). Flagging it for a separate fix rather than widening this PR's scope.

llun:
## Sub-agent code review — round 2: clean, loop closed

Round 2 verified the round-1 fix and hunted for anything missed. **No new actionable findings**, so the review loop converges here (2 of a maximum 20 rounds).

### The `kysely` fix verified

- `kysely@^0.29.4` is back in **`dependencies`** (not devDependencies), satisfying `@better-auth/core`'s required peer range `^0.28.5 || ^0.29.0`. Its `peerDependenciesMeta` marks only `@cloudflare/workers-types` and `@opentelemetry/api` optional, confirming kysely is non-optional.
- `git diff origin/main -- package.json | grep -c kysely` → **0**. Byte-identical to `main`; the removal is fully reverted.
- `yarn install` → **YN0002 = 0**, YN0060 = 2. Both remaining YN0060s (`better-call`, `better-sqlite3`) are pre-existing on `main`, not introduced here.
- Lockfile still **821** resolutions before and after install — restoring kysely added nothing.

### The systematic peer-dependency audit (the check round 1 only did for one package)

Walked the **entire** `node_modules` tree (nested + scoped, depth 6), reading every `package.json`'s `peerDependencies`/`peerDependenciesMeta` and cross-referencing against all 18 names, with kysely as a positive control.

**Complete hit list:**

| Package | Peer | Optional? |
|---|---|---|
| `@better-auth/core@1.6.25` | `kysely: ^0.28.5 \|\| ^0.29.0` | **REQUIRED** — now satisfied |
| `@better-auth/kysely-adapter@1.6.25` | `kysely: ^0.28.17 \|\| ^0.29.0` | optional |

**Zero** packages anywhere in the tree declare a peer dependency on any of the other 17. Round 1's finding was the only one of its kind.

### Phantom-dependency sweep

The lockfile lost 283 packages and gained 0. Every bare import/require/dynamic-import specifier was extracted from all **2093** tracked source files (84 distinct packages) and checked against both the removed set and `node_modules`: **0 imports name a removed package, 0 imports fail to resolve.**

This is the check that matters most for `scripts/` — 31 files that are excluded from lint, never built, and have no tests, so a phantom-dependency break there would pass all four CI gates silently. Clean.

### Gate, re-run on the final state

| Gate | Result |
|---|---|
| `yarn lint` | 0 errors, 31 warnings (identical to `main`) |
| `yarn build` | exit 0; no "Module not found" |
| `yarn test` | 712 files / 6835 tests passed |
| `yarn install --immutable` | exit 0 |

All four commits use valid `chore:` prefixes → patch bump. `version` remains `1.116.1` and is not in the diff. CI is green on the pushed fix, including all four test shards and the aggregate `CI Success`.

### Leads investigated and cleared (so a future round needn't repeat them)

- `vitest-fetch-mock`'s removal does not orphan the fetch mocking in `vitest.setup.ts` / `lib/client.test.ts` — that is `jest-fetch-mock`, an independently declared devDependency. On `main`, `vitest-fetch-mock@0.4.5` had only a `vitest` peer and no relationship to it.
- `@simplewebauthn/browser@13.3.0` and `@simplewebauthn/server@13.3.2` remain physically present, supplied transitively by `@better-auth/passkey` (a direct dep). No repo source imports them, so this is correct hoisting rather than a phantom-dependency hazard.
- No doc updates are owed: the only regex hits across docs/configs/workflows were the English word "unified" in `AGENTS.md`/`CLAUDE.md` prose about content width — not the `unified` npm package.

PR #1313, opened 2026-07-25
minor: widen fitness privacy radius to 50m-1km and tidy the route map overlays

llun:
## Summary

Two fitness changes reported from the activity detail page.

### 1. The privacy hide radius was capped far too tight

The radius options were `5m / 10m / 20m / 50m`, so **50m was the maximum**. That is barely more than consumer GPS drift — a route still ended on the doorstep the zone was supposed to hide.

50m is now the **smallest** enabled zone rather than the largest, and the option set is **50m, 100m, 200m, 500m, 1km**.

Because zones already saved under the old options (5m/10m/20m) would sanitize to `0` and **silently stop hiding anything**, `sanitizePrivacyRadiusMeters` now snaps an unlisted positive radius **up** to the next supported option instead of dropping it (`5/10/20 → 50`, `300 → 500`, `5000 → 1000`). Non-numeric, non-finite and non-positive values still resolve to `0` (disabled). The snap runs on read as well as write, so an existing zone keeps hiding at least as much as before with no backfill needed.

The picker and the saved-location list render `1000` as `1km`.

### 2. Route map overlays

- **Removed the "GPS trace" badge** from the map's top-right corner. The route map is self-evidently a GPS trace, so the badge was label-only chrome sitting over the map.
- **The green "Green segments are hidden from other viewers" notice is now dismissible.** It is an acknowledgement, not a persistent legend, so it is a button that clears itself on tap and stops covering the route.

## Notes for reviewers

`POST /api/v1/fitness/general` validates the radius against the enumerated option set, so it now rejects `5`/`10`/`20`. Only this app's own settings UI calls that endpoint, and it always sends a sanitized value from the new set.

## Testing

- `yarn run prettier --write .` → clean
- `yarn lint` → 0 errors
- `yarn build` → green
- `yarn test` → **712 files / 6848 tests passed**

New regression coverage:
- `sanitizePrivacyRadiusMeters` table-driven cases for the new option set, the legacy snap-up, the in-between snap-up, the cap, and the invalid-input paths.
- `getFitnessPrivacyLocations` keeps a legacy 20m zone enabled by snapping it to 50m.
- `FitnessPrivacyLocationSettings` renders exactly `50m / 100m / 200m / 500m / 1km`.
- `FitnessStatusDetail` asserts there is no "GPS trace" badge, and that the privacy notice disappears when clicked. Both new detail tests were confirmed to **fail** against the pre-change component, so they are not vacuous.

### Browser verification (local SQLite + mock data)

Verified signed in against a local `dev.sqlite3` at `localhost:3100`:

1. `/settings/fitness/privacy` — the Hide Radius select offers `50m, 100m, 200m, 500m, 1km` and defaults to `50m`.
2. Saved a `1km` zone end-to-end; the API accepted it and the saved list shows `Hide radius: 1km`.
3. Imported a GPX route crossing that 1km zone and opened the activity detail page: the map renders with no "GPS trace" badge, the green notice appears over the partly-hidden route, and clicking it removes it and uncovers the map.

(The one console error on the detail page is a pre-existing hydration mismatch on the mobile nav's "More" dropdown trigger — unrelated to this change.)

## Known limitations

**Pre-generated route-map images and cached heatmaps are not re-rendered.** The snap runs on read, so a zone stored at 20m immediately masks at 50m on every live surface — but the route-map PNG that was baked with the 20m mask, and any cached heatmap rows, still reflect the old radius until they are regenerated. The image never exposes more than the radius the user originally chose, and this is the same drift that occurs whenever a user edits their radius, which is what the existing **"Regenerate maps for old statuses"** button in the same settings panel is for.

One sub-case worth recording: if the widened zone leaves fewer than two drawable visible samples, `shouldRenderInteractiveMap` goes false and the panel falls back to that stale static image. Auto-queueing regeneration for every affected actor on deploy is a larger change than this PR; happy to do it as a follow-up.

## Screenshots

Verified in a real browser (see Testing above) — captures were taken locally at each step, but `gh` cannot attach images to a PR body, so they are described rather than embedded:

1. `/settings/fitness/privacy` — Hide Radius select open, showing `50m / 100m / 200m / 500m / 1km`, defaulting to `50m`.
2. Same page after saving — Saved Privacy Locations shows `13.756300, 100.501800` / `Hide radius: 1km` with the "Fitness privacy location settings saved." confirmation.
3. Activity detail with a GPX route crossing the zone — map renders with no "GPS trace" badge; green notice sits bottom-left over the partly hidden route.
4. Same view after tapping the notice — notice gone, map fully visible.

Say the word if you want them attached via the web UI.

## Checklist

- [x] Commit messages and PR title use the correct conventional prefix
- [x] `package.json` version untouched (CI handles the bump)
- [x] `yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test` all green
- [x] Tests added for the new behavior, and confirmed to fail against the pre-change code
- [x] UI changes verified in a real browser
- [x] No migration added, so no schema dump regeneration needed
- [x] Docs checked for staleness (no markdown references the radius values or the removed badge)
- [x] Sub-agent code review run; findings posted, addressed, and resolved on this PR


lib/services/fitness-files/privacy.ts
```diff
@@ -1,8 +1,20 @@
...
+  0, 50, 100, 200, 500, 1000
+] as const
 
 export type FitnessPrivacyRadiusMeters =
   (typeof FITNESS_PRIVACY_RADIUS_OPTIONS)[number]
 
+// Derived, not a second literal: a hand-maintained copy would silently cap
+// below a newly added option.
+export const MAX_FITNESS_PRIVACY_RADIUS_METERS =
+  FITNESS_PRIVACY_RADIUS_OPTIONS[FITNESS_PRIVACY_RADIUS_OPTIONS.length - 1]
```
llun, line 16: **Cap duplicated the array's last element as a second literal.** If someone appended `2000` to the option set and forgot to update this constant, `sanitizePrivacyRadiusMeters(5000)` would return `1000` — capping *below* a radius the UI now offers.

Also, the upward snap silently depends on the array being sorted ascending: reordering it to `[0, 1000, 50, ...]` would make `sanitize(60)` return `1000` and jump every legacy zone to 1km.

Fixed: the cap is now derived from the list, both invariants are documented in the comment, and `privacy.test.ts` asserts them.

lib/services/fitness-files/privacy.ts
```diff
@@ -39,15 +51,17 @@ const toRadians = (degrees: number) => (degrees * Math.PI) / 180
...
 ): FitnessPrivacyRadiusMeters => {
-  if (typeof value !== 'number' || !Number.isFinite(value)) {
+  if (typeof value !== 'number' || !Number.isFinite(value) || value <= 0) {
     return 0
   }
 
-  return FITNESS_PRIVACY_RADIUS_OPTIONS.includes(
-    value as FitnessPrivacyRadiusMeters
+  // Snap an unlisted radius up to the next supported option rather than
+  // dropping it to 0. Zones saved under the older, smaller option set (5m,
```
llun, line 59: **Accepted, not fixed — stale pre-generated artifacts.** The snap runs on *read*; stored rows keep `hideRadiusMeters: 20` and nothing rewrites them or re-renders the route-map PNGs / cached heatmaps that were baked with the old 20m mask. So live surfaces now hide more than the already-generated image does.

Why this is not a blocker: the image never leaks more than the radius the user originally chose, and this is the same drift that occurs any time a user edits their radius — which is exactly what the existing **"Regenerate maps for old statuses"** button in this settings panel is for. Auto-queueing a regeneration fan-out for every actor on deploy is a much larger change than this PR's scope.

The sharper sub-case is real and worth recording: if the widened zone leaves fewer than two drawable visible samples, `shouldRenderInteractiveMap` goes false and the panel falls back to that stale static image. Documented in the PR body; happy to open a follow-up that invalidates or re-queues affected artifacts if you want it handled automatically.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -876,6 +881,14 @@ const ActivityMapPanel: FC<{
     [drawableRouteSegments]
   )
 
+  // The panel is not keyed per file, so switching activity files in an
+  // aggregated post swaps `routeSegments` underneath the same instance. Without
+  // this reset a dismissal would carry over and the next file would draw green
+  // segments with no explanation.
+  useEffect(() => {
+    setIsPrivacyNoticeDismissed(false)
+  }, [routeSegments])
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:890]: **Bug: the dismissal leaked across activity files.** `ActivityMapPanel` is rendered at a stable tree position with no `key`; the multi-file switcher only swaps the `routeSegments` prop, so the component never unmounts.

Failure: on an aggregated post where two files both have hidden segments, dismissing the notice for file 1 and then switching to file 2 drew green segments on the map with no explanation at all. (Inverse inconsistency: switching Overview ↔ Analysis *did* remount the panel and brought the notice back, so the behaviour was arbitrary from the user's side.)

Fixed with a reset effect on `routeSegments`. `key={fitness?.id}` would also work but tears down and rebuilds the GL map. Regression test `brings the privacy notice back when another activity file is selected` was confirmed to fail without the effect.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1197,13 +1210,16 @@ const ActivityMapPanel: FC<{
...
           ) : null}
-          <div className="absolute right-3 top-3 inline-flex items-center gap-1.5 rounded-md border bg-background/95 px-2.5 py-1 text-xs font-medium text-muted-foreground shadow-sm">
-            <Route className="size-3.5" /> GPS trace
-          </div>
-          {hasHiddenPrivacySegments ? (
-            <div className="absolute bottom-3 left-3 rounded-md border border-green-300 bg-background/95 px-3 py-2 text-xs font-medium text-green-700 shadow-sm dark:border-green-900 dark:text-green-400">
+          {hasHiddenPrivacySegments && !isPrivacyNoticeDismissed ? (
+            <button
+              type="button"
+              onClick={() => setIsPrivacyNoticeDismissed(true)}
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1216]: **The notice was a button with nothing to say it was one.** No `cursor-pointer`, no hover state, no dismiss icon — and Tailwind v4's preflight leaves buttons on the default arrow cursor, which is why `lib/components/ui/button.tsx` and the sibling raw `<button>` 40 lines above both add it explicitly. The adjacent zoom controls also carry `hover:bg-muted`.

That directly defeated the point of the change: the notice permanently occluded the bottom-left of the map (frequently the route start/finish) and nothing suggested it could be dismissed.

Separately, the accessible name was just the notice text, so a screen reader announced "…, button" with no indication of what activating it does.

Fixed: `cursor-pointer hover:bg-muted`, an `X` icon (`aria-hidden`), and `aria-label="Dismiss notice: …"`.

One related point accepted rather than fixed: activating the button unmounts the focused element, so keyboard focus falls to `<body>`. Moving focus into the map container would add a tab stop to a decorative overlay; for a transient leaf notice I judged that the worse trade. Flag it if you disagree.

lib/services/fitness-files/privacy.test.ts
```diff
@@ -69,30 +72,32 @@ describe('getDistanceMeters', () => {
...
     expect(segments).toHaveLength(2)
     expect(segments[0]).toHaveLength(2)
     expect(segments[1]).toHaveLength(2)
     expect(segments.flat()).toEqual([
-      { lat: 52.0002, lng: 5.0002 },
-      { lat: 52.0003, lng: 5.0003 },
-      { lat: 52.0002, lng: 5.0002 },
-      { lat: 52.0003, lng: 5.0003 }
+      { lat: 52.002, lng: 5.002 },
+      { lat: 52.003, lng: 5.003 },
```
llun, line 98: **Narrowing `FitnessPrivacyRadiusMeters` broke this file's types.** `tsc --noEmit` reported three real errors — `radiusMeters: 5` is no longer assignable to `0 | 50 | 100 | 200 | 500 | 1000`. CI never type-checks `*.test.ts`, so nothing caught it.

The fix was not a number swap: at `radiusMeters: 50` the old `0.0002`-degree offsets are only ~26m from home, so those points become *hidden* and both assertions flip. Rescaled the fixture to `0.002`/`0.003` (~261m/~392m) so the visible/hidden split still holds, and added a comment recording the sizing.

(The second test needed only `5` → `50`; its zones are half a degree apart.)

app/api/v1/fitness/general/route.test.ts
```diff
@@ -478,6 +523,85 @@ describe('Fitness General Settings API', () => {
       expect(mockDb.createFitnessSettings).not.toHaveBeenCalled()
     })
 
+    // The Zod schemas are the only thing keeping an out-of-set radius out of
+    // the database, and this is the release that redefined the set.
+    it.each([
+      { description: 'rejects a legacy 5m radius', hideRadiusMeters: 5 },
+      { description: 'rejects a legacy 10m radius', hideRadiusMeters: 10 },
```
llun, line 530: **No test covered the API rejecting the now-removed radii.** These Zod schemas are the only thing keeping an out-of-set radius out of the database, and this is the release that redefined the set — widening the array back or dropping the `.refine()` would have let arbitrary radii persist with every test still green.

Added a table-driven 422 case for 5/10/20/75/2000, plus a positive case for 1000, which is the new upper bound and had zero API-layer coverage.

app/api/v1/fitness/general/route.test.ts
```diff
@@ -149,12 +149,54 @@ describe('Fitness General Settings API', () => {
...
+          hideRadiusMeters: 200
         }
       ])
       expect(data.privacyHomeLatitude).toBe(13.7563)
       expect(data.privacyHomeLongitude).toBe(100.5018)
-      expect(data.privacyHideRadiusMeters).toBe(20)
+      expect(data.privacyHideRadiusMeters).toBe(200)
+    })
+
+    it('snaps a zone saved under the older option set up to 50m', async () => {
```
llun, line 160: **The legacy snap — the whole motivation for the sanitizer rewrite — had no coverage above the unit level.** My bulk fixture edit (20 → 200) had removed the only legacy value in the API tests, and with it the natural place to observe the migration.

This is the "an existing user's 20m zone must keep hiding something" contract; a regression here silently de-anonymises home locations. Added a GET test with a stored `20` asserting the response snaps to `50`.

app/api/v1/fitness/general/route.test.ts
```diff
@@ -347,17 +389,17 @@ describe('Fitness General Settings API', () => {
...
             latitude: 35.6764,
             longitude: 139.65,
-            hideRadiusMeters: 10
+            hideRadiusMeters: 100
           }
         ],
         privacyHomeLatitude: 13.7563,
         privacyHomeLongitude: 100.5018,
-        privacyHideRadiusMeters: 20,
+        privacyHideRadiusMeters: 200,
```
llun, line 402: **My bulk edit weakened this test.** The legacy `privacyHideRadiusMeters` here went `5` → `50`. Under the new option set `5` is invalid, and keeping it made the test strictly stronger: it proves that when `privacyLocations` is present the legacy branch is neither validated nor applied, *even with a value the schema would reject*. Changing it to a now-valid `50` made the assertion pass for a weaker reason.

Reverted to `5` with a comment explaining why the invalid value is deliberate.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.test.tsx
```diff
@@ -469,6 +479,78 @@ describe('FitnessStatusDetail', () => {
...
 
+  it('renders the route map without a GPS trace badge', async () => {
+    renderDetail()
+
+    // Positive anchor first: the zoom control lives in the same overlay
+    // fragment the badge used to, so its presence proves the overlay really
+    // rendered and the badge assertion below is not vacuous.
+    expect(await screen.findByLabelText('Zoom in map')).toBeInTheDocument()
+    expect(screen.queryByText('GPS trace')).not.toBeInTheDocument()
+  })
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.test.tsx:490]: **Deleting the badge removed the overlay layer's only positive anchor.** This test was a purely negative assertion, and the `waitFor(() => expect(mockGetFitnessRouteData).toHaveBeenCalledTimes(1))` in front of it resolved on the first synchronous check, so it was not a barrier for the state commit. It passed for the right reason today (verified: it fails when the badge is restored), but any change to the map-panel gating or the 3-sample fixture would have made it vacuous with no signal.

Added `findByLabelText('Zoom in map')` — same JSX fragment the badge lived in — as the anchor, and added the missing negative case: nothing asserted the notice was *absent* when no segment is hidden, so a regression dropping the `hasHiddenPrivacySegments` gate would have shown "Green segments are hidden from other viewers" on every route with the suite still green.

lib/components/settings/FitnessPrivacyLocationSettings.tsx
```diff
@@ -1000,7 +1003,8 @@ export const FitnessPrivacyLocationSettings: FC<Props> = ({ mapProvider }) => {
                       {location.longitude.toFixed(6)}
                     </p>
                     <p className="text-xs text-muted-foreground">
-                      Hide radius: {location.hideRadiusMeters}m
+                      Hide radius:{' '}
+                      {formatRadiusLabel(location.hideRadiusMeters)}
```
llun, line 1007: **The saved-list call site of `formatRadiusLabel` was uncovered.** The function's two branches are covered indirectly by the option assertions, but nothing asserted this `Hide radius: …` line — and the JSX now splits into two text nodes (`Hide radius:{' '}` + label), exactly the shape that silently breaks a normalised `getByText` match later.

Added a table-driven test rendering a saved location at 200 / 1000 / legacy 20, asserting `200m` / `1km` / `50m` respectively (the last also pins the snap through the display path).

llun (commented):
## Sub-agent review — round 1

Four sub-agents reviewed the whole diff in parallel (correctness, project invariants, test coverage, UI/a11y). Findings are posted inline below; all code-level ones are already fixed in 1c9f305.

**Fixed:** derived cap constant + ascending-sort invariant; privacy-notice dismissal leaking across activity files; missing affordance/accessible name on the notice; type errors introduced in `privacy.test.ts`; six test-coverage gaps.

**Accepted with reasoning (no code change):** stale pre-generated route-map images, and the strict API rejection of legacy radii — see the threads.

**Confirmed clean by the invariant reviewer:** server/client module boundary (`privacy.ts` has no `'use client'` and no imports; the client component importing *from* it is the sanctioned direction), `useInstanceLimits` correctly *not* used (this is a per-actor preference, not an instance policy), Tailwind v4 variable syntax, logging/response/Zod rules, and documentation (zero markdown hits for the old radius values or "GPS trace").

---

### Finding that has no line in this diff to anchor to

`app/api/v1/fitness/general/route.ts` is unchanged by this PR, so GitHub will not take an inline comment on it.

**Accepted, not fixed — the write path stays strict while the read path snaps.** `PrivacyRadiusSchema` / `PrivacyLocationRadiusSchema` still require exact membership, so a POST carrying a legacy `20` gets a bare 422 rather than being snapped.

This is deliberate and consistent in practice: GET snaps, so the settings UI always round-trips a legal value, and this route exports no `OPTIONS` and sits behind `AuthenticatedGuard` — per `REVIEW.md` it is consumed exclusively by the internal web client. The only exposure is a browser tab loaded before the deploy. Making the write path snap too would mean silently persisting something other than what was sent, which is worse than an explicit 422.

This is also why the version bump is `minor:` and not `major:` — no documented integration contract changes, and no migration or operator action is required.

lib/services/fitness-files/privacy.ts
llun: Fixed in 1c9f305 — cap derived via `FITNESS_PRIVACY_RADIUS_OPTIONS[length - 1]`, ascending-sort and cap invariants both asserted in `privacy.test.ts`.

lib/services/fitness-files/privacy.ts
llun: Accepted as a documented limitation, no code change. Recorded in the PR body under Known limitations, with the existing "Regenerate maps for old statuses" button as the remedy.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:890]: Fixed in 1c9f305 — reset effect on `routeSegments`. Regression test added and verified to fail without the effect.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1216]: Fixed in 1c9f305 — `cursor-pointer hover:bg-muted`, `X` icon, and an action-naming `aria-label`. Focus-restoration deliberately not added; reasoning in the comment above.

lib/services/fitness-files/privacy.test.ts
llun: Fixed in 1c9f305 — fixture rescaled to 0.002/0.003 offsets; `tsc --noEmit` is now clean for this file.

app/api/v1/fitness/general/route.test.ts
llun: Fixed in 1c9f305 — table-driven 422 cases for 5/10/20/75/2000 plus a 200 case for 1000.

app/api/v1/fitness/general/route.test.ts
llun: Fixed in 1c9f305 — GET test with a stored 20m zone asserting it snaps to 50m.

app/api/v1/fitness/general/route.test.ts
llun: Fixed in 1c9f305 — reverted to `5` with a comment explaining the invalid value is deliberate.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.test.tsx:490]: Fixed in 1c9f305 — zoom-control anchor added, plus the missing 'absent when nothing is hidden' case.

lib/components/settings/FitnessPrivacyLocationSettings.tsx
llun: Fixed in 1c9f305 — table-driven saved-list test at 200 / 1000 / legacy 20.

llun:
## Sub-agent review — round 2 (clean)

Re-ran the review after the round-1 fixes, since fixes can introduce new problems. Two sub-agents: correctness/regression-hunting on the round-1 changes, and a dedicated silent-failure/fail-open pass (this is a privacy control, so a fail-open here de-anonymises a home address).

### No new issues in this change

The correctness pass verified each round-1 fix rather than taking it on trust:

- **The reset effect is safe.** `routeSegments` is parent `useState` set only inside the `[fitness?.id]` effect, and is passed as the state variable (not an inline literal) at both call sites — so its identity is stable across the hover-driven re-renders that fire several times a second on the Analysis tab. `setIsPrivacyNoticeDismissed(false)` while already `false` is an `Object.is`-equal update, so React bails out: no loop, no extra render, no spurious re-show. `drawableRouteSegments` would be exactly equivalent (it's a `useMemo` on the same value).
- **The derived cap still types correctly.** Indexing a `readonly` tuple with a `number` yields the member union, so `MAX_FITNESS_PRIVACY_RADIUS_METERS` is `0 | 50 | 100 | 200 | 500 | 1000`, not widened to `number`, and `sanitizePrivacyRadiusMeters`'s return type still holds. `tsc` is clean for every changed non-test file and for `privacy.test.ts`.
- **The rescaled fixture was independently recomputed** against the module's own haversine: `52.00001` → 1.31m (hidden), `52.002` → 261.16m (visible), `52.003` → 391.73m (visible). The rescale was *necessary*, not cosmetic — the old offsets measure 26.12m and 39.17m, both inside the new 50m zone, so the old fixture would have produced zero visible segments.
- **The new 422 assertions are meaningful.** `vi.clearAllMocks()` runs before the `mockResolvedValue` setup and only clears call logs, and `getFitnessSettings` defaults to `null`, so an accepted radius necessarily reaches `createFitnessSettings` — the assertion really does discriminate accept from reject.
- **`X` import is correct** and `Route` is still used at line 1922, so removing the badge orphaned nothing.

The fail-open pass confirmed the important invariants: **the snap can never shrink a stored zone** (`find(option => option >= value)` is always ≥ input; the cap-down branch is unreachable because no write path can persist > 1000), the `hideRadiusMeters: "20"` string concern does not materialise (SQLite TEXT and PG `jsonb` both land on numbers), and the route-data endpoint **fails closed** — a settings-lookup throw returns 500, and `isOwnerAccount` is false on any lookup miss, so hidden segments are dropped.

### Pre-existing issues found nearby — reported, not fixed here

Both verified against `origin/main` as untouched by this PR. Flagging rather than fixing, to keep this PR scoped to the radius/overlay change:

1. **`lib/database/sql/fitnessSettings.ts:92` swallows a JSON parse failure into `[]` with no logging.** If `privacyLocations` ever holds a non-JSON string, every read returns `[]`, `getFitnessPrivacyLocations` falls through to the legacy single-zone path, and a user with five configured zones silently drops to one — the other four published unmasked, with no signal to user or operator. The catch is also broader than needed: only `JSON.parse` can throw there. Worth narrowing it and adding a `logger.error`.

2. **A failed settings load can be laundered into a destructive save** in `FitnessPrivacyLocationSettings.tsx`. If the GET 500s, `privacyLocations` stays `[]`, the auto-locate effect fires *because* the list is empty and prefills coordinates, and Save / Clear all are gated only on `isLoading || isSaving || …` — all false. Saving then POSTs a list built from the empty state and `updateFitnessSettings` replaces the whole JSON array, destroying every previously configured zone. The only signal was a load-error banner that `handleSave` clears before the write. A `hasLoadedSettings` flag gating both buttons would close it.

3. **Persisting the snap** (rather than only applying it on read) would be a stronger fix than the read-time patch: as it stands the stored value keeps saying `20` forever, so if a future change ever relaxes the snap those zones silently revert. A one-shot data migration over `privacyLocations` / `privacyHideRadiusMeters`, plus queuing map regeneration, is the complete version of this change — deliberately left out of scope here and captured under Known limitations in the PR body.

### Loop status

Round 1: 11 findings, all addressed (10 fixed in 1c9f305, 1 accepted with reasoning). Round 2: no new issues in the diff. Stopping at 2 rounds per the `AGENTS.md` loop — a clean round ends it. Every round-1 thread is replied to and resolved.

PR #1314, opened 2026-07-25
fix: stop fitness privacy locations vanishing silently on a parse failure

llun:
Follow-up 1 of 3 from the review of #1313. Each finding gets its own PR.

## The problem

`parseStoredPrivacyLocations` in `lib/database/sql/fitnessSettings.ts` wrapped both the JSON parse *and* the sanitizer in a single catch that returned `[]` with nothing logged:

```ts
try {
  const parsedValue =
    typeof value === 'string' ? getCompatibleJSON<unknown>(value) : value
  return sanitizePrivacyLocationSettings(parsedValue)
} catch {
  return []
}
```

`[]` means "this actor has no privacy zones". So if the column ever fails to parse — a manual DB edit, a restore from a mis-typed dump, a future write path that forgets `JSON.stringify` — every read reports no zones and republishes the route segments those zones were hiding.

It degrades worse than it looks. `getFitnessPrivacyLocations` falls through to `getLegacyFitnessPrivacyLocation`, which reconstructs at most **one** zone from `privacyHomeLatitude`/`privacyHomeLongitude`. An actor with five configured zones silently drops to one; the other four homes or workplaces are published unmasked. The settings page renders the reduced list without comment, and nothing reaches the log or an error tracker.

## The change

- Narrow the catch to the parse. `sanitizePrivacyLocationSettings` is total — it cannot throw — so wrapping it only creates a place for a real bug to hide.
- Skip the guard entirely on the non-string path: SQLite hands back the `json` column as TEXT, PostgreSQL `jsonb` arrives already parsed, and only the string branch can throw.
- `logger.error` on parse failure, so an operator can see it and restore the column.

## What this deliberately does not do

It still returns `[]` rather than throwing. Throwing would be strictly more fail-closed — `getFitnessSettings` would propagate, the route-data endpoint would 500, and nothing would be served unmasked — but it would take out the whole fitness surface for that actor (status rendering, heatmaps, import jobs) with no self-service recovery, for a corruption that our own writes should never produce. Logging makes the failure visible without that blast radius. Happy to switch it if you would rather fail closed.

## Testing

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**712 files / 6864 tests**) all green.
- New `parseStoredPrivacyLocations` tests: one asserts the error is logged and the result is `[]` for unparseable JSON — confirmed to **fail** against the pre-change code; one asserts that valid JSON the sanitizer merely rejects does **not** log, pinning the narrowed catch so a future over-broad catch is caught.

### DIFF
--- a/lib/database/sql/fitnessSettings.test.ts
+++ b/lib/database/sql/fitnessSettings.test.ts
@@ -1,7 +1,6 @@
-import {
-  getTestSQLDatabase,
-  getTestSQLDatabaseWithInstance
-} from '@/lib/database/testUtils'
+import { parseStoredPrivacyLocations } from '@/lib/database/sql/fitnessSettings'
+import { getTestSQLDatabase } from '@/lib/database/testUtils'
+import { SQLFitnessSettings } from '@/lib/types/database/fitnessSettings'
 import { logger } from '@/lib/utils/logger'
 
 describe('FitnessSettings database operations', () => {
@@ -745,55 +744,23 @@ describe('FitnessSettings database operations', () => {
 })
 
 describe('parseStoredPrivacyLocations', () => {
-  const actorId = 'test-actor-corrupt-privacy'
+  const context = { actorId: 'actor-1', serviceType: 'general' }
+  const zone = { latitude: 13.7563, longitude: 100.5018, hideRadiusMeters: 200 }
 
-  const setup = async () => {
-    const { database, instance } = getTestSQLDatabaseWithInstance()
-    await database.migrate()
-    await database.createActor({
-      actorId,
-      username: 'corruptprivacy',
-      domain: 'test.example.com',
-      inboxUrl: `https://test.example.com/users/corruptprivacy/inbox`,
-      outboxUrl: `https://test.example.com/users/corruptprivacy/outbox`,
-      followersUrl: `https://test.example.com/users/corruptprivacy/followers`,
-      sharedInboxUrl: 'https://test.example.com/shared/inbox',
-      publicKey: 'test-public-key-corruptprivacy',
-      privateKey: 'test-private-key-corruptprivacy',
-      createdAt: Date.now()
-    })
-    await database.createFitnessSettings({
-      actorId,
-      serviceType: 'general',
-      privacyLocations: [
-        { latitude: 13.7563, longitude: 100.5018, hideRadiusMeters: 200 }
-      ]
-    })
-    return { database, instance }
-  }
-
-  it('logs an error when the stored privacy locations cannot be parsed', async () => {
-    const { database, instance } = await setup()
+  it('logs an error and reports no zones when the stored JSON cannot be parsed', () => {
     const loggerErrorSpy = vi
       .spyOn(logger, 'error')
       .mockImplementation(() => undefined)
 
     try {
-      await instance('fitness_settings')
-        .where({ actorId, serviceType: 'general' })
-        .update({ privacyLocations: 'not-json{' })
-
-      const settings = await database.getFitnessSettings({
-        actorId,
-        serviceType: 'general'
-      })
-
-      // Degrading to "no privacy zones" republishes every segment those zones
-      // were hiding, so it must never be silent.
-      expect(settings?.privacyLocations).toEqual([])
+      // Reporting "no privacy zones" republishes every route segment those
+      // zones were hiding, so it must never be silent.
+      expect(parseStoredPrivacyLocations('not-json{', context)).toEqual([])
       expect(loggerErrorSpy).toHaveBeenCalledWith(
         expect.objectContaining({
-          message: 'Failed to parse stored fitness privacy locations'
+          message: 'Failed to parse stored fitness privacy locations',
+          actorId: 'actor-1',
+          serviceType: 'general'
         })
       )
     } finally {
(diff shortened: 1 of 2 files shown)


llun (commented):
## Sub-agent review — round 1

Two sub-agents (fail-open hunting, project conventions). Three findings, all fixed in the follow-up commit.

**Fixed:**
1. The catch was not actually narrowed — the sanitizer was still inside the `try` on the string path, so the documented invariant held only on the PostgreSQL branch. On SQLite (local dev, the tests, and SQLite deployments) a sanitizer throw would still be swallowed *and* mislabelled as a parse failure.
2. The log line had no `actorId`/`serviceType`, so an operator had no way to find the offending row.
3. The tests drove the helper through a real database, which can only reach the SQLite TEXT branch — the already-parsed PostgreSQL path this PR touched had **zero** coverage — and leaked an in-memory SQLite handle per test. Replaced with direct unit tests on the now-exported helper, which covers both branches and leaks nothing.

**Confirmed by the fail-open reviewer:** `sanitizePrivacyLocationSettings` is total for every value that can reach it (`Array.isArray` gate, flat iteration, guarded property reads, no recursion), and hostile objects cannot be smuggled in because both backends produce plain `JSON.parse` output. The `getCompatibleJSON` `?? ({} as T)` fallback was already dead on this path before the change, so the restructure alters no behavior.

**Not changed — logger key convention.** One reviewer suggested `err` instead of `error`, because `lib/utils/logger.ts`'s `log` hook reads `logObject.err?.stack` to attach `stack_trace` for Cloud Error Reporting. That is a genuine observation, but all 12 existing `logger.error` call sites in `lib/` and `app/` use `error`, and it is the form `AGENTS.md` documents. Switching one call site would just make it inconsistent — worth raising as a repo-wide question instead.

llun (commented):
## Sub-agent review — round 2

Two findings, both real, both fixed.

**1. The log line serialized to nothing.** The reviewer ran the repo pino with this exact payload and got `"error":{}`. I verified independently: an Error's `message` and `stack` are non-enumerable, so `JSON.stringify({error: new SyntaxError("…")})` → `{"error":{}}`. The whole purpose of this line is telling an operator *why* the column failed to parse, and it was telling them nothing.

This reverses my round-1 call. I had kept `error` to match the 25-odd existing call sites, but consistency with a convention that discards the error is not worth having — and `lib/utils/logger.ts`'s own formatter reads `logObject.err?.stack` to emit `stack_trace`, with precedent at `lib/services/auth/getSession.ts:39` and `lib/services/wellknown/nodeinfo.ts:95`. Switched to `err`, with a comment recording why. The test now asserts `err: expect.any(SyntaxError)` so it cannot regress.

**2. The narrowing guard could not fail.** The reviewer reverted the sanitizer back inside the `try` and all 6 tests still passed — `[{ latitude: "nope" }]` parses fine and the sanitizer returns `[]` without throwing, so both shapes behaved identically. The test pinned nothing.

Replaced with one that forces a sanitizer failure through a mocked module and asserts the throw *surfaces* rather than being swallowed and mislabelled. Verified: reverting the narrowing now fails that test specifically (1 failed / 37 passed).

**Verified clean this round:** all 3 call sites updated with no other callers; the `let parsedValue` restructure is behaviourally identical for every input shape (string→array, string→non-array, already-parsed array/object, null/undefined); no new lint or tsc errors; the DB-handle leak is gone entirely since these are now pure unit tests.

Gate: prettier, lint (0 errors), build, **712 files / 6869 tests** green.

llun:
## Sub-agent review — round 3 (clean)

Final confirmation round, targeted at what the round-2 fixes could have broken. **No defects found.**

Verified this round:

- **The file-wide `vi.mock` does not contaminate the other 30+ tests.** `importOriginal` + spread preserves every export (the module is all named consts, no default, no side effects), the toggle initialises to `null` so it is off everywhere except the one test, and the reset sits in a `finally` that runs even when the `expect(...).toThrow` assertion itself fails. `vitest.config.ts` sets no `restoreMocks`, so the explicit `mockRestore()` calls are required and present. The wider fitness suite (14 files / 158 tests) passes.
- **`vi.hoisted` is the right tool and matches existing practice** — same pattern as `lib/database/sql/collectionFeedCap.test.ts` and five other `importOriginal` files. `AGENTS.md`'s `vi.importMock` rule governs *reading* a mocked module; *defining* one with `vi.mock(path, factory)` is separately endorsed.
- **The `err` switch was reproduced against the real pino config**: `err` emits `{"err":{"type":"SyntaxError","message":…,"stack":…},"stack_trace":…}`, `error` emits `"error":{}`. Nothing in the repo consumes an `error` key for this call.
- **The narrowing invariant is genuinely pinned** — moving the sanitizer back inside the `try` fails exactly one test.
- Lint 0 errors; the only tsc hits in the changed files are the standard `importOriginal` baseline artifacts that appear identically in all five pre-existing files using that pattern.

### Loop status
Round 1: 3 findings, all fixed. Round 2: 2 findings, both fixed (one reversed a round-1 decision of mine after the reviewer produced evidence the log line serialized to `{}`). Round 3: clean. Stopping at 3 rounds.

Gate: prettier, lint (0 errors), build, 712 files / 6869 tests — all green.

PR #1316, opened 2026-07-25
fix: stop a failed settings load from wiping saved fitness privacy zones

llun:
Follow-up 2 of 3 from the review of #1313.

## The problem

A read failure could be laundered into a destructive write, silently destroying every privacy zone an actor had configured. The chain, all in `lib/components/settings/FitnessPrivacyLocationSettings.tsx`:

1. `GET /api/v1/fitness/general` 500s → `getFitnessGeneralSettings` throws → the catch sets an error banner. `privacyLocations` stays `[]` and `isLoading` goes false.
2. The auto-locate effect fires **precisely because** `privacyLocations.length === 0` and both coordinate inputs are empty, prefilling latitude/longitude from the browser. The form now presents as a fresh, unconfigured one with a live location already in it.
3. Save and Clear all are gated only on `isLoading || isSaving || isRegeneratingMaps || isLocatingCurrentPosition` — all false. Both are clickable.
4. `handleSave` builds `locationsToSave` from the empty `privacyLocations` plus the prefilled draft and POSTs it. `updateFitnessSettings` **replaces the entire stored JSON array**, so every previously configured zone is gone. "Clear all" posts `[]` and wipes them outright.

From that point `getFitnessPrivacyLocations` returns at most the one new zone, and every subsequent route-data response and regenerated map is unmasked for the destroyed ones. The user's only warning was a banner that `handleSave` clears before the write.

## The change

- Track `hasLoadedSettings`, set only after a successful read, and require it for **Save** and **Clear all**.
- Gate the auto-locate prefill on it too, so a failed load can never dress an empty form up as a configured one.
- Add a **Retry loading** button so the disabled state is recoverable without a full page reload (the fetch effect is now keyed on a reload token).
- Say in the error message *why* saving is disabled, rather than leaving the user with two dead buttons and no explanation.

`Regenerate maps for old statuses` is deliberately left ungated — it is non-destructive and reads server-side state.

## Testing

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**712 files / 6872 tests**) all green.
- Three new tests, each verified to **fail** against the pre-change component:
  - Save and Clear are disabled after a failed load.
  - Coordinates are not prefilled after a failed load. This one needed a resolving MapLibre stub — the auto-locate effect is gated on `isMapReady`, and with the default never-resolving loader the test would have passed for the wrong reason. It asserts on the form fields rather than on `getCurrentPosition`, because the map's own initial-view lookup calls geolocation through a different path.
  - Retrying after a failure restores the saved zones and re-enables Save.

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -2446,6 +2446,25 @@ export interface RegenerateFitnessMapsResponse {
   queuedCount?: number
 }
 
+// A bare cast would let any 200 with any body count as a loaded settings
+// object. `privacyLocations` would then read as an empty list, and because a
+// save REPLACES the whole stored list, the next save would wipe every zone the
+// actor configured. Reject the body instead so the caller's load-failure path
+// handles it.
+const assertFitnessGeneralSettings = (
+  value: unknown
+): FitnessGeneralSettingsResponse => {
+  if (
+    !value ||
+    typeof value !== 'object' ||
+    !Array.isArray((value as FitnessGeneralSettingsResponse).privacyLocations)
+  ) {
+    throw new Error('Unexpected fitness privacy settings response')
+  }
+
+  return value as FitnessGeneralSettingsResponse
+}
+
 export const getFitnessGeneralSettings =
   async (): Promise<FitnessGeneralSettingsResponse> => {
     const response = await fetch('/api/v1/fitness/general', {
@@ -2459,7 +2478,7 @@ export const getFitnessGeneralSettings =
       throw new Error('Failed to load fitness privacy settings')
     }
 
-    return (await response.json()) as FitnessGeneralSettingsResponse
+    return assertFitnessGeneralSettings(await response.json())
   }
 
 export const updateFitnessGeneralSettings = async (
@@ -2476,7 +2495,15 @@ export const updateFitnessGeneralSettings = async (
   })
 
   const data = (await response.json()) as FitnessGeneralSettingsResponse
-  return { ok: response.ok, data }
+
+  // On success the caller rebuilds its list from this body, so an unrecognised
+  // 200 must not be reported as a save — it would blank the list under a
+  // "saved" message and invite the user to save the emptiness for real.
+  if (response.ok) {
+    return { ok: true, data: assertFitnessGeneralSettings(data) }
+  }
+
+  return { ok: false, data }
 }
 
 export const regenerateFitnessMaps = async (): Promise<{
--- a/lib/components/settings/FitnessPrivacyLocationSettings.test.tsx
+++ b/lib/components/settings/FitnessPrivacyLocationSettings.test.tsx
@@ -28,6 +28,12 @@ describe('FitnessPrivacyLocationSettings', () => {
 
   afterEach(() => {
     vi.restoreAllMocks()
+    // `restoreAllMocks` does not reset an implementation set on a vi.fn() from
+    // a module mock factory, so a test that makes the GL loader resolve would
+    // otherwise leak a live map into every later test in this file.
+    vi.mocked(loadMaplibreModule).mockImplementation(
+      () => new Promise(() => {})
+    )
     Object.defineProperty(global.navigator, 'geolocation', {
       configurable: true,
       value: originalGeolocation
@@ -360,8 +366,11 @@ describe('FitnessPrivacyLocationSettings', () => {
     })
 
     const postBodies = fetchMock.mock.calls
-      .filter(([, init]) => {
-        return (init?.method ?? 'GET') === 'POST'
+      .filter(([input, init]) => {
+        return (
+          input === '/api/v1/fitness/general' &&
+          (init?.method ?? 'GET') === 'POST'
+        )
       })
       .map(([, init]) => JSON.parse(String(init?.body)))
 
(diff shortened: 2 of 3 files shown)


llun (commented):
## Sub-agent review — round 1

Two sub-agents (conventions/React correctness, data-loss hunting). Seven findings; five fixed here, two deferred with reasoning.

### Fixed — introduced by this PR

1. **The guard explanation lived in the erasable `error` slot.** Every action handler calls `setError(null)`, so clicking anything wiped the banner and left the user with a greyed-out Save and no on-screen reason. The banner is now derived from `!isLoading && !hasLoadedSettings` — the same expression that renders Retry — so nothing can clear it.
2. **Gating was asymmetric.** `Add location to list`, the inputs, the radius select and `Use current location` stayed enabled, so the user could build a list and be told *"Privacy location added to list. Save settings to apply."* against a permanently disabled Save. All of them now share one `isEditingDisabled` flag. This also closes the follow-on finding that a successful retry silently discarded whatever the user had added while unloaded.
3. **`isHydratingSettingsRef` was never re-armed**, so a retry hydrated differently from a first load (map flying to the marker mid-hydration). Now set at the start of every attempt.
4. **The MapLibre mock leaked across tests.** `vi.restoreAllMocks()` only undoes `spyOn`, not an implementation set on a `vi.fn()` from a module factory — so my resolving-loader stub turned every later test in the file into one running against a live map. Reproduced by moving the new describe to the top (an existing test failed), fixed by reinstating the never-resolving loader in `afterEach`, and re-verified: the reordering now passes 17/17.
5. A body-less `regenerate-maps` POST could land in the pre-existing clear-all collector, which matched *any* POST and then `JSON.parse(undefined)`. Scoped it to the settings endpoint and made the banner test await its own request settling.

### Fixed — pre-existing, but the same hole this PR targets

6. **Neither response body was validated.** `getFitnessGeneralSettings` and `updateFitnessGeneralSettings` used bare casts, so any 200 with any body counted as success. A route-shape change or a proxy rewriting the body would present as "no privacy locations", with Save enabled — the identical data-loss bug this PR exists to close, just reached through a 200 instead of a 500. Worse on the POST path: an unrecognised 200 blanked the list *under a "settings saved" message*, and the user's natural next click would save the emptiness for real. Both now reject the body and route into the load-failure path. New test covers the 200-with-junk case.

### Deferred, with reasoning

7. **A degraded server-side parse is invisible to the client.** When `parseStoredPrivacyLocations` hits corrupt JSON it returns `[]` (and logs — #1314), the API falls back to the single legacy zone, and the client sees an ordinary 200 with one zone. Closing this needs a `privacyLocationsDegraded` flag threaded through the GET response and the API contract — a server-side change beyond this component fix. Worth its own PR; flagging rather than smuggling it in.
8. **The load `catch` discards the error**, so a 307 to `/auth/signin` (session expired) surfaces as "failed to load" with a Retry that can never succeed, instead of "sign in again". Real, but it is an error-classification improvement rather than a data-loss path, and it needs a shared approach across this file's three bare catches.

Gate: prettier, lint (0 errors), build, **712 files / 6875 tests** green.

llun:
## Sub-agent review — round 2 (clean)

**No defects at reportable confidence.** The reviewer mutation-tested each round-1 fix and confirmed every one is pinned by a specific failing test:

- removing `!hasLoadedSettings` from the prefill gate → the prefill test fails
- reverting `isEditingDisabled` to `isLoading || isSaving` → 3 tests fail
- reverting both `lib/client.ts` assertions → the unrecognised-200 test fails

Also verified: `assertFitnessGeneralSettings` cannot cause a spurious failure — every 200 body from `app/api/v1/fitness/general/route.ts` is built through `toPrivacyLocationsResponse`, whose `privacyLocations` is unconditionally an array, and the component is the only caller of either client function. `isEditingDisabled` is byte-equivalent to the old expressions once loaded, so nothing became newly disabled during normal operation or regeneration. The banner cannot flash (`isLoading` starts `true`, so SSR and first render both fail its condition — no hydration mismatch). `isHydratingSettingsRef` is reset by an unconditional `finally`, and overlapping attempts are impossible since Retry only renders while `!isLoading`.

### Two sub-threshold notes acted on anyway

- **The map click still wrote into the disabled inputs** after a failed load — no data-loss risk since Add/Save/Clear are all gated, but it contradicted the "disables the whole editing surface" claim. Now guarded via a ref, since the handler is registered once inside the init effect and would otherwise close over stale state.
- **`expect(loadMaplibreModule).toHaveBeenCalled()` could not fail** — there is no `clearMocks` in the vitest config, so call history carries over between tests in the file. Removed; the `getCurrentPosition` wait that follows is the real sync point.

Left alone: the POST-side branch of `assertFitnessGeneralSettings` has no direct test, because Save is disabled once a load fails so the component cannot reach it. It is defensive code and the GET path — the one the original bug travelled — is covered.

### Loop status
Round 1: 7 findings (5 fixed, 2 deferred with reasoning). Round 2: clean, plus 2 optional polish items taken. Stopping at 2 rounds.

Gate: prettier, lint (0 errors), build, 712 files / 6875 tests — all green.

PR #1317, opened 2026-07-25
fix: persist the fitness privacy radius snap instead of only applying it on read

llun:
Follow-up 3 of 3 from the review of #1313 — the last outstanding finding.

## The problem

#1313 changed the hide-radius options from 5/10/20/50m to 50/100/200/500m/1km and made `sanitizePrivacyRadiusMeters` snap an unlisted radius **up** to the next supported option, so a zone saved at 20m already masks 50m everywhere the app reads it.

But the snap runs on **read** only. The stored row still says `20`, which means:

- stored and effective values disagree permanently, and nothing ever reconciles them; and
- if a future change relaxes the snap or reintroduces a sub-50m option, every one of these zones silently reverts to a radius the product no longer considers private — with no migration to point at.

## The change

A data-only migration that backfills both `privacyHideRadiusMeters` and the per-entry `hideRadiusMeters` inside the `privacyLocations` JSON.

Snapping upward **never shrinks a zone**, so this can only increase how much is hidden — it cannot expose a GPS point that was previously masked.

Following the repo's backfill convention (`20260611090000_lowercase_account_emails.js`): self-contained with no app imports, keyset-paginated in chunks of 500 so peak memory stays bounded, and wrapped in one transaction so it is atomic. Re-running is a no-op.

`down` is a deliberate no-op. The original sub-50m radii are not recoverable from the snapped values, and restoring them would shrink zones below the enforced minimum — re-exposing GPS points around a home address. Rolling back leaves the widened radii in place, which is the safe direction.

## Not included

**Map images and heatmaps are still not re-rendered.** Migrations run outside the worker and have no queue access, so regeneration stays a deliberate action via the existing "Regenerate maps for old statuses" button on the fitness privacy settings page. There is an operator note to that effect at the top of the migration.

## Testing

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**713 files / 6878 tests**) all green.
- **Functionally verified** against a throwaway local SQLite by seeding rows and running the migration:

  | case | before | after |
  |---|---|---|
  | legacy radius + mixed legacy locations | `20`, entries at `20` and `5` | `50`, both entries `50` |
  | already valid | `50` | unchanged |
  | disabled | `0` | `0` — stays disabled |
  | `null` radius, unparseable JSON | `null`, `not-json{` | untouched, left exactly as found |
  | legacy radius, empty list | `5`, `[]` | `50`, `[]` |

  Re-running produced no further changes (idempotent).
- **Both reference schema dumps regenerated** — against a throwaway local SQLite and a throwaway `postgres:17` container — and confirmed byte-identical, since this migration adds no DDL. `knex_migrations` = 142 on both, matching the 142 files in `migrations/`.
- New guard test: the migration duplicates the option list (migrations must not import app code), so a test fails if it drifts from `FITNESS_PRIVACY_RADIUS_OPTIONS`. Verified to fail when the lists are made to disagree.


llun (commented):
## Sub-agent review — round 1

Two sub-agents (migration correctness/conventions, data-destruction hunting). Both independently landed on the same headline problem.

### The big one: the tests never ran the migration

All three assertions read the migration as **text**. One reviewer mutation-tested it — changing the snap to round to the wrong option left **all 3 green** — and the second confirmed test 2 passes even if the migration file is deleted entirely, because it re-implements the snap inline and then asserts a property of its own local expression. The JSON round-trip, the keyset loop, and idempotency had **zero** behavioural coverage, on a migration that rewrites production privacy data.

I had also missed that the repo already has the right pattern: ten `lib/database/sql/*Migration.test.ts` files import the migration and run `up()` against in-memory SQLite — including `lowercaseAccountEmailsMigration.test.ts`, the very migration this one is modelled on. Replaced with a behavioural test in that directory. It now catches a snap-direction mutation (5 failures) and the entry-guard regression below (1 failure).

### Real defect that gap was hiding

`snapLocations` **destroyed an entry radius it could not interpret**. `snapRadius` returns `0` for a non-number, and unlike the column-level call site the entry-level one had no guard — so a stored `hideRadiusMeters: "20"` was rewritten to `0`, meaning "no zone", permanently. No live impact today (the runtime already drops such entries), but it directly contradicts the file's own stated rule — "never turn a recoverable column into a lost one" — and would defeat a later fix that coerces numeric strings. Both call sites now leave unknown values exactly as found.

### Also fixed

- **The drift guard forced edits to an already-applied migration.** The header argues the option list is duplicated *because* migrations must keep working when app code changes — then the test asserted that copy must track `privacy.ts` forever. Editing an applied migration is a guaranteed no-op (deployments never re-run it, a fresh DB has no rows) whose only real effect is to make migrate history irreproducible. The expectation is now a frozen literal, with a comment saying to add a new migration instead.
- **"Snapping upward never shrinks a zone" was stated absolutely and is false above the largest option** (5000 → 1000, verified by running it). Unreachable today — the API never accepted a radius above 50 — but the claim is now qualified.
- **The migration was completely silent.** It now logs how many rows it snapped and warns with the ids of rows whose `privacyLocations` could not be parsed. Those actors already read as having *no* privacy zones at runtime, and this is the one moment every row gets scanned — the cheapest possible place to tell an operator.
- Added the `rows.length < CHUNK_SIZE` break the sibling backfill has (saves a round-trip, and closes a theoretical non-termination if an externally-inserted row carried a NULL id).

### Confirmed sound by both reviewers

`JSON.stringify` into a `jsonb` column does **not** double-encode — node-pg binds it as an untyped text param and PG infers the type from the column, which is exactly what the app already does on this same column. Keyset pagination is collation-safe and the PK is never updated, so no skips or repeats. Every write is inside the transaction. Unknown fields on a location entry survive the rewrite (verified with a `label` field). Array length and `null` entries are preserved. The empty `down` has precedent and `migrate:down` rolls back cleanly.

### Noted, not acted on

A read-then-write across the transaction could recompute over a concurrent save. The window is sub-second per row and it is the same shape as the existing lowercase-emails backfill, so it is documented in the header as "run before the new version serves traffic" rather than adding partial compare-and-set semantics.

The chunk-loop break is defensive and is **not** covered by a test — the loop terminates via the empty-page check regardless, so I am not claiming coverage it does not have.

Gate: prettier, lint (0 errors), build, **713 files / 6887 tests**. Migration chain re-run and both schema dumps re-verified byte-identical.

llun (commented):
## Sub-agent review — round 2

Three findings, all in code I added in round 1. All fixed. **The data path itself was confirmed sound** — no way found for a row to end up with less privacy than before.

### The warning I added was wrong in a way that would cause a false emergency

Two separate errors, both verified:

1. **It reported the settings-row UUID, not the actor.** `fitness_settings.id` is a `crypto.randomUUID()` unrelated to `actors.id`. An operator pasting it into an actor lookup finds nothing. Now selects and reports `actorId`.
2. **"these actors currently have NO privacy zones at runtime" is false whenever the row still has legacy home coordinates.** `parseStoredPrivacyLocations` returns `[]` for an unusable column, and `getFitnessPrivacyLocations` then falls back to the legacy `privacyHomeLatitude`/`Longitude` zone. I confirmed this directly: such a row still yields `[{ lat: 1, lng: 2, radiusMeters: 50 }]` — the actor is fully protected. My warning would have named them as actively exposed, in a privacy migration's own output, which is exactly the thing an operator would act on immediately.

The message now states only what the migration can actually see, and caps the id list at 20 rather than emitting one unbounded log line during a bad-data incident.

### The "frozen option list" test did the opposite of what its comment said

It compared a literal against the live `FITNESS_PRIVACY_RADIUS_OPTIONS` — so it *tracked the app constant*, precisely the coupling the comment claimed it avoided. Two consequences: a future PR that correctly adds a new option and a new migration would see this test fail and be pointed at an already-applied migration that is behaving correctly; and it never touched the migration's own copy, so a typo there passed everything, because the behavioural cases only ever exercised `5` and `20` → `50`.

Replaced with boundary cases driven through the migration itself: 1→50, 50→50, 51→100, 150→200, 250→500, 600→1000, 1000→1000, and the 5000→1000 clamp. Verified that dropping `100` from the migration's array now fails, and that reporting `row.id` instead of `row.actorId` fails.

### Confirmed sound this round

The reviewer traced the round-1 entry guard against the runtime for every input class and found no stored-vs-effective divergence: `20.5`→50 both ways, `0` unchanged, `-5` writes 0 and the runtime drops the entry either way, `'20'` left untouched and dropped at runtime either way. The short-page break was tested at 499/500/501/1000 rows with zero skips. `if (!locations && row.privacyLocations)` does not false-flag `'[]'` (an empty array is truthy). Test hygiene is fine — `destroy()` per test, layered console spies unwound by `restoreAllMocks`, and `@/migrations/*` resolves via both the tsconfig path alias and the vitest alias.

Gate: prettier, lint (0 errors), build, **713 files / 6894 tests**. Migration chain re-run; both schema dumps still byte-identical.

llun:
## Sub-agent review — round 3 (clean)

Final confirmation round, targeted at what the round-2 fixes could have broken. **No findings.**

Verified this round:

- **`actorId` is safe to select and can never print `undefined`** — `NOT NULL` in both dumps (`schema.sql:459`, `schema.sqlite.sql:48`), present since the table was created, and this migration is last by timestamp.
- **The `slice(0, 20)` suffix has no off-by-one** — exercised at n = 1 / 19 / 20 / 21 / 25: exactly-20 prints no suffix, 21 prints `(1 more)`.
- **Each boundary case gets a fresh DB** — `beforeEach` builds a new `:memory:` instance, so the repeated `id: '1'` cannot collide (a leak would trip the PK constraint immediately). 19/19 pass.
- **No dangling reference** to the removed import; lint 0 errors and zero tsc errors in either changed file.
- **Data safety re-driven end to end** through the real `up()`: negative, 0, fractional (20.5→50), exact 1000, over-max (99999→1000), null/empty locations, a null array entry, a boolean radius, and a JSON scalar. Every outcome matches what `sanitizePrivacyRadiusMeters` already returns on read, so nothing effective changes and nothing recoverable is destroyed.
- **CI's Schema Dump Sync will pass** — the reviewer replayed the exact pipeline: 142 migrations applied, `sqlite3 .schema` through CI's strip filters diffs identical to the committed dump, and the migration-count gate matches.

Two sub-threshold notes left alone, both unreachable: a non-numeric top-level `privacyHideRadiusMeters` would be written as 0, but the column is `integer` in both dialects so only non-coercible text could survive — which no future fix could recover either; and a `privacyLocations` of `''` is skipped by the warning, but the column is `NOT NULL DEFAULT '[]'` so that needs a manual DB write, and the runtime reads it as `[]` too.

### Loop status
Round 1: 6 findings (all fixed), including the headline one that the tests never executed the migration. Round 2: 3 findings, all in code I added in round 1 — including a warning that would have falsely told an operator that protected actors were exposed. Round 3: clean. Stopping at 3 rounds.

Gate: prettier, lint (0 errors), build, 713 files / 6894 tests — all green.

PR #1320, opened 2026-07-25
minor: add status emoji reactions data model and inbound federation

llun:
Epic 5.1, PR 1 of 3. Adds the storage and inbound-federation half of Misskey/Pleroma-style emoji reactions on statuses. The write API + outbound federation (5.1b) and the picker/chips UI (5.1c) follow.

## The one invariant

**A reaction is never a favourite.** It writes no `likes` row, never touches `CounterKey.totalLike`, and never moves `favourites_count`/`favourited`. Inbound `Like` **with** reaction content becomes a reaction only; a **plain** `Like` stays a favourite, unchanged.

Behaviour change worth calling out: a Misskey `❤` used to land here as an ordinary favourite and inflate `favourites_count`. It now shows as a `❤` reaction chip instead. Existing `likes` rows are **not** migrated — they are indistinguishable from genuine Mastodon favourites in storage, so the historical skew is accepted. Documented under *Intentional divergences* in the compat doc.

## What's in it

**Migration + both schema dumps.** `status_reactions`, PK `(statusId, actorId, name)` — Pleroma/glitch-style multi-reaction rather than Misskey's one-per-actor rule, so a legitimate second reaction is never silently dropped or destructively replaced. Cap of 8 distinct reactions per actor per status (glitch `MAX_REACTIONS` parity), enforced in the mixin for local *and* inbound writes; the cap drops the overflow rather than evicting an earlier reaction, which would desync us from a sender that still believes it stands. `name` is stored colon-free (unicode emoji, local shortcode, or `shortcode@domain`); `url` holds a remote custom emoji's image while local shortcodes resolve live from `customEmojis` so an admin re-upload propagates. Same migration adds nullable `notifications.reactionName`. Both `migrations/schema.sql` and `migrations/schema.sqlite.sql` regenerated from a fresh local PG 17 / SQLite run — additions only, and the PG dump reloads cleanly.

**JSON-LD (the interop safety net).** `CANONICAL_CONTEXT` gains `litepub`, `EmojiReact` and `_misskey_reaction`. Two of the new fixtures genuinely failed before the aliases: an `EmojiReact` typed with the full litepub IRI stayed an IRI, and Misskey's inline-defined `_misskey_reaction` compacted to a CURIE. A fourth case — a sender that emits `_misskey_reaction` *without* defining the term — needed more than an alias: the bundled AS2 context's `@vocab: "_:"` turns an undefined property into a blank node that `stripJsonLdArtifacts` deletes, so the reaction would have silently degraded to a favourite. `normalizeInputContext` now prepends a lowest-precedence extension-term fallback, exactly the trick already used for `security/v1`, so a sender that *does* define the terms still wins.

**Inbound, at both inboxes.**

| Delivered | Per-user inbox | Shared inbox |
|---|---|---|
| `EmojiReact` (FEP-c0e0 / litepub) | reaction | `EmojiReactionJob` |
| `Like` + `content`/`_misskey_reaction` | reaction | `EmojiReactionJob` |
| `Undo` of either | reaction removed | `EmojiReactionJob` |
| plain `Like` | favourite (unchanged) | `null` → 404 (unchanged) |

Plain-`Like` semantics at the shared inbox are deliberately untouched — see *Out of scope* below.

**Remote custom emoji + anti-spoofing.** A `:shortcode:` from a remote sender is namespaced `shortcode@senderdomain`. It is deliberately never stored bare: that would resolve against *our* `customEmojis` table and let a remote actor render any local emoji it names. The tag's `icon.url` is only stored when it is http(s), length-bounded, and served by the **sending actor's own host**; anything else is stored without a URL and renders as text (and is logged). Note this also rejects emoji served from a separate media CDN — the reaction still lands, it just renders as its name.

**Serialization.** Every `Status` carries the same rollups under both ecosystem names — `pleroma.emoji_reactions` (Pleroma/Akkoma; what Husky and the Megalodon-family clients read) and `reactions` (glitch-soc). They are derived from one rollup by one function, so they cannot disagree; a test asserts field-for-field equality. The batch path resolves the whole page — metric statuses *plus* embedded quoted statuses — in **one** grouped query seeded into `reactionsCache`; single-status routes fall back to a per-status lookup, mirroring `getStatusReblogsCount`. An `Announce` wrapper reports empty arrays and the boosted status carries the reactions on `reblog`, exactly as `favourites_count` already behaves.

**Notifications.** New internal `emoji_reaction` type, `groupKey` `emoji_reaction:<status>:<name>`, `reactionName` persisted, and the notifications page renders "reacted with 🔥 to your post".

## Deviations from the plan

1. **The `pleroma:emoji_reaction` type mapping and the `emoji` field landed here, not in 5.1b.** The plan puts them in 5.1b, but 5.1a is where the notification is first *created*, and `internalTypeToMastodon` falls through to `'mention'` by default — so deferring would have shipped a reaction notification that every client renders as a mention. ~15 lines, in both the shared mapping module and the duplicated copy inside `getMastodonNotification`.
2. **`normalizeInputContext` gained an extension-term fallback** (see above). The plan only asked for the `CANONICAL_CONTEXT` aliases; those alone cannot satisfy the plan's own fixture #4. Three existing `normalizeInputContext` expectations were updated for the extra (lowest-precedence) context entry.
3. **`deleteActorStatusReactions` is not implemented.** The plan lists it as a Misskey-style replace helper "unused until 5.1c", and 5.1c's contract is a multi-reaction toggle that does not need it. Left out rather than shipping dead code; trivial to add if 5.1c wants it.
4. **`MAX_REACTIONS_PER_ACTOR` lives in `lib/services/statuses/reactionLimits.ts`**, a dependency-free module next to `pollDurations`, so 5.1c can read the same constant client-side without crossing the server/client module boundary.
5. **Per-user inbox route tests assert dispatch, not rows.** That file mocks its database and every action; the storage semantics (including the explicit "no `likes` row, no `totalLike` bump" assertion) are covered in `lib/actions/emojiReaction.test.ts` against a real database instead.

## Federation fixture evidence

`lib/activities/jsonld/index.test.ts` — all four dialect shapes come out as bare terms after `compactActivityPub`: a litepub-context `EmojiReact`, a full-IRI `EmojiReact` (with its `Emoji` tag surviving), the FEP-c0e0 example verbatim, a real-shaped Misskey `Like` with inline-defined `_misskey_reaction`, and `_misskey_reaction` from a sender that defines nothing. `app/api/inbox/getJobMessage.test.ts` pins the shared-inbox routing table including the plain-`Like` → `null` case; `app/api/users/[username]/inbox/route.test.ts` pins the per-user Like/reaction fork in both directions.

## Verification

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **716 files / 6967 tests green**. Schema-dump sanity: `grep status_reactions` hits both dumps; the SQLite dump loads cleanly via `sqlite3`; the PG dump reloads into a fresh database with `ON_ERROR_STOP=1`.

No UI screenshots: the only visible surface this PR adds is the notification row's verb, and the reaction chips arrive in 5.1c.

## Out of scope (found while reading)

The shared inbox still returns 404 for **plain** `Like`/`Undo{Like}`. Mastodon targets personal inboxes for favourites so nothing is lost today, but any server batching them through `sharedInbox` silently loses favourites. Worth a separate audit + favourite job — this PR deliberately routes only reaction-bearing shapes.

## Review

Three rounds of the sub-agent review loop (fan out across dimensions → each finding handed to an independent skeptic instructed to refute it → address → reply → resolve). **20 findings raised, 6 real defects fixed.** All threads posted inline and resolved.

Round 1 refuted all 11 of its findings; I re-judged them myself and fixed four anyway where the refutation was right on the facts but missed a consequence — notably that a *replayed* reaction produced a fresh notification every time, which neither the finding nor its refutation had considered.

Rounds 2 and 3 found defects that survived refutation:

- **`notifications.groupKey` overflowed `varchar(255)` on PostgreSQL.** The key is a concatenation of a status-id URL and the reaction name, so bounding the name (round 1) did not bound it — my own round-1 commit message claimed it did. A verifier reproduced `ERROR: value too long for type character varying(255)` on a real PG 17. The failure mode was silent and permanent: the reaction row commits in its own transaction, the notification insert then throws, and on retry the idempotency check makes the handler early-return, so the notification is lost for good. **No SQLite test can catch this** — SQLite ignores varchar lengths — so the tests assert the length bound directly instead of relying on the database to reject it.
- **A peer could bind its own image to another instance's emoji namespace.** The image URL was validated against the *sending* host, which is right when the sender uses its own shortcode — but a reaction can arrive already namespaced, and that path took the sender's claim at face value. `evil.test` relaying `:blobcat@good.test:` with an image on `evil.test` passed the check, and since the rollup unwraps one url per `(statusId, name)` group, it could **displace** the real image good.test's own reactors stored.
- **Adding `content` to `Like` as a bare `z.string()` could break plain favourites.** A sender whose document carries a default language has `content` compacted to a language-tagged object; the whole activity would then fail the union and 400 — a regression on favourites that federated fine before this epic. All three added fields now degrade to `undefined` instead, per the AGENTS.md rule that inbound schemas stay liberal and narrow at the consumption boundary.
- Plus: reactions were not removed by `deleteStatus`/`deleteActorData`, a colon-*less* shortcode escaped namespacing and resolved against our own `customEmojis`, and the derived `shortcode@domain` name was unbounded.

I also caught one in my own fix that no reviewer raised: I had been reading `.onConflict().ignore()`'s return value, which on PostgreSQL is the raw driver `Result` — **always truthy** regardless of rows inserted. Removed; the in-transaction pre-check is authoritative and the residual race is documented.

Consciously **not** fixed, each with reasoning in the resolved threads: the advisory per-actor cap race (an attacker can register N actors anyway, and locking would serialise the hot inbound path), and the unbounded count of *distinct* reaction names per status — the plan scopes a per-status cap out of v1, serializer truncation would break the `me` flag for a viewer whose reaction falls past the cut, and a per-group SQL `LIMIT` needs window functions the repo's MySQL-compat guidance discourages. That one is now recorded as a **Known limitation** in the compat doc rather than silently accepted.

## Cross-backend verification

The repo's pg test harness races on a shared database name (CI pins SQLite), so I exercised the reaction data layer directly against a local PostgreSQL 17: create returns `true`/`false`/`false` for first/repeat/unknown-status, delete returns `true`/`false`, rollups group with the right `count`/`me`/first-reaction order, local shortcodes resolve from `customEmojis` while remote ones keep their stored url, and the per-actor cap holds at 8 — identical to SQLite. Both schema dumps were regenerated from a fresh local PG 17 / SQLite migration run, and the PG dump reloads cleanly with `ON_ERROR_STOP=1`.


llun (commented):
Automated sub-agent review round 1 — four dimension reviewers (data layer · federation/JSON-LD · serializer+notifications · tests/security/docs), each finding then handed to an independent skeptic instructed to refute it.

**11 findings raised, 0 survived adversarial refutation.** The verifiers default to refuting, so I did not treat that as a clean bill of health — I re-judged each myself and acted on six in 377dd593 (full gate re-run green: 716 files / 6975 tests). Dispositions are in the inline threads.

One finding could not be anchored inline because its file is not in this diff:

**`lib/services/notifications/getNotificationGroup.ts` — the v2 grouped-notifications entity carries no reaction emoji**, so a `pleroma:emoji_reaction` group is unrenderable even though the v1 entity gained `emoji` here. *Verdict: refuted; agreed and deliberately deferred.* Grouped notifications are 5.1b scope, and `emoji_reaction` is intentionally **not** in `DEFAULT_GROUPABLE_TYPES` (Mastodon groups only favourite/reblog/follow by default), so a reaction only reaches the v2 grouped path when a client explicitly opts in with `grouped_types[]`. The v1 `emoji` field landed here only because this PR is where the notification is first created and `internalTypeToMastodon` would otherwise have rendered it as a `mention`.

lib/database/sql/status.ts
```diff
@@ -2169,6 +2169,15 @@ export const StatusSQLDatabaseMixin = (
       statusIdsToDelete
     )
     await deleteRowsByColumnChunks(trx, 'likes', 'statusId', statusIdsToDelete)
+    // Emoji reactions follow the status, like favourites do. A remote status
+    // that is deleted and later re-fetched under the same id would otherwise
+    // come back wearing its old reaction chips.
+    await deleteRowsByColumnChunks(
+      trx,
+      'status_reactions',
```
llun, line 2177: **`status_reactions` was not wired into any deletion cascade.** `deleteStatus` cleans `likes`, `bookmarks`, `status_pins`, `status_quotes` … but not the new table, and `deleteActorData` mirrored it. Orphan rows accumulate; a remote note deleted and re-fetched under the same id comes back wearing its old chips, and — once 5.1b adds local reactions — a recreated actor URL inherits reactions it never made.

**Verdict: refuted.** The verifier showed `announcement_reactions`, the exact structural analogue this table was copied from, is likewise never purged on actor deletion, along with ~13 other actor-keyed tables; and in this PR `actorId` is always a remote URI, so the actor-reuse scenario is unreachable *today*.

**Fixed anyway** (this hunk + `actor.ts`). The precedent argument is fair for `deleteActorData`, but `deleteStatus` cleans `likes` two lines above — leaving reactions behind is inconsistent with its own contract and leaks storage forever. And "unreachable today" stops being true in 5.1b, which I would rather not have to remember. Regression tests in `status.test.ts` and `actor.test.ts`.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,249 @@
...
+ *   names.
+ * - The image URL is taken from the activity's matching `Emoji` tag only when it
+ *   is http(s), bounded, and served from the sending actor's own host. Anything
+ *   else is stored without a URL and renders as text.
+ */
+const resolveReactionEmoji = (
+  reaction: string,
+  activity: ReactionActivity
+): { name: string; url: string | null } | null => {
+  if (!isCustomEmojiReference(reaction)) return { name: reaction, url: null }
```
llun, line 96: **A colon-less shortcode was stored unqualified and then resolved against our own `customEmojis`.** Flagged independently by three of the four reviewers. `resolveReactionEmoji` only namespaced `:shortcode:`, so a remote sending bare `partyparrot` was stored as `partyparrot`, and `isLocalShortcodeCandidate` (any name without `@` is local) resolved it to the instance's own `:partyparrot:` image — exactly what the comment claimed to prevent.

**Verdict: refuted** on severity — the image surfaced is an admin-uploaded emoji already served publicly at `GET /api/v1/custom_emojis`, so nothing is disclosed.

**Fixed anyway.** The disclosure framing is right but beside the point: via `:partyparrot:` the sender gets `partyparrot@remote.test` rendered as *text*, so the bare form is a strictly **larger** capability — it lets a remote instance choose which local emoji image its chip displays. A shortcode is exactly `[A-Za-z0-9_]+`, which no unicode emoji matches (keycaps like `1️⃣` carry U+FE0F/U+20E3), so `isCustomEmojiReference` now covers both spellings and the DB-layer comment is true again.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (senderHost === getConfig().host) return shortcode
+    return `${shortcode}@${senderHost}`
+  })()
+
+  // The raw reaction is bounded, but the namespaced name is not: a long
+  // shortcode plus a long hostname can overrun the varchar(255) columns it is
+  // stored in (`status_reactions.name`, `notifications.reactionName`, and the
+  // notification `groupKey` built from it). Postgres rejects the write, which
+  // the shared-inbox job would surface as a failed job rather than a 400.
+  if (name.length > MAX_STORED_REACTION_NAME_LENGTH) {
```
llun, line 124: **`MAX_REACTION_NAME_LENGTH` bounded the raw reaction but not the derived `shortcode@senderHost`**, which can reach ~354 chars (100 + 253) and overrun the `varchar(255)` columns it lands in — `status_reactions.name`, `notifications.reactionName`, and the `groupKey` built from it.

**Verdict: refuted** — the per-user inbox wraps its whole handler in `try/catch` and returns 400, never a 500, and the transaction rolls back cleanly.

**Fixed anyway.** The refutation only checked the per-user inbox; the shared-inbox path runs through `emojiReactionJob`, where `createJobHandle` **re-throws** — so this was a failed job and a retry storm, not a clean rejection. New `MAX_STORED_REACTION_NAME_LENGTH = 255` guard drops the reaction with a `logger.warn`, covered by a test. This also bounds the `emoji_reaction:<status>:<name>` groupKey, which a separate finding raised.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,249 @@
...
+  const stored = await database.createStatusReaction({
+    statusId,
+    actorId: activity.actor,
+    name,
+    url
+  })
+  // Nothing changed — an unknown status, a redelivery of a reaction we already
+  // hold, or an actor past the per-status cap. Notifying anyway would let a
+  // sender replay one activity into unbounded notifications.
+  if (!stored) return
```
llun, line 236: **A reaction dropped at the per-actor cap still produced a notification**, so `MAX_REACTIONS_PER_ACTOR` bounded storage but not notification volume.

**Verdict: refuted** — the cap is documented as a storage/conformance limit mirroring glitch-soc's `MAX_REACTIONS`, not a notification-rate control.

**Fixed anyway**, for a case both the finding and the refutation missed: re-delivering the *same* emoji is an idempotent no-op in storage yet created a **new notification every time**, so one replayed activity produced unbounded notifications. `createStatusReaction` and `deleteStatusReaction` now report whether they changed state, and the action only notifies on a real insert. PR 5.1b needs exactly this signal to keep outbound federation idempotent.

lib/database/sql/statusReaction.ts
```diff
@@ -0,0 +1,198 @@
...
+      // query answers both "already reacted with this name?" and "at the cap?".
+      const existing = await trx('status_reactions')
+        .where({ statusId, actorId })
+        .select<{ name: string }[]>('name')
+      // Re-reacting with a name already stored is an idempotent no-op, not an
+      // overflow. A genuine overflow drops the new reaction rather than evicting
+      // an earlier one: eviction would desynchronise us from the sender, which
+      // still believes the evicted reaction stands.
+      if (existing.some((row) => row.name === name)) return false
+      if (existing.length >= MAX_REACTIONS_PER_ACTOR) return false
```
llun, line 62: **The per-actor cap is a read-then-write with no locking**, so concurrent transactions on PostgreSQL READ COMMITTED all observe the pre-insert count and the cap is exceeded.

**Verdict: refuted — and I agree; no change.** The cap is advisory (glitch-soc `MAX_REACTIONS` parity) and does not bound reactions per *status* at all: a hostile instance can register N actors and place 8 each entirely within it. Winning the race yields strictly less than that, so locking would buy nothing while serialising every reaction write on a popular status on the hot inbound path.

The restructured code does now read the actor's existing names in **one** query instead of two, and `onConflict().ignore()` reporting zero affected rows means a lost race still correctly reports "nothing changed" to the caller.

lib/database/sql/statusReaction.ts
```diff
@@ -0,0 +1,198 @@
...
+            (row) => !row.reactionUrl && isLocalShortcodeCandidate(row.name)
+          )
+          .map((row) => row.name)
+      )
+    ]
+    // A disabled custom emoji stops rendering as an image and falls back to its
+    // shortcode text, matching how the picker and emoji list treat it.
+    const localEmojiChunks = localShortcodes.length
+      ? await Promise.all(
+          chunkArray(localShortcodes, getWhereInBatchSize(database, 1)).map(
```
llun, line 152: **The `customEmojis` chunk reserved 0 bindings** although the query also binds `disabled = false`, so a full SQLite chunk emits 1000 bindings against the module's own 999 ceiling.

**Verdict: refuted** — the verifier probed the installed driver: better-sqlite3 ships SQLite 3.53.2 with `SQLITE_LIMIT_VARIABLE_NUMBER = 32766`; the pre-3.32 999-variable build the failure depends on does not exist here, so `SQLITE_MAX_BINDINGS` is a conservative floor.

**Changed anyway** as a consistency tidy, not a bug fix: the rollup query above already passes `getWhereInBatchSize(database, 1)`, and a sibling call differing invites the same question from the next reader.

lib/services/mastodon/getMastodonStatus.test.ts
```diff
@@ -580,6 +580,163 @@ describe('getMastodonStatus', () => {
...
+      expect(getStatusReactionRollups).toHaveBeenCalledTimes(1)
+      expect(getStatusReactionRollups).toHaveBeenCalledWith({
+        statusIds: [firstStatus.id, secondStatus.id],
+        currentActorId: ACTOR2_ID
+      })
+      expect(mastodonStatuses[0].reactions).toHaveLength(2)
+      expect(mastodonStatuses[1].reactions).toEqual([])
+    })
+
+    it('surfaces a boosted status reactions on the reblog, not the wrapper', async () => {
```
llun, line 710: **This test never checked the reblog**, despite its name.

**Verdict: refuted** — the verifier ran a mutation test (deleting the `reactions: []` / `pleroma` lines from `baseData`) and the assertions *did* fail, because both fields are `.optional()` in `BaseStatus`, so `undefined !== []`. Not a dead test. (That mutation was left behind in my working tree; reverted.)

**Strengthened anyway**: the boosted status (`post-3`) now actually carries a reaction and the test asserts it appears on `reblog.reactions` while the wrapper stays empty — which is what the name promises.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * Groups an actor's reaction notifications per (status, emoji). Both parts are
+ * remote-controlled and unbounded-ish — the status id is a URL and the name can
+ * be a full `shortcode@domain` — so the readable form is used only while it fits
+ * the varchar(255) column, with a digest of the same inputs as the overflow
+ * fallback. The digest is deterministic, so a group stays stable across
+ * deliveries either way.
+ */
+export const getReactionGroupKey = (statusId: string, name: string) => {
```
llun, line 205: **CONFIRMED (×3 reviewers) — `notifications.groupKey` overflows on PostgreSQL.** Round 1 bounded the derived reaction *name* at 255 and my commit message claimed that covered the groupKey. It did not: the key is a **concatenation**, `emoji_reaction:` + a status-id URL + `:` + the name, so ~89 chars of overhead for a typical local status id. A 40-char shortcode from a 139-char host yields a 180-char name (passes every existing bound) and a 269-char key.

A verifier reproduced it on a real PostgreSQL 17: `ERROR: value too long for type character varying(255)` (22001). The failure mode is nasty — `createStatusReaction` commits in its own transaction *first*, then the notification insert throws, so the reaction row persists while the recipient never gets notified; and on the sender's retry `createStatusReaction` now returns `false`, so the early-return means the notification is **permanently** lost. No SQLite test can catch it: SQLite ignores `varchar` lengths entirely.

**Fixed** in 50e0bf2c. `getReactionGroupKey` keeps the readable form only while it fits and otherwise falls back to a deterministic digest of the same `(statusId, name)` inputs, so grouping stays stable either way. Tests assert the length bound directly rather than relying on the DB to reject it.

lib/services/statuses/reactionLimits.ts
```diff
@@ -0,0 +1,28 @@
...
+// derived `shortcode@domain` and so can be much longer than the raw reaction.
+// Matches the `varchar(255)` columns it lands in: `status_reactions.name` and
+// `notifications.reactionName`.
+export const MAX_STORED_REACTION_NAME_LENGTH = 255
+
+// `notifications.groupKey` is also varchar(255), but it is a *concatenation* of
+// a status id (a URL) and the reaction name, so bounding the name alone does not
+// bound it. Postgres rejects the oversized write; SQLite ignores varchar lengths
+// entirely, so no SQLite-only test can catch a regression here.
+export const MAX_NOTIFICATION_GROUP_KEY_LENGTH = 255
```
llun, line 28: **CONFIRMED — the round-1 comment here was factually wrong.** It claimed `MAX_STORED_REACTION_NAME_LENGTH = 255` "matches … the `groupKey` derived from it". As the verifier pointed out, that cannot be true by arithmetic: even a *zero-length* status id overflows at `15 + 1 + 255 = 271`.

**Fixed** in 50e0bf2c: the comment now says what the constant actually covers (`status_reactions.name` and `notifications.reactionName`), and a separate `MAX_NOTIFICATION_GROUP_KEY_LENGTH` documents that the groupKey is a concatenation and that SQLite cannot catch a regression in it.

lib/database/sql/statusReaction.ts
```diff
@@ -0,0 +1,202 @@
...
+      .where({ statusId, actorId, name })
+      .delete()
+    return deleted > 0
+  },
+
+  async getStatusReactionRollups({
+    statusIds,
+    currentActorId
+  }: GetStatusReactionRollupsParams) {
+    const uniqueStatusIds = [...new Set(statusIds)]
```
llun, line 102: **CONFIRMED — no bound on the number of *distinct* reaction names per status.** The per-actor cap of 8 does not bound the per-status total (a peer with 125 actors reaches 1000 distinct names), `getStatusReactionRollups` applies no `LIMIT`, and the serializer emits the full list **twice** per status (`reactions` and `pleroma.emoji_reactions`), so a hot status can inflate every timeline page that carries it.

**Won't fix in this PR — documented instead** (5133c618). Reasoning: the epic plan explicitly scopes this out for v1 ("no per-status global cap v1"), and the obvious mitigations each have a real cost I don't want to pay silently. Truncating the rollups in the serializer would drop a viewer's own reaction out of the list when it falls beyond the cap, so `me` goes false and the UI would invite them to re-add a reaction they already have. A per-group SQL `LIMIT` needs window functions, which the repo's MySQL-compatibility guidance discourages. A write-time per-status cap would contradict the plan's stated design.

Storage is unaffected; the cost is response size, and it now appears as a **Known limitation** in `docs/mastodon-api-compatibility.md` so it is an explicit trade-off rather than an oversight. The natural place to revisit is 5.1c, which owns the chips-row truncation contract.

lib/database/sql/actor.ts
```diff
@@ -1675,6 +1681,10 @@ export const ActorSQLDatabaseMixin = (database: Knex): SQLActorDatabase => ({
       // Delete likes made by this actor
       await trx('likes').where('actorId', actorId).delete()
 
+      // Delete emoji reactions made by this actor, so a recreated actor URL
+      // cannot resurface them as the new owner's own reactions.
+      await trx('status_reactions').where('actorId', actorId).delete()
```
llun, line 1686: **Round 2 flagged the per-status half of this cleanup as untested** — "deleting the hunk leaves the entire suite green".

**Verdict: refuted** by the verifier's own mutation test, which found existing tests *do* fail.

**Covered anyway** in 50e0bf2c, and the exercise paid for itself: while writing the test I found a round-2 agent had left the per-status hunk deleted in my working tree from *its* mutation test, and the new test failed. Restoring the hunk turned it green — so the test is demonstrably load-bearing, which is exactly what the finding asked for. It reacts as a *remote* actor on a deleted actor's status, so only the `statusId` sweep (not the `actorId` one) can clean it up.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,265 @@
...
+ */
+export const emojiReactionRequest = async ({
+  activity,
+  database
+}: EmojiReactionRequestParams) => {
+  const reaction = getReactionContent(activity)
+  if (!reaction) return
+
+  const resolved = resolveReactionEmoji(reaction, activity)
+  if (!resolved) return
```
llun, line 223: **Round 2 flagged `.onConflict().ignore()` as always-truthy**, so the documented concurrent-writer guard never fired and `createStatusReaction` reported `true` for a write it did not make.

**Verdict: refuted as a stale snapshot** — and correctly so: I had found and fixed the same thing independently in a6dd5c7d before this reviewer's read landed. With no `returning`, knex hands back the raw driver response, which on PostgreSQL is a `Result` object that is truthy regardless of rows inserted, so the check was dialect-dependent and misleading. The code now relies on the in-transaction read (which definitively establishes the row did not exist) and documents the residual race as one duplicate notification at worst — the same exposure `createLike`'s identical read-then-insert has always had.

llun (commented):
Automated sub-agent review **round 2** — three reviewers over the round-1 fix commit and the whole final diff, each finding handed to an independent skeptic.

**7 findings raised, 4 confirmed.** Unlike round 1, these survived refutation: one verifier spun up a PostgreSQL 17 cluster and reproduced the `groupKey` overflow with `ERROR: value too long for type character varying(255)`.

Three of the four confirmed findings are the same root cause (the `groupKey` bound), fixed in 50e0bf2c. The fourth is documented in 5133c618. Dispositions inline.

Also worth recording: two round-2 agents mutated the working tree to run mutation tests (deleting the `baseData` reaction lines, and deleting the `deleteActorData` per-status hunk) and did not restore them. Both reverted; the accidental second one usefully confirmed the new `deleteActorData` test is load-bearing rather than vacuous.

lib/actions/emojiReaction.ts
```diff
@@ -0,0 +1,276 @@
...
+  }
+}
+
+// Whether the sender is entitled to supply the image for this reaction name.
+// Only the instance that owns the namespace is: otherwise a hostile peer could
+// relay `:blobcat@good.test:` with an image on its own host, pass the
+// same-host URL check, and — because the rollup unwraps one url per
+// (statusId, name) group — have its image displace the real one that
+// good.test's own reactors stored.
+const ownsReactionNamespace = (name: string, senderHost: string) =>
```
llun, line 146: **CONFIRMED — a peer could bind its own image to another instance's emoji namespace.** The round-1 fix validates the `Emoji` tag's icon URL against the *sending* actor's host, which is right for a sender using its own shortcode. But a reaction can arrive already namespaced (`:blobcat@good.test:`), and that path accepted the sender's namespace claim without checking the sender owns it. So `evil.test` relays `:blobcat@good.test:` with an image on `evil.test` — the same-host check passes — and because `getStatusReactionRollups` unwraps a single `MAX(url)` per `(statusId, name)` group, the hostile image can **displace** the legitimate one that good.test's own reactors stored for that same chip.

**Fixed**: `ownsReactionNamespace` gates the image on the stored name ending in `@<senderHost>`. A third party's reaction still counts toward the chip — it just renders as text instead of carrying a picture, which is the same conservative fallback used when the URL fails the scheme/host/length checks.

lib/types/activitypub/activities.ts
```diff
@@ -83,19 +83,44 @@ export const Reject = z.object({
...
   id: z.string(),
   actor: z.string(),
-  object: z.union([z.string(), Note])
+  object: z.union([z.string(), Note]),
+  // Liberal per the ActivityPub & JSON-LD rules: a sender whose document
+  // carries a default language can have compaction render `content` as a
+  // language-tagged object rather than a scalar. Degrade that to `undefined`
+  // instead of failing the whole activity — a `Like` we cannot read a reaction
+  // from is simply an ordinary favourite, which is the pre-reaction behaviour.
+  content: z.string().optional().catch(undefined),
```
llun, line 111: **CONFIRMED — typing `content` as a bare `z.string()` could break plain favourites that worked before this epic.** Adding `content`/`_misskey_reaction`/`tag` to `Like` made the whole activity fail validation whenever compaction renders those values as something other than a scalar — a language-tagged object, for example, which is exactly what happens for a sender whose document carries a default language (the reason `normalizeInputContext` already strips an *undetermined* one). A failed `Like` falls through the inbox `Activity` union and becomes a **400**, so an ordinary favourite that federated fine yesterday would start bouncing.

**Fixed**: all three fields now `.catch(undefined)`, so an unreadable value degrades to absent and the activity parses as the plain favourite it is. This is what AGENTS.md's ActivityPub & JSON-LD section asks for — liberal schemas, narrowed at the consumption boundary, which is `getReactionContent`'s job. Verified the zod behaviour directly (string → kept; language-tagged object / array / number → `undefined`, `success: true`), and added a regression table for the three shapes.

llun (commented):
Automated sub-agent review **round 3** — two reviewers over the round-2 fixes and the whole final diff.

**2 findings, both real and both fixed** in b7281b18. Neither had been caught by rounds 1 or 2, and both are inbound-federation trust issues rather than style. Dispositions inline.

I also self-checked the round-2 fix while round 3 ran: the digest fallback key is a fixed 79 chars (`emoji_reaction:` + sha256 hex), `status_reactions.url` is `text` in both dumps so it needs no length bound, and `status_reactions.actorId` is `varchar(255)` — the same exposure the existing `likes` table has, so not something this PR introduces.

Gate after each round: prettier → eslint (0 errors) → next build → vitest. **716 files / 6984 tests green**, plus a cross-backend run of the reaction data layer against a real PostgreSQL 17.

llun:
Automated sub-agent review **round 4 — clean**. Two reviewers (round-3 fix verification + a from-scratch end-to-end sweep) returned **zero findings**, which closes the review loop per the AGENTS.md exit condition.

**Loop summary across 4 rounds: 20 findings raised, 7 real defects fixed.**

| Round | Raised | Confirmed by the skeptic | Fixed |
|---|---|---|---|
| 1 | 11 | 0 | 4 (re-judged; refutations were right on facts, wrong on consequence) |
| 2 | 7 | 4 | 2 (+2 dedupes) |
| 3 | 2 | 0\* | 2 |
| 4 | 0 | — | — |

\* Round 3's verifiers refuted both only because I had already applied the fixes while verification was running — both explicitly confirmed the defects were real at the commit they reviewed.

One round-3 verifier also left a genuine follow-up: the `MAX(url)` comment in the rollup had become an approximation once namespace ownership was enforced. Corrected in a313958d.

Final state on `a313958d`: prettier → eslint (0 errors) → `next build` → `vitest run` all green locally (**716 files / 6984 tests**), CI green on the head commit (build, lint+prettier, schema-dump sync, CodeQL, all 4 test shards), and the reaction data layer separately exercised against a real PostgreSQL 17.

Merging.

PR #1321, opened 2026-07-25
minor: reply by email for mentions and direct messages

llun:
## Summary

Makes mention and reply notification emails **repliable**. Hitting Reply in any mail client and sending posts a real fediverse reply, in the right thread and at the same visibility as the post being answered — a reply to a DM stays `direct`, without the sender ever typing an `@mention`.

The outbound half already existed and is reused as-is (`sendMail` across all four providers already supported `replyTo`; `Message.replyTo` was defined and unused). Everything new is on the inbound side: an address to reply *to*, a way to receive the mail, and a safe path from a received message to `createNoteFromUserInput`.

Off by default at every level — this only does anything once an operator configures inbound email **and** an account opts in.

## How it works

**Outbound.** When a mention/reply notification email is built, the producer mints a 128-bit capability token bound to exactly one `(account, post)` pair and passes `reply+<token>@<inbound domain>` as `Reply-To`. The body then leads with a sentinel line (`--- Reply above this line to respond ---`) — mail clients quote the original *below* the new text, so that line is a precise cut point on the way back in. Minting is a database write on a path that was previously pure reads, so a failure degrades to a normal non-repliable email rather than losing the notification.

**Inbound.** `POST /api/v1/webhooks/email` authenticates the provider, answers `202` as soon as the request checks out, and enqueues `ReplyByEmailJob`. The job re-resolves the token, re-checks the account and both switches *at post time*, strips the quoted history, and posts through `createNoteFromUserInput` with **no `visibility`** — which is what makes visibility inheritance work: `getVisibilityFromReplyStatus` supplies it, and `getMentions` already appends the parent author.

## Gating — three switches, all required

| Level | Where | Default |
| --- | --- | --- |
| Environment | `ACTIVITIES_EMAIL_INBOUND_{SECRET,DOMAIN}` (+ outbound email) | unset |
| Instance | **Admin → Posts & media → Reply by email** (`replyByEmail.enabled`) | on |
| Account | **Settings → Notifications → Reply by email** | **off** |

The account toggle defaults off deliberately: it turns an email address into a posting capability. When the instance cannot receive replies at all, the account toggle is **hidden entirely** — it is per-account state with nothing to opt in to.

The admin switch behaves differently on purpose, because it is a database row *every* instance honours and it defaults to on. It stays visible and always shows its **stored** value, and it locks in one direction only: it cannot be switched on where the feature would be inert, but a stored *on* stays readable and writable. Rendering a stored `true` as `false` — or locking it — would put the cluster-wide kill switch out of reach of whichever pod the admin happened to be talking to, while a pod that *does* have the variables kept minting. Its help line names whichever half of the email config is missing, inbound or outbound, rather than blaming both.

**Admin → Posts & media → Configure environment** also gained a **Reply by email** area that assembles the `ACTIVITIES_EMAIL_INBOUND_*` block. This turned `EnvTemplateArea` into a discriminated union: the existing areas pick a backend and their fields follow from it, while this one has a single fixed set of variables and so renders no selector.

## Security

The reply address travels in plaintext email and lands in mail logs, so it is treated as a bearer capability that will leak eventually.

| Risk | Defense |
| --- | --- |
| Forged webhook request | HMAC-SHA256 over `"<timestamp>.<rawBody>"`, 5-minute window, constant-time compare, `404` when unconfigured |
| Replay of a captured request | The timestamp is signed *with* the body — a body-only signature stays valid forever — plus per-`Message-Id` idempotency |
| Duplicate provider delivery | Same idempotency key (`email-reply:<sha256(messageId)>`, scoped to the actor) |
| Leaked database → forged addresses | Only the keyed HMAC is stored; the raw token exists solely in the sent email |
| Token leaked via a forwarded thread | 30-day expiry, 20-use ceiling, scoped to one `(actor, status)` pair — it cannot post anywhere else |
| Account suspended/opted out after minting | Actor, moderation state and both switches are re-checked in the job, not at mint time |
| Blocked/muted conversation | `createNoteFromUserInput` already runs block and mute filtering; nothing bypasses it |
| Enormous body / zip bomb | 10 MB body cap on the route, `validateStatusContentLimits` on the text, `exceedsMaxMediaUploadSize` per file |

**On `From`:** checked *softly* on purpose. Without the DKIM/SPF result — which a normalized JSON webhook has already discarded — it is trivially forgeable, while requiring it would break replying from an alias, a plus-address, or a work account. The token is the real authorization; a mismatch is logged and accepted.

**Silent data loss:** a reply that cannot be posted (empty body, over the limit, expired token, opted out, thread gone) gets a short notice back explaining why. That notice deliberately carries no `Reply-To` of its own, so it cannot loop.

## Operator setup

The inbound domain needs an **MX record pointing at whichever provider POSTs the webhook** (Mailgun Routes, SendGrid Inbound Parse, Postmark, an SES receipt rule + Lambda, …). Full contract, JSON body shape, and signature recipe are in [`docs/features.md#reply-by-email`](docs/features.md).

## Notes for review

- **Migration:** adds `email_reply_tokens`. Both reference schema dumps were regenerated against fresh local databases (PostgreSQL 17 in a throwaway container, SQLite via the migration chain); each diff is purely additive and contains only the new table, its two constraints and its two indexes. Docker is unavailable on this machine now, so the PG dump could not be re-run for a final check — it was instead verified against the migration column-for-column, including knex's lowercased constraint names and dump-order placement. CI's Schema Dump Sync gate covers the SQLite side only.
- **`NoQueue`:** `publish` runs the handler inline, so a no-queue deployment does the posting work inside the webhook request. Acceptable, and the job deliberately uses no `delaySeconds` (NoQueue drops delayed messages).
- **`.env.example` is not updated** — a local permission rule blocked reading it. `docs/environment-variables.md` documents all three variables and `lib/config/envDocumentation.test.ts` only checks that file, so CI is unaffected. The block to add is in the first review comment.

## Verification

`yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test` all green (**748 test files, 7771 tests**).

New unit coverage: `extractReplyText` (sentinel, wrapped attribution, `>`-quoting, `-- ` signature, HTML-only, empty), `replyToken` (mint/resolve round trip, expiry, use ceiling, hash-only storage, all three gates), the webhook route (good/bad/stale/tampered signature, unconfigured → 404, malformed → 400/422, oversize → 413), the job (DM stays direct, public inherits, duplicate posts once, suspended refused, disallowed attachment dropped without failing the reply), and the database mixin on both backends.

**Browser-verified against a local SQLite instance** (`yarn dev`, mock user):

1. Account toggle renders off by default with the capability warning; switching it on persists `settings.replyByEmail = true`.
2. Admin switch renders in its own section, saves `replyByEmail.enabled = false`, and with it off the account toggle disappears entirely — even though the account's own setting is still `true`.
2b. On an instance with **no** inbound email configured and the setting at its stored default of `true`, the switch reads *"Reply by email is allowed, but unavailable here"*, stays writable, and turning it off persisted `replyByEmail.enabled = false` — the case that is unreachable if the control is simply disabled. It then re-rendered disabled, since there is nothing left to do. Selecting the builder's new area produced `ACTIVITIES_EMAIL_INBOUND_DOMAIN=…` plus a masked `ACTIVITIES_EMAIL_INBOUND_SECRET`, with the optional prefix omitted until given a value.
3. A correctly signed `curl` to `/api/v1/webhooks/email` returned `202`, the reply appeared in the thread with the quoted history stripped, `useCount` went to 1, and the idempotency key was recorded.
4. A tampered signature returned `401`. Re-POSTing the same `Message-Id` returned `202` and created **nothing** new (still 1 reply, `useCount` still 1).

Screenshots of (1), (2) and (3) are attached in a follow-up comment.


llun:
### `.env.example` block to add

A local permission rule (`Read(./.env.*)`) blocked me from reading `.env.example`, so it is the one file this PR does not touch. Everything else is documented — `docs/environment-variables.md` has all three variables, and `lib/config/envDocumentation.test.ts` only checks that file, so CI is green either way.

This belongs under the existing `# Email` section:

```bash
# ---------------------------------------------------------------------------
# Inbound email (reply by email) — optional
# ---------------------------------------------------------------------------
# Makes mention/reply notification emails repliable. Needs the outbound
# ACTIVITIES_EMAIL_* settings above, an MX record for the domain below pointing
# at whichever provider POSTs the webhook, the instance-wide switch on
# Admin -> Posts & media, and a per-account opt-in in Settings -> Notifications.
# Both SECRET and DOMAIN are required; setting only one disables the feature.

# Shared secret the provider signs inbound webhook payloads with, as
# `x-activities-signature: t=<unix-seconds>,v1=<hex>` where
# v1 = HMAC-SHA256(secret, "<t>.<raw request body>").
# ACTIVITIES_EMAIL_INBOUND_SECRET=

# Bare domain that receives replies, e.g. reply.example.tld. A value with a
# scheme, a user part, or no dot disables the feature.
# ACTIVITIES_EMAIL_INBOUND_DOMAIN=

# Local-part prefix of the reply address. Optional, defaults to `reply`
# (so replies arrive at reply+<token>@<domain>). Restricted to A-Z a-z 0-9 . _ -
# ACTIVITIES_EMAIL_INBOUND_LOCAL_PART_PREFIX=reply
```

llun:
## Review round 1 — 5 reviewers, 22 raw findings, 8 survived adversarial verification, all addressed

Each finding was checked by two independent refuters instructed to *kill* it, defaulting to refuted when uncertain; 14 died there. I reproduced each survivor myself before changing anything. Four of the eight were the same NUL-byte issue reported by different dimensions, so five distinct defects:

### 1. `resend.ts` dropped the Reply-To entirely — HIGH · fixed in cd0ce53

This PR is the first caller ever to set `Message.replyTo`, which made four previously-dead transport branches load-bearing. SMTP and SES were right; Resend was not. `resend-node` v4 renamed the option to `replyTo` and its own serializer emits the API's `reply_to` — so passing the wire name meant the SDK never saw it. Verified in `node_modules/resend/dist/index.mjs:195` (`reply_to: email.replyTo`).

The whole feature was silently inert on any Resend instance: the email arrived with the "Reply above this line" sentinel and no Reply-To, so replies went to the noreply sender and vanished. TypeScript couldn't catch it because the key arrives via a spread, so no excess-property check applies. `lib/services/email/resend.test.ts` is new and fails against the old key.

### 2. Every Gmail HTML-only reply was thrown away — HIGH · fixed in cd0ce53

`dropTrailingAttribution` deleted the last line whenever it ended in `wrote:`, even when the walk-back found no `On …` opener. `htmlToPlainText` collapses an HTML reply onto **one** line, so that line is `"<user text> On <date> … wrote:"` — and deleting it returned `''`, which the job reads as `empty` and answers with a "we could not post your reply" notice.

Reproduced before fixing: a realistic Gmail `gmail_quote` body returned `""`. So did the plain-text sentence `"Here is what she wrote:"`. Now the attribution-shaped **suffix** is trimmed and the sender's text in front of it survives; a sentence that merely ends in `wrote:` is left alone. Both cases are regression-tested.

### 3. Expired and spent tokens vanished silently — MEDIUM · fixed in cd0ce53

`resolveReplyToken` collapsed *unknown hash*, *expired*, and *use-exhausted* into `null`, and the job logged all three and stopped. That reasoning only holds for an unknown hash — an expired row still carries its `actorId`, so the owner was perfectly reachable and just never told. The tell was already in the codebase: the `'expired'` failure reason existed with user-facing copy and **nothing could ever produce it**.

`resolveReplyToken` now returns a discriminated `{ status: 'ok' | 'expired' | 'exhausted' | 'unknown' }`. Unknown still logs only; the other two send the notice. `'exhausted'` was added as its own reason rather than reusing the expiry wording.

### 4. Two raw NUL bytes made the job file binary to git — MEDIUM · fixed in d03ff06

`git diff main...HEAD --stat` reported `lib/jobs/replyByEmailJob.ts | Bin 0 -> 9335 bytes`. The single most security-relevant file in the change — token resolution, the authorization re-checks, the attacker-supplied base64 decode loop, the outbound mail path — rendered as "Binary file not shown" on GitHub, so the review this PR is meant to get could not see it. `grep` skipped it too, and prettier/eslint both passed clean, so nothing in the gate caught it.

The separator is now written as a `` escape in an array join instead of raw bytes. Identical hash input; that file diffs as **287 readable lines**.

### 5. The attachment test asserted on the mock, not the post — MEDIUM · fixed in cd0ce53

`stores a real image attachment on the reply` only checked that `saveMedia` was called. It couldn't check more: the fake media id resolved to nothing through the live `getAttachmentsFromMediaIds`, so `storeAttachments` returned `[]` and the reply was posted with **no** attachments — the test name was false. Deleting `attachments` from the `createNoteFromUserInput` call would not have failed a single test. Now stubbed at the `mediaIds` boundary and asserted on the posted status.

### Also: coverage for the `createNote.ts` minting sites

Flagged as untested, and it matters more than a coverage gap: the `actorId` passed there decides *whose account* a leaked reply address can post as. The two call sites use different expressions (`replyStatus.actorId`, `mentionedActorId`) from the remote producer's `currentActor.id`, which is exactly the shape of thing a future edit copies wrong.

`lib/actions/createNote.replyByEmail.test.ts` asserts the minted token resolves to the **recipient**. Verified by mutation: swapping in `currentActor.id` fails the test, restoring passes it.

### One I found before the reviewers

Probing my own parser on real mail shapes: Outlook puts an *unquoted* divider + `From:/Sent:/Subject:` block between the reply and the sentinel, so neither the sentinel cut nor the quoted-tail cleanup removed it and those headers were posted as part of the status. Fixed in b1398d9.

---

Gate re-run green after all of it: prettier → lint (0 errors) → build → **722 test files, 7036 tests**. Starting the next review round.

llun:
## Review round 2 — 10 raw findings, 4 survived, all addressed

Round 2 was aimed squarely at the round-1 fixes, on the principle that fixes introduce bugs. It was right to be: **the worst finding of the whole review is a regression I introduced in round 1.**

### 1. The round-1 fix silently truncated published posts — fixed in 80e0188

Round 1 replaced "delete the last line if it ends in `wrote:`" with "trim the attribution-shaped suffix", using one regex anchored only at its end:

```js
/(?:^|\s)(?:On|Le|Am|El)\s.*\bwrote:\s*$/
```

JavaScript matches **leftmost**. On the collapsed-HTML path — where `htmlToPlainText` puts the sender's every sentence on the same line as the attribution — that starts at the *first* ` On `/` Am `/` El `/` Le ` in the line and deletes everything after it. Reproduced through the real function before changing anything:

| Reply body (HTML) | Posted |
| --- | --- |
| `Yes, that works.` / `On Tuesday I am offline.` | `Yes, that works.` |
| `Thanks! Am I right that this ships today?` | `Thanks!` |

Worse than the bug it replaced. Round 1's failure was loud — the reply came back empty and the sender got a "we could not post" notice. This one is silent: something survives, so the `empty` guard never fires, the length check passes on the *already truncated* string, and a mangled status is published irreversibly with the sender believing they sent the whole thing.

Now cuts at the **last** opener. The trade-off is deliberate and asymmetric: cutting late can leave a stray date fragment in a post, cutting early deletes words someone actually wrote. Noise beats data loss. (I also avoided the tempered-wildcard form — `(?:(?!\s(?:On|Le|Am|El)\s).)*` — because it backtracks per candidate start, which is O(n²) on a line that can be megabytes after the collapse.)

### 2. The Outlook divider cut never worked on the HTML shape — fixed in 80e0188

My own earlier fix (b1398d9) closed this for `text/plain`, where Outlook emits a row of underscores. Its **HTML** reply uses `<hr>` instead — which `htmlToPlainText` drops, since it isn't in `ALLOWED_STRUCTURE_TAGS` — so neither divider pattern could match and the `From:/Sent:/Subject:` block was posted as part of the status. Same leak, other body part. Now cut inline, requiring the full `From: … Sent:|Subject:` sequence so an ordinary sentence mentioning "From:" isn't mistaken for it.

### 3. The instance kill switch was never tested at redemption — fixed in 80e0188

The job gates on `!serverSettings.replyByEmail.enabled || actorSettings?.replyByEmail !== true`, and only the second half was covered. The reviewer proved it by deleting the first half — 62 tests still passed.

That is not a cosmetic gap. Tokens live 30 days, so the switch **has** to be re-checked at redemption; minting-side coverage is no substitute. An admin turning the feature off after abuse would otherwise find every address issued in the previous month still posting. Mutation-verified: deleting the check now fails the new test.

### 4. A test that proved nothing about the guard it named — fixed in 80e0188

`inbound email is not configured` passed `token: 'irrelevant'`, which hashes to no row — so the job returned early at the token lookup whether or not the config guard existed. Deleting `if (!email || !emailInbound) return` left every test green. It now uses a real minted token, so the guard is the only thing stopping the post. Mutation-verified.

---

Six further findings were refuted by the verification pass and are not changes.

Gate green: prettier → lint (0 errors) → build → **723 test files, 7041 tests**. Round 3 running.

llun:
## Review round 3 — 14 raw findings, 8 survived, all addressed

Seven of the eight were the same defect wearing different clothes, so this round stopped patching regexes and fixed the root cause.

### The pattern

Every cut heuristic in `extractReplyText` keyed on **shape alone** — a line starting with `On`/`Le`/`Am`/`El`, or containing `From:`. Those are ordinary English. Three rounds running, each fix removed one false positive and left the identical flaw in the regex two lines below it:

| Round | Cut at | Result |
| --- | --- | --- |
| 1 | last line ending in `wrote:` | every Gmail HTML-only reply destroyed |
| 2 | first `On` in the line | `"Yes, that works. On Tuesday I am offline."` → `"Yes, that works."` |
| 3 | first `From:` in the line | `"The bounce says From: postmaster@… — do you know why?"` → `"The bounce says"` |

Round 3 also found the *older* branches had it too: `cutAtAttribution` joined its 3-line window **across blank lines**, fusing a sender's `On Friday I will be away.` with the client's attribution two lines below and deleting the paragraph between; and the `dropTrailingAttribution` walk-back did the same, reducing `Am I late? / Read what she wrote:` to nothing.

### The fix — evidence, not shape

A machine-generated header always carries a **date, a time or an address**. A sentence does not. `ATTRIBUTION_EVIDENCE = /[0-9]|@/` is now required before any attribution cut, windows join only across **contiguous** lines (a header never spans a blank one), and the Outlook run must be all three of `From:` → `Sent:`/`Date:` → `Subject:` in order within bounded gaps.

The governing rule is now written down in the module: *leaving a stray fragment of a quoted header in a post is cosmetic; deleting a sentence the sender wrote is silent, published and irreversible.*

### A denial of service in the same regexes — HIGH

The unbounded lazy spans were quadratic: for each `From:` on a line the engine rescanned the remainder looking for `Sent:`/`Subject:`. An HTML body collapses to **one** line of up to megabytes, and `validateStatusContentLimits` runs on the *extracted* text — i.e. after the parser — so the size limit was no protection. Measured: 4× cost per doubling, ~5.5s at 400 KB, extrapolating to minutes of blocked event loop for a body well under the 10 MB webhook cap. Any account with a live reply address could send it repeatedly.

Bounded gaps make each scan constant, and input is now capped at 256 KB before parsing — safe because the sender's text is always at the **top** of a reply, so truncating the tail cannot lose a word of it. There is a timing test.

### The use ceiling was advisory — MEDIUM

`resolveReplyToken` read `useCount` and compared it to the ceiling; `recordEmailReplyTokenUse` ran at the very *end* of the job, after media storage and federation. Classic check-then-act with a window of hundreds of milliseconds. QStash delivers concurrently and its dedup only collapses retries of the same `Message-Id`, so 100 distinct emails bearing one token would all read the same stale count and all post — past a ceiling `docs/features.md` advertises as a bound.

Replaced with a conditional `UPDATE … WHERE useCount < ? AND expiresAt > ?` that re-tests both inside the statement, taken **before** any expensive work so an abusive burst cannot even do the media work. Mutation-verified: removing the WHERE clauses fails three tests, including one asserting exactly `maxUses` of 10 concurrent claims succeed.

The trade-off is deliberate: a post that then fails still costs a use. That is the right way round for a control whose purpose is bounding abuse.

---

Six findings were refuted and are not changes.

Gate green: prettier → lint (0 errors) → build → **723 test files, 7051 tests**. Round 4 running — every round so far has found a defect in the previous round's fix, so the loop continues until one comes back clean.

llun:
## Review round 4 — 7 raw findings, 6 survived, all addressed

### 1. The round-3 bound leaked the replier's own email address — HIGH · fixed in 29f5247

Round 3 bounded both gaps in the Outlook header run at 80 characters. Real Outlook puts `To:` — and often `Cc:` or `Importance:` — between `Sent:` and `Subject:`, so the second gap routinely exceeds it and the run stopped matching. Because Outlook's HTML reply emits `<hr>` (which `htmlToPlainText` drops) rather than the underscore rule, that regex is the **only** marker on the path. Reproduced:

> `My actual reply. From: Activities <noreply@example.tld> Sent: Saturday, July 25, 2026 9:00 PM To: Bob Smith <bob.smith@example.tld> Cc: Team <team@example.tld> Subject: someone mentioned you`

— published and federated as the status. The `To:` line is the **replying user's own address**; `Cc:` is everyone else's.

Only the second gap is widened, to 300. The first must stay tight: widening it lets a prose `From:` reach the real block and delete the sender's text instead, which I verified fails the existing test. Three shapes are now regression-tested and assert the output contains no address at all.

### 2. The evidence rule was too weak — MEDIUM · fixed in 29f5247

Round 3's `ATTRIBUTION_EVIDENCE = /[0-9]|@/` accepted any bare digit, so a sentence mentioning any date or number still satisfied it and was still deleted:

> `Some context.` / `On 25 July I asked Alice about the outage and here is what she wrote:` → **`Some context.`**

Now `/@|\d{1,2}:\d{2}/` — an address or a clock time. Every client probed (Gmail desktop and mobile, Apple Mail, Thunderbird, Outlook.com, Yahoo, Proton) emits one; prose does not.

### 3. Reply by email ignored `accounts.disabledAt` — HIGH · fixed in 29f5247

The job checked only `actor.suspendedAt`. The repo's canonical gate is `isActorModerationBlocked` — suspended **or** account disabled — and it guards every other authenticated surface. `disabledAt` is the Mastodon admin "disable" action; it also blocks sign-in.

So an admin could disable an abusive account, and that account could keep posting by replying to any notification email it still held. Now uses `isActorModerationBlocked` at redemption, and at minting so a disabled account is never sent a repliable email. Mutation-verified: reverting to the bare `suspendedAt` check fails the new test.

### 4/5. Failure notices were an unbounded outbound-mail stream — MEDIUM · fixed in 29f5247

Every failure branch — `disabled`, `thread-missing`, `empty`, `too-long`, `expired`, `exhausted` — emailed the account owner *before* any use was claimed, and none consumed one. The use ceiling is documented as the bound on a leaked address, but it bounded only the posts. 10,000 messages to a leaked address meant 10,000 "we could not post your reply" emails into the owner's inbox, from the instance's own sending domain.

The claim now sits above every notice, so the whole surface is bounded by the ceiling — and below the idempotency check, so a provider retry cannot eat the budget. A spent or expired token is dropped in silence.

That last part reverses round 3's fix for those two cases, deliberately: the round-3 concern was silent loss with no feedback anywhere, and the fix went too far by answering *every* message forever. Bounded feedback is the right middle. `docs/features.md` now says so.

### 6. The "disabled" notice blamed the wrong thing — LOW · fixed in 29f5247

One reason string covered both the instance switch and the account toggle, and said "for this account". With the instance switch off, the account toggle is **hidden** from Settings → Notifications — so the notice sent people looking for a control that was not there. Split into `disabled-instance` and `disabled-account`.

### Also: a flaky test of my own

A wall-clock assertion I added in round 3 (`< 1000ms`) failed once under full-suite parallel load and passed 12/12 in isolation. Rather than leave that in, the guarantee worth asserting is the deterministic one — input above `MAX_PARSE_CHARS` is truncated — with the timing kept only as a generous smoke bound. Full suite green twice in a row afterwards.

---

One finding was refuted and is not a change.

Gate green: prettier → lint (0 errors) → build → **723 test files, 7059 tests**. Round 5 running.

llun:
## Review round 5 — 3 raw findings, 2 real, both fixed · plus a merge with main

Round 5's first attempt aborted: both reviewers hit a session limit, so it returned zero findings because zero agents ran. Re-running it once the limit reset is what produced the results below — worth stating, because an aborted round looks exactly like a clean one.

The re-run's aggregation then scored all three findings "refuted", but that rule requires **both** verifiers to agree. Two findings had one verifier that reproduced them by running the code, so I checked those myself rather than let a split verdict dismiss them. Both were real.

### 1. A long instance name re-opened the address leak — fixed in 29f5247

Round 4 fixed the Outlook header block by widening the *second* gap and keeping the first bounded at 80 characters. But the first gap spans the instance's own `From` value — so on an instance whose service name is long, the run stops matching and the whole block, **including the replier's own address**, is published again. Reproduced:

| Instance `From` | Extracted |
| --- | --- |
| `Activities <noreply@example.tld>` | `My actual reply.` |
| `Activities Social Network <noreply@social.example-company.example>` | `My actual reply.` |
| `Activities Social Network for Teams and Communities <…>` | **the entire header block** |

The mistake was bounding by *distance*. What separates a sender's typed `From:` from the real block is **content**: a header value never contains sentence-ending punctuation, and the prose between someone's typed `From:` and the real header almost always does. The gap is now tempered on that instead — still bounded, so still linear (15 ms on the 400 KB adversarial input).

### 2. Photos emailed from an iPhone were silently dropped — fixed in 29f5247

Every part carrying a `Content-ID` was treated as decoration. Apple Mail and iOS Mail mark a photo someone **deliberately attached** as inline too, so those were dropped — the reply posted as text only, with nothing said, while the docs promise attachments ride along. A part is now skipped only when the body actually references it with `cid:`, which is what the signature-logo case was always about.

The third finding — a stale clause in a doc comment — was genuinely cosmetic and both verifiers refuted it.

---

## Merged with main

Main moved underneath this branch: emoji reactions (#1320, with its own migration) and an account-email redesign (#1322).

The redesign rewrote the notification templates onto a design-system layout — `build*Email()` returning a `RenderedEmail` through a shared `renderEmail`. The sentinel could no longer be a string each template prepends, so it is now a **`replyMarker` option on the layout**, emitted as the first line of both media. First is not cosmetic: the inbound parser keeps everything *above* the marker, so anything rendered ahead of it — the preheader, the header wordmark — would be swept into the published reply. There is a test asserting that ordering, and an end-to-end one that quotes the rendered email and recovers only the sender's text.

Taking main's rewritten template test files dropped my sentinel assertions, so that coverage was rebuilt at the layout level where the invariant now lives.

Also in the merge:

- **Both schema dumps regenerated** against the combined 144 migrations, against fresh local databases (throwaway PostgreSQL 17 container + SQLite), never hand-edited.
- **This change's migration renamed** to `20260726010000_add_email_reply_tokens.js`. It had landed on the same timestamp as main's `20260726000000_add_status_reactions.js`; two migrations sharing a timestamp order only by the rest of the filename, which is not a property worth depending on. The rename changes table creation order, so the dumps were regenerated again afterwards.

Gate green on the merged tree: prettier → lint (0 errors) → build → **7411 tests**.

Round 6 will run against the merged result, since the merge brought in code paths none of the earlier rounds saw.

llun:
## Review round 6 — 10 raw findings, 5 confirmed + 3 split, all addressed

This round reviewed the merged tree. I also changed the harness: split verdicts are now reported separately instead of being dropped, because round 5 had two real defects that one verifier reproduced and the other refuted. All three splits here turned out to be real.

### 1. Unsanitised attachment filename reached a path-traversal write — HIGH · fixed in 892112f

An inbound attachment's `filename` went from the wire straight into `new File()`. The object-storage video path then builds `join(tmpdir(), <random> + file.name)` and calls `writeFile` — and `join` resolves `..`:

```
join('/tmp', 'deadbeef' + '../../../../etc/passwd')  ->  /var/folders/wg/etc/passwd
join('/tmp', 'deadbeef' + 'deadbeef../../evil.mp4')  ->  <tmpdir>/evil.mp4
```

Every other upload surface receives a basename from a browser's multipart encoder. Inbound email is the first that takes this string straight off the wire, so it is the first place the weakness is reachable by a remote party. Now reduced to a basename with control characters stripped, with regression tests for a POSIX traversal, a Windows path and a bare `..`.

The underlying weakness is in the shared media layer and is reachable from the non-browser upload APIs too. That is a separate change needing its own review, so it is filed as follow-up work rather than widened into this PR.

### 2. The failure notice was the one email left on the old template shape — MEDIUM ×2 + LOW · fixed in 892112f

The merge brought in main's redesign, where every template builds through `renderEmail`. My `replyByEmailFailure.ts` predates it and still exported the old `getSubject`/`getTextContent`/`getHTMLContent` trio with hand-rolled escaping and raw markup — so it sent a bare `<h3>`/`<p>` fragment: no doctype, no header, no footer, quirks mode, beside eleven designed emails. Verified by rendering it (233 bytes, `doctype:false`) against `buildReplyEmail` (4473 bytes, `doctype:true`).

It also silently escaped the design conformance suite, which enumerates templates through a hand-maintained import list rather than scanning the directory.

Now `buildReplyByEmailFailureEmail(): RenderedEmail`, composed from shared blocks with an `account` footer and no reply marker (a notice about a failed reply must not itself be repliable). Registered in `emailDesign.test.ts` and in `buildPreviews()` — `renderEmailPreviews.ts` now renders 11 previews, so the branch's only new email finally has the visual-verification path the docs promise.

That also makes the merged claims in `AGENTS.md`, `REVIEW.md` and `docs/maintenance.md` true again: `activityImport.ts` really is the one template left. Counts bumped from "ten of eleven" to "eleven of twelve".

### 3. `<title>` could be published from a whole-document quote — MEDIUM (split) · fixed in 892112f

`<title>` carries the subject as real text that sanitize-html keeps, and it sits above the reply marker. A client quotes the rendered body, so this needs an unusual provider — but it costs nothing to close. The `<head>` region is now removed before conversion.

Worth recording how the first attempt at this went wrong: I cut everything before `<body>`, which deletes the sender's own text, since in a quoted reply that text comes *ahead* of the quoted document. My own test caught it. That is the same over-eager-cut mistake as rounds 1–4 — the fix removes only the head region.

### 4. Reply tokens outlived the account — MEDIUM (split) · fixed in 892112f

`deleteActorData` cleaned twenty tables and not `email_reply_tokens`. The rows are inert once the actor is gone (the job re-checks), but they name the actor and the statuses they were minted for, so they should go with the account rather than linger until an operator runs the prune script. Now deleted in the same transaction.

### 5. The docs promised M4A attachments — LOW (split) · fixed in 892112f

`ACCEPTED_FILE_TYPES` includes `audio/mp4`, so a voice memo passes the filter — but both storage backends reject anything that is not `image/*` or `video/*`, so it is dropped a step later. `docs/features.md` said M4A rides along. Corrected rather than "fixed" in code: making audio work is a media-pipeline feature, not something to smuggle into this PR.

---

Two findings were refuted outright and are not changes.

Gate green on the merged tree: prettier → lint (0 errors) → build → **7427 tests**.

llun:
Closing unmerged — too large to land safely in one piece. Worth breaking into smaller PRs if we revisit.

Nothing is lost: the branch `claude/reply-by-email-mentions-dms-694d19` is preserved at `6a0f2b0af`, rebased on `main` with CI green (748 test files / 7847 tests).

Where it got to, so a future attempt does not rediscover it:

- The last review rounds found five inputs where the quoted-text parser published a **truncated** status with no notice — a one-line HTML reply beginning "On" lost everything; a sentence containing a clock time was cut; a pasted bounce header ate the question after it; a postscript below a `--` delimiter was dropped. All are fixed on the branch with regression tests, but the parser is the risky part of this feature and deserves its own PR and its own scrutiny.
- One known limit is documented in the code rather than guessed at: a reply ending on a markdown `>` blockquote loses that run, because a client quote and a sender's quote are not distinguishable at that point.
- Two commits touch shared config (`lib/config/mediaStorage.ts`, `lib/config/fitnessStorage.ts`): a blank required storage variable used to boot a live-but-broken backend, and an empty path resolved to the process CWD. That fix is independent of reply-by-email and would stand on its own.

Natural split: (1) the blank-storage-value guard, (2) the token store + webhook + job, (3) the quoted-text parser, (4) the admin surface and env-block builder.

PR #1322, opened 2026-07-25
fix: redesign account and security emails to match the design system

llun:
## Summary

First of three PRs bringing every email the server sends in line with the approved [Email templates](https://claude.ai/design/p/a61009c5-7207-4e18-8d2b-a0cace8d672b?file=ui_kits%2Fweb%2Femails.html) design.

Today every template returns a **bare HTML fragment** — no document, no branding, no inline CSS, no footer. The whole "verify your email" body is one unstyled sentence with a link. This PR adds the shared skeleton the design specifies (600px, nested tables, fully inline styles, bulletproof button, hidden preheader) and moves the four account/security emails onto it.

The layout lands already used by four templates, so nothing dead ships. PR 2 puts the notification emails on it; PR 3 adds the fitness email.

## What changed

**New shared layout** — `lib/services/email/layout/`
- `theme.ts` — design tokens. Email has no stylesheet and no CSS variables, so every colour and size must be inlined literally; this is the one place they are written down.
- `blocks.ts` — seven block builders. Each emits the HTML **and** its plain-text twin, so a block cannot exist in one medium and be missing from the other.
- `actorDisplay.ts` — short name for headlines, two-letter monogram, stable per-handle colour.
- `renderEmail.ts` — wraps blocks in the document skeleton and derives the plain-text alternative from the same block list.

**Four account emails moved into template modules**, each returning a `RenderedEmail`:
`verifyEmail.ts`, `changeEmail.ts`, `resetPassword.ts`, and a redesigned `actorDeleted.ts`. Their call sites drop the inline subject/text/html literals; the surrounding failure handling (the 500 path, the reset-code restore) is untouched.

## Defects fixed along the way

- **Broken links on non-https deployments.** `app/api/v1/accounts/email/route.ts` and the password-reset route both built links from a hardcoded `https://${config.host}`, which is wrong when `ACTIVITIES_INSECURE_AUTH=true`. Both now use `getBaseURL()`, matching what `sendConfirmationEmail` already did.
- **Escaping is now structural.** It lives inside the block builders, so a template hands in plain strings and *cannot* forget it. URLs are protocol-checked to http/https/mailto; an unusable href degrades to plain text, and an unusable button URL renders no button rather than a dangerous one.
- **`actorDeleted.test.ts` was passing vacuously.** Its local `vi.mock('@/lib/config')` shadowed the global mock without `getBaseURL`, so the template threw and `deleteActorJob`'s catch swallowed it — a broken deletion email would have gone unnoticed. The mock now supplies `getBaseURL` and the test asserts the mail is actually sent.
- **The actor-deleted notice said "please contact us"** with no way to do so. It now links the admin-configured `instance.contactEmail`, falling back to the generic wording when none is set.
- **Missing route coverage.** Nothing asserted the email-change verification was sent at all, let alone where it pointed.

## Client-compatibility choices

All load-bearing, all commented in the code:
- `mso-hide:all` on the preheader — Outlook ignores `display:none` alone and would print it above the header
- button background on the `<td>`, not the anchor — Outlook drops the anchor's `border-radius` and would render white-on-white
- `table-layout:fixed` — Outlook ignores `word-break`, so a long verification URL would otherwise stretch the column
- explicit `width`/`height` on the logo — most clients block remote images, and the header must still read as three columns

Outlook renders square corners and Arial. That is the design's stated, accepted degradation.

## Testing

`prettier` clean · `eslint` 0 errors · `next build` ✅ · **6981 tests passed (719 files)**

New: `actorDisplay.test.ts` (21), `blocks.test.ts` (24), `renderEmail.test.ts` (18), plus per-template tests and the new route assertion.

## Verification

Rendered all four templates with `scripts/mock/renderEmailPreviews.ts` and checked them in a browser at 1280px and 375px. All four match the design: header, card, orange button, fallback URL, hairline-ruled note, and the correct account footer. `actor-deleted` correctly renders no button. The 600px column is fixed by design — mail clients scale it on mobile rather than reflowing.

> A browser is a lower bar than a mail client. Before this ships widely, one of these should also be sent to a real Gmail and Apple Mail inbox, dark mode included. Noted in `docs/maintenance.md`.

## Notes

- The design's button orange is `#E66A0F`; the app's `--primary` is `hsl(24 95% 46%)` = `#E55F06`. The design is the reviewed source of truth for email, so the literal value ships. Flagging the ~1.5% delta deliberately.
- `version` in `package.json` untouched.


llun:
## Review round 1 — findings and resolutions

Three sub-agent reviewers ran over the full diff: correctness/security, project conventions + test quality, and email-client compatibility. Findings and what changed:

### Blocking — fixed

**1. `app/api/v1/accounts/password/reset/request/route.test.ts` was passing for the wrong reason** (`4e6207583`)

The same trap already fixed in `deleteActorJob.test.ts`, still present here. The local `vi.mock('@/lib/config', …)` factory shadowed the global mock without `getBaseURL`, so `buildResetPasswordEmail` threw inside the route's `try` and the `catch` reported it as a delivery failure. **`sendMail` was never called — 0 times.**

Consequences: the three tests named "…when email sending fails" mocked a `sendMail` rejection that was never reached. They would have kept passing with `sendMail` deleted from the route entirely. The happy path had zero coverage of the email.

Fixed by supplying `getBaseURL`, adding the explanatory comment, and adding a happy-path assertion on the subject and link shape.

### Non-blocking — addressed

**2. Dead code** (`8604b7758`) — `actorDisplay.ts` (+161 test lines) and the `INSET_BACKGROUND` / `MAP_BACKGROUND` / `RADIUS_INSET` / `MONOGRAM_PALETTE` tokens have no consumer in this PR; they exist for the notification emails, which quote an actor. Removed — they land with the templates that use them. The PR description's "nothing dead ships" is now accurate.

**3. Coverage gaps** (`309f1dcee`) — the preheader was the one escaped field in the skeleton with no escaping test, and nothing pinned rendering with zero blocks. Both added.

**4. Undocumented convention** (`dc1cd3512`) — AGENTS.md requires a new convention to be recorded in `AGENTS.md` and `REVIEW.md`; neither had an email entry. Added: the module contract, escaping belonging to the block builders, the absolute-URL rule, and the `vi.mock` config trap that has now silently disabled an email in two separate test files.

### Clean

**Security/correctness — no defects found.** The reviewer enumerated every interpolation site in `blocks.ts` and `renderEmail.ts` and confirmed complete escaping coverage, then tested `safeUrl` against 28 bypass attempts — leading whitespace/C0 control chars, `JAVASCRIPT:` casing, embedded tab/newline (`ja\tvascript:`), `data:`, `vbscript:`, protocol-relative `//evil.com`, and attribute-breakout payloads. All rejected. It specifically probed whether URL-parser normalization could let a string parse as `https:` while a mail client re-reads it as `javascript:` — it cannot, since HTML applies the same stripping.

It also confirmed the password-reset restore path is now a *superset* of before (a template failure triggers the restore too), that `encodeURIComponent` is a no-op for base64url and round-trips correctly through all three consuming endpoints, and that `getResolvedServerSettings` cannot throw in `deleteActorJob` — it catches internally and returns cached/default values, so `contactEmail` degrades to the no-contact copy and the email still sends.

**Conventions — clean.** No `console.*` outside `scripts/`, no `../` imports (including in `vi.mock` paths), no env reads outside `lib/config/`, server/client boundary intact, test naming correct, `package.json` `version` untouched. The reviewer swept every other test file that mocks `@/lib/config` and confirmed the reset route was the only remaining instance of the trap.

### Verification after fixes

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **6963 tests passed (718 files)**

Test count moves from 6981/719 because `actorDisplay.test.ts` (21 tests) defers to the next PR; net new here is the preheader, empty-card, and reset-route assertions.

llun:
## Review round 1 (cont.) — email-client compatibility

The third reviewer audited the generated documents against Gmail, Apple Mail, Outlook desktop/Word, Outlook.com and Yahoo. It confirmed structure is sound — all four documents tag-stack-parse balanced, `role="presentation"` on 4/4 tables, `lang` present, MSO conditional valid and correctly placed, 3.4–4.0 KB (no Gmail clipping), and **every design-spec value matches exactly**.

It also found five real rendering defects. All fixed.

### `8467b8a55` — mobile, Apple Mail, Outlook headings

- **The column could not shrink on a phone.** It was a rigid `width:600px`, and the `width=device-width` viewport meta is precisely what suppresses iOS Mail's shrink-to-fit — so a 600px table just overflowed a 375pt screen. Most opens are mobile. Outlook reads the `width` *attribute*, so keeping `width="600"` and moving the style to `width:100%;max-width:600px` costs nothing there. **Verified in a browser: 375px viewport now fits; it was 632px before.**
- **Apple Mail linkified the footer.** Data detectors auto-linked the recipient address and "expires in 24 hours" as blue underlined links inside muted grey copy. Added `format-detection` and `x-apple-disable-message-reformatting`.
- **Outlook mixed two typefaces.** `h1` was missing from the MSO font override, so the headline stayed on the inline stack (Segoe UI) while everything else became Arial.
- **The inbox preview backfilled with the header.** Only 16 characters of preheader padding against Gmail's ~100-character snippet, so the preview read "…account on example.com. **Activities example.com Verify your email**". Padding now outruns the snippet.
- **The logo alt duplicated the wordmark.** With images blocked the header read "Activities Activities example.com". The logo sits beside real text, so it is decorative — empty alt.

### `37526be60` — Outlook button and long links

- **The button collapsed to a ~20px strip.** The Word engine ignores `display:block`, so the anchor stayed inline and its vertical padding added no height; the `td` carried the background but no padding of its own. `mso-padding-alt` restores it and no other client sees the property.
- **Long links overflowed the card.** `word-break:break-all` is unsupported in Word. The links are also longer than the preview suggested — see below. Added `word-wrap`, the Microsoft-origin property Word does honour, plus the same guard on `paragraph`/`note`.

### `d71236b29` — the preview was hiding the bug

The fixtures used 10-character codes where production emits **43** (`crypto.randomBytes(32).toString('base64url')`). That rendered a 65-character link sitting comfortably inside the card, where a real link is ~98 characters against 544px of interior. A visual review could never have caught the overflow. Fixtures now match production length — and the re-rendered preview shows the link wrapping correctly.

### Deliberately not changed

**`color-scheme: light dark` with no `prefers-color-scheme` rules.** The reviewer is right that this is a claim we don't honour: it tells Apple Mail and Outlook macOS/iOS to skip their automatic dark adaptation, so dark-mode readers get the full-brightness card. `content="light"` would be the honest declaration.

I've left it as-is because **the approved design specifies `light dark` in all 11 templates**, and silently diverging from a reviewed spec on a debatable point seemed worse than flagging it. @llun — worth a design decision: either drop to `light`, or add real dark-mode rules in a follow-up.

Two other low-severity notes were judged not worth the churn: unitless line-heights without `mso-line-height-rule:exactly` (minor Outlook leading drift), and grapheme-cluster handling in the monogram helper (which has moved to the next PR anyway).

### Verification after fixes

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **6967 tests passed (718 files)** · re-rendered and re-checked in a browser at 1280px and 375px.

llun:
## Review round 2 — findings and resolutions

Re-reviewed after the round-1 fixes, since fixes can introduce new problems. The reviewer verified the sound fixes empirically — measuring the rendered documents in a browser at 1400/1000/375px, and reverting each fix to confirm its test actually fails without it.

**Confirmed sound:** the `theme.ts` token removal orphaned nothing; `width:100%;max-width:600px` measures exactly 600px centered at 1400px and 343px at 375px with no overflow (the pre-fix version overflows at 632px, so the bug was real); `alt=""` keeps the header at 28×28 with images blocked; `repeat(30)` costs 360 bytes of a 4.5 KB document; and every `renderEmail.ts` assertion fails when its fix is reverted.

It found six more real problems. All fixed.

### `5c256976b` — I had guessed, and the guess was load-bearing

My comment claimed *"the `width` ATTRIBUTE is what Outlook's Word engine reads, so it still gets a 600px table."* That was unverified and runs against the general rule — the reviewer measured that in standards engines the inline style **beats** the attribute, which is precisely why the column can shrink. Word doesn't support `max-width`, so if it reads the style the same way, the card stretches to the full reading pane: a regression the old fixed width did not have.

Replaced with the MSO ghost table, which removes the question entirely — Outlook sees a 600px table, everything else sees a comment. Verified the ghost is invisible in a browser and the layout is unchanged.

### `19d1c5d07` — half a recipe, and three untested fixes

- **`mso-padding-alt` was only on the cell.** Word *does* apply horizontal padding to inline elements, so without the matching `mso-padding-alt:0` on the anchor the button gets the cell's 20px plus the anchor's 20px per side. Added.
- **Every fix in the previous commit was untested.** The reviewer reverted each and watched the suite stay green: `mso-padding-alt` → 97 passed, `word-wrap` on `fallbackUrl` → 97 passed, `word-wrap` on `paragraph`/`note` → 97 passed. All three could have been dropped by a later refactor as dead style — and being Outlook-only, nothing in a browser would have shown it. Now guarded.

### `a22cd8111` — the docs I added were wrong in three ways

AGENTS.md is normative for agents, so an inaccuracy there is a defect, not a typo:

- It described a **`SanitizedBody` type that has never existed** in any source file — `git log -S SanitizedBody --all` returns only the docs commit itself. It belongs to the notification templates that aren't written yet. Nothing in the layout emits an unescaped value today.
- It stated the module contract **as universal when only 4 of 11 templates follow it**. The other seven still export the old three-getter trio, write raw markup, and hardcode `https://${config.host}` — the exact pattern the section calls wrong. A future agent would have read the migration as finished. Now carved out explicitly, the way the `fetch()` rule names its frozen legacy list.
- It claimed the preview script renders **"every template" when it covers 4**, in three places. A change to `reply.ts` would have looked verifiable when it isn't previewable at all.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **6970 tests passed (718 files)** · re-rendered and re-checked in a browser.

### Still open for a human

The `color-scheme: light dark` question from round 1 stands — the design specifies it, but we ship no `prefers-color-scheme` rules, so it tells Apple Mail to skip its own dark adaptation. And per the AGENTS.md note this PR itself adds: the Outlook-specific fixes (ghost table, `mso-padding-alt`, `mso-hide:all`) are **not observable in a browser** and want one real send before this reaches volume.

llun:
## Review round 3 — findings and resolutions

Round 3 verified the round-2 fixes hold. Notably it parsed the generated document under **both** interpretations of the MSO conditionals — as comments (every other client) and as real markup (Outlook) — and confirmed the tag tree is balanced either way, with the ghost `<td>` correctly wrapping the real table inside the centered cell. It also re-confirmed every new test is non-vacuous by reverting each fix individually, and ran the full sweep: build exit 0, eslint 0 errors, prettier clean, 6970 tests passing.

Two findings. One I've fixed as reported; on the other I reached the opposite conclusion.

### `6567c121c` — my docs were wrong again, in a new way

Round 2's correction said all seven notification templates hardcode `https://${config.host}`. Verified against the code: **only four do.**

| template | reality |
|---|---|
| `like`, `mention`, `reply`, `reblog` | hardcode `https://${config.host}` |
| `follow`, `followRequest` | link the **remote** `actor.id`, never touching the host |
| `activityImport` | no link at all |

The distinction matters for the migration ahead: a wrong *scheme* and a wrong *destination* are different problems, and pointing at a remote actor URI instead of the local profile is the more serious one. Corrected in both files.

### `c32f0839f` — `{ tight: true }`: right analysis, wrong conclusion

The reviewer found that `tight` has no visible effect and recommended moving or deleting it. The measurement is correct — I confirmed it in the browser:

```
p1 (mb 20px) → p2   : gap 20px
p2 (mb  4px) → note : gap 20px   ← the 4px collapses with the note's 20px margin-top
```

But the recommendation would make things worse. **Word does not collapse margins**, so in Outlook the values sum:

| | standards engines | Outlook |
|---|---|---|
| with `tight` (4px) | 20px | **24px** |
| without `tight` (20px) | 20px | **40px** |

The design intends 20px. `tight` keeps Outlook within 4px of that; removing it puts Outlook 20px off. It is not dead code — it is compensating for exactly the engine difference the reviewer identified, in the opposite direction from the one suggested.

Kept, and documented on both the option and its test so the next reader doesn't re-derive "this does nothing". The test asserts the CSS string rather than a rendered gap for the same reason: the only client where the difference is visible is the one a browser can't show.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **6970 tests passed (718 files)**

### Cleared, not defects

The reviewer explicitly checked and cleared: `mso-padding-alt:0` placement (property order is irrelevant, and the unknown property doesn't break parsing of subsequent declarations — verified in a browser); the ghost table against `table-layout:fixed` and the `width="600"` attribute (both resolve to 600px in Word, so they can't disagree); "nothing in the layout emits an unescaped value" (audited every interpolation — now true, and `SanitizedBody` no longer appears anywhere); the preview-coverage and 43-char-code claims; and that `sendNotificationAlerts.test.ts`'s bare `vi.mock('@/lib/config')` resolves to the manual mock, which *does* export `getBaseURL` — so it is not a landmine waiting for PR 2.

llun:
## Review round 4 — merge readiness: **approved**

A fourth reviewer ran the full gate, traced all 18 doc claims to code, and read the whole diff fresh.

**Gate:** `prettier --check` clean · `eslint` exit 0 (0 errors; 31 pre-existing `no-explicit-any` warnings, none in the 37 email files) · `next build` exit 0 · **718 files, 6970 tests, 0 failures**.

**Verdict: approve, 0 merge-blockers.**

It verified the measurable claims rather than trusting them: the codes are exactly 43 chars, the verify link exactly 98, the longest preheader exactly 89, the card interior 544px, `/logo-nav.png` present, all four linked routes in the build output, and every interpolation in both layout files passing through `escapeHtml`. It ran the documented preview command verbatim — it works, and `git status` stayed clean, confirming the "no database is opened" claim. It also swept once more for the silent-failure trap and found **no third case**.

On the `{tight:true}` reasoning it confirmed the premise independently, including that `note`'s `border-top` correctly does *not* interfere — borders block parent/child collapsing, not adjacent-sibling.

### `04583655b` — four nits fixed rather than deferred

Three were judged follow-ups and one cosmetic, but two are the same class of doc overstatement rounds 2 and 3 already corrected, so they're worth closing now:

- **AGENTS.md said "the four account emails" inlined their body at the call site. Three did.** `actorDeleted` was already a template module on `main`; its problem was hand-written raw markup, not a call-site literal.
- **`docs/maintenance.md` listed Gmail and Apple Mail** for the real-inbox check but omitted **Outlook** — the only client where any of the `mso-` work or the ghost table does anything, and therefore the one a browser preview cannot stand in for. That omission sat in the file a maintainer actually reads before sending.
- The repo's single `vi.mocked(await import(...))` now uses the `vi.importMock` form AGENTS.md prescribes.
- A test name had the hairline rule on the wrong side of the note.

**Gate after the fixes:** `next build` exit 0 · **6970 tests passed**.

---

## Four rounds, closed

| Round | Found | Nature |
|---|---|---|
| 1 | 9 | A test passing while `sendMail` was never called; 5 client-rendering defects; dead code; missing convention docs |
| 2 | 6 | An unverified claim in my own comment about Outlook; a half-applied button recipe; 3 untested Outlook fixes; 3 false doc claims |
| 3 | 2 | A wrong doc claim; one finding where the analysis was right but the fix would have made Outlook worse |
| 4 | 4 nits | Approved |

The recurring theme is worth recording: **every round found something that could not fail loudly** — a test that passed with the code path unreachable, Outlook-only fixes with no assertions, a preview whose fixtures were too short to show the bug, and documentation asserting things that were never true. None of it would have been caught by CI.

PR #1324, opened 2026-07-25
minor: add the status reaction write API and outbound federation

llun:
Epic 5.1, PR 2 of 3. 5.1a ([#1320](https://github.com/llun/activities.next/pull/1320)) stored reactions and accepted them inbound; this makes them writable and sends them out. The picker UI is 5.1c.

## The write path

`reactStatus` / `unreactStatus` (`lib/services/reactions/reactStatus.ts`) resolve the status through `getReadableStatus`, validate, write, and — **only when the write actually changed state** — federate or notify. That gate is why 5.1a made `createStatusReaction`/`deleteStatusReaction` report whether they stored anything: a redelivered or capped reaction must not re-federate or re-notify.

**A locally-authored status notifies its author directly instead of federating.** This deliberately departs from how favourites work. `sendLike` posts to `status.actor.inboxUrl` unconditionally, so a local favourite notifies via a self-POST that comes back through our own inbox — but that cannot work for reactions: the inbound handler would find the reaction already stored, `changed` would be false, and the notification would be skipped. Federation and local notification are therefore exclusive here.

## Validation is stricter locally than inbound — on purpose

Inbound federation stays liberal (senders disagree about what a reaction may be, and storage is opaque). A reaction *we* originate has to round-trip, so it must be either:

- a **single emoji grapheme**, measured with `Intl.Segmenter` so a flag (regional-indicator pair), a ZWJ family, a skin-toned hand and a keycap each count as one — a bare letter or digit does not; or
- a **shortcode naming an enabled custom emoji on this instance**, which is also what supplies the outbound `Emoji` tag.

Anything else is `422`, including `shortcode@domain`: this instance cannot vouch for another's emoji. Removal deliberately skips revalidation so an emoji stays removable after an admin disables it.

## Routes — two dialects, one store

| Endpoint | Dialect | Scopes |
|---|---|---|
| `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` | Pleroma (primary) | `write` \| `write:favourites` |
| `GET /api/v1/pleroma/statuses/:id/reactions` | Pleroma | `read` \| `read:statuses`, auth optional |
| `GET /api/v1/pleroma/statuses/:id/reactions/:emoji` | Pleroma | `read` \| `read:statuses`, auth optional |
| `POST /api/v1/statuses/:id/react/:name` | glitch-soc | `write` \| `write:favourites` |
| `POST /api/v1/statuses/:id/unreact/:name` | glitch-soc | `write` \| `write:favourites` |

All five run one shared `reactionWriteHandler` over one service, so the dialects cannot drift; a parity test asserts the Pleroma `:emoji` segment and the glitch `:name` segment reach the service identically. Writes return the refetched `Status` so the caller sees its own reaction; reads return `{name, count, me, url, static_url, accounts}` in first-reaction order, with one batched account lookup rather than one per reaction. All four modules are registered in `routeScopeWiring.test.ts`.

## Outbound

`sendReaction`/`sendUndoReaction` emit the **Misskey shape** — a `Like` with the emoji on both `content` and `_misskey_reaction`, plus an `Emoji` tag for a custom one. That is the only spelling every family renders something for: Misskey/Firefish/Sharkey read it natively, Pleroma/Akkoma rewrite it into their own `EmojiReact`, and vanilla Mastodon — which has no `EmojiReact` handler at all and drops that activity silently — ignores `content` on a `Like` and shows a **favourite**.

That degradation is the deliberate trade-off, and it is now stated plainly in the compat doc: **reacting from here shows up as a favourite to someone on vanilla Mastodon.** The alternative (emitting FEP-c0e0 `EmojiReact`, which is cleaner) would be invisible to most of the network.

The activity id is `#emoji-reactions/<md5 status>/<md5 reaction>` — distinct per reaction *and* distinct from `sendLike`'s `#likes/<md5 status>`, so a receiver that dedups by activity id can never conflate a favourite with a reaction, or have an `Undo` retract the wrong one. Tests assert exactly that.

## Deviations from the plan

1. **`getReactionGroupKey` moved** from `lib/actions/emojiReaction.ts` to `lib/services/reactions/reactionGroupKey.ts` (with its tests), so the inbound action and the local service share one implementation instead of a service importing an action.
2. **`getStatusReactionActors` gained no pagination.** The plan's GET contract does not specify any, and Pleroma's does not either; a status with very many reactors returns them all. Related to the per-status rollup limitation already recorded in the compat doc, and best revisited together.
3. **Grouped notifications** deliberately left alone: `emoji_reaction` is not in `DEFAULT_GROUPABLE_TYPES` (Mastodon groups only favourite/reblog/follow by default), so reactions only reach the v2 grouped path when a client opts in with `grouped_types[]`.

## Verification

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **721 files / 7043 tests green**.

No UI in this PR, so no screenshots — the picker and chips land in 5.1c, which is where browser verification belongs.

### DIFF
--- a/app/api/v1/accounts/email-notifications/route.ts
+++ b/app/api/v1/accounts/email-notifications/route.ts
@@ -11,6 +11,7 @@ const EmailNotificationSettingsRequest = z.object({
   mention: z.boolean().optional(),
   reply: z.boolean().optional(),
   reblog: z.boolean().optional(),
+  emoji_reaction: z.boolean().optional(),
   activity_import: z.boolean().optional(),
   actorId: z.string().optional()
 })
--- a/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts
+++ b/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts
@@ -1,0 +1,82 @@
+import {
+  OAuthGuardAnyScope,
+  OptionalOAuthGuard,
+  corsErrorResponse
+} from '@/lib/services/guards/OAuthGuard'
+import { getStatusReactionList } from '@/lib/services/reactions/getStatusReactionList'
+import {
+  reactionRouteAttributes,
+  reactionWriteHandler
+} from '@/lib/services/reactions/reactionRouteHandlers'
+import { Scope } from '@/lib/types/database/operations'
+import { HttpMethod } from '@/lib/utils/http-headers'
+import { ERROR_404, apiResponse, defaultOptions } from '@/lib/utils/response'
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+// Pleroma/Akkoma dialect — the primary reaction surface, and the one with real
+// deployed client support (Husky, the Megalodon family). Not core Mastodon API.
+const CORS_HEADERS = [
+  HttpMethod.enum.OPTIONS,
+  HttpMethod.enum.GET,
+  HttpMethod.enum.PUT,
+  HttpMethod.enum.DELETE
+]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+const guardOptions = { errorResponse: corsErrorResponse(CORS_HEADERS) }
+const WRITE_SCOPES = [Scope.enum.write, Scope.enum['write:favourites']]
+
+interface Params {
+  id: string
+  emoji: string
+}
+
+export const GET = traceApiRoute(
+  'getStatusReactionsByEmoji',
+  OptionalOAuthGuard<Params>(
+    [Scope.enum.read, Scope.enum['read:statuses']],
+    async (req, { database, currentActor, params }) => {
+      const { id, emoji } = await params
+      const reactions = await getStatusReactionList({
+        database,
+        currentActor,
+        statusId: idToUrl(id),
+        name: emoji
+      })
+      if (!reactions) {
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_404,
+          responseStatusCode: 404
+        })
+      }
+
+      return apiResponse({ req, allowedMethods: CORS_HEADERS, data: reactions })
+    },
+    guardOptions
+  ),
+  { addAttributes: reactionRouteAttributes }
+)
+
+export const PUT = traceApiRoute(
+  'addStatusReaction',
+  OAuthGuardAnyScope<Params>(
+    WRITE_SCOPES,
+    reactionWriteHandler('react', CORS_HEADERS),
+    guardOptions
+  ),
+  { addAttributes: reactionRouteAttributes }
+)
+
+export const DELETE = traceApiRoute(
+  'removeStatusReaction',
+  OAuthGuardAnyScope<Params>(
+    WRITE_SCOPES,
+    reactionWriteHandler('unreact', CORS_HEADERS),
+    guardOptions
+  ),
+  { addAttributes: reactionRouteAttributes }
+)
--- a/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts
+++ b/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts
@@ -56,7 +56,9 @@ export const GET = traceApiRoute(
 
       return apiResponse({ req, allowedMethods: CORS_HEADERS, data: reactions })
     },
-    guardOptions
+    // `any`, so a token holding only the granular `read:statuses` is accepted;
+    // the guard otherwise defaults to requiring every listed scope.
+    { ...guardOptions, matchMode: 'any' }
   ),
   { addAttributes: reactionRouteAttributes }
 )
--- a/app/api/v1/pleroma/statuses/[id]/reactions/route.ts
+++ b/app/api/v1/pleroma/statuses/[id]/reactions/route.ts
@@ -1,0 +1,52 @@
+import { OptionalOAuthGuard } from '@/lib/services/guards/OAuthGuard'
+import { getStatusReactionList } from '@/lib/services/reactions/getStatusReactionList'
+import { Scope } from '@/lib/types/database/operations'
+import { HttpMethod } from '@/lib/utils/http-headers'
+import { ERROR_404, apiResponse, defaultOptions } from '@/lib/utils/response'
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+// Pleroma/Akkoma dialect, auth optional — an anonymous reader sees the same
+// rollups with `me: false`. Not core Mastodon API.
+const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
+
+export const OPTIONS = defaultOptions(CORS_HEADERS)
+
+interface Params {
+  id: string
+}
+
+export const GET = traceApiRoute(
+  'getStatusReactions',
+  OptionalOAuthGuard<Params>(
+    [Scope.enum.read, Scope.enum['read:statuses']],
+    async (req, { database, currentActor, params }) => {
+      const statusId = idToUrl((await params).id)
+      const reactions = await getStatusReactionList({
+        database,
+        currentActor,
+        statusId
+      })
+      if (!reactions) {
+        return apiResponse({
+          req,
+          allowedMethods: CORS_HEADERS,
+          data: ERROR_404,
+          responseStatusCode: 404
+        })
+      }
+
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: reactions
+      })
+    }
+  ),
+  {
+    addAttributes: async (_req, context) => {
+      const params = await context.params
+      return { statusId: params?.id || 'unknown' }
+    }
+  }
+)
--- a/app/api/v1/pleroma/statuses/[id]/reactions/route.ts
+++ b/app/api/v1/pleroma/statuses/[id]/reactions/route.ts
@@ -1,4 +1,7 @@
-import { OptionalOAuthGuard } from '@/lib/services/guards/OAuthGuard'
+import {
+  OptionalOAuthGuard,
+  corsErrorResponse
+} from '@/lib/services/guards/OAuthGuard'
 import { getStatusReactionList } from '@/lib/services/reactions/getStatusReactionList'
 import { Scope } from '@/lib/types/database/operations'
 import { HttpMethod } from '@/lib/utils/http-headers'
@@ -41,7 +44,11 @@ export const GET = traceApiRoute(
         allowedMethods: CORS_HEADERS,
         data: reactions
       })
-    }
+    },
+    // `any`, so a token holding only the granular `read:statuses` is accepted;
+    // the guard otherwise defaults to requiring every listed scope. The CORS
+    // error responder keeps guard-level 401/403 readable cross-origin.
+    { errorResponse: corsErrorResponse(CORS_HEADERS), matchMode: 'any' }
   ),
   {
     addAttributes: async (_req, context) => {
--- a/app/api/v1/routeScopeWiring.test.ts
+++ b/app/api/v1/routeScopeWiring.test.ts
@@ -34,6 +34,23 @@ vi.mock('@/lib/utils/traceApiRoute', () => ({
 // guarded methods (deduped). Kept independent of the route source so a wrong
 // literal is caught.
 const EXPECTED: Array<{ module: string; scopes: string[] }> = [
+  // status emoji reactions (ecosystem dialects, one store)
+  {
+    module: '@/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route',
+    scopes: ['read', 'read:statuses', 'write', 'write:favourites']
+  },
+  {
+    module: '@/app/api/v1/pleroma/statuses/[id]/reactions/route',
+    scopes: ['read', 'read:statuses']
+  },
+  {
+    module: '@/app/api/v1/statuses/[id]/react/[name]/route',
+    scopes: ['write', 'write:favourites']
+  },
+  {
+    module: '@/app/api/v1/statuses/[id]/unreact/[name]/route',
+    scopes: ['write', 'write:favourites']
+  },
   // account actions
   {
     module: '@/app/api/v1/accounts/[id]/follow/route',
--- a/app/api/v1/routeScopeWiring.test.ts
+++ b/app/api/v1/routeScopeWiring.test.ts
@@ -253,6 +253,16 @@ const MULTI_METHOD: Array<{
   methods: Record<string, string[]>
 }> = [
   {
+    // Reads and writes on one reaction URL: a GET<->mutation scope swap here
+    // would pass the flattened EXPECTED assertion above.
+    module: '@/app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route',
+    methods: {
+      GET: ['read', 'read:statuses'],
+      PUT: ['write', 'write:favourites'],
+      DELETE: ['write', 'write:favourites']
+    }
+  },
+  {
     module: '@/app/api/v1/notifications/route',
     methods: {
       GET: ['read', 'read:notifications'],
--- a/lib/actions/emojiReaction.ts
+++ b/lib/actions/emojiReaction.ts
@@ -3,6 +3,7 @@ import { z } from 'zod'
 import { getConfig } from '@/lib/config'
 import { Database } from '@/lib/database/types'
 import { createNotificationWithPolicy } from '@/lib/services/notifications/createNotificationWithPolicy'
+import { sendNotificationAlerts } from '@/lib/services/notifications/sendNotificationAlerts'
 import { shouldCreateNotification } from '@/lib/services/notifications/shouldNotify'
 import { getReactionGroupKey } from '@/lib/services/reactions/reactionGroupKey'
 import {
@@ -228,7 +229,7 @@ export const emojiReactionRequest = async ({
     return
   }
 
-  await createNotificationWithPolicy(database, {
+  const notification = await createNotificationWithPolicy(database, {
     actorId: status.actorId,
     type: NotificationType.enum.emoji_reaction,
     sourceActorId: activity.actor,
@@ -236,6 +237,33 @@ export const emojiReactionRequest = async ({
     reactionName: name,
     groupKey: getReactionGroupKey(status.id, name)
   })
+  if (!notification || notification.filtered) return
+
+  // Fan out the push alert, mirroring `likeRequest`. Without this an inbound
+  // federated reaction records a notification row but never reaches the
+  // recipient's device — so the whole reaction push path would only ever fire
+  // for a local actor reacting to a local post. Fire-and-forget: alert delivery
+  // must not fail the inbound activity. No email content: there is no reaction
+  // email template yet, and sendNotificationAlerts only mails events with one.
+  database
+    .getActorFromId({ id: activity.actor })
+    .then((sourceActor) => {
+      if (!sourceActor) return
+      sendNotificationAlerts({
+        database,
+        actorId: status.actorId,
+        sourceActorId: activity.actor,
+        sourceActor,
+        statusId: status.id,
+        events: [
+          {
+            type: NotificationType.enum.emoji_reaction,
+            notificationId: notification.id
+          }
+        ]
+      })
+    })
+    .catch(() => undefined)
 }
 
 /**
--- a/lib/actions/emojiReaction.ts
+++ b/lib/actions/emojiReaction.ts
@@ -4,14 +4,13 @@ import { getConfig } from '@/lib/config'
 import { Database } from '@/lib/database/types'
 import { createNotificationWithPolicy } from '@/lib/services/notifications/createNotificationWithPolicy'
 import { shouldCreateNotification } from '@/lib/services/notifications/shouldNotify'
+import { getReactionGroupKey } from '@/lib/services/reactions/reactionGroupKey'
 import {
-  MAX_NOTIFICATION_GROUP_KEY_LENGTH,
   MAX_REACTION_NAME_LENGTH,
   MAX_STORED_REACTION_NAME_LENGTH
 } from '@/lib/services/statuses/reactionLimits'
 import { EmojiReact, Like } from '@/lib/types/activitypub'
 import { NotificationType } from '@/lib/types/database/operations'
-import { getHashFromString } from '@/lib/utils/getHashFromString'
 import { logger } from '@/lib/utils/logger'
 
 // A reaction-bearing activity is either a litepub `EmojiReact` (Pleroma/Akkoma,
@@ -193,20 +192,6 @@ const getReactionEmojiUrl = ({
 
 const getReactionTarget = (activity: ReactionActivity) =>
   typeof activity.object === 'string' ? activity.object : activity.object.id
-
-/**
- * Groups an actor's reaction notifications per (status, emoji). Both parts are
- * remote-controlled and unbounded-ish — the status id is a URL and the name can
- * be a full `shortcode@domain` — so the readable form is used only while it fits
- * the varchar(255) column, with a digest of the same inputs as the overflow
- * fallback. The digest is deterministic, so a group stays stable across
- * deliveries either way.
- */
-export const getReactionGroupKey = (statusId: string, name: string) => {
-  const groupKey = `emoji_reaction:${statusId}:${name}`
-  if (groupKey.length <= MAX_NOTIFICATION_GROUP_KEY_LENGTH) return groupKey
-  return `emoji_reaction:${getHashFromString(`${statusId}:${name}`)}`
-}
 
 /**
  * Store an inbound emoji reaction. A reaction is never a favourite: it writes no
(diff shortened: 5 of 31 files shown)


app/api/v1/pleroma/statuses/[id]/reactions/route.ts
```diff
@@ -0,0 +1,59 @@
...
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: reactions
+      })
+    },
+    // `any`, so a token holding only the granular `read:statuses` is accepted;
+    // the guard otherwise defaults to requiring every listed scope. The CORS
+    // error responder keeps guard-level 401/403 readable cross-origin.
+    { errorResponse: corsErrorResponse(CORS_HEADERS), matchMode: 'any' }
```
Comment (llun) [inline app/api/v1/pleroma/statuses/[id]/reactions/route.ts:51]: **CONFIRMED — the reaction GETs required *both* scopes.** `OptionalOAuthGuard` defaults `matchMode` to `'all'`, and the scope hierarchy is coarse→granular only, so `hasGrantedScope(['read:statuses'], 'read')` is false. A token holding only the granular `read:statuses` got a **401 — worse than an anonymous caller**, which gets a 200 with `me: false`. It also contradicted this PR's own doc line ("the reads take `read` or `read:statuses`").

A verifier reproduced it by driving the real guard with a seeded opaque token: `['read']` → 200, `['read:statuses']` → 401, `['read:statuses']` + `matchMode: 'any'` → 200, anonymous → 200.

Every other multi-scope optional-auth route in the repo passes `'any'` (19 call sites), and `routeScopeWiring.test.ts` records only the scope *array*, never the match mode — so nothing caught it. **Fixed** on both GETs. This route also gained the `corsErrorResponse` it was missing (below).

app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts
```diff
@@ -0,0 +1,84 @@
...
+          data: ERROR_404,
+          responseStatusCode: 404
+        })
+      }
+
+      return apiResponse({ req, allowedMethods: CORS_HEADERS, data: reactions })
+    },
+    // `any`, so a token holding only the granular `read:statuses` is accepted;
+    // the guard otherwise defaults to requiring every listed scope.
+    { ...guardOptions, matchMode: 'any' }
```
Comment (llun) [inline app/api/v1/pleroma/statuses/[id]/reactions/[emoji]/route.ts:61]: **CONFIRMED — same missing `matchMode: 'any'` on the per-emoji GET**, and an inconsistency *within one file*: the PUT/DELETE use `OAuthGuardAnyScope` and so accept granular-only `write:favourites`, while the GET rejected granular-only `read:statuses`.

**Fixed** with `{ ...guardOptions, matchMode: 'any' }`, leaving `guardOptions` intact for the writes.

app/api/v1/pleroma/statuses/[id]/reactions/route.ts
```diff
@@ -0,0 +1,59 @@
...
+} from '@/lib/services/guards/OAuthGuard'
+import { getStatusReactionList } from '@/lib/services/reactions/getStatusReactionList'
+import { Scope } from '@/lib/types/database/operations'
+import { HttpMethod } from '@/lib/utils/http-headers'
+import { ERROR_404, apiResponse, defaultOptions } from '@/lib/utils/response'
+import { traceApiRoute } from '@/lib/utils/traceApiRoute'
+import { idToUrl } from '@/lib/utils/urlToId'
+
+// Pleroma/Akkoma dialect, auth optional — an anonymous reader sees the same
+// rollups with `me: false`. Not core Mastodon API.
```
Comment (llun) [inline app/api/v1/pleroma/statuses/[id]/reactions/route.ts:13]: **CONFIRMED — CORS-enabled route without a CORS error responder.** This module exports `OPTIONS`, but its guard got no `errorResponse: corsErrorResponse(CORS_HEADERS)`, so guard-level 401/403/500 came back through `apiErrorResponse` **without CORS headers** — a cross-origin client sees an opaque browser failure instead of the error. AGENTS.md is explicit that OPTIONS-exporting routes use `apiResponse` even for errors.

**Fixed** in the same options object as the `matchMode` change.

lib/services/reactions/getStatusReactionList.ts
```diff
@@ -0,0 +1,95 @@
...
+    statusId,
+    currentActor,
+    withReplies: false
+  })
+  if (!status) return null
+
+  // The `:emoji` URL segment is free text, and the write path stores a custom
+  // emoji colon-free — so normalise identically here or `GET .../reactions/:x:`
+  // would find nothing that `PUT .../reactions/:x:` had just created.
+  const storedName = name ? normalizeStoredReactionName(name) : undefined
```
llun, line 41: **The read path matched the `:emoji` segment verbatim while the write path normalises it.** `PUT .../reactions/:partyparrot:` stores `partyparrot` (colon-free, so the rollup can resolve the image from `customEmojis`), but `GET .../reactions/:partyparrot:` filtered on the raw `:partyparrot:` and returned nothing — the two surfaces disagreeing on the very same URL. Trimming differed too.

**Verdict: refuted** (the reviewer's mechanics were accepted; the severity was argued down).

**Fixed anyway** — a GET that cannot find what the PUT just created on the identical URL is a bug regardless of how one grades it. Both paths now go through one shared `normalizeStoredReactionName`.

lib/services/reactions/reactStatus.ts
```diff
@@ -0,0 +1,229 @@
...
+  if (!status) return { ok: false, reason: 'not-found' }
+
+  const resolved = await resolveLocalReaction(database, name)
+  if (!resolved) return { ok: false, reason: 'invalid-emoji' }
+
+  // Reacting to a boost reacts to the post it boosts. The serializer renders an
+  // Announce wrapper with empty reactions by design (they belong to the
+  // original, and surface on `reblog`), so storing against the wrapper would
+  // federate a reaction that is then invisible on every Status entity.
+  const target = getOriginalStatus(status)
```
llun, line 140: **Reacting to a boost stored the reaction against the `Announce` wrapper**, which the serializer renders with `reactions: []` by design (they belong to the original and surface on `reblog`). So the reaction was stored, federated *and* notified — then invisible on every Status entity.

**Verdict: refuted**, on the grounds that it clones the pre-existing favourite path.

**Fixed anyway.** The parallel does not hold: a favourite on a boost still surfaces through the counters, whereas a reaction on a wrapper is unreachable by any client. Reactions now resolve through `getOriginalStatus`, which is also what Mastodon does for favourites on boosts. Covered by a test asserting the reaction lands on the boosted post and *not* on the wrapper.

lib/services/reactions/reactStatus.ts
```diff
@@ -0,0 +1,229 @@
...
+  const existing = await database.getStatusReactionRollups({
+    statusIds: [target.id],
+    currentActorId: currentActor.id
+  })
+  const mine = existing.filter((rollup) => rollup.me)
+  if (
+    mine.length >= MAX_REACTIONS_PER_ACTOR &&
+    !mine.some((rollup) => rollup.name === resolved.name)
+  ) {
+    return { ok: false, reason: 'cap-reached' }
```
llun, line 154: **A reaction past the per-actor cap was a silent no-op.** The storage layer drops it and the route answered **200 with a Status that does not contain the caller's reaction** — the exact silent-failure shape `REVIEW.md` warns about.

**Verdict: refuted** as deliberate and documented.

**Fixed anyway**, cheaply: the service counts the actor's own rollups (`me === true`) before writing and returns `cap-reached` → **422**. No new DB method, one extra query on the write path, and re-reacting with an emoji already held stays idempotent rather than reporting the cap. Racy under concurrency, but the cap was always advisory.

lib/services/reactions/reactStatus.ts
```diff
@@ -0,0 +1,229 @@
...
+    name: storedName
+  })
+  if (changed && !target.isLocalActor) {
+    // A vanilla-Mastodon receiver resolves `Undo{Like}` by (account, status),
+    // not by activity id — and it stored our reaction Like as a favourite. If
+    // the actor also genuinely favourited this status, sending the Undo would
+    // destroy that favourite there. Withhold it: the favourite is the state
+    // that should survive, and the reaction is already gone locally.
+    if (
+      await database.isActorLikedStatus({
```
llun: **The most serious finding: retracting a reaction destroyed the actor's genuine favourite on vanilla Mastodon.** Mastodon resolves `Undo{Like}` by **(account, status)**, not by activity id — so the distinct per-reaction activity id we carefully mint buys nothing on removal. Our reaction Like landed there as a favourite, and Mastodon dedups a reaction and a real favourite into a single row, so `Undo` removed whichever survived.

**Verdict: refuted** as an unavoidable corollary of Like-based emission.

**Fixed anyway, because it is avoidable**: the `Undo` is now withheld when the actor still favourites the status. That is precisely the case where the favourite must survive — and where it does not favourite, the only favourite on the remote came from our reaction, so undoing it is correct. Both branches are tested, and the trade-off is now written down in the compat doc.

lib/services/reactions/getStatusReactionList.ts
```diff
@@ -0,0 +1,95 @@
...
+  const requestedActorIds = [...new Set(reactors.map((r) => r.actorId))]
+  const requestedActorIdSet = new Set(requestedActorIds)
+  const accounts = await database.getMastodonActorsFromIds({
+    ids: requestedActorIds
+  })
+  // Key by the actor URI decoded from the opaque `account.id`, falling back to
+  // `account.url` — for some actors `url` is a profile URL (`/@name`) rather
+  // than the actor URI, and keying on it alone drops them. Mirrors how
+  // getMastodonStatuses builds its account cache.
+  const accountByActorId = new Map<string, Mastodon.Account>()
```
llun, line 69: **Found while checking a reviewer's account-mapping claim — no reviewer reported this one.** The account map was keyed on `account.url`, but for some actors `url` is a profile URL (`/@name`) rather than the actor URI. Those reactors were silently dropped from the `GET .../reactions` response.

`getMastodonStatuses` already handles exactly this, keying on `idToUrl(account.id)` with an `account.url` fallback, and has a test named "keys hydrated account cache by actor id when account url is a profile url". Mirrored here, with a regression test that stubs the lookup to return profile URLs and asserts both reactors still resolve.

llun (commented):
Automated sub-agent review round 1 on this PR — three dimension reviewers (service layer · HTTP surface · outbound federation), each finding handed to an independent skeptic instructed to refute it.

**8 findings raised, 3 confirmed.** All 8 are addressed in 99fa1634: the 3 confirmed ones, plus 4 refuted ones where the mechanics the reviewer described were right even though the severity was disputed, plus one bug no reviewer raised that I found while checking their account-mapping claim.

Gate re-run: prettier → eslint (0 errors) → `next build` → `vitest run`, **722 files / 7053 tests green**.

lib/services/notifications/pushNotification.ts
```diff
@@ -29,6 +29,7 @@ const NOTIFICATION_TYPE_TO_ALERT: Partial<
   reblog: 'reblog',
   quote: 'quote',
   quoted_update: 'quoted_update',
+  emoji_reaction: 'pleroma:emoji_reaction',
```
llun, line 32: **CONFIRMED — the `pleroma:emoji_reaction` alert flag was advertised but inert.** I added it to `PushAlerts` and both defaults in round 0, and stopped there. It was missing from all three places that make an alert key mean anything:

- `NOTIFICATION_TYPE_TO_ALERT` — so `alertKey` was `undefined` and the `sub.alerts[alertKey] === false` filter was skipped entirely;
- `ALERT_KEYS` in the subscription route — so `parseAlertsInput` silently dropped the key from POST and PUT, and **no client could set it**;
- `getNotificationContent` — so the push rendered as the generic "New Notification / You have a new notification".

Since this PR is the first to actually push for reactions, the net effect was: a user's subscription reports `"pleroma:emoji_reaction": false`, their phone buzzes anyway with meaningless text, and there is no way to turn it off. Every other pushable type was wired into both maps when it was added, and `pushNotification.test.ts` even pins that invariant for `activity_import`.

**Fixed**: all three, plus the per-actor `emoji_reaction` toggle — the push-settings Zod schema omitted it and the notification settings page did not list it, so the actor-level gate was equally unreachable.

lib/services/reactions/reactStatus.test.ts
```diff
@@ -0,0 +1,407 @@
...
+      await reactStatus({
+        database,
+        currentActor: reactor,
+        statusId,
+        name: '🍎'
+      })
+    ).toMatchObject({ ok: true, changed: false })
+  })
+
+  it('does not federate again when nothing changed', async () => {
```
llun, line 226: **CONFIRMED — the idempotency test could not fail.** It reacted to `${ACTOR2_ID}/statuses/post-2`, which is authored by the reactor itself. `announceReaction` therefore takes the local branch (no `sendReaction`) and then bails at `shouldCreateNotification` (self-reaction, so no notification) — on the **first**, `changed: true` call just as much as on the repeat. Both assertions held regardless of the `if (changed)` guard, so deleting that guard left the entire suite green, and every duplicate `PUT` against a *remote* status would have re-posted the `Like` to the remote inbox unnoticed.

**Fixed**: the test now uses a remote-authored status and asserts `sendReaction` fires **exactly once** across two identical calls. I verified it is load-bearing by mutation — replacing `if (changed)` with `if (true)` fails it, and only it — then restored the file and confirmed the worktree was clean.

llun (commented):
Automated sub-agent review **round 2** — two reviewers over the round-1 fixes and the whole final diff.

**2 findings, both confirmed, both fixed** in f9b85c3b. Both were mine, and neither was caught by round 1. Dispositions inline.

Gate: prettier → eslint (0 errors) → `next build` → `vitest run`, **722 files / 7053 tests green**.

app/api/v1/accounts/email-notifications/route.ts
```diff
@@ -11,6 +11,7 @@ const EmailNotificationSettingsRequest = z.object({
   mention: z.boolean().optional(),
   reply: z.boolean().optional(),
   reblog: z.boolean().optional(),
+  emoji_reaction: z.boolean().optional(),
```
llun, line 14: **CONFIRMED — round 2's own fix broke the Email channel.** The settings page drives **both** the Email and Push columns from one shared `notificationTypes` list. Round 2 added `emoji_reaction` to that list and to the *push* route's schema, but not the email route's. A plain `z.object` strips an undeclared key silently, so the POST 200s and the UI says "Saved" while nothing is stored.

The load-bearing consequence is not the reaction row — it is the **master toggle**, recomputed as `some((nt) => stored[nt.key] !== false)`. With `emoji_reaction` missing from storage, `undefined !== false` is true, so switching the **entire Email channel off** reads back as ON after a reload. That worked correctly before this PR. A verifier reproduced it against the real handler: status 200, stored object contains the seven other keys and no `emoji_reaction`, master-on-after-reload `true`.

Worth recording the verifier's scope correction: no unwanted *email* is actually sent, because the reaction paths pass no `emailContent` and `sendNotificationAlerts` only emails events that carry it. The defect is the persistence regression, not spam.

**Fixed** with the one-line schema mirror, plus a route test asserting `emoji_reaction: false` round-trips into `emailNotifications`.

lib/database/sql/pushSubscription.ts
```diff
@@ -42,6 +42,13 @@ export const DEFAULT_PUSH_ALERTS: PushAlerts = {
   update: false,
   quote: false,
   quoted_update: false,
+  // Deliberately true, unlike every Mastodon-documented flag above.
+  // `pleroma:emoji_reaction` is an ecosystem extension that no Mastodon client
+  // will ever request, and `normalizeAlerts` resolves a key absent from a
+  // stored row to this default — so `false` would permanently suppress reaction
+  // pushes for every subscription that predates the key, i.e. all of them. An
+  // Akkoma-aware client that explicitly sets it false is still honoured.
+  'pleroma:emoji_reaction': true,
```
llun, line 51: **`pleroma:emoji_reaction` defaulted to false, which silently disabled reaction pushes for everyone.** `normalizeAlerts` starts from `DEFAULT_PUSH_ALERTS` and only overrides keys present in the stored row, and `parseStoredAlerts` runs on **read** — so every subscription created before this key existed (i.e. all of them) resolved it to `false`, and the filter suppressed the push. No Mastodon client will ever request an ecosystem-extension alert key to opt back in, so the flag round 2 wired up would have meant reaction pushes never fire at all for token subscriptions.

**Verdict: refuted** (the all-false default matches the Mastodon WebPushSubscription docs).

**Changed anyway to `true`**, because the Mastodon default does not apply to a key Mastodon does not define: `false` here is not a conservative default, it is an off switch with no reachable on position. The comment states the reasoning, an Akkoma-aware client that explicitly sets it false is still honoured, and both directions are now tested via `parseStoredAlerts`.

I also verified the two gates independently rather than trusting the fix: `skipSettingsCheck` only bypasses the actor-level re-check, so the per-subscription `alertKey` filter genuinely runs, and `shouldSendPushForNotification` reads the actor setting the settings page now writes.

llun (commented):
Automated sub-agent review **round 3** — two reviewers over the round-2 fixes and the whole diff.

**3 findings, 2 confirmed** (both the same root cause), fixed in 312a8f43. Round 3 caught a regression that round 2's own fix introduced, which is exactly what the loop is for.

Gate: prettier → eslint (0 errors) → `next build` → `vitest run`, **722 files / 7056 tests green**. CI green on the previous head.

lib/services/reactions/reactStatus.ts
```diff
@@ -0,0 +1,232 @@
...
+  // Removal accepts whatever is stored rather than re-validating: an emoji that
+  // was legal when it was added must stay removable after an admin disables it.
+  const storedName = normalizeStoredReactionName(name)
+  const target = getOriginalStatus(status)
+
+  const changed = await database.deleteStatusReaction({
+    statusId: target.id,
+    actorId: currentActor.id,
+    name: storedName
+  })
```
llun, line 224: **I reversed my own round-1 fix here, because round 4's verifier was right and I was wrong.**

Round 1 (and its round-4 extension) withheld the outbound `Undo{Like}` when the actor still held a favourite or another reaction, to stop Mastodon — which resolves `Undo{Like}` by *(account, status)* — from clearing more than intended. The verifier pointed out what I had missed: **every reaction-native receiver resolves the Undo by reaction content.** `undoEmojiReactionRequest` in this very repo does exactly that (`deleteStatusReaction({statusId, actorId, name})`), and Pleroma/Akkoma behave the same way. So withholding left the reaction **visible forever** on the Misskey family, Akkoma, and other Activity.next instances — the servers this feature is actually for — with nothing the user could do about it.

Weighed properly: a permanent, unrecoverable bug on the primary interop targets versus a cosmetic artifact on a server that never rendered the reaction and where the user can re-favourite. Not close. One activity shape cannot satisfy both Undo semantics and per-instance software detection was rejected in the epic design, so the compromise is now explicit — **always send** — with the reasoning in the code and the consequence written down in the compat doc. Tests now assert the Undo goes out in both situations.

Worth noting the loop worked in both directions here: rounds 1 and 4 pushed me toward a fix, and round 4's *verifier* pulled me back off it with a better argument.

lib/actions/emojiReaction.ts
```diff
@@ -243,14 +229,41 @@ export const emojiReactionRequest = async ({
...
     reactionName: name,
     groupKey: getReactionGroupKey(status.id, name)
   })
+  if (!notification || notification.filtered) return
+
+  // Fan out the push alert, mirroring `likeRequest`. Without this an inbound
+  // federated reaction records a notification row but never reaches the
+  // recipient's device — so the whole reaction push path would only ever fire
+  // for a local actor reacting to a local post. Fire-and-forget: alert delivery
+  // must not fail the inbound activity. No email content: there is no reaction
```
llun, line 246: **CONFIRMED — the inbound path never fanned out the push alert**, so everything this PR wired up (the `NOTIFICATION_TYPE_TO_ALERT` entry, the "Reaction" payload body, the `true` default, the user-facing "Reactions" toggle) only ever fired for a **local** actor reacting to a **local** post. A federated reaction — the dominant source on any real instance — recorded a notification row and never reached the recipient's device.

The epic plan assigns this fan-out to this PR ("5.1a's action grows the alert fan-out here"); I missed it. **Fixed** by mirroring `likeRequest`: resolve the source actor, then fire-and-forget `sendNotificationAlerts`, gated on a non-null unfiltered notification. No `emailContent` — there is no reaction email template yet, and `sendNotificationAlerts` only mails events that carry one. The verifier confirmed there is no double-push risk: `announceReaction` federates for remote-hosted statuses and notifies for local ones exclusively, so the inbound and local paths never both run for one reaction.

app/api/v1/routeScopeWiring.test.ts
```diff
@@ -235,6 +252,16 @@ const MULTI_METHOD: Array<{
   module: string
   methods: Record<string, string[]>
 }> = [
+  {
+    // Reads and writes on one reaction URL: a GET<->mutation scope swap here
```
llun, line 256: **The multi-method reactions route was registered only in the flattened `EXPECTED` table**, which dedupes scopes across all of a module's methods — so swapping the GET's read scopes with the PUT/DELETE's write scopes would still have produced the same flattened set and passed. That is precisely the gap the `MULTI_METHOD` table exists to close, and it is the same class of blind spot that let round 1's missing `matchMode` through.

**Fixed** by adding the route to `MULTI_METHOD` with per-method expectations: `GET: [read, read:statuses]`, `PUT`/`DELETE: [write, write:favourites]`.

llun (commented):
Automated sub-agent review **round 4** — two reviewers over the round-3 fixes and the whole diff.

**3 findings, 1 confirmed** — but the most valuable outcome of this round was a *refutation* that overturned one of my own earlier fixes. Addressed in 49beaffa and cb63f7df.

Gate: prettier → eslint (0 errors) → `next build` → `vitest run`, **722 files / 7058 tests green**.

lib/services/reactions/reactStatus.ts
```diff
@@ -0,0 +1,256 @@
...
+    withReplies: false
+  })
+  // A status readable when the actor reacted may since have narrowed its
+  // audience (the author edits its visibility, or an accepted follow goes away).
+  // Refusing here would trap the reaction forever: it keeps counting in every
+  // viewer's rollups and, for a remote post, is never retracted upstream — the
+  // same unrecoverable-for-the-user shape that made us always send the Undo.
+  // Holding the row is itself proof the actor could read the status at the time,
+  // so fall back on ownership, exactly as unfavourite/unbookmark/unreblog do.
+  const status =
```
llun, line 196: **CONFIRMED — a reaction became permanently unretractable once the post's audience narrowed.** `unreactStatus` returned `not-found` whenever `getReadableStatus` came back empty, so: B posts publicly, A reacts 🔥, B narrows the post to followers-only with `PUT /api/v1/statuses/:id` — and now **neither** dialect can remove A's reaction. It keeps counting in every remaining viewer's rollups, and for a remote-authored post the `Undo` is never federated either. Unfollowing gets you there too.

The verifier made the connection that sold it: this is the *same* unrecoverable-for-the-user failure mode that round 4's `Undo` reversal explicitly rejected, arrived at from a different direction. And there are three in-repo precedents doing the opposite — `unfavourite`, `unbookmark` (whose comment says "its visibility changed after they bookmarked it") and `unreblog` all fall back to an ownership check — with `REVIEW.md` codifying it: *"Mastodon-compat mutation responses return the affected entity even when the actor can't otherwise read it."*

**Fixed** the same way: fall back to `database.getStatus` (no visibility filter) and proceed only when the actor actually holds that reaction — holding the row is itself proof they could read the status when they reacted, so nothing leaks. An actor with no reaction on an unreadable status still gets `not-found`. `refetchedStatusResponse` already bypasses the visibility filter, so the response shape is unchanged. Verified by mutation that the new test fails without the fallback.

llun (commented):
Automated sub-agent review **round 5** — two reviewers over the round-4 changes and the whole diff.

**2 findings, 1 confirmed**, fixed in 62435a9d.

Gate: prettier → eslint (0 errors) → `next build` → `vitest run`, **722 files / 7060 tests green**.

llun:
Automated sub-agent review **round 6 — clean**. Two reviewers returned **zero findings**, closing the review loop per the AGENTS.md exit condition.

**Loop summary across 6 rounds: 19 findings raised, 11 real defects fixed, 1 earlier fix reversed.**

| Round | Raised | Confirmed | Outcome |
|---|---|---|---|
| 1 | 8 | 3 | 8 addressed (3 confirmed + 4 refuted-but-real + 1 I found myself) |
| 2 | 2 | 2 | both fixed — inert push-alert wiring, vacuous idempotency test |
| 3 | 3 | 2 | fixed a regression **round 2's own fix** introduced |
| 4 | 3 | 1 | 1 fixed, 1 test gap closed, **1 earlier fix reversed** |
| 5 | 2 | 1 | reaction was unretractable once visibility narrowed |
| 6 | 0 | — | clean |

Two rounds are worth singling out:

**Round 4 reversed my own round-1 fix.** I had withheld the outbound `Undo{Like}` to protect a Mastodon-side favourite. A verifier showed that reaction-native receivers resolve the Undo *by reaction content* — so withholding left the reaction visible forever on Misskey/Akkoma/Activity.next, the servers this feature is for. Always sending is now the documented compromise.

**Round 3 caught a regression created by round 2.** Adding `emoji_reaction` to the shared settings list wired the Push column but not Email, and since the email route silently strips undeclared keys, turning the **whole Email channel** off stopped persisting — for every notification type, not just reactions.

Final state on `72d1b464` (rebased onto main, which had moved three commits): prettier → eslint (0 errors) → `next build` → `vitest run` all green locally (**7324 tests**), CI green on the rebased head. The one earlier CI failure was the unrelated flaky `FitnessStatusDetail` `waitFor` test, cleared by the rebase re-run.

Merging.

PR #1325, opened 2026-07-25
fix: redesign notification emails to match the design system

llun:
## Summary

Second of three PRs bringing every email in line with the approved [Email templates](https://claude.ai/design/p/a61009c5-7207-4e18-8d2b-a0cace8d672b?file=ui_kits%2Fweb%2Femails.html) design. Builds on the shared layout from #1322.

Moves the six notification emails — follow, follow request, mention, reply, like, boost — onto that layout. **Ten of eleven templates are now migrated**; only the fitness activity import remains, and nothing sends it (PR 3).

## Defects fixed

These templates were the worst of the set. Four separate problems, all live:

- **Every hashtag link was dead.** `convertMarkdownText` renders hashtags as `<a href="/tags/foo">`. A mail client has no origin to resolve that against, so every hashtag in a mention or reply email pointed nowhere. New `absolutizeLinks` rewrites root-relative hrefs on both the markdown and sanitize paths — deliberately *not* protocol-relative `//host` URLs, which point at another origin and must not be silently re-homed.
- **`like` and `reblog` injected raw remote markup.** Both interpolated `status.text` straight into the HTML with no sanitizing. Only `mention`/`reply` had the pipeline; now all four share it.
- **`follow`/`followRequest` linked the remote `actor.id`**, dropping the reader on another server with no session and no follow button. They now link the local profile.
- **Four templates hardcoded `https://${config.host}`** in duplicated `getLocalStatusUrl` copies (one had drifted). Consolidated into `getEmailStatusUrl`, built on `getStatusDetailPath` so an email link is byte-identical to the web UI's.

The plain-text part was also printing `status.text` raw — for a remote actor, a wall of markup. It is now derived from the sanitized HTML.

## Copy change

`reblog.ts` says **"boosted"** in subject and body, per the design. Internal vocabulary is untouched — `NotificationType.reblog`, the stored value and the Mastodon API field are the fediverse wire term. The file keeps its name to preserve the 1:1 module↔notification-type mapping.

> Known, deliberate inconsistency: Settings still reads "Reblogs" while the email says "boosted". That label was scoped out.

## What changed

- `layout/actorDisplay.ts` — short name, two-letter monogram, stable per-handle colour (held back from #1322 as it had no consumer there)
- `layout/blocks.ts` — the `quote` builder, and `SanitizedBody`: the one value the layout emits unescaped, a named type precisely so a bare string can't be handed in by accident
- `templates/statusUrl.ts`, `templates/statusBody.ts` — the two shared helpers
- Six templates rewritten to `build*Email(): RenderedEmail`
- The recipient threaded through **9 call sites** (all already had it in scope — no extra DB reads); `EmailContent` is now `RenderedEmail & { recipientEmail }`

## Testing

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7242 tests passed (725 files)**

New `emailDesign.test.ts` — one table-driven conformance suite over all ten templates, rendered with a fixture whose every string slot carries a hostile sentinel: document shape, fluid column + Outlook ghost table, preheader, button count, footer variant, and that no hostile input, `javascript:` url, relative href or unresolved template value survives. **134 assertions.**

Two assertions there are deliberately narrower than the obvious version, both because the obvious one produces false failures:
- it asserts the sentinel never appears **verbatim**, rather than that `"><` is absent — that sequence occurs at every legitimate tag boundary
- the plain-text check looks for **layout markup**, not a bare `<` — the fixtures put one in the display name, and a literal angle bracket in a `text/plain` body is a character, not a tag

Writing the suite also corrected an assumption of mine: I expected the hostile fixtures to yield **zero** buttons via the `javascript:` fallback. They yield one — the hostile actor still produces a derivable local path, so `getEmailStatusUrl` never reaches its fallback. The refusal is now covered by a separate case using an actor-less status, which is the only way to reach it.

## Verification

Rendered all ten templates and checked them in a browser. Confirmed in the DOM that the boost email's `#running` hashtag now resolves to `https://llun.social/tags/running` — **zero relative links in the document**.

> Same caveat as #1322: the Outlook-specific work (ghost table, `mso-padding-alt`, `mso-hide:all`) is invisible in a browser by construction and wants one real send.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -262,18 +262,12 @@ section-navigation patterns; pick by section type.
 Emails go through one shared skeleton, so a design or copy change lands in one
 place instead of eleven.
 
-> **Migration in progress — this describes the target, not the whole tree yet.**
-> Only the four account/security emails are on the shared layout:
-> `verifyEmail.ts`, `changeEmail.ts`, `resetPassword.ts`, `actorDeleted.ts`.
-> The seven notification templates — `follow.ts`, `followRequest.ts`, `like.ts`,
-> `mention.ts`, `reply.ts`, `reblog.ts`, `activityImport.ts` — still export the
-> old `getSubject`/`getTextContent`/`getHTMLContent` trio, write raw markup, and
-> hand-maintain both media. Their links are broken in two different ways:
-> `like`/`mention`/`reply`/`reblog` build a local status URL from a hardcoded
-> `https://${config.host}`, while `follow`/`followRequest` link the **remote**
-> `actor.id` directly instead of the local profile page (`activityImport` has no
-> link at all). They are being migrated, not grandfathered: apply the rules below
-> to any template you touch, and do not copy the old shape into a new one.
+> **Migration in progress — one template left.** Ten of the eleven are on the
+> shared layout. Only `activityImport.ts` still exports the old
+> `getSubject`/`getTextContent`/`getHTMLContent` trio and writes raw markup —
+> and nothing sends it, so it is dead code until the fitness import wiring
+> lands. Apply the rules below when migrating it, and do not copy the old shape
+> into a new template.
 
 - **One module per email** in `lib/services/email/templates/`, exporting a single
   `build<Name>Email(params): RenderedEmail` (`{ subject, text, html }`). **Never
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -169,14 +169,11 @@ change doesn't touch.
 
 ## Emails
 
-- **Migration in progress.** Only the four account/security emails
-  (`verifyEmail`, `changeEmail`, `resetPassword`, `actorDeleted`) are on the
-  shared layout. The seven notification templates still use the old
-  `getSubject`/`getTextContent`/`getHTMLContent` trio with raw markup, and their
-  links are wrong in two different ways: `like`/`mention`/`reply`/`reblog`
-  hardcode `https://${config.host}`, while `follow`/`followRequest` link the
-  remote `actor.id` rather than the local profile. Apply the rules below to any
-  template a change touches; don't copy the old shape into a new one.
+- **Migration in progress — one template left.** Ten of eleven are on the shared
+  layout; only `activityImport.ts` still uses the old
+  `getSubject`/`getTextContent`/`getHTMLContent` trio, and nothing sends it.
+  Apply the rules below when migrating it; don't copy the old shape into a new
+  template.
 - Every migrated email is built by a `build<Name>Email(params): RenderedEmail`
   module in `lib/services/email/templates/`. No subject/HTML/text literals at a
   call site.
@@ -194,9 +191,9 @@ change doesn't touch.
   omission does not fail loudly: the template throws, the catch swallows it, and
   the test passes while the email silently stops sending. This has bitten twice.
 - Template changes are verified by rendering:
-  `./scripts/mock/renderEmailPreviews.ts` (see `docs/maintenance.md`). It covers
-  only the four migrated templates — a PR migrating another must add it to
-  `buildPreviews()`, and fixtures must be production-shaped (43-char codes).
+  `./scripts/mock/renderEmailPreviews.ts` (see `docs/maintenance.md`). A PR
+  migrating a template must add it to `buildPreviews()`, and fixtures must be
+  production-shaped (43-char codes).
 - Outlook-only properties (`mso-padding-alt` on both the button cell and its
   anchor, `mso-hide:all`, the MSO ghost table pinning the column to 600px) are
   load-bearing and invisible in a browser. Don't drop them as dead style.
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -129,8 +129,8 @@ Creates a test user for development/testing:
 
 Renders email templates to standalone HTML files so a template change can be
 checked visually. Covers the templates listed in `buildPreviews()` — currently
-the four account/security emails; the notification templates are not on the
-shared layout yet. Emails are not pages, so there is no dev-server route to open —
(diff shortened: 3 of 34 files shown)


llun:
## Review round 1 — one real vulnerability, fixed

### `286579e40` — a local author could put raw HTML in someone else's inbox

`getStatusBody` ran markdown **OR** sanitize. It has to be markdown **THEN** sanitize — the order `processStatusTextContent` uses for the web UI — because `marked` has shipped no sanitizer since v5.

I reproduced it before fixing. A local account posting:

```
Hi <img src="https://tracker.evil.example/p.gif" onerror="alert(1)"><script>alert(2)</script>
```

produced exactly that markup, verbatim, in the mention email delivered to whoever it mentioned:

```
<p>Hi <img src="https://tracker.evil.example/p.gif" onerror="alert(1)"><script>alert(2)</script></p>
```

`quote()` emits the body unescaped by design, so this sanitize step was the only thing between a post body and the recipient's mail client. `<script>` mostly won't execute in a mail client, but nothing else here needs to: a remote image is a **read receipt plus the reader's IP**, a `<style>` block can hide or replace the legitimate card, and arbitrary anchor markup is a phishing primitive.

The remote branch was always safe — `sanitizeText` handled it. Only the local-author path was exposed.

### The test that should have caught it couldn't

`emailDesign.test.ts` asserts `not.toContain('<script')` across all ten templates — and still missed this, because **every status fixture in it was `isLocalActor: false`**. The vulnerable branch was never exercised. That is the same failure mode as the vacuous tests found in #1322: an assertion that looks like coverage while the code path stays untouched.

Both suites now carry a local-author hostile fixture. Verified non-vacuous by reverting the fix: **5 tests fail**, all pass with it restored.

### Scope note

The either/or predates this PR — the old `mention.ts`/`reply.ts` had it, and the old `like`/`reblog` were worse still (raw `status.text`, no pipeline at all). But it's this PR's problem to fix: it introduces `SanitizedBody`, whose documentation asserts the value "has ALREADY been through the sanitize/markdown pipeline". Asserting an invariant the code doesn't enforce is worse than not asserting it.

### Cleared

The reviewer traced and cleared the rest:

- **`absolutizeLinks`** — verified against ten inputs. Rewrites `href="/a"` and `src="/a"`; leaves `//evil`, absolute URLs, `href=""` and `data-href="/a"` alone. **No open redirect**: the replacement always yields `${base}/…`, and `(?!\/)` blocks `//host` re-homing. Two cosmetic gaps (`HREF=`, single quotes) are unreachable — sanitize-html lowercases attribute names and normalizes quoting, and now runs first.
- **`quote()`'s unescaped body** — all six templates traced; only `getStatusBody` output reaches it, and `SanitizedBody` makes a bare string un-passable.
- **`getEmailStatusUrl`** — all four consumers feed it only to `button({url})`, which runs `safeUrl()`. No unchecked href.
- **All 9 call sites** — recipient is correct at every one, including `mentionNotification.ts`, which reads differently (`currentActor`) but is right: it's the inbound-remote path and `currentActor` is the notified local actor.
- **`isLocalActor` undefined** → falsy → `sanitizeText`, the safe branch.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7248 tests passed (725 files)**

### Known inconsistency

The boost footer says "email notifications for **boosts**" and links to a settings page where the toggle reads "**Reblogs**". Deliberate and scoped out of this PR; the other five match their toggles exactly.

llun:
## Review round 2 — verifying the sanitize fix is complete

The round-2 sub-agent hit a session limit mid-run, so I did this verification directly rather than leave the fix unchecked.

**The question that mattered:** sanitizing *after* markdown could strip things the renderer legitimately emits. I rendered ten representative bodies through the new order and inspected the output.

| input | output |
|---|---|
| `a #running tag` | `<a href="https://test.llun.dev/tags/running" class="hashtag" rel="tag">#<span>running</span></a>` |
| `hi @someone@example.com` | `<span class="h-card"><a href="…" target="_blank" class="u-url mention" rel="noopener noreferrer">…` |
| `see https://example.com/x` | `<a href="…" target="_blank" rel="noopener noreferrer">` |
| `**bold**` / `*em*` | `<strong>` / `<em>` |
| list / code / blockquote | `<ul>` / `<code>` / `<blockquote>` |
| `# Big heading` | dropped to text |
| `alt` | dropped |
| `:blobcat:` | left as literal text |

**Nothing legitimate is lost.** Hashtag and mention links keep their `class` and `rel`; the `h-card` wrapper survives; links keep `rel="noopener noreferrer"`. Headings and images drop to text — but *exactly as they do on the web*, because both paths share `sanitizeText`. Consistency with the web UI is preserved because the fix matches `processStatusTextContent`'s first two steps precisely.

**The hashtag output also confirms the ordering question:** it comes out absolute (`https://test.llun.dev/tags/running`), so running `absolutizeLinks` after `sanitizeText` still works — sanitize-html doesn't normalize the href in a way that defeats the regex. Running it second is also strictly safer: only hrefs that survived sanitizing get rewritten.

### `53ff64928` — one real difference, documented

The email path stops where the web pipeline continues: it omits `convertEmojisToImages` and `sanitizeTrustedStatusText`. **Custom emoji therefore stay as literal `:shortcode:` text** rather than becoming `` tags.

I've kept that rather than matching the web, and recorded why: mail clients block remote images by default, so a row of broken-image placeholders reads worse than the shortcode. Flagging it as a conscious call rather than an oversight — if you'd prefer emoji images, it means threading `status.tags` through and emitting more remote images.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7248 tests passed (725 files)** · CI green.

The round-1 fix remains verified non-vacuous: reverting it fails 5 tests across both suites.

PR #1327, opened 2026-07-26
minor: send an email when a fitness activity import completes

llun:
## Summary

Last of three PRs bringing every email in line with the approved [Email templates](https://claude.ai/design/p/a61009c5-7207-4e18-8d2b-a0cace8d672b?file=ui_kits%2Fweb%2Femails.html) design, after #1322 and #1325.

**This one turns on an email that has never sent.** The `activity_import` notification row has been created since the fitness feature shipped, but nothing ever fanned it out — no email *and no push* — despite the settings toggle existing and a template being written and tested. `activityImport.ts` was dead code.

All eleven templates are now on the shared layout.

## Where it sends from, and why that matters

It sends from the **end of `processFitnessFileJob`**, not from where the import is enqueued. That distinction is the substance of this PR:

- Under `NoQueue` — what local dev uses — `importFitnessFilesJob` runs the processing job **inline**, so notifying at enqueue time looks perfectly correct.
- Under **QStash in production** the job is merely *queued*. At that moment `hasMapData` is false, `mapImagePath` is `NULL`, there is no attachment and `processingStatus` is `'pending'`.

Notifying at the enqueue point would therefore have shipped an email with an empty card in production while looking flawless in dev — the worst shape a bug can take. The premature `createNotificationWithPolicy` call has been removed from `importStravaActivityJob` and its test now asserts the notification is *not* created there, with the reason recorded.

## What earns an email

A new `notifyOnComplete` flag on the job payload, set only for an unattended **first** import (`!existingStatus` in `importFitnessFilesJob`, which is the reliable new-status signal).

It is deliberately **not** derived from `publishSendNote`, which is `false` for the Strava path, the user-triggered `/retry-fitness` endpoint **and** the `scripts/fitness/*` backfills alike — reusing it would mass-mail on a backfill. A direct upload also stays silent: the user is watching the composer.

Because the flag lives on the payload, adding Wahoo or any future importer is one line.

## Provider-agnostic, per your note

Nothing in the email names a provider, in the copy or the data. Every value comes from the **status and its fitness file**:

| Slot | Source |
|---|---|
| card title | first line of the status text, truncated; else activity type; else "Fitness activity" |
| timestamp | `fitnessFile.activityStartTime`, else `status.createdAt` |
| Distance / Time / Pace-or-Avg speed | the existing `lib/utils/fitness.ts` formatters |
| footnote | `Imported from activity file · {fileName}` |
| button | `getEmailStatusUrl(status)` → **View status** |

A test asserts the word "strava" appears nowhere in subject, html or text. The third stat follows the activity — `getFitnessPaceOrSpeed` returns **Pace** for a run and **Avg speed** for a ride — rather than being hardcoded as the design sample shows.

## Degradation

Every tier still sends:

1. map + stats — the full designed card
2. no map (indoor, privacy-masked, or map generation failed) → stats only
3. no fitness file at all → headline, lead and button

Tier 3 is the Strava fallback path, for an activity with no exportable streams. It never creates a fitness file so never reaches `processFitnessFileJob`; it keeps its own notification and gets the minimal variant.

## Scope: the JPEG twin is deferred

The approved plan flagged this PR as splittable, and I've taken that option. The Outlook JPEG twin needs a `format` option threaded through **every upload path in the product**, plus a migration and both schema dumps regenerated — that deserves its own focused review rather than riding along here.

Consequence: the map is the existing WebP, which Gmail, Apple Mail and new Outlook render fine. **Outlook desktop and Windows Mail get the `alt` text instead of the map.** Everything else in this PR works everywhere. I'll open the follow-up separately.

## Testing

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7281 tests passed (725 files)**

New coverage includes the two that matter most: `processFitnessFileJob` **sends** with the right recipient, subject and button on a first import, and **stays silent** on a reprocess. Plus 14 template tests covering each degradation tier and the dynamic Pace/Avg-speed label, and the fitness email added to the 11-template conformance suite.

## Verification

Rendered all eleven templates and checked the fitness one in a browser — card, stats row, footnote, **View status** button and the fitness-imports footer all match the design.

The preview deliberately passes **no map URL**. A fixture cannot fabricate a stored media URL, and I tried a stand-in: the generated maps are 4:3, so a square placeholder rendered the card a third taller than it ever will be and made the preview look broken. Showing the genuine no-GPS degradation is more honest; the map slot is covered by `blocks.test.ts`.

> Same standing caveat as the previous two: the Outlook-specific work is invisible in a browser and wants one real send.

## Note for deploy

`shouldSendEmailForNotification` **defaults to true when unset**, so everyone with a background importer starts receiving this email on their first import after deploy.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -259,15 +259,9 @@ section-navigation patterns; pick by section type.
 
 ## Transactional & Notification Emails
 
-Emails go through one shared skeleton, so a design or copy change lands in one
-place instead of eleven.
-
-> **Migration in progress — one template left.** Ten of the eleven are on the
-> shared layout. Only `activityImport.ts` still exports the old
-> `getSubject`/`getTextContent`/`getHTMLContent` trio and writes raw markup —
-> and nothing sends it, so it is dead code until the fitness import wiring
-> lands. Apply the rules below when migrating it, and do not copy the old shape
-> into a new template.
+Every email the server sends goes through one shared skeleton, so a design or
+copy change lands in one place instead of eleven. All eleven templates are on
+it; there is no legacy shape left to copy.
 
 - **One module per email** in `lib/services/email/templates/`, exporting a single
   `build<Name>Email(params): RenderedEmail` (`{ subject, text, html }`). **Never
@@ -305,14 +299,16 @@ place instead of eleven.
   (`deleteActorJob.test.ts` and the password-reset route test), in both cases
   hiding that `sendMail` was never reached at all.
 - **Verify a template change by rendering it**:
-  `./scripts/mock/renderEmailPreviews.ts` writes each template it covers to HTML
-  with fixture data, plus an index showing every one beside its plain-text twin
-  (see `docs/maintenance.md`). Emails are not pages, so this is the real-browser
-  check Definition of Done item 6 asks for. **It currently covers only the four
-  account emails** — add a template to `buildPreviews()` in the same PR that
-  migrates it, or the change ships unpreviewable. Keep the fixture values
-  production-shaped: the codes are 43-char base64url, and a short placeholder
-  hides the link-wrapping problems a real one exposes.
+  `./scripts/mock/renderEmailPreviews.ts` writes every template to HTML with
+  fixture data, plus an index showing each one beside its plain-text twin (see
+  `docs/maintenance.md`). Emails are not pages, so this is the real-browser check
+  Definition of Done item 6 asks for. **A new template must be added to
+  `buildPreviews()` in the same PR**, or the change ships unpreviewable. Keep the
+  fixture values production-shaped — the codes are 43-char base64url, and a short
+  placeholder hides the link-wrapping problems a real one exposes — and leave out
+  a fixture the preview cannot represent honestly: the fitness card passes no map
+  URL because the generated maps are 4:3 and any stand-in image renders the card
+  a third taller than it ever will be.
 - A browser is a lower bar than a mail client. For a change to the shared layout,
   also send one to a real inbox and check Gmail, Apple Mail and Outlook —
   Outlook's Word engine is the one that needs `mso-` properties and the ghost
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -169,14 +169,9 @@ change doesn't touch.
 
 ## Emails
 
-- **Migration in progress — one template left.** Ten of eleven are on the shared
-  layout; only `activityImport.ts` still uses the old
-  `getSubject`/`getTextContent`/`getHTMLContent` trio, and nothing sends it.
-  Apply the rules below when migrating it; don't copy the old shape into a new
-  template.
-- Every migrated email is built by a `build<Name>Email(params): RenderedEmail`
-  module in `lib/services/email/templates/`. No subject/HTML/text literals at a
-  call site.
+- Every email is built by a `build<Name>Email(params): RenderedEmail` module in
+  `lib/services/email/templates/`. No subject/HTML/text literals at a call site.
+  All eleven templates follow this; there is no legacy shape left to copy.
 - Templates compose blocks from `@/lib/services/email/layout/blocks` and render
   through `renderEmail`; they never write markup. Escaping belongs to the block
   builders, so a template hands in plain strings, and nothing in the layout emits
--- a/app/(timeline)/settings/notifications/page.tsx
+++ b/app/(timeline)/settings/notifications/page.tsx
@@ -56,8 +56,8 @@ const notificationTypes: {
   {
     key: 'activity_import',
     label: 'Fitness Activity Imported',
-    description:
-      'A Strava fitness activity has been imported and is ready to view'
+    // Provider-agnostic: this fires for any background import, not just Strava.
+    description: 'A fitness activity has been imported and is ready to view'
   }
 ]
 
--- a/app/api/v1/webhooks/strava/[webhookToken]/route.test.ts
+++ b/app/api/v1/webhooks/strava/[webhookToken]/route.test.ts
@@ -102,7 +102,11 @@ describe('Strava Webhook API', () => {
         data: {
           actorId: 'actor-1',
           stravaActivityId: '987654',
-          visibility: 'unlisted'
+          visibility: 'unlisted',
+          // The webhook is the only entry point that opts into the import
+          // email. Asserted explicitly so the feature cannot be switched off
+          // by dropping this one field with a green build.
+          notifyOnComplete: true
         }
       })
     )
@@ -143,7 +147,8 @@ describe('Strava Webhook API', () => {
         data: {
           actorId: 'actor-1',
           stravaActivityId: '13579',
-          visibility: 'private'
+          visibility: 'private',
+          notifyOnComplete: true
         }
       })
     )
@@ -193,7 +198,8 @@ describe('Strava Webhook API', () => {
         data: {
           actorId: 'actor-1',
           stravaActivityId: '24680',
-          visibility: 'private'
+          visibility: 'private',
+          notifyOnComplete: true
         }
       })
     )
--- a/app/api/v1/webhooks/strava/[webhookToken]/route.ts
+++ b/app/api/v1/webhooks/strava/[webhookToken]/route.ts
@@ -184,7 +184,12 @@ export const POST = traceApiRoute(
         data: {
           actorId: fitnessSettings.actorId,
           stravaActivityId,
-          visibility
+          visibility,
+          // The one entry point that emails: a single activity arriving
+          // unattended, which is the whole reason to tell the user. Retry-all
+          // and the scripts/fitness recovery tools drive the same job in bulk
+          // and deliberately leave this off.
+          notifyOnComplete: true
         }
       })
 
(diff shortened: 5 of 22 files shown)


llun:
## Review round 1 — a mass-mail regression I introduced, fixed

### `219362c3a` — a 500-ride archive import would have sent 500 emails

`notifyOnComplete` was inferred inside `importFitnessFilesJob` as `!existingStatus`. That distinguishes *a new status from a reprocess* — **not** *an unattended single import from a bulk one*, which is what the flag is actually for. My own comment on the field stated the intent correctly ("only an unattended importer sets it at all"); the implementation didn't match it.

`importFitnessFilesJob` is the single funnel for **every bulk import in the repo**. Each activity in a batch is its own overlap group with no pre-existing status, so every one evaluated `true`:

| Path | Scenario |
|---|---|
| `importStravaArchiveJob` | A user uploads their Strava archive ZIP. The job walks the entire activity list → **one email per activity.** |
| `POST /api/v1/fitness/import` | The settings UI is an explicit *multi*-file picker with a progress bar → one email per file, while the user watches the progress UI. |
| `scripts/fitness/importStoredFitnessFile.ts` | Orphan recovery — its whole job is files that never got a status. An operator recovering N crash orphans mails the user N times about old rides. |
| `scripts/fitness/repairFailedFitnessImports.ts` | Same. |
| `retryImports.ts` | Batch retry. |

The reviewer verified it empirically — 25 files in one batch produced 25 publishes, all `notifyOnComplete: true` — and confirmed the archive path produced **zero** notifications on `main`, so this was a pure regression, introduced by me in this PR.

There is no dedup anywhere on the path: `createNotificationWithPolicy` is a plain INSERT with a fresh UUID, and `groupKey` is only read back at query time for display grouping. And `shouldSendEmailForNotification` defaults to **true** when unset. So the emails would all have sent.

**The fix:** the flag now comes from the *originating publisher* and defaults to `false`, so a caller is silent until it opts in — the safe direction to be wrong in. Only `importStravaActivityJob` sets it: one activity, delivered while the user is doing something else. Every bulk path is silent without needing to be touched, including the two recovery scripts that call the job directly.

The regression test drives a three-file batch and asserts every publish stays silent. **Verified non-vacuous:** it fails against the exact logic it replaces.

### `894bb06c6` — title truncation could split an emoji

`slice` cuts on UTF-16 code units. Auto-generated fitness captions lead with an emoji (`buildActivitySummary` emits one per activity type), so a cut landing mid-surrogate leaves a lone half that renders as `�`. Now iterates code points. Also adds the missing tests for the shared group-key helper, and consolidates the duplicated copy in `importStravaActivityJob` onto it.

### Cleared

The reviewer traced and cleared the highest-risk areas:

- **The notify seam cannot fire on a failed import.** It sits inside the try, *after* the file is written `completed`, so a processing failure throws before reaching it. Because the file is already `completed`, a hang there cannot strand it in `processing`. The internal try covers every `await`, and `sendNotificationAlerts` installs its own `.catch` on both the push and email chains.
- **`mapImageUrl` is correct on all three branches** — assigned only after `saveMedia` returned non-null *and* the attachment write succeeded; stays `undefined` if the map block threw or storage returned null, which `statCard` handles.
- **No duplicate emails on single-activity paths.** The overlap merge is safe (`withImportLock` serialises the two webhooks; the second finds the sibling's `statusId`). QStash retries are 0 and `publish` sets a dedup id.
- **The Strava fallback path cannot double up** — it creates no fitness file and returns before reaching `importFitnessFilesJob`.
- **Fixed-UTC timestamps are correct, not a bug** — the fitness detail page renders the same field with `formatUtcDate`, so a viewer-local email would *disagree* with the web UI for the same activity, and no per-user timezone is stored.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7290 tests passed (726 files)**

### Still open

One behaviour the reviewer flagged that I've left as-is: an import that **fails and is then retried** via `/retry-fitness` never emails, because retry correctly passes no flag. That's deliberate — but it does mean the "your activity arrived" promise silently doesn't hold for any import that needed a retry. Worth a follow-up decision rather than a silent gap.

llun:
## Review round 2 — the round-1 fix was one level too low

Round 2 correctly rejected the previous fix as incomplete. Both findings are now addressed.

### `7863f00c7` — the opt-in was still hardcoded, just further down

Round 1 moved `notifyOnComplete` out of `importFitnessFilesJob`, but set it as a **literal inside `importStravaActivityJob`**. So it was never "the webhook opts in" — it was "*anything that runs this job* opts in", and the webhook is one of **five** callers:

| Caller | Consequence |
|---|---|
| `POST /api/v1/fitness/retry-failed` | Loops over **every** retriable batch for the actor. A re-import there *does* create a brand-new status — precisely what "retry failed" targets — so one email per recovered activity. This is a user-facing endpoint, not a script. |
| `scripts/fitness/repairFailedFitnessImports.ts` | Documented use is orphaned files with no post; every one is a new status. One email per repaired activity, with no way for the operator to suppress it. |
| `scripts/fitness/retrigerStravaActivities.ts` | N `--activity-id` args → N emails. |
| `scripts/fitness/runImportStravaActivity.ts` | Manual single run → email. |

The comment I wrote — *"every other caller of this job is a batch … and each leaves this false"* — was simply **wrong**. Those bulk paths run *this* job and inherited the literal. Worse, the no-exportable-streams fallback branch emailed with **no gate at all**, firing on `isNewFallback` alone.

The flag is now a parameter of `importStravaActivityJob`, defaulting to `false`, forwarded to both the import call and the fallback branch, and set **only** by the webhook route.

### The coverage gap that let this through

Round 2 ran a second mutation I hadn't: **deleting the opt-in entirely — switching the whole feature off — and every one of 7290 tests still passed.** The suite proved the flag *propagated* and that the default was silent, but nothing asserted the feature was actually wired on.

The webhook route tests now pin it. Verified both directions:
- delete the opt-in → **3 tests fail**
- restore → 5 pass

### `db9a2e7` — three smaller corrections

- **The provider-agnostic claim was overstated.** The module said no provider is named "in the copy or the data". The copy part holds; the data part doesn't — the footnote prints the real file name, and a Strava import stores `strava-<activityId>.tcx`. That's the user's own file, not wording we chose, so I corrected the claim rather than mangling the filename. The passing test only passed because its fixture used a non-Strava name.
- **The Settings toggle still said "A Strava fitness activity has been imported"** — the one user-visible string describing this now provider-agnostic email.
- **Documented a behaviour change round 1 missed:** an import whose *processing* fails now produces no `activity_import` notification at all, where the old enqueue-time call produced one as soon as the status existed. Intended — "your activity arrived" shouldn't fire for an activity that didn't finish arriving, and the failure is already surfaced in the fitness UI with a retry — but it wasn't written down anywhere.

### Cleared

Round 2 confirmed: no other route to `processFitnessFileJob` sets the flag (composer upload, per-post retry, archive, resume all correctly default to false); the single-activity webhook cannot double-mail (QStash dedups on a deterministic id, `withImportLock` serialises concurrent redelivery, and the overlap merge finds the sibling's status); and `processingStatus='completed'` is written *before* the email, so a crash mid-notify loses the email rather than duplicating it.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7292 tests passed (726 files)**

llun:
## Review round 3 — approved, with one more leak found and closed

Round 3 verified the opt-in chain hop by hop and confirmed **no third email leak**: the flag is forwarded correctly at all seven hops, only the webhook sets it, and the feature genuinely works — the reviewer rendered the real email and confirmed the full card.

It also ran a mutation I hadn't, and found a gap on the adjacent channel.

### `dd9d0dd` — push was never gated on the fallback path

The no-exportable-streams branch gated only `emailContent`, leaving `sendNotificationAlerts` itself to fire on `isNewFallback` alone. Since `main` produced **no push at all** from that branch — it created the notification row and stopped — an ungated fan-out would have been a fresh one-per-activity regression on the *push* channel for any bulk recovery run. Same shape as rounds 1 and 2, one channel over.

Gating the whole block also makes this path consistent with the main one, where the entire notify step sits inside the same check.

**Mutation (d) — removing that gate — previously left all 7292 tests green.** That branch had no coverage whatsoever, which is exactly why two rounds of regressions kept landing near it. It now has a guard that fails the moment the gate goes.

### Mutation results, all four levels

| Mutation | Before | Now |
|---|---|---|
| (a) webhook `notifyOnComplete: true` deleted | caught | caught |
| (b) flag dropped from the `importFitnessFilesJob` call | caught | caught |
| (c) `notifyOnComplete && !existingStatus` → `!existingStatus` | caught | caught |
| (d) fallback gate removed | **NOT caught** | **caught** |

### A test I dropped rather than fake

I tried adding a positive assertion for the fallback path — that it *does* email when opted in — and couldn't make it work honestly: stubbing `getStatus` makes `getOrCreateStravaFallbackNote` believe the note already exists, so `isNewFallback` goes false and the branch never runs. Rather than contort the harness into something that passes without exercising the path, I kept the silent-by-default guard, which is the assertion that actually catches the regression, and noted why in the test. The positive case is covered by the group-key test above it, which opts in.

### Also documented

Round 3 flagged that in-app notification coverage narrowed: a crash-orphan repaired by retry-all or a recovery script now creates the post with no bell entry, where `main` created one for every `isNewImport`. That is intended — a repair run shouldn't spam the bell — and is now stated in the code rather than left implicit.

### Verification

`prettier --check` clean · `eslint` 0 errors · `next build` exit 0 · **7293 tests passed (726 files)**

### Remaining nits, not addressed

Round 3's minor observations, left as-is deliberately: `formatStartTime` hardcodes `en-GB`/UTC — it's the only date formatting in the email system and is consistent with the UTC-day `groupKey`, and there is no per-user timezone stored to use instead. `hasCard` is redundant (`stats.length > 0` implies `fitness`) but harmless and readable. `status as EditableStatus` is an unchecked union narrowing, unreachable in practice since both statuses are created as Notes.

PR #1329, opened 2026-07-26
fix: sanitize uploaded media file names before they reach storage

llun:
## Summary

`lib/services/medias/S3StorageFile.ts` built the temp path for a video upload as `join(tmpdir(), randomHex + file.name)`. `path.join` resolves `..` segments, and the random prefix was concatenated with **no separator**, so it became part of the first path segment. Two distinct failures follow (both verified on this machine, prefix = 16 hex chars):

| `file.name` | resulting path | effect |
| --- | --- | --- |
| `../../../../etc/cron.d/x` | `<tmpdir>/../../etc/cron.d/x` | **escapes `tmpdir()`** — arbitrary file write |
| `clip.mp4/../../../evil.html` | a directory above `tmpdir()` | escapes |
| `../../evil.mp4` | `<tmpdir>/evil.mp4` | does *not* escape, but the `..` cancels the prefix out — a **predictable** path, so one upload can overwrite another's temp file mid-probe |

Three or more `..` are needed to escape, because the first is absorbed by the prefix's own segment. `fs.writeFile(tmpVideoFile, buffer)` then wrote the attacker-supplied bytes at whichever path came out. Browser multipart uploads send a basename so the common path was safe, but any non-browser API client (`POST /api/v1/media`, `POST /api/v2/media`, `POST /api/v1/medias/presigned`) controls the string directly.

There was a second, worse consequence on the **local** driver. The stored filename was `randomHex + extname(file.name)`, and `extname('clip.mp4/../../../evil.html')` is `.html`; `app/api/v1/files/[...pathname]/route.ts` serves `mime.contentType(extname(path))`, so an mp4/HTML polyglot (ffprobe only inspects the bytes) came back as `text/html` on the instance origin — stored XSS. No traversal on that driver, though: its path was always `prefix + ext`, and `extname` can never return a separator.

The same raw name also fed the stored object key via `extname(file.name)` and was persisted verbatim as `medias.originalFileName`.

## What changed

New `lib/services/medias/fileName.ts`, and every supplied name now goes through it at the point it enters storage (both drivers, all three upload paths):

- **`sanitizeStoredFileName`** — reduces a name to one inert path segment: cuts at the last `/` **or** `\` (`path.basename` only knows the running platform's separator, and the value is also handed to object keys / `Content-Disposition`), drops control characters, rejects `.` and `..`, and caps the result at 200 bytes so it fits both `varchar(255)` and a filesystem name. Truncation walks code points so a multi-byte character is never split. Anything that reduces to nothing becomes `file`.
- **`createMediaTempFilePath`** — adds the missing separator between the random prefix and the name, and asserts the result's parent is still `tmpdir()` (defence in depth; a sanitized name can't contain a separator).
- **`getStoredMediaExtension(contentType, fileName)`** — derives a generated path's extension from the already-validated content type instead of the name.

### On `extname()` (the second half of the report)

`extname` on an unsanitised name is steerable too, and the results are worth listing:

| `file.name` | `extname` | effect |
| --- | --- | --- |
| `clip.mp4/../../../evil.html` | `.html` | attacker picks the stored object key's suffix |
| `clip.mp4\0.exe` | `.exe` | plus a NUL that makes `fs` throw `ERR_INVALID_ARG_VALUE` |
| `x.` + 300 chars | 301-char ext | key/filename no filesystem or bucket accepts |
| `clip..` | `.` | key ends in a bare dot |
| `MOVIE.MOV` | `.MOV` | `endsWith('.mov')` is case-sensitive, so uppercase QuickTime was **not** rewritten to `.mp4` |

The extension is cosmetic today — the served `Content-Type` comes from the stored mime type, not the key — but it was fully attacker-chosen and nothing downstream re-checked it. Mapping from the validated content type (`image/jpeg`→`.jpg`, `image/png`→`.png`, `video/mp4`/`video/quicktime`→`.mp4`, `video/webm`→`.webm`, `audio/mp4`→`.m4a`) closes all five rows and makes the suffix describe the bytes actually stored. Unknown content types (which the upload routes already reject) fall back to the sanitized name's extension only when it is a short alphanumeric token.

### Drive-bys in the same functions

- `LocalFileStorage._saveImageBuffer` took a `fileName` parameter it never used — removed, so there is no longer an apparent-but-dead sink for a user-controlled name. Images are re-encoded to WebP under a generated name.
- `_uploadVideoToS3` unlinked its temp file only on the success path; a probe/preview failure leaked it. Now in a `finally`, matching `_uploadImageBufferToS3`.

## Behaviour changes

- `MOVIE.MOV` (content type `video/quicktime`) is now stored as `.mp4`, as `movie.mov` always was.
- `photo.JPEG` is stored as `.jpg`; `photo.PNG` as `.png`. Existing rows are untouched — only new uploads.
- `medias.originalFileName` now stores the sanitized basename rather than the raw string.

## Tests

`lib/services/medias/fileName.test.ts` (77 cases) unit-tests the new module — the sanitizer, the extension mapping, the containment assertion, and that no hostile name can produce a path outside `tmpdir()`, including writing each one to prove the result is a usable single segment (a NUL or an unbounded name would otherwise turn an upload into a 500). One test binds `ACCEPTED_FILE_TYPES` to `EXTENSION_BY_CONTENT_TYPE`, so adding an accepted type without a mapping fails CI instead of silently falling back to the supplied name.

Entry-point tests were added at the three places a name actually arrives. Confirmed **failing against `main`**, i.e. genuine regression tests:

- temp path for a traversing video name stays in `tmpdir()`; the random prefix stays effective for the two-`..` case
- the temp file is removed when probing fails, and when preview extraction fails (the new `finally`)
- object-key extension per content type; `MOVIE.MOV` → `.mp4`
- presigned `fileName` reduced to a basename; stored `originalFileName` sanitized on the video, image, and local paths

Three are **guards, not regressions** — they pass equally before and after, and are kept to hold the behaviour: the happy-path temp unlink, `clip.webm` → `.webm`, and "a traversing name stays inside the local media root" (that driver never had the traversal).

## Verification

`yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (732 files, 7389 tests, all passing).

No UI surface changes, so no screenshots.

## Docs

New convention documented in `AGENTS.md` (**Uploaded file names are untrusted input**), with the matching entries in `REVIEW.md`, `CLAUDE.md`, and `.github/copilot-instructions.md`.

## Note

PR #1321 added `safeAttachmentFilename` in `lib/jobs/replyByEmailJob.ts` for its own inbound-email input and deliberately left the shared media layer alone. That job is not on this branch; when it lands, it can drop its local helper in favour of `sanitizeStoredFileName`.


llun:
## Sub-agent review — round 1

Three sub-agents reviewed the whole diff (security/bypass, project conventions + docs, test quality). Findings below, all addressed in a95491ea5.

### Addressed — correctness

1. **The headline regression test did not prove the traversal.** `'../../evil.mp4'` does not escape `tmpdir()` under the old code: the 16-char random prefix's own segment absorbs the first `..`. `join(tmpdir(), 'deadbeefdeadbeef' + '../../evil.mp4')` → `<tmpdir>/evil.mp4`. So `expect(resolve(dirname(tempPath))).toBe(resolve(tmpdir()))` *passed* against the vulnerable code, and only the naming assertion failed. Now uses `../../../../etc/cron.d/evil.mp4` (verified to escape); the two-`..` case became its own test for what it really is — the random prefix cancelled out into a predictable path. The PR description had the same error and is corrected.

2. **`SAFE_EXTENSION` shape-checked instead of allowlisting.** `/^\.[a-z0-9]{1,10}$/` still admits `.html`, `.svg`, `.js`. The fallback is unreachable today, but nothing bound `ACCEPTED_FILE_TYPES` to `EXTENSION_BY_CONTENT_TYPE` — adding an accepted type without a mapping would compile, pass every test, and either re-open the name-derived `.html` or produce an extension-less local path that `getFile` can never serve (`mime.contentType('')` is `false`). Fallback is now a media-extension allowlist, and a test asserts the map covers `ACCEPTED_FILE_TYPES`.

3. **`extractVideoImage` mocked as resolving `null`.** It only ever resolves a `Buffer` or rejects, so the thumbnail branch was exercised on a state production cannot reach, and the `localFile` `readdir` assertion described a directory shape that never occurs. Mocks now use a real 1×1 PNG; added a test for the realistic failure (preview extraction rejecting), which is what the new `finally` exists for.

4. **The S3 image branch of `saveFile` had no test** — a separate `createMedia` call from the video branch, so the video tests never reached it. Covered.

### Addressed — hardening

5. Only C0/DEL were stripped. C1, zero-width and bidi characters survived; the stored name is federated and rendered to other users as the attachment name, where a `U+202E` makes it display as something other than what is stored. Now stripped.
6. `.trim()` ran before truncation, so cutting at the byte cap could re-expose trailing whitespace. Trimmed again after.
7. The containment check was untestable dead defence (the only uncovered line). Extracted to an exported `assertDirectChildOf` with its own tests; the defence stays.
8. Temp video now written with `wx` (`O_EXCL`) so it cannot follow a planted symlink or clobber an existing file.

### Addressed — docs

9. The new `AGENTS.md`/`REVIEW.md`/`CLAUDE.md` prose asserted a repo-wide invariant the codebase does not hold: `lib/services/fitness-files/` still persists `file.name` raw into `fitness_files.fileName varchar(255) not null`. Its paths are safe (extension from the `getFitnessFileType` allowlist, writes via `assertFitnessStoragePath`), but the name is unbounded and unsanitized. Rules are now scoped to the media drivers and the gap is documented as known. **Follow-up, not this PR.**
10. Docs omitted the fallback branch, called a 300-char extension "a key no filesystem accepts" (true for the local driver; S3 keys allow 1024 bytes), and conflated `sanitizeStoredFileName` with the `extname` behaviour. Corrected.
11. A comment justified cutting at `\` by citing object keys and `Content-Disposition` as sinks. Neither exists today — the object key is `medias/<date>/<random><ext>` and there is no `Content-Disposition` response header anywhere in `lib/` or `app/`. Reworded to the real reason (the value is persisted and returned to clients).

### Considered, not changed

- **Route-level tests** for `/api/v1/media`, `/api/v2/media`, `/api/v1/medias/presigned`: all three route tests mock the storage layer, and the routes forward `File`/`PresigedMediaInput` untouched. A route assertion would only observe the raw name arriving at a mock. The driver is the correct boundary.
- **A repo-wide guard test** (in the style of `tailwindCssVariableSyntax.test.ts`) asserting no `extname(file.name)`/`join(tmpdir(), …)` anywhere: attractive, but any grep-based rule would be either trivially evadable or noisy against the fitness-files exception. Revisit if fitness-files is brought in.

Gate re-run after the fixes: prettier → lint (0 errors) → build → test, 732 files / 7415 tests passing.

llun:
## Sub-agent review — round 2

Re-reviewed the round-1 fixes for new problems. **One finding, fixed in 17c71bfb0.**

### Addressed

**The character-stripping predicate over-stripped orthographic joiners.** The range `0x200b–0x200f` swallowed U+200C ZWNJ and U+200D ZWJ along with the bidi marks. Those two are joiners, not controls, and neither can reorder text — so they don't belong in a set aimed at Trojan Source spoofing. Because the stored name federates as the attachment's `name` and renders as alt text (`lib/services/statuses/mediaIds.ts:44`, `lib/components/posts/media.tsx:31`), the damage was user-visible:

```
می‌خواهم.mp4   -> میخواهم.mp4    a misspelling
क्‍ष.png       -> क्ष.png         collapses two distinct conjunct forms
क्‌ष.png       -> क्ष.png         ...into the same name
👨‍👩‍👧‍👦 trip.mp4  -> 👨👩👧👦 trip.mp4
🏳️‍🌈.png       -> 🏳️🌈.png
```

The comment also claimed to cover zero-width characters generally while U+2060 WORD JOINER, U+2061–2064, U+180E and U+061C ALM all survived. The set is now explicitly two groups — bidi controls (ALM, LRM/RLM, LS/PS, embeddings, overrides, isolates) and invisible spacing with no orthographic meaning — with the joiners documented as a deliberate exclusion. Tests cover the newly stripped characters, both joiners surviving, and a Thai name passing through untouched.

### Verified clean (no action)

- **`EXTENSION_BY_CONTENT_TYPE` coverage test is genuinely enforced.** Proven by adding `image/webp` to a scratchpad copy of `ACCEPTED_FILE_TYPES`: 1 failed / 77 passed. `FALLBACK_EXTENSIONS` derives from the map at module load, so it tracks automatically.
- **`assertDirectChildOf`** holds for win32, UNC, extended-length, trailing-separator and relative directories, and for symlinked tmpdirs (neither operand goes through `realpath`, so a symlinked `/tmp` cannot trip it). A 20,000-iteration fuzz never escaped.
- **`flag: 'wx'`** does what it claims: an existing symlink at the path now yields `EEXIST` with the target untouched, where the previous code wrote straight through. Collision → `EEXIST` → 500, at ~1 in 1.8e19; a 500 is the right answer in the symlink case anyway.
- **The double `.trim()`** cannot produce `''`, `'.'`, `'..'`, a separator, an untrimmed value or >200 bytes — 220,000 fuzzed inputs. The function is also idempotent, which matters because the presigned path sanitizes and then `getStoredMediaExtension` sanitizes again.
- Character ranges have no off-by-one: U+202F NNBSP and U+FE0F VS16 both survive.

### Noted, deliberately out of scope

- **No backfill.** Rows written before this PR keep their unsanitized names and still federate them. A separate concern, and per AGENTS.md operational SQL doesn't belong in a PR description.
- **`audio/mp4` can never actually be stored** — both drivers' `saveFile` return `null` for anything not `image/*` or `video/*` (→ 422), while `/api/v1/instance` advertises it in `supported_mime_types`. Entirely pre-existing; the coverage test just makes that map entry look load-bearing. Not touching it here.

Gate re-run: prettier → lint (0 errors) → build → test, 732 files / 7420 tests passing.

llun:
## Sub-agent review — round 3: CLEAN

No new actionable findings. Per AGENTS.md → Code Review Loop, that ends the loop at 3 rounds of 20.

### Verification of the round-2 fix

- **Bidi set is complete.** All 12 characters with `Bidi_Control=Yes` are stripped (U+061C, U+200E, U+200F, U+202A–202E, U+2066–2069), plus U+2028/U+2029. All 14 Trojan-Source reordering primitives reduce `photo<X>gnp.exe` → `photognp.exe`.
- **Exhaustive BMP sweep** of every code point with GC `Cc`/`Cf`/`Zl`/`Zp` or `Default_Ignorable_Code_Point`: 87 stripped, 60 survive. Every survivor is `Bidi_Class` BN or L — nothing with reordering capability gets through. The predicate strips zero non-invisible characters across all 65,536 BMP code points, and zero astral code points, so emoji ZWJ sequences, tag-sequence flags, skin-tone modifiers and variation selectors are untouched.
- **35 boundaries audited**, all correct — notably U+202E stripped vs U+202F NNBSP kept (a real space in French typography), U+2064 vs U+2065/2066, U+FEFF vs U+FE0F VS16, and U+200B stripped vs U+200C/U+200D kept.
- 28 real-world names (Persian with ZWNJ, Arabic, Hebrew, Thai, Devanagari with ZWJ, Japanese, Korean, Greek, Cyrillic, Vietnamese, keycaps, 👨‍👩‍👧‍👦) go in and come out byte-identical.
- Exhaustive per-code-point injection into 9 hostile shapes plus a 200,000-input fuzz: no separator, no `.`/`..` segment, no untrimmed or >200-byte output, and `createMediaTempFilePath` never escaped `tmpdir()`.

### Independently re-confirmed

- `{ flag: 'wx' }` is awaited **outside** the `Promise.all`, so a failed write never reaches the `.finally()` unlink.
- `MediaValidationError` from `assertDirectChildOf` maps to **422**, not 500.
- The presigned S3 key and `originalFileName` are fully decoupled; nothing re-derives a key from the stored name.
- Every factual claim in the AGENTS.md / REVIEW.md / CLAUDE.md additions verified against the implementation, including the `..`-count behaviour (2 → predictable path, 3+ → escape) reproduced on two tmpdir shapes.

### Judgment recorded, not changed

`U+180E MONGOLIAN VOWEL SEPARATOR` is stripped as "invisible spacing", but in traditional Mongolian it does affect shaping — structurally the same argument that kept ZWNJ/ZWJ. It has been `Cf` + `Default_Ignorable` since Unicode 6.3 and appears on essentially every invisible-character blocklist, and traditional Mongolian in a filename is vanishingly rare. Keeping it stripped; noting the trade-off here so the choice is on the record rather than accidental.

Gate on the final commit: prettier → lint (0 errors) → build → test, 732 files / 7420 tests passing.

PR #1330, opened 2026-07-26
minor: add the status reaction picker and chips

llun:
Epic 5.1, PR 3 of 3. [#1320](https://github.com/llun/activities.next/pull/1320) stored reactions and accepted them inbound; [#1324](https://github.com/llun/activities.next/pull/1324) made them writable and federated them out. This puts them on screen.

## Reactions had to reach the domain Status

5.1a added rollups to the **Mastodon** `Status` entity, but the shared `Post` components render a **domain** `Status` — which had no reactions. Fetching per post from the client would have been one request per rendered post, so instead the rollups now hydrate onto the domain status through the existing batch seam next to the like/bookmark state: **one extra grouped query per timeline page**, with a single-status fallback for the detail routes (the same `has`-vs-`get` contract `statusMetricsCache` uses).

`reactions` is `.optional()` rather than `.default([])` on purpose — a zod default is *required* on the output type, which would have forced every existing status literal and fixture in the repo to declare it. Freshly-created statuses do declare `reactions: []` so the created shape matches the hydrated one.

## The UI

- **`ReactionRow`** sits between the post content and the action row, rendered by `Post` — so reactions look and behave identically on every surface, per the **Status Posts & Actions** rule, with no per-page wiring. Optimistic toggle that reverts on failure and reconciles against the server's authoritative rollups; `aria-pressed` chips; a custom emoji renders as its image and a unicode one (or a remote emoji whose image we could not vouch for) as text; a `+N` overflow once a post collects more chips than a row can usefully carry.
- **`ReactionPicker`** reuses the post-box picker's data layer (`EMOJI_GROUPS`, `searchSystemEmojis`, the instance custom emoji) rather than duplicating an emoji table, and returns a reaction *name* instead of caret text. It loads the custom emoji itself, memoized once per page load — threading them as a prop would touch every surface that renders a post, which is the prop-drilling `AGENTS.md` warns against for instance-wide values.
- A logged-out reader sees existing chips read-only with no picker, and a post with no reactions renders nothing at all for them.
- All calls go through new `lib/client.ts` helpers (`reactToStatus`, `unreactFromStatus`); no component calls `fetch` directly.

## Browser verification

Ran against a throwaway local SQLite database on port 3044 with the sanctioned mock user. Note: `.env.local` is unreadable in this environment, so rather than trust it I passed the local SQLite config **inline**, which wins over `dotenv-flow` — the resolved target was provably `./dev-reactions.sqlite3`.

Verified end to end, signed in as `testuser@localhost:3044`:

1. Reaction rows render on every post (7 add buttons on the timeline); `GET /api/v1/custom_emojis` fires once and returns 200.
2. The picker opens with `aria-expanded="true"` and 30 emoji — screenshot below.
3. Picking one issues `PUT /api/v1/pleroma/statuses/apurl_…/reactions/%F0%9F%98%83` → **200**, and the chip renders as `😃 1` with `aria-pressed="true"` in the "mine" style.
4. The database holds exactly one row: the correct status URL, actor, `name: 😃`, `url: NULL` (unicode).
5. **The chip survives a full page reload** with `me: true` — which is the real proof the new server-side domain hydration works.
6. Toggling the chip off deletes the row (`select count(*)` → 0).
7. A self-reaction creates **no** notification, as designed.

One thing worth recording for anyone repeating this: `scripts/mock/createMockStatuses.ts` seeds statuses with **bare UUID ids** rather than ActivityPub URLs, so `urlToId` cannot round-trip them and any status-addressed API 404s on them — including the pre-existing favourite route. Verification therefore used a post created through the app, which gets a proper `apurl_` id. Not a defect in this PR, but it will bite the next person who tests a status-addressed endpoint against mock data.

I could not attach the screenshots to this description (the Browser pane returned blank captures on the image-heavy timeline, and there is no upload path from the CLI); the picker and chip screenshots were captured locally and the functional evidence above is the substantive verification.

## Verification

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **732 files / 7333 tests green**, including 9 new jsdom tests for the row.

Two existing expectations were updated rather than worked around: the mixin-wiring test now asserts the new `statusReactionDatabase` argument, and `createNote`'s timeline assertion passes because freshly-created statuses now declare `reactions: []`.

### DIFF
--- a/app/(timeline)/[actor]/ActorTimelines.tsx
+++ b/app/(timeline)/[actor]/ActorTimelines.tsx
@@ -22,6 +22,7 @@ import {
   StatusPoll,
   StatusType
 } from '@/lib/types/domain/status'
+import { StatusReaction } from '@/lib/types/mastodon/statusReaction'
 
 import { ActorMediaGallery } from './ActorMediaGallery'
 
@@ -230,6 +231,22 @@ export const ActorTimelines: FC<Props> = ({
     []
   )
 
+  // Reactions have to be synced alongside likes and bookmarks: those handlers
+  // replace the status object in this page's own copy, which re-renders the
+  // reaction row from its (now stale) `reactions` prop and would otherwise
+  // revert a chip the viewer just added.
+  const handleReactionsChanged = useCallback(
+    (status: StatusNote | StatusPoll, reactions: StatusReaction[]) => {
+      setCurrentStatuses((previousStatuses) =>
+        updateMatchingStatus(previousStatuses, status.id, (target) => ({
+          ...target,
+          reactions
+        }))
+      )
+    },
+    []
+  )
+
   const loadMoreStatuses = useCallback(async () => {
     const nextPageUrl = currentStatusPagination.nextPageUrl
     if (isLoadingRef.current || !nextPageUrl) return
@@ -331,6 +348,7 @@ export const ActorTimelines: FC<Props> = ({
         onPostDeleted={handlePostDeleted}
         onLikeChanged={handleLikeChanged}
         onBookmarkChanged={handleBookmarkChanged}
+        onReactionsChanged={handleReactionsChanged}
       />
     ) : (
       <EmptyState>{emptyMessage}</EmptyState>
--- a/app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
+++ b/app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
@@ -46,6 +46,7 @@ import { RepostButton } from '@/lib/components/posts/actions/repost-button'
 import { ActorAvatar } from '@/lib/components/posts/actor'
 import { Media } from '@/lib/components/posts/media'
 import { Post } from '@/lib/components/posts/post'
+import { ReactionRow } from '@/lib/components/posts/reaction-row'
 import { StatusReplyBox } from '@/lib/components/posts/status-reply-box'
 import { Button } from '@/lib/components/ui/button'
 import {
@@ -1958,6 +1959,20 @@ export const FitnessStatusDetail: FC<Props> = ({
             )}
           </div>
         </div>
+
+        {/* Reactions belong to the post, so they sit above the action bar here as
+            they do in `Post`. This page composes its own action row instead of
+            using `Posts`, so the row has to be added explicitly — without it a
+            fitness post is the one surface where an existing reaction is
+            invisible and no new one can be added. */}
+        {(currentActor || (status.reactions?.length ?? 0) > 0) && (
+          <div className="border-t px-4 pt-2.5">
+            <ReactionRow
+              currentActor={currentActor ?? undefined}
+              status={status}
+            />
+          </div>
+        )}
 
         <div className="flex flex-wrap items-center justify-between gap-x-4 gap-y-2 border-t px-4 py-2.5">
           {sourceHref ? (
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -204,6 +204,9 @@ const Page: FC<Props> = async ({ params }) => {
     replies = await database.getStatusReplies({
       statusId,
       url: statusUrl,
+      // Separate from the visibility filter below: this is the viewer whose
+      // like, bookmark and reaction state each reply is hydrated with.
+      currentActorId: currentActor?.id,
       ...(currentActor
         ? { visibleToActorId: currentActor.id }
         : { publicOnly: true })
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -285,7 +285,10 @@ const Page: FC<Props> = async ({ params }) => {
           // which already supplies its own top padding, so only the signed-in
           // surface needs this gap.
           currentActorProfile && 'mt-4',
-          'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
+          // No `overflow-hidden`: this card contains posts, and a post's
+          // non-portaled overlays (the reaction picker) would be clipped by it —
+          // the same reason `Posts` dropped it.
+          'rounded-2xl border bg-background/80 shadow-sm'
         )}
       >
         {currentActorProfile ? (
@@ -332,7 +335,10 @@ const Page: FC<Props> = async ({ params }) => {
         // logged-out `PublicShell` already provides its own, so scope the gap
         // to the signed-in card to keep the spacing consistent across both.
         currentActorProfile && 'mt-4',
-        'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
+        // No `overflow-hidden`: this card contains posts, and a post's
+        // non-portaled overlays (the reaction picker) would be clipped by it —
+        // the same reason `Posts` dropped it.
+        'rounded-2xl border bg-background/80 shadow-sm'
       )}
     >
       {currentActorProfile ? (
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -285,10 +285,7 @@ const Page: FC<Props> = async ({ params }) => {
           // which already supplies its own top padding, so only the signed-in
           // surface needs this gap.
           currentActorProfile && 'mt-4',
-          // No `overflow-hidden`: this card contains posts, and a post's
-          // non-portaled overlays (the reaction picker) would be clipped by it —
-          // the same reason `Posts` dropped it.
-          'rounded-2xl border bg-background/80 shadow-sm'
+          'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
         )}
       >
         {currentActorProfile ? (
@@ -338,7 +335,7 @@ const Page: FC<Props> = async ({ params }) => {
         // No `overflow-hidden`: this card contains posts, and a post's
         // non-portaled overlays (the reaction picker) would be clipped by it —
         // the same reason `Posts` dropped it.
-        'rounded-2xl border bg-background/80 shadow-sm'
+        'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
       )}
     >
       {currentActorProfile ? (
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -230,7 +230,11 @@ const Page: FC<Props> = async ({ params }) => {
   if (status.type !== StatusType.enum.Announce && status.reply) {
     let replyStatus = await database.getStatus({
       statusId: status.reply,
-      withReplies: false
+      withReplies: false,
+      // Ancestors render the same interactive chips as the focused status, so
+      // they need the same viewer — otherwise one post shows its reaction
+      // highlighted on the timeline and unhighlighted as a thread ancestor.
+      currentActorId: currentActor?.id
     })
     while (previouses.length < 3 && replyStatus) {
       // `getStatus` does no visibility filtering, so without this guard a public
@@ -257,7 +261,8 @@ const Page: FC<Props> = async ({ params }) => {
       }
       replyStatus = await database.getStatus({
         statusId: replyStatus.reply,
-        withReplies: false
+        withReplies: false,
+        currentActorId: currentActor?.id
       })
     }
   }
--- a/app/(timeline)/[actor]/[status]/page.tsx
+++ b/app/(timeline)/[actor]/[status]/page.tsx
@@ -79,7 +79,8 @@ const Page: FC<Props> = async ({ params }) => {
   const resolvedStatus = await resolveStatusFromPath({
     database,
     actorParam: actor,
-    statusParam
+    statusParam,
+    currentActorId: currentActor?.id
   })
   if (!resolvedStatus) return notFound()
 
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
@@ -78,12 +78,14 @@ export const resolveStatusFromPath = async ({
   if (isStatusHash) {
     status = await database.getStatusFromUrlHash({
       urlHash: decodedStatusParam,
-      actorId: actorIdFromPath
+      actorId: actorIdFromPath,
+      currentActorId
     })
 
     if (!status && actorIdFromPath) {
       const unscopedStatus = await database.getStatusFromUrlHash({
-        urlHash: decodedStatusParam
+        urlHash: decodedStatusParam,
+        currentActorId
       })
 
       if (unscopedStatus) {
--- a/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
+++ b/app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
@@ -8,6 +8,9 @@ interface ResolveStatusFromPathParams {
   >
   actorParam: string
   statusParam: string
+  // The signed-in viewer, so the focused status is hydrated with their own
+  // like/bookmark/reaction state. Omitted for anonymous readers.
+  currentActorId?: string
 }
 
 interface ResolveStatusFromPathResult {
@@ -43,7 +46,8 @@ const getStatusForPathActor = (status: Status, actorId: string) => {
 export const resolveStatusFromPath = async ({
   database,
   actorParam,
-  statusParam
+  statusParam,
+  currentActorId
 }: ResolveStatusFromPathParams): Promise<ResolveStatusFromPathResult | null> => {
   const decodedActor = decodePathParam(actorParam)
   const decodedStatusParam = decodePathParam(statusParam)
@@ -91,13 +95,15 @@ export const resolveStatusFromPath = async ({
   if (!status && !isStatusHash) {
     status = await database.getStatus({
       statusId: fullStatusId,
-      withReplies: false
+      withReplies: false,
+      currentActorId
     })
   }
 
   if (!status && !isStatusHash && !isFullStatusUrl) {
     status = await database.getStatus({
       statusId: decodedStatusParam,
+      currentActorId,
       withReplies: false
     })
   }
--- a/app/api/v1/tags/[tag]/route.ts
+++ b/app/api/v1/tags/[tag]/route.ts
@@ -108,6 +108,11 @@ export const GET = traceApiRoute(
         fetchBatch: ({ maxStatusId, limit }) =>
           database.getStatusesByHashtag({
             hashtag: tag,
+            // This is the load-more endpoint behind the web hashtag page, so it
+            // must hydrate for the same viewer the server-rendered first page
+            // did — otherwise the viewer's own reaction, like, bookmark and
+            // boost state flip to false halfway down one scroll.
+            currentActorId: currentActor?.id,
             limit,
             maxStatusId: maxStatusId ?? undefined
           })
--- a/docs/features.md
+++ b/docs/features.md
@@ -46,7 +46,7 @@ This document tracks the implemented and planned features for Activity.next.
 - ✅ **Push notifications** — Web Push subscriptions with VAPID configuration
 - ✅ **Unread count** — Badge showing unread notification count
 
-- ✅ **Status emoji reactions** — React to any post with a unicode emoji or an instance custom emoji, separately from favouriting it. Chips under each post show who reacted with what, and a picker adds your own. Reactions federate both ways: inbound from Pleroma/Akkoma (`EmojiReact`) and the Misskey family (a `Like` carrying `content`/`_misskey_reaction`), outbound as a Misskey-style `Like` that vanilla Mastodon shows as a favourite. Served on every status as `pleroma.emoji_reactions` and `reactions`, with `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` (plus glitch-soc `react`/`unreact` aliases)
+- ✅ **Status emoji reactions** — React to any post with a unicode emoji or an instance custom emoji, separately from favouriting it. Chips under each post show each emoji with its count (yours highlighted), and a picker adds your own; the accounts behind a reaction are available from the API rather than the chips. Reactions federate both ways: inbound from Pleroma/Akkoma (`EmojiReact`) and the Misskey family (a `Like` carrying `content`/`_misskey_reaction`), outbound as a Misskey-style `Like` that vanilla Mastodon shows as a favourite. Served on every status as `pleroma.emoji_reactions` and `reactions`, with `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` (plus glitch-soc `react`/`unreact` aliases)
 
 ### Storage & Media
 
--- a/docs/features.md
+++ b/docs/features.md
@@ -46,7 +46,7 @@ This document tracks the implemented and planned features for Activity.next.
 - ✅ **Push notifications** — Web Push subscriptions with VAPID configuration
 - ✅ **Unread count** — Badge showing unread notification count
 
-- 🚧 **Status emoji reactions** — React to a post with a unicode emoji or an instance custom emoji, separately from favouriting it. Reactions federate both ways: inbound from Pleroma/Akkoma (`EmojiReact`) and the Misskey family (a `Like` carrying `content`/`_misskey_reaction`), outbound as a Misskey-style `Like` that vanilla Mastodon shows as a favourite. Served on every status as `pleroma.emoji_reactions` and `reactions`, with `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` (plus glitch-soc `react`/`unreact` aliases) to write them. The picker UI is still landing
+- ✅ **Status emoji reactions** — React to any post with a unicode emoji or an instance custom emoji, separately from favouriting it. Chips under each post show who reacted with what, and a picker adds your own. Reactions federate both ways: inbound from Pleroma/Akkoma (`EmojiReact`) and the Misskey family (a `Like` carrying `content`/`_misskey_reaction`), outbound as a Misskey-style `Like` that vanilla Mastodon shows as a favourite. Served on every status as `pleroma.emoji_reactions` and `reactions`, with `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` (plus glitch-soc `react`/`unreact` aliases)
 
 ### Storage & Media
 
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -468,8 +468,39 @@ export const likeStatus = async ({ statusId }: DefaultStatusParams) => {
 }
 
 /**
+ * The outcome of a reaction write. A 4xx that the user cannot retry away (the
+ * per-actor cap, an emoji this instance rejects) carries the server's own
+ * message so the row can say what actually went wrong instead of inviting a
+ * retry that will always fail the same way.
+ */
+export type ReactionUpdateResult =
+  | { ok: true; reactions: MastodonStatusReaction[] }
+  | { ok: false; error?: string }
+
+const toReactionUpdateResult = async (
+  response: Response
+): Promise<ReactionUpdateResult> => {
+  if (!response.ok) {
+    // 4xx is a verdict on the request itself; 5xx and network faults are
+    // transient, and get the caller's generic retry copy.
+    if (response.status >= 400 && response.status < 500) {
+      const body = await response.json().catch(() => null)
+      const error = (body as { error?: unknown } | null)?.error
+      if (typeof error === 'string' && error.length > 0)
+        return { ok: false, error }
+    }
+    return { ok: false }
+  }
+  const status = (await response.json()) as MastodonStatus
+  return {
+    ok: true,
+    reactions: status.pleroma?.emoji_reactions ?? status.reactions ?? []
+  }
+}
+
+/**
  * Adds the current actor's emoji reaction to a status and returns the updated
- * reaction rollups, or null on failure so the caller can revert its optimistic
+ * reaction rollups, or a failure the caller can use to revert its optimistic
  * chip. `name` is a unicode emoji or a local custom-emoji shortcode.
  *
  * Uses the Pleroma/Akkoma dialect, which is the primary reaction surface (the
@@ -480,9 +511,7 @@ export const likeStatus = async ({ statusId }: DefaultStatusParams) => {
 export const reactToStatus = async ({
   statusId,
   name
-}: DefaultStatusParams & { name: string }): Promise<
-  MastodonStatusReaction[] | null
-> => {
+}: DefaultStatusParams & { name: string }): Promise<ReactionUpdateResult> => {
   const response = await fetch(
     `/api/v1/pleroma/statuses/${urlToId(statusId)}/reactions/${encodeURIComponent(name)}`,
     {
@@ -492,21 +521,17 @@ export const reactToStatus = async ({
       }
     }
   )
-  if (!response.ok) return null
-  const status = (await response.json()) as MastodonStatus
-  return status.pleroma?.emoji_reactions ?? status.reactions ?? []
+  return toReactionUpdateResult(response)
 }
 
 /**
  * Removes the current actor's emoji reaction from a status and returns the
- * updated rollups, or null on failure.
+ * updated rollups, or a failure the caller can use to revert.
  */
 export const unreactFromStatus = async ({
   statusId,
   name
-}: DefaultStatusParams & { name: string }): Promise<
-  MastodonStatusReaction[] | null
-> => {
+}: DefaultStatusParams & { name: string }): Promise<ReactionUpdateResult> => {
   const response = await fetch(
     `/api/v1/pleroma/statuses/${urlToId(statusId)}/reactions/${encodeURIComponent(name)}`,
     {
@@ -516,9 +541,7 @@ export const unreactFromStatus = async ({
       }
     }
   )
-  if (!response.ok) return null
-  const status = (await response.json()) as MastodonStatus
-  return status.pleroma?.emoji_reactions ?? status.reactions ?? []
+  return toReactionUpdateResult(response)
 }
 
 export const bookmarkStatus = async ({ statusId }: DefaultStatusParams) => {
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -27,6 +27,7 @@ import type { ListEntity } from '@/lib/types/mastodon/list'
 import type { MediaAttachment } from '@/lib/types/mastodon/mediaAttachment'
 import type { PreviewCard } from '@/lib/types/mastodon/previewCard'
 import type { Status as MastodonStatus } from '@/lib/types/mastodon/status'
+import type { StatusReaction as MastodonStatusReaction } from '@/lib/types/mastodon/statusReaction'
 import type { Tag } from '@/lib/types/mastodon/tag'
 import type { Translation } from '@/lib/types/mastodon/translation'
 import { normalizeActorId } from '@/lib/utils/activitypub'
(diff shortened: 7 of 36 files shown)


lib/database/sql/status.ts
```diff
@@ -1201,6 +1215,35 @@ export const StatusSQLDatabaseMixin = (
     })
   }
 
+  // Batched emoji-reaction rollups for a page of statuses, in the shape the
+  // domain Status carries. Every hydration site uses this so none of them falls
+  // back to the per-status lookup — that fallback exists for single-status
+  // routes, not for list paths.
+  async function buildReactionRollupContext(
```
llun, line 1226: **N+1 (confirmed, fixed).** Only `getStatusesByIds` got the batched rollup prefetch — `getActorStatuses` and `getStatusReplies` built their hydration context without `reactionRollups`, so every row fell through to the per-status fallback: one grouped `status_reactions` query per status (plus a `customEmojis` query per local-shortcode row) on profiles, threads, the unauthenticated AP outbox and the Mastodon list routes. Worse on the outbox, where `toActivityPubObject` discards `reactions` entirely, and on `/api/v1/accounts/:id/statuses`, where `getMastodonStatuses` re-batches them anyway.

All three sites now share `buildReactionRollupContext`, so a fourth batch site can't drift again — which is exactly how this one got missed.

lib/database/sql/status.ts
```diff
@@ -1237,19 +1280,24 @@ export const StatusSQLDatabaseMixin = (
...
       hydrationStatusIds.size > 0
         ? statusDetectedLanguageDatabase.getDetectedLanguages({
             statusIds: [...hydrationStatusIds]
           })
         : Promise.resolve({}),
       hydrationStatusIds.size > 0
         ? getStatusQuoteEdges([...hydrationStatusIds])
-        : Promise.resolve(new Map<string, StatusQuote>())
+        : Promise.resolve(new Map<string, StatusQuote>()),
+      buildReactionRollupContext(
```
llun: **`me` flag on list paths (fixed).** The rollup context now carries `visibleToActorId`, so a viewer's own reaction reads `me: true` in lists instead of always `false`.

A reviewer raised this separately for `getActorStatuses` and a verifier refuted it as pre-existing systemic behaviour (likes and bookmarks are also unhydrated on that path). That's accurate, and I'm not widening the function's contract to chase it — but the viewer id is already a parameter, so passing it into the batch costs nothing and strictly improves the flag.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
```diff
@@ -43,7 +46,8 @@ const getStatusForPathActor = (status: Status, actorId: string) => {
 export const resolveStatusFromPath = async ({
   database,
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:47]: **Detail page `me` flag (confirmed, fixed).** The resolver called `getStatus` with no `currentActorId`, so on the one surface where chips are interactive your own reaction always rendered as "Add" — click it and the server returns `changed: false` and the correct rollups, so the UI silently self-corrected, which is how this slipped past manual testing.

The resolver now takes an optional `currentActorId` and the page passes the already-resolved `currentActor?.id`.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,232 @@
...
+      setPendingName(null)
+    }
+  }
+
+  // Nothing to show and nothing to do: a logged-out reader on an unreacted post
+  // gets no empty row.
+  if (reactions.length === 0 && !currentActor) return null
+
+  const visible = reactions.slice(0, MAX_VISIBLE_CHIPS)
+  const hiddenCount = reactions.length - visible.length
```
llun, line 227: **Read-only chip accessibility (refuted, fixed anyway).** A verifier argued the `disabled` button is deliberate and matches convention. The intent is deliberate; the mechanism was wrong. A disabled button drops out of the tab order and renders at 50% opacity with a not-allowed cursor — so on every logged-out surface the reaction counts became unreadable to keyboard and screen-reader users, and visually degraded for everyone. "Not actionable" and "not readable" aren't the same requirement.

A reader who can't react now gets a plain `<span>` at full contrast with an `aria-label` carrying the emoji and count. Signed-in chips are unchanged.

lib/services/reactions/reactionRouteHandlers.ts
llun, line 1: **Permanent 422 shown as retryable (refuted, fixed anyway).** The verifier's ground for refusal was that picker-side cap awareness is documented as deferred (`reactionLimits.ts` says "(later) the picker UI"). Fair — but that defers *pre-emptive* cap display, not the obligation to report a refusal honestly. Hitting the 8-reaction cap showed "Failed to add reaction. Please try again.", and retrying fails identically forever.

The two 422 reasons now carry real messages. Surfacing the previous generic `{ error: 'Unprocessable entity' }` verbatim would have been worse than the retry copy, which is why the fix is here rather than only in the client. A descriptive `error` string is also more Mastodon-dialect, not less.

lib/client.ts
```diff
@@ -466,6 +467,83 @@ export const likeStatus = async ({ statusId }: DefaultStatusParams) => {
...
+ * The outcome of a reaction write. A 4xx that the user cannot retry away (the
+ * per-actor cap, an emoji this instance rejects) carries the server's own
+ * message so the row can say what actually went wrong instead of inviting a
+ * retry that will always fail the same way.
+ */
+export type ReactionUpdateResult =
+  | { ok: true; reactions: MastodonStatusReaction[] }
+  | { ok: false; error?: string }
+
+const toReactionUpdateResult = async (
```
llun, line 480: **Client error contract (fixed).** `reactToStatus`/`unreactFromStatus` collapsed every non-ok response to `null`, so the row couldn't tell a transient 500 from a permanent 422. They now return a `ReactionUpdateResult` discriminated union; only a 4xx carrying a message surfaces it, and 5xx/network faults keep the generic retry copy.

lib/components/posts/posts.tsx
```diff
@@ -51,6 +52,10 @@ interface Props {
     isBookmarked: boolean
   ) => void
   onLikeChanged?: (status: StatusNote | StatusPoll, isLiked: boolean) => void
```
llun, line 54: **`onReactionsChanged` reachability (refuted, threaded anyway).** The verifiers correctly caught that the "unreachable" claim was wrong — `Post` is exported and `StatusBox`/`FitnessStatusDetail` render it directly, so the prop was already available.

I still threaded it through `Posts`. AGENTS.md documents the data-sync callbacks as a set (`onStatusCreated`, `onPostUpdated`, `onPostDeleted`, `onLikeChanged`, `onBookmarkChanged`), and a reaction callback that works on two detail surfaces but not on any list is the kind of per-surface asymmetry that contract exists to prevent. Four lines, mirroring `onLikeChanged`; AGENTS.md updated to list it.

lib/database/sql/status.test.ts
```diff
@@ -352,11 +352,78 @@ describe('StatusDatabase', () => {
           isLocalActor: true,
           totalLikes: 0,
           totalShares: 0,
+          reactions: [],
           attachments: [],
           tags: []
         })
       })
 
+      describe('emoji reaction hydration', () => {
```
llun, line 361: **Hydration coverage (refuted as "missing test, not a defect", added anyway).** True as stated — but the second test here fails against the pre-fix code, because `me: true` on `getActorStatuses` only holds via the batched path. It's a regression guard for the confirmed finding above, not speculative coverage.

docs/features.md
```diff
@@ -46,7 +46,7 @@ This document tracks the implemented and planned features for Activity.next.
...
+- ✅ **Status emoji reactions** — React to any post with a unicode emoji or an instance custom emoji, separately from favouriting it. Chips under each post show each emoji with its count (yours highlighted), and a picker adds your own; the accounts behind a reaction are available from the API rather than the chips. Reactions federate both ways: inbound from Pleroma/Akkoma (`EmojiReact`) and the Misskey family (a `Like` carrying `content`/`_misskey_reaction`), outbound as a Misskey-style `Like` that vanilla Mastodon shows as a favourite. Served on every status as `pleroma.emoji_reactions` and `reactions`, with `PUT`/`DELETE /api/v1/pleroma/statuses/:id/reactions/:emoji` (plus glitch-soc `react`/`unreact` aliases)
```
llun, line 49: **Doc overclaim (refuted, fixed anyway).** The verifier's argument was that nothing points agents at `docs/features.md`, so the inaccuracy is inert. Whether or not an agent reads it, the sentence was wrong: I wrote that chips "show who reacted with what", and the shipped chips show an emoji and a count with no account names anywhere. The accounts exist only behind `GET /api/v1/pleroma/statuses/:id/reactions`. Reworded to match what ships.

lib/components/posts/reaction-picker.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  return (
+    <>
+      {/* Outside-click overlay. aria-hidden keeps this full-viewport target out
+          of the accessibility tree. */}
+      <div className="fixed inset-0 z-30" aria-hidden onClick={onClose} />
+      <div
+        role="dialog"
+        aria-label="Choose a reaction"
+        className="bg-background absolute bottom-9 left-0 z-40 w-72 overflow-hidden rounded-xl border shadow-lg"
```
llun: **Picker positioning — rejected.** The claim was that `absolute bottom-9 left-0` opens off-screen and clips on narrow phones. The verifier rebuilt the exact layout chain in Chrome at 360×800 with the real internals and measured it: the panel's own `<Input autoFocus>` triggers React's `commitMount` focus with no `preventScroll`, so the browser scrolls it into view. No change.

lib/components/posts/reaction-picker.tsx
```diff
@@ -0,0 +1,229 @@
...
+  if (!customEmojiRequest) {
+    customEmojiRequest = getCustomEmojis()
+      .then((emojis) =>
+        emojis
+          .filter((emoji) => emoji.visible_in_picker)
+          .map((emoji) => ({ shortcode: emoji.shortcode, url: emoji.url }))
+      )
+      // A failure degrades to unicode-only rather than breaking the picker, and
+      // is not cached so the next open retries.
+      .catch(() => {
```
llun, line 58: **Dead `.catch` — rejected.** The claim was that `getCustomEmojis` resolves `[]` rather than rejecting, so the cache-clearing `.catch` can never run. It's chained *after* the `.then` mapper, which throws on any non-array body (a proxy or middleware answering 200 with JSON that isn't an array). The catch is reachable and the retry-on-failure behaviour is real. No change.

lib/database/sql/status.ts
```diff
@@ -1612,34 +1665,41 @@ export const StatusSQLDatabaseMixin = (
...
+        ? database('bookmarks')
+            .where('actorId', currentActorId)
+            .whereIn('statusId', [...hydrationStatusIds])
+            .select<{ statusId: string }[]>('statusId')
+        : Promise.resolve([]),
+      currentActorId && hasHydrationStatusIds
+        ? database('likes')
+            .where('actorId', currentActorId)
+            .whereIn('statusId', [...hydrationStatusIds])
+            .select<{ statusId: string }[]>('statusId')
```
llun: **Duplicate rollup query on Mastodon paths — rejected.** A Mastodon timeline does issue two grouped `status_reactions` queries: one hydrating the domain `Status` (which the web UI's `ReactionRow` reads) and one filling `getMastodonStatuses`' own `reactionsCache`.

The proposed fix — have the serializer prefer `status.reactions` — is unsafe. The domain field is hydrated with whatever viewer that DB call had (`currentActorId` on `getStatusesByIds`, `visibleToActorId` on the list paths, `undefined` on the outbox), and the serializer has its own `currentActorId`. Where they differ, preferring the domain value would serve a wrong `me` flag — a correctness bug traded for one grouped query per page. The two-layer cost is deliberate; the N+1 above was not.

app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts
Comment (llun) [inline app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts:47]: **Round-1 fix covered the wrong branch (confirmed, fixed).** Five of five reviewers flagged this independently, and they were right: `getStatusDetailPath` returns `/@user@host/<sha256(url)>` whenever `isLocalActor !== false`, so *every in-app link to a local post* — timeline click, notification, email — lands on the url-hash branch. That branch called `getStatusFromUrlHash`, which hydrated with no viewer at all, so "your own reaction renders as Add" survived on the common path while the replies right below it (fetched with a viewer) rendered correctly — the same page contradicting itself.

`GetStatusFromUrlHashParams` now takes `currentActorId` and the resolver forwards it on both hash lookups. Pinned by a test asserting the forwarding, since round 1's test only covered the `getStatus` call and that's exactly why the gap shipped.

lib/database/sql/timeline.ts
```diff
@@ -261,19 +260,17 @@ export const TimelineSQLDatabaseMixin = (
...
-            await statusDatabase.getStatus({
-              statusId,
-              currentActorId: actorId
-            })
-          )
-        }
-
-        return statuses.filter((status): status is Status => !!status)
+        // One batched hydration rather than a status-at-a-time loop: this is the
+        // busiest page in the app, and the per-status path re-queries detected
```
llun, line 264: **Home timeline N+1 (confirmed, fixed).** `MAIN`/`HOME`/`DIRECT` loaded statuses one at a time while `LOCAL_PUBLIC`/`FEDERATED_PUBLIC` already batched — so the busiest page in the app re-queried reaction rollups, detected languages, quote edges, bookmarks and likes once per row. My PR added the rollup query to that per-row path, so this was partly a regression I introduced.

Now uses the same `getStatusesByIds` the sibling branches use. Order is preserved (it maps over the requested ids), and the comment I replaced was about avoiding a *parallel fan-out* overflowing RSC async tracing — a single batched call doesn't fan out at all.

lib/database/sql/status.ts
llun, line 1: **Viewer vs visibility filter (confirmed, fixed).** Profile and hashtag timelines had no viewer to hydrate with, so their chips reported `me: false` while being fully interactive.

My first attempt reused `visibleToActorId` for this and that was wrong — it is a *visibility filter*: passing a viewer id there flips `recipientActorIds` from `null` to a recipient whitelist and changes **which statuses come back**. `getActorStatuses`, `getStatusReplies` and `getStatusesByHashtag` now take an explicit hydration-only `currentActorId`, documented as distinct from the filter, and the round-1 test that had conflated the two was corrected.

lib/services/reactions/getStatusReactionList.ts
```diff
@@ -35,18 +36,23 @@ export const getStatusReactionList = async ({
   })
   if (!status) return null
 
+  // Reactions are stored against the boosted post, never the boost — so read
+  // them back from the same target the write path used, or reacting through a
+  // boost's id would store a row that a GET on that same id cannot see.
+  const targetStatusId = getOriginalStatus(status).id
```
llun, line 42: **Write/read asymmetry on boosts (fixed).** `reactStatus`/`unreactStatus` resolve a boost to `getOriginalStatus(status)` before writing, but the read path queried the id it was given — so `PUT .../reactions/🔥` through a boost's id stored a row that `GET .../reactions/🔥` on that same id could not see. Both sides now resolve the same target.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,288 @@
...
+          aria-haspopup="dialog"
+          aria-expanded={isPicking}
+          className="border-border bg-background text-muted-foreground hover:bg-muted flex h-7 w-7 items-center justify-center rounded-full border transition-colors"
+          onClick={(event) => {
+            event.stopPropagation()
+            setIsPicking((value) => !value)
+          }}
+        >
+          <SmilePlus className="size-3.5" />
+        </button>
```
llun, line 308: **The picker could undo a reaction (fixed).** Found by the completeness critic, not by any dimension reviewer. `onPick` routed into `toggle`, which removes when `existing?.me` — so a picker item labelled "React with 🔥" *removed* a 🔥 the viewer had already added. `toggle` now takes an `intent`, and a pick never undoes.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,288 @@
...
+// A hot post can accumulate more distinct emoji than a row can usefully show.
+// The rollups arrive in first-reaction order, so the oldest — which is also the
+// most established — stay visible.
+const MAX_VISIBLE_CHIPS = 12
+
+// The write path accepts a unicode emoji or one of THIS instance's enabled
+// shortcodes — nothing else. So a chip is only offered as a control when the
+// viewer could actually succeed:
+//   • unicode                        → always
+//   • local shortcode, still enabled → the rollup carries a url (the rollup
```
llun, line 28: **Dead affordances (fixed).** A chip is only rendered as a control when this instance could actually react with it. A remote `shortcode@domain` is never locally reactable, and a local shortcode whose emoji has since been disabled has no url in the rollup (the rollup resolves urls only for enabled local emoji) — both 422'd on every click. They now render read-only. The second case came from the completeness critic; my first pass only handled the `@` form.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,288 @@
...
+  useEffect(() => {
+    // A write in flight owns the local state: resyncing from the prop mid-flight
+    // would discard the optimistic chip, then be overwritten by the server's
+    // rollups a moment later, flickering the count.
+    if (pendingNameRef.current) return
+    setReactions(status.reactions ?? [])
+    setError(null)
+  }, [status.id, status.reactions, setError])
+
+  const toggle = async (name: string, intent: 'toggle' | 'add' = 'toggle') => {
```
llun, line 146: **Overlapping writes (fixed).** Each response carries the status's *full* rollups, so two in-flight writes race and the loser's chip vanishes. The single-flight guard stays, but every chip and the picker trigger are now disabled while a write is pending, so the refusal is visible rather than silent.

Also fixed a bug I introduced while doing this: adding `pendingName` to the resync effect's deps made it re-run when a write settled and overwrite the server's rollups with the still-stale prop. It reads from a ref now. The strengthened reconcile test below is what caught it.

lib/components/posts/reaction-row.test.tsx
```diff
@@ -0,0 +1,384 @@
...
+    )
+
+    expect(screen.getByLabelText('Remove 🔥 reaction, 2')).toHaveAttribute(
+      'aria-pressed',
+      'true'
+    )
+  })
+
+  it('adds a reaction optimistically and reconciles with the server', async () => {
+    // Deliberately NOT the optimistic guess (which would be count 3, me true,
```
llun, line 74: **Unfalsifiable assertion (confirmed, fixed).** The served rollups were identical to the optimistic guess (count 3, me true), so `setReactions(result.reactions)` was never actually verified. They now differ from anything the optimistic path could invent — and this test immediately caught a real clobber bug in the same session, so it is load-bearing.

app/(timeline)/[actor]/ActorTimelines.tsx
Comment (llun) [inline app/(timeline)/[actor]/ActorTimelines.tsx:1]: **Reaction state not synced (confirmed, fixed).** Round 1 threaded `onReactionsChanged` through `Posts` but no surface supplied it, so a reaction was discarded on any remount — Radix `Tabs.Content` unmounts inactive content, so switching profile tabs and back dropped the chip while a like made the same way survived. Wired here and in `FavoritesTimeline`, the two surfaces that sync like state.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1959,6 +1960,20 @@ export const FitnessStatusDetail: FC<Props> = ({
           </div>
         </div>
 
+        {/* Reactions belong to the post, so they sit above the action bar here as
+            they do in `Post`. This page composes its own action row instead of
+            using `Posts`, so the row has to be added explicitly — without it a
+            fitness post is the one surface where an existing reaction is
+            invisible and no new one can be added. */}
+        {(currentActor || (status.reactions?.length ?? 0) > 0) && (
+          <div className="border-t px-4 pt-2.5">
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1969]: **Missing surface (fixed).** Also from the completeness critic. This page composes its own action row (reply/boost/like/bookmark/menu) rather than going through `Posts`, so it had no reaction row at all — a fitness post was the one surface where an existing reaction was invisible and no new one could be added. `ReactionRow` is now rendered above the action bar, and AGENTS.md records that this page has to opt in by hand.

lib/client.ts
```diff
@@ -466,6 +467,92 @@ export const likeStatus = async ({ statusId }: DefaultStatusParams) => {
...
```
llun, line 480: **Error-copy leak (refuted, fixed anyway).** My round-1 change forwarded the `error` field of *any* 4xx to the UI, and every generic 4xx here answers with the bare HTTP reason phrase — so a 401 would have shown the user "Unauthorized" as the explanation for a failed reaction. The route now tags its two deliberate 422s with a `reason`, and the client surfaces `error` only when that marker is present. Bodies are covered by tests.

llun:
**Round-2 finding outside this diff — recording, not fixing here.**

`GET /api/v1/pleroma/statuses/:id/reactions` returns every reactor account with no `limit` and no pagination. A reviewer raised it as an unbounded-response concern on an endpoint that also serves unauthenticated requests; a verifier refuted it as disproportionate, and I agree for *this* PR — the file is not in this diff (it landed in #1324) and the per-actor cap of 8 bounds distinct emoji per actor, not the number of reactors on a popular post.

Flagging it explicitly rather than letting it disappear into a refuted pile: capping the accounts array the way Mastodon paginates `favourited_by` is worth a follow-up.

lib/database/sql/status.ts
```diff
@@ -1201,12 +1215,71 @@ export const StatusSQLDatabaseMixin = (
...
+      })
+    }
+    return context
+  }
+
+  // The viewer's own like/bookmark state for a page of statuses, batched. Every
+  // list path that hydrates for a viewer uses this: passing a `currentActorId`
+  // into per-status hydration without it turns on a like lookup, a bookmark
+  // lookup and an announce lookup *per row*.
+  async function buildViewerHydration(
```
llun, line 1255: **Round-2 fix introduced an N+1 (confirmed, fixed).** Self-inflicted, and the more embarrassing kind: giving `getStatusReplies`/`getActorStatuses` a hydration viewer switched on three per-row lookups that had been skipped entirely while `currentActorId` was `undefined` — a like, a bookmark, and an announce query per status. An unbounded reply thread paid three round trips per reply, directly under a comment claiming that batch block exists to prevent exactly this.

All three list paths now share `buildViewerHydration`. I also refactored `getStatusesByIds` onto it rather than leaving its own inline copy — the two having separate implementations is precisely how they drifted apart in the first place.

Note the fix is to *extend the batch*, not to drop the viewer: before round 2, replies and profile posts rendered `isActorLiked`/`isActorBookmarked` as false for every signed-in viewer.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,322 @@
...
+const MAX_VISIBLE_CHIPS = 12
+
+// The write path accepts a unicode emoji or one of THIS instance's enabled
+// shortcodes — nothing else. So a chip is only offered as a control when the
+// viewer could actually succeed:
+//   • unicode                        → always
+//   • local shortcode, still enabled → the rollup carries a url (the rollup
+//     query resolves urls only for non-disabled local emoji)
+//   • local shortcode, disabled/gone → no url, so withheld
+//   • remote `shortcode@domain`      → namespaced, never locally reactable
```
llun, line 31: **A new reaction could be invisible (confirmed, fixed).** Rollups arrive oldest-first and the row truncates at `MAX_VISIBLE_CHIPS`. On a post that already carried 12 distinct emoji — routine on Akkoma/Misskey-heavy threads — a viewer's brand-new reaction sorted last and was dropped, so they'd add one and see nothing change except a `+N` counter that isn't expandable. No error, no confirmation, and the chip they'd need to click to undo it never rendered.

The viewer's own reactions are no longer truncated; the remaining slots go to the rest, still oldest-first. The per-actor cap of 8 is below the 12-chip limit, so a viewer's own chips can never crowd the row out entirely.

lib/database/sql/timeline.ts
llun, line 264: **Timeline batching — reviewed and cleared.** Round 3 put a dimension on this specifically, since replacing the deliberately-sequential loop on the busiest page was the riskiest thing in the PR. The finding was refuted, matching my own read: `getStatusesByIds` maps over the requested ids so order (including the min_id reverse) is preserved, it filters missing rows as the loop did, applies no visibility filter when `visibleToActorId` is absent, and passes the same `withReplies`. The un-chunked `whereIn` is the one sharp edge, but the two public-timeline branches already used this same call with the same `limit`, so it is existing risk rather than new.

The existing `Timeline.MAIN` tests covering ordering, out-of-order pagination and min_id-vs-since_id all passed unchanged, and I added one pinning what the change was *for* — the viewer's own reaction coming back flagged from the batch.

lib/database/sql/status.ts
```diff
@@ -1201,12 +1219,92 @@ export const StatusSQLDatabaseMixin = (
...
+        count: rollup.count,
+        me: rollup.me,
+        url: rollup.url,
+        static_url: rollup.staticUrl
+      })
+    }
+    return context
+  }
+
+  // The viewer's own like/bookmark state for a page of statuses, batched. Every
```
llun, line 1251: **Third lookup still N+1'd (confirmed, fixed).** Round 3 batched two of the three per-row lookups the viewer threading switched on and I stopped there. `getActorAnnounceStatus` was still running once per status — and worse, each hit fully hydrated an Announce (recipients ×2, attachments, tags, actor, counters, plus a complete nested hydration of the original status) purely to read `.id` and discard everything else.

It now comes from the same batch as likes and bookmarks, as a plain `originalStatusId → announceId` map, and the per-status fallback uses `getActorAnnouncedStatusId` — the id-only query that was already in this file. Pinned by a test asserting the boost id still resolves on a list path.

The verifier's correction is worth recording: these were issued inside `Promise.all`, so the cost was DB load and pool pressure rather than serial latency (full serialization on SQLite).

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -226,7 +230,11 @@ const Page: FC<Props> = async ({ params }) => {
   if (status.type !== StatusType.enum.Announce && status.reply) {
     let replyStatus = await database.getStatus({
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:231]: **Thread ancestors had no viewer (confirmed, fixed).** The two `getStatus` calls building `previouses` were the last viewer-less lookups on this page — so the same post rendered its reaction highlighted in the timeline and unhighlighted as an ancestor of the post you were reading, on one screen. This PR threaded the viewer into the focused status and the replies and missed the ancestors between them.

The reviewer also spotted the same gap on the signed-in fitness list (`app/(timeline)/fitness/page.tsx`); both are fixed.

app/api/v1/tags/[tag]/route.ts
```diff
@@ -108,6 +108,11 @@ export const GET = traceApiRoute(
         fetchBatch: ({ maxStatusId, limit }) =>
           database.getStatusesByHashtag({
```
Comment (llun) [inline app/api/v1/tags/[tag]/route.ts:109]: **Half-done fix, plus an inverted comment (confirmed, fixed).** Round 2 gave the hashtag page's server-rendered first page a viewer and stopped there — the load-more endpoint behind the very same page kept none. So the viewer's own reaction, like, bookmark and boost state flipped to false partway down a single scroll, and their own 🔥 on a page-2 post rendered as "Add", optimistically counted up on click, then snapped back when the server reported the reaction already existed.

Worse, the comment I wrote alongside that round-2 change claimed the opposite — that the client-paginated pages already had it. Both are fixed, and the route test now pins the argument rather than only asserting the call happened, which is why nothing caught this.

This came from the whole-diff sweep, not from re-reading the newest commit. It's the third time in this review a fix addressed the reaction-shaped instance of a problem and left its neighbour, which is what that dimension was added to find.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,332 @@
...
+              aria-label={`${reaction.name} reaction, ${reaction.count}`}
+            >
+              {body}
+            </span>
+          )
+        }
+
+        return (
+          <button
+            key={reaction.name}
```
llun, line 265: **One of my fixes defeated another (confirmed, fixed).** Round 2 added focus restoration to the picker; rounds 2–3 added disable-while-pending to prevent racing writes. Together they cancel out: the browser blurs a disabled control, so pressing Enter on a picker item focused the trigger and then immediately disabled it — dumping the keyboard user on `<body>` with their tab position reset to the top of the document. Exactly the outcome the focus-restoration comment two lines away says it exists to prevent.

Now `aria-disabled` plus the busy styling: assistive tech is told the control is unavailable, the click handler returns early, and nothing leaves the tab order or loses focus. Pinned by a test asserting the trigger still has focus across a full picker-driven round trip.

lib/types/database/operations.ts
```diff
@@ -575,10 +577,19 @@ export type DeleteStatusParams = BaseStatusParams & {
 
 export type GetStatusFromUrlParams = {
```
llun, line 578: **Same pattern, one more instance (confirmed, fixed).** `getStatusFromUrl` was the un-hydrated twin of `getStatusFromUrlHash`, which this PR fixed in round 2 for exactly this reason.

The reviewer surfaced why it matters more than it looks: `createNote` stores `id = <actor>/statuses/<uuid>` but `url = https://<domain>/@user/<uuid>`, so a status's permalink is *never* its id — meaning the id lookup on the line above essentially never hits for a pasted link and this fallback is the real resolution path, including for your own posts. The resolved hit rendered with no reaction/like/bookmark state while the indexed hits beside it in the same response rendered with it.

Notably my own parallel sweep missed this: I enumerated the *pages* that render chips and checked each one's data source, which is why the search route's URL-resolution fallback never came up. Enumerating the status-resolution functions instead would have caught it.

lib/components/posts/reaction-row.tsx
```diff
@@ -0,0 +1,338 @@
...
+            .filter((reaction) => reaction.count > 0)
+        : existing
+          ? previous.map((reaction) =>
+              reaction.name === name
+                ? { ...reaction, me: true, count: reaction.count + 1 }
+                : reaction
+            )
+          : [
+              ...previous,
+              { name, count: 1, me: true, url: null, static_url: null }
```
llun, line 180: **Refuted, fixed anyway (small).** A reviewer called this the same bug as the disabled-control focus theft; the verifier correctly refuted that — the chip unmounts because the user deleted the reaction it represents, which is not the same as a control that stays rendered being made unfocusable — and confirmed it is not a regression from this PR.

The refutation is right on both counts, but a keyboard user removing their only reaction still lands on `<body>`. Three lines to move focus to the adjacent trigger, so worth closing rather than leaving as a known wrinkle.

lib/components/posts/reaction-picker.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  return (
+    <>
+      {/* Outside-click overlay. aria-hidden keeps this full-viewport target out
+          of the accessibility tree. */}
+      <div className="fixed inset-0 z-30" aria-hidden onClick={onClose} />
+      <div
+        role="dialog"
+        aria-label="Choose a reaction"
+        // Opens downward (like the composer's emoji picker) rather than upward.
```
llun: **Picker rendered off-screen (confirmed, fixed).** A non-portaled popover growing ~294px upward from `bottom-9`, on a page where the focused post's chip row sits ~161px below the top of its card — and that card was `overflow-hidden`, so the panel's head was clipped into territory that can never be scrolled back into view (block-start overflow is not part of the scrollable region).

I measured it on the running app with the real compiled CSS rather than trusting the estimate:

| placement | panel top | vs card top (89px) |
|---|---|---|
| old `bottom-9` | **-141px** | 230px above it — and off the viewport entirely |
| new `top-full mt-2` | **197px** | fully inside the card and the 720px viewport |

So it was worse than reported: not merely clipped, but rendered off-screen. Now opens downward like the composer's emoji picker, and the detail card drops `overflow-hidden` — which is exactly what `Posts` already did, deliberately, for this same class of non-portaled overlay.

Worth noting round 2 had a reviewer claim the picker was mispositioned and a verifier refute it by measuring — on a *timeline* layout, where it genuinely is fine. The refutation was sound for what it measured; the detail-page card was a different context.

lib/database/sql/collection.ts
```diff
@@ -304,11 +306,14 @@ const readCollectionFeed = async ({
   const rows = await query
```
llun, line 306: **Fourth instance of the projection/viewer conflation (refuted, fixed anyway).** The public collection projection hydrated anonymously, with a comment calling that intentional. The verifier refuted the finding on the strength of that comment — but it predates interactive reaction chips, and `CollectionDetail` renders `<Posts currentActor showActions={Boolean(currentActor)} />`. So a signed-in visitor to a public collection saw actionable controls reporting their own reaction, like and bookmark as un-acted; clicking their own 🔥 was treated as an add, counted up, then snapped back.

The same reasoning would have refuted the hashtag fix two rounds ago — that timeline is equally public, and hydrating it for the viewer was right there too. `projection` still decides *which* statuses the feed contains; a separate hydration-only viewer decides *whose state* they carry. Fixed on the page, the load-more route, and the owner's public-preview tab.

That makes four surfaces where this exact conflation shipped (status detail, hashtag, search-by-permalink, collections), which is why I stopped treating it as a series of one-off misses and enumerated every status-resolution function instead.

lib/components/posts/reaction-picker.tsx
```diff
@@ -0,0 +1,315 @@
...
+  onPick: (name: string) => void
+  onClose: () => void
+  // The button the picker hangs off. Its viewport rect anchors the panel.
+  anchorRef: RefObject<HTMLButtonElement | null>
+}
+
+const PANEL_WIDTH = 288
+const VIEWPORT_MARGIN = 8
+
+// The panel is rendered in a portal at the document root and positioned from
```
llun, line 76: **Both confirmed findings fixed — but not the way either suggested.**

Round 8 found the picker clipped in two more containers (the fitness header card, the search results section) after round 7 fixed it in the two status-detail cards. Both suggested fixes were "drop `overflow-hidden` from this container too".

I checked how many containers could do this: **four files** render posts inside an `overflow-hidden` ancestor — the two the reviewers found, plus `CollectionDetail`, which nobody had flagged. Every card that wraps posts has it, for its rounded corners, and the panel is taller than the space inside them. Fixing them one at a time is what produced rounds 7 and 8 in the first place.

So the picker now renders in a **portal at the document root**, positioned from the trigger's viewport rect. Ancestor overflow cannot reach it by construction — the container list stops mattering. It opens below the trigger, flips above when there's no room, follows the trigger on scroll and resize, and clamps to the viewport. Round 7's `overflow-hidden` removal is reverted, so the cards go back to clipping their own corners.

Pinned by tests: one asserting the panel is not a descendant of the post (it renders inside a deliberately `overflow: hidden` wrapper in the test), and three asserting the placement math.

One extra bug came out of verifying this. Measuring against a zero-width viewport, the panel landed at `left: -296px` — my clamp was `min(max(margin, left), width - panel - margin)`, and when the viewport is narrower than the panel that inner expression goes negative and `min` takes it, pushing the panel off-screen. Clamping the right edge first fixes it; the narrow-viewport test case fails against the old ordering.

llun:
**Round 9 — two confirmed findings, deliberately NOT fixed here.**

Both are real and worth fixing. Neither is in this PR's scope, and both verifiers said so themselves:

1. `app/api/v1/statuses/[id]/context/route.ts` — `collectDescendants` omits the viewer while `collectAncestors` passes it, so one `/context` response reports the viewer's own state correctly for ancestors and as `false` for every reply.
2. `lib/services/notifications/getMastodonNotification.ts:119` — `getStatus` omits `currentActorId` although it is destructured two lines above and handed to `getMastodonStatus` immediately after. The v2 grouped path does it correctly, so `/api/v1/notifications` and `/api/v2/notifications` disagree about the same status.

Why they're out of scope:

- **Neither file is in this diff** (`git diff origin/main --stat` on both is empty).
- **Neither is a reaction bug.** Reactions resolve correctly on both paths — `/context` via the batched `reactionsCache`, notifications via `getStatusReactions`' viewer-scoped fallback query. What's wrong on both is `favourited` / `bookmarked` / `reblogged`.
- They are **pre-existing**; this PR only makes the first one *fixable* by adding `currentActorId` to `GetStatusRepliesParams`.

Fixing them would mean changing Mastodon API responses for favourites, bookmarks and boosts from inside a reaction-UI PR. That's a separate change with its own blast radius, and it deserves its own review rather than riding along at round 9 of this one.

Recording them here so they don't vanish into a closed review: both are two-line fixes (`currentActorId: currentActor?.id` at each call site), and the notifications one is the more visible of the two since it affects every row of the v1 notifications list.

PR #1331, opened 2026-07-26
test: de-flake the heatmap setData assertion

llun:
## Summary

`RouteHeatmapMap > pushes updated geometry to the existing source when the cache changes` is flaky in CI (always **Shard 4 of 4**) and intermittently blocks unrelated PRs:

```
AssertionError: expected "vi.fn()" to be called 1 times, but got 2 times
 ❯ app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx:1355:41
```

**The test was wrong, not the component.**

## Root cause

The component's contract is one `setData` per `routeGeoJson` change (`lib/components/fitness/RouteHeatmapMap.tsx:373`). The "second" call is not a redundant re-render — it is the **initial** push, which the test intends to exclude with `setData.mockClear()` but never actually waits for:

- `OpenFreeMap` is rendered in the commit that flips `isMapLoaded`
- the first `setData` runs in that commit's **passive effect** — a later macrotask
- `await screen.findByText('OpenFreeMap')` can resolve between the two

When it does, `mockClear()` clears nothing, both the initial push and the rerender push land afterwards, and the count is permanently 2 — so `waitFor` can only retry until it times out.

Reproduced deterministically by observing the label outside React Testing Library's act drain: `raceAfterFind=0` → `raceAfterSettle=2`.

Evidence it is pre-existing and not caused by any particular change: the same commit SHA (`04583655b`, #1322) produced one passing and one failing CI run. Prior instances: runs `30117916423` and `30107189026`, both Shard 4.

## Second finding: the test never checked the geometry

The fixture's "updated" third point `(52.37, 4.89)` is **exactly collinear** with the endpoints, so `simplifySegmentsToBudget`'s Douglas-Peucker pass at the 1m tolerance dropped it. Dumping both payloads showed them byte-identical:

```
[[4.88,52.36],[4.91,52.39]]   // initial
[[4.88,52.36],[4.91,52.39]]   // "updated"
```

A test named *pushes updated geometry* was asserting only that `setData` fired again — it would have passed while pushing the old geometry.

## Changes

- **Wait for the initial push before clearing** (`waitFor(() => expect(setData).toHaveBeenCalled())`), so the clear can no longer race the effect it is meant to exclude. This is a logical guarantee, not a timing one.
- **Move the third fixture point off the line** so it survives simplification, and assert the **last call's payload** plus `expect(mapConstructor).toHaveBeenCalledTimes(1)` (the update went to the *existing* source, no remount) — instead of counting calls.
- The two sibling `toHaveBeenCalledTimes(1)` assertions inside `waitFor` (the polling tests) have the same shape and both immediately capture the real count afterwards, so they become `toHaveBeenCalled()` gates. The trailing `toHaveBeenCalledTimes(callsAfterInitialLoad)` is a settled assertion outside `waitFor` and is correct as-is — left alone.

No component guard was added: a redundant `setData` would be harmless and idempotent, so exactly-once is not a real invariant worth enforcing.

## Test results

- Replayed the exact CI interleaving against the new structure: `atLabel=0` (raced) yet passes, 5/5
- **20/20** clean runs of the file
- Full gate: `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**731 files, 7324 tests passed**)

## Notes

Test-only change. No migrations, no config, no user-facing behavior, no docs affected.

### DIFF
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1207,9 +1207,9 @@ describe('FitnessHeatmapView', () => {
       />
     )
 
-    // A gate, not a count assertion — the exact number is captured below. An
-    // exact count inside waitFor can never recover if the extra call lands
-    // before the first poll.
+    // A gate, not a count assertion — the exact number is captured below. Call
+    // counts only ever grow, so an exact count inside waitFor can never recover
+    // if an extra call lands before the count is captured below.
     await waitFor(() => {
       expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalled()
     })
@@ -1347,7 +1347,7 @@ describe('RouteHeatmapMap', () => {
       <RouteHeatmapMap
         heatmap={worldHeatmap({
           updatedAt: 3,
-          // The third point must sit off the straight line between the
+          // The interior point must sit off the straight line between the
           // endpoints: simplifySegmentsToBudget runs Douglas-Peucker at a 1m
           // tolerance, so a collinear point is dropped and the "updated"
           // geometry would be byte-identical to the original.
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1209,7 +1209,7 @@ describe('FitnessHeatmapView', () => {
 
     // A gate, not a count assertion — the exact number is captured below. Call
     // counts only ever grow, so an exact count inside waitFor can never recover
-    // if an extra call lands before the count is captured below.
+    // if an extra call lands before waitFor's first successful check.
     await waitFor(() => {
       expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalled()
     })
@@ -1338,8 +1338,10 @@ describe('RouteHeatmapMap', () => {
     // Wait for the *initial* push before clearing, not just for the provider
     // label. The label appears in the commit that flips isMapLoaded, while the
     // first setData runs in that commit's passive effect — a later macrotask.
-    // Clearing on the label alone can therefore land between the two, leaving
-    // the initial push to arrive after the clear and inflate every later count.
+    // Clearing on the label alone can therefore land between the two, in which
+    // case the clear silently does nothing and the initial push stays in the
+    // record — which is what made the old exact-count assertion here
+    // unsatisfiable.
     await waitFor(() => expect(setData).toHaveBeenCalled())
     setData.mockClear()
 
--- a/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
+++ b/app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
@@ -1207,8 +1207,11 @@ describe('FitnessHeatmapView', () => {
       />
     )
 
+    // A gate, not a count assertion — the exact number is captured below. An
+    // exact count inside waitFor can never recover if the extra call lands
+    // before the first poll.
     await waitFor(() => {
-      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(1)
+      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalled()
     })
 
     const callsAfterInitialLoad = mockGetFitnessRouteHeatmaps.mock.calls.length
@@ -1245,8 +1248,9 @@ describe('FitnessHeatmapView', () => {
       />
     )
 
+    // A gate, not a count assertion — see the sibling test above.
     await waitFor(() => {
-      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalledTimes(1)
+      expect(mockGetFitnessRouteHeatmaps).toHaveBeenCalled()
     })
 
     const callsAfterInitialLoad = mockGetFitnessRouteHeatmaps.mock.calls.length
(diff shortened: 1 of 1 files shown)


app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -1331,18 +1335,27 @@ describe('RouteHeatmapMap', () => {
...
+    // Clearing on the label alone can therefore land between the two, leaving
+    // the initial push to arrive after the clear and inflate every later count.
+    await waitFor(() => expect(setData).toHaveBeenCalled())
     setData.mockClear()
 
     rerender(
       <RouteHeatmapMap
         heatmap={worldHeatmap({
           updatedAt: 3,
+          // The third point must sit off the straight line between the
```
llun: **Comment is inaccurate as placed.** The point this PR actually moved is `points[1]` — the middle/interior point (`lng` 4.89 -> 4.91). Read literally against the array it annotates, "the third point" is `{ lat: 52.39, lng: 4.91 }`, which is an *endpoint* and therefore lies on the line by definition, making the sentence false.

The intended sense is "third" relative to the 2-point base fixture. Reword to "the interior point" so it matches the array it sits above.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -1207,8 +1207,11 @@ describe('FitnessHeatmapView', () => {
       />
     )
 
+    // A gate, not a count assertion — the exact number is captured below. An
+    // exact count inside waitFor can never recover if the extra call lands
+    // before the first poll.
```
llun: **Ambiguous wording.** This file uses "poll" for the component's 5s heatmap polling interval, so "before the first poll" reads as if it refers to that — but it means `waitFor`'s own retry tick.

Worth disambiguating, because the mechanics differ: under fake timers `waitFor` advances the clock only in 50ms steps up to its 1000ms timeout, so the 5000ms heatmap poll can never fire inside this gate. Suggest "before the count is captured below".

llun (commented):
Sub-agent review round 1 (3 reviewers: concurrency/determinism, project conventions, assertion strength).

All three independently verified the root cause and the fix:

- The race was reproduced: with an artificially slowed `waitFor` callback, 1/20 runs had `setData` still uncalled at the moment the `OpenFreeMap` label resolved.
- The collinearity claim was confirmed against the real pipeline: the old middle point's perpendicular distance is **exactly 0 m** at a 1 m tolerance, and the old "updated" payload came out byte-identical to the original. The old assertion genuinely could not detect a stale payload.
- The new point sits **~1159 m** off the chord (~1159x the tolerance), so there is no marginal-tolerance flake risk.
- `expect(mapConstructor).toHaveBeenCalledTimes(1)` is not vacuous. Because `getSource` returns a fresh wrapper around the same spy, the payload assertion alone cannot distinguish "pushed to the existing source" from "torn down and re-seeded" — this is the only assertion carrying the "existing source" half of the test name. Verified by mutation: changing `bounds` on the rerender gives 2 constructions while `lastCall` stays identical.
- No other structurally-racy assertion remains in the file; every other exact-count assertion sits outside `waitFor`, after a gate.

Two comment-accuracy findings below.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in b1546eadf — reworded to "The interior point must sit off the straight line between the endpoints", which matches the array it annotates.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in b1546eadf. Now reads: "Call counts only ever grow, so an exact count inside waitFor can never recover if an extra call lands before the count is captured below." That drops the "poll" collision and states the monotonicity that makes the old form unrecoverable.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
```diff
@@ -1331,18 +1335,27 @@ describe('RouteHeatmapMap', () => {
     const { rerender } = render(
       <RouteHeatmapMap heatmap={worldHeatmap()} mapProvider={{ type: 'osm' }} />
     )
-    await screen.findByText('OpenFreeMap')
+    // Wait for the *initial* push before clearing, not just for the provider
+    // label. The label appears in the commit that flips isMapLoaded, while the
+    // first setData runs in that commit's passive effect — a later macrotask.
+    // Clearing on the label alone can therefore land between the two, leaving
+    // the initial push to arrive after the clear and inflate every later count.
```
llun: **Stale clause introduced by this PR.** The first two sentences are accurate, but "and inflate every later count" no longer describes anything: this same commit deleted the only later `setData` count assertion. The one remaining `toHaveBeenCalledTimes` is on `mapConstructor`, which a leaked `setData` push cannot inflate.

Documentation-only (no runtime consequence — React flushes a commit's pending passive effects before processing the next render, so `lastCall` would be the new geometry either way), but a reader following this comment will hunt for a "later count" that does not exist. Reword to say what the guard actually buys: the clear otherwise silently does nothing and the initial push stays in the record.

llun (commented):
Sub-agent review round 2 (fresh-eyes reviewer, no context from round 1).

Confirmed clean on: mutation-sensitivity of the payload assertion (flipping the expected interior coordinate makes the test fail with a proper diff), non-vacuity of both new assertions, `mapConstructor` scope resolution, types under `mock.lastCall?.[0]`, no dead variables or unused imports, and that neither loosened gate can go vacuous (`callsAfterInitialLoad` can never be 0 because the gate guarantees >= 1).

Verification: 44/44 on 3 consecutive runs, `prettier --check` clean, `eslint` exit 0. CI is green on both commits including **Test Shard 4 of 4** — the shard this flake was failing on.

One stale-comment finding below.

app/(timeline)/fitness/heatmap/FitnessHeatmapView.test.tsx
llun: Fixed in 1ab4b9efd. Now reads: "…in which case the clear silently does nothing and the initial push stays in the record — which is what made the old exact-count assertion here unsatisfiable." No dangling reference to a count that no longer exists.

Also took the sub-nit: the sibling gate comment now says "before waitFor's first successful check" rather than "before the count is captured below", which was one step too wide.

PR #1332, opened 2026-07-26
fix: store a JPEG copy of the route map for the import email

llun:
Follow-up to #1327, which added the fitness activity import email with a route map.

## Problem

The email's map `` pointed at the stored WebP media URL. Gmail, Apple Mail and new Outlook render WebP; **Outlook desktop (Word rendering engine) and Windows Mail do not** — those recipients saw the `alt="Route map"` text instead of their route. This was deliberately deferred from #1327 to keep that PR reviewable.

An on-demand JPEG rendition off `/api/v1/files/[...pathname]` is not viable: with object storage that route answers `Response.redirect(redirectUrl, 308)`, so there are no bytes to transcode. It has to be a stored asset.

## What changed

**Storage takes an output format.** Both backends hardcoded `.webp({ quality: 95, smartSubsample: true, nearLossless: true })`, so every image `saveMedia` stores is WebP. `_saveImageBuffer` / `_uploadImageBufferToS3` now take an optional `format`, **defaulting to `'webp'` so nothing else changes**, and both encode through one shared `encodeImageOutput` helper instead of duplicating the encoder options. JPEG is baseline (not progressive) with `chromaSubsampling: '4:4:4'` — a route map is a saturated 4px polyline over pale tiles, which is exactly what chroma subsampling smears.

**The copy is not a `medias` row.** `saveMediaImageRendition` stores an extra encoding and returns `{ path, url, bytes, mimeType, metaData }` without creating one. Two reasons: it would be a `medias` row attached to nothing, and (per the note in the issue) `medias.original.mimeType` records the *input* type — which is why today's map attachment claims `image/png` while serving WebP bytes — so a JPEG stored that way would claim `image/png` too. The rendition's `mimeType` is the type actually written. Quota is still checked, like `saveThumbnail`.

**New `fitness_files.mapImageEmailPath` column.** `processFitnessFileJob` stores a JPEG copy of the same map buffer, records its path there, and points the email at it — falling back to the WebP when there is no copy, which is every activity imported before this shipped. The copy is only made when an email is actually going out (`notifyOnComplete`): a direct upload notifies nobody, so a copy would be storage spent on an image no one would ever fetch. It is not attached to the status and never federates.

**Lifecycle.** Reprocessing clears the column alongside `mapImagePath`. **Regenerate maps** deletes the copy and clears the column: the email it served was sent when the activity arrived, and after a privacy-location change the copy may show a route the owner has since hidden. `scripts/backup/productionArchive.ts` now collects the path so the copy travels with an archive.

Both schema dumps (`migrations/schema.sql`, `migrations/schema.sqlite.sql`) are regenerated from the full migration chain against fresh local PostgreSQL 17 and SQLite databases — the diff in each is exactly the one new column.

## Sizing

Measured on a representative 800×600 map: the existing pipeline upscales to 4000×3000 (pre-existing `resize(MAX_WIDTH, MAX_HEIGHT)` without `withoutEnlargement`), giving a 271 KB WebP and a 442 KB JPEG — the same order of magnitude as the image the email already linked, so the copy stays on the identical pipeline rather than gaining a second size knob. The upscaling itself affects every upload in the product and is out of scope here.

## Verification

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**734 files, 7394 tests, all passing**).
- New tests: the shared format helper (real `sharp`, asserts the encoded format and that JPEG is baseline); `LocalFileStorage` writes a real JPEG to disk (default `saveFile` still writes WebP), reports the *written* file's size/dimensions, returns `null` for non-image input, and refuses over quota; `S3FileStorage` uploads with `ContentType: image/jpeg` and a `.jpg` key (WebP by default); `saveMediaImageRendition` dispatch across all three storage types and the unconfigured case; the column's round-trip and clearing; the job storing the copy, the email HTML carrying the `.jpg` and never the `.webp`, both fallback paths (`null` and throw) still completing the import with its map, and no copy stored when no email is going out; regeneration deleting the copy.
- `./scripts/mock/renderEmailPreviews.ts` renders all 11 templates, and the rendered output is **byte-identical to `main`** — no template markup changed, only the URL passed in. (Per `AGENTS.md`, the fitness fixture deliberately passes no map URL, since a stand-in image has the wrong aspect ratio.)
- Eyeballed the JPEG the encoder actually produces from a map-shaped source: the route line is clean, with no chroma bleeding and thin grid lines intact.

**Not verified here:** delivery to a real Outlook desktop / Windows Mail inbox. That needs an actual send, which this environment cannot do — it remains the outstanding real-inbox check for the email work.

## Docs

`docs/fitness-file-storage.md` (column + pipeline step), plus the convention itself in the `AGENTS.md` email section and the `REVIEW.md` checklist: an email image never points at a stored media path directly.

---

## Review round 1 (sub-agent loop)

Three agents reviewed the media-storage code, the job/DB/migration/script code, and the docs + email completeness. All 10 findings are posted as inline comments on this PR and fixed in the second commit. The substantive ones:

**A copy has no `medias` row, so nothing generic can find it — and two places needed to.**
- `scripts/maintenance/cleanupMediaStorage.ts` classified every copy as orphaned and would have deleted files the database still points at. It now reads the column.
- Running that script to verify also exposed a **pre-existing bug on `main`**: it rebased already-relative stored paths with `path.relative`, producing `../../…` strings that matched nothing, so for local storage it reported *every* media file as orphaned and `--yes` would have deleted all of them. Only absolute paths are rebased now. Verified against a local SQLite DB with a referenced WebP, a referenced JPEG copy and one genuine orphan: 3 orphans reported before, exactly 1 after.

**A reprocess nulled the column without deleting the file.** A retry or recovery script leaves `notifyOnComplete` false, so nothing rewrote it — the object was orphaned, and if the owner had just added a privacy location, the old **unfiltered** route stayed fetchable at its unchanged URL. Deleting the copy is now the job of one shared `deleteEmailMapImage`, used wherever the reference is dropped: import, reprocess, re-import, map regeneration, and activity deletion. `importFitnessFilesJob` had the same hole from the other side (it reset `mapImagePath` but not `mapImageEmailPath`, and a file that ends up non-primary never reaches `processFitnessFileJob` to rewrite it).

**Storing a copy is now gated on an email genuinely going out** — `notifyOnComplete` plus the instance having email configured plus the owner's activity-import preference. Previously a self-hosted instance with no `ACTIVITIES_EMAIL_*` stored a second copy of every imported map that nothing could ever fetch.

**Quota:** the pre-check refuses a write by an over-quota account, but the bytes are not added to the usage counters, which are only maintained alongside `medias` rows. Metering would need new database-interface surface (there is no public counter-adjust operation), which seemed disproportionate for one small file per emailed import — so the comments and docs now state the actual behaviour instead of implying enforcement. Say the word if you'd rather have it metered.

**Tests hardened where the reviewers proved coverage was absent, by mutation rather than by reading:**
- The WebP encoder options are now pinned by byte-comparing output — preserving them for existing uploads is the one invariant this refactor rests on, and `quality: 95 → 60` previously passed the whole suite.
- `regenerateFitnessMapsJob`'s non-primary-heal and no-visible-coordinates branches were both green with the new code deleted; both are now covered, and the failed-regeneration case asserts the reference *survives*.
- Reprocess cleanup and the no-email gate are covered, each verified to fail when its production line is reverted.

**Docs corrected where they overstated.** The WebP fallback is **live** (it fires on unconfigured storage, over-quota, or a failed encode), not legacy-only — the old wording would have licensed deleting `?? storedMap.url` as dead code. "Every image the media storages write is WebP" became false three lines above the sentence describing the JPEG, and that absolute is what justified the two hardcoded `mimeType: 'image/webp'` literals this PR removed. Also aligned the stale "route-map PNGs" wording in `docs/environment-variables.md` and `docs/features.md` (the PNG is the render buffer; the stored form is WebP).

Confirmed clean by the reviewers: the `importStravaActivityJob` call site passes no map URL (it is the no-exportable-streams fallback note, which has no map — the real Strava path routes through `processFitnessFileJob`), and no other email template embeds stored media, so the route map was the only exposure.

Gate re-run after these changes: prettier → lint (0 errors) → build → test, **735 files / 7405 tests passing**.

Separately filed, not fixed here: the shared pipeline upscales every image below 4000px (`resize(MAX_WIDTH, MAX_HEIGHT)` with no `withoutEnlargement`), which is why the copy is 442 KB instead of ~52 KB. It affects every upload in the product, so it wants its own PR.


scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -27,25 +27,37 @@ async function getAllMediaPathsFromDatabase(
...
-        paths.add(originalPath)
-      }
-      if (media.thumbnail) {
-        // For local file storage, normalize to relative paths
-        const thumbnailPath = basePath
-          ? path.relative(basePath, media.thumbnail)
-          : media.thumbnail
-        paths.add(thumbnailPath)
-      }
+      addPath(media.original)
```
llun, line 55: **Finding (fixed): this script classified every JPEG copy as orphaned and would have deleted files the database still points at.**

The referenced set was built only from `medias.original`/`thumbnail`. A copy stored by `saveMediaImageRendition` deliberately has no `medias` row — `fitness_files.mapImageEmailPath` is its only reference — so it looked orphaned. Failure scenario: 50 Strava imports with GPS store 50 objects under `medias/<date>/<hex>.jpg`; an operator runs the documented `--yes` and all 50 vanish while the DB still references them, so already-delivered emails render the `media-removed.svg` placeholder. The sibling `productionArchive.ts` *was* taught about the column in the first commit; this script was missed.

Fixed by reading the column here too.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -27,25 +27,37 @@ async function getAllMediaPathsFromDatabase(
   const database = knex(config.database)
 
   try {
+    const paths = new Set<string>()
```
llun, line 36: **Pre-existing bug found while verifying the fix above, by actually running the script.**

My first dry-run reported the JPEG copy as orphaned *and also the WebP that definitely has a `medias` row* — all three test files. Cause: stored paths are already relative to the storage root, but `path.relative(basePath, value)` resolved a relative value against the CWD, producing a `../../…` string that matched nothing. For local storage this script therefore reported **every** media file as orphaned, and `--yes` would have deleted all of it.

This is on `main`, not introduced here, but my fix above is inert without it. Now only absolute values are rebased. Verified against a local SQLite DB with a referenced WebP, a referenced JPEG copy and one genuine orphan: before, 3 orphans reported; after, exactly 1.

lib/jobs/processFitnessFileJob.ts
```diff
@@ -240,6 +332,18 @@ export const processFitnessFileJob = createJobHandle(
         buffer: fitnessBuffer
       })
 
+      // The reset below de-references any copy an earlier run stored, so delete
+      // the file too. A retry or a recovery script leaves notifyOnComplete
+      // false, so nothing would rewrite the column: without this the object is
+      // orphaned, and worse, if the owner added a privacy location before
+      // reprocessing, the old UNFILTERED route would stay fetchable at its
+      // unchanged URL.
+      await deleteEmailMapImage({
```
llun, line 362: **Finding (fixed): a reprocess nulled the column without deleting the file, and could leave an unfiltered route publicly fetchable.**

The per-run reset clears `mapImageEmailPath`, but a retry or recovery script leaves `notifyOnComplete` false, so nothing rewrote it. Failure scenario: a Strava import stores a copy, the worker is OOM-killed before `completed` (a documented failure mode here, hence `isFitnessProcessingStuck`), the owner clicks retry — the column goes NULL and the object is orphaned, reclaimable only by `cleanupMediaStorage`.

Worse, the privacy edge this PR already cites as the reason the regenerate job deletes: add a privacy location, then reprocess, and the new run draws a filtered map while the **old unfiltered** route JPEG stays fetchable at its unchanged URL.

Fixed by deleting the copy wherever the reference is dropped, via one shared `deleteEmailMapImage` now used by this job, `importFitnessFilesJob`, `regenerateFitnessMapsJob` and `deleteFitnessFile`. `importFitnessFilesJob` had the same hole from the other direction: it reset `mapImagePath` but not `mapImageEmailPath`, and a file that ends up non-primary never reaches this job to rewrite it. Covered by a test that fails under mutation.

lib/jobs/processFitnessFileJob.ts
```diff
@@ -321,7 +427,26 @@ export const processFitnessFileJob = createJobHandle(
...
 
-              mapImageUrl = storedMap.url
+              // The WebP is what the post and every web surface use; the email
+              // prefers a JPEG copy, because Outlook desktop cannot decode
+              // WebP. Only store one when an email is genuinely going out,
+              // otherwise it is storage spent on an image nobody will ever
+              // fetch: a direct upload notifies no one, an instance with no
+              // email configured sends nothing, and the owner may have turned
+              // activity-import emails off.
+              mapImageUrl = (await willSendImportEmail({
```
llun, line 458: **Finding (fixed): `notifyOnComplete` only approximates "an email is going out", so instances that send no email still stored a copy of every map.**

The gate was decided long before any email gate: the notification can be filtered by policy, the instance may have no `ACTIVITIES_EMAIL_*` configured at all (`sendNotificationAlerts` wraps the whole email block in `if (config.email)`), and the owner may have activity-import emails switched off. On a self-hosted instance without email, every Strava import with GPS wrote a second copy of the map that nothing could ever fetch — and rendition bytes are not metered, so that growth is invisible in the usage counters.

Fixed with `willSendImportEmail`, which also checks `getConfig().email`, `actor.account` and the existing `shouldSendEmailForNotification` predicate. The one case still not knowable here is the notification policy filtering the event, which is decided after the map exists; the docs now say "an import that is going to email the owner" rather than claiming certainty.

lib/services/medias/types.ts
```diff
@@ -169,6 +183,19 @@ export interface MediaStorage {
     actor: Actor,
     file: File
   ): Promise<ThumbnailStorageOutput | null>
+  // Processes and stores an image in a specific output format, without creating
```
llun, line 186: **Finding (fixed): the doc comment claimed this "enforces the account storage quota", which overstated it.**

Usage is counter-based (`CounterKey.mediaUsage` / `fitnessUsage`) and those counters are only maintained alongside `medias` rows, which a rendition has none of — unlike `saveThumbnail`, whose bytes are folded in by `updateMedia`'s `thumbnailUsageDelta`. So the check refuses a write by an already-over-quota account but the bytes are never added.

Metering properly would mean new database-interface surface (there is no public counter-adjust operation), which I judged disproportionate for one small file per emailed import — especially now that a copy is only stored when an email is genuinely going out. Fixed by making the comments here and in both backends state the actual behaviour, and documenting it in `docs/fitness-file-storage.md` so the trade-off is visible rather than implied. Happy to add the counter plumbing instead if you'd rather have it metered.

lib/jobs/regenerateFitnessMapsJob.test.ts
```diff
@@ -232,22 +236,25 @@ describe('regenerateFitnessMapsJob', () => {
       })
       expect(fitnessFile).toBeDefined()
 
+      const oldEmailMapPath = `medias/old-route-map-${postId}-${index}.jpg`
       await database.updateFitnessFileActivityData(fitnessFile!.id, {
         hasMapData: true,
-        mapImagePath: oldMedia!.original.path
+        mapImagePath: oldMedia!.original.path,
```
llun, line 242: **Finding (fixed): two of the three branches were untested — confirmed by mutation, not by reading.**

`setupStatusWithMultipleOldMaps` never set `mapImageEmailPath`, so the non-primary heal test hit the cleanup helper's `undefined` guard and returned immediately, and no test drove the no-visible-coordinates branch at all. Deleting the cleanup call and both `mapImageEmailPath: null` resets left all 4 tests green.

Fixed: the multi-map setup now seeds a copy, the non-primary test asserts the copy is deleted and the column cleared, a new test covers the no-visible-coordinates branch, and the failed-regeneration test now asserts the reference *survives* (the invariant that keeps the copy from being orphaned). Re-verified by mutation: dropping the non-primary cleanup now fails the test.

lib/services/medias/imageOutputFormat.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    expect(getImageOutputFormatDetail(format)).toEqual({
+      extension,
+      contentType
+    })
+  })
+})
+
+describe('encodeImageOutput', () => {
+  it('defaults to webp so existing uploads are unchanged', () => {
+    expect(DEFAULT_IMAGE_OUTPUT_FORMAT).toBe('webp')
```
llun, line 58: **Finding (fixed): nothing pinned the WebP encoder options — the one invariant this refactor claims to preserve.**

The tests asserted only `metadata().format` and `isProgressive`, so changing `quality: 95 → 60` or dropping `nearLossless`/`smartSubsample` would have silently re-encoded every upload in the product with the whole suite green.

Fixed by byte-comparing the helper's WebP output against a locally constructed `sharp(...).webp({ quality: 95, smartSubsample: true, nearLossless: true })` buffer, plus an assertion that the JPEG branch really emits `4:4:4` (the reason chroma subsampling is disabled: a route map is a saturated thin polyline over pale tiles). Also added the `::1` / `[::1]` cases that `getMediaFileUrl`'s localhost check carries but the test skipped.

AGENTS.md
```diff
@@ -309,6 +309,24 @@ it; there is no legacy shape left to copy.
...
   URL because the generated maps are 4:3 and any stand-in image renders the card
   a third taller than it ever will be.
+- **An email must never point an `<img src>` at a stored image path directly.**
+  The media storages write WebP unless the caller asks for another format
+  (`_saveImageBuffer` / `_uploadImageBufferToS3`), and Outlook desktop (Word
+  rendering engine) and Windows Mail have no WebP decoder — those recipients get
+  the `alt` text. So an email image needs a stored JPEG copy
+  (`saveMediaImageRendition(database, actor, file, 'jpeg')`) plus a column to
+  remember it; the route map's lives in `fitness_files.mapImageEmailPath`. Keep
+  the WebP as a **live** fallback for whenever the copy is missing — no media
```
llun, line 319: **Finding (fixed): this rule said the WebP fallback exists "only for activities imported before that column existed" — which would have licensed deleting live code.**

The actual fallback (`?? storedMap.url`) fires whenever storing the copy returns undefined: no media storage configured, over quota, or any encode/DB failure — and this PR has two dedicated tests for exactly that. For pre-column activities there is no live fallback path at all, since their email was already sent. A reviewer trusting the old wording — and `REVIEW.md` is the checklist reviewers enforce — would read `?? storedMap.url` as legacy dead code and remove the only thing that keeps a map in the email when the JPEG write fails.

Also fixed in the same bullet: "Every image the media storages write is WebP" became false three lines above the sentence describing the JPEG, and that absolute is what justified the two hardcoded `mimeType: 'image/webp'` literals this PR had to remove. Now "WebP unless the caller asks for another format", with the same corrections in `REVIEW.md`. The 308-redirect justification is also narrowed: it holds when object storage sits behind a public hostname; without one the route streams the body.

docs/fitness-file-storage.md
```diff
@@ -52,6 +52,7 @@ Important columns include:
...
... recipients saw the image's alt text instead of their route. The copy is not attached to the status and never federates. Nullable, and only written by an import that is going to email the owner — checked against `notifyOnComplete`, the instance having email configured, and the owner's activity-import email preference, so instances that send no email store no copies. An activity with no GPS data has no map at all, and activities imported before this column existed keep pointing their (already sent) email at the WebP. Because it has no `medias` row, the fitness file is its only reference: it is deleted wherever that reference is dropped — deleting the activity, reprocessing or re-importing it, and **Regenerate maps** replacing or removing the map it was made from — `scripts/maintenance/cleanupMediaStorage.ts` treats it as referenced rather than orphaned, and `scripts/backup/productionArchive.ts` archives it with the other media. Its bytes are checked against the account quota before being written but are not added to the usage counters, which are maintained alongside `medias` rows. Deleting only the _post_ keeps the copy, because the activity itself — and so the reference — survives.
```
llun: **Finding (fixed): "only written when an import actually sends an email" was not what the code did, and the deletion points I enumerated were incomplete.**

Both are now true rather than aspirational: the write is gated on the instance having email configured and the owner's preference as well as `notifyOnComplete` (see the code change), and the enumeration covers deleting the activity, reprocessing, re-importing and regenerating. The quota-metering caveat is stated here too.

While here: `docs/environment-variables.md` and `docs/features.md` described the *stored* route maps as PNGs, which read as contradicting this PR now that the stored encoding is load-bearing. The PNG is the render buffer; the stored form is WebP. Reworded.

lib/jobs/processFitnessFileJob.ts
```diff
@@ -119,6 +123,94 @@ const buildActivitySummary = (data: FitnessActivityData): string => {
...
+  } catch (error) {
+    // Best-effort: the map is already stored and attached by the time this
+    // runs, so failing here costs Outlook recipients their image and nothing
+    // else. It must never fail the import.
+    logger.warn({
+      message: 'Failed to store the route map copy for email',
+      actorId: actor.id,
+      statusId,
+      fitnessFileId,
+      error: toImportErrorMessage(error, 'Unknown route map copy error')
```
llun, line 229: **Nit (fixed): `(error as Error).message` logs `undefined` for a non-`Error` rejection.** Switched to the existing `toImportErrorMessage`, which this repo already uses for exactly this and which `importFitnessFilesJob` carries a warning comment about.

**Not changed, for the record:** the reviewer noted the pre-existing `saveMedia` dispatch in `lib/services/medias/index.ts` omits `MediaStorageType.S3Storage`, so an instance with `ACTIVITIES_MEDIA_STORAGE_TYPE=s3` gets a silent `null`. That is on `main` and unrelated to this change; the new `saveMediaImageRendition` handles all three types. Worth its own issue.

Also confirmed clean by the reviewers and worth recording: the `importStravaActivityJob` call site passes no map URL (it is the no-exportable-streams fallback note, which has no map at all — the real Strava path routes through this job), and no other email template embeds stored media (the only other `` in any email is the static `/logo-nav.png`; avatars are text monograms and custom emoji stay as `:shortcode:` text). So the route map was the only exposure.

llun (commented):
Round 1 of the sub-agent review loop: three agents over the media-storage code, the job/DB/migration/script code, and the docs + email completeness. Every finding below is posted inline where it applies. All of them are already addressed in b2130b708 — each comment records the finding and what changed, and I'm resolving the threads as I go.

Two findings were verified by mutation rather than by reading, and two came out of actually running the affected script.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -280,6 +351,20 @@ async function cleanupMediaStorage() {
 
   console.log(`   Found ${storageFiles.length} files in storage`)
 
+  // Fitness files may live inside media storage but are not this script's to
```
llun, line 428: **Finding (HIGH, fixed): this script would have deleted every fitness activity file — and round 1's fix is what made that dangerous.**

It lists the whole media root, and with no `ACTIVITIES_FITNESS_STORAGE_TYPE` configured fitness files live *inside* it (`<mediaPath>/fitness` locally, the `fitness/` key prefix on S3). Their references are `fitness_files.path`, relative to the FITNESS root, so they could never match a media reference: every `.fit`/`.gpx`/`.tcx` original and every Strava archive ZIP was reported as orphaned. `--yes` would have deleted them all, permanently breaking route data, retries and Regenerate maps.

What makes this a finding on *this* PR rather than pre-existing rot: before round 1 the `path.relative` bug meant **everything** looked orphaned, so the output was self-evidently broken. Fixing that made the list look correct while it still contained all the fitness data — and the guardrail I'd added to `docs/maintenance.md` ("if it names every file you have, stop") no longer catches it.

Fixed by skipping that namespace outright rather than translating a second path convention (which is what `productionArchive.ts` does for its media plan), and reporting the skip count. Verified against a local SQLite DB with a `.fit` and a Strava ZIP in the media root: they are now skipped and only the genuine orphan is listed.

app/api/v1/statuses/[id]/route.ts
```diff
@@ -630,6 +646,16 @@ export const DELETE = traceApiRoute(
         }
       }
 
+      for (const emailMapImage of emailMapImagesToDelete) {
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.ts:649]: **Finding (fixed, privacy): `delete_media=true` destroyed the WebP the post displayed but left the JPEG copy of the same map fetchable forever.**

I checked the reviewer's claim rather than taking it: the route map attachment is *not* matched by `isFitnessStatusAttachment` (its mediaType is `image/png`, its URL is under `/api/v1/files/medias/`, and its name carries no fitness extension), so it really is included in `mediaIdsToDelete` and really is deleted. The copy has no `medias` row, so it isn't in `status.attachments` at all and the media-manager flow could not reach it.

Failure scenario: someone deletes a post because the route starts at their front door. The WebP starts returning `media-removed.svg`; `…/route-map.jpg` keeps serving the full route, unauthenticated, with `max-age=31536000, immutable`. Their only remedy was deleting the whole activity, which is not what "delete post" implies.

Fixed here, capturing the copies while the fitness files are still linked (deleting the status only nulls their `statusId`), then deleting the objects and clearing the column. Deleting a post *without* `delete_media` still keeps both encodings, which is now what the docs say.

lib/jobs/processFitnessFileJob.ts
```diff
@@ -321,7 +448,26 @@ export const processFitnessFileJob = createJobHandle(
                 mapImagePath: getAttachmentMediaPath(storedMap.url)
               })
 
-              mapImageUrl = storedMap.url
+              // The WebP is what the post and every web surface use; the email
```
llun, line 451: **Finding (fixed): the new gate could throw inside the map `try`, dropping the email image entirely — including the WebP fallback the same round wrote into `AGENTS.md`.**

`willSendImportEmail` awaits `shouldSendEmailForNotification`, a DB read, and that `await` sits inside the try whose catch logs `'Map generation failed; continuing without route map'`. Before round 2 the expression was a plain boolean and could not throw.

Failure scenario: a transient error reading actor settings. The map was already generated, stored, attached and recorded, but the log blames map generation, control jumps past the assignment, and `mapImageUrl` stays `undefined` — so `?? storedMap.url` never applies and the email renders its stat card with no image at all. Exactly the opposite of the rule.

Fixed by catching inside the gate and failing *towards* storing the copy: the cost of guessing wrong is one small unused file, versus an Outlook recipient losing their route.

lib/services/medias/imageOutputFormat.test.ts
```diff
@@ -0,0 +1,118 @@
+import sharp from 'sharp'
```
llun, line 1: **Finding (fixed): the WebP pin I added in round 1 was insensitive to every option it claimed to pin except `nearLossless`.**

The fixture was an 8×6 **solid** red image, and the encoders produce identical bytes for a flat image whatever quality is asked for. Mutation table from the reviewer: `quality: 95 → 90` passed, `→ 20` passed, `smartSubsample: false` passed, `effort: 2` added passed. Only `nearLossless` failed. So the test added specifically to stop a silent re-encode of every upload in the product would not have stopped one.

Fixed with a deterministic non-uniform fixture (pale background plus a saturated diagonal, the shape a route map actually has) plus a guard-the-guard test asserting the fixture *reacts* to quality. Re-verified: `quality: 95 → 20` now fails.

One honest correction that came out of this — `smartSubsample` is inert under `nearLossless: true`, because near-lossless puts libwebp in lossless mode where chroma subsampling never applies. It is kept only because the encoders this replaced passed it, and the comment claiming the JPEG's `4:4:4` is "the same reason" was wrong. Both comments now say so.

lib/jobs/importFitnessFilesJob.ts
```diff
@@ -346,6 +356,7 @@ export const importFitnessFilesJob = createJobHandle(
           activityStartTime: activityData.startTime ?? null,
           hasMapData: false,
           mapImagePath: null,
+          mapImageEmailPath: null,
```
llun, line 359: **Finding (fixed): round 2's change here had zero coverage — the reviewer proved it by deleting both production edits and watching 48 tests across three suites stay green.**

This is the same gap class round 1 flagged for `regenerateFitnessMapsJob`, repeated one file over: the case the comment calls out (a file that ends up non-primary never reaches `processFitnessFileJob` to rewrite the column) is exactly the case nothing asserted, and the one where the object is orphaned permanently if the call is later dropped.

Now covered by a test that seeds a copy, re-imports, and asserts both the delete and the cleared column — verified to fail when either production line is reverted.

lib/services/fitness-files/index.ts
```diff
@@ -171,6 +185,19 @@ export const deleteFitnessFile = async (
 
       if (storageDeleted) {
         await database.deleteFitnessFile({ id: fileId })
+        await deleteEmailMapImage({
```
llun: **Two findings fixed here.**

**The row kept referencing an object it had just deleted.** `database.deleteFitnessFile` is a *soft* delete, so the row survives with `mapImageEmailPath` intact pointing at a file that is gone. `productionArchive.ts` **throws and aborts** on a referenced-but-missing file unless `--allow-missing-storage`, and `--storage-scope referenced` is the default — so after any activity deletion the next archive run would abort. The column is now cleared alongside the delete.

**Only the local branch was tested.** The byte-identical object-storage branch had no coverage; the test is now table-driven over both.

Related, also fixed: `scripts/maintenance/cleanupMediaStorage.ts` counted soft-deleted rows as live references, which meant anything the delete paths missed was pinned as "referenced" forever and could never be reclaimed. It now filters `deletedAt`.

scripts/fitness/repairStravaActivityFiles.ts
```diff
@@ -350,6 +355,18 @@ async function repairStravaActivityFiles(args = process.argv.slice(2)) {
           new DeleteObjectCommand({ Bucket: bucket, Key: s3Key })
         )
         await database.deleteFitnessFile({ id: file.id })
+        // The route map's JPEG copy for the import email has no `medias` row,
```
llun, line 358: **Finding (fixed): `--delete-missing` was a sixth reference-drop site that the round-1 `AGENTS.md` rule claimed was covered and wasn't.**

It deleted the S3 fitness object, soft-deleted the row and deleted the status, but never the email copy — so the JPEG leaked permanently. Now routed through the same `deleteEmailMapImage` helper, with the column cleared.

The `AGENTS.md` and `docs/fitness-file-storage.md` wording is corrected too: rather than a list I have to keep exhaustive by hand, the rule now states that `deleteEmailMapImage` is the one helper and every site dropping a reference must call it, then names the current sites.

lib/jobs/regenerateFitnessMapsJob.test.ts
```diff
@@ -367,6 +440,14 @@ describe('regenerateFitnessMapsJob', () => {
     expect(refreshedNonPrimary?.mapImagePath).toBeUndefined()
     expect(refreshedNonPrimary?.processingStatus).toBe('completed')
 
+    // The stray map's email copy goes with it: it has no `medias` row, so the
```
llun, line 443: **Nit (fixed): the comment on the new test misdescribed its own setup** — it said "a privacy location that swallows the whole route", but the test stubs `parseFitnessFile` to return empty coordinates; no privacy location is involved. The assertions were sound. Reworded to say it reaches the branch directly at the parse boundary.

**Accepted and not fixed — a race the reviewer traced carefully.** `regenerateFitnessMapsJob` can be enqueued alongside an in-flight `processFitnessFileJob` for the same file (a Strava webhook redelivery outside QStash's dedup window satisfies `!isNewImport && isPrimary !== false && !hasMapData`), and `fitness_import_locks` does **not** serialize them — `withImportLock` is only taken at one Strava-import site and released before this job dequeues. Either interleaving can orphan or delete a copy moments after its email went out, degrading that email's image.

The race pre-dates this PR — it can already produce duplicate map attachments — and closing it properly means extending the locking to the processing/regeneration paths, which is a bigger change than this `fix:` should carry. This PR extends its blast radius from "duplicate attachment" to "stale email image", which is a milder symptom than the race's existing one. Flagging rather than silently shipping it; happy to open a follow-up.

llun (commented):
Round 2 of the sub-agent review loop: two agents, one on the second commit's mechanics and one on cross-cutting angles (concurrency, multi-file statuses, the serving path, email escaping, rollback). Eight findings, one **high severity** and directly caused by round 1's fix. All fixed in db85d03ec except the last, which I'm accepting with reasons.

Four findings were proven by mutation rather than by reading — including that the WebP pin I added in round 1 didn't actually pin anything.

Also confirmed clean by these agents: no new import cycle (BFS from the media service reaches no fitness module), delete ordering uses the pre-update snapshot at every site, the JPEG goes through the identical `safeUrl`/`escapeHtml` path as the WebP and cannot be silently dropped, the map PNG is opaque so JPEG flattening can't blacken it, the merge/non-primary lifecycle has no hole, and rollback is safe (nullable column, unread by older code).

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -280,6 +371,19 @@ async function cleanupMediaStorage() {
...
```
llun, line 428: **Round 3 finding (fixed): my own round-2 fix put the script back into the high-severity case it was written to prevent.**

`getSharedFitnessPrefix` returned `null` when fitness storage resolves to the *same* directory as media storage (`path.relative` gives `''`, which my `!relative` guard treated as "no shared prefix"). With identical roots nothing in a storage listing distinguishes a fitness file from a media file, so every stored activity would again be reported as an orphaned media file — the exact outcome round 2 fixed, just reachable through a different configuration.

It now aborts with an explanation instead of guessing, and the check runs before the database is opened so it fails fast on what is a configuration problem. Verified both ways against a local SQLite DB: identical roots abort; the normal `<mediaPath>/fitness` layout still skips the fitness namespace and reports only the genuine orphan.

Also confirmed here: `'fitness/'` really is the default prefix in `lib/config/fitnessStorage.ts` (both the explicit and the fallback path), so the `|| 'fitness/'` default matches reality.

app/api/v1/statuses/[id]/route.ts
```diff
@@ -630,6 +646,29 @@ export const DELETE = traceApiRoute(
...
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.ts:649]: **Round 3 finding (fixed): this cleanup could turn a completed delete into a 500.**

The loop ran after `deleteStatusFromUserInput`, so the status was already gone, but an unhandled throw from `updateFitnessFileActivityData` (or anything else in the loop) would surface as a 500 — which a client reads as "the delete failed and your post is still there", when it isn't. The media loop directly above it is best-effort and logs per item; this one now matches.

The capture itself stays before deletion, where a throw is safe because nothing has been deleted yet.

lib/services/fitness-files/index.test.ts
```diff
@@ -0,0 +1,167 @@
+import { getConfig } from '@/lib/config'
```
llun, line 1: **Round 3 note: the object-storage coverage I claimed in round 2 was not actually proven, and I only found that by re-running the mutation properly.**

My first mutation script located the branch with `s.index('case FitnessStorageType.S3Storage:')`, which matches an **earlier switch in the same file** (`saveFitnessFile`/`getFitnessFile` are defined above `deleteFitnessFile`). So it had been deleting the first `deleteEmailMapImage` after *that* point — the local branch — and the companion "break the local branch" run was a no-op that mutated nothing. The result looked like the object-storage case passing while the local case failed, which is what made me look again.

Re-done by removing each of the two occurrences individually: breaking the local branch now fails only the `local file storage` case, and breaking the object-storage branch fails only the `object storage` case. Both branches are genuinely covered.

Recording this because the same trap applies to any mutation check on a file with repeated switch shapes — matching on a `case` label is not enough to locate a branch.

llun (commented):
Round 3 of the sub-agent review loop. The reviewer agent hit an account usage limit partway in, so I ran this round myself against the third commit rather than leave it unreviewed — treating my own fixes as the thing most likely to be wrong, which they were.

Two real defects in round 2's fixes, both fixed in aabf8509e, plus a correction to a verification claim I made in round 2. Gate re-run green after these: prettier → lint (0 errors) → build → 735 files / 7409 tests.

llun (commented):
Round 4 of the review loop came back **clean** — no new defects in the fourth commit. It confirmed the sentinel has exactly one caller that checks it before use, that `--help` still short-circuits ahead of the new guard, and that each root-comparison case lands on the right branch.

It left two non-blocking notes and one coverage gap; I've taken all three rather than deferring them, since two concern a script that deletes files and the third is untested code I added. Gate green: prettier → lint (0 errors) → build → 735 files / 7412 tests.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -280,6 +420,19 @@ async function cleanupMediaStorage() {
...
```
llun, line 428: **Round 4 note, taken: the guard was spelling-based, so two spellings of the same directory still reached the delete-everything branch.**

I'd already closed the case-insensitivity half after checking the edge cases by running them rather than reasoning: on macOS/Windows, media `/data/Media` and fitness `/data/media/fitness` are one directory tree, so the activity files really are in the listing, but `path.relative` reports `../media/fitness` and the script skipped nothing.

This commit closes the symlink half — both roots are resolved with `realpathSync` before comparison, falling back to the given path when the directory doesn't exist yet (realpath throws ENOENT, and a missing directory has nothing to skip anyway).

Verified live for every layout: fitness inside media → skips; case-mismatched roots → skips; symlinked roots → aborts; identical roots → aborts; fitness as a parent of media, and siblings → skip nothing, which is correct because those files are outside the media root and never appear in the listing.

app/api/v1/statuses/[id]/route.ts
```diff
@@ -630,6 +646,34 @@ export const DELETE = traceApiRoute(
...
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.ts:649]: **Round 4 note, taken: reference-first ordering.**

`updateFitnessFileActivityData` is the only statement in the block that can throw, so the previous delete-then-null order meant the new catch swallowed precisely the failure that strands a live row pointing at an already-deleted object — the state that makes `productionArchive` abort under its default `--storage-scope referenced`.

Reversed here and in `deleteFitnessFile`: a failed null now skips the delete and leaves a consistent row, and a failure after it leaves a plain orphan that `cleanupMediaStorage` reclaims, because only live rows count as references.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -3144,6 +3144,78 @@ describe('GET /api/v1/statuses/[id]', () => {
       )
     })
 
+    const createNoteWithFitnessEmailCopy = async (suffix: string) => {
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:3159]: **Round 4 note, taken: this loop had no test at all.**

The reviewer checked the 150 KB route test for `mapImageEmailPath`, `deleteEmailMapImage` and `getFitnessFilesByStatus` and found none — so the whole email-copy path, including the catch branch added in the previous commit, was untested. That is the same gap class rounds 1–3 kept turning up, in code I'd added two commits earlier.

Covered now for three cases: `delete_media=true` deletes the copy and clears the column; `delete_media` absent keeps both (matching what happens to the WebP); and a storage failure during cleanup still returns 200, because the status is already deleted by then and a 500 would read as "your post is still there". Verified to fail when the loop is removed.

llun (commented):
Round 5: one real finding — a test of mine that passed for an unrelated reason — plus one note I took anyway. Both fixed in fbf36e9d6.

Round 5 also cleared the things I was least sure of: reference-first ordering is correct at both sites, the 1:1 path-to-row claim holds, `toRealPath`'s error swallowing cannot flip a skip decision unsafely, and the other two new tests are non-vacuous.

Gate green: prettier → lint (0 errors) → build → 735 files / 7413 tests.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -3144,6 +3144,111 @@ describe('GET /api/v1/statuses/[id]', () => {
...
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:3159]: **Round 5 finding (fixed): the test I added in round 4 to prevent this exact class of bug was itself passing for the wrong reason.**

`mockRejectedValueOnce` was consumed by the **first** `deleteMediaFile` call of the request — the status's own media in the media-manager loop — so the email copy never failed. The reviewer proved it from the emitted log line naming the `.webp`, not by inference. And even with the order reversed it could not have reached the handler: `deleteEmailMapImage` swallows storage errors itself, so `updateFitnessFileActivityData` is the only statement in that block that can throw, and nothing made it fail. Removing the entire try/catch left all three tests green.

Fixed two ways: the storage failure is now injected by **path** rather than by call order, and a second test rejects the null-out to exercise the catch itself. Re-verified with the reviewer's own mutation — removing the try/catch now fails.

This is the fourth time in this PR that a test looked like coverage and wasn't, and the second time in code I'd written specifically to close a gap. The lesson I'd draw: `mockRejectedValueOnce` on a shared mock asserts nothing about *which* call failed.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -280,6 +422,19 @@ async function cleanupMediaStorage() {
...
```
llun, line 428: **Round 5 note, taken: the last length assumption is gone.**

I'd already replaced `realTo.slice(realTo.length - insensitive.length)` after finding that `'İstanbul'.toLowerCase()` is 9 characters rather than 8. The reviewer spotted the mirror image in the replacement — `realTo.slice(realFrom.length + 1)` indexes into one string using a length taken from the other, while containment was proven on the lowercased copies — and constructed the input where it returns `'/fitness'` instead of `'fitness'`.

They correctly noted it always degrades to the pre-PR behaviour and can never make a *worse* deletion decision, so it was a note rather than a defect. I took it anyway: every reasoning-based assumption in this file has turned out wrong at least once, and this is a tool with a `--yes` flag. Comparing segment by segment and rebuilding from the real-cased segments removes the question in both directions.

Verified across seven layouts, including one whose lowercase changes length and a prefix-but-not-parent case (`/data/media` vs `/data/mediaX/fitness`), which segments make structurally impossible to get wrong.

Also recorded from round 5's clean findings: `mapImageEmailPath` really is 1:1 with a row — every rendition is named from `crypto.randomBytes(8)`, so nulling one row's column can never de-reference another row's file — and `toRealPath` swallowing errors is safe, because resolving `to` requires traversing every component of `from`.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -3144,6 +3156,139 @@ describe('GET /api/v1/statuses/[id]', () => {
...
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts:3159]: **Round 6 finding (fixed): my `afterEach` reset was wrong in exactly the way its own commit message claimed it was right.**

I wrote that it reset "back to the bare `vi.fn()` the file starts with". The file does not start with a bare `vi.fn()` — the module mock is `vi.fn().mockResolvedValue(true)` (line 78), so `mockReset()` dropped it to `undefined`. From the second test in the describe onward every delete resolved `undefined`, `deleteEmailMapImage` did `undefined.catch(...)`, and the route swallowed the resulting TypeError.

So "deletes the route map email copy when delete_media is true" was passing **only because production code swallows a TypeError** — the success path of `deleteEmailMapImage` never executed. It was order-dependent too: the first test in the describe still saw `true` purely by position, and inserting anything above it would have flipped it silently.

**Second finding, same area: the ordering assertion was not testing ordering.** `deleteEmailMapImage` swallows storage errors, so `mapImageEmailPath` ends up null under *either* order — the column state cannot distinguish them. The reviewer proved it: fix the mock and reverse the two statements together, and all 129 tests pass. Ordering is now asserted on `invocationCallOrder`, and reversing the statements fails it — verified.

That is the fifth time in this PR a test looked like coverage and wasn't, and the third in code I wrote specifically to close a gap.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -280,6 +425,27 @@ async function cleanupMediaStorage() {
...
```
llun, line 428: **Round 6 observation, fixed: a third spelling route into the same catastrophe.**

The reviewer flagged this as out of scope — it predates the commit they were reviewing and was neither introduced nor worsened by it — but it is in code this PR added, and it is a deletion tool, so I took it.

The skip compared the **configured** spelling against **on-disk** names. `ACTIVITIES_FITNESS_STORAGE_PATH=…/Fitness` against a directory created as `fitness` skipped nothing on a case-insensitive filesystem, so every stored activity file was offered for deletion. They confirmed it end-to-end on real directories.

That is now three separate spellings that reached the same disaster — case-mismatched roots, symlinked roots, and configured-vs-on-disk casing — each found by running the thing rather than reading it. Prefix matching now folds case on the platforms that fold case, sharing one predicate with the containment comparison.

Re-verified live across six layouts: normal, mis-cased config, mis-cased media root, symlinked and identical roots (both abort), and a fitness root outside the media root.

Also recorded from round 6's clean findings: the segment comparison survives trailing separators, doubled separators, root paths and prefix-substring false positives; `path.win32.relative` is already case-insensitive so the branch is effectively darwin-only; and the `vi.spyOn` rejection cannot leak into a later test, because `mockRejectedValueOnce` builds the rejected promise lazily.

llun (commented):
Round 6: **not clean** — two real findings in the tests, plus one observation the reviewer explicitly scoped out that I fixed anyway because it is a deletion tool. All in 3a741daf5.

Both test findings were in code I had just written to close earlier gaps, and both were proven by mutation rather than argued. Gate green: prettier → lint (0 errors) → build → 735 files / 7414 tests.

llun:
## Review loop complete — round 7 clean

Seven rounds, stopping here per `AGENTS.md` ("repeat until a full round surfaces no new issues", max 20).

Round 7 found **no defects** and independently confirmed the things I was least sure of: `mock.calls` and `mock.invocationCallOrder` really are parallel arrays (checked against `@vitest/spy` source) and the invocation counter is module-global, so the cross-mock ordering comparison is valid; only one `.jpg` delete can exist, because `deleteMediaForAccount` returns only stored `.webp` paths; `mockReset().mockResolvedValue(true)` restores the exact original; and case-folding the prefix match can only **broaden** the skip set, which is the direction a deletion tool should err in. It also verified the ordering test by mutation independently of my own check.

One thing it raised and I'm consciously not acting on: an under-skip is constructible for a fitness directory whose name ends in a capital Greek sigma (`'ΑΣΒ'.startsWith('ΑΣ')` is true while the lowercased comparison is false). Not a credible directory name, and the failure direction is "skips less", which the guard's other layers catch.

### What the loop actually found

The email change has been stable since the first commit. **Every defect after it was in a fix**, and nearly all lived in `scripts/maintenance/cleanupMediaStorage.ts` — a file this PR only touches because the JPEG copy deliberately has no `medias` row.

Three separate spellings reached the same catastrophe in that script, each found by *running* it rather than reading it:
1. Mis-cased roots (`/data/Media` vs `/data/media/fitness`) — `path.relative` calls them unrelated.
2. Symlinked roots — same, via a different spelling.
3. Configured-vs-on-disk casing (`…/Fitness` configured, `fitness` on disk) — the prefix never matched the listing.

Plus the pre-existing `path.relative` bug on `main` that made the script report **every** media file as orphaned, which is what made the fixed version dangerous: it turned obviously-broken output into plausible-looking output that still proposed deleting all fitness data.

**Five times a test looked like coverage and wasn't**, three of them in code written specifically to close an earlier gap:
- `mockRejectedValueOnce` consumed by the wrong call.
- An ordering assertion that couldn't distinguish either ordering.
- A `mockReset()` that stripped a module mock's default, so a test passed only because production swallowed a TypeError.
- A mock implementation leaking into ~800 later tests (`vi.clearAllMocks()` does not reset implementations, and this repo sets neither `mockReset` nor `restoreMocks` globally).
- A spy restored outside a `finally`.

None were caught by reading. Each took a mutation, a live run, or a probe of mock state.

### Open by choice

- **Quota metering.** The copy's bytes are quota-*checked* but not added to the usage counters, which only exist alongside `medias` rows. Metering needs new database-interface surface; happy to add it if you'd prefer.
- **A pre-existing race.** `regenerateFitnessMapsJob` can run alongside `processFitnessFileJob` (webhook redelivery outside QStash's dedup window); `fitness_import_locks` doesn't serialize them. It already produces duplicate map attachments on `main`; this PR adds "stale email image" to the symptom list. Closing it means extending locking.

### Not verified

Delivery to a real Outlook desktop / Windows Mail inbox. Everything upstream is confirmed — the stored bytes are baseline JPEG at 4:4:4, and the email HTML carries the `.jpg` and never the `.webp` — but the actual send needs a real inbox.

PR #1333, opened 2026-07-26
fix: sanitize stored fitness file names

llun:
Stacked on #1329 — base is `claude/elegant-swartz-d1fe53`, since this PR uses the `sanitizeStoredFileName` helper that PR adds. Retarget to `main` once #1329 merges.

## Problem

`lib/services/fitness-files/` persisted the client-supplied upload file name verbatim into `fitness_files.fileName`, which is `varchar(255) not null` in both `migrations/schema.sql` and `migrations/schema.sqlite.sql`.

There was no path traversal: the storage path is fully generated, the extension comes from the `getFitnessFileType` allowlist, and local writes go through `assertFitnessStoragePath`. Two things were wrong anyway:

- **An unbounded name is an insert failure.** On PostgreSQL a name over 255 characters fails the insert, turning an otherwise valid upload into a 500.
- **The raw name kept control, zero-width and bidi characters**, and it is rendered back to users in `lib/components/settings/FitnessFileManagement.tsx` and `lib/components/settings/FitnessImport.tsx`. A U+202E before `xpg.gpx` renders the tail reversed.

AGENTS.md already carried this as a known gap under **Uploaded file names are untrusted input**.

## Change

All five sites now go through `sanitizeStoredFileName` — it reduces a name to a single inert path segment (cuts at the last `/` or `\`, strips C0/DEL/C1/zero-width/bidi, rejects `.`/`..`, caps at 200 bytes) and falls back to `file`:

- `localFile.ts` — the `createFitnessFile` record and the `FitnessStorageSaveFileOutput` payload
- `S3StorageFile.ts` — the same two in `saveFile`, plus `input.fileName` on the presigned path

### Where the helper lives

Left under `lib/services/medias/fileName.ts` rather than moved to `lib/utils/`. Both callers are server services, and `lib/services/fitness-files/` already reaches into that directory for the upload machinery it shares — `localFile.ts` and `S3StorageFile.ts` both import `checkQuotaAvailable` from `@/lib/services/medias/quota`. Following that precedent keeps the diff off #1329's in-review files. The module header now says so explicitly, and the docs describe the rule as covering both upload drivers.

### `getFitnessFileType` still reads the raw name

Deliberate, and covered by a test. The 200-byte cap can truncate a long name past its extension, and `getFitnessFileType` throws when neither the name nor the MIME type identifies a type — so sanitizing first would turn a 500-character `….gpx` upload with a generic MIME type into a hard failure. Its return is one of four literals and is the only part of a supplied name that reaches a storage path.

## Tests

`lib/services/fitness-files/localFile.test.ts` and `S3StorageFile.test.ts` gain entry-point coverage for both drivers, sharing one case table from `testUtils.ts`: POSIX and Windows directory prefixes, a NUL byte, a bidi override, the `.` and `..` fallbacks, an ordinary name left alone, the byte cap (asserted as an exact value, not a bound), type detection surviving a truncating name, and path/key containment guards.

13 fail against the pre-fix code. The rest pass either way by design: two control rows asserting an ordinary name is left alone, two path/key containment guards, and two ordering guards that lock the sanitize-*after*-detect sequence — mutation-verified that hoisting `sanitizeStoredFileName` above `getFitnessFileType` fails them, and that a catastrophic truncation (`.slice(0, 3)`) fails the cap assertions.

Full suite: 732 files, 7438 tests passing. `yarn lint` reports 0 errors, `yarn build` succeeds.

## Docs

Removes the "Known gap" bullet from `AGENTS.md` and the matching parentheticals in `REVIEW.md`, `CLAUDE.md` and `.github/copilot-instructions.md`, and widens their scope from the media drivers to both upload drivers. Adds the `varchar(255)` insert-failure rationale and a note about the deliberate raw-name type detection.

---

## Out of scope, noticed in passing

Two adjacent issues surfaced during review. Neither is introduced by this PR and neither is fixed here:

1. **`S3StorageFile.ts:293` persists `mimeType: input.contentType` unbounded.** `fitness_files.mimeType` is also `varchar(255) NOT NULL`, and its only caller validates it as `contentType: z.string().default('application/zip')` with no allowlist or length check (`app/api/v1/fitness/strava/archive/presigned/route.ts:29`). So `{"fileName":"a.zip","contentType":"x".repeat(500),"size":1}` is exactly the PostgreSQL insert failure this PR describes — in the sibling column, in the same function. The multipart routes are safe because their allowlists bound `file.type`.

2. **`PresignedArchiveInput.fileName` has no `.max(255)`**, which `REVIEW.md`'s API-routes checklist calls for on string fields backed by a sized column. This PR closes the failure mode at the storage layer, so it is defensible as-is; adding `.max(255)` would return 422 instead of silently truncating. Left alone deliberately — it would reject uploads that now succeed, which is a product call rather than a bug fix.


lib/services/medias/fileName.ts
```diff
@@ -12,13 +12,22 @@ import { MediaValidationError } from '@/lib/services/medias/errors'
 // + '../../../evil.mp4')` lands a directory above `tmpdir()`. Everything the
 // storage drivers derive from a supplied name goes through this module, so the
 // reduction happens once, at the point the name enters storage.
+//
+// `sanitizeStoredFileName` is shared with the fitness storage drivers, which
+// take the same untrusted name from `POST /api/v1/fitness-files` and its
+// presigned variant. It lives here rather than in a neutral module because
```
llun: **Factually wrong endpoint.** `/api/v1/fitness-files` has no presigned variant. The only caller of `getPresignedForSaveFileUrl` is `app/api/v1/fitness/strava/archive/presigned/route.ts:82`, via `getPresignedFitnessFileUrl` (`lib/services/fitness-files/index.ts:182`).

Should name both endpoints explicitly: `POST /api/v1/fitness-files` and `POST /api/v1/fitness/strava/archive/presigned`.

AGENTS.md
```diff
@@ -663,13 +663,13 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
-- `sanitizeStoredFileName` reduces a name to one inert path segment (cuts at the last `/` **or** `\`, drops control, C1, zero-width and bidi characters, rejects `.`/`..`, caps it at 200 bytes so it fits both `varchar(255)` and a filesystem name). Use it for anything persisted or handed to another system — the stored name is federated and rendered to other users.
+- **In the upload storage drivers (`lib/services/medias/`, `lib/services/fitness-files/`), never join, `extname`, or persist a supplied file name directly — put it through `@/lib/services/medias/fileName` first.** `File.name` and the presigned flows' `fileName` field are plain client-controlled strings: only a browser multipart upload is guaranteed to send a bare basename, and every non-browser Mastodon client (`POST /api/v1/media`, `POST /api/v2/media`, `POST /api/v1/medias/presigned`) puts whatever it likes there. The module lives under `medias/` and is shared, the same way `medias/quota` already is. Apply the same treatment to any new code that accepts an uploaded name.
```
llun: **Entry-point list wasn't widened along with the scope.** The bullet now governs both drivers but still enumerates only the three media endpoints, framed as "every non-browser Mastodon client". The two fitness entry points are missing — and `POST /api/v1/fitness/strava/archive/presigned` isn't a Mastodon endpoint, so the framing needs adjusting rather than just appending to the list.

.github/copilot-instructions.md
```diff
@@ -21,8 +21,8 @@ Always refer to and follow the guidelines in [AGENTS.md](../AGENTS.md) for all c
 
 ### Uploaded File Names
 
-- A supplied file name (`File.name`, the presigned upload's `fileName`) is untrusted client input — only browser multipart uploads send a bare basename.
-- In the media storage drivers, never join it to a path, pass it to `extname`, or persist it raw. Use `sanitizeStoredFileName`, `createMediaTempFilePath`, and `getStoredMediaExtension` from `@/lib/services/medias/fileName`.
+- A supplied file name (`File.name`, the presigned uploads' `fileName`) is untrusted client input — only browser multipart uploads send a bare basename.
+- In the upload storage drivers (`lib/services/medias/`, `lib/services/fitness-files/`), never join it to a path, pass it to `extname`, or persist it raw. Use `sanitizeStoredFileName`, `createMediaTempFilePath`, and `getStoredMediaExtension` from `@/lib/services/medias/fileName`.
```
llun: **Contradicts `AGENTS.md:671` and the module header.** This now tells you to use all three helpers in `lib/services/fitness-files/`, but `createMediaTempFilePath` and `getStoredMediaExtension` are media-specific — fitness deliberately uses `getFitnessFileType` for the extension and never builds a temp path from a supplied name.

Scope the sentence: `sanitizeStoredFileName` for any stored name, the other two for media uploads only.

REVIEW.md
```diff
@@ -59,12 +59,13 @@ change doesn't touch.
...
+- A supplied file name (`File.name`, the presigned flows' `fileName`) is
   attacker-controlled: only browser multipart uploads send a bare basename. In
-  the media storage drivers it must never be joined to a path, passed to
+  the upload storage drivers (`lib/services/medias/`,
+  `lib/services/fitness-files/`) it must never be joined to a path, passed to
   `extname`, or persisted raw — it goes through `@/lib/services/medias/fileName`
-  first. (`lib/services/fitness-files/` is a known unfixed exception; its paths
-  are safe but its stored names are not sanitized.)
+  first. `medias.originalFileName` and `fitness_files.fileName` are both
+  `varchar(255)`, so an unsanitized name is also a PostgreSQL insert failure.
```
llun: **Imprecise.** An unsanitized name isn't an insert failure — only an *over-long* one is. `AGENTS.md:667` and `CLAUDE.md:43` both say "unbounded" correctly; this line should match.

lib/services/fitness-files/S3StorageFile.test.ts
```diff
@@ -129,3 +137,136 @@ describe('S3FitnessStorage presigned upload verification', () => {
...
+
+    const output = await createStorage().saveFile(actor, { file })
+
+    expect(storedRecord().fileName).toBe(expected)
+    expect(output.fileName).toBe(expected)
+  })
+
+  // `fitness_files.fileName` is `varchar(255) not null`, so an unbounded name is
+  // an insert failure on PostgreSQL — a 500 on an otherwise valid upload.
+  it('saveFile caps an over-long name at the stored column width', async () => {
```
llun, line 204: **Missing the ordering guard that `localFile.test.ts:143` has**, despite `S3StorageFile.ts:172-176` carrying the identical justifying comment.

Mutation-verified: applying `getFitnessFileType(sanitizeStoredFileName(file.name), file.type)` to `S3StorageFile.ts` leaves all 12 S3 tests green, while the same mutation on `localFile.ts` fails. So a future refactor that hoists the sanitizer breaks S3 uploads of long-named files with a generic MIME type (`getFitnessFileType` throws) and CI stays green.

Mirror the localFile test: save `` `${'a'.repeat(500)}.gpx` `` with `type: 'application/octet-stream'` and assert `fileType === 'gpx'`.

lib/services/fitness-files/localFile.test.ts
```diff
@@ -38,3 +39,127 @@ describe('LocalFileFitnessStorage path containment', () => {
...
+    expect(stored.fileName).toBe(expected)
+    expect(output.fileName).toBe(expected)
+  })
+
+  // `fitness_files.fileName` is `varchar(255) not null`, so an unbounded name is
+  // an insert failure on PostgreSQL — a 500 on an otherwise valid upload.
+  it('caps an over-long name at the stored column width', async () => {
+    const { output, stored } = await saveFile(`${'a'.repeat(500)}.gpx`)
+
+    expect(Buffer.byteLength(stored.fileName)).toBeLessThanOrEqual(200)
```
llun: **Assertion is too weak to pin the behavior.** `toBeLessThanOrEqual(200)` is satisfied by *any* truncation — mutation-verified that `sanitizeStoredFileName(file.name).slice(0, 3)` also passes.

The exact value is trivially known: `'a'.repeat(500) + '.gpx'` sanitizes to exactly `'a'.repeat(200)`. Assert that literal instead. Same at `S3StorageFile.test.ts:228`.

That also strengthens the next line — `expect(output.fileName).toBe(stored.fileName)` compares two values derived from the same variable, so it catches a divergence between record and payload but not a shared bug, which is exactly the pre-fix state.

lib/services/fitness-files/S3StorageFile.test.ts
```diff
@@ -129,3 +137,136 @@ describe('S3FitnessStorage presigned upload verification', () => {
...
+
+    expect(storedRecord().path).toMatch(
+      /^\d{4}-\d{2}-\d{2}\/[0-9a-f]{16}\.gpx$/
+    )
+    expect(vi.mocked(PutObjectCommand).mock.calls[0][0].Key).toBe(
+      storedRecord().path
+    )
+  })
+
+  it.each([
```
llun, line 237: **Presigned table doesn't cover the byte cap.** It covers directory-stripping and the `file` fallback, but the presigned flow is the one that takes `fileName` as a plain JSON request field, and the `varchar(255)` insert failure is a stated motivation for this PR. Worth one more row with a 500-character name.

lib/services/fitness-files/localFile.test.ts
```diff
@@ -38,3 +39,127 @@ describe('LocalFileFitnessStorage path containment', () => {
...
+  const saveFile = async (fileName: string, type = 'application/gpx+xml') => {
+    const file = new File([Buffer.from('<gpx/>')], fileName, { type })
+    const output = await createStorage().saveFile(actor, { file })
+    return {
+      output,
+      stored: vi.mocked(database.createFitnessFile).mock.calls[0][0]
+    }
+  }
+
+  it.each([
```
llun: **These two tables have already drifted apart within this PR.** `localFile.test.ts:93-123` and `S3StorageFile.test.ts:182-207` are near-copies: localFile has a bidi-override row that S3 lacks, and S3 uses `'.'` for the fallback row where localFile uses `'..'`. Neither driver has any reason to behave differently.

Hoist one shared fixture array so both drivers are held to the same table — the drift happening *inside a single PR* is evidence the duplication is actively harmful.

llun (commented):
Round 1 of the sub-agent code-review loop: three reviewers (correctness, tests, docs). Correctness came back clean — it traced all three `createFitnessFile` writers, every consumer of `fitness_files.fileName`, and reproduced the raw-name rationale. Findings below are from the tests and docs reviewers.

lib/services/medias/fileName.ts
llun: Fixed in 6c4300ef — now names both fitness entry points explicitly: `POST /api/v1/fitness-files` and `POST /api/v1/fitness/strava/archive/presigned`. You were right that no presigned variant of `/api/v1/fitness-files` exists.

AGENTS.md
llun: Fixed in 6c4300ef. Split the sentence so the Mastodon framing stays attached to the media endpoints, and added the two fitness entry points separately.

.github/copilot-instructions.md
llun: Fixed in 6c4300ef. Now scopes `sanitizeStoredFileName` to any stored name, keeps `createMediaTempFilePath`/`getStoredMediaExtension` as media-only, and states that fitness takes its extension from `getFitnessFileType`.

REVIEW.md
llun: Fixed in 6c4300ef — "over-long", matching the wording in `AGENTS.md:667` and `CLAUDE.md:43`.

lib/services/fitness-files/S3StorageFile.test.ts
llun: Fixed in 6c4300ef. Added `saveFile still detects the file type for a name the byte cap truncates` to the S3 driver. Re-ran your mutation to confirm it now bites: hoisting the sanitizer into `getFitnessFileType(sanitizeStoredFileName(file.name), file.type)` fails this test, where it previously left all 12 S3 tests green.

lib/services/fitness-files/localFile.test.ts
llun: Fixed in 6c4300ef. Both drivers now assert the exact truncated value via the shared `OVER_LONG_FITNESS_FILE_NAME_TRUNCATED` constant. Verified with your mutation: `sanitizeStoredFileName(file.name).slice(0, 3)` now fails the cap test, where it previously passed.

lib/services/fitness-files/S3StorageFile.test.ts
llun: Fixed in 6c4300ef — added a 500-character `.zip` row to the presigned table.

lib/services/fitness-files/localFile.test.ts
llun: Fixed in 6c4300ef. Hoisted the table into `lib/services/fitness-files/testUtils.ts` (following the `lib/database/testUtils.ts` convention) as `STORED_FITNESS_FILE_NAME_CASES`, and folded in both drifted rows — the bidi-override case plus separate `.` and `..` fallback rows — so each driver runs all seven.

AGENTS.md
```diff
@@ -663,13 +663,13 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
-- `sanitizeStoredFileName` reduces a name to one inert path segment (cuts at the last `/` **or** `\`, drops control, C1, zero-width and bidi characters, rejects `.`/`..`, caps it at 200 bytes so it fits both `varchar(255)` and a filesystem name). Use it for anything persisted or handed to another system — the stored name is federated and rendered to other users.
+- **In the upload storage drivers (`lib/services/medias/`, `lib/services/fitness-files/`), never join, `extname`, or persist a supplied file name directly — put it through `@/lib/services/medias/fileName` first.** `File.name` and the presigned flows' `fileName` field are plain client-controlled strings: only a browser multipart upload is guaranteed to send a bare basename. Every non-browser Mastodon client (`POST /api/v1/media`, `POST /api/v2/media`, `POST /api/v1/medias/presigned`) puts whatever it likes there, and so do the fitness uploads (`POST /api/v1/fitness-files`, `POST /api/v1/fitness/strava/archive/presigned`). The module lives under `medias/` and is shared, the same way `medias/quota` already is. Apply the same treatment to any new code that accepts an uploaded name.
```
llun: **The fitness entry-point list reads as exhaustive but isn't.** Two more HTTP routes push a client-supplied `File.name` into `saveFitnessFile`:

- `app/api/v1/fitness/import/route.ts:90` — multipart `files` (a `getAll('files')` loop)
- `app/api/v1/fitness/strava/archive/route.ts:352` — multipart `archive`

Nothing is unsafe — they all funnel through the same two sanitizing drivers — but an agent reading this as the complete set of untrusted-name entry points would be wrong. Same omission in the `lib/services/medias/fileName.ts` header comment.

lib/services/fitness-files/S3StorageFile.test.ts
```diff
@@ -129,3 +142,126 @@ describe('S3FitnessStorage presigned upload verification', () => {
...
+      description: 'falls back when the name reduces to nothing usable',
+      fileName: `${String.fromCharCode(0x200b)}   `,
+      expected: 'file'
+    },
+    {
+      // This is the flow that takes the name as a plain JSON request field,
+      // so it is the likeliest source of an over-long one.
+      description: 'caps an over-long name at the stored column width',
+      fileName: `${'a'.repeat(500)}.zip`,
+      expected: 'a'.repeat(200)
```
llun: **Re-hardcodes the constant this commit just centralized.** `expected` here is character-for-character `OVER_LONG_FITNESS_FILE_NAME_TRUNCATED`, so the 200 now lives in two places and a future cap change updates only one — exactly the drift the shared-fixture refactor was meant to eliminate.

The *input* legitimately can't reuse `OVER_LONG_FITNESS_FILE_NAME` (this route requires `.zip`), but `expected` should reference the constant.

lib/services/medias/fileName.ts
```diff
@@ -12,13 +12,23 @@ import { MediaValidationError } from '@/lib/services/medias/errors'
 // + '../../../evil.mp4')` lands a directory above `tmpdir()`. Everything the
 // storage drivers derive from a supplied name goes through this module, so the
 // reduction happens once, at the point the name enters storage.
+//
+// `sanitizeStoredFileName` is shared with the fitness storage drivers, which
+// take the same untrusted name from `POST /api/v1/fitness-files` and
+// `POST /api/v1/fitness/strava/archive/presigned`. It lives here rather than
+// in a neutral module because
```
llun: Cosmetic: the round-1 edit left this as a 4-word orphan line. Prettier doesn't reflow comments, so it stays until reflowed by hand.

lib/services/fitness-files/S3StorageFile.test.ts
```diff
@@ -129,3 +142,126 @@ describe('S3FitnessStorage presigned upload verification', () => {
...
+  it('saveFile caps an over-long name at the stored column width', async () => {
+    const output = await saveFile(OVER_LONG_FITNESS_FILE_NAME)
+
+    expect(storedRecord().fileName).toBe(OVER_LONG_FITNESS_FILE_NAME_TRUNCATED)
+    expect(output.fileName).toBe(OVER_LONG_FITNESS_FILE_NAME_TRUNCATED)
+  })
+
+  // The type comes from the raw name because sanitizing can truncate a long
+  // name past its extension, and `getFitnessFileType` throws when neither the
+  // name nor the MIME type identifies a type. This pins the ordering — hoisting
```
llun, line 213: Cosmetic: this ordering-guard comment gained a sentence its twin at `localFile.test.ts:117-119` didn't. Minor, but the whole rationale for round 2's shared fixture was that these two files drift — worth keeping the comments in step too.

llun (commented):
Round 2: reviewed the round-1 follow-up commit `6c4300ef`. **No behavior bugs** — the new `testUtils.ts` isn't bundled into the build (verified against `.next`) and doesn't trip `clientModuleBoundary`; the exact-value cap assertions are genuinely exact (mutation-verified by changing `MAX_STORED_FILE_NAME_BYTES` 200→199, which the old `<= 200` bound would not have caught); the new S3 ordering guard fails under exactly the intended mutation and nothing else; and the `saveFile`/`storedRecord()` refactor weakens no assertion. Four minor findings below.

AGENTS.md
llun: Fixed in a0cffb56. AGENTS.md now lists all four fitness upload routes, and the `fileName.ts` header was reworded to "every fitness upload route" with the four named, so neither reads as a partial list any more. Confirmed `app/api/v1/fitness/import/route.ts:90` and `app/api/v1/fitness/strava/archive/route.ts:352` both reach `saveFitnessFile` — safe, but they belonged in the list.

lib/services/fitness-files/S3StorageFile.test.ts
llun: Fixed in a0cffb56 — `expected` now references `OVER_LONG_FITNESS_FILE_NAME_TRUNCATED`. The input keeps its literal since the route requires `.zip`, with a comment noting the cap drops the extension either way, so the shared expectation still applies.

lib/services/medias/fileName.ts
llun: Fixed in a0cffb56 — comment reflowed.

lib/services/fitness-files/S3StorageFile.test.ts
llun: Fixed in a0cffb56 — copied the ordering sentence onto the local driver's twin at `localFile.test.ts` so both read identically.

llun:
Round 3 (convergence check) — **clean, no new actionable findings**. The review loop is closed at 3 rounds, well under the 20-round cap; all 12 threads from rounds 1 and 2 are replied-to and resolved.

Round 3 verified independently rather than taking the earlier rounds on trust:

- **Enumerated the write path bottom-up.** `createFitnessFile` has exactly three call sites (`localFile.ts:165`, `S3StorageFile.ts:202`, `S3StorageFile.ts:290`), `lib/database/sql/fitnessFile.ts:350` is the sole SQL write, and no `updateFitnessFile*` touches `fileName`. All three are sanitized — **no remaining unsanitized write**.
- **Confirmed the four-route list is exactly right**, no fifth route. `LocalFileFitnessStorage.getPresignedForSaveFileUrl` returns `null` without creating a row, so there is one presigned write site and it is covered.
- **Adversarial pass on the production change**: `sanitizeStoredFileName` never throws, so it adds no upload failure mode, and the `file` fallback satisfies `NOT NULL`. Every downstream consumer is display-only — render, a `localeCompare` sort, and `${fileName} route map` description strings. No path, S3 key, or header is built from the stored name; there is no `Content-Disposition` sink for fitness files, `PutObjectCommand` carries no name metadata, and `verifyPresignedUpload` keys off `path`.
- **Cross-document consistency** across `AGENTS.md` / `REVIEW.md` / `CLAUDE.md` / `.github/copilot-instructions.md` / the module header / the code, with no residue of the old "known gap" framing anywhere in `*.md` or `docs/`.

Local gate on the final commit: prettier clean, `yarn lint` 0 errors, `yarn build` compiles, **732 files / 7438 tests passing**.

Two items deliberately left for a human, both noted in the PR description:

1. This PR targets `claude/elegant-swartz-d1fe53` and **needs retargeting to `main` once #1329 merges**.
2. `getPresignedForSaveFileUrl` also persists `mimeType: input.contentType` unbounded into a `varchar(255)` column — the same insert failure, two lines away — but it is pre-existing and outside this PR's scope, so it is flagged rather than folded in.

PR #1334, opened 2026-07-26
fix: stop upscaling stored images to the 4000x4000 cap

llun:
## Summary

`sharp`'s `fit: 'inside'` **enlarges by default** (`withoutEnlargement` is
false), so the image pipeline in both storage drivers

```ts
sharp(buffer).resize(MAX_WIDTH, MAX_HEIGHT, { fit: 'inside' }).rotate()
```

was not applying a 4000×4000 cap — it was an **upscale**. Every image the server
wrote below 4000×4000 was blown up to fill the box.

This adds `withoutEnlargement: true` through a single shared
`STORED_IMAGE_RESIZE_OPTIONS` constant, so the two duplicated call sites
(`lib/services/medias/localFile.ts`, `lib/services/medias/S3StorageFile.ts`)
cannot drift apart again. Images **above** the cap are still scaled down exactly
as before.

Noticed while adding the JPEG twin of the route map for the import email
(#1332), which deliberately kept the twin on the same pipeline rather than
widening that PR's scope.

## Measured impact

Stored WebP for representative sources, before vs. after. Measured on detailed,
non-flat content — a flat fill compresses to nothing at any size and would
understate this badly:

| Source | Before | After | Ratio |
| --- | --- | --- | --- |
| 64×64 custom emoji | 4000×4000, 1,040,740 B | 64×64, 4,144 B | 251× |
| 400×400 avatar | 4000×4000, 4,248,906 B | 400×400, 133,998 B | 32× |
| 800×600 route map | 4000×3000, 276,096 B | 800×600, 11,862 B | 23× |
| 1600×1200 photo | 4000×3000, 2,862,224 B | 1600×1200, 1,546,542 B | 1.9× |
| 3024×4032 photo (over the cap) | 3000×4000, unchanged | 3000×4000, unchanged | 1× |

## Scope: which uploads this affects

Only the **server-side write** path (`saveMedia` / `saveMediaThumbnail`) runs
sharp at all. That covers the sync upload endpoint, the Mastodon-API multipart
avatar/header handler, custom emojis, thumbnail replacement, and the fitness
import jobs — both the route maps the server generates and the arbitrary-sized
activity photos pulled from Strava.

The **presigned direct-to-S3** path never touched sharp and is unchanged. On an
object-storage instance that is what the web composer and the Settings/Account
image pickers actually use, so this fix does not reach them. `docs/architecture.md`
now documents that split, since it was not written down anywhere.

## Behaviour changes

- **New uploads on the server-side path** are stored at their own dimensions
  when they fit the cap.
- **Thumbnails** (`saveThumbnail`, and `saveFile` with a thumbnail) record
  `bytes`/`metaData` from `outputInfo` — the real stored file — so `meta.small`
  stops reporting upscaled dimensions and the per-account quota counter stops
  being charged for them.
- **`meta.original` is unchanged.** It is read from the *input* image, so it
  always reported the source dimensions. What changes is that the stored file
  now agrees with it.
- **Existing stored media is untouched.** Nothing re-encodes it. Re-encoding
  stored user media would rewrite paths, bytes and the quota counter for no
  display benefit, so this PR does not attempt it. Note the recorded row was
  never inflated for originals — only the file on disk was — so an affected
  attachment serves a file several times larger than the size it advertises, and
  a query for oversized rows will not find it. That is now written down in
  `docs/architecture.md`.

## Tests

New regression tests drive both storage drivers end to end and assert the
**actual stored bytes**, not the recorded metadata — the local driver reads the
file back out of its media root, the S3 driver drains the `PutObjectCommand`
body. Coverage per driver: below-cap (stored at source dimensions), above-cap
(still scaled down to fit), and the thumbnail path, which is where the upscale
reached the database.

All five below-cap assertions fail without `withoutEnlargement`, with the exact
upscale signature (`4000x3000` where `800x600` is expected). The two above-cap
cases stay green, so they independently guard the cap itself.

No pre-existing test asserted post-resize dimensions, so nothing else needed
updating.

## Verification

- [x] `yarn run prettier --write .`
- [x] `yarn lint` — 0 errors (31 pre-existing warnings, none in touched files)
- [x] `yarn build`
- [x] `yarn test` — 732 files, 7376 tests passing

No UI change, so no screenshots: this alters bytes on disk, not rendering.

## Follow-ups (deliberately not in this PR)

- `medias.original.bytes` is the **input file's** size while the stored file is
  the re-encoded WebP, so the quota counter for originals has never matched
  actual disk usage in either direction.
- `medias.original.metaData` is read from the input image, so an upload **above**
  the cap still reports pre-downscale dimensions (an 8000×4000 upload stores
  4000×2000 but reports 8000×4000). The thumbnail path already uses `outputInfo`
  for exactly this reason.
- `saveMedia` in `lib/services/medias/index.ts` omits `MediaStorageType.S3Storage`
  from its switch and returns `null` for it, unlike every other function in that
  file. On an instance configured with the documented `s3` type, every
  server-side image save silently no-ops.
- `S3FileStorage.saveFile`'s image branch ignores `media.thumbnail` entirely,
  while `LocalFileStorage` handles it — a client-supplied thumbnail is dropped on
  object storage.

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -221,6 +221,18 @@ change doesn't touch.
   is separately `vi.mock`'d. (Some review bots wrongly flag `vi.importMock` as
   non-existent — it is a valid, documented Vitest API.)
 
+## Stored media
+
+- Stored-image resizes go through `STORED_IMAGE_RESIZE_OPTIONS`
+  (`lib/services/medias/constants.ts`), never an inline `{ fit: 'inside' }`.
+  sharp's `fit: 'inside'` **enlarges** by default, so a bare
+  `MAX_WIDTH`/`MAX_HEIGHT` box is an upscale, not a cap — it inflates every
+  stored image below the cap, silently, with no error and no test failure.
+- `original.metaData`/`original.bytes` describe the **uploaded** file, while
+  `thumbnail.*` describes the **stored** WebP (`outputInfo`). Know which one a
+  change reads: only the latter moves when the encode pipeline changes, and only
+  the latter feeds `meta.small`.
+
 ## Docs hygiene
 
 - `docs/` is durable, general-purpose reference only. No implementation plans,
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -160,14 +160,22 @@ Media files (images and video) and fitness files (.fit, .gpx, .tcx) support mult
 - **S3** — Amazon S3
 - **Object storage** — Any S3-compatible service (MinIO, DigitalOcean Spaces, Cloudflare R2, etc.)
 
-Uploaded images (originals, thumbnails, and generated fitness route maps) are
-re-encoded to WebP and bounded by a 4000x4000 pixel box. That box is a **cap,
-not a target**: an image already inside it is stored at its own dimensions and
-is never upscaled. Images stored before this became true were enlarged to fill
-the box, so older rows can be several times larger than their source and carry
-dimensions the source never had. They are still served correctly and are left
-as-is — nothing re-encodes them — so an instance's media directory may hold a
-mix of both until those attachments are deleted.
+Every stored image — uploaded originals, their thumbnails, avatars, headers,
+custom emojis, and generated fitness route maps — is re-encoded to WebP and
+bounded by a 4000x4000 pixel box. That box is a **cap, not a target**: an image
+already inside it is stored at its own dimensions and is never upscaled.
+
+Images that predate that rule were enlarged on disk to fill the box. The
+`medias` row was not: `original.metaData` and `original.bytes` are read from the
+uploaded file, so they always described the source. An affected attachment
+therefore **serves a file several times larger than the dimensions and byte
+count it advertises**, and per-account storage usage under-counts it — a query
+for oversized rows will not find these. Only thumbnails recorded the enlarged
+numbers (`thumbnail.metaData`/`thumbnail.bytes`, surfaced as `meta.small`), so
+those over-counted usage instead.
+
+Nothing re-encodes existing media, so an instance's storage keeps both shapes
+until the affected attachments are deleted.
 
 ## Database Schema (Simplified)
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -167,10 +167,13 @@ of the surface the user is on — the same picker can take either.
 **Server-side write** (`saveMedia` / `saveMediaThumbnail` in
 `lib/services/medias/`) re-encodes to WebP and bounds the result by a 4000x4000
 pixel box. That box is a **cap, not a target**: an image already inside it is
-stored at its own dimensions and is never upscaled. This runs for the sync
-upload endpoint, the Mastodon-API multipart avatar/header
-(`PATCH /api/v1/accounts/update_credentials`), custom emojis, and the fitness
-route maps the server generates itself.
+stored at its own dimensions and is never upscaled. This runs whenever the
+server itself holds the bytes: the sync upload endpoint, the Mastodon-API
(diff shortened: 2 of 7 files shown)


llun:
## Sub-agent code review record

Per `AGENTS.md` → **Code Review Loop (Sub-Agents)**. Five rounds, converged on a
clean round. Every finding was addressed on the branch in the commit named
beside it.

### Round 1 — correctness / tests / conventions (3 agents)

The core fix cleared on correctness. Three findings, all addressed in
`9daabd195`:

1. **`docs/architecture.md` was factually wrong** — it claimed older *rows*
   carried inflated dimensions and byte counts. They did not: `original.metaData`
   and `original.bytes` are read from the uploaded file, so the row always
   described the source while only the file on disk was enlarged. An operator
   querying for oversized rows would have found nothing and wrongly concluded the
   instance was unaffected. Rewritten to state what actually diverged.
2. **The `meta.original` assertions in both new tests could not fail** —
   verified empirically: that field is input-derived, so it read `800x600`
   before the fix too. Replaced with coverage of the **thumbnail** path, which is
   where the upscale genuinely reached the database (`thumbnail.metaData`/`bytes`
   come from `outputInfo` and surface as `meta.small` and in the quota counter).
3. **Missing convention record** — added a `REVIEW.md` entry, per
   `AGENTS.md` → Documentation Maintenance. Deliberately *not* added to
   `AGENTS.md` or `CLAUDE.md`: there are exactly two `.resize()` call sites, they
   now share one constant carrying the rationale, so the invariant is enforced by
   construction rather than by prose.

Also collapsed the below/above-cap cases to a table-driven `it.each`, made the
local driver's file lookup pick original vs. thumbnail explicitly, and renamed
the constant to `STORED_IMAGE_RESIZE_OPTIONS` (the browser-side cap in
`lib/utils/resizeImage.ts` is a separate mechanism).

### Round 2 — round-1 fixes + silent-failure sweep (2 agents)

Tests and `REVIEW.md` cleared. One finding, addressed in `4b1e8c1ff`:

- **The docs overstated the guarantee.** "Every stored image … is re-encoded to
  WebP and bounded by a 4000x4000 box" is false for the presigned direct-to-S3
  path, which stores client bytes as-is — and that is the path the web composer
  prefers whenever object storage is configured. Scoped the claim to the
  server-side write path.

Two pre-existing bugs on adjacent paths were surfaced and **left out of this
PR's scope**; both are listed as follow-ups in the description.

### Round 3 — final docs verification (1 agent)

One finding, addressed in `cec6af620`:

- **Avatars and headers were listed on the wrong path.** The Settings/Account
  image pickers call `uploadAttachment` — the same helper the composer uses —
  which prefers the presigned route and submits only a URL string to the profile
  route. `saveMedia` runs for those images only via the Mastodon-API multipart
  endpoint. Enumerating *surfaces* is what kept getting this wrong, so the
  section now describes the two routes by *mechanism*.

### Round 4 — enumeration completeness (1 agent)

One finding, addressed in `7dc9b25ea`:

- **The server-side enumeration was incomplete.** The Strava import jobs put
  arbitrary-sized user activity photos through `saveMedia`, not just
  server-generated route maps — those photos are exactly the images the enlarging
  box was upscaling. Also added the `saveMediaThumbnail` caller
  (`PUT /api/v1/media/:id`) and the second route the avatar/header handler serves.

### Round 5 — convergence check (1 agent)

**Clean.** Every `saveMedia`/`saveMediaThumbnail` caller is represented by the
enumeration and nothing is overclaimed; no defects in the production code, tests,
or `REVIEW.md`. Loop terminated at 5 of a maximum 20 rounds.

### Gate

`yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` →
`yarn test` (732 files, 7376 tests) all green on the final commit.

PR #1337, opened 2026-07-28
fix: save media on s3 storage type

llun:
## Summary

`saveMedia`'s switch on `mediaStorage?.type` in `lib/services/medias/index.ts` handled only `MediaStorageType.LocalFile` (`'fs'`) and `MediaStorageType.ObjectStorage` (`'object'`), then fell through to `default: return null`. It omitted `MediaStorageType.S3Storage` (`'s3'`) — a documented, supported value: it appears in `docs/environment-variables.md` and `docs/setup.md`, and `MediaStorageS3Config` in `lib/config/mediaStorage.ts` accepts both `'object'` and `'s3'`.

`saveMedia` was the **only** function in that module with the gap. `saveMediaThumbnail`, `saveMediaImageRendition`, `getPresignedUrl`, `completePresignedMediaUpload`, `getMedia`, and `deleteMediaFile` all list `S3Storage` alongside `ObjectStorage`. I swept the rest of the repo for the same pattern — `lib/services/fitness-files/index.ts`, `scripts/maintenance/cleanupMediaStorage.ts`, and `scripts/backup/productionArchive.ts` all pair the two types correctly.

## Impact

On an instance configured with `ACTIVITIES_MEDIA_STORAGE_TYPE=s3`, every server-side image save silently returned `null`:

- `lib/jobs/processFitnessFileJob.ts` logged `Failed to store generated route map image` and still marked the import `'completed'`, so every fitness post permanently lost its map with no user-visible signal.
- `lib/services/accounts/updateCredentialsHandler.ts` does `if (saved) …`, so an avatar/header change returned HTTP 200 while silently keeping the old image.
- `lib/services/medias/handleSyncMediaUpload.ts` returned a bare 422 "Validation failed" for what was actually a server misconfiguration.
- `app/api/v1/admin/custom_emojis/route.ts` dropped custom emoji uploads.
- `lib/jobs/importStravaArchiveJob.ts`, `lib/jobs/importStravaActivityJob.ts`, and `lib/jobs/regenerateFitnessMapsJob.ts` lost their stored media.

Deployments on `'fs'` or `'object'` were unaffected, which is why this went unnoticed.

## Changes

**`lib/services/medias/index.ts`** — one line, falling through to the same `S3FileStorage` branch as `ObjectStorage`, matching every other function in the file.

**`lib/services/medias/index.test.ts`** — the omission survived because the test file only exercised all three storage types for `deleteMediaFile`. Replaced the duplicated per-storage blocks with a delegation matrix: `describe.each` over all seven exported functions x `it.each` over all three storage types. Each case asserts that the expected driver's `getStorage` was called with `(mediaStorage, host, database)`, that the right method received the right arguments, that the return value passes through, and — the part that catches an omission — that the *other* driver was never touched.

`getPresignedUrl` / `completePresignedMediaUpload` declare only the S3 types as supported, so local file storage is asserted to hit the fallback rather than silently counting as coverage. Each function also gets a no-storage-configured case with its own fallback value (`null`, or `false` for `deleteMediaFile`).

This replaced four hand-duplicated `deleteMediaFile` blocks that were near-identical pass-through assertions against a mocked driver (the "returns true when file does not exist (ENOENT)" case never reached any filesystem). The two genuinely distinct ones — false pass-through and path-unchanged — are kept. The now-unused `fs` and `@aws-sdk/client-s3` mocks went with them, since both storage modules are auto-mocked and neither was ever reachable.

## Verification

Confirmed the new matrix actually catches the bug: stashing the source fix and re-running produces exactly one failure, `saveMedia > delegates to 's3 storage'` — expected `{delegated: true}`, received `null`.

Definition of Done gate, in order:

- `yarn run prettier --write .` — no files changed beyond the two touched
- `yarn lint` — 0 errors (31 pre-existing warnings, none in `medias/`)
- `yarn build` — passed
- `yarn test` — 737 files, 7590 tests, all passing (30 in this file, up from 15)

No docs are made stale — this restores the behavior `docs/environment-variables.md` and `docs/setup.md` already describe. No migration, and no UI change, so no browser verification applies.

Found during review of #1334, which deliberately did not widen its scope; re-verified as still valid after #1332 merged.

### DIFF
--- a/lib/services/medias/index.test.ts
+++ b/lib/services/medias/index.test.ts
@@ -8,16 +8,6 @@ import * as mediaStorageService from './index'
 import * as LocalFileStorage from './localFile'
 import { MediaSchema, PresigedMediaInput } from './types'
 
-const {
-  completePresignedMediaUpload,
-  deleteMediaFile,
-  getMedia,
-  getPresignedUrl,
-  saveMedia,
-  saveMediaImageRendition,
-  saveMediaThumbnail
-} = mediaStorageService
-
 vi.mock('@/lib/config')
 vi.mock('@/lib/utils/logger', () => ({
   logger: {
@@ -113,77 +103,87 @@ const S3_STORAGE_TYPES = [
 const NON_DELEGATING_EXPORTS = ['PresignedUploadValidationError']
 
 interface MediaStorageFunctionCase {
-  name: string
+  name: keyof typeof mediaStorageService
   /** Method on the resolved storage driver the function must delegate to. */
   method: keyof StorageMock
-  expectedArguments: unknown[]
+  /**
+   * Arguments after `database`. Every function in this module hands these
+   * straight to its driver method, so the harness both invokes the export with
+   * them and asserts the driver received them. A future function that reshapes
+   * its arguments fails loudly here rather than needing a silent second field.
+   */
+  callArguments: unknown[]
   /** Storage types the function delegates for; the rest fall back. */
   supportedTypes: MediaStorageType[]
   /** Value returned when no driver handles the configured storage type. */
   fallbackValue: null | false
-  call: () => Promise<unknown>
 }
 
 const MEDIA_STORAGE_FUNCTIONS: MediaStorageFunctionCase[] = [
   {
     name: 'saveMedia',
     method: 'saveFile',
-    expectedArguments: [actor, media],
+    callArguments: [actor, media],
     supportedTypes: ALL_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => saveMedia(mockDatabase, actor, media)
+    fallbackValue: null
   },
   {
     name: 'saveMediaThumbnail',
     method: 'saveThumbnail',
-    expectedArguments: [actor, file],
+    callArguments: [actor, file],
     supportedTypes: ALL_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => saveMediaThumbnail(mockDatabase, actor, file)
+    fallbackValue: null
   },
   {
     name: 'saveMediaImageRendition',
     method: 'saveImageRendition',
-    expectedArguments: [actor, file, 'jpeg'],
+    callArguments: [actor, file, 'jpeg'],
     supportedTypes: ALL_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => saveMediaImageRendition(mockDatabase, actor, file, 'jpeg')
+    fallbackValue: null
   },
   {
     name: 'getPresignedUrl',
     method: 'getPresigedForSaveFileUrl',
-    expectedArguments: [actor, presignedInput],
+    callArguments: [actor, presignedInput],
     // Presigned uploads go straight to the bucket, so local file storage has
     // nothing to sign.
     supportedTypes: S3_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => getPresignedUrl(mockDatabase, actor, presignedInput)
+    fallbackValue: null
   },
   {
     name: 'completePresignedMediaUpload',
     method: 'completePresignedUpload',
-    expectedArguments: [actor, 'media-1'],
+    callArguments: [actor, 'media-1'],
     supportedTypes: S3_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => completePresignedMediaUpload(mockDatabase, actor, 'media-1')
+    fallbackValue: null
   },
   {
     name: 'getMedia',
     method: 'getFile',
-    expectedArguments: ['medias/test.jpg'],
+    callArguments: ['medias/test.jpg'],
     supportedTypes: ALL_STORAGE_TYPES,
-    fallbackValue: null,
-    call: () => getMedia(mockDatabase, 'medias/test.jpg')
+    fallbackValue: null
   },
   {
     name: 'deleteMediaFile',
     method: 'deleteFile',
-    expectedArguments: ['medias/test.jpg'],
+    callArguments: ['medias/test.jpg'],
     supportedTypes: ALL_STORAGE_TYPES,
-    fallbackValue: false,
-    call: () => deleteMediaFile(mockDatabase, 'medias/test.jpg')
+    fallbackValue: false
   }
 ]
+
+/**
+ * Invokes the live export named by the row. Looking the function up by `name`
+ * rather than capturing it in a per-row closure is what keeps a copy-pasted row
+ * honest: a row labelled one function cannot end up exercising another.
+ */
+const invoke = (mediaFunction: MediaStorageFunctionCase) => {
+  const exported = mediaStorageService[mediaFunction.name] as (
+    ...args: unknown[]
+  ) => Promise<unknown>
+  return exported(mockDatabase, ...mediaFunction.callArguments)
+}
 
 const configureStorage = (mediaStorage: unknown) => {
   mockGetConfig.mockReturnValue({
@@ -212,10 +212,10 @@ describe('Media Storage Service', () => {
   })
 
   // The delegation matrix below only protects what it enumerates, and an
-  // `it.each([])` registers zero tests without failing. Without these guards the
-  // matrix can silently shrink: dropping a STORAGE_CONFIGS entry, adding a
-  // fourth MediaStorageType, or exporting an eighth function all leave a green
-  // suite that no longer covers the case this file exists to catch.
+  // `it.each([])` registers zero tests without failing, so the matrix can
+  // silently shrink: exporting an eighth function, dropping a STORAGE_CONFIGS
+  // entry, or adding a fourth MediaStorageType would each leave a green suite
+  // that no longer covers the case this file exists to catch.
   describe('matrix coverage', () => {
     it('describes every function exported from the module', () => {
       const exported = Object.keys(mediaStorageService).filter(
@@ -237,12 +237,14 @@ describe('Media Storage Service', () => {
       expect(exercised.sort()).toEqual(Object.values(MediaStorageType).sort())
     })
 
-    it('declares at least one supported storage type per function', () => {
-      const withoutSupportedTypes = MEDIA_STORAGE_FUNCTIONS.filter(
-        (mediaFunction) => mediaFunction.supportedTypes.length === 0
-      ).map((mediaFunction) => mediaFunction.name)
-
-      expect(withoutSupportedTypes).toEqual([])
+    // Without this, a new enum member is auto-classified as unsupported: adding
+    // it to STORAGE_CONFIGS alone turns the previous guard green again while
+    // every function merely asserts that returning null for it is correct —
+    // which is precisely the silent-null bug this file exists to catch.
+    it('treats every configurable storage type as supported by default', () => {
+      expect([...ALL_STORAGE_TYPES].sort()).toEqual(
+        Object.values(MediaStorageType).sort()
+      )
     })
   })
 
@@ -267,7 +269,7 @@ describe('Media Storage Service', () => {
           : S3FileStorage.S3FileStorage.getStorage
       expectedStorage[mediaFunction.method].mockResolvedValue(DELEGATED_RESULT)
 
-      const result = await mediaFunction.call()
+      const result = await invoke(mediaFunction)
 
       expect(result).toBe(DELEGATED_RESULT)
       expect(expectedGetStorage).toHaveBeenCalledWith(
@@ -276,7 +278,7 @@ describe('Media Storage Service', () => {
         mockDatabase
       )
       expect(expectedStorage[mediaFunction.method]).toHaveBeenCalledWith(
-        ...mediaFunction.expectedArguments
+        ...mediaFunction.callArguments
       )
       expect(otherStorage[mediaFunction.method]).not.toHaveBeenCalled()
     })
@@ -287,7 +289,7 @@ describe('Media Storage Service', () => {
         async (storage) => {
           configureStorage(storage.mediaStorage)
 
-          const result = await mediaFunction.call()
+          const result = await invoke(mediaFunction)
 
           expect(result).toBe(mediaFunction.fallbackValue)
           expect(localStorageMock[mediaFunction.method]).not.toHaveBeenCalled()
@@ -299,7 +301,7 @@ describe('Media Storage Service', () => {
     it('returns the fallback value when no storage is configured', async () => {
       configureStorage(undefined)
 
-      const result = await mediaFunction.call()
+      const result = await invoke(mediaFunction)
 
       expect(result).toBe(mediaFunction.fallbackValue)
       expect(localStorageMock[mediaFunction.method]).not.toHaveBeenCalled()
@@ -315,7 +317,10 @@ describe('Media Storage Service', () => {
     it('returns false when the driver reports the deletion failed', async () => {
       localStorageMock.deleteFile.mockResolvedValue(false)
 
-      const result = await deleteMediaFile(mockDatabase, 'medias/test.jpg')
+      const result = await mediaStorageService.deleteMediaFile(
+        mockDatabase,
+        'medias/test.jpg'
+      )
 
       expect(result).toBe(false)
       // Without this the assertion above cannot tell a driver that returned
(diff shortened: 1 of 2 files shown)


lib/services/medias/index.test.ts
```diff
@@ -31,249 +30,263 @@ vi.mock('@/lib/services/medias/localFile')
...
-  mockLocalStorage as unknown as ReturnType<
-    typeof LocalFileStorage.LocalFileStorage.getStorage
-  >
-)
-vi.spyOn(S3FileStorage.S3FileStorage, 'getStorage').mockReturnValue(
-  mockS3Storage as unknown as ReturnType<
-    typeof S3FileStorage.S3FileStorage.getStorage
-  >
-)
+const MEDIA_STORAGE_FUNCTIONS: MediaStorageFunctionCase[] = [
```
llun, line 122: **Nothing forces a new exported function into the table (silent gap).**

`MEDIA_STORAGE_FUNCTIONS` is hand-maintained with no completeness assertion, so an eighth export is simply never tested.

Demonstrated: appending a `saveMediaVariant` to `index.ts` with a storage switch that omits `case MediaStorageType.S3Storage:` — literally the bug this PR fixes — and leaving the table untouched yields **30/30 passing, green**.

The table covers all 7 current exports correctly; the gap is only about the next one. Since the stated purpose of this rewrite is that the next omission fails a test rather than shipping, this is worth closing.

lib/services/medias/index.test.ts
```diff
@@ -31,249 +30,263 @@ vi.mock('@/lib/services/medias/localFile')
...
+    const unsupported = STORAGE_CONFIGS.filter(
+      (storage) =>
+        !mediaFunction.supportedTypes.includes(storage.mediaStorage.type)
+    )
 
-        expect(mockDeleteFile).toHaveBeenCalledWith(
-          'media/file with spaces.jpg'
-        )
-      })
+    it.each(supported)('delegates to $description', async (storage) => {
```
llun, line 260: **`it.each([])` is silent in Vitest 4, so shrinking `STORAGE_CONFIGS` deletes tests without failing.**

`it.each(supported)` registers zero tests and raises no error when `supported` is empty. There is an `if (unsupported.length > 0)` guard at line 243, but none on `supported` — and `supported` gates every delegation assertion.

Demonstrated end to end:

1. Delete the `'s3 storage'` entry from `STORAGE_CONFIGS` (lines 80-86) — a plausible future edit, e.g. "`s3` and `object` are the same code path, this is redundant." Suite goes **30 → 23 tests, still green**. Seven tests vanish with no signal.
2. With that drift in place, reintroducing the exact bug this PR fixes (removing `case MediaStorageType.S3Storage:` from `saveMedia`) gives **23/23 passing**. The regression is fully invisible again.

The same class of gap covers `ALL_STORAGE_TYPES`/`S3_STORAGE_TYPES` at lines 99-107: a fourth `MediaStorageType` enum member would be covered by nothing.

lib/services/medias/index.test.ts
```diff
@@ -31,249 +30,263 @@ vi.mock('@/lib/services/medias/localFile')
...
+      configureStorage({ type: MediaStorageType.LocalFile, path: '/tmp/media' })
     })
 
-    describe('No storage configured', () => {
-      beforeEach(() => {
-        mockGetConfig.mockReturnValue({
-          host: 'https://example.com'
-        } as unknown as ReturnType<typeof getConfig>)
-      })
+    it('returns false when the driver reports the deletion failed', async () => {
```
llun, line 317: **Minor: this test cannot fail for the reason its name states.**

It asserts only `expect(result).toBe(false)`, but `deleteMediaFile`'s `default` branch also returns `false` — so the assertion cannot distinguish "driver returned false" from "no driver ran at all".

Verified: deleting the entire `case MediaStorageType.LocalFile` block from `deleteMediaFile` fails `delegates to 'local file storage'` and `passes the path through unchanged`, but **this test stays green**. Redundant rather than harmful given those siblings, but an `expect(localStorageMock.deleteFile).toHaveBeenCalledWith(...)` would make it mean what its name says.

llun (commented):
Sub-agent review round 1 — three dimensions (fix correctness/scope, test quality, project invariants).

The test-quality pass mutation-tested the harness rather than eyeballing it: all 19 `case MediaStorageType.X:` labels across the 7 functions are killed by the matrix, plus 10 semantic mutations (driver swap, method swap, argument drops, `getStorage` arg reorder, `null`→`undefined`, `false`→`null`, await-then-return-null). Zero survivors — so the matrix has teeth for every function/type combination that exists **today**.

The findings below are all about the harness drifting **tomorrow**. Each was demonstrated end to end, not theorised.

lib/services/medias/index.test.ts
llun: Fixed in 56079834e. Added a `matrix coverage` describe block; the first guard asserts the table describes every module export:

```ts
const exported = Object.keys(mediaStorageService).filter(
  (name) => !NON_DELEGATING_EXPORTS.includes(name)
)
expect(exported.sort()).toEqual(
  MEDIA_STORAGE_FUNCTIONS.map((mediaFunction) => mediaFunction.name).sort()
)
```

`NON_DELEGATING_EXPORTS` holds only `PresignedUploadValidationError`, with a comment saying why it is exempt — a re-exported error class, not a storage-delegating function. Any other new export has to be added to the table or explicitly exempted.

To read the real module I switched the named imports to a namespace import plus a destructure (`const { saveMedia, … } = mediaStorageService`), which avoids importing the same module twice.

Verified against your exact repro: appending `export const saveMediaVariant` to `index.ts` and leaving the table untouched now fails `describes every function exported from the module` (1 failed | 32 passed).

lib/services/medias/index.test.ts
llun: Fixed in 56079834e. Two guards in the new `matrix coverage` block close both halves of this:

```ts
expect(exercised.sort()).toEqual(Object.values(MediaStorageType).sort())
```

pins `STORAGE_CONFIGS` to the enum, so deleting an entry fails and a fourth `MediaStorageType` member fails until it is exercised. And:

```ts
const withoutSupportedTypes = MEDIA_STORAGE_FUNCTIONS.filter(
  (mediaFunction) => mediaFunction.supportedTypes.length === 0
).map((mediaFunction) => mediaFunction.name)
expect(withoutSupportedTypes).toEqual([])
```

closes the same empty-`it.each` hole on the per-function axis, which was the other way `supported` could silently reach zero.

Verified both mutations: deleting the `s3

lib/services/medias/index.test.ts
llun: Fixed in 56079834e. Added the delegation assertion so the test means what its name says:

```ts
expect(result).toBe(false)
// Without this the assertion above cannot tell a driver that returned
// false from the `default` branch, which returns false too.
expect(localStorageMock.deleteFile).toHaveBeenCalledWith("medias/test.jpg")
```

Agreed it was redundant rather than harmful given the sibling matrix cases, but it cost one line to make it load-bearing.

lib/services/medias/index.test.ts
```diff
@@ -31,249 +32,307 @@ vi.mock('@/lib/services/medias/localFile')
...
+    mediaStorage: {
+      type: MediaStorageType.ObjectStorage,
+      bucket: 'test-bucket',
+      region: 'us-east-1',
+      endpoint: 'https://s3.example.com'
+    }
+  }
+]
+
+const ALL_STORAGE_TYPES = [
```
llun, line 91: **`ALL_STORAGE_TYPES` has no guard, so the "fourth storage type" hole is only half closed.**

Guard 2 constrains `STORAGE_CONFIGS`, but `ALL_STORAGE_TYPES` is a *second* hand-maintained enum list and it is the one that decides supported-vs-fallback. A new enum member is therefore auto-classified as legitimately unsupported.

Verified scenario: add `GcsStorage = 'gcs'` to `MediaStorageType` → guard 2 fails → the natural fix is to add a `STORAGE_CONFIGS` entry → suite goes green at **40/40**, with all seven functions each gaining a *passing* `returns the fallback value for 'gcs storage'` test asserting the `null`/`false` return is correct.

That is exactly the shape of the bug this PR fixes: a configured storage type where `saveMedia` silently returns `null` and media vanishes. The guard converts total invisibility into one transient red test whose obvious resolution re-blesses the broken behaviour.

lib/services/medias/index.test.ts
```diff
@@ -31,249 +32,307 @@ vi.mock('@/lib/services/medias/localFile')
...
-  mockLocalStorage as unknown as ReturnType<
-    typeof LocalFileStorage.LocalFileStorage.getStorage
-  >
-)
-vi.spyOn(S3FileStorage.S3FileStorage, 'getStorage').mockReturnValue(
-  mockS3Storage as unknown as ReturnType<
-    typeof S3FileStorage.S3FileStorage.getStorage
-  >
-)
+const MEDIA_STORAGE_FUNCTIONS: MediaStorageFunctionCase[] = [
```
llun, line 122: **The completeness guard matches names, not functions, so a copy-pasted table entry keeps it green.**

`MediaStorageFunctionCase.name` is a free-text string with no relationship to the `call` closure, so the guard proves only that *a table row bearing that name exists*.

Verified scenario: appended a real `purgeMediaCache` export to `index.ts` whose switch handles only `LocalFile` (omitting `S3Storage`/`ObjectStorage` — the branch's own bug), added a table entry with `name: 'purgeMediaCache'` but left the duplicated `call: () => getMedia(...)`. Result: **37/37 green**, including four tests *labelled* `purgeMediaCache` that never invoke it. Duplicating a neighbouring row and missing the closure is the realistic way that entry gets written.

Same class: `NON_DELEGATING_EXPORTS` is a one-line escape hatch and the guard's failure message does not distinguish "add to the table" from "add to `NON_DELEGATING_EXPORTS`" — the easier move is the wrong one for a delegating function.

lib/services/medias/index.test.ts
```diff
@@ -31,249 +32,307 @@ vi.mock('@/lib/services/medias/localFile')
...
+      )
 
-        const result = await deleteMediaFile(mockDatabase, 'media/test.jpg')
+      expect(exercised.sort()).toEqual(Object.values(MediaStorageType).sort())
+    })
 
-        expect(result).toBe(true)
-        expect(mockDeleteFile).toHaveBeenCalledWith('media/test.jpg')
-      })
+    it('declares at least one supported storage type per function', () => {
```
llun: **Correction to my round-1 commit message: this guard is redundant and the rationale I gave for it was wrong.**

I wrote that an empty `supportedTypes` "silently skipped a function's delegation tests via the same empty-`it.each` hole". It does not: emptying it moves all three configs into `unsupported`, and `unsupported.length > 0` is then satisfied, so it fails loudly.

Setting `saveMedia.supportedTypes = []` fails **4** tests — this guard plus the three `saveMedia > returns the fallback value for X`. Delete this guard and it is still 3 failures. Harmless, but not load-bearing, and it should not be described as closing a hole it does not close.

llun (commented):
Sub-agent review round 2 — re-review of the round-1 fix commit 56079834e, since fixes introduce new problems.

Confirmed sound empirically: `Object.keys(mediaStorageService)` returns exactly the 8 value exports (no `default`, no symbols) and `export type`/`export interface` are erased so they do not trip the guard; `Object.values(MediaStorageType)` returns exactly `['fs','object','s3']` with no reverse mapping (string enum). Guard 1 kills an 8th export omitting `S3Storage`; guard 2 kills both a dropped `STORAGE_CONFIGS` entry and a 4th enum member.

Two gaps remain in the guards' teeth, plus one correction to my own round-1 commit message.

lib/services/medias/index.test.ts
llun: Fixed in 4f98f04c7. Added the third guard you specified:

```ts
it("treats every configurable storage type as supported by default", () => {
  expect([...ALL_STORAGE_TYPES].sort()).toEqual(
    Object.values(MediaStorageType).sort()
  )
})
```

Verified against your exact scenario: adding `GcsStorage = "gcs"` to the enum and then taking the natural fix of adding a `STORAGE_CONFIGS` entry now fails `treats every configurable storage type as supported by default` (1 failed | 39 passed) where it previously went green at 40/40. The new member lands in `supported`, so its delegation tests stay red until `index.ts` actually handles it.

lib/services/medias/index.test.ts
llun: Fixed in 4f98f04c7, structurally rather than by assertion — a `fn` field would still leave `call` free to invoke something else.

Rows no longer carry a closure at all. `expectedArguments` and `call` merged into one `callArguments` field (everything after `database`), and the harness looks the export up by name:

```ts
const invoke = (mediaFunction: MediaStorageFunctionCase) => {
  const exported = mediaStorageService[mediaFunction.name] as (
    ...args: unknown[]
  ) => Promise<unknown>
  return exported(mockDatabase, ...mediaFunction.callArguments)
}
```

`name` is now typed `keyof typeof mediaStorageService`, so a typo is a type error rather than a silently-never-invoked row. Merging the two argument fields is sound because every function in this module hands its post-`database` arguments straight to the driver method; one that reshapes them would fail loudly here rather than needing a second field.

Verified against your repro: the `purgeMediaCache` export omitting both S3 cases, with a row whose args were copy-pasted from `getMedia`, now fails `purgeMediaCache > delegates to "s3 storage"` and `> delegates to "object storage"` (2 failed | 35 passed) instead of passing 37/37.

Dropping the closures also removed the module-level destructure, which as you noted was a namespace snapshot rather than a live binding.

lib/services/medias/index.test.ts
llun: You are right on both counts, and thank you for checking the claim rather than the code. Emptying `supportedTypes` moves all three configs into `unsupported`, `unsupported.length > 0` is satisfied, and it fails loudly — it never went through the empty-`it.each` hole at all.

Removed the guard in 4f98f04c7 rather than keeping a redundant test with a corrected comment, and the new commit message states plainly that the round-1 rationale was wrong. The empty-`it.each` hazard is real, but `STORAGE_CONFIGS` was the only axis that could actually reach zero through it, and the enum guard covers that.

llun:
## Sub-agent review round 3 — clean, loop terminated

Round 3 re-reviewed the round-2 harness restructure (4f98f04c7), since each round's fix is itself new code. **No findings.** 36 mutations run, 35 killed.

**Merging `expectedArguments` into `callArguments` is not circular.** `callArguments` is fixed literal data the harness both feeds in and asserts against — never derived from what `index.ts` actually did. Every argument-shape mutation still goes red: dropping `format` from `saveImageRendition`, reordering `saveFile`/`saveThumbnail`/`getPresigedForSaveFileUrl` args, substituting a different path in `getFile`, wrapping `deleteFile`'s path in `encodeURIComponent`, a literal `'media-2'` in `completePresignedUpload`, an extra trailing arg, a branch resolving the wrong driver, a branch calling the wrong method, and both fallback-value swaps.

**The original bug still fails, and the full label sweep still passes.** Dropping the `S3Storage` case in each of the 7 functions → 7/7 red. Dropping the `LocalFile` branch in each of the 5 that have one → 5/5 red.

**The self-protection guards work.** 8th export with no row → red. `STORAGE_CONFIGS` losing the s3 entry → red. A new enum member alone → red. And the case round 2 was written for — new enum member *plus* the "natural fix" of a `STORAGE_CONFIGS` entry, which used to go green at 40/40 → red.

**`invoke()` is safe.** All seven exports are arrow consts with no `this` usage and `database` first, so the unbound call is correct; the namespace properties are live bindings with no side-effecting getters. A typo'd row name fails both ways — `TS2820` at typecheck and `TypeError: exported is not a function` at runtime. A row naming `PresignedUploadValidationError` also fails loudly (completeness guard plus the class throwing without `new`), so there is no silent path.

**Removing the `supportedTypes`-non-empty guard lost nothing**, confirming the correction in that commit message: an empty `supportedTypes` moves every config into `unsupported`, where the function still delegates and the un-stubbed mock returns `undefined` — failing with `expected undefined to be null`.

### Recorded, deliberately not changed

- **One surviving mutation, pre-existing:** replacing `database` with a fresh `{}` inside `saveMedia`'s `getStorage(...)` stays green, because `mockDatabase` is `{}` and `toHaveBeenCalledWith` is structural. Identical in all three commits and untouched by this branch; swapping `host`/`database`, passing `undefined`, or omitting the argument all still go red. No realistic bug shape.
- **`S3_STORAGE_TYPES` stays hand-maintained.** A new enum member would land in `unsupported` for the two presigned functions. Not a silent hole — the same mutation turns the five `ALL_STORAGE_TYPES` functions red, so the suite forces a decision rather than green-blessing it quietly.

---

**Loop summary:** 3 rounds, 6 findings, all addressed and resolved. Round 1 — one production-code review (clean: `S3FileStorage` never inspects `config.type`, and `saveMedia` was the only omission in the repo), one invariants review (clean: docs already document `s3`, so this restores documented behavior), one test review (3 findings). Round 2 — 3 findings, including a correction to a wrong claim in the round-1 commit message. Round 3 — clean.

PR #1338, opened 2026-07-28
fix: dispatch s3 media storage type in saveMedia

llun:
## Summary

`saveMedia` in `lib/services/medias/index.ts` switched on `mediaStorage?.type` but handled only `MediaStorageType.LocalFile` and `MediaStorageType.ObjectStorage` — it was missing `MediaStorageType.S3Storage` (`'s3'`). Every other dispatcher in the same file (`saveMediaThumbnail`, `saveMediaImageRendition`, `getPresignedUrl`, `completePresignedMediaUpload`, `getMedia`, `deleteMediaFile`) already lists `case MediaStorageType.S3Storage:` alongside `ObjectStorage`.

`'s3'` is a fully supported value: `getMediaStorageConfig` accepts `ACTIVITIES_MEDIA_STORAGE_TYPE=s3` and builds a `MediaStorageS3Config` from it, and the docs (`docs/environment-variables.md`, `docs/setup.md`, `docs/maintenance.md`, `.env.example`) list it as a peer of `object` with no stated limitation.

**Impact:** on an instance configured with `ACTIVITIES_MEDIA_STORAGE_TYPE=s3`, `saveMedia` fell through to `default: return null`, and `handleSyncMediaUpload` converts that null into a 422 — so `POST /api/v1/media` and `POST /api/v2/media` rejected **every** synchronous upload, while thumbnails, image renditions, presigned uploads and reads all kept working. That asymmetry is what made it easy to miss.

## Changes

- `lib/services/medias/index.ts` — add the missing `case MediaStorageType.S3Storage:` above `ObjectStorage` so both S3-backed config types route to `S3FileStorage.getStorage(...).saveFile(...)`.
- `lib/services/medias/index.test.ts` — new `describe('saveMedia', …)` block. `saveFile` is mocked **per driver** (`mockLocalSaveFile` / `mockS3SaveFile`) rather than shared like the existing `deleteFile` mocks, so each case pins *which* storage was reached instead of only that something was called:
  - table-driven `it.each` over `s3` and `object` → the S3 driver receives `(actor, media)` and the local driver is untouched
  - `fs` → the local driver, S3 untouched
  - no storage configured → `null`, neither driver called

## Testing

The regression test was confirmed to fail without the fix — stashing the one-line change makes `'delegates s3 storage to the s3 driver'` fail with `expected null to deeply equal { id: 'media-1' }`, which is exactly the null `handleSyncMediaUpload` was turning into a 422.

Full gate, run in order, all green:

- `yarn run prettier --write .` — no files reformatted beyond the two changed
- `yarn lint` — 0 errors (31 pre-existing `no-explicit-any` warnings, untouched)
- `yarn build` — clean
- `yarn test` — 737 files / 7579 tests passed

## Notes

- No migration, no schema-dump change, no config or environment-variable change.
- No docs update needed: the docs already document `s3` and `object` as equivalent backends, so this restores documented behavior rather than changing it.
- No UI surface, so no screenshots.

### DIFF
--- a/lib/services/medias/index.test.ts
+++ b/lib/services/medias/index.test.ts
@@ -7,8 +7,9 @@ import { Database } from '@/lib/database/types'
 import { Actor } from '@/lib/types/domain/actor'
 
 import * as S3FileStorage from './S3StorageFile'
-import { deleteMediaFile, saveMediaImageRendition } from './index'
+import { deleteMediaFile, saveMedia, saveMediaImageRendition } from './index'
 import * as LocalFileStorage from './localFile'
+import { MediaSchema } from './types'
 
 vi.mock('fs', () => ({
   promises: {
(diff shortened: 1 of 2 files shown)


llun:
Superseded by #1337, which landed the identical fix on `main` (commit e1cea90e2).

`git diff origin/main HEAD -- lib/services/medias/index.ts` on this branch is **empty** — #1337 added the same single `case MediaStorageType.S3Storage:` label to `saveMedia`.

The only remaining difference was the test, and #1337's is a strict superset: it replaced the per-storage blocks with a delegation matrix over all seven exports × all three storage types (so `saveMedia` × `s3`/`object` is covered), plus coverage guards asserting the table describes every module export and every `MediaStorageType` enum value — which would catch the next omission of this shape rather than just this one. Nothing here to carry over.

Closing as redundant.

PR #1341, opened 2026-07-28
fix: store a caller-supplied thumbnail on the S3 media driver

llun:
## Summary

`S3FileStorage.saveFile` handled `media.thumbnail` only in its **video** branch, where it built one from the extracted preview frame. The **image** branch called `createMedia` with no `thumbnail` key at all, so a thumbnail uploaded beside an image was silently discarded.

`MediaSchema` accepts an optional `thumbnail` and `handleSyncMediaUpload` passes the parsed schema straight through, so the same `POST /api/v1/media` request produced a different `meta.small` and `preview_url` depending on which storage backend the instance runs: stored on a filesystem instance, dropped on an object-storage one. (Upload-response only — `getMastodonAttachment` hardcodes `preview_url: null` and the ActivityPub `Document` has no thumbnail field, so nothing federated.)

The video branch had the matching half of the gap — it always used the extracted frame and ignored a caller-supplied thumbnail, which the local driver prefers.

Five rounds of sub-agent review turned this from "store the thumbnail on S3 too" into "both drivers answer a supplied thumbnail identically, and neither leaves a stored file behind when it can't". The review comments record what each round found.

## What changed

**`lib/services/medias/S3StorageFile.ts`** — the image and video branches share one `createMedia` call with the local driver's precedence: a caller-supplied thumbnail wins → else a video's extracted preview frame → else no thumbnail. The stored thumbnail is recorded from `outputInfo` (the stored WebP), so `meta.small` and the metered `thumbnail.bytes` describe what actually landed in the bucket. `_uploadImageBufferToS3` returns `previewImage: null` so an image and a video upload share one result shape, exactly as `_saveImageBuffer` does locally.

**`lib/services/medias/thumbnailInput.ts`** (new) — a supplied thumbnail is unvalidated client input, and both drivers read one through `readValidThumbnail` so the answer cannot differ by backend. It checks the declared type (`FileSchema` accepts video and audio too) and fully decodes the bytes — `sharp().stats()`, not the cheaper `metadata()`, which parses only the header and so lets a truncated image through to the encoder, where the failure is no longer distinguishable from a storage fault. It runs before anything is stored and returns the bytes, so the drivers store them without reading the file again.

**`lib/services/medias/localFile.ts`** — same validation, same quota reservation, same reclaim. Its behaviour changes too: it previously fed an unvalidated thumbnail straight to sharp.

**Reclaim on failure, both drivers** — a `medias` row is the only handle anything has on a stored path, so a file written without one is reachable only by `scripts/maintenance/cleanupMediaStorage.ts`. Whatever fails between the first write and the row now takes the files back out: a thumbnail that fails to store, `createMedia` rejecting, and `createMedia` returning null. The error itself is never relabelled — a storage fault stays a logged 500 rather than becoming a 422 the client will not retry.

**Quota** — `getUploadQuotaReservation` (in `medias/quota`) reserves `file.size + thumbnail.size`, since `createMedia` meters `original.bytes + thumbnail.bytes`. Best-effort rather than exact: a lossy JPEG can re-encode into a *larger* near-lossless WebP, so this can still under-reserve — `original.bytes` under-counts the same way and always has. Reserving nothing was strictly worse: an upload that fits only without its thumbnail was accepted and left the account over quota. This tightens acceptance slightly — an upload at the quota edge carrying a thumbnail now 422s where it previously succeeded.

**`saveThumbnail`** (`PUT|PATCH /api/v1/media/:id`) uses the same validation, so the dedicated thumbnail endpoint answers with 422 what it used to 500 on.

**Docs** — `docs/architecture.md` gains a paragraph on thumbnail handling in the media-storage section; `REVIEW.md` records the two invariants this surfaced (the drivers must answer identically; a write that loses its row must be reclaimed, not relabelled as client error).

## Testing

`thumbnailInput.test.ts` pins the validator directly, including the two inputs that isolate each half of it: real image bytes declared `video/mp4` (only the type guard rejects it) and a truncated PNG (only the full decode does).

Both drivers carry a `saveFile with a caller-supplied thumbnail` block covering: the thumbnail is stored at its own dimensions under a `-thumbnail` key; `createMedia` records path/bytes/mimeType/metaData from the *stored* file; the attachment reports `meta.small`/`preview_url`; an above-cap thumbnail (the only case that distinguishes `outputInfo` from the input metadata); `description`/`focus` survive the unified call; the quota reservation; a non-image thumbnail; a truncated one; and reclaim on both `createMedia` failure modes. The S3 block additionally injects a storage failure to pin that the original is reclaimed and the error is not relabelled. Video precedence is a table on both drivers. Coverage is not identical between the two files — some cases exist only where they can be driven, and the S3 block is the superset.

Every assertion was mutation-checked: reverting each behaviour in turn — `outputInfo`→`metaData`, dropping the `description`/`focus` spreads, dropping the type guard, `stats()`→`metadata()`, removing each reclaim, inverting the precedence on either driver — fails exactly the tests that cover it and nothing else. The type-guard mutation initially failed *nothing*, which is why that test exists.

Gate: `prettier --write .`, `eslint .` (0 errors), `next build`, `vitest run` — **738 files / 7623 tests passed**.

No UI surface: the observable change is the Mastodon attachment JSON and the upload status codes, which the tests assert.

## Notes

- Found during review of #1334, which deliberately did not widen its scope; re-verified as still valid after #1332 and #1337 merged.
- Out of scope, filed separately: `LocalFileStorage._saveVideoFile` writes the video into the media root and *then* runs `extractVideoImage`, so an ffmpeg failure orphans the video — the same class of leak this PR fixes elsewhere, on a path it does not touch (the S3 driver extracts before its PUT and has no such gap).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -181,8 +181,10 @@ and both storage drivers treat it identically: it is re-encoded and stored the
 same way, recorded on the media row from the **stored** file's size and
 dimensions (so it is metered against the account's storage usage), and surfaced
 as the attachment's `meta.small` and `preview_url`. For a video it replaces the
-frame the server would otherwise extract; without one, a video still gets that
-extracted frame and an image gets no thumbnail at all.
+extracted frame — the frame is extracted either way, so it saves no work — while
+without one a video keeps that frame and an image gets no thumbnail at all. A
+thumbnail that is not an image is refused before anything is stored, so an
+unusable one cannot leave an orphaned original behind.
 
 **Presigned direct-to-S3** stores the bytes exactly as uploaded — original
 format, no re-encode, no server-side dimension cap — because the browser PUTs
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -183,11 +183,14 @@ dimensions (so it is metered against the account's storage usage), and surfaced
 as the attachment's `meta.small` and `preview_url`. For a video it replaces the
 extracted frame — the frame is extracted either way, so it saves no work — while
 without one a video keeps that frame and an image gets no thumbnail at all. A
-supplied thumbnail is client input, so both drivers validate it — declared type
-and actual bytes — through the same module before anything is stored, and
-unusable input is a 422 with nothing written. Should storing it fail after that,
-the original stored alongside it is reclaimed: a `medias` row is the only handle
-anything has on a stored path, so a file written without one is unreachable.
+supplied thumbnail is client input, so both drivers validate it through the same
+module: the declared type and a header parse before anything is stored, and — if
+storing it fails anyway, as a truncated image only can once the encoder reaches
+the missing bytes — a full decode to decide what went wrong. Unusable input is a
+422; a storage fault keeps its own error and stays a logged 500. Either way the
+original stored alongside it is reclaimed, because a `medias` row is the only
+handle anything has on a stored path and a file written without one is
+unreachable.
 
 **Presigned direct-to-S3** stores the bytes exactly as uploaded — original
 format, no re-encode, no server-side dimension cap — because the browser PUTs
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -183,10 +183,11 @@ dimensions (so it is metered against the account's storage usage), and surfaced
 as the attachment's `meta.small` and `preview_url`. For a video it replaces the
 extracted frame — the frame is extracted either way, so it saves no work — while
 without one a video keeps that frame and an image gets no thumbnail at all. A
-thumbnail that is not an image is refused before anything is stored; one that
-merely claims to be an image is caught by the encoder instead, and the original
-it arrived with is reclaimed rather than left orphaned. Either way the upload
-answers 422, not 500.
+supplied thumbnail is client input, so both drivers validate it — declared type
+and actual bytes — through the same module before anything is stored, and
+unusable input is a 422 with nothing written. Should storing it fail after that,
+the original stored alongside it is reclaimed: a `medias` row is the only handle
+anything has on a stored path, so a file written without one is unreachable.
 
 **Presigned direct-to-S3** stores the bytes exactly as uploaded — original
 format, no re-encode, no server-side dimension cap — because the browser PUTs
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -176,6 +176,14 @@ and `PATCH /api/v1/profile`), custom emojis, thumbnail replacement
 and the fitness import jobs — which cover both the route maps the server
 generates and the arbitrary-sized activity photos it fetches from Strava.
 
+The sync upload endpoint also accepts an optional `thumbnail` beside the file,
+and both storage drivers treat it identically: it is re-encoded and stored the
+same way, recorded on the media row from the **stored** file's size and
+dimensions (so it is metered against the account's storage usage), and surfaced
+as the attachment's `meta.small` and `preview_url`. For a video it replaces the
+frame the server would otherwise extract; without one, a video still gets that
+extracted frame and an image gets no thumbnail at all.
+
 **Presigned direct-to-S3** stores the bytes exactly as uploaded — original
 format, no re-encode, no server-side dimension cap — because the browser PUTs
 straight to the bucket. `uploadAttachment` (`lib/client.ts`) tries this first
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -183,12 +183,10 @@ dimensions (so it is metered against the account's storage usage), and surfaced
 as the attachment's `meta.small` and `preview_url`. For a video it replaces the
 extracted frame — the frame is extracted either way, so it saves no work — while
 without one a video keeps that frame and an image gets no thumbnail at all. A
-supplied thumbnail is client input, so both drivers validate it through the same
-module: the declared type and a header parse before anything is stored, and — if
-storing it fails anyway, as a truncated image only can once the encoder reaches
-the missing bytes — a full decode to decide what went wrong. Unusable input is a
-422; a storage fault keeps its own error and stays a logged 500. Either way the
-original stored alongside it is reclaimed, because a `medias` row is the only
+supplied thumbnail is client input, so both drivers validate it the same way and
+before anything is stored: unusable input is a 422 with nothing written, while a
+storage fault keeps its own error and stays a logged 500. Whatever fails after
+the first write reclaims what it stored, because a `medias` row is the only
 handle anything has on a stored path and a file written without one is
 unreachable.
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -183,8 +183,10 @@ dimensions (so it is metered against the account's storage usage), and surfaced
 as the attachment's `meta.small` and `preview_url`. For a video it replaces the
 extracted frame — the frame is extracted either way, so it saves no work — while
 without one a video keeps that frame and an image gets no thumbnail at all. A
-thumbnail that is not an image is refused before anything is stored, so an
-unusable one cannot leave an orphaned original behind.
+thumbnail that is not an image is refused before anything is stored; one that
+merely claims to be an image is caught by the encoder instead, and the original
+it arrived with is reclaimed rather than left orphaned. Either way the upload
+answers 422, not 500.
 
 **Presigned direct-to-S3** stores the bytes exactly as uploaded — original
 format, no re-encode, no server-side dimension cap — because the browser PUTs
--- a/lib/services/medias/S3StorageFile.test.ts
+++ b/lib/services/medias/S3StorageFile.test.ts
@@ -22,6 +22,7 @@ import {
 import { MediaValidationError } from '@/lib/services/medias/errors'
 import { extractVideoImage } from '@/lib/services/medias/extractVideoImage'
 import { extractVideoMeta } from '@/lib/services/medias/extractVideoMeta'
+import { getQuotaLimit } from '@/lib/services/medias/quota'
 import { getMaxMediaUploadSize } from '@/lib/services/medias/uploadSizeLimit'
 import { Actor } from '@/lib/types/domain/actor'
 import { StreamByteLimitError } from '@/lib/utils/streamLimit'
@@ -600,9 +601,9 @@ describe('S3FileStorage saveFile with a video', () => {
     )
   })
 
-  // The image and video branches share one `createMedia` call, but only the
-  // image branch reaches it through `_uploadImageToS3`, so it needs its own
-  // coverage.
+  // The image and video branches share one `createMedia` call, but the
+  // original's path and metadata come from a different helper in each, so the
+  // image branch needs its own coverage.
   it('stores a sanitized original file name for an image', async () => {
     database.createMedia.mockResolvedValue({
       id: 'media-2',
@@ -650,21 +651,29 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
   // deletes its temp file as soon as `send` resolves, so an image stream has to
   // be drained inside the mock.
   let uploads: { key: string; body: Buffer }[]
+  let deletedKeys: string[]
 
   beforeEach(() => {
     vi.clearAllMocks()
     uploads = []
+    deletedKeys = []
     ;(S3Client as jest.MockedClass<typeof S3Client>).mockImplementation(
       function () {
         return { send } as unknown as S3Client
       }
     )
     send.mockImplementation(async (command) => {
+      if (command instanceof DeleteObjectCommand) {
+        deletedKeys.push(String(command.input.Key))
+        return {}
+      }
       if (!(command instanceof PutObjectCommand)) {
         throw new Error('Unexpected command')
       }
       const { Key, Body } = command.input
       const chunks: Buffer[] = []
+      // A video is uploaded as a Buffer and an image as a stream. `for await`
+      // over a Buffer walks it byte by byte, so the two need separate handling.
       if (Buffer.isBuffer(Body)) {
         chunks.push(Body)
       } else {
@@ -838,6 +847,47 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
     ).rejects.toThrow(MediaValidationError)
     expect(uploads).toHaveLength(0)
     expect(database.createMedia).not.toHaveBeenCalled()
+  })
+
+  // `createMedia` meters the thumbnail's bytes too, so the pre-check has to
+  // reserve for them — otherwise an upload that fits only without its thumbnail
+  // is accepted and leaves the account over its quota.
+  it('counts the thumbnail against the account quota', async () => {
+    const file = await createPngFile(800, 600)
+    const thumbnail = await createPngFile(400, 300)
+    // Exactly enough room for the original on its own.
+    database.getStorageUsageForAccount.mockResolvedValue(
+      getQuotaLimit() - file.size
+    )
+
+    await expect(
+      createStorage().saveFile(actor, { file, thumbnail })
+    ).rejects.toThrow(MediaValidationError)
+    expect(uploads).toHaveLength(0)
+    // The same upload without the thumbnail still fits, so the thumbnail's
+    // bytes are what tipped it over.
+    await expect(
+      createStorage().saveFile(actor, { file })
+    ).resolves.toBeTruthy()
+  })
+
+  // The declared type is only a claim, and sharp is the real arbiter — so the
+  // guard above cannot be the whole story. By the time sharp rejects, the
+  // original is already in the bucket and no `medias` row will ever reference
+  // it.
+  it('reclaims the stored original when the thumbnail is not a readable image', async () => {
+    const thumbnail = new File([Buffer.from('not-an-image')], 'evil.png', {
+      type: 'image/png'
+    })
+
+    await expect(
+      createStorage().saveFile(actor, {
+        file: await createPngFile(800, 600),
+        thumbnail
+      })
+    ).rejects.toThrow(MediaValidationError)
+    expect(database.createMedia).not.toHaveBeenCalled()
+    expect(deletedKeys).toEqual([uploaded('original').key])
   })
 
   it('stores no thumbnail for an image uploaded without one', async () => {
--- a/lib/services/medias/S3StorageFile.test.ts
+++ b/lib/services/medias/S3StorageFile.test.ts
@@ -849,6 +849,19 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
     expect(database.createMedia).not.toHaveBeenCalled()
   })
 
+  // The dedicated thumbnail endpoint (PUT /api/v1/media/:id) has to answer the
+  // same bytes the same way — it used to reach sharp unguarded and 500.
+  it('refuses unreadable bytes on the standalone thumbnail path', async () => {
+    const thumbnail = new File([Buffer.from('not-an-image')], 'evil.png', {
+      type: 'image/png'
+    })
+
+    await expect(
+      createStorage().saveThumbnail(actor, thumbnail)
+    ).rejects.toThrow(MediaValidationError)
+    expect(uploads).toHaveLength(0)
+  })
+
   // `createMedia` meters the thumbnail's bytes too, so the pre-check has to
   // reserve for them — otherwise an upload that fits only without its thumbnail
   // is accepted and leaves the account over its quota.
@@ -872,10 +885,10 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
   })
 
   // The declared type is only a claim, and sharp is the real arbiter — so the
-  // guard above cannot be the whole story. By the time sharp rejects, the
-  // original is already in the bucket and no `medias` row will ever reference
-  // it.
-  it('reclaims the stored original when the thumbnail is not a readable image', async () => {
+  // type guard above cannot be the whole story. Validating the bytes up front
+  // is what keeps this a 422 with nothing stored, rather than a 500 with an
+  // original in the bucket that no `medias` row will ever reference.
+  it('refuses unreadable thumbnail bytes before storing anything', async () => {
     const thumbnail = new File([Buffer.from('not-an-image')], 'evil.png', {
       type: 'image/png'
     })
@@ -886,8 +899,51 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
         thumbnail
       })
     ).rejects.toThrow(MediaValidationError)
+    expect(uploads).toHaveLength(0)
     expect(database.createMedia).not.toHaveBeenCalled()
-    expect(deletedKeys).toEqual([uploaded('original').key])
+  })
+
+  // What is left after that check is a storage fault, not bad input: it must
+  // keep its own error — a logged 500, not a 422 telling the caller its
+  // perfectly good thumbnail was rejected — while still reclaiming the original.
+  it('reclaims the stored original when the thumbnail upload fails', async () => {
+    const failure = new Error('S3 unavailable')
+    let puts = 0
+    send.mockImplementation(async (command) => {
+      if (command instanceof DeleteObjectCommand) {
+        deletedKeys.push(String(command.input.Key))
+        return {}
+      }
+      puts += 1
+      // The second PutObject is the thumbnail; the original is already stored.
+      if (puts === 2) throw failure
+      uploads.push({ key: String(command.input.Key), body: Buffer.alloc(0) })
+      return {}
+    })
+
+    await expect(
+      createStorage().saveFile(actor, {
+        file: await createPngFile(800, 600),
+        thumbnail: await createPngFile(400, 300)
+      })
+    ).rejects.toThrow(failure)
+    expect(database.createMedia).not.toHaveBeenCalled()
+    expect(deletedKeys).toEqual([uploads[0].key])
+  })
+
+  // A row is the only handle anything else has on these paths, so without one
+  // both stored objects are unreachable.
+  it('reclaims both stored objects when the media row cannot be created', async () => {
+    database.createMedia.mockResolvedValue(null as never)
+
+    await expect(
+      createStorage().saveFile(actor, {
+        file: await createPngFile(800, 600),
+        thumbnail: await createPngFile(400, 300)
+      })
+    ).rejects.toThrow('Fail to store media')
+    expect(deletedKeys).toEqual(uploads.map((upload) => upload.key))
+    expect(deletedKeys).toHaveLength(2)
   })
 
   it('stores no thumbnail for an image uploaded without one', async () => {
@@ -901,33 +957,42 @@ describe('S3FileStorage saveFile with a caller-supplied thumbnail', () => {
     )
   })
 
-  it('prefers a caller-supplied thumbnail over the extracted video frame', async () => {
+  it.each([
+    {
+      description:
+        'prefers a caller-supplied thumbnail over the extracted video frame',
+      suppliedThumbnail: { width: 400, height: 300 },
+      storedThumbnail: { width: 400, height: 300 }
+    },
+    {
+      description:
+        'falls back to the extracted video frame when no thumbnail is supplied',
+      suppliedThumbnail: null,
+      // The mocked `extractVideoImage` frame.
+      storedThumbnail: { width: 1, height: 1 }
+    }
+  ])('$description', async ({ suppliedThumbnail, storedThumbnail }) => {
     const file = new File([Buffer.from('video-bytes')], 'clip.mp4', {
       type: 'video/mp4'
     })
 
     await createStorage().saveFile(actor, {
       file,
-      thumbnail: await createPngFile(400, 300)
+      ...(suppliedThumbnail
+        ? {
+            thumbnail: await createPngFile(
+              suppliedThumbnail.width,
+              suppliedThumbnail.height
+            )
+          }
+        : null)
     })
 
-    // The video plus one thumbnail — the 1x1 preview frame is not uploaded.
+    // The video plus exactly one thumbnail — the losing source is not uploaded.
     expect(uploads).toHaveLength(2)
     await expect(
       sharp(uploaded('thumbnail').body).metadata()
-    ).resolves.toMatchObject({ width: 400, height: 300 })
-  })
-
-  it('falls back to the extracted video frame when no thumbnail is supplied', async () => {
-    const file = new File([Buffer.from('video-bytes')], 'clip.mp4', {
-      type: 'video/mp4'
-    })
-
-    await createStorage().saveFile(actor, { file })
-
-    await expect(
-      sharp(uploaded('thumbnail').body).metadata()
-    ).resolves.toMatchObject({ width: 1, height: 1 })
+    ).resolves.toMatchObject(storedThumbnail)
   })
 })
 
--- a/lib/services/medias/S3StorageFile.test.ts
+++ b/lib/services/medias/S3StorageFile.test.ts
@@ -627,6 +627,192 @@ describe('S3FileStorage saveFile with a video', () => {
   })
 })
 
+// Regression: the image branch called `createMedia` with no `thumbnail` key at
+// all, so a client that uploaded one (MediaSchema accepts it, and
+// handleSyncMediaUpload passes it straight through) got it stored on a
+// filesystem instance and silently dropped on an object-storage one — the same
+// upload produced a different `meta.small`/`preview_url` per backend. The video
+// branch had the matching gap: it always used the extracted frame and ignored a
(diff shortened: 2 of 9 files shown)


llun (commented):
Sub-agent review round 1 — three reviewers (correctness/parity, tests, invariants+docs+security). Findings posted inline below, plus two that cannot be anchored because they fall outside the diff:

**`lib/services/medias/localFile.test.ts` — the "mirrors localFile.test.ts" claim is backwards.** The local driver has exactly one caller-supplied-thumbnail case in `saveFile` (`reports the stored thumbnail as meta.small`, :343) and **none** for the video precedence. Inverting the precedence at `localFile.ts:151-155` so the extracted frame wins over `media.thumbnail` leaves all 15 tests in that file green — i.e. the exact bug this PR fixes on S3 can be reintroduced on the local driver with nothing failing. The S3 tests are a superset; the mirror belongs in `localFile.test.ts`. CONFIRMED by mutation.

**`lib/services/medias/S3StorageFile.test.ts:602` — stale comment.** "`saveFile`'s image branch is a separate `createMedia` call from the video branch above" was true before this PR unified the two branches into one call. Cosmetic, test-file only.

Cleared and explicitly not flagged: the untouched-file-name invariant (the thumbnail's `File.name` never reaches the key, the `Content-Type`, or the row — the key is `crypto.randomBytes(8)` + the output format's extension); `REVIEW.md` → Stored media (`original.*` from the upload, `thumbnail.*` from `outputInfo`); the quota pre-check/metering asymmetry (pre-existing and symmetric across drivers); commit and PR-title prefixes; no SQL in the PR body.

lib/services/medias/S3StorageFile.ts
```diff
@@ -419,54 +419,22 @@ export class S3FileStorage implements MediaStorage {
...
-      // thumbnail when we actually have a preview image to avoid sharp(null).
-      const thumbnail = previewImage
+    const { path, metaData, previewImage } = file.type.startsWith('video')
+      ? await this._uploadVideoToS3(currentTime, file)
+      : await this._uploadImageToS3(currentTime, file)
+    // Same precedence as the local driver: a caller-supplied thumbnail wins
+    // over a frame extracted from a video, and video preview extraction can
+    // fail (no decodable frame), so only fall back to the preview when we
+    // actually have one — sharp(null) would throw.
+    const thumbnail = media.thumbnail
```
llun: **A non-image `thumbnail` now turns a 200 into a 500 and orphans the uploaded original.** CONFIRMED (all three reviewers, verified against the repo's sharp build).

`MediaSchema.thumbnail` is `FileSchema.optional()`, and `FileSchema` accepts every entry of `ACCEPTED_FILE_TYPES` — which includes `video/mp4`, `video/quicktime`, `video/webm` and `audio/mp4`, not just images. That file now goes straight into `_uploadImageToS3` → `sharp(buffer)`.

Failure scenario: `POST /api/v1/media` with `file=photo.png` and `thumbnail=clip.mp4`. sharp rejects with a plain `Error` (`Input buffer contains unsupported image format`), which `handleSyncMediaUpload` maps to **500**, not the 422 every other bad-input case returns — and the original was already `PutObject`'d on the line above, with `createMedia` never reached, so the bucket keeps an object with no `medias` row (reclaimable only by `scripts/maintenance/cleanupMediaStorage.ts`). Repeatable cheaply, bounded only by `media.maxFileSize`.

Before this PR the S3 path silently dropped that thumbnail and returned 200, so this is a **new** failure mode here. It is parity with `localFile.ts:151`, which has the identical hole — but this PR is specifically about making the drivers agree, and this is the one place the agreed-on behaviour is wrong. Note `saveThumbnail` on both drivers already guards (`if (!file.type.startsWith('image')) return null`); `saveFile` does not, and `PUT /api/v1/media/:id` answers the same input with 422.

Fix in **both** drivers, before anything is uploaded, so no orphan is created.

lib/services/medias/S3StorageFile.ts
```diff
@@ -419,54 +419,22 @@ export class S3FileStorage implements MediaStorage {
...
-        file
-      )
-      // Video preview extraction can fail (no decodable frame); only build a
-      // thumbnail when we actually have a preview image to avoid sharp(null).
-      const thumbnail = previewImage
+    const { path, metaData, previewImage } = file.type.startsWith('video')
+      ? await this._uploadVideoToS3(currentTime, file)
+      : await this._uploadImageToS3(currentTime, file)
+    // Same precedence as the local driver: a caller-supplied thumbnail wins
+    // over a frame extracted from a video, and video preview extraction can
```
llun: **This comment states a rationale that cannot happen.** CONFIRMED.

"video preview extraction can fail (no decodable frame), so only fall back to the preview when we actually have one — `sharp(null)` would throw": `extractVideoImage` is typed `Promise<Buffer>` and returns `fs.readFile(...)`. On ffmpeg failure it **rejects**, and `_uploadVideoToS3` awaits it inside `Promise.all`, so the whole call throws long before `saveFile` inspects `previewImage`. The video arm can never yield `null`.

The ternary's real reason for existing is the **image** arm, which now returns `previewImage: null`. The wording was inherited from the pre-refactor code, but this PR rewrote and expanded it, so it should be corrected rather than propagated — it sends the next reader looking for a null-returning extraction path that does not exist.

lib/services/medias/S3StorageFile.ts
```diff
@@ -479,6 +447,20 @@ export class S3FileStorage implements MediaStorage {
...
+              bytes: thumbnail.outputInfo.size,
+              mimeType: thumbnail.contentType,
+              metaData: {
+                width: thumbnail.outputInfo.width,
+                height: thumbnail.outputInfo.height
+              }
+            }
+          }
+        : null),
       ...(media.description ? { description: media.description } : null),
```
llun: **`description` / `focus` survive this refactor untested.** CONFIRMED by mutation.

The PR collapsed two `createMedia` call sites into one and rewrote these two spreads. Deleting **both of them entirely** leaves all 276 tests across `lib/services/medias`, `app/api/v1/media` and `app/api/v2/media` green.

Failure scenario: a future refactor drops alt text on upload; every attachment federates with `description: null`, and no test fails. The new describe block already has the `createMedia` spy wired, so this is a two-line addition.

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -627,6 +627,192 @@ describe('S3FileStorage saveFile with a video', () => {
...
+    expect(uploads).toHaveLength(2)
+    await expect(
+      sharp(uploaded('thumbnail').body).metadata()
+    ).resolves.toMatchObject({ width: 400, height: 300, format: 'webp' })
+  })
+
+  // `thumbnail.bytes` is metered: `createMedia` adds it to the account's usage
+  // counter, so it has to describe the stored WebP (`outputInfo`) rather than
+  // the uploaded PNG.
+  it('records the stored thumbnail on the media row', async () => {
```
llun, line 769: **The `outputInfo` contract is only half-tested — the dimension half cannot fail.** CONFIRMED by mutation.

The comment above claims this pins `thumbnail.*` to the stored WebP rather than the uploaded PNG. `bytes` does that. `metaData` does not: both fixtures are 400x300, below the 4000x4000 cap, so input dims == output dims. Replacing `thumbnail.outputInfo.width/height` with `thumbnail.metaData.width/height` in `S3StorageFile.ts:458-459` leaves **all 31 tests in this file green**.

Failure scenario: a client uploads a 4200x2100 thumbnail; the driver stores a 4000x2000 WebP but reports `meta.small` and `thumbnail.metaData` as 4200x2100 — exactly the bug class `REVIEW.md` → Stored media names and that #1334 fixed on the original path.

Add an above-cap thumbnail case (`MAX_WIDTH + 200` x `(MAX_HEIGHT + 200) / 2`): verified to pass on this branch and fail under the mutation.

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -627,6 +627,192 @@ describe('S3FileStorage saveFile with a video', () => {
...
+      .png()
+      .toBuffer()
+    return new File([new Uint8Array(buffer)], 'route-map.png', {
+      type: 'image/png'
+    })
+  }
+
+  // Thumbnails are uploaded under the same prefix as the original, suffixed
+  // `-thumbnail`, so each helper has to pick out the object it means.
+  const uploaded = (kind: 'original' | 'thumbnail') => {
```
llun, line 741: Minor: `uploaded('original')` is dead — the helper is only ever called with `'thumbnail'`, so the captured original/video body is never asserted (which is also why the `Buffer.isBuffer(Body)` branch of the mock has no observable effect on any assertion). Either use it or drop the parameter.

docs/architecture.md
```diff
@@ -176,6 +176,14 @@ and `PATCH /api/v1/profile`), custom emojis, thumbnail replacement
 and the fitness import jobs — which cover both the route maps the server
 generates and the arbitrary-sized activity photos it fetches from Strava.
 
+The sync upload endpoint also accepts an optional `thumbnail` beside the file,
+and both storage drivers treat it identically: it is re-encoded and stored the
+same way, recorded on the media row from the **stored** file's size and
+dimensions (so it is metered against the account's storage usage), and surfaced
+as the attachment's `meta.small` and `preview_url`. For a video it replaces the
```
llun, line 183: **"replaces the frame the server would otherwise extract" overstates it.** CONFIRMED, minor.

The frame is still extracted: `_uploadVideoToS3` runs `extractVideoImage` unconditionally (and so does `localFile._saveVideoFile`) before `saveFile` ever looks at `media.thumbnail` — a supplied thumbnail only causes the result to be discarded. The current wording reads as if the ffmpeg run is skipped, which matters because that run carries a 30-second timeout and is the expensive part of a video upload. "replaces the extracted frame" would be exact.

lib/services/medias/S3StorageFile.ts
llun: Fixed in f8ceda68b. Both drivers now refuse a non-image thumbnail with a `MediaValidationError` (→ 422) at the top of `saveFile`, before the quota check and before anything is uploaded, so no original is left orphaned. `saveFile` now matches `saveThumbnail`, which has always guarded on `startsWith('image')`.

Covered by `refuses a thumbnail that is not an image before storing anything` in both `S3StorageFile.test.ts` and `localFile.test.ts` — each asserts the rejection *and* that nothing was uploaded/written and `createMedia` was never called. Verified both fail when the guard is deleted.

lib/services/medias/S3StorageFile.ts
llun: Fixed in f8ceda68b — you are right that the video arm can never yield null. Reworded to say the image arm is what makes `previewImage` null here, and that a video whose frame cannot be decoded rejects rather than resolving null.

lib/services/medias/S3StorageFile.ts
llun: Fixed in f8ceda68b. Added `keeps the description and focus alongside the thumbnail`, asserting both on the `createMedia` payload and on the returned attachment (`description`, `meta.focus`). Verified it fails when both spreads are deleted — previously that mutation left all 276 tests green.

lib/services/medias/S3StorageFile.test.ts
llun: Fixed in f8ceda68b. Added `records an above-cap thumbnail at the dimensions it was stored at`: a 4200x2100 thumbnail beside an 800x600 original, asserting `createMedia` records 4000x2000, that the uploaded bytes really are 4000x2000, and that the original is bounded independently and stays 800x600. Verified it fails under the `outputInfo` → `metaData` mutation, which the previous cases could not catch.

lib/services/medias/S3StorageFile.test.ts
llun: Fixed in f8ceda68b — the new above-cap case uses `uploaded('original')` to assert the original is bounded independently of the thumbnail, so both arms of the helper are now live. The `Buffer.isBuffer(Body)` branch of the mock stays: a video body is a real `Buffer`, and `for await` over one iterates it as bytes rather than chunks, so dropping the branch would silently corrupt the captured video upload.

docs/architecture.md
llun: Fixed in f8ceda68b — reworded to "replaces the extracted frame — the frame is extracted either way, so it saves no work". Also added a line recording that a non-image thumbnail is refused before anything is stored, per the first thread.

llun:
Round 1 — the two findings from the review body that had no anchorable line are also fixed in f8ceda68b:

- **`localFile.test.ts` had no video-precedence coverage.** Added `prefers a caller-supplied thumbnail over the extracted video frame` and `falls back to the extracted video frame when no thumbnail is supplied`, plus the non-image refusal case. Verified the precedence test fails when the precedence is inverted in `localFile.ts` — previously that mutation left all 15 tests in the file green, so the exact bug this PR fixes on S3 could have been reintroduced on the local driver undetected.
- **Stale comment at `S3StorageFile.test.ts:602`** ("a separate `createMedia` call from the video branch") — rewritten to say the branches share one call and the image branch needs its own coverage because it reaches it through `_uploadImageToS3`.

Every new assertion was mutation-checked: `outputInfo`→`metaData` on the thumbnail dims, deleting the `description`/`focus` spreads, deleting the non-image guard (both drivers), and inverting the local precedence each fail exactly one test and no others. Gate re-run green: prettier, `eslint .` 0 errors, `next build`, 737 files / 7605 tests.

Starting review round 2 on the updated diff.

lib/services/medias/S3StorageFile.ts
```diff
@@ -406,6 +406,14 @@ export class S3FileStorage implements MediaStorage {
     if (!file.type.startsWith('image') && !file.type.startsWith('video')) {
       return null
     }
+    // `MediaSchema.thumbnail` is a `FileSchema`, which accepts every entry of
+    // ACCEPTED_FILE_TYPES — video and audio included. Refuse an unusable one
+    // here, before anything is uploaded: reaching sharp with it rejects with a
+    // plain Error, which is a 500 rather than a 422 and leaves the already
+    // stored original behind with no `medias` row to reclaim it by.
+    if (media.thumbnail && !media.thumbnail.type.startsWith('image')) {
+      throw new MediaValidationError('Thumbnail must be an image')
```
llun: **Round 1's guard is MIME-string-only, so the 500 + orphaned original is still reachable.** CONFIRMED — reproduced with a throwaway vitest against the real `S3FileStorage`.

`media.thumbnail.type` is a client-supplied claim, not the bytes. sharp is the actual arbiter and takes nothing on trust.

Failure scenario: `POST /api/v1/media` with a valid PNG as `file` and arbitrary non-image bytes as `thumbnail`, declared `image/png` (`curl -F "thumbnail=@notes.txt;type=image/png"`). `MediaSchema` accepts it, this guard passes, **the original is PUT to the bucket**, then `_uploadImageToS3` → `sharp(...).toFile()` rejects with a plain `Error: Input buffer contains unsupported image format`. `createMedia` is never reached → 500, and the stored original is orphaned with no `medias` row to reclaim it by. Byte-for-byte the round-1 defect, triggered by a mislabeled file instead of an honestly-labeled one.

A stricter type check cannot fix this — the fix is ordering or cleanup: reclaim the stored original when thumbnail processing throws, and report a caller-supplied thumbnail's failure as the invalid input it is. Applies to `localFile.ts:141` identically.

Consequently `docs/architecture.md:186` and the source comments here overstate what the code does: the "cannot leave an orphaned original behind" claim currently holds only for a truthfully-declared MIME type.

llun (commented):
Sub-agent review round 2 — two reviewers over the round-1 fixes and the full diff. Round 1's guard turns out to be incomplete: it trusts a client-declared MIME type, so the 500 + orphaned original is still reachable. Findings inline.

Also noted, not flagged as findings: `localFile.test.ts:462` duplicates a `createPngFile` identical to the one at :278 (the file already carries three variants, one per describe — consistent with how it is organised, so left alone), and `app/api/v1/media/[id]/route.ts:174`'s comment claims `FileSchema` validates "an image File" when it accepts every `ACCEPTED_FILE_TYPES` entry (that route still answers 422 correctly, via `saveThumbnail` returning null — pre-existing, another file, not touched here).

lib/services/medias/S3StorageFile.ts
```diff
@@ -406,6 +406,14 @@ export class S3FileStorage implements MediaStorage {
...
+    }
 
     // Check quota before saving
     const quotaCheck = await checkQuotaAvailable(
```
llun, line 421: **Thumbnail bytes are metered but not quota-checked — and this PR makes that newly reachable on S3.** CONFIRMED.

`checkQuotaAvailable(..., file.size)` counts only the original, while `createMedia` increments usage by `original.bytes + thumbnail.bytes` (`lib/database/sql/media.ts:197`). Verified scenario: report usage as `limit - file.size - 1`, upload file + thumbnail → the save succeeds and the account's counter ends **above** its quota (the next upload is then refused).

I called this out as out of scope in the PR body on the grounds that it was symmetric across drivers. That reasoning no longer holds: on `main` the S3 driver dropped the thumbnail entirely, so this PR is what makes the overshoot reachable there. One line, both drivers: `file.size + (media.thumbnail?.size ?? 0)`. Over-reserving slightly (the uploaded thumbnail is bigger than the stored WebP) is the safe direction.

lib/services/medias/localFile.test.ts
```diff
@@ -458,4 +458,74 @@ describe('LocalFileStorage.saveFile with a video', () => {
...
+    const files = await fs.readdir(mediaRoot)
+    const match = files.find((file) => file.endsWith('-thumbnail.webp'))
+    if (!match) throw new Error(`No stored thumbnail in [${files.join(', ')}]`)
+    return sharp(await fs.readFile(path.join(mediaRoot, match))).metadata()
+  }
+
+  // Which of the two thumbnail sources wins was untested on this driver, so the
+  // precedence could be inverted here — reintroducing exactly the bug the S3
+  // driver had — with nothing failing.
+  it('prefers a caller-supplied thumbnail over the extracted video frame', async () => {
```
llun: **Missing the file-count assertion its S3 twin has, which leaves a plausible refactor undetected — and order-dependent.** CONFIRMED by mutation.

A refactor that computes both thumbnails and then picks the supplied one:

```ts
const supplied = media.thumbnail ? await this._saveImageFile(media.thumbnail, { isThumbnail: true }) : null
const extracted = previewImage ? await this._saveImageBuffer(previewImage, { isThumbnail: true }) : null
const thumbnail = supplied ?? extracted
```

orphans the extracted frame on disk forever — no `medias` row references it, so `deleteMediaForAccount` never reclaims it. The full local suite passes **18/18** under that mutant, and running this test alone 8 times gave **4 passes / 4 failures**: `readStoredThumbnail` takes the first `-thumbnail.webp` in `readdir` order, which is not creation order. The S3 counterpart is immune because it asserts `expect(uploads).toHaveLength(2)`.

Add `expect(await fs.readdir(mediaRoot)).toHaveLength(2)` — it closes the gap and removes the nondeterminism.

lib/services/medias/localFile.test.ts
```diff
@@ -458,4 +458,74 @@ describe('LocalFileStorage.saveFile with a video', () => {
...
+    })
+
+    // 400x300 rather than the 1x1 extracted frame.
+    await expect(readStoredThumbnail()).resolves.toMatchObject({
+      width: 400,
+      height: 300
+    })
+  })
+
+  it('falls back to the extracted video frame when no thumbnail is supplied', async () => {
```
llun: This case and the one above it differ only by input and expected value — same `clip.mp4` setup, same `readStoredThumbnail()` → `toMatchObject` assertion shape — so `AGENTS.md` → Testing Guidelines calls for a table-driven `it.each([...])('$description', …)`. The sibling describe in this same file already uses that pattern at :307.

Separately, on its own this case kills no mutant the suite doesn't already kill: removing the `previewImage` fallback also fails the pre-existing `keeps a traversing file name inside the media root` (:427, `toHaveLength(2)`). Folding the pair into one table keeps the intent explicit without pretending it is independent coverage.

(The S3 pair does **not** qualify — `prefers…` there carries an extra `expect(uploads).toHaveLength(2)`, a distinct assertion shape.)

lib/services/medias/S3StorageFile.test.ts
```diff
@@ -599,8 +600,9 @@ describe('S3FileStorage saveFile with a video', () => {
     )
   })
```
llun, line 603: Minor: the round-1 rewrite of this comment is still imprecise. A **video** upload that carries a caller-supplied thumbnail now also calls `_uploadImageToS3`, so "only the image branch reaches it through `_uploadImageToS3`" is not quite right. The claim you want is that the *original*'s `path`/`metaData` come from a different helper per branch.

lib/services/medias/S3StorageFile.ts
llun: Fixed in fe7f58963. You are right that a type check cannot close this, so the fix is cleanup, not a stricter guard: thumbnail processing is now wrapped in both drivers, and on failure the stored original is deleted before the error propagates. A caller-supplied thumbnail's failure is rethrown as `MediaValidationError`, so the mislabeled-bytes case answers 422 like every other bad input instead of 500; a failure encoding an extracted video frame still rethrows as-is (not the client's input), but the original is reclaimed either way.

The cheap type guard stays as the first half — it avoids storing the original at all for an honestly-declared video, and keeps `saveFile` consistent with `saveThumbnail`.

New test on both drivers: `reclaims the stored original when the thumbnail is not a readable image` — non-image bytes declared `image/png`, asserting `MediaValidationError`, no `createMedia` call, and that the original was deleted (S3 asserts the `DeleteObjectCommand` key equals the uploaded original's; local asserts the media root is empty). Verified both fail when the reclaim is removed, and again when the 422 remap is removed.

`docs/architecture.md` reworded to match: refused up front if it is not an image, caught by the encoder if it only claims to be, original reclaimed either way, 422 not 500.

lib/services/medias/S3StorageFile.ts
llun: Fixed in fe7f58963 — `file.size + (media.thumbnail?.size ?? 0)` on both drivers. Agreed the earlier out-of-scope call no longer held once the S3 driver started storing thumbnails at all; I have corrected that in the PR body.

Covered by `counts the thumbnail against the account quota` on both drivers: usage is set to exactly `getQuotaLimit() - file.size`, so the upload is refused with its thumbnail and the same upload without one still succeeds — which pins the boundary on the thumbnail rather than on an arbitrary threshold. Verified it fails when the pre-check goes back to `file.size` alone.

lib/services/medias/localFile.test.ts
llun: Fixed in fe7f58963 — the table now asserts `expect(await fs.readdir(mediaRoot)).toHaveLength(2)` before reading the thumbnail, so a variant that writes both and picks one fails on the count rather than on `readdir` ordering. Good catch on the flake: the `.find()` was only deterministic because exactly one thumbnail can exist today.

lib/services/medias/localFile.test.ts
llun: Fixed in fe7f58963 — merged into an `it.each` with a `description` column and a `suppliedThumbnail` input, matching the pattern at :307. Agreed on the coverage point: with the file-count assertion added, the fallback row now pins something the traversal test only covers incidentally.

lib/services/medias/S3StorageFile.test.ts
llun: Fixed in fe7f58963 — reworded to "the original's path and metadata come from a different helper in each, so the image branch needs its own coverage."

llun:
Sub-agent review round 3 — two reviewers (adversarial on the round-2 error path; fresh-eyes on the whole diff). Both independently found that round 2 made things worse in one respect. Findings and resolutions, all in 71a18be4d:

**1. The catch relabelled every failure as bad input, unlogged. CONFIRMED by both, reproduced.** Round 2 wrapped thumbnail processing and rethrew `MediaValidationError` whenever a thumbnail was supplied — but `s3client.send`, `fs.open` and `toFile` are all inside that try, so an S3 5xx, expired credentials or `ENOSPC` became a 422 with the original deleted, the client told its perfectly good thumbnail was rejected (a permanent error it will not retry), and no log at all, since `handleSyncMediaUpload` only calls `logger.error` on the non-validation branch. A bucket outage would have looked like user error.

Fixed by validating up front rather than classifying after the fact: `sharp(...).metadata()` is a header parse, cheap enough to run before anything is stored, and it is definitive about whether bytes are an image. Unusable input is now refused with nothing written; anything failing after that point keeps its own error and stays a logged 500. The reclaim stays for those.

**2. `saveThumbnail` still 500d on the exact bytes `saveFile` now 422s on. CONFIRMED.** `PUT /api/v1/media/:id` is the dedicated thumbnail endpoint and had no such check. It now shares the same validation.

**3. The fix was copy-pasted into both drivers — reproducing the drift this PR exists to fix. CONFIRMED.** The shared checks moved into `lib/services/medias/thumbnailInput.ts`, imported by both. Each driver’s `saveFile` is smaller than before this round.

**4. The quota comment was inverted. CONFIRMED with measurements** — a lossy JPEG re-encodes into a *larger* near-lossless WebP (up to ~13x on photo-like input), so the reservation is best-effort, not conservative. Comment corrected to say so, and to note `original.bytes` under-counts the same way and always has. I have also corrected the PR body, which still carried my round-1 "out of scope" reasoning.

**5. `createMedia` returning null left both stored files unreachable. CONFIRMED, low.** Two lines below the new reclaim, so it now reclaims there too.

**6. Test placement/shape. CONFIRMED, minor.** Four image-only tests were living under `describe(... with a video)`; they move to a `LocalFileStorage.saveFile with a caller-supplied thumbnail` describe mirroring the S3 file. The S3 precedence pair becomes an `it.each` like its local twin.

Not fixed, with reasons: the S3 test file still carries its own `createPngFile`/`beforeEach` rather than sharing the `image sizing` block’s — the new block needs video mocks and delete capture that block deliberately lacks, and merging them would rewrite passing tests. `image: resizedImage` in `_uploadImageBufferToS3`/`_saveImageBuffer` is returned and read by nobody, but it is pre-existing and unrelated to thumbnails.

Also surfaced and filed separately rather than widened into this PR: `LocalFileStorage._saveVideoFile` writes the video into the media root and *then* runs `extractVideoImage`, so an ffmpeg failure orphans the video — the same class of leak this round fixed elsewhere, but on a path this diff does not touch (the S3 driver extracts before the PUT and has no such gap).

New invariants recorded in `REVIEW.md`: the two drivers must answer the same input identically, and a write that loses its row must be reclaimed rather than reported as client error.

Each fix mutation-checked: removing the byte probe, removing it from `saveThumbnail`, removing the missing-row reclaim, and restoring the round-2 relabelling each fail exactly the tests that cover them. Gate green: prettier, `eslint .` 0 errors, `next build`, 737 files / 7614 tests.

Starting round 4.

llun:
Sub-agent review round 4 — two reviewers (adversarial on the round-3 error path; regressions + project rules across the whole diff). Both landed on the same substantive finding. All fixed in 140b432c8.

**1. The round-3 probe was header-only, so a truncated image still became a logged 500 — and on S3 that was worse than `main`. CONFIRMED by both, measured.** `sharp().metadata()` parses the header: a PNG or JPEG truncated to 90/60/40/20% still reports `400x300`, and only the encoder trips over the missing bytes. So the "validated up front, therefore anything failing later is ours" reasoning did not hold for the one input class most likely to occur, and `main` answered that same request with a 200 (thumbnail dropped).

I measured all three sharp calls across those truncations: `stats()` rejects every case `metadata()` accepts, and unlike `toBuffer()` it decodes without re-encoding. So the cheap header check stays up front (bytes that are not an image at all are refused with nothing stored), and the catch now runs `stats()` to classify: unreadable bytes are the caller's 422, anything else keeps its own error and stays a logged 500. `saveThumbnail` makes the same split, which also closes finding 2 below.

**2. `saveThumbnail` still 500d on the exact bytes `saveFile` 422s on. CONFIRMED.** Same classifier, so `PUT /api/v1/media/:id` now agrees with the upload endpoints.

**3. `createMedia` *rejecting* was not reclaimed — only a null return was. CONFIRMED, reproduced.** A deadlock or lost connection left both files stored with no row. Both drivers now share a `_reclaimStored` helper used on all three failure paths.

**4. `REVIEW.md` said the divergence "federated a different `meta.small`". CONFIRMED wrong.** `getMastodonAttachment` hardcodes `preview_url: null` and the AP `Document` has no thumbnail field — only the upload-response entity differed. Corrected in `REVIEW.md`, `docs/architecture.md` and the PR body.

**5. The PR body contradicted the code. CONFIRMED.** It still carried my round-1 "quota pre-check unchanged, out of scope" note after round 3 changed exactly that, and its file list omitted `localFile.ts`, `thumbnailInput.ts` and `REVIEW.md`. Rewritten.

**6. The PR's own new REVIEW.md rule ("a change to one driver's `saveFile` needs the matching test in both") was not met. CONFIRMED.** The local driver had no reclaim-on-storage-failure test; the truncated-image case now drives that branch on both.

**7. `getUploadQuotaReservation` was a quota helper in a validation module.** Moved to `medias/quota`.

Also added `thumbnailInput.test.ts` pinning the header-vs-decode distinction directly — including the case that documents why both checks exist, so that if the cheap one is ever made a full decode the second is deleted rather than both kept.

Worth noting: `next build` caught a type error the whole suite did not (`sharp.Sharp` is not a namespace under this config) — a reminder that `yarn test` passing is not a typecheck.

Mutation-checked again: removing the decode classifier, weakening `stats()` back to `metadata()`, removing the row-write reclaim, and removing the `saveThumbnail` classifier each fail exactly the tests that cover them. Gate green: prettier, `eslint .` 0 errors, `next build`, 738 files / 7627 tests.

Starting round 5.

llun:
Sub-agent review round 5 — one adversarial reviewer on the round-4 error path, one maintainer-fresh-eyes pass on the whole diff.

**The adversarial pass came back clean** and said the loop can stop. It verified, with measurements rather than reading: a pixel bomb is refused up front (`metadata()` does enforce `limitInputPixels` — 30000x30000 throws in 1ms); truncated PNG/JPEG at 30/60/90% leave no partial file on disk, so the "nothing stored" assertions are honest; a thumbnail that decodes but fails to write is a 500 with the original reclaimed; `_reclaimStored` resolves to the same files both drivers wrote; and every doc/commit claim checks out against the code. Its only finding was the pre-existing local video orphan already filed as a follow-up.

**The fresh-eyes pass found no bug either, but pushed back on the design** — and it was right, so I took it:

> the two exported names are synonyms in ordinary English… you have to read both bodies and the comments to tell them apart. Then you meet the second one inside a `catch`, which is where "patch on patch" is the honest first impression.

It measured the tradeoff instead of asserting it: `stats()` rejects exactly what the encoder rejects, on both formats, for ~15% of the encode it precedes (38ms vs 253ms on a 12MP image). So in 1d3aae6e8 there is now **one** check, up front: `readValidThumbnail` (declared type + full decode). That deletes the second exported function, four try/catch classifier blocks, their `let` temporaries, and four copies of the same explanatory comment. `thumbnailInput.test.ts:47` had already written the argument down — *"Anything that changes this to a full decode should delete the second function rather than leave both."*

Two things came free with it: `readValidThumbnail` returns the bytes it read, so the drivers store them via `_saveImageBuffer`/`_uploadImageBufferToS3` instead of reading the file again (two fewer full-size copies per upload, on a service with a history of worker OOM), and the thumbnail/preview-frame ternary collapses now that both arms are buffers.

Also taken: `_reclaimStored(originalPath, thumbnailPath?)` instead of variadic-plus-filter; the quota comment now points at `getUploadQuotaReservation` rather than implying reserved and metered are the same number; the duplicate garbage-bytes driver test dropped per driver (the truncated case is the stronger claim and the unit test pins the distinction); `expect(uploads).toHaveLength(0)` added to the S3 standalone-thumbnail case; the `docs/architecture.md` sentence trimmed to the behaviour, leaving the mechanism to the module.

**Mutation checking caught something the review did not:** removing the declared-type guard entirely broke no test, because a video-typed thumbnail fails the decode anyway. The guard is still worth keeping — it is what makes `saveFile` agree with `saveThumbnail`, which refuses a non-image type outright — so it is now pinned by the only input that distinguishes it: real PNG bytes declared `video/mp4`. Four other mutations (`stats()`→`metadata()`, dropping the reclaim, inverting precedence, and the ones from earlier rounds) each fail exactly the tests that cover them.

Not taken, with reasons: the S3 test block keeps its own `createPngFile`/`beforeEach` rather than sharing the `image sizing` block’s — it needs video mocks and delete capture that block deliberately lacks. The suggested "encode the thumbnail to a buffer before storing the original" variant would remove the last of the error handling, but it needs an encode/write seam in each driver and the current shape is already unambiguous.

Gate: prettier, `eslint .` 0 errors, `next build`, **738 files / 7623 tests**.

Five rounds, each of which found the previous round’s fix incomplete; this one found no defect in the diff and the changes above are readability and one test gap. Stopping here.

PR #1342, opened 2026-07-28
minor: match the design system's full-width post action row and reactions

llun:
Brings the status action row and the emoji reactions in line with the web UI kit (`ui_kits/web/Post.jsx` + `Reactions.jsx`).

## What changed on screen

**The picker moved into the action row.** It used to be a lone round `SmilePlus` button tacked onto the end of the chip row. It is now an action next to reply/boost/like — `ReactionButton`, carrying the post's running reaction total — which is what the kit does and what `Reactions.jsx` calls out explicitly ("the picker trigger MOVED into the action bar … instead of a lone smiley button in the status body"). Two consequences fall out of that:

- A post with **no** reactions yet still offers one, because the trigger no longer depends on the chip row existing.
- The chip row is now purely a read-out, so it renders **nothing at all** on an unreacted post instead of an empty row with a lone button in it.

**Both rows span the whole status.** Each is pulled `-ml-13` — 13 spacing steps, which is the `size-10` avatar column plus its `gap-3`, so the pull tracks the root font size the way those two already do — and each starts at the post's own left edge rather than under the text. The action row is `justify-between` so the actions are evenly spaced with the ⋯ menu pinned to the far right. That meant taking `ml-auto` off the ⋯ wrapper: an auto margin absorbs all the free space *before* `justify-content` gets to see it, which would have quietly collapsed the row back to a left-packed cluster.

The chips only go full-bleed when there is an action row under them to align with. On a read-only surface the row beneath is `ReadOnlyStats`, which starts at the text, so chips pulled further left would line up with nothing.

**A narrow post hands bookmark and react to the ⋯ menu**, as "Bookmark" and "React to post" above the menu's own items; reply, boost and like keep their place. The width comes from a `ResizeObserver` on the row itself rather than a viewport breakpoint — a post can sit in a narrow column, a modal or an embed on a wide window, and a viewport query would collapse the wrong ones. Where `ResizeObserver` is missing (server render, jsdom) the full row is the default and the first client layout corrects it before paint.

## What that forced in the code

Splitting one control across two places means its state can no longer live inside either of them:

- **`useReactionState`** owns the rollups, the single-flight write, the optimistic update and the picker anchor. `Post` holds it; `FitnessStatusDetail` holds its own, because that page composes its own action row and so now renders both halves (`ReactionRow` **and** `ReactionButton`) explicitly.
- **`useBookmarkState`** does the same for the bookmark, held by `Actions`. This is not tidiness: on a narrow post the bookmark is a menu item rather than a button, and if each spelling kept its own `useState` then bookmarking on a wide window and then narrowing it would show the menu offering "Bookmark" for a status that already is one.
- **`PostMenu`** takes `extraItems` and a `triggerRef`. The ref is what lets the picker anchor to the ⋯ button when it is opened from there. `deferUntilClosed` on an item runs it from `onCloseAutoFocus`, which also suppresses Radix's own focus restore — otherwise that restore lands after the panel has taken focus and pulls it straight back to the ⋯ trigger. Items also carry `disabled` while their own write is in flight: a menu item has none of the busy styling the button it replaced had.

The action row's `role="group"` is now `"Post actions"` on the row itself rather than `"Post primary actions"` on an inner wrapper — the wrapper had to go (it would have been a single flex item and defeated `justify-between`), and the ⋯ menu is an action too, so one honestly-named group is the accurate description.

## One judgement call worth confirming

The kit's bar is `justify-between`, and the outstanding note on the design project says *"All actions should have the same spacing"* — so that is what this implements: evenly distributed actions across the full width. Reading the attached mockups alone, they can also be read as tightly-grouped actions on the left with only the ⋯ pushed right. Say the word and it's a one-line change (`gap-6` on the row plus `ml-auto` back on the ⋯ wrapper).

## Browser verification

Local dev server on port 3100 against a throwaway SQLite `dev.sqlite3` (`ACTIVITIES_DATABASE_CLIENT=better-sqlite3`, `ACTIVITIES_HOST=localhost:3100`), signed in as the sanctioned mock `testuser`, with three reaction rollups seeded on one post (👍 12, ✨ 4, 🤔 2 — 18 total, one of them the viewer's own).

Desktop (1280px), timeline and status detail:

- Chip row and action row both start at x=336, the article's own content edge (article at x=320 + `px-4`) — full-bleed, exactly aligned with each other.
- Row children measured at x = 336 / 616 / 897 / 1178 across an 874px row: evenly distributed, ⋯ flush right.
- The react action reads `Add reaction, 18 reactions` and shows `18`; the viewer's own 🤔 chip renders in the "mine" style.
- Opening the picker from the action button anchors it under the button; picking 😀 issued the write, the chip appeared as `😀 1` and the total went 18 → 19.

Mobile (375px), fresh load:

- Row collapses to reply / boost / like / ⋯ — measured children x = 33 / 125 / 218 / 310 in a 309px row, evenly spaced with **no empty slot** where the trigger used to be (see the last commit — that was a real bug the measurement caught).
- The ⋯ menu opens with "React to post" and "Bookmark" at the top, above the separator and the menu's usual items, matching the mockup.
- Selecting "React to post" opens the picker anchored to the ⋯ trigger. (The search field's focus is pinned by a test rather than by eye — round two found that it was *not* focused in a real browser, because the panel is `visibility: hidden` until measured and jsdom focuses regardless.)

I could not attach the captures to this description — there is no upload path from the CLI — so the measurements above are the substantive evidence.

Worth recording for the next person: `ResizeObserver` callbacks are only delivered when the browser pane actually paints, so reading the DOM straight after a programmatic resize returns the *stale* tree. Every "the compact mode isn't reacting" reading here turned out to be that, not the app.

## Verification

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **739 files / 7622 tests green**, including new coverage for the action-bar trigger, the compact collapse, opening the picker from the menu, bookmarking from the menu (which asserts the state is shared rather than a second copy), the width threshold and its zero-width guard, both compact error surfaces, and the fitness detail's reactions.

## Review

Three rounds of sub-agent review, recorded in the comments below: **14 findings, all addressed.** Round one found eight, including a pre-paint flash on every narrow post and a picker that opened with focus on `<body>` in a real browser (jsdom focuses regardless, which is why the test passed). Round two found three, two of them introduced by round one's own fixes. Round three found nothing in the production code but mutation-tested every test the earlier rounds added and caught three that passed with their own fix reverted; those are repaired and re-checked by reverting each fix in turn.

Docs updated in the same PR: **Status Posts & Actions** in `AGENTS.md`, the matching `CLAUDE.md` reminder, the `REVIEW.md` checklist, and the reactions line in `docs/features.md`.


llun:

Four reviewers over the whole diff: React correctness (hooks/refs/focus/races), project invariants + a11y, test coverage, and CSS/layout + cross-surface impact. Every finding below is recorded with its resolution; all of them were in code this PR introduced, and all are fixed in 1dec8915b.

### Confirmed and fixed

1. **`lib/components/posts/actions/useCompactActionBar.ts` — the compact row was applied after the first paint.** The layout effect only *subscribed*; the first `ResizeObserver` delivery lands in the browser's rendering step and the `setState` it schedules is not discrete, so React ran it after the frame painted. Every narrow post flashed the full six-control row and then dropped two buttons out of it and re-spread the rest. Now measured in the layout effect, where a `setState` is flushed before paint.

   Fixing that surfaced a second problem: a zero width is what a `display: none` ancestor, a detached subtree and jsdom all report, and reading it as "narrow" collapsed the row in every unrelated test that stubs `ResizeObserver` for some other component (`ExplorePageClient`, `BookmarksTimeline`). Zero is now ignored rather than treated as a measurement.

2. **`lib/components/posts/reaction-picker.tsx` — the picker opened with focus on `<body>`.** It relied on `autoFocus`, but the panel is `visibility: hidden` until it has been measured and a browser will not focus inside a hidden subtree. jsdom focuses regardless, which is exactly why the existing test passed and only a real browser would have shown it. It now focuses once placed, and only once, so a scroll-driven reposition cannot pull focus off an emoji the viewer has tabbed to. This also makes `deferUntilClosed`'s stated rationale true rather than aspirational; `post.test.tsx` now asserts the search field has focus after opening from the menu.

3. **`lib/components/posts/actions/actions.tsx` — the compact menu items bypassed the single-flight guard.** In the row the trigger is guarded twice and shows a busy state; the "React to post" / "Bookmark" menu items had neither, so a tap during an in-flight write was swallowed by `pendingName` with nothing on screen to explain it. Both now carry `disabled` while their own write is pending (new `PostMenuExtraItem.disabled`).

4. **`lib/components/posts/actions/actions.tsx` — a failed bookmark was silent on a compact row.** `BookmarkButton` is the only thing that renders `bookmark.error`, and it is not in the row at that width. The row renders it instead. Both this and the reaction error are `position: absolute`, so they add no flex item and cannot skew the `justify-between` spacing — which also fixes the reviewer's separate point that the reaction error's wrapper made the row jitter for the 4s the message was up.

5. **`lib/components/posts/actions/reaction-button.tsx` — an open picker was stranded by a width flip.** The panel is placed from the anchor's rect when it opens and `usePanelPosition` does not re-run on a ref swap, so crossing the threshold with the picker open left it hanging over the button that had just unmounted. It closes on the flip.

6. **`lib/components/posts/actions/edit-history-button.tsx` — the history panel now overflowed the post.** It is `absolute left-0` and `25rem` wide; `justify-between` moved its trigger to the right end of the post, so the panel ran past the right edge and was clipped by the card (`overflow-hidden` on the detail card, `overflow-x-clip` on the timeline `main`, the search results section). It opens leftward now. Desktop/tablet only — below `md` it is `fixed inset-x-4`.

7. **`lib/components/posts/post.tsx` — `fullBleed` was unconditional, so read-only posts had misaligned chips.** On any `showActions={false}` surface there is no action row to align with, and the row that *is* underneath (`ReadOnlyStats`) starts at the text — so the chips hung 52px further left than everything else. The landing feed, logged-out profile/hashtag/collection, admin hashtag posts, and every reply row on the status detail page were affected. `fullBleed` is now tied to the same gate `Actions` applies to itself.

8. **`app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx` — the fitness action row could overflow.** Adding the react trigger made it a six-control non-wrapping row inside `px-4`; on a ~320px phone that overflows and the fitness card's `overflow-hidden` clips the ⋯ menu. It wraps now.

### Also taken

- The hardcoded `-ml-[52px]` is now `-ml-13` (13 spacing steps). The reviewer's point stands: the avatar is `size-10` and the gap is `gap-3`, both rem-based, so a fixed 52px drifted from the column it compensates for at any non-16px root font size. Verified the utility actually compiles — `.-ml-13{margin-left:calc(var(--spacing) * -13)}` is in the build output, which matters because a Tailwind class that doesn't generate fails silently.
- Test gaps closed: a `useCompactActionBar` unit suite (threshold boundary at 399/400, the zero-width guard, growing back to wide, a caller-supplied threshold, `disconnect` on unmount); the wide→compact transition asserting the bookmark taken while wide survives (which is the entire reason that state sits in the row); both compact error surfaces; the picker's focus outcome; `FitnessStatusDetail`'s reactions, which had none; and the singular "1 reaction" label, which the previous regex assertion could not catch.
- Docs corrected in `AGENTS.md` / `REVIEW.md` where they described the focus sequence that finding 2 shows does not happen.

### Reviewed and deliberately not changed

- **Action positions vary post-to-post within a feed** (different counts change button widths, and an edit-history button changes the divisor). That is inherent to `justify-between`, which is what the kit specifies and what "all actions should have the same spacing" asks for. Flagged in the PR description as the one judgement call worth confirming.
- **`app/(timeline)/[actor]/[status]/page.tsx:335` has a comment claiming the card has no `overflow-hidden`, immediately above the line that sets it.** Pre-existing and out of scope here; harmless for the reaction picker (portalled and `fixed`), but it is the `overflow-hidden` behind finding 6, so worth a follow-up.

### Verified as non-issues

The `triggerRef` hand-off between `ReactionButton` and `PostMenu` (React detaches changed refs before attaching, in both directions); `deferUntilClosed` vs. Radix's `DismissableLayer` cleanup; the `pendingNameRef` resync pattern (relocated verbatim, no loop); hooks ordering against the early returns; `'use client'` on `post.tsx` (every importer was already a client module, and `lib/clientModuleBoundary.test.ts` covers it); no `w-[--var]` bracket-variable form; RTL (the app sets no `dir`); the chip row's wrapping, which is strictly better now that the row is 52px wider; and no `overflow-hidden` ancestor clipping the negative margin — every `Post` wrapper has at least 16px of horizontal padding.

llun:

Two reviewers over the updated diff: one checking whether the round-one fixes introduced new defects, one fresh-eyes on everything the first round did not look at (behaviour drift vs. `main`, dead code, the new modules' public API, edge cases, commit messages). Both independently found the same three things. Fixed in 874577c4d.

### Confirmed and fixed

1. **`lib/components/posts/actions/reaction-button.tsx` — closing the picker on the width flip dropped focus on `<body>`.** Round one added an effect that closes the picker when the trigger moves to the ⋯ menu, because the panel is placed from its anchor's rect and does not re-measure on a ref swap. But every other close path in that file calls `focusTrigger()` for a reason, and this one didn't: a keyboard user whose window crossed the threshold with the picker open had the panel vanish and their next Tab restart at the top of the document. It hands focus to whichever trigger now exists.

   Doing that properly needed `focusTrigger` to become stable (`useCallback` in `useReactionState`) — it was an inline arrow recreated every render, so as an effect dependency it would have closed the picker the instant it opened. The effect reads `isPicking` through a ref so an ordinary open/close doesn't re-run it. `post.test.tsx` now covers the whole sequence: open while wide, resize to 320, assert the panel is gone and the ⋯ trigger has focus.

2. **`lib/components/posts/actions/actionButtonShared.tsx` — the two relocated errors shared one anchor.** Round one moved the bookmark and reaction errors into the row as `absolute right-0 top-full`, which is the same box for both — and the writes are independent, so both can be live inside one 4s dismiss window. Two opaque tooltips rendered exactly on top of each other, which is the same "a failure is invisible on a compact row" defect round one set out to close. New `ActionRowErrors` stacks them in one left-anchored column, which also keeps them clear of `PostMenu`'s own right-anchored error. Covered by a test asserting both errors share one parent.

3. **`AGENTS.md` / `CLAUDE.md` / `REVIEW.md` — the docs still prescribed `-ml-[52px]`.** The docs commit landed before the commit that replaced it with `-ml-13` and was never re-synced. `CLAUDE.md` is loaded at the start of every task in this repo, so the stale rule was an active invitation for the next agent to "restore" the hardcoded pixel value and undo the fix. All three updated, with the reasoning (13 spacing steps = the `size-10` avatar column + its `gap-3`) rather than just the new literal.

### Also taken

- `ReactionState` now exposes `canReact: boolean` instead of the whole `ActorProfile`. The single consumer only reads it as a truthiness test, and publishing the profile invites a future consumer to key off an actor id that the hook never reconciles against the status.
- Leftovers: `getReactionButtonLabel` and `CustomEmojiOption` are no longer exported (nothing outside their modules used them), and a test local still named `primaryActions` after the group was renamed to "Post actions".

### Reviewed and deliberately not changed

- **`useCompactActionBar`'s `threshold` parameter has no production caller.** It is a default parameter next to an exported constant and it is what makes the boundary testable; the alternative is a hook whose only threshold is a module constant that no test can vary.
- **`edit-history-button.tsx`'s `right-0` could clip leftward** on a row between ~400px and ~440px wide *at* `md`+. No surface in the repo renders a post in a ~400px container on a ≥768px viewport, and below the compact threshold the row is a different shape anyway.

### Verified as non-issues

The close-on-flip effect does not fire spuriously (`setIsPicking` is a stable dispatch, `hideTrigger` is a boolean, and the mount run bails out); `ReactionPicker`'s focus guard resets correctly because the component is conditionally mounted; `-ml-13` compiles to `calc(var(--spacing) * -13)` and nothing else referenced the old literal; the zero-width guard has no real-browser counter-example; `Announce` resolves to the same object in `Post` and `Actions`, so both halves share one status; every path that unmounts `ReactionButton` is also a path where `canReact` is false, so no error can be produced with nowhere to show it; and the commit messages match their diffs.

### Gate

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **739 files / 7619 tests green**. Re-verified in a real browser at 1280px and 375px after the fixes.

llun:

A verification round: does the branch hold up after eleven fixes, and would each new test actually catch the thing it was written for? The reviewer mutation-tested every test the previous rounds added, in a throwaway worktree.

**No correctness, accessibility or convention defects.** The areas checked, so the conclusion is auditable: `ActionRowErrors`' live-region behaviour and left anchoring against the full-bleed row; the close-on-flip effect traced through all six states (mount, open, close, flip-while-open, flip-while-closed, unmount) including whether `focusTrigger()` runs before or after the new trigger's ref attaches; `canReact` replacing `currentActor` (no consumer left on the old shape, `toggle` unchanged since it gates on the closed-over param); `'use client'` on `post.tsx`; the ⋯ separator hygiene; and `justify-between`'s child count with a hidden trigger and an empty error stack.

One adjacent race the reviewer chased and cleared: on the compact path `usePanelPosition`'s layout effect runs before `PostMenu`'s ref attaches in that commit, so the first `place()` sees a null anchor. It self-corrects because the panel's callback ref forces a second render and the panel stays `visibility: hidden` until then — no stranded panel, no flash.

### Three tests that passed with their own fix reverted

Fixed in e90fa40bc, and each re-checked by reverting its fix and confirming the test now fails.

1. **`post.test.tsx` — "stacks a bookmark and a reaction failure".** It asserted only that the two tooltips share a parent, which was true of the broken version too: both were already direct children of the action row. The defect was that each anchored *itself* to the same `right-0 top-full`. It now asserts neither positions itself and that the stack does.

2. **`useCompactActionBar.test.tsx` — "ignores a zero width".** It delivered 320, then 0, and expected "compact" both times — which is also what the unguarded code returns. It runs from both starting widths now; the wide one is the case the guard exists for. (The reviewer noted the guard was accidentally covered by a sibling test, via jsdom reporting 0 from the layout-effect measure — incidental coupling that would evaporate the moment that test's setup changed.)

3. **`FitnessStatusDetail.test.tsx` — "renders no chip row".** It queried by *text* for `/reaction,/`, which only ever appears in an `aria-label`, so it matched nothing whether or not chips rendered. What can actually regress is the empty `border-t` strip left behind when the wrapper is ungated — a bare rule with 10px of dead space — so that is what it asserts, via a testid on the wrapper.

### Coverage the reviewer named as missing, now added

- The `disabled` flags on both ⋯ menu items, including that they are independent rather than one shared busy flag (settle the bookmark, confirm the reaction item stays disabled).
- The row's own `-ml-13` / `justify-between` / no-`ml-auto` classes and the chip row's `-ml-13`. None of it is observable in jsdom and all of it is load-bearing — `ml-auto` on the ⋯ wrapper would silently absorb the free space and re-cluster the row, which is now a rule in three docs with nothing pinning it. `tailwindCssVariableSyntax.test.ts` is the precedent for guarding a class this way.

Left uncovered deliberately: the pre-paint `measure(...)` in `useCompactActionBar`, which jsdom cannot express (it reports 0 for every element, which is exactly what the guard ignores), and `edit-history-button.tsx`'s `right-0`, whose only failure mode is visual overflow at a width no surface in the repo produces.

### Gate

`prettier --write .` → `eslint .` (0 errors, 31 pre-existing warnings) → `next build` → `vitest run`: **7622 tests green**.

Three rounds, 14 findings, all addressed. Round three surfaced nothing in the production code, so this is where the loop stops.

PR #1343, opened 2026-07-28
fix: extract a video's preview frame before storing the video

llun:
## Summary

`LocalFileStorage._saveVideoFile` wrote the uploaded video into the media root with `fs.writeFile(filePath, buffer)` and only **then** called `extractVideoImage(filePath)`. That helper is typed `Promise<Buffer>` and rejects whenever ffmpeg finds no decodable frame; the rejection propagates out of `saveFile`, so `createMedia` is never reached. The video is left on disk with no `medias` row — and a row is the only handle anything has on a stored path, so nothing but `scripts/maintenance/cleanupMediaStorage.ts` can ever find it again.

The S3 driver never had that gap: `_uploadVideoToS3` extracted the frame before the `PutObjectCommand`.

## What changed

Both drivers now run the same sequence, and it is the sequence the S3 driver *should* have had:

1. probe the buffer (`extractVideoMeta`),
2. validate the container **from the probe alone**,
3. extract the preview frame from a temp copy,
4. only then store.

Nothing reaches the media root or the bucket until every step has succeeded, so there is nothing to reclaim on failure. `_saveVideoFile` and `_uploadVideoToS3` are now the same code up to the store step.

The temp-copy step lives in **`lib/services/medias/videoPreview.ts`**, which both drivers import — these two files are edited one at a time and drift silently, which is how the local driver came to have this gap in the first place.

### Two defects the review rounds turned up, both shared with the S3 driver

**The temp copy's name must be server-derived.** ffmpeg picks its demuxer from the path as well as from the bytes, and `image2`/`mjpeg` beat content probing for an image extension paired with a `%0Nd` pattern or a `*` glob. A perfectly good H.264 mp4 uploaded as `filename="IMG_%04d.jpg"` sent ffmpeg hunting for a numbered image sequence:

```
/tmp/<hex>.mp4            -> exit 0 (frame extracted)
/tmp/<hex>-IMG_%04d.jpg   -> exit 254  [image2] Could find no file with path ... and index in the range 0-4
/tmp/<hex>-vid*.jpg       -> exit 69   [mjpeg] dqt: len 55930 is too large
```

so the instance answered 500 for a file it can store fine. The path still comes from `createMediaTempFilePath` — random prefix, `tmpdir()` assertion — but the name handed to it is `video<ext>` from the validated content type. Taking nothing from the supplied name also makes the traversal question moot.

**The container has to be validated before the frame is extracted.** An audio-only `.mp4` — a voice memo, or an mp4 with its video track stripped, which the browser labels `video/mp4` from the extension — is exactly the input ffmpeg cannot pull a frame from. Deciding it from the probe keeps it a **422** that never spawns ffmpeg; extracting first turned it into a logged **500** the client would retry. This follows the invariant #1341 documents: validate input first, then let a genuine fault stay a logged 500. A failed extraction still propagates its own error unchanged — it is never relabelled as a `MediaValidationError`.

**Temp-copy ownership.** `wx` (O_EXCL) creates the file before the bytes go in, and `writeFile` does not remove what it created when the write itself fails — an ENOSPC partway through a 200 MB upload leaked the partial copy for the lifetime of the container, one per retry. The call opens the file explicitly so ownership is structural: the open is what creates the path, so everything after it is cleaned up and an EEXIST collision — the one case where the path is *not* ours — is left alone.

## Tests

- `videoPreview.test.ts` — the shared helper: the frame comes from a temp copy holding the uploaded bytes, the copy is removed on success, on extraction failure and on a mid-write failure, the name is server-derived, and a collision neither proceeds nor deletes the file it refused to clobber.
- `localFile.test.ts` / `S3StorageFile.test.ts` — the same guarantees per driver, plus the previously-uncovered `Invalid video format` branch and probe-failure path in both.
- The `extractVideoImage` mock in all three suites now **reads the file it is handed**, so an implementation that never writes the temp copy fails 20 tests instead of passing green.

Every fix in this PR was mutation-checked: reverting it fails a named test. That includes the original bug (before the fix, the regression test rejected correctly but left `<hex>.mp4` in the media root), the validate-first ordering, the ownership split, and the cleanup placement.

## Review

Four sub-agent rounds (correctness / conventions / tests, then docs+tests, then a whole-diff pass). Rounds 1–3 produced 13 findings, all addressed and resolved in the threads above; round 4 was clean.

## Notes

- Both drivers write the video buffer to `tmpdir()` as well as to storage. The S3 driver always did; it is the cost of extracting before storing, and it means a local-file deployment wants temp-dir headroom equal to `media.maxFileSize`. Worth knowing if `/tmp` is a tmpfs.
- The `REVIEW.md` "two storage drivers must answer the same input identically" bullet is being added by #1341 and is not duplicated here. #1341 also touches `saveFile` (thumbnail handling + `_reclaimStored`) but not the video path, so the two changes are complementary.
- `docs/` needed no change; the new invariant is recorded in `AGENTS.md`, `REVIEW.md` and `CLAUDE.md`.

## Gate

`prettier --write .` -> `eslint .` (0 errors) -> `next build` ✅ -> `vitest run` ✅ **738 files / 7609 tests passed.**


lib/services/medias/localFile.ts
```diff
@@ -315,10 +319,26 @@ export class LocalFileStorage implements MediaStorage {
...
+  // both read a temp copy, so nothing reaches the media root until both have
+  // succeeded. Writing the video there first and extracting from it afterwards
+  // left the bytes on disk whenever ffmpeg found no decodable frame, with no
+  // `medias` row — the only handle anything but
+  // `scripts/maintenance/cleanupMediaStorage.ts` has on a stored path.
   private async _saveVideoFile(videoFile: File) {
     const uploadPath = this._config.path
     const buffer = Buffer.from(await videoFile.arrayBuffer())
-    const probe = await extractVideoMeta(Buffer.from(buffer))
+    const tmpVideoFile = createMediaTempFilePath(videoFile.name)
```
llun: **Correctness — ffmpeg now picks its demuxer from the client-supplied file name, so a valid video can fail to upload.**

`extractVideoImage` used to receive the *final stored path*, whose extension comes from `getStoredMediaExtension` (always `.mp4`/`.webm` for video). It now receives `createMediaTempFilePath(videoFile.name)`, which preserves the client's own extension. ffmpeg's `image2`/`mjpeg` demuxers probe the **filename**: an image extension plus a `%0Nd` number pattern or a `*` glob beats content probing and the input never opens.

Verified with a real H.264 mp4, byte-identical in both cases:

```
old  /tmp/<hex>.mp4              -> exit 0 (frame extracted)
new  /tmp/<hex>-IMG_%04d.jpg     -> exit 254  [image2] Could find no file with path ... and index in the range 0-4
new  /tmp/<hex>-vid*.jpg         -> exit 69   [mjpeg] dqt: len 55930 is too large
```

**Failure scenario:** a non-browser client — the exact threat model `fileName.ts` documents — POSTs a good H.264 mp4 with `Content-Type: video/mp4` and `filename="IMG_%04d.jpg"`. Before this commit the upload succeeded; now `extractVideoImage` rejects, and `handleSyncMediaUpload` returns **500** for a file the instance can store fine.

This is parity with the S3 driver, which already had it — but it is new for the local driver, so the fix belongs on both sides.

lib/services/medias/localFile.ts
```diff
@@ -315,10 +319,26 @@ export class LocalFileStorage implements MediaStorage {
...
+    // `wx` (O_EXCL) so the write fails rather than following a symlink someone
+    // planted at the path, or clobbering an existing file. The 64-bit random
+    // prefix already makes that infeasible to aim at; this makes it impossible.
+    await fs.writeFile(tmpVideoFile, buffer, { flag: 'wx' })
+    // `finally` so a probe/preview failure still removes the temp file instead
+    // of leaking it for the lifetime of the container.
+    const [probe, previewImage] = await Promise.all([
```
llun: **Correctness — an mp4/webm with no video stream turns a clean 422 into a logged 500.**

Container validation now runs *after* the `Promise.all`, and `Promise.all` rejects on the first rejection, so the `MediaValidationError` below is unreachable whenever `extractVideoImage` also fails. The PR body waives only "ffprobe ok, container not mp4/webm, extraction fails". It does **not** cover an mp4 that *is* an accepted container but has no video stream — which is precisely the input ffmpeg cannot extract a frame from, so it is the systematic case, not a corner.

Verified end to end:

```
$ ffprobe -show_format -show_streams audio_only.mp4
format_name = mov,mp4,m4a,3gp,3g2,mj2      codec_types = ['audio']
$ ffmpeg -loglevel error -i audio_only.mp4 -frames:v 1 -y out.jpg
[out#0/image2] Output file does not contain any stream    exit=234
```

**Failure scenario:** a user uploads an audio-only `.mp4` (voice memo, or an mp4 with the video track stripped). The browser labels it `video/mp4` from the extension. Before: `!videoStream` -> `MediaValidationError` -> **422**, no subprocess spawned, nothing logged. After: ffmpeg exits 234 and the user gets a **500** the client will treat as retryable, plus a `logger.error` on every such upload.

REVIEW.md's own wording for this is "Validate input first, then let a genuine fault stay a logged 500."

lib/services/medias/localFile.ts
```diff
@@ -315,10 +319,26 @@ export class LocalFileStorage implements MediaStorage {
...
+  // `scripts/maintenance/cleanupMediaStorage.ts` has on a stored path.
   private async _saveVideoFile(videoFile: File) {
     const uploadPath = this._config.path
     const buffer = Buffer.from(await videoFile.arrayBuffer())
-    const probe = await extractVideoMeta(Buffer.from(buffer))
+    const tmpVideoFile = createMediaTempFilePath(videoFile.name)
+    // `wx` (O_EXCL) so the write fails rather than following a symlink someone
+    // planted at the path, or clobbering an existing file. The 64-bit random
+    // prefix already makes that infeasible to aim at; this makes it impossible.
+    await fs.writeFile(tmpVideoFile, buffer, { flag: 'wx' })
```
llun: **Test coverage — the newly added, security-motivated `{ flag: 'wx' }` has no coverage:** removing it leaves the whole suite green. The comment claims a symlink/clobber guarantee that nothing enforces. Hard to test directly because the path is randomised, so this is a note rather than a blocker.

lib/services/medias/localFile.test.ts
```diff
@@ -445,6 +445,61 @@ describe('LocalFileStorage.saveFile with a video', () => {
...
+
+  it('removes the temp video once probing finishes', async () => {
+    const file = new File([Buffer.from('video-bytes')], 'clip.mp4', {
+      type: 'video/mp4'
+    })
+
+    await createStorage().saveFile(actor, { file })
+
+    const tempPath = vi.mocked(extractVideoImage).mock.calls[0][0]
+    await expect(fs.access(tempPath)).rejects.toThrow()
```
llun, line 509: **Test quality — no test asserts the temp copy exists, with the video's bytes, at the moment `extractVideoImage` is called.** Every assertion is about the path's *shape* or its *absence afterwards*, both of which hold for a file that was never created.

Verified: deleting `await fs.writeFile(tmpVideoFile, buffer, { flag: 'wx' })` from `_saveVideoFile` leaves **all 18 local tests passing** — i.e. an implementation that hands ffmpeg a path that was never written (so every real video upload breaks) ships green.

Suggested fix: have the mock verify its input, e.g. `mockImplementation(async (p) => { extractedFrom = await fs.readFile(p); return ONE_PIXEL_PNG })`.

lib/services/medias/localFile.test.ts
```diff
@@ -445,6 +445,61 @@ describe('LocalFileStorage.saveFile with a video', () => {
...
+    )
+
+    expect(await fs.readdir(mediaRoot)).toEqual([])
+    expect(database.createMedia).not.toHaveBeenCalled()
+    // Nor may the temp copy the extraction read outlive the failure.
+    const tempPath = vi.mocked(extractVideoImage).mock.calls[0][0]
+    await expect(fs.access(tempPath)).rejects.toThrow()
+  })
+
+  it('removes the temp video once probing finishes', async () => {
```
llun: **Test coverage — no local counterpart to the S3 suite's `removes the temp video when probing fails`,** so temp cleanup is only pinned for the *extraction* failure, not the *probe* failure.

Verified: rewriting the fix as `const probe = await extractVideoMeta(buffer)` followed by `extractVideoImage(tmp).finally(unlink)` — which leaks the temp file for the container's lifetime on every ffprobe failure — passes all 18 local tests. The `Promise.all(...).finally(...)` shape is exactly what makes the probe path safe, and nothing here guards it.

lib/services/medias/localFile.test.ts
```diff
@@ -445,6 +445,61 @@ describe('LocalFileStorage.saveFile with a video', () => {
     )
   })
 
+  // Regression: the video was written into the media root and the preview frame
+  // extracted from it only afterwards, so a clip ffmpeg cannot decode left the
+  // bytes on disk with no `medias` row — and a row is the only handle anything
+  // but `scripts/maintenance/cleanupMediaStorage.ts` has on a stored path. The
+  // S3 driver never had the gap: it extracts from its temp file before it
+  // uploads, so a failed extraction stores nothing.
+  it('stores nothing when the preview frame cannot be extracted', async () => {
```
llun, line 480: **Test coverage — the `Invalid video format` branch is exercised by no test in the repo,** so only the extraction-failure half of this PR's "nothing reaches the media root unless everything succeeded" invariant is covered.

Verified: moving `await fs.writeFile(filePath, buffer)` above the container-validation block — which re-creates the exact orphan-with-no-`medias`-row bug this PR fixes, on every invalid-container rejection — passes **all 44 tests across both files**. `extractVideoMeta` is only ever mocked to a valid mp4 probe or to a rejection; a probe returning e.g. `{ streams: [], format: { format_name: 'matroska' } }`, asserting `rejects.toThrow(MediaValidationError)` plus an empty media root, is the missing case.

lib/services/medias/localFile.test.ts
```diff
@@ -445,6 +445,61 @@ describe('LocalFileStorage.saveFile with a video', () => {
...
+    // The extraction's own error, not a `MediaValidationError`: a storage-side
+    // fault stays a logged 500 rather than the 422 a client will not retry.
+    await expect(createStorage().saveFile(actor, { file })).rejects.toThrow(
+      'ffmpeg failed'
+    )
+
+    expect(await fs.readdir(mediaRoot)).toEqual([])
+    expect(database.createMedia).not.toHaveBeenCalled()
+    // Nor may the temp copy the extraction read outlive the failure.
+    const tempPath = vi.mocked(extractVideoImage).mock.calls[0][0]
```
llun, line 496: **Minor (diagnosability) — `mock.calls[0][0]` is read without the `expect(extractVideoImage).toHaveBeenCalledTimes(1)` guard** used further down (same at line 480). If a regression stopped calling `extractVideoImage`, the test fails with `TypeError: Cannot read properties of undefined` instead of a readable assertion. Not a false pass, just noise when it breaks.

llun (commented):
Sub-agent review round 1 (three reviewers: correctness / project-invariant compliance / test quality). Findings below. Reviewer 2 (AGENTS.md + REVIEW.md compliance) found no violations.

lib/services/medias/localFile.ts
llun: Fixed in 7c483c1. The temp copy is now named `video<ext>` where `ext` comes from `getStoredMediaExtension(file.type, file.name)` — the supplied name contributes nothing to it, so neither the demuxer nor the directory can be steered by a client. Applied on the S3 driver too, since the defect was shared; the shared step now lives in `lib/services/medias/videoPreview.ts` and both drivers import it.

Regression coverage: `builds the temp video name from the content type, not the supplied name` (localFile.test.ts) uses `IMG_%04d.jpg` with `Content-Type: video/mp4` and asserts the temp basename matches `^[0-9a-f]{16}-video\.mp4$`; the S3 suite runs the same assertion table-driven over the traversing name, the parent reference, and the image-sequence pattern.

lib/services/medias/localFile.ts
llun: Fixed in 7c483c1. The `Promise.all` is gone: both drivers now probe, validate, and only then extract. An audio-only mp4 is rejected from the probe alone and never spawns ffmpeg, so it is a 422 again, and a genuine extraction fault still propagates unchanged as a logged 500.

Regression coverage: `rejects a container with no video stream without extracting a frame` in both suites — asserts `MediaValidationError`, `extractVideoImage` not called, nothing stored. Verified it bites: re-introducing the concurrent shape fails exactly that test.

Losing the concurrency is a fair trade — validating first means an unusable upload does no ffmpeg work at all, where the concurrent version always spawned it.

lib/services/medias/localFile.ts
llun: Covered in 7c483c1 by `videoPreview.test.ts` -> `refuses to write over an existing temp path`. The randomised path is worked around by forcing one write at a known path with `mockImplementationOnce`, then asserting the call rejects with `EEXIST`, `extractVideoImage` is never called, and the pre-existing file still holds its original bytes. That pins the flag itself rather than the prefix.

lib/services/medias/localFile.test.ts
llun: Fixed in 7c483c1, and made suite-wide rather than a single test. The `extractVideoImage` mock in both driver suites (and in `videoPreview.test.ts`) now reads the file it is handed:

```ts
vi.mocked(extractVideoImage).mockImplementation(async (filePath) => {
  extractedFrom = await fs.readFile(filePath)
  return ONE_PIXEL_PNG
})
```

so an implementation that never writes the temp copy fails everywhere instead of passing green — verified: deleting the `fs.writeFile` now fails 20 tests across the three files. `hands the extraction a temp copy of the uploaded video` asserts the captured bytes explicitly so the intent is visible.

lib/services/medias/localFile.test.ts
llun: Fixed in 7c483c1, though the shape changed with the ordering fix. Now that validation precedes extraction, a probe failure means no temp file is ever written, so the guarantee to pin is stronger than cleanup: `stores nothing and writes no temp video when probing fails` (both suites) asserts `extractVideoImage` is never called and nothing is stored. Cleanup on the paths that do write a temp copy is pinned by `removes the temp copy when the extraction fails` in `videoPreview.test.ts` plus the per-driver assertions.

lib/services/medias/localFile.test.ts
llun: Fixed in 7c483c1. `rejects a container with no video stream without extracting a frame` now exercises the `Invalid video format` branch in both suites, asserting `MediaValidationError`, an empty media root / no `PutObjectCommand`, and no `createMedia`. This is the same test that pins the ordering fix, so the mutation you describe — hoisting the store above the validation — is caught by it.

lib/services/medias/localFile.test.ts
llun: Fixed in 7c483c1 — `expect(extractVideoImage).toHaveBeenCalledTimes(1)` now precedes every `mock.calls[0][0]` read in both suites.

lib/services/medias/videoPreview.ts
```diff
@@ -0,0 +1,37 @@
...
+  extension: string
+): Promise<Buffer> => {
+  const tempFilePath = createMediaTempFilePath(`video${extension}`)
+  // `wx` (O_EXCL) so the write fails rather than following a symlink someone
+  // planted at the path, or clobbering an existing file. The 64-bit random
+  // prefix already makes that infeasible to aim at; this makes it impossible.
+  await fs.writeFile(tempFilePath, buffer, { flag: 'wx' })
+  // `finally` so a failed extraction still removes the temp copy instead of
+  // leaking it for the lifetime of the container.
+  return extractVideoImage(tempFilePath).finally(() =>
```
llun: **Correctness — a failed `fs.writeFile` leaves the partially-written temp copy behind forever.**

The unlink is attached only to the `extractVideoImage(...)` promise, so it only runs once the write has already succeeded. `fs.promises.writeFile` opens the file (`O_CREAT|O_EXCL` via `wx`), writes, and closes in a `finally` — it never unlinks on failure. So a write that fails *after* the file is created orphans it.

**Failure scenario:** an actor uploads a 200 MB mp4 while the temp filesystem is short on space. `open(tempFilePath, 'wx')` succeeds, the write hits `ENOSPC` (or `EDQUOT`/`EIO`) partway, `writeFile` rejects, `extractVideoImage` is never called, and nothing removes `<tmpdir>/<16-hex>-video.mp4`. The upload 500s and the partial file stays for the lifetime of the container; each retry adds another under a fresh random name, so a full-disk condition compounds rather than self-heals.

This shape existed in the old `_uploadVideoToS3` too, but the refactor extends it to `LocalFileStorage`, which previously never wrote a temp video at all. Wrapping from the write onward in `try/finally` fixes it and reads more directly than the `.finally()` chain.

lib/services/medias/videoPreview.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(path.resolve(path.dirname(filePath))).toBe(path.resolve(os.tmpdir()))
+    expect(path.basename(filePath)).toMatch(
+      new RegExp(`^[0-9a-f]{16}-video\\${extension}$`)
+    )
+  })
+
+  // `wx` (O_EXCL): the write must fail rather than follow a symlink planted at
+  // the path or clobber a file already there.
+  it('refuses to write over an existing temp path', async () => {
+    const collision = path.join(os.tmpdir(), 'existing-video.mp4')
```
llun: **Test flake — the `wx` test uses a fixed, machine-global path (`<os.tmpdir()>/existing-video.mp4`), so two concurrent vitest runs on the same machine race on it.**

Reproduced: launching 10 simultaneous runs of just this test, **1 of 10 failed** with `expected Error: ENOENT … to match object { code: 'EEXIST' }`. Mechanism: run A's `finally` unlinks the shared file between run B's setup write and B's assertion, so B's redirected `wx` write succeeds, `extractVideoImage` is called with a path nothing ever wrote, and the mock's `readFile` rejects with ENOENT.

Not a CI risk (shards are separate runners), but very likely here: a `--watch` session alongside a manual run, or tests running in two `.claude/worktrees/` checkouts at once. Fix is one line — put the collision file inside an `fs.mkdtemp()` directory, as `localFile.test.ts` already does for its media root.

Secondarily: the setup write happens *before* the `try`, so if `vi.spyOn` ever throws, the file leaks into tmpdir.

lib/services/medias/videoPreview.ts
```diff
@@ -0,0 +1,37 @@
...
+ * `scripts/maintenance/cleanupMediaStorage.ts`. Keeping the step here is what
+ * stops the two drivers drifting apart on it again.
+ *
+ * `extension` comes from the already-validated content type, and the temp name
+ * carries no part of the supplied file name. ffmpeg chooses its demuxer from
+ * the path as well as from the bytes, and `image2`/`mjpeg` beat content probing
+ * for a name that pairs an image extension with a `%0Nd` or `*` pattern — so an
+ * upload named `IMG_%04d.jpg` sent ffmpeg looking for a numbered image sequence
+ * and failed to open a perfectly good mp4.
+ */
```
llun, line 22: **Docs — the new cross-driver invariant is recorded only in this JSDoc.** Both round-2 reviewers raised this independently.

`AGENTS.md` -> **Uploaded file names are untrusted input** and the matching `REVIEW.md` checklist section are not *wrong* — "`createMediaTempFilePath` is the only sanctioned way to build a temp path from a supplied name" still holds — but they are now the only doc-level statement about media temp paths, and they describe a defense no production call site exercises any more: after this PR `createMediaTempFilePath` is called from exactly one place, with the server-derived literal `video${extension}`.

The genuinely new, non-obvious rules — *both drivers extract the preview frame through `extractVideoPreviewFrame` before storing*, and *the temp name carries no part of the supplied name because ffmpeg's `image2` demuxer wins on `IMG_%04d.jpg`* — exist nowhere outside this file. Per AGENTS.md -> Documentation Maintenance, and given this module is what stops the two drivers drifting apart again, that belongs in `AGENTS.md` and `REVIEW.md` in this PR.

llun (commented):
Sub-agent review round 2 (correctness + tests/conventions) over the reworked diff. Round-1 findings all verified fixed; the correctness reviewer confirmed the two drivers are now line-for-line equivalent, that `extension` can never be attacker-influenced on any call path into `saveFile`, and that the `wx` test genuinely kills its mutant. Three new items.

lib/services/medias/videoPreview.ts
llun: Fixed in d0a6054, though not with the `try` shape suggested — that one has a bug of its own. A `try` wrapping `fs.writeFile(..., { flag: 'wx' })` also covers the EEXIST case, and EEXIST is precisely the case where the path is *not* ours: the `finally` would delete the file the flag just refused to clobber.

So the call opens the file explicitly and ownership becomes structural — the open is what creates the path, so everything after it owns the path and nothing before it does:

```ts
const tempFile = await fs.open(tempFilePath, 'wx')
try {
  await tempFile.writeFile(buffer)
  await tempFile.close()
  return await extractVideoImage(tempFilePath)
} finally {
  await tempFile.close().catch(() => undefined)
  await fs.unlink(tempFilePath).catch(() => undefined)
}
```

(The second `close()` resolves as a no-op on Node 24 — checked — and is caught regardless.)

Both halves are pinned: `removes the temp copy when the write fails partway` forces the handle's `writeFile` to reject with ENOSPC and asserts the partial copy is gone, and the existing collision test already asserted the colliding file survives. Verified both mutants die — the naive `try`-over-the-open version and the original extraction-only `.finally()` version each fail two tests.

lib/services/medias/videoPreview.test.ts
llun: Fixed in d0a6054. The collision file now lives in its own `fs.mkdtemp()` directory, matching what `localFile.test.ts` does for its media root, so concurrent runs cannot see each other's path. The setup write also moved inside the `try`, and cleanup is `fs.rm(scratchDir, { recursive: true, force: true })` — so nothing leaks if `vi.spyOn` throws.

Thanks for actually racing it 10x; a fixed tmpdir path is exactly the kind of thing that only shows up on someone else's machine.

lib/services/medias/videoPreview.ts
llun: Fixed in d0a6054. Added to `AGENTS.md` -> **Uploaded file names are untrusted input** as two bullets, and a matching entry in the `REVIEW.md` -> **Uploaded file names** checklist. Between them they now state the three rules that only lived in the JSDoc:

- the preview frame is extracted through the shared `extractVideoPreviewFrame` and runs *before* the video is stored, never from the stored file, because a stored file with no `medias` row is unreachable by everything except the cleanup script;
- the temp copy takes no part of the supplied name, with the `image2`/`IMG_%04d.jpg` reason spelled out, so `createMediaTempFilePath`'s sanitizer is a second line of defence here rather than the only one;
- validate the container from the probe first, so an audio-only mp4 stays a 422 and never spawns ffmpeg.

Deliberately put in the **Uploaded file names** section rather than **Stored media**, since #1341 is editing the latter and there is no reason to make you resolve a conflict over it.

llun (commented):
Sub-agent review round 3. The correctness reviewer returned clean — it verified the Node 24 double-`close` semantics, that `return await` is load-bearing (a bare `return` would unlink the input while ffmpeg was still reading), that the ENOSPC test is faithful to the real failure, and that the two drivers are now sequence-identical. The docs/tests reviewer found three, including a self-inflicted one: the collision test does not guard the mutant its commit message claims it does.

lib/services/medias/videoPreview.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      path.join(os.tmpdir(), 'video-preview-')
+    )
+    const collision = path.join(scratchDir, 'existing-video.mp4')
+    const spy = vi.spyOn(fs, 'open')
+
+    try {
+      await fs.writeFile(collision, 'someone-elses-bytes')
+      // The random prefix makes a real collision infeasible to aim at, so the
+      // path has to be forced to prove the flag is the thing preventing it.
+      spy.mockImplementationOnce((_target, flags) => fs.open(collision, flags))
```
llun: **Test — the collision test cannot fail for the mutant it was written to catch.**

The mock diverts `fs.open` to a *different* path (`<scratchDir>/existing-video.mp4`) than the `tempFilePath` production computed and will `unlink`, so `expect(fs.readFile(collision))` is vacuous. The "cleanup must not start before the open" invariant — the one d0a6054's message claims this test guards — is therefore unprotected.

Verified: replacing `videoPreview.ts` with exactly that mutant (`fs.open` moved inside the `try`, `finally` doing `tempFile?.close()` + `fs.unlink(tempFilePath)`) leaves **all 7 tests passing**, because production unlinks `tempFilePath`, which is never the collision path.

Colliding on the *real* path does discriminate it (fails on the mutant, passes on HEAD — both run):

```ts
spy.mockImplementationOnce(async (target, flags) => {
  await fs.writeFile(String(target), 'someone-elses-bytes')
  return fs.open(target, flags) // EEXIST
})
```

Note the fixed version has to unlink that file in its own `finally` — on the correct implementation nothing removes it, so it would otherwise leak into the real `os.tmpdir()`.

REVIEW.md
```diff
@@ -75,6 +75,16 @@ change doesn't touch.
   `getStoredMediaExtension()`, not from the name. Every `ACCEPTED_FILE_TYPES`
   entry needs a mapping in `EXTENSION_BY_CONTENT_TYPE`, or it falls through to
   the name.
+- A video's preview frame is extracted through the shared
+  `extractVideoPreviewFrame` (`medias/videoPreview`) **before** the video is
+  stored, never from the stored file: `extractVideoImage` rejects when ffmpeg
+  finds no decodable frame, and a stored file with no `medias` row is
+  unreachable by everything except `scripts/maintenance/cleanupMediaStorage.ts`.
+  Its temp copy is named `video<ext>` and takes nothing from the supplied name —
```
llun: **Docs — "Its temp copy is named `video<ext>`" is not what the code produces.**

`videoPreview.ts` passes `` `video${extension}` `` to `createMediaTempFilePath`, which prepends a 16-hex random prefix and a separator, so the real basename is `<16 hex>-video<ext>` — as the test's own `/^[0-9a-f]{16}-video\.mp4$/` asserts.

As written the bullet describes a fixed `<tmpdir>/video.mp4`, which is exactly the predictable-path pattern the bullet ten lines above it calls a bug. A reviewer using the checklist could wave through a temp path with no random prefix, or flag the correct name as non-conforming.

(`AGENTS.md` uses the same phrasing but reconciles it by naming `createMediaTempFilePath` in the next clause, so it is defensible there — REVIEW.md never mentions the helper.)

AGENTS.md
```diff
@@ -694,6 +694,8 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The temp copy that ffmpeg decodes carries no part of the supplied name.** ffmpeg picks its demuxer from the path as well as from the bytes, and the `image2`/`mjpeg` demuxers beat content probing for an image extension paired with a `%0Nd` number pattern or a `*` glob — so a perfectly good H.264 mp4 uploaded as `IMG_%04d.jpg` sent ffmpeg hunting for a numbered image sequence and answered 500 for a file the instance can store. The name is `video<ext>` from `getStoredMediaExtension`, which makes `createMediaTempFilePath`'s sanitizer a second line of defence here rather than the only one. Validate the container **first**, too: `MediaValidationError` is the caller's 422, and deciding it from the probe alone is what keeps an audio-only mp4 — a voice memo the browser labels `video/mp4` — from spawning ffmpeg and coming back as a logged 500 the client will retry.
```
llun: **Docs — `CLAUDE.md`'s key-reminders bullet (line 43) now guides an agent straight back into the bug this PR fixes.**

It still lists only the three helpers and frames `createMediaTempFilePath` as *the* safe way to build a temp path from `file.name`. On `main` that is literally the line this PR deletes (`_uploadVideoToS3`: `createMediaTempFilePath(file.name)`); after this PR the helper has exactly one production call site and it is passed a server-derived name.

AGENTS.md -> Documentation Maintenance mandates the update on both counts: "New or changed coding conventions and patterns -> the matching `AGENTS.md` section, the `REVIEW.md` checklist, and ... the `CLAUDE.md` key reminders", and "Changes to AGENTS.md rules themselves -> keep the thin per-tool pointer files in sync (`CLAUDE.md`, ...)". An agent that reads only CLAUDE.md at task start would put a client-supplied name back into a path ffmpeg opens.

Checked and **not** needing changes: `GEMINI.md` and `.cursor/rules/agents.mdc` are pure pointers with no media rule content; `.github/copilot-instructions.md` mirrors rule content but every statement in it stays true.

lib/services/medias/videoPreview.test.ts
llun: Fixed in the latest commit — and thank you for actually running the mutant. My round-2 reply claimed this test guarded the ownership invariant; it did not, and the claim was the more dangerous half of the mistake.

It now plants the file at the path production chose, as you suggested:

```ts
spy.mockImplementationOnce(async (target, flags) => {
  collision = String(target)
  await fs.writeFile(collision, "someone-elses-bytes")
  return fs.open(target, flags) // EEXIST
})
```

and unlinks `collision` in its own `finally`, since correct behaviour deliberately leaves that file alone — your point about it otherwise leaking into the real tmpdir.

Verified both directions: HEAD passes 7/7, and the mutant (open moved inside the `try`, `finally` doing `tempFile?.close()` + `fs.unlink(tempFilePath)`) now fails exactly this test.

REVIEW.md
llun: Fixed in the latest commit. Both entries now say the path still comes from `createMediaTempFilePath` — keeping the random prefix and the `tmpdir()` assertion — and that only the *name handed to it* is server-derived:

> Its temp copy goes through `createMediaTempFilePath` like any other, but the name it hands over is the server-derived `video<ext>` (so the path is `<random prefix>-video<ext>`) and takes nothing from the supplied name

You are right that as written it read as a fixed `<tmpdir>/video.mp4`, which is the exact shape the bullet above it calls a bug. AGENTS.md got the same treatment even though it was defensible there, so the two do not have to be read together to be correct.

AGENTS.md
llun: Fixed in the latest commit. `CLAUDE.md`'s key-reminders bullet now says a temp path should prefer a server-derived name over a sanitized supplied one, names `extractVideoPreviewFrame` as the one production caller, gives the `IMG_%04d.jpg` reason, and carries the extract-before-store / validate-the-container-first invariant.

Agreed on your reading of the other pointer files, and I took it: `GEMINI.md` and `.cursor/rules/agents.mdc` carry no media rule content, and every statement in `.github/copilot-instructions.md`'s "Uploaded File Names" section is still true after this PR — so none of the three are touched.

llun:
Round 4 (whole-diff, fresh eyes, explicitly told a clean report was the expected outcome) returned no findings, so the sub-agent review loop has converged — 4 rounds of 20.

Across rounds 1-3 it produced 13 findings, all fixed and resolved above. The two that mattered most were regressions this PR would otherwise have shipped, both introduced by mirroring the S3 driver faithfully and therefore shared with it: ffmpeg selecting its demuxer from the client-supplied temp file name, and container validation moving after extraction so an audio-only mp4 became a logged 500 instead of a 422. One more was self-inflicted: my round-2 collision test did not guard the mutant its commit message claimed it did, which round 3 caught by running the mutant.

Every fix is mutation-checked — reverting it fails a named test.

PR #1344, opened 2026-07-28
fix: treat a blank storage env value as unset

llun:
## Summary

A storage environment variable that is **set but empty** was accepted as a valid value and booted a live-but-broken backend. A *missing* variable fails loudly — `z.string()` rejects `undefined`, so `Config.parse` throws — but a *blank* one satisfies the same schema.

Worst case, in `lib/config/mediaStorage.ts`:

- `ACTIVITIES_MEDIA_STORAGE_TYPE=fs` + `ACTIVITIES_MEDIA_STORAGE_PATH=` yielded `{type:'fs', path:''}`.
- The local driver does `path.resolve(this._config.path)`, and `path.resolve('')` is `process.cwd()`.
- So uploads landed in the application root — and because `app/api/v1/files/[...pathname]/route.ts` passes the requested path straight to the local driver with no DB lookup, an unauthenticated `GET /api/v1/files/package.json` (or `README.md`, or any CWD-relative file with a known mime type) read and served it. The traversal guard only prevents escaping *above* the root; it does not help when the root is the app itself.

Same class of problem for S3: a blank `..._BUCKET` or `..._REGION` produced `{type:'s3', bucket:'', region:''}`, which parses fine, boots, renders as `S3 — ` in the admin Posts & media page, and only fails inside the AWS SDK at request time.

`lib/config/fitnessStorage.ts` had it in **both** branches — its own `ACTIVITIES_FITNESS_STORAGE_PATH/BUCKET/REGION` reads, and its fallback onto the `ACTIVITIES_MEDIA_STORAGE_*` variables, where a blank path fell through `|| 'uploads'` and `path.join('uploads','fitness')` resolved under the CWD.

A whitespace-only value was worse still: `'   '` is truthy, so it survived every existing check and became a literal three-space directory name.

## What changed

A shared `requiredStorageValue(key, subject)` helper trims each required storage value and, when the result is empty, logs a warning and makes the resolver return `null` — the already-supported "no storage configured" state (the composer hides upload, uploads 422, `/api/v1/files` serves the media-removed placeholder, and the admin page renders "Not configured").

Applied to all nine required reads: `ACTIVITIES_MEDIA_STORAGE_PATH/BUCKET/REGION`, `ACTIVITIES_FITNESS_STORAGE_PATH/BUCKET/REGION`, and the fitness fallback onto the media variables.

Two deliberate choices worth a reviewer's attention:

- **The helper takes the variable *name*, not the value.** Passing both would let a call site pair one name with another variable's read — a copy-paste bug the warning message would then misreport. The name is still a literal at the call site, so it stays greppable and `envDocumentation.test.ts` still sees it.
- **It lives in its own module, not `lib/config/utils.ts`.** `utils.ts` is reachable from `proxy.ts` (via `config/host` and, through `utils/http-headers/csp`, `config/securityHeaders`), so it ends up in the middleware Edge Runtime bundle — adding the pino logger there is exactly the class of import `AGENTS.md` warns about.

**Unset is unchanged**: the value stays `undefined` and `Config.parse` still fails loudly. Only blank behaves differently. `hostname`/`endpoint` are untouched — they are optional and already use `|| undefined`.

Uses `logger` from `@/lib/utils/logger` (not `console.*`), following the `lib/config/email.ts` warning style: `` `${key} is set but empty; ${subject} will be disabled` ``.

## Testing

`yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**738 files, 7647 tests, all passing**) — re-run green after every review round.

New table-driven tests cover blank and whitespace-only values for every required variable in both resolvers, plus padded-value trimming and the unset-stays-loud case.

**The tests were verified by mutation**, not just by passing:

- Gutting the helper so blank passes through fails **30** tests across the three files.
- Reverting each of the **nine** guarded call sites individually to its raw `process.env.X` read fails tests every time — so no site is covered only incidentally by another:

  | Mutated call site | Tests failed |
  | --- | --- |
  | media fs path | 3 |
  | media s3 bucket | 4 |
  | media s3 region | 3 |
  | fitness fallback media path | 3 |
  | fitness fallback media bucket | 1 |
  | fitness fallback media region | 1 |
  | fitness own path | 3 |
  | fitness own bucket | 3 |
  | fitness own region | 3 |

Every guard added later in review was mutation-checked the same way: relaxing `=== null` to `== null` fails 9 tests, dropping the fitness fallback guard fails the service test, and short-circuiting either resolver's paired reads fails its both-blank test. The `scripts/` change is outside the test suite, so it was verified by running the script against a throwaway SQLite both ways — a typo'd fitness type refuses, an unset one proceeds past the guard.

Config-only change, so there is no UI to screenshot; the admin "Not configured" and upload-disabled states it falls back to are the existing no-storage paths.

## Notes

- No migration, so no schema-dump regeneration.
- `docs/environment-variables.md` and `docs/fitness-file-storage.md` updated in the same PR.
- Found while reviewing the now-closed #1321 (reply by email); it is independent of that feature and re-derived here rather than cherry-picked.

---

## Review loop (3 rounds, converged)

Per AGENTS.md → Code Review Loop. Every finding was posted inline, addressed, replied to and resolved. Round 3 surfaced nothing actionable, so the loop terminated.

**Round 1** (4 reviewers: correctness, project invariants, tests, security) found that the fix was incomplete on the fitness side. `getEffectiveFitnessStorageConfig` falls back to media storage whenever `config.fitnessStorage` is absent, and could not distinguish "never configured" from "configured, then disabled here" — so `ACTIVITIES_FITNESS_STORAGE_TYPE=s3` with a blank bucket warned that fitness storage was disabled and then wrote every fitness file into the **media** bucket. That was worse than before this branch, where the empty bucket failed loudly in the AWS SDK. `hasExplicitFitnessStorageType()` now makes the runtime fallback match its own documented contract. Round 1 also found that nothing pinned the `=== null` distinction — relaxing it to `== null` passed all 181 `lib/config` tests while silently converting the loud unset failure into a quiet disable.

**Round 2** (2 reviewers) found that round 1's fix gave `scripts/maintenance/cleanupMediaStorage.ts` a path to **delete live data**: it excludes the shared `fitness/` namespace from its orphan scan, and read "no fitness config" as "nothing to exclude". The script now refuses in that state. Round 2 also established that round 1's `it.each` title change was backwards — Vitest truncates an interpolated `$description` at 40 characters but prints a literal title in full — so the descriptions were rewritten to lead with their distinguishing tokens.

## Behaviour changes worth knowing at deploy time

- **An instance currently running with `type=fs` and a blank path is "working"** — media is written to and served from the application directory — and will flip to disabled: uploads rejected, existing media URLs serving the removed-media placeholder, on the strength of one startup warning. That is the point of the fix, but it is a visible change for such an install rather than a no-op. Blank bucket/region carry no such concern: those deployments never worked.
- **Setting `ACTIVITIES_FITNESS_STORAGE_TYPE` to an unrecognised value no longer falls back to media storage** — it disables fitness storage and warns. Previously it silently inherited the media backend.
- **An unrecognised (or missing-but-implied) `ACTIVITIES_MEDIA_STORAGE_TYPE` now warns.** It already disabled media storage; it did so silently.

## Not done

- **`.env.example` still needs the same note as `docs/environment-variables.md`.** AGENTS.md requires both when validation changes. This environment denies read access to `.env*` paths, so it could not be edited here — it is the one outstanding item.

## Follow-ups (deliberately out of scope)

- `path.resolve('.')` is also the CWD, so `ACTIVITIES_MEDIA_STORAGE_PATH=.` reproduces the original exposure. The durable fix belongs at the driver (refuse a root equal to the CWD) or in the files route (look the path up in `medias.original.path` before serving), not as a config blocklist.
- The storage **type** variables are not trimmed, so `ACTIVITIES_MEDIA_STORAGE_TYPE=' fs '` is an unrecognised type. Now that it warns, this is at least observable.
- `_HOSTNAME`/`_ENDPOINT`/`_PREFIX` accept whitespace: `prefix: '  '` prefixes every stored object key with two literal spaces.
- The QStash signing keys have the identical unset-vs-blank asymmetry. It fails closed at the pinned library version, but every queued job is then dropped with `MAX_JOB_RETRIES = 0`.
- A blank entry in `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` collapses the CSP list to the `https:` wildcard — a fail-open.
- `scripts/backup/productionArchive.ts` omits the fitness plan entry in the same unresolved state. Non-destructive (an incomplete backup, not a deletion), so it was left alone.


lib/config/fitnessStorage.ts
```diff
@@ -113,24 +130,42 @@ export const getFitnessStorageConfig = (): {
...
     case FitnessStorageType.ObjectStorage: {
+      const bucket = requiredStorageValue(
+        'ACTIVITIES_FITNESS_STORAGE_BUCKET',
+        'fitness storage'
+      )
+      const region = requiredStorageValue(
+        'ACTIVITIES_FITNESS_STORAGE_REGION',
+        'fitness storage'
+      )
+      if (bucket === null || region === null) return null
```
llun, line 178: **Blocking — this `null` does not disable fitness storage; it silently redirects it to the media bucket.**

`lib/services/fitness-files/index.ts:20-40` (`getEffectiveFitnessStorageConfig`) falls back to media storage whenever `config.fitnessStorage` is falsy. It cannot tell "never explicitly configured" (the case its own doc comment describes) from "explicitly configured but disabled here because a required value was blank".

With `ACTIVITIES_FITNESS_STORAGE_TYPE=s3`, a blank `..._BUCKET`, and a valid media S3 config, boot logs `ACTIVITIES_FITNESS_STORAGE_BUCKET is set but empty; fitness storage will be disabled` — and then every fitness upload succeeds into the **media** bucket under `fitness/`, inheriting `mediaStorage.hostname`. The warning is false.

This is a regression this PR introduces: before it, `bucket: ''` reached the S3 SDK and failed loudly on every request. Fix by making the runtime fallback match its documented contract — only fall back when fitness storage was never explicitly configured. The `fs` branch is unaffected (there is no LocalFile fallback).

lib/config/fitnessStorage.ts
```diff
@@ -77,26 +78,42 @@ export const getFitnessStorageConfig = (): {
...
+        )
+        if (mediaPath === null) return null
         return {
           fitnessStorage: {
             type: process.env.ACTIVITIES_MEDIA_STORAGE_TYPE,
-            path: path.join(
-              process.env.ACTIVITIES_MEDIA_STORAGE_PATH || 'uploads',
-              'fitness'
-            ),
+            path: path.join(mediaPath ?? 'uploads', 'fitness'),
```
llun, line 109: `??` vs the previous `||` is a no-op here (at this point `mediaPath` is either a non-empty trimmed string or `undefined`), so this is fine — but note the `'uploads'` branch is unreachable in a booting app: with `TYPE=fs` and no path, `getMediaStorageConfig()` yields `{path: undefined}` and `Config.parse` throws first. Worth keeping for the preserved default, but the test comment calling it live behaviour overstates it.

lib/config/mediaStorage.ts
```diff
@@ -67,13 +75,23 @@ export const getMediaStorageConfig = (): {
...
     case MediaStorageType.ObjectStorage: {
+      const bucket = requiredStorageValue(
+        'ACTIVITIES_MEDIA_STORAGE_BUCKET',
+        'media storage'
+      )
+      const region = requiredStorageValue(
+        'ACTIVITIES_MEDIA_STORAGE_REGION',
+        'media storage'
+      )
+      if (bucket === null || region === null) return null
```
llun, line 90: **Missing coverage on the `=== null` distinction.** Changing this line to `== null` — the single most likely future "simplification" — makes an *unset* bucket or region silently disable storage instead of failing loudly in `Config.parse`, which is precisely the regression this PR promises not to introduce. That mutation passes all 181 tests in `lib/config`.

Only the `fs` path branch has an unset test. Add unset-bucket and unset-region tests here and the three equivalents in `fitnessStorage.test.ts` (own bucket, own region, and the fallback pair).

lib/config/mediaStorage.ts
```diff
@@ -67,13 +75,23 @@ export const getMediaStorageConfig = (): {
             undefined
         }
       }
+    }
     case MediaStorageType.S3Storage:
     case MediaStorageType.ObjectStorage: {
+      const bucket = requiredStorageValue(
```
llun, line 82: Reading both values before the null check is a deliberate choice — an operator with two blank variables sees both warnings rather than fixing them one boot at a time. Nothing pins it: rewriting this to short-circuit (`if (bucket === null) return null` before reading region) still passes 21/21, because every table row blanks exactly one variable. Add a both-blank case asserting two warns.

It also has a consequence for the docs: with bucket *unset* and region *blank*, this returns `null` and the app boots with media disabled, where previously `Config.parse` threw. The new doc paragraph claims unset is unconditionally a hard startup failure.

lib/config/mediaStorage.test.ts
```diff
@@ -147,5 +157,122 @@ describe('MediaStorage config', () => {
...
+          ACTIVITIES_MEDIA_STORAGE_BUCKET: '',
+          ACTIVITIES_MEDIA_STORAGE_REGION: 'auto'
+        },
+        warnedVariable: 'ACTIVITIES_MEDIA_STORAGE_BUCKET'
+      }
+    ])('returns null when $description', ({ env, warnedVariable }) => {
+      Object.assign(process.env, env)
+
+      expect(getMediaStorageConfig()).toBeNull()
+      expect(mockWarn).toHaveBeenCalledWith(
```
llun, line 260: Two issues with the table:

1. **Title pattern.** AGENTS.md → Testing Guidelines specifies `it.each([...])('$description', …)`. Wrapping it as `'returns null when $description'` pushes the distinguishing text into the truncated tail, so rows become indistinguishable in the reporter — the mutation run printed `× returns null when 'the fallback media object storage reg…'` twice for two different cases. Move the whole behaviour statement into `description`.
2. **`toHaveBeenCalledWith` without a count** tolerates extra warnings, so a call site that also warns for a sibling variable it should not have read is invisible. Add `toHaveBeenCalledTimes(1)`.

lib/config/mediaStorage.test.ts
```diff
@@ -147,5 +157,122 @@ describe('MediaStorage config', () => {
...
+    // disabling storage — only blank changed behaviour.
+    it('leaves an unset required value undefined for schema validation', () => {
+      process.env.ACTIVITIES_MEDIA_STORAGE_TYPE = 'fs'
+      delete process.env.ACTIVITIES_MEDIA_STORAGE_PATH
+
+      const config = getMediaStorageConfig()
+
+      expect(config).not.toBeNull()
+      expect((config?.mediaStorage as { path?: string }).path).toBeUndefined()
+      expect(() => MediaStorageFileConfig.parse(config?.mediaStorage)).toThrow()
```
llun: Bare `.toThrow()` passes if *any* field fails validation, not necessarily `path` — and it also passes on `parse(undefined)`, so it only means anything because of the preceding `not.toBeNull()`. Prefer `safeParse` and assert the failing issue path is `path`.

The fitness twin of this test (`leaves an unset fitness path undefined for schema validation`) never runs a schema at all — it asserts `toBeUndefined()` and stops, so nothing in that file proves the "unset still fails loudly" half of the contract. `FitnessStorageFileConfig` is already exported.

lib/config/mediaStorage.test.ts
```diff
@@ -1,15 +1,25 @@
...
+    warn: vi.fn()
+  }
+}))
+
 describe('MediaStorage config', () => {
   const originalEnv = process.env
+  const mockWarn = logger.warn as jest.Mock
 
   beforeEach(() => {
     process.env = { ...originalEnv }
```
llun, line 21: `process.env = { ...originalEnv }` copies the developer's **real** environment, and `matcher('ACTIVITIES_MEDIA_STORAGE_')` keys off any variable with that prefix. AGENTS.md documents `set -a; . ./.env.local; set +a` as the local workflow, so exported `ACTIVITIES_*` is realistic.

Verified: `ACTIVITIES_FITNESS_STORAGE_TYPE=s3 yarn vitest run lib/config/fitnessStorage.test.ts` fails 11 tests, 8 of them added by this PR. Worse, a lone ambient `ACTIVITIES_MEDIA_STORAGE_HOSTNAME` leaves everything green while `returns null when no … env vars are set` starts passing for the wrong reason. Delete the `ACTIVITIES_MEDIA_STORAGE_`/`ACTIVITIES_FITNESS_` keys in `beforeEach`.

lib/config/storageValue.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(
+      requiredStorageValue('ACTIVITIES_MEDIA_STORAGE_PATH', 'media storage')
+    ).toBeUndefined()
+    expect(mockWarn).not.toHaveBeenCalled()
+  })
+
+  it.each([
+    { description: 'an empty string', value: '' },
+    { description: 'spaces', value: '   ' },
+    { description: 'a tab', value: '\t' },
```
llun: The blank table covers ASCII whitespace only. `.trim()` also strips `\u00a0`, `\ufeff` and `\u3000` — the realistic copy-paste-from-a-web-page case, and the thing that distinguishes relying on `trim()` from a hand-rolled `/^[ \t\n]*$/`. Worth one more row.

docs/environment-variables.md
```diff
@@ -203,6 +203,8 @@ Optional. Enables `POST /api/v1/statuses/:id/translate` and the Translate contro
 
 Required for media uploads (images and video in posts). If no media storage is configured, media uploads are disabled.
 
+`ACTIVITIES_MEDIA_STORAGE_PATH`, `ACTIVITIES_MEDIA_STORAGE_BUCKET`, and `ACTIVITIES_MEDIA_STORAGE_REGION` are required for their storage type and have no default. Surrounding whitespace is trimmed, and a value that is **set but empty** (or whitespace only) is treated as unset: the app logs a warning and leaves media storage disabled instead of starting a broken backend. An empty `ACTIVITIES_MEDIA_STORAGE_PATH` would otherwise resolve to the application's working directory, and an empty bucket or region would only fail later inside the AWS SDK. Leaving one of them **unset** while its storage type is configured remains a hard startup failure.
```
llun: Two problems in this paragraph:

1. **It contradicts itself.** It says a blank value "is treated as unset", then two sentences later that unset "remains a hard startup failure". Blank and unset now behave *differently* — that is the whole point of the change. Say "treated as **not configured**".
2. **The unset claim is not unconditional.** Both bucket and region are evaluated before the null check, so bucket *unset* + region *blank* returns `null` and boots with media disabled rather than throwing. Qualify it.

llun (commented):
Round 1 of the sub-agent review loop (AGENTS.md → Code Review Loop). Four reviewers ran over the whole diff: correctness/consumer-tracing, project-invariant conformance, test quality (with independent mutation testing), and security. Findings posted inline below; two more that have no line to anchor to:

**`.env.example` was not updated.** AGENTS.md:586 requires `docs/environment-variables.md` **and** `.env.example` when validation changes. It is the file operators copy, and it already ships blank-valued lines (`# ACTIVITIES_MEDIA_STORAGE_HOSTNAME=`) that model the exact `NAME=` shape this PR changes the meaning of.

**`docs/fitness-file-storage.md:32` is now conditional.** It states the media-storage fallback unconditionally; with a blank media path/bucket/region there is now no fallback at all and fitness storage is disabled too.

Out of scope, recorded as follow-ups rather than fixed here: `path.resolve('.')` is also the CWD so `ACTIVITIES_MEDIA_STORAGE_PATH=.` reproduces the original exposure (belongs at the driver or as a DB lookup in the files route, not as a config blocklist); the storage **TYPE** variables are still whitespace-truthy (`ACTIVITIES_MEDIA_STORAGE_TYPE=' fs '` silently disables storage with no warning); `_HOSTNAME`/`_ENDPOINT`/`_PREFIX` accept whitespace (`prefix: '  '` prefixes every object key with two literal spaces); the QStash signing keys have the identical unset-vs-blank asymmetry; and a blank entry in `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` collapses the CSP list to the `https:` wildcard.

lib/config/fitnessStorage.ts
llun: Fixed in 1d0e018d9 — confirmed real and the more serious of the two halves. `hasExplicitFitnessStorageType()` (exported from `lib/config/fitnessStorage.ts`, so the env read stays inside `lib/config/`) now gates the runtime fallback in `getEffectiveFitnessStorageConfig`, making it match its own doc comment: it falls back only when fitness storage was *never* configured. Setting the type at all opts out.

Since an explicit-but-unusable type now disables fitness storage outright rather than borrowing media, an unrecognised type would have become a silent total disable — so the explicit switch's `default:` now warns, following the `lib/config/email.ts` precedent.

Covered by a new `getEffectiveFitnessStorageConfig` describe in `lib/services/fitness-files/index.test.ts` (fallback still works when unconfigured; no fallback when a type is set). Verified by mutation: removing the guard line fails that test.

lib/config/fitnessStorage.ts
llun: Acknowledged — no code change. Agreed `??` is a no-op here, kept for precision about what the value can actually be at that point. The test comment overstated it and is now corrected to say a booting app cannot reach the `'uploads'` branch (`Config.parse` throws on `{path: undefined}` first) and that the test pins the resolver in isolation.

lib/config/mediaStorage.ts
llun: Fixed in 1d0e018d9 — this was the sharpest finding: the `== null` mutation passing all 181 tests meant the PR's central promise was unpinned. Unset is now table-driven across all five required values (media path/bucket/region, fitness path/bucket/region, and the two fallback media values), each asserting the value is `undefined` AND that `safeParse` fails with that specific field in the issue path. Re-ran the mutation: `=== null` → `== null` now fails 9 tests.

lib/config/mediaStorage.ts
llun: Both fixed in 1d0e018d9. Added `warns for every blank required value before disabling storage` to both resolvers, asserting `toHaveBeenCalledTimes(2)` and both messages — the short-circuit rewrite you describe now fails it.

The docs claim was wrong as written and is now qualified: unset is a hard startup failure *unless another required variable for the same storage type is blank*, which disables storage before validation runs.

lib/config/mediaStorage.test.ts
llun: Both fixed in 1d0e018d9. Titles are now `'$description'` with the whole behaviour statement in the column, per AGENTS.md — the two truncated rows are distinguishable again. Every table row also asserts `toHaveBeenCalledTimes(1)`, and the new both-blank test asserts exactly 2.

lib/config/mediaStorage.test.ts
llun: Fixed in 1d0e018d9. Both files now use `safeParse` and assert the failing issue path contains the field under test, so the assertion cannot be satisfied by an unrelated field failing. The fitness twin is no longer assertion-free either — it is now a five-row table that imports `FitnessStorageFileConfig`/`FitnessStorageS3Config` and runs them.

lib/config/mediaStorage.test.ts
llun: Fixed in 1d0e018d9 — verified the failure mode you describe before fixing. `beforeEach` now deletes every `ACTIVITIES_MEDIA_STORAGE_*` key from the copied env (and `ACTIVITIES_FITNESS_*` too in the fitness file, since both prefixes select the branch there). The pass-for-the-wrong-reason case matters as much as the outright failures, so the comment explains why rather than just doing it.

lib/config/storageValue.test.ts
llun: Fixed in 1d0e018d9 — added U+00A0 and U+3000 rows. Agreed on the reasoning: they are what distinguishes relying on `trim()` from a hand-rolled ASCII whitespace class, and they are the realistic paste case.

docs/environment-variables.md
llun: Both fixed in 1d0e018d9. Blank is now described as treated as **not configured**, and the unset sentence is qualified per your other comment. Also added the point that the storage **type** variables are *not* trimmed, so `ACTIVITIES_MEDIA_STORAGE_TYPE=' fs '` is an unrecognised type — worth stating explicitly in the same paragraph rather than leaving a reader to generalise the trimming rule.

scripts/maintenance/cleanupMediaStorage.ts
```diff
@@ -151,7 +154,14 @@ const getContainedRelativePath = (from: string, to: string) => {
  */
 const getSharedFitnessPrefix = (mediaStorage: MediaStorageConfig) => {
   const fitnessStorage = getEffectiveFitnessStorageConfig()
-  if (!fitnessStorage) return null
+  if (!fitnessStorage) {
+    // "Never configured" and "configured but unusable" both resolve to null,
+    // and only the first means there is no fitness namespace to protect. A
+    // deployment that ran on the media fallback and later acquired a blank or
```
llun, line 160: **Blocking, and caused by round 1's own fix.** `getSharedFitnessPrefix` reads "no fitness config" as "no fitness namespace to protect" — which was true while an unusable fitness config still fell back to the media bucket. After round 1 it resolves to `null` instead.

So: a deployment runs with `ACTIVITIES_FITNESS_STORAGE_TYPE` unset, accumulating .fit/.gpx/.tcx and Strava archives under the media backend's `fitness/` prefix. Someone later sets that variable to a typo'd value. Those files are referenced by `fitness_files.path`, which this script never reads — so with nothing excluded, every one of them is reported as orphaned media and deleted under `--yes`.

The script now refuses in that state, matching how it already handles indistinguishable media/fitness paths. Verified by running it against a throwaway SQLite both ways: a typo'd type refuses with the new message, an unset type proceeds past the guard.

(The *blank*-value cases were already `null` here before this branch — a blank bucket produced a config whose bucket did not match media's, so the equality check rejected it. The newly-widened case is specifically an unrecognised type with S3 media.)

lib/config/mediaStorage.ts
```diff
@@ -94,6 +113,18 @@ export const getMediaStorageConfig = (): {
       }
     }
     default:
+      // Reached only when some ACTIVITIES_MEDIA_STORAGE_* variable is set, so
+      // the operator meant to configure storage. Disabling it silently is the
+      // same failure mode as a blank required value. Mirrors lib/config/email.ts.
+      if (process.env.ACTIVITIES_MEDIA_STORAGE_TYPE) {
+        logger.warn(
```
llun, line 120: Round 2 raised that the docs paragraph implies media and fitness behave alike on an unrecognised type, when only fitness warned. Rather than document the asymmetry, this is the last silent-disable path in these two resolvers — the exact failure mode the branch exists to remove — so it now warns, following `lib/config/email.ts`.

The unset-type case is separated because `default:` is also reached when other `ACTIVITIES_MEDIA_STORAGE_*` variables are set with no type, which is a different operator mistake. An instance with no storage variables at all still returns before the switch and stays silent; there is a test for each of the three.

lib/config/fitnessStorage.test.ts
```diff
@@ -108,5 +132,317 @@ describe('FitnessStorage config', () => {
...
+      {
+        description: 'whitespace fallback region disables storage',
+        env: {
+          ACTIVITIES_MEDIA_STORAGE_TYPE: 'object',
+          ACTIVITIES_MEDIA_STORAGE_BUCKET: 'bucket',
+          ACTIVITIES_MEDIA_STORAGE_REGION: '\t'
+        },
+        warnedVariable: 'ACTIVITIES_MEDIA_STORAGE_REGION'
+      }
+    ])('$description', ({ env, warnedVariable }) => {
```
llun, line 245: **Round 1's title change was backwards — thanks for re-checking it rather than taking it as settled.** Vitest truncates an *interpolated* `$description` at 40 characters (loupe's `objDisplay`) but prints a *literal* title in full, so moving the statement into the column collapsed five fallback rows to `'returns null when the fallback media …'`.

AGENTS.md prescribes the `$description` form, so the fix is in the descriptions: they now lead with the distinguishing tokens (`blank fallback bucket disables storage`). Re-ran the fallback-guard mutation and every failing row is individually named.

Also fixed here: the fallback branch's two reads were not pinned against short-circuiting the way the explicit branch was — verified that rewriting it passed all 30 tests. The both-blank test is now table-driven over both branches, and that mutation now fails.

docs/environment-variables.md
```diff
@@ -222,6 +228,10 @@ S3 credentials are not `ACTIVITIES_*` variables: the AWS SDK resolves them from
 
 For fitness activity file uploads (.fit, .gpx, .tcx). Falls back to media storage configuration if not set.
 
+`ACTIVITIES_FITNESS_STORAGE_PATH`, `ACTIVITIES_FITNESS_STORAGE_BUCKET`, and `ACTIVITIES_FITNESS_STORAGE_REGION` follow the same rule as their media counterparts: whitespace is trimmed, and a **set but empty** value disables fitness storage with a warning rather than starting a broken backend. The media-storage fallback applies the rule to the `ACTIVITIES_MEDIA_STORAGE_*` variables it reads, so a blank `ACTIVITIES_MEDIA_STORAGE_PATH` no longer falls through to `uploads/fitness` under the application's working directory.
```
llun, line 231: Correct on both counts, fixed.

"At all" was false for an empty value: `if (!fitnessStorageType)` and `Boolean(...)` are both truthiness checks, so `ACTIVITIES_FITNESS_STORAGE_TYPE=` still selects the fallback while `'   '` opts out. That is the one blank/unset asymmetry this branch does not remove — the type variable was outside the scope of the reported bug — so it is now stated explicitly rather than papered over, and trimming the type variables is recorded as a follow-up in the PR body.

The "applies to blank values only" sentence was also muddled — trimming applies to any padded value, not just blank ones — and now says the rule covers the three value variables rather than the type.

llun (commented):
Round 2 of the review loop, run against round 1's own fixes — which is where the serious finding was. Two reviewers: correctness/consumer-tracing and tests+docs, both re-verified independently.

All findings addressed in a2b69b8af; details on each thread. Full gate green: prettier, lint (0 errors), build, **738 files / 7647 tests**.

llun:
Round 3 of the review loop: re-reviewed rounds 1-2 (the `UNRESOLVED` sentinel's placement before any listing or deletion, sentinel collision, the new `default:` warnings' reachability and memoization, cross-function agreement on what "configured"/"disabled"/"not configured" mean, and the three docs paragraphs against the final code). **No actionable findings** — the loop converges at 3 of the 20 permitted rounds.

One observation recorded as a follow-up rather than fixed: `scripts/backup/productionArchive.ts` omits the fitness plan entry in the same unresolved state. Unlike the cleanup script that is non-destructive — an incomplete backup, not a deletion — and it is pre-existing, so it stays out of this PR.

PR #1345, opened 2026-07-28
fix: stop the env-block builder emitting placeholders as values

llun:
## The bug

The admin **Configure environment** builder (Admin → Posts & media) filled a required field's value from its **placeholder** when nothing had been typed:

```ts
value: value || field.placeholder
```

A placeholder is an example for the input beside it. Emitting it as a value meant an admin could open the builder, click **Copy .env block**, paste, restart, and boot a real configuration out of values they do not own. With nothing typed:

| Choice | Block before | Block after |
| --- | --- | --- |
| `storage/s3` | `..._BUCKET=media.example.social`<br>`..._REGION=eu-central-1` | `..._BUCKET=`<br>`..._REGION=` |
| `storage/fs` | `..._PATH=./uploads` | `..._PATH=./uploads` (unchanged — a real value) |
| `maps/mapbox` | `..._MAPBOX_ACCESS_TOKEN=pk.eyJ1Ijoi…` | `..._MAPBOX_ACCESS_TOKEN=` |
| `maps/apple` | fake team id, key id and PEM | all three empty |

## The change

Splits the two ideas `placeholder` was carrying, in `lib/config/environmentTemplates.ts`:

- `placeholder` — an example, shown in the input, **never** emitted into the block.
- `defaultValue` (new, optional) — a genuinely correct value the operator can keep, which **is** emitted.

A required field with nothing typed emits `NAME=<defaultValue>` if it has one and `NAME=` otherwise; optional fields are unchanged (omitted until given a value). Only `ACTIVITIES_MEDIA_STORAGE_PATH` carries a `defaultValue` (`./uploads`), which keeps the filesystem block pasteable as-is. It is a default of the *field*, not of the variable — no storage variable has an app-level default.

### Relationship to #1344

`NAME=` is only inert if the resolvers reject a blank value. This branch originally carried its own implementation of that guard; **#1344 landed it independently and goes further** (the `getEffectiveFitnessStorageConfig` fallback, the `cleanupMediaStorage.ts` refusal). On rebase that duplicate was dropped in favour of main's — the two implementations had independently converged on the same `requiredStorageValue(name, subject)` shape and the same leaf-module placement, so there was nothing left to reconcile.

Two things are kept from this branch's review rounds because main lacks them:

**A guard for the boundary #1344 relies on.** `lib/config/storageValue.ts` exists precisely because `lib/config/utils.ts` is reachable from `proxy.ts` — but nothing enforced that. Putting a logger import there passes lint, build and every test while dragging **264 pino files** into `middleware.js` (`lib/utils/logger` calls `pino({…})` at module top level and there is no `sideEffects` field, so it cannot be tree-shaken). `proxy.test.ts` now fails on `pino` / `@/lib/utils/logger` anywhere in the proxy's static import graph, reusing the existing walk, and both cases assert the walk still reaches `lib/config/utils.ts` so a resolution failure cannot make them vacuous. Verified by reintroducing the exact import.

**What a CWD-rooted media path actually exposed.** `GET /api/v1/files/<path>` reads off the storage root with no database lookup — only a containment check and a MIME-type check — so a legacy root `config.json` (`secretPhase`, database credentials), `backups/production-archives/*.tar.gz`, and `.p8`/`.pem`/`.p12` key material were readable unauthenticated. Affected operators need to **rotate**, not just re-point a variable. There is an upgrade note for instances that were running that way, since they lose media and fitness storage with only a boot warning to say so. `.env.example` also gets the blank-value note #1344 explicitly left undone.

## Tests

The invariant is asserted against **the string the copy button actually writes**, not the preview's `textContent` — the preview renders each line as a `<div>`, so `textContent` has no newlines and splitting it yields one concatenated blob that can only observe the last value of a block.

It sweeps **every area and every choice** (`fs`, `s3`, `osm`, `mapbox`, `apple`), asserting per choice:

1. every required variable is present, with a matching line count — so an emptied block cannot pass vacuously;
2. no field's placeholder appears as a value;
3. an unfilled required field is exactly its `defaultValue ?? ''`;
4. a typed value beats the default, and a whitespace-only one does not.

`environmentTemplates.test.ts` separately pins the permitted defaults **by name and value**, since assertion 2 must skip a field whose default *is* its placeholder — correct for `./uploads`, a silent hole for anything else.

**Every assertion was proved by mutation.**

## Verification

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (742 files, 7757 tests) all green on the rebased tree.
- `middleware.js.nft.json` re-checked after the rebase: zero pino entries.
- All five area/choice blocks verified in a real browser against a local SQLite instance with a mock admin, plus a typed value correctly overriding `./uploads`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -133,6 +133,11 @@ ACTIVITIES_SECRET_PHASE=
 # ============================================================================
 # Required for media uploads (images, etc.)
 # Without this, media uploads are disabled.
+#
+# Leave the whole block commented out to disable uploads. Do NOT set a variable
+# to an empty value instead: a blank required value (e.g. a bare
+# ACTIVITIES_MEDIA_STORAGE_PATH=) is rejected and disables storage with a
+# warning, rather than being ignored.
 
 # Storage type: fs (local), s3, object (S3-compatible)
 # ACTIVITIES_MEDIA_STORAGE_TYPE=fs
@@ -154,6 +159,11 @@ ACTIVITIES_SECRET_PHASE=
 # OPTIONAL - Fitness File Storage Configuration
 # ============================================================================
 # Falls back to media storage config if not set separately.
+#
+# The same blank-value rule applies to _PATH / _BUCKET / _REGION here: leave a
+# variable commented out rather than setting it to an empty value. (An empty
+# ACTIVITIES_FITNESS_STORAGE_TYPE= is different — it selects the media-storage
+# fallback, exactly as leaving it unset does.)
 
 # ACTIVITIES_FITNESS_STORAGE_TYPE=fs
 # ACTIVITIES_FITNESS_STORAGE_PATH=./uploads/fitness
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -75,6 +75,22 @@ inert — nothing typed into it is submitted, stored, or sent anywhere, and the
 server still only reads those values from the environment at boot, so a change
 needs a restart.
 
+A required variable you have not filled in is listed in the block carrying a
+value only when the example is itself a correct one to keep — today that is just
+`ACTIVITIES_MEDIA_STORAGE_PATH=./uploads` — and empty otherwise (`NAME=`), never
+its example. The example belongs to the input beside it: a block that carried
+examples as values could be pasted verbatim and boot a real configuration out of
+them, pointing a live instance's uploads at a bucket the operator does not own.
+Note that `./uploads` is a default of the builder's field, not of the variable:
+none of the storage variables has an app-level default, so omitting a required
+one from your `.env` fails config validation rather than falling back. An
+**optional** variable is left out of the block entirely until you type something
+into it, which is why `ACTIVITIES_MEDIA_STORAGE_ENDPOINT` does not appear by
+default.
+
+An emitted `NAME=` is inert because a blank required value is rejected rather
+than accepted — see [Media Storage](#media-storage) below for that rule.
+
 ## Core Configuration
 
 | Variable                         | Required | Description                                                                                                                                                                                                                                            |
@@ -205,6 +221,15 @@ Required for media uploads (images and video in posts). If no media storage is c
 
 `ACTIVITIES_MEDIA_STORAGE_PATH`, `ACTIVITIES_MEDIA_STORAGE_BUCKET`, and `ACTIVITIES_MEDIA_STORAGE_REGION` are required for their storage type and have no default. Surrounding whitespace is trimmed, and a value that is **set but empty** (or whitespace only) is treated as **not configured**: the app logs a warning and leaves media storage disabled instead of starting a broken backend. An empty `ACTIVITIES_MEDIA_STORAGE_PATH` would otherwise resolve to the application's working directory, and an empty bucket or region would only fail later inside the AWS SDK.
 
+If you have been running with a blank media path, treat it as a **credential exposure**, not merely a misconfiguration. `GET /api/v1/files/<path>` reads straight off the storage root with no database lookup — it only checks that the path stays inside the root and that the extension maps to a known MIME type — so with the root set to the working directory, any file in the application checkout with a recognised extension was readable without authentication. `.env`, `.env.local` and `*.sqlite3` happen not to map to a MIME type and so were not readable, but that is not an all-clear; plenty of secret-bearing files do map, including:
+
+- a legacy root `config.json` (`application/json`), which on older instances held the whole configuration: `secretPhase`, database credentials, and auth and email secrets (S3 credentials were never in it — those come from the AWS SDK chain);
+- `backups/production-archives/*.tar.gz` (`application/gzip`), the default output directory of `scripts/backup/productionArchive.ts` — a full database and media archive;
+- key material such as `.p8` (the Apple Maps signing key), `.pem`, `.crt`, `.p12` and `.pfx`;
+- `.conf`, `.ini`, `.toml` and `.txt`, plus source files (`.ts`, `.js`, `.json`, `.md`, `.yml`, `.sql`).
+
+Audit what was sitting in the application directory and rotate anything reachable from that list, rather than assuming only source code leaked.
+
 Leaving one of them **unset** while its storage type is configured is different, and still a hard startup failure — unless another required variable for the same storage type is blank, which disables storage before validation runs.
 
 The rule covers those three value variables, not `ACTIVITIES_MEDIA_STORAGE_TYPE`, which is neither trimmed nor blank-checked: a padded `' fs '` is simply an unrecognised type. An unrecognised type — and a type left unset while other `ACTIVITIES_MEDIA_STORAGE_*` variables are set — disables media storage with a warning naming the variable.
@@ -223,6 +248,8 @@ The rule covers those three value variables, not `ACTIVITIES_MEDIA_STORAGE_TYPE`
 S3 credentials are not `ACTIVITIES_*` variables: the AWS SDK resolves them from its standard chain, so set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (or leave them unset when the host already supplies an IAM role or instance profile).
 
 > Upgrade note: If you previously set `ACTIVITIES_MEDIA_STORAGE_HOSTNAME` or `ACTIVITIES_FITNESS_STORAGE_HOSTNAME` to a MinIO, Cloudflare R2, DigitalOcean Spaces, or other S3-compatible API endpoint, move that value to the matching `*_STORAGE_ENDPOINT` variable. `*_STORAGE_HOSTNAME` is for a public hostname/CDN origin, not for S3 API operations or browser presigned uploads.
+
+> Upgrade note: An instance that was running with a **blank** required storage variable (most likely `ACTIVITIES_MEDIA_STORAGE_TYPE=fs` with an empty `ACTIVITIES_MEDIA_STORAGE_PATH=`) had a working-looking media backend rooted at the process working directory. That configuration now disables media and fitness storage: uploads stop, and every existing media URL serves the "media removed" placeholder. The signal is a pair of boot warnings — `ACTIVITIES_MEDIA_STORAGE_PATH is set but empty; media storage will be disabled` and the matching `… fitness storage will be disabled` — so grep for `is set but empty` after upgrading. To recover, set a real path and move your existing files into it. They are loose in the directory the app was started from, named `<16 hex characters><extension>`: `.webp` for stored images, `-thumbnail.webp` for their thumbnails, `.jpg` for the JPEG renditions generated for email, and `.mp4`/`.webm` for video attachments — so move all of those, not just the `.webp` files. Fitness activity files were not loose in that directory: the fallback defaulted the empty path to `uploads`, so move `./uploads/fitness/` into `<new path>/fitness/` as well, or every `.fit`/`.gpx`/`.tcx` and imported Strava `.zip` archive goes missing. Route maps and their email JPEGs are **not** in there — they go through media storage, so they are among the loose files and the first move already covers them. Set the path even if you do not want media storage: leaving the `ACTIVITIES_MEDIA_STORAGE_*` variables unset entirely is the supported way to turn uploads off.
 
 ## Fitness File Storage
 
--- a/lib/components/admin/settings/EnvBlockBuilder.test.tsx
+++ b/lib/components/admin/settings/EnvBlockBuilder.test.tsx
@@ -10,6 +10,8 @@ import {
   within
 } from '@testing-library/react'
 
+import { ENV_TEMPLATE_AREAS } from '@/lib/config/environmentTemplates'
+
 import { EnvBlockBuilder } from './EnvBlockBuilder'
 
 const STORAGE_AREA = 'Media storage — filesystem or S3'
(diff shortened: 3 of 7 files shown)


llun:
## Browser verification

Local SQLite instance, mock admin, `/admin/posts` → **Configure environment**. Nothing typed in either case.

**S3 (default choice)** — the inputs still show `media.example.social` / `eu-central-1` as *placeholders*, but the generated block no longer carries them:

```
ACTIVITIES_MEDIA_STORAGE_TYPE=s3
ACTIVITIES_MEDIA_STORAGE_BUCKET=
ACTIVITIES_MEDIA_STORAGE_REGION=
```

**Filesystem** — the one real default is still emitted, so this block stays pasteable as-is:

```
ACTIVITIES_MEDIA_STORAGE_TYPE=fs
ACTIVITIES_MEDIA_STORAGE_PATH=./uploads
```

**Fitness maps**, read off the live page for all three providers:

```
osm    → ACTIVITIES_FITNESS_MAP_PROVIDER=osm
mapbox → ACTIVITIES_FITNESS_MAP_PROVIDER=mapbox
         ACTIVITIES_FITNESS_MAPBOX_ACCESS_TOKEN=
apple  → ACTIVITIES_FITNESS_MAP_PROVIDER=apple
         ACTIVITIES_FITNESS_APPLE_MAPS_TEAM_ID=
         ACTIVITIES_FITNESS_APPLE_MAPS_KEY_ID=
         ACTIVITIES_FITNESS_APPLE_MAPS_PRIVATE_KEY=
```

Before this change those last two emitted a fake `pk.eyJ1Ijoi…` token and a fake PEM respectively.

llun:
## Review round 1 — findings and resolutions

Three sub-agents reviewed the diff (builder/descriptor, storage resolvers, docs/security). No blocking findings; the fix itself was confirmed correct on all five area/choice blocks. Everything actionable is addressed in c36217102.

### Test gaps that were real (each now pinned by mutation)

| Finding | Mutation that used to survive | Now |
| --- | --- | --- |
| Nothing ever typed into a field that **has** a `defaultValue` | `field.defaultValue \|\| value` — silently discards everything typed into "Media directory" | `prefers a typed value over $field.name defaultValue` |
| Whitespace-only input untested | dropping `.trim()` on the typed value | `treats a whitespace-only $field.name as unfilled` |
| The `undefined` arm of the tri-state untested | collapsing to `if (!value) return null` — turns a loud `Config.parse` failure into a silent "storage disabled" | `passes an unset required value through rather than disabling storage` |
| Warning asserted only to contain `"empty"` | swapping the `BUCKET`/`REGION` names — points the operator at the wrong variable | both suites assert the exact message; the fitness suite asserted nothing at all before |

All 8 round-2 mutations are killed.

### Code

- **`requiredStorageValue` moved to `lib/config/utils.ts`** (beside `matcher`, which both resolvers already import) and now reads `process.env` itself. Passing a name *and* a separately-written value nine times compiles happily when the two disagree — the same bug class as the warning finding above. This also drops the `fitnessStorage → mediaStorage` edge added for it.
- **`*_STORAGE_TYPE` gets the same blank-is-unset treatment.** It was the one variable in these resolvers not getting this PR's own invariant: untrimmed, `ACTIVITIES_MEDIA_STORAGE_TYPE='fs '` missed every `switch` case and disabled storage with **no warning at all**, and a whitespace-only `ACTIVITIES_FITNESS_STORAGE_TYPE` was truthy enough to skip the media-storage fallback before doing the same.
- **`|| 'uploads'` dropped from the fitness media fallback.** Post-guard it was reachable only for an *unset* media path — where the media resolver hands the same `undefined` to `Config.parse` and the app refuses to boot — so it had this one resolver claiming a working backend for a configuration that cannot start, which `diagnoseFitnessImport.ts` would then report.

### Two claims that were wrong

- The builder comment said the resolvers *"treat a blank value as unset"*. That is backwards, and the sibling module documents the opposite: **blank disables one backend; missing stops the whole app** (`z.string()` rejects `undefined` in `Config.parse`). A future reader acting on the old wording could "simplify" the helper and take down the entire config.
- The docs **understated** the empty-path consequence. `GET /api/v1/files/<path>` reads straight off the storage root with no `medias` row lookup — only a containment check and a MIME-type check — so a CWD root exposed any file in the checkout with a recognised extension, unauthenticated. Documented as a source-disclosure exposure rather than a misconfiguration, with an **upgrade note**: an instance running with a blank path loses media and fitness storage on upgrade, and existing media URLs degrade to the "media removed" placeholder, with only a boot warning to say so.

### Won't fix

- **`logger.warn('string')` vs `logger.warn({ message })`** — `lib/config/email.ts` and `lib/config/translation.ts` all use the bare-string form and pino's `messageKey: 'message'` makes both serialize identically. Consistent with its siblings.
- **A `logger.warn` in `mapProvider.ts`** to match the storage resolvers' new warning. Real asymmetry — an unfilled maps block now downgrades to OpenStreetMap silently — but it is outside this diff. Noted for follow-up.

Full suite re-verified: 738 files / 7661 tests, build and lint clean, and all five blocks re-checked in the browser against the rebuilt server (plus a typed value correctly overriding `./uploads`).

llun:
## Review rounds 2–3

Both rounds reviewed the *fixes*, not the original diff, and both found real regressions introduced by the previous round's fixes. Addressed in 255b43496 and efee42d31.

### Round 2 — two regressions in the round-1 fixes

**Moving `requiredStorageValue` into `lib/config/utils.ts` pulled pino into the middleware bundle.** `proxy.ts` → `lib/config/host` → `./utils` → `lib/utils/logger`. A fresh build traced **264 pino files** into `middleware.js.nft.json`. `lib/utils/logger.ts` calls `pino({…})` at module top level and `package.json` declares no `sideEffects`, so it cannot be tree-shaken — one import anywhere in that graph drags in pino, its serializers and sonic-boom. AGENTS.md forbids exactly this, and the existing `proxy.test.ts` guard could not see it because it only flags literal `fs`/`path` specifiers.

Fixed by moving both helpers to `lib/config/storageValues.ts`, a leaf module only the two resolvers import. Trace is back to **zero** pino files.

**Trimming `*_STORAGE_TYPE` could take a booting instance down.** This was scope I added beyond the stated bug, and it was wrong. With `ACTIVITIES_MEDIA_STORAGE_TYPE` set to `"s3 "`, a bucket set and a region never set, that typo *currently* leaves storage inert and the app booting. A repaired selector matches the case, hands `region: undefined` to `Config.parse`, and throws on every `getConfig()` — the instance is down.

Reverted to normalising **only emptiness**, which is all the builder's `NAME=` needs. That also resolved a knock-on: `getSecurityHeaderConfig` reads the same variable raw, so a repaired type would have activated S3 storage while the CSP omitted the bucket — uploads store fine, then don't render. Round 3 rebuilt the full selector matrix for both resolvers against `main` and confirmed every cell now agrees.

**The upgrade note would have caused data loss.** It named only `.webp`, so `mv *.webp` would have left behind every video attachment and every JPEG email rendition. Now names all extensions, the exact pair of boot warnings to grep for, and the `uploads/fitness/` directory.

### Round 3 — verified the fixes, then found the exposure note was reassuring where it shouldn't be

Round 3 confirmed all three fixes empirically (built both trees, ran the selector matrix side by side, mutation-tested the new assertions). Its main finding:

**The "not readable" sentence read as an all-clear.** Saying `.env` and `*.sqlite3` don't map to a MIME type is true, but with the storage root at the CWD the unauthenticated files route also served:

- a legacy root `config.json` (`application/json`) — on older instances that held `secretPhase`, database credentials and storage keys;
- `backups/production-archives/*.tar.gz` (`application/gzip`) — the default output of `scripts/backup/productionArchive.ts`, a full database and media archive;
- `.p8` (the Apple Maps signing key), `.pem`, `.crt`, `.p12`, `.pfx`.

An operator following the old wording would have concluded only source leaked and skipped rotating credentials. The note now says to audit and rotate.

Also fixed: the padded-selector example relied on a trailing space inside a code span, which **renders invisibly** — it read as a claim that a correctly-spelled `ACTIVITIES_MEDIA_STORAGE_TYPE=s3` matches no backend — and doesn't reproduce from an unquoted `.env` line anyway, since dotenv strips the padding first.

### The guard that would have caught it

The bundle regression happened silently, inside this PR, with a green build. `proxy.test.ts` now also fails on `pino` / `@/lib/utils/logger` anywhere in the proxy's static import graph, reusing the existing walk. Verified by reintroducing the exact import — it fails with `lib/config/utils.ts imports @/lib/utils/logger`.

---

Gate green after each round; currently 738 files / 7663 tests, build and lint clean. Round 4 is running.

llun:
## Review rounds 4–5

### Round 4 — one wrong claim in the recovery note

**Route maps are not under `uploads/fitness/`.** The note said a blank media path left "every `.fit`/`.gpx` and route map" there. Every map write goes through `saveMedia` / `saveMediaImageRendition` (`processFitnessFileJob.ts:200,422`, `regenerateFitnessMapsJob.ts:254`) — **media** storage — so maps and their email JPEGs are among the loose files in the working directory that the preceding sentence already covers. An operator following the old wording would have grepped `uploads/fitness/`, found no maps, and concluded they were lost.

Also dropped "storage keys" from the `config.json` exposure item: S3 credentials come from the AWS SDK chain and were never in that file, so naming them overstates what needs rotating.

**Proxy guards folded into one `it.each`.** They were byte-identical apart from the `Set`, which is exactly the input-only variation AGENTS.md's table-driven rule covers. The shared walk now returns its visited set, and both cases assert it reached `lib/config/utils.ts` — otherwise a future resolution failure collapses the walk to `proxy.ts` alone and leaves both guards green and inert. Verified by mutation: reintroducing the logger import fails the logger case; disabling descent fails both.

### Round 5 — one finding accepted, one rejected

**Accepted:** calling `./uploads` a "real default" overclaims. It is a default of the **builder's field**, not of `ACTIVITIES_MEDIA_STORAGE_PATH`. No storage variable has an app-level default — omitting a required one fails config validation and the app does not boot, which `mediaStorage.test.ts` pins. The old wording invited a reader to leave the variable out and expect a fallback. Reworded in both the doc and the field comment, and the `.env.example` blank-value note now covers the fitness block too.

**Rejected:** the review claimed the `.zip` in the fitness recovery list was wrong, on the grounds that no `.zip` is ever written to a filesystem fitness backend — citing the presigned S3 flow as the only writer and `FitnessFileSchema` rejecting `.zip` on the multipart path.

That check missed a route. `app/api/v1/fitness/strava/archive/route.ts:352` calls `saveFitnessFile` with the raw archive directly, with **no S3 gating** and without going through `FitnessFileSchema`. On a filesystem backend that reaches `localFile.saveFile`, where `getFitnessFileType` (`types.ts:128-149`) returns `'zip'` for a `.zip` name or a `/zip` MIME type and the file is written as `.${fileType}`. So a `.zip` archive can and does land in `<path>/fitness/` on a local install. The `.zip` clause stays.

---

Five rounds; the code has been stable since round 3, and rounds 4–5 were documentation accuracy only. Gate green after each: 738 files / 7663 tests, build and lint clean.

llun:
## Rebased onto `main` — conflicts resolved

The conflict was substantive, not textual: **#1344 landed the blank-storage-value guard independently** while this branch was in review, and this branch carried its own implementation of the same thing.

Rebased onto `adfb11698` and **dropped the duplicate in favour of main's**, which goes further — it also fixes the `getEffectiveFitnessStorageConfig` fallback (a disabled fitness storage borrowing the media bucket) and makes `cleanupMediaStorage.ts` refuse to run when fitness storage cannot be resolved. Neither was in this branch.

Worth noting the two implementations had independently converged: same `requiredStorageValue(name, subject)` signature, same tri-state (`undefined` unset / trimmed string / `null` blank), same warning text, and the same conclusion that it must live outside `lib/config/utils.ts` because `proxy.ts` reaches that module. So there was nothing left to reconcile — main's version simply supersedes.

**Dropped from this branch** (now main's): `lib/config/storageValues.ts`, and all changes to `mediaStorage.ts` / `fitnessStorage.ts` and their tests.

**Also dropped:** `storageTypeValue`, which mapped a blank `*_STORAGE_TYPE` to unset. Main deliberately left that asymmetry in place and documented it (*"Leaving the type empty … still selects the fallback, exactly as leaving it unset does"*), and main now warns on an unrecognised type, which removes the silent-failure part I was solving. That is the maintainer's call already made, so this branch does not relitigate it.

**Kept**, because main lacks them:

- **`proxy.test.ts` logger guard.** `storageValue.ts`'s own doc comment says it lives outside `lib/config/utils.ts` because that module reaches the middleware bundle — but nothing enforced it. Re-verified after the rebase: reintroducing the import fails with `lib/config/utils.ts imports @/lib/utils/logger`, and `middleware.js.nft.json` still traces zero pino files.
- **The credential-exposure detail and upgrade note.** Main documents that a blank path "would otherwise resolve to the application's working directory"; it does not say that made a legacy root `config.json`, production backup archives and `.p8`/`.pem` key material readable unauthenticated, or that affected operators should rotate.
- **The `.env.example` notes** — #1344's commit message records this as left undone ("this environment denies read access to .env\* paths").

The docs were merged by hand rather than by conflict resolution: main's blank-value prose is authoritative and mine was removed, including one place where my wording **contradicted** main (I had written that a blank `TYPE` counts as unset; main documents that `TYPE` is deliberately not blank-checked).

Diff is now 7 files / +320 −9, all builder-and-guard, no storage-resolver overlap. Gate re-run on the rebased tree: 742 files / 7757 tests, build and lint clean.

PR #1358, opened 2026-07-29
fix: match the fitness status page to the design system

llun:
Brings the fitness activity detail page (`/[actor]/[status]` for a status carrying a `.fit`/`.gpx`/`.tcx` file) in line with the design kit's `ui_kits/web/FitnessActivity.jsx` spec.

### 1. The action bar matches the global status

The page hand-rolled its own row (`flex flex-wrap gap-0.5` inside a `justify-between` parent — six controls clustered at the right on a 2px gap), so a fitness post's actions sat at different spacing from every other surface.

It now renders the shared `<Actions>`: same six controls, same `justify-between` across the full width, and it inherits the narrow-post `⋯` collapse it never had. `Actions` gains a `fullBleed` prop — **on by default**, so every existing caller is unchanged — turned off here because the card footer's own padding already puts the row at the status's left edge, with no avatar column to pull back over. The source-file link moves onto its own line so the row gets the full width.

This deviates from the mock, which shows 5 buttons with `⋯` in the header. The ask was explicitly to match the *global* bar, and the mock's layout would have re-broken the rule that a post's action set is identical everywhere.

### 2. Reactions match the design

The chips move into the `p-5` card body directly under the stat grid, where the spec puts them, instead of a bordered strip wedged between body and footer.

### 3. Analysis merges into a single table

With "All graphs", every visible chart lives in **one** bordered panel split by 1px dividers — no gaps. The load-bearing part is subtler than the border: the min/max scale labels moved out of a `grid-cols-[auto_1fr]` gutter into overlays on the plot corners, so every row's plot starts at the same x. Without that the rows wouldn't line up as a table regardless of the shared border.

### 4. Hovering a graph shows the value at that point

A bordered chip with value + unit and a pointer arrow, pinned to the hover dot on **every** visible chart, since they share one highlighted instant. It flips to the dot's left near the right edge and is clamped vertically. Each series' crosshair and dot take that series' own colour instead of all sharing the speed chart's blue. Works on touch as well as pointer.

## Scope beyond the four asks — please read

Most of what follows was **found by the review loop** rather than planned, but the diff is meaningfully wider than "restyle a page" and that shouldn't be buried:

- **A behavioural change to how heart rate is plotted.** The Analysis chart drew the *filtered* series, which drops 0 bpm dropouts and so **compacts** the array — while the chart maps sample index to elapsed time positionally. Any dropout slid the whole heart-rate axis left: a strap taking 10 minutes of a 30-minute ride to pick up put the halfway crosshair on the reading recorded at 20:00, with the three graphs beside it correctly on 15:00 and nothing on screen to say so. It now holds the last good reading across a gap. `heartRateStats` and the zone buckets are order-free tallies and keep the filtered array. **This is a data-presentation change with its own risk profile riding inside a design PR — reasonable to ask for it to be split.**
- A `-0` formatting guard applied at nine call sites (`(-0.3).toFixed(0)` is `"-0"`, and downsampled elevation bins straddling sea level hit it).
- The hover dot moved from an SVG `<circle>` to HTML — under `preserveAspectRatio="none"` a circle renders as an ellipse, measured at 10.7×6.8px on desktop and 3.7×6.8px on mobile.
- The coordinate projection unified into `getChartXPosition`/`getChartYPosition` (it existed in three copies; this PR raised the stakes by putting the dot in a different rendering system from the line).
- Touch scrubbing, plus `touch-action: pan-y pinch-zoom`.
- A WCAG AA contrast fix on the new label backdrops.
- A dead `flush` prop removed, and an unread `sampleCount` parameter dropped.

## Verification

- `yarn prettier:check` clean, `yarn lint` 0 errors, `yarn build` compiled, `yarn test` **746 files / 7850 passed**.
- Verified in a real browser in **light and dark**, at 1280×900 and 390×844, with **pointer and touch**, against a locally-seeded activity with a real `.tcx` (SQLite + local fitness storage). Touch scrubbing checked end-to-end with dispatched `TouchEvent`s: `touch-action: pan-y`, values updating across a drag, all four readouts clearing on `touchend`.
- Every behavioural fix is **mutation-tested** — the fix was reverted and the new assertion confirmed to fail. Twenty-one mutations across the loop.

## Review

Eight rounds of the sub-agent loop, one comment per round below. Twenty-six findings, all addressed or explicitly declined with reasons. Six of them were tests that could not fail, found by mutating the code rather than reading it — including four in the first round.

Round 8 was asked to judge whether to continue and said stop: R6 found real defects (two of them in R5's own fix), R7 returned zero defects and two improvements, and an eighth round was more likely to introduce a defect than find one.

## Known gaps, not fixed here

- **The `⋯` menu still lacks Edit and Quote** on this page (pre-existing — `main` passed `canEdit={false}` too). Not half-fixed: enabling `canEdit` without wiring `onEdit` renders a menu item that does nothing, and doing it properly means giving the page the shared `useInlineComposer`. Recorded in `AGENTS.md` as a known gap rather than a licensed difference.
- **The charts are unreachable by keyboard and screen reader** — the `<svg>` has no `role`/`aria-label`, and the readout is deliberately `aria-hidden` (a bare number stream on a control you cannot drive is worse than nothing). Pre-existing; the net effect is that this feature has no value for AT users.
- **The readout's flip threshold is a viewBox fraction against a fixed-pixel chip.** A pixel-aware threshold measured off the plot, the way `useCompactActionBar` measures the action row, would serve narrow and wide widths alike and end the per-round retuning.
- Deleting your own fitness post from `⋯` still strands you on a dead page (`onPostDeleted` unwired) — now a one-line fix on a shared component rather than a bespoke one.


llun:

Four reviewers, fanned out across correctness of the `Actions` refactor, the chart rewrite, project invariants + docs, and test coverage. **One reviewer found no defects; the other three found ten.** All ten are addressed in 9299a41a6.

Every fix below was **mutation-tested**: I reverted the fix, confirmed the new assertion fails, and restored it. Where a finding was that a test *couldn't* fail, that's stated explicitly.

### Correctness (3)

**1. The Analysis heart-rate chart reported the wrong instant across a sensor dropout.** `positiveHeartRateSeries` filters out 0 bpm readings, which **compacts** the array — but `ChartPanel` maps sample index to elapsed time positionally, and `buildXAxisLabels` labels the full width `0 → duration`. Any dropout therefore slid the whole heart-rate axis left. Concretely: a 30-minute ride at 1 Hz where the strap doesn't pick up for the first 10 minutes → hovering the halfway point puts the HR crosshair on the reading recorded at **20:00** while the three graphs beside it correctly show 15:00, with nothing on screen to say so. Both series are 120 points after downsampling, so the two crosshairs even land at the same x.

The compaction predates this PR, but the readout is what turns it from an unlabelled dot into a specific wrong number, and the merged panel is what frames the stack as time-aligned. Fixed by holding the last good reading across a gap (back-filling a leading one), which preserves length and still keeps 0 bpm off the plot. `heartRateStats` and `computeHeartRateZones` are order-free tallies and keep the filtered array.

**2. The readout could render `-0`.** `toFixed` keeps the sign of a value that rounds to zero — `(-0.3).toFixed(0) === '-0'`. Elevation uses `fractionDigits: 0` and legitimately goes negative (the local repro reached −16 m), and `downsampleSeries` averages bins, so a bin straddling sea level reads **"-0 m"**. Fixed with a `formatReadoutValue` helper that rounds then adds zero (`-0 + 0` is `+0`).

**3. The hover dot rendered as an ellipse, and flipped orientation between breakpoints.** `<circle r={4.5}>` inside a `preserveAspectRatio="none"` viewBox gets a non-uniform transform; `vectorEffect="non-scaling-stroke"` fixes the ring but not the geometry. Measured: **10.73 × 6.84 px** (1.57:1 wide) at a 972px viewport, **3.66 × 6.84 px** (0.53:1 tall) at 375px. Pre-existing, but this PR made the dot the focal point the readout's arrow points at. It is now an HTML span positioned by the same percentage mapping as the readout — which a reviewer verified is *exact* under `preserveAspectRatio="none"` by comparing against `getScreenCTM()` at every corner and the centre of a 906×190 box against a 760×250 viewBox.

### Test gaps (4) — all found by mutating the diff

**4. `fullBleed={false}` was untested.** Deleting it left all 33 tests passing, with the row then yanked 3.25rem out of the card entirely. (The *default* was already guarded by `post.test.tsx:1477`.)

**5. `flush` was untested** — the exact regression the new test's own comment claimed to prevent. Removing it restored `rounded-xl border` on all four charts inside the shared wrapper (literally four bordered cards again) with all 33 tests passing.

**6. The hover test could not detect a broken x→time mapping.** jsdom returns an all-zero rect, so `Math.max(bounds.width, 1)` made the divisor 1 and **every `clientX >= 1` saturated the clamp at ratio 1** — the test only ever probed the last sample. Replacing the handler body with `onHighlightElapsedSeconds(durationSeconds)` — ignoring the pointer entirely — still passed. It now stubs a real box (`left: 100, width: 400`) and hovers a quarter of the way in, so the offset, the division and the clamp are all exercised.

**7. Per-series crosshair colours and `mt-3` were unpinned.** Hardcoding both colour props back to sky-500 (reverting the whole point of `ANALYSIS_GRAPH_STYLES`) passed; so did dropping `mt-3` from `Actions`' `fullBleed` default, which silently closes the gap between every post body and its action row on every surface.

Also: `expect(getAllByTestId('analysis-graphs')).toHaveLength(1)` could never fail — `openAnalysis` returns `findByTestId`, a single-element query that throws on duplicates first. Replaced with the `flush` assertion, so its comment is now true. And an `it.each` covers both the normal and the flipped readout placement (the old test only ever hit the flipped branch).

### Docs (3)

**8. `REVIEW.md`** still carried the unqualified "chips and action row are both full-bleed" rule this PR carves an exception into — `AGENTS.md` says convention changes must update the REVIEW.md checklist too, and a reviewer following it would have flagged `fullBleed={false}` as a violation.

**9. `useReactionState`'s own doc comment and a test comment** still said the fitness page "composes its own action row" — the thing this PR removed.

**10. `AGENTS.md`/`CLAUDE.md` conflated two different props.** My added sentence attributed the pull for *both* rows to `Actions`' `fullBleed`, but `ReactionRow` has a separate same-named prop with the **inverse** default (`Post` must opt in with `fullBleed={showsActionRow}`). Both docs now say so.

### Raised and deliberately not actioned

Three pre-existing issues, all confirmed unchanged by this PR and none a regression from it — flagged so they're on record rather than silently dropped:

- Error tooltips clip inside the header card (`overflow-hidden` + a `top-full`-anchored bubble). `posts.tsx` documents omitting `overflow-hidden` for exactly this reason.
- Deleting your own fitness post from `⋯` strands you on a detail page for a status that no longer exists — `onPostDeleted` isn't passed and `DeleteDialog` neither refreshes nor navigates. Easier to fix now that the page goes through the shared `Actions`.
- The charts are mouse-only: the `<svg>` has no `role`/`aria-label`, so the new readout is unreachable by keyboard and screen reader. The charts had no accessible name before this PR either.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **7845 passed**.

llun:

Two reviewers, both pointed at round 1's fixes (fixes are where new bugs come from): a re-review of the fixes themselves, and an adversarial pass for silent degradations. **Both cleared the fixes — no defects.** Between them they found three minor issues, all addressed in 494dc12fe.

### Cleared, with the evidence

**The round-1 fixes are sound.**

- `formatReadoutValue` was run over **800,000+ random samples** at 0 and 1 decimals plus 40 hand-picked edges: 130,731 renderings changed and *every one* was a `-0`-family string becoming `0`. No other value's rendering moved, so no double-rounding artifact. `NaN` / `±Infinity` / `±1e21` byte-identical to the bare `toFixed`.
- The HTML hover dot is **geometrically identical** to the circle it replaced. Preflight `box-sizing: border-box` makes `size-[11px]` + `border-2` a 3.5px fill radius inside a 2px ring, outer r=5.5 — exactly `r={4.5}` + a 2px centred non-scaling stroke. Percentage translate resolves against the border box, so it lands dead on the point. All four extremes checked: max overflow 5.5px against `p-4`, and it's a strict improvement, because the old `<circle>` was cut in half by the viewBox at the first and last samples. `border-background` is also a dark-mode fix — `main` used `stroke-white`, a white ring on a near-black card.
- `heartRateChartSeries`: `heartRateChartSeries.length > 0 ⟺ positiveHeartRateSeries.length > 0`, so `hasHeartRate` and the chart-option gate can never disagree; and the held values form the same **multiset** as the positive readings, so pre-downsample min/max is unchanged and the printed scale can only move inward toward the true range. A lone trailing positive reading now draws a flat line where the old code emitted a single `M` with no `L` and drew literally nothing.
- Each new test confirmed to fail on the bug it names — the dropout test would read `140bpm` under the old code, not `120bpm`.

**No silent failures.** Every new Tailwind class was compiled through the project's own `@tailwindcss/postcss@4.3.3` and emits a real declaration (`size-[11px]`, `-left-[4.5px]`/`-right-[4.5px]`, `stroke-[2]`/`stroke-[1.5]`, `border-background`, all four `bg-*`, `-ml-13`, `z-10`/`z-20`). All 13 new `cn()` call sites were run through `twMerge` — nothing dropped, and `stroke-[2]` + `stroke-slate-400` correctly land in separate groups. `ANALYSIS_GRAPH_STYLES` is keyed by a closed 4-member union with all four present, so a missing key is a type error rather than a runtime `undefined`. `visibleAnalysisCharts` going empty is guarded twice and falls through to the existing "No analysis data" card. The map's static-preview fallback and the `routeDataError` banner are untouched, and the removed `showHoverMessage` prop was dead — all four call sites passed `false`, and the live "Selected time:" line is the parent's.

### Fixed (3)

**1. The overlaid scale labels sat on top of the plotted line — a regression from this PR.** On `main` the min/max labels lived in a gutter column with nothing to collide with. Moving them onto the plot put them exactly where the line is at x≈0: speed, power and heart rate nearly always *start* at their minimum, so the first path point is at exactly `y = height`, straight through the bottom-left label. Each label now carries a `rounded bg-background/85 px-1` backdrop. Verified in the browser with a series deliberately shaped to start at its minimum — the 2px stroke previously ran right through "4.5 km/h" and "63 w".

**2. The hover dot shared `z-10` with those labels** and came later in the DOM, so at the very first sample an 11px coloured dot painted over a label's corner. Labels move to `z-20` — above the dot, still below the readout, which is the thing being read.

**3. `AGENTS.md` overstated the alignment.** It said the fitness card's padding "puts each row at the status's left edge" — true of the action row, not of the chips. The card body is `p-5` and the footer `px-4`, so the two rows sit 4px apart, each aligned to its own content: the chips with the title and the stat grid, the action row with the source-file link. That's what the design kit does too (`FitnessActivity.jsx` has the chips inside the `p-5` body and the action row in a `px-4` footer), so the layout stays and the wording is now precise. `CLAUDE.md` also implied the card passes `fullBleed={false}` to both rows when it only does so for `Actions` and leaves `ReactionRow`'s prop unset.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **746 files / 7845 passed**.

Round 3 next — re-reviewing round 2's changes.

llun:

Two reviewers: one re-reviewing round 2's changes, one taking a fresh read of the whole diff for **structure and comment accuracy** rather than re-treading correctness. Addressed in f41750309 + 0c9b3c2cb.

### Round 2 cleared — no defects

Verified empirically, not by reading:

- `bg-background/85` compiled through the project's real Tailwind pipeline emits a `color-mix` composite; the `@theme inline` indirection does **not** break the `/85` modifier. Live computed value confirmed `oklab(… / 0.85)`, and `--background` is themed under `.dark`, so the backdrop is white-ish in light and near-black in dark.
- Measured `label.left - plot.left === 0` — `px-1` insets the glyphs but the backdrop box still starts at the plot's left edge, so the "every stacked chart's plot starts at the same x" claim holds.
- All five overlay elements are direct siblings of one `z-index: auto` parent, so no stacking-context subtlety; the readout wins over the labels on DOM order at equal z, exactly as the comment claims.
- `document.elementsFromPoint` at the label's centre returns the `svg` topmost — a `pointer-events: none` element doesn't become hit-testable by gaining a background, so the crosshair is unaffected.

It also confirmed those three changes are genuinely presentation-only: jsdom computes no paint order or compositing, so a test for them would assert nothing real.

### Fixed (2)

**1. Three copies of the same coordinate projection.** `buildChartPath` re-derived the (index, value) → (x, y) mapping that `getChartYPosition` already owned, and the hover crosshair derived the x term a third time. The line and the dot must agree to the pixel or the dot floats off the line it sits on — and that agreement was held together by two copies of the arithmetic. Both now go through `getChartXPosition` / `getChartYPosition`.

Worth being precise about the attribution: the reviewer said this PR introduced `getChartYPosition`; it did not, the duplication predates it. What this PR changed is the **stakes** — the dot used to be an SVG `<circle>` sharing the path's coordinate space, and is now an HTML element positioned from those same numbers as a percentage, so the agreement spans two rendering systems. The refactor is proven behaviour-preserving: the emitted path string is byte-identical to the previous inline version over a mixed test series.

**2. `ChartPanel`'s `flush` prop was dead.** One call site, always passing `true` in every filter state, so the `!flush` "own border" branch became unreachable the moment the charts were merged into one bordered panel. Dropped, with the reason the panel carries no border of its own stated where the class list is.

### And a gap the refactor exposed

Unifying the projection removed the drift risk but nothing asserted the projection itself. Mutating `getChartXPosition` to divide by `count` instead of `count - 1` left the suite **green** — the existing cases only checked the flip threshold, which both clear by a margin wide enough to absorb a wrong scale. Inverting `getChartYPosition` outright was green too.

So (0c9b3c2cb):
- the placement cases now assert `left` on the readout **and** the dot, so the two cannot move independently;
- a new case hovers the elevation fixture's minimum and maximum and pins where each lands. It also records the deliberate difference between the two elements — the dot is **unclamped** because it marks the actual point (100%/0%), while the readout is held inside the plot (92%/8%). I had that backwards in my first draft of the test and the baseline run caught it.

All three mutations now fail: broken x scale, inverted y, and removed clamp.

### Not actioned

`downsampleSeries` and the heart-rate carry-forward could live in `lib/utils/fitness.ts` for standalone unit testing. The reviewer scored this below its own reporting threshold — nothing else needs them yet and both are covered by component tests — so it stays a follow-up rather than churn in this PR.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **746 files / 7846 passed**.

llun:
## Sub-agent code review — round 4

One reviewer, on round 3's refactor and test additions. **No defects.**

- **The projection refactor is output-identical.** `getChartYPosition` now recomputes `range` per point instead of once outside the map, but from the same two doubles — IEEE arithmetic is deterministic, there is no accumulation, and the `values.length === 0` early return still precedes it. `getChartXPosition(index, values.length, width)` expands to exactly the previous inline form at both call sites. `ElevationProfileChart` on the Overview tab therefore emits an identical `d` string and renders unchanged.
- **Perf is a non-issue:** one subtraction and one `Math.max` per point over ≤120 points, and `ChartPanel` memoizes `path` on `[maxValue, minValue, values]`, so hover-driven re-renders don't rebuild it at all.
- **The `flush` removal is equivalent** — `cn('bg-background p-4', !flush && …)` with `flush={true}` produced exactly the hardcoded `"bg-background p-4"`. No reference survives anywhere, and the explanatory comment sits above `return (` rather than between it and the root element, so the parse error that removal briefly introduced is genuinely gone.
- **The new test numbers were re-derived independently** from the fixture and all four check out. All four series are length 6, so `downsampleSeries(…, 120)` is a no-op and the shared `left` assertion is valid across every chart.
- **`toHaveStyle({ top: '0%' })` was probed directly** against this jsdom/cssstyle/jest-dom combination: `0%` does not normalise to `0px`, and the assertion correctly throws against `50%`, `8%`, `0px` and a missing `top`. Not vacuous.
- **The assertions were re-mutated independently** and all three bite.

One nice catch on scope: `getBarColor` also uses `index / Math.max(1, total - 1)`, but it's a colour-interpolation ratio rather than a plot projection — correctly left out of the unification.

A final confirmation round is running on two angles the loop hasn't touched: design fidelity against the kit's own `FitnessActivity.jsx` spec, and a completeness critic asking what four rounds of review never looked at.

llun:
## Sub-agent code review — round 5

Two reviewers on angles the loop had not touched: **design fidelity** against the kit's own `FitnessActivity.jsx`, and a **completeness critic** asked what four rounds never looked at. Between them, six findings. Addressed in d6cb5c999 + fa8e2e1eb.

The completeness pass was the most valuable round of the five, because it found a defect sitting eleven pixels from something three earlier rounds had mutation-tested.

### Fixed (5)

**1. The `-0` fix was applied to one of nine call sites.** Round 1 fixed the hover readout. The scale labels are derived from the *same* downsampled bins and still went through a bare `toFixed`. The PR's own regression fixture (`altitudeSeries: [10, -0.3, …]`, six samples so `downsampleSeries` is a no-op) rendered **`Scale -0 m - 55 m`** in the panel header and `-0 m` in the overlay label, while the test asserted the readout beside it read `0m`. Renamed `formatReadoutValue` → `formatChartValue` and routed all eight min/max labels through it; the existing test now asserts the scale text.

**2. The round-2 backdrop was below WCAG AA — in the case it was added for.** `bg-background/85` composites `text-muted-foreground` to **4.10:1** over `rose-500`, 4.22 over `violet-500`, 4.40 over `sky-500` in light theme. The backdrop exists precisely because speed, power and heart rate start at their minimum and put the first path point under that label, so the failing case is the designed case. `/95` gives 4.74–4.89:1.

This is also the answer to "what modality was never run": **no contrast check happened on this diff**, and `app/globals.contrast.test.ts` structurally cannot cover it — it checks `--muted-foreground` against six *opaque* surface tokens, and a translucent surface over a raw palette colour is outside its scope.

**3. The readout was mouse-only.** `onMouseMove`/`onMouseLeave` and nothing else, so on a phone the feature this PR adds did not exist — and the charts carry an explicit mobile height. Pointer and touch now share one `scrubToClientX`; `touch-pan-y` leaves vertical swipes scrolling the page (which matters under four full-width charts) while claiming horizontal drags. CSS rather than `preventDefault`, so it works with passive listeners. Verified in a real browser: `touch-action: pan-y`, values updating across a drag (19m→45m, 12.2→7.4 km/h), all four readouts clearing on `touchend`.

**4. The readout clipped below ~360px.** Its flip threshold is a fraction of the viewBox while the chip is a fixed pixel width; at 320px — the WCAG 1.4.10 reflow target — the two stop agreeing and the merged panel's `overflow-hidden` cut up to 10.6px off. Capped its width. 390px was verified earlier and passes, which is why this was missed.

**5. The readout is now `aria-hidden`.** It was the only new *text* on the page an AT user could reach, in no live region, on a control they cannot drive — a bare figure with no context. Its numbers are already in the panel header's "Scale …".

Also: `buildXAxisLabels` took a `sampleCount` it never read, making the ticks look series-dependent when they are identical on every row.

### Documented, not changed (1)

**The compact-collapse threshold differs between surfaces.** `useCompactActionBar` measures the row's own border box *including* the `-ml-13` pull, so a `fullBleed={false}` row is 52px narrower and collapses in a ~466–499px band where the same status in the timeline does not.

I disagree that this is a bug. `AGENTS.md` is explicit that the width comes from a `ResizeObserver` on the row and **not** a viewport breakpoint, precisely because available width is what matters — and the fitness footer genuinely has less of it. The rule is working. It was undocumented, so it is now.

Likewise **Edit and Quote are still missing from this page's `⋯` menu** — pre-existing (`main` passed `canEdit={false}` too). I deliberately did not half-fix it: turning `canEdit` on without wiring `onEdit` renders a menu item that does nothing, and wiring it properly means giving the page the shared `useInlineComposer`, well beyond these asks. Recorded in `AGENTS.md` as a known gap rather than a licensed difference, since my own doc edit had implied more parity than the code delivers.

### A test that looked right and wasn't

The critic's suggested one-word fix for the `mt-3` gap was `not.toHaveClass('-ml-13', 'mt-3')`. I applied it, then mutated — and **hoisting `mt-3` out of the `fullBleed` branch still passed**. jest-dom's negated multi-class form passes when *any* listed class is absent, so with `-ml-13` gone the assertion never looked at `mt-3`. Split into two assertions; both directions now fail correctly.

### Corrections to earlier rounds

- Round 2's "the held values form the same **multiset** as the positive readings" is wrong — carry-forward duplicates readings, so same *set*, not multiset. The min/max conclusion stands; the mean does not, which is exactly why `heartRateStats` correctly keeps using the filtered array.
- Two changes appear in no commit message: the path stroke thinned `stroke-[2.5]` → `stroke-[2]`, and `vectorEffect="non-scaling-stroke"` was added to the `<path>` (previously only the crosshair had it). Both deliberate — they match the design kit's `strokeWidth="2"` + `vectorEffect`, and without the latter `preserveAspectRatio="none"` distorts stroke width with the container. Recording them here since this PR documents everything else.
- **Dark mode had only ever been inferred**, never observed. Now checked in the browser at 1280×900: readout chips, label backdrops and per-series crosshairs all render correctly on the near-black card.

### Surfaces the critic ruled out as non-gaps

`processing`/`failed`/`pending` never reach this component (`StatusBox` and `page.tsx` both gate on `completed`); a remote actor's fitness post is impossible (`status.fitness` comes only from the local `fitness_files` table); logged-out and the multi-file switcher are both covered by existing tests.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **746 files / 7848 passed**. Verified in a real browser in light and dark, at 1280×900 and 390×844, with pointer and touch.

llun:
## Sub-agent code review — round 6

One reviewer, on round 5's touch and contrast work. **Three findings, all real** — and two of them were defects in the fix I had added the round before. Addressed in bcbb86f0e.

### Fixed (3)

**1. `touch-pan-y` silently disabled pinch-zoom over the whole Analysis section.** The shorthand compiles to *exactly* `touch-action: pan-y` — I compiled it through the project's own Tailwind 4.3.3 to confirm:

```css
.touch-pan-y { --tw-pan-y: pan-y;
  touch-action: var(--tw-pan-x,) var(--tw-pan-y,) var(--tw-pinch-zoom,); }
```

With only `--tw-pan-y` set, `pinch-zoom` is dropped. Four `h-[190px]` charts stacked cover most of a phone viewport, so this took magnification — and double-tap-to-zoom — away over the one part of the page a low-vision reader most wants to magnify. `touch-pinch-zoom` composes it back to `pan-y pinch-zoom`, which still leaves the single-finger horizontal drag unclaimed for scrubbing.

**2. A tap left the readout stuck on.** `touchend` cleared the highlight, and the browser then fired its compatibility `mousemove` at the tap point, which re-entered the scrub — and no `mouseleave` ever follows a touch, so it stayed on indefinitely. Drags were unaffected, because movement past the tap slop suppresses the compat events, so the surface behaved one way for a drag and the opposite for a tap. `preventDefault()` on `touchend` suppresses the sequence; the chart has no click behaviour to lose, and a gesture the browser claims for scrolling fires `touchcancel` instead. Not reproducible in jsdom, so the test asserts the event comes back cancelled rather than simulating the compat sequence.

**3. The `max-w-[calc(100%-1.5rem)]` I added last round could not do what its comment claimed.** The class is valid and `100%` does resolve against the right containing block — but the chip is clipped by where it is *positioned* (`left: X%` plus a 12px translate), and `max-width` constrains neither. Worse, a cap wide enough never to truncate the text is 196px against a ~76px chip, so it was inert in every reachable geometry.

The real fix is the flip threshold. The design kit's `0.72` is a fraction of the viewBox while the chip is a fixed pixel width, so the two only agree above some container width. At the 320px WCAG 1.4.10 reflow target the plot is ~220px and 0.72 puts the chip's right edge 10px past what `overflow-hidden` will show. `0.62` is the widest round value that fits there; it flips slightly sooner in a desktop column, which costs nothing.

### Checked and sound

- `scrubToClientX`/`clearScrub` being new closures each render is fine — they are React props on a delegated event system, not `addEventListener` registrations, so no listener churn and no stale capture.
- `touches[0]` is correct for `touchstart`/`touchmove` (the `changedTouches` trap doesn't apply, since `touchend` reads no touch data). Multi-touch degrades gracefully: `ratio` is clamped, so a second finger can only pin the readout to a chart edge.
- `aria-hidden` on the readout has no focusable descendants, so no axe violation, and is the right call over `role="status"` — which would announce a stream of bare numbers during a drag.
- `formatChartValue` on the eight scale labels changes nothing but `-0` cases; speed's `toFixed(1)` round-trips byte-identically.
- Both AGENTS.md additions re-verified by tracing the container chains: timeline row is `W − 66` and the fitness row `W − 100`, so the quoted 466–499px compact band is exactly right.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **746 files / 7849 passed**. Both new behaviours mutation-tested: dropping `touch-pinch-zoom` and dropping the `preventDefault` each fail.

llun:
## Sub-agent code review — round 7

One reviewer, on round 6's touch fixes. **No blocking defects.** Two of its secondary observations were worth taking; addressed in 33f86a258.

### Cleared, with the evidence

- **`preventDefault()` on `touchend` genuinely works.** Confirmed by reading React 19's event system in `node_modules`: it passes `passive: true` for exactly `touchstart`, `touchmove` and `wheel`; every other event including `touchend` is attached with `passive` unset. No passive-listener warning.
- **No collateral breakage.** The chart has no click behaviour, and the page's one Radix overlay (the section dropdown) is `modal` by default, so the body gets `pointer-events: none` and the `<svg>` isn't hit-tested while it's open — the suppressed compat `click` can't block outside-dismissal, which was the one path that would have mattered.
- **The 0.62 arithmetic was re-derived by measurement, not estimate.** Plot at a 320px viewport is exactly 220px through the real container chain, and chip widths measured in a browser with the project font stack: `13.5 km/h` **77.2px** (widest realistic), `166 bpm` 70.5, `2456 m` 67.1, `240 w` 57.0. At 0.62 the worst case lands 10.4px inside; at the old 0.72 it clipped by 11.6px — matching the "10px past" in the previous commit message, so the diagnosis was right and the fix works.
- Removing `max-w-*` regressed nothing: it resolved to 196px against a 77px chip, so it could never bind.

### Fixed (2)

**1. The `preventDefault()` was unguarded.** Once a scroll is underway Chrome keeps dispatching `touchend` with `cancelable: false` rather than switching to `touchcancel` — so the call was a no-op that logs *"Ignored attempt to cancel a touchend event with cancelable=false"* on every vertical swipe begun on a chart, which under four stacked full-width charts is most of them. Now guarded on `event.cancelable`, and the comment's claim about `touchcancel` corrected.

**2. Nothing pinned the flip threshold, which had been retuned twice.** The existing placement cases sit at ratios 0.20 and 1.00 — clearing both 0.62 and the old 0.72 by a margin wide enough to notice neither, as their own comment admits.

The six-sample fixture's 0.2 grid cannot reach a point between the two values, so the new case uses an eleven-sample series to hover exactly 0.70. It asserts the **bound**, not the constant: retuning up to 0.66 (which also fits) stays green, while a return to the kit's 0.72 fails. Verified both ways — 0.72 fails, 0.66 passes.

Also corrected a claim I had written: 0.62 was not "the widest round value that still fits" — 0.66 fits and 0.67 is the first that doesn't. The comment now states the margin it actually buys.

### Recorded, not actioned

The reviewer's structural suggestion is right and worth saying plainly: **a pixel-aware flip** (`ratio*W + 12 + chipWidth > W`, measured off the plot the way `useCompactActionBar` measures the action row) would satisfy both narrow and wide widths and end the per-round retuning. At a 832px desktop plot, 0.62 flips with ~29% of the width still free, versus ~17% at the kit's 0.72. That is a better shape than a magic fraction, but it is a new mechanism at round 7 of a review loop for a cosmetic detail, so it stays a follow-up rather than a same-PR rewrite.

### Gate

`prettier:check` clean · `lint` 0 errors · `build` compiled · `test` **746 files / 7850 passed**.

llun:
## Sub-agent code review — round 8 (final)

One reviewer, with two jobs: review round 7's changes, and judge honestly whether this loop should stop.

### Job 1 — round 7's changes: no findings

- The `cancelable` guard is the only `preventDefault` call site, and skipping the call when `cancelable === false` is behaviourally right, not merely quieter: a browser that has already committed the gesture to scrolling doesn't fire the compat mouse sequence, so there is nothing left to suppress.
- The eleven-sample fixture's mismatch with its six-sample siblings is **required**, not laziness — elevation lands at `left: 70%` while the other three land at 80%, which is exactly why the new test asserts `[0]` rather than looping like the pre-existing cases.
- `left: '70%'` was checked for float drift (`(7/10) × 760` is the classic way to get `69.99999999999999%`): it rounds clean at both steps, 3.4e-14 from 532.0 against a half-ulp of 5.7e-14.
- The threshold is genuinely **bracketed** to `[0.20, 0.70)` — the new test alone would pass with a threshold of 0.0; the pre-existing `clientX: 200` case supplies the lower bound.
- Mutation-verified both directions independently: 0.72 fails, 0.66 passes.

Two comment inaccuracies, fixed in 5cbef7a:
- the geometry left this panel's own 16px right padding implicit, so a reader re-deriving it lands on 0.59 and concludes the comment is wrong;
- "the widest chip the four series **can** produce" overstated a measurement of the widest they *realistically* produce (a 3-digit speed or 4-digit power renders wider — 0.62 still fits those, with an 87.6px budget against a ~84px chip).

I also removed the closing clause proposing a pixel-aware threshold. It's a design proposal, and those belong in the PR body and an issue, not in a source comment — which should say what the code does and why it is this way, not argue for what it should become. It's now in the PR description under known gaps.

### Job 2 — stop

The trajectory is the argument: **R6 found three real defects, two of them in R5's own fix** (`touch-pan-y` silently dropping pinch-zoom, with a compiled-CSS proof; a tap leaving the readout stuck on; a `max-w` that couldn't do what its comment claimed). **R7 returned zero defects and two improvements** — and its headline fix was console-warning spam, which is the clearest possible convergence signal. An eighth round is now more likely to introduce a defect than find one, which is precisely the pattern R5→R6 demonstrated on this PR.

It went looking for an unexamined risk category and found none new. It re-checked the highest-blast-radius file that rounds 2–7 never revisited — `actions/actions.tsx`, whose `fullBleed` default reaches every post surface — and confirmed the `cn()` rewrite emits the same class set with no twMerge-conflicting groups.

The one thing genuinely still open is what **round 1 named on day one and nobody returned to**: the charts have no keyboard or screen-reader story. That's now in the PR description rather than filed and forgotten.

### What it said a human would push back on

Taken and acted on: **the scope is wider than "restyle a page" and shouldn't pass unremarked.** The PR description now names it, and singles out the heart-rate carry-forward as a data-presentation change with its own risk profile riding inside a design PR — reasonable to ask for it to be split.

Not taken: `getAllByTestId(…)[0]` assumes elevation is the first visible chart, which the pre-existing cases already assume. Real coupling, opaque failure mode, but not worth a ninth round.

### Loop closed

**8 rounds · 26 findings · 21 mutation tests · 10 commits.** Six findings were tests that could not fail. Gate green throughout: `prettier:check` · `lint` 0 errors · `build` · **746 files / 7850 tests**.

PR #1362, opened 2026-07-30
minor: add a combined overlay option to the activity analysis graphs

llun:
## Summary

The activity **Analysis** section could only draw its series stacked one above another. A single-select "graph display" filter chose one series or all of them, so comparing (say) speed against heart rate over the same climb meant reading two charts a screen apart and aligning them by eye.

This adds a **Combined** display mode alongside the existing **Separate** one, matching the design:

- The graph-display card now leads with a **Separate / Combined** segmented toggle and turns its series chips into a **multi-select** (each with its series-colour dot) instead of a single-select "All graphs" filter. Every series starts selected, so the default view is unchanged; toggling a chip adds or drops that series in whichever mode is active.
- **Combined** mode overlays every selected series in one chart, **each scaled to its own range** so a shallow trace is not flattened by a taller one. It carries a legend, the shared time x-axis, and the "Each graph is scaled to its own range." note, and it scrubs like the stacked charts do: one crosshair, a coloured dot on each line, and a single readout listing every series' value at the hovered instant — driving the same map highlight.
- The stacked **Separate** view keeps its per-series panels and behaviour. A hint covers the case where every series is toggled off.

Implementation lives entirely in `app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx` (new `CombinedChartPanel`, `graphDisplayMode` + `selectedGraphKeys` state, multi-select chips + toggle) with tests in the co-located spec. No API, schema, migration, env var, or docs surface changed.

## Screenshots

Verified in a real browser (local SQLite + a seeded public GPX ride carrying elevation/heart-rate/speed track data). Screenshots below are shared with the reviewer directly (they are not committed — the repo forbids committing PR screenshots).

- **Separate mode (default, unchanged):** stacked Elevation / Speed / Heart rate panels, each scaled to its own range, with the new multi-select chips.
- **Combined mode:** the three series overlaid in one "Combined" chart with a legend, `0:00 … 42:20` time axis, and the "scaled to its own range" note.
- **Combined hover:** crosshair + per-series dots + a single readout box ("67 m · 23.2 km/h · 161 bpm").

## Checklist

- [x] Docs updated: I grepped `*.md` and `docs/` for every command, env var, route, script, or convention this PR renames, removes, or reshapes (AGENTS.md → Documentation Maintenance) — this change touches only one client component's internal UI, nothing documented

---
_Generated by [Claude Code](https://claude.ai/code/session_019bBrm4F8mwZAxgqAHDuCWE)_

### DIFF
--- a/app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
+++ b/app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
@@ -167,7 +167,15 @@ type SectionKey =
   | 'comments'
 
 type AnalysisGraphKey = 'elevation' | 'speed' | 'power' | 'heart-rate'
-type AnalysisGraphFilter = 'all' | AnalysisGraphKey
+// How the selected graphs are drawn: `separate` stacks each into its own row of
+// one bordered panel (the original behaviour); `combined` overlays them all in a
+// single chart, each series scaled to its own range.
+type GraphDisplayMode = 'separate' | 'combined'
+
+const GRAPH_DISPLAY_MODES: Array<{ id: GraphDisplayMode; label: string }> = [
+  { id: 'separate', label: 'Separate' },
+  { id: 'combined', label: 'Combined' }
+]
 
 interface SectionTab {
   id: SectionKey
@@ -242,11 +250,13 @@ interface MapboxModule {
   ) => MapboxLngLatBounds
 }
 
+// One toggle per series (no "all" pseudo-option): the graphs to show are picked
+// as a multi-select, so "show everything" is simply every chip on — which is the
+// default — rather than a distinct mode.
 const ANALYSIS_GRAPH_OPTIONS: Array<{
-  id: AnalysisGraphFilter
+  id: AnalysisGraphKey
   label: string
 }> = [
-  { id: 'all', label: 'All graphs' },
   { id: 'elevation', label: 'Elevation' },
   { id: 'speed', label: 'Speed' },
   { id: 'power', label: 'Power' },
@@ -255,15 +265,32 @@ const ANALYSIS_GRAPH_OPTIONS: Array<{
 
 // One colour per series, used for the line, the hover crosshair and the hover
 // dot alike — so a stacked graph is identifiable by its own colour rather than
-// every crosshair sharing the speed chart's blue.
+// every crosshair sharing the speed chart's blue. `chipBorder` tints a selected
+// picker chip and the combined-chart legend with that same series colour.
 const ANALYSIS_GRAPH_STYLES: Record<
   AnalysisGraphKey,
-  { stroke: string; dot: string }
+  { stroke: string; dot: string; chipBorder: string }
 > = {
-  elevation: { stroke: 'stroke-slate-400', dot: 'bg-slate-400' },
-  speed: { stroke: 'stroke-sky-500', dot: 'bg-sky-500' },
-  power: { stroke: 'stroke-violet-500', dot: 'bg-violet-500' },
-  'heart-rate': { stroke: 'stroke-rose-500', dot: 'bg-rose-500' }
+  elevation: {
+    stroke: 'stroke-slate-400',
+    dot: 'bg-slate-400',
+    chipBorder: 'border-slate-400'
+  },
+  speed: {
+    stroke: 'stroke-sky-500',
+    dot: 'bg-sky-500',
+    chipBorder: 'border-sky-500'
+  },
+  power: {
+    stroke: 'stroke-violet-500',
+    dot: 'bg-violet-500',
+    chipBorder: 'border-violet-500'
+  },
+  'heart-rate': {
+    stroke: 'stroke-rose-500',
+    dot: 'bg-rose-500',
+    chipBorder: 'border-rose-500'
+  }
 }
 
 // `toFixed` keeps the sign of a value that rounds to zero, so an elevation bin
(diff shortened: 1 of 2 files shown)


app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1681,8 +1936,24 @@ export const FitnessStatusDetail: FC<Props> = ({
...
+    useState<GraphDisplayMode>('separate')
+  // Which series are drawn, as a multi-select. Every graph starts on, so the
+  // default view is every available series (a key with no data is filtered out
+  // downstream), and the picker chips toggle each on and off. Intersecting this
+  // intent with the series that actually have data is what lets it stay stable
+  // as the selected file changes underneath it.
+  const [selectedGraphKeys, setSelectedGraphKeys] = useState<
+    AnalysisGraphKey[]
+  >(() => ANALYSIS_GRAPH_OPTIONS.map((option) => option.id))
+  const toggleGraphKey = (key: AnalysisGraphKey) =>
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1951]: **Test coverage:** `graphDisplayMode` and `selectedGraphKeys` are independent state driving two separate render branches, but nothing tests that a chip toggled off in one mode stays off after switching Separate↔Combined. This cross-cutting interaction is exactly what can silently regress. Add a test that deselects a series, flips the mode, and asserts it stays out.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2197,6 +2489,10 @@ export const FitnessStatusDetail: FC<Props> = ({
     return `rgb(${r}, ${g}, ${b})`
   }
 
+  // The chips to offer — only series that actually have data. A chip the user
+  // toggled off is simply dropped from `visibleAnalysisCharts`, and one whose
+  // series is absent from the current file never appears here, so the picker
+  // needs no reset effect to recover from an empty selection.
   const analysisGraphOptions = useMemo(() => {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2496]: **Test coverage (higher priority):** this diff removed the old reset `useEffect` that recovered a filter stuck on a series with no data. The replacement safety is this availability filter (chips only render for data-bearing series) plus the `values.length > 0` guard in `visibleAnalysisCharts`. That whole data-availability path is only exercised by the happy case — every series in the `routeData` fixture is populated. Add a test rendering a file with an empty series (e.g. `powerSeries: []`) that asserts the Power chip is absent and the combined chart draws N−1 lines, so the removed effect can't regress silently.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2653,63 +2938,138 @@ export const FitnessStatusDetail: FC<Props> = ({
...
+                      </button>
+                    )
+                  })}
                 </div>
                 <p className="mt-2 text-xs text-muted-foreground">
                   {highlightedElapsedLabel
                     ? `Selected time: ${highlightedElapsedLabel}`
-                    : 'Hover any graph below to follow that time point on the map.'}
+                    : graphDisplayMode === 'combined'
+                      ? 'Pick the graphs to overlay in one chart. Hover it to follow that time point on the map.'
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:3006]: **Minor:** the combined-mode helper copy isn't asserted anywhere (only the separate-mode empty hint is). Cheap to pin so a copy change is caught.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1230,6 +1257,234 @@ const ChartPanel: FC<{
...
+            />
+          ))}
+          {crosshairX !== null ? (
+            <line
+              x1={crosshairX}
+              y1={0}
+              x2={crosshairX}
+              y2={height}
+              vectorEffect="non-scaling-stroke"
+              className="stroke-muted-foreground stroke-[1.5] opacity-50"
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1420]: **Cosmetic (note, likely won't-fix):** the single crosshair is drawn at the continuous `ratio * width` (the hovered *time*), while each series' hover dot snaps to its own nearest sample x. When two overlaid series were downsampled to different lengths, the dots don't land exactly on the crosshair (sub-pixel for the driver series, up to ~3 viewBox-px for a 120-point series). The readout **values** are correct. Keeping the dot on its own line (not on the crosshair) is the intended visual, and snapping the readout to the nearest sample matches the Separate charts' semantics — so this is a deliberate trade-off rather than a bug. Flagging for the record.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2653,63 +2938,138 @@ export const FitnessStatusDetail: FC<Props> = ({
...
-                    </button>
-                  ))}
+                <div className="flex flex-wrap items-center justify-between gap-3">
+                  <p className="text-[11px] font-semibold uppercase tracking-wide text-muted-foreground">
+                    Graph display
+                  </p>
+                  {/* Segmented toggle: `separate` stacks each graph, `combined`
+                      overlays them in one chart. */}
+                  <div
+                    role="group"
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2948]: **Accessibility (suggestion):** Separate/Combined is a mutually-exclusive single-select, so a `role="radiogroup"` + `role="radio"`/`aria-checked` pattern would convey "one of two" more precisely than two independent `aria-pressed` toggle buttons. The current `role="group"` + `aria-pressed` toggle pattern is a well-established accessible segmented-control pattern and needs no extra arrow-key handling, so either is defensible. Raising only as a consideration.

---

llun (commented):
Sub-agent code-review round 1 (correctness, project invariants, tests/a11y/design). No correctness bugs and no invariant violations found. Findings are test-coverage gaps plus two notes; inline comments below. Addressing the actionable ones next.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1951]: Fixed in 70d46d4 — added `keeps a deselected series out after switching display modes`: it deselects Elevation in Separate mode, flips to Combined and asserts `aria-pressed=false` plus only 3 overlaid lines, then flips back to Separate and asserts Elevation is still absent from the stack.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2496]: Fixed in 70d46d4 — added `offers no chip for a series that has no data`: it renders with `powerSeries: []`, asserts the Power chip is absent (while Elevation/Speed/Heart rate remain), and asserts the combined chart overlays only 3 lines. That exercises the availability filter that replaced the removed reset effect.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:3006]: Fixed in 70d46d4 — added `explains the overlay in combined mode`, which asserts the exact combined-mode helper string.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1420]: Keeping as-is (deliberate). In the common case the overlaid series share a length (uniform full-duration downsampling), so the offset is sub-pixel. Placing each dot on its own line — rather than on the shared crosshair — is the intended read, and snapping the readout to the nearest sample keeps combined-mode values identical to the Separate charts. Interpolating each dot onto the crosshair would fix the pixels but diverge the readout values between the two modes, which is a worse trade. No change.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2948]: Keeping the `role="group"` + `aria-pressed` toggle pattern. A true `radiogroup` would also require arrow-key roving-tabindex handling to be correct, which the segmented-toggle pattern doesn't; state is already conveyed programmatically via `aria-pressed` and each button has a text label. No change.

---

PR #1363, opened 2026-07-30
fix: trim only the start and end of a route for fitness privacy zones

llun:
## Summary

Fitness privacy zones decided visibility **one track point at a time** — `isPointHiddenByPrivacyLocations` was true for any point inside any zone circle, and `buildPrivacySegments` then started a brand-new segment every time that flag flipped. A route that started at home, left the zone, came back through it mid-ride and left again was drawn as **several disconnected polylines** with a circular hole punched at every crossing.

A zone now only ever trims the **head and the tail** of a route. The radius does double duty: it is both the area test (does the activity start, or finish, inside the circle?) **and** the length of route hidden from that end, measured as cumulative along-route distance.

Precisely:

- `R_start` is the largest `radiusMeters` among zones containing the first point, `0` if none contains it. `firstVisibleIndex` is the smallest `i` where `points[i]` is outside **every** zone **and** the cumulative distance from the first point is `>= R_start`.
- The tail mirrors it from the last point with `R_end`. **Each end is judged independently** — a point-to-point activity that starts at home and finishes elsewhere loses only its head, and its line runs right through to the final track point.
- Everything between the trimmed ends is exposed as **one contiguous run**, even where the route passes back through a zone.
- If no qualifying point exists on an end, or the two trims meet, nothing is exposed and no route map is produced (the existing "no exposed route" paths — `removeRecordedMap`, skipping map generation — are unchanged).

The whole change lands in `lib/services/fitness-files/privacy.ts` as a new `getPrivacyVisibleRange`, with `annotatePointsWithPrivacy` and `getVisibleSegments` reimplemented on top of it. **All four consumers keep their existing plumbing untouched**: the static route map (`processFitnessFileJob`, `regenerateFitnessMapsJob`), the interactive status map (`/api/v1/fitness-files/[id]/route-data`), and route heatmaps. `buildPrivacySegments` now sees at most two flag flips and yields at most `hidden head → visible run → hidden tail`.

Two supporting changes:

- **Renamed** `isPointHiddenByPrivacy` / `isPointHiddenByPrivacyLocations` to `isPointInsidePrivacyLocation` / `isPointInsideAnyPrivacyLocation`. Being inside a circle is no longer the same thing as being hidden, and conflating the two is precisely the bug. Neither had any consumer outside the module.
- **Zero-radius zones are filtered in `normalizePrivacyLocations`.** `getFitnessPrivacyLocations` never emits one, but a caller can pass one, and a point sitting exactly on such a centre read as "inside" (`distance <= 0`) while contributing a hide length of `0` — so it could block a trim nothing asked for.

Route heatmaps annotate **per file, inside the page loop, before any cross-file accumulation**, so every activity's own ends are trimmed rather than only the first and last of the pooled point list. There is a regression test for exactly that.

### The trade-off, stated plainly

**A mid-route pass through a privacy zone is now visible.** If a route leaves a zone and later passes back through it — a loop that returns past your house, laps of a block — those coordinates are exposed to every viewer of the post, the route data, and the heatmap.

This is a deliberate exchange rather than an oversight. Hiding each pass individually punched a **circular hole into the polyline**, and a hole centred on the private location advertises it at least as loudly as the coordinates do — the same reasoning already documented for the public heatmap embed in `lib/services/fitness-files/publicHeatmap.ts`. It also fragmented one activity into several disconnected lines, which is what was reported.

Two smaller consequences of the same rule:

- Because the trim is now measured as cumulative along-route distance and not just circle membership, a short activity that starts or finishes in a zone without ever covering the hide length is hidden **entirely**, where it previously showed whatever lay outside the circle (e.g. a 200 m warm-up lap under a 500 m zone).
- When the head trim and the tail trim meet, nothing is exposed and no route map is produced. Reachable with a start/finish near the *edge* of a large zone on a short out-and-back.

### Operational notes

No migration — no schema change, so neither reference schema dump is touched. Existing artefacts were clipped under the old rule and are re-derived by existing user-facing actions:

- **Route map images** — **Regenerate maps for old statuses** on the fitness privacy settings page. A previously fragmented map becomes one continuous line; a route whose window now collapses takes the `removeRecordedMap` path and loses its attachment, which is correct but not reversible for that image.
- **Owner route heatmaps** — per-region **Regenerate**, or `scripts/fitness/recreateFitnessRouteHeatmaps.ts` in bulk. Note that saving privacy settings does not auto-invalidate heatmap caches; that is pre-existing and out of scope.
- **Public heatmap embeds are byte-identical before and after.** `flattenPrivacySegmentsForPublic` drops the flag but keeps every coordinate, so the flag moving position changes nothing there.
- The interactive status map needs nothing — `route-data` recomputes per request.

## Screenshots

Verified in a real browser against a local SQLite instance with a hand-written GPX that starts on the zone centre, runs ~1.4 km out, returns **through** the centre, heads out again and finishes ~2.6 km away, with a 200 m zone saved at the start point.

I cannot attach image files to a PR body from this environment, so here is exactly what the browser showed:

**The generated static route map** (the artefact every viewer of the post sees — produced by `getVisibleSegments`, which has no owner special-case) rendered as **one unbroken red line**: the green start marker sits offset from the true start by the trimmed head, the line runs out to the NE vertex, comes back **straight through the zone centre with no gap**, heads out to the NW vertex and continues to the red end marker. No fragmentation, no circular hole. The tail is not trimmed, because the activity finishes 2.6 km clear of the zone.

**Route data** exposed by the server for that activity:

| activity | owner segments |
| --- | --- |
| starts in zone, passes back through, finishes 2.6 km away | `[hidden 4 pts][visible 117 pts]` |
| starts **and** finishes in the zone | `[hidden 4 pts][visible 113 pts][hidden 4 pts]` |

Four hidden points is the 200 m trim at ~55 m per point, and in both cases the middle is a single unbroken run containing the mid-route return to the centre. Under the old rule the first activity produced *two* visible segments with a hole between them.

**Privacy settings page** renders the new copy: the card now reads "Trim the start and finish of your routes around your saved privacy locations, on shared maps and generated route images", and the Hide Radius helper reads "When a route starts or finishes here, that end is hidden from other viewers until it leaves the area and has covered this distance. The middle of a route is never cut, so a route that later passes back through the area stays visible."

Two caveats, reported rather than papered over:

- The sandbox's egress proxy intermittently answers OSM tile requests with `503`, so two of the three uploads could not render their static map. That surfaced correctly as the owner-only "The route map image could not be generated. Everything else in this activity is intact." banner with a **Retry** button — i.e. `mapError` persisted and retryable, exactly as documented — and it is an environment limitation, not a code path this PR touches.
- The logged-out route-data view was not exercised in the browser: the composer created these fitness posts as `private`, so an anonymous request correctly 404s. That path is covered by tests instead, including a new one asserting a non-owner gets exactly **one** visible segment containing the mid-route pass. The static map image above is in any case the non-owner geometry.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — lint 0 errors, **746 test files / 7908 tests passing**. Run on Node 22 (only 20/21/22 available in this container; `engines` pins 24 and CI uses 24).
- [x] Docs updated: `docs/fitness-file-storage.md` (pipeline step 3, the Security and Privacy bullet, and the two "entirely inside a privacy zone" phrasings that are now incomplete), `docs/features.md`, `docs/environment-variables.md` (`ACTIVITIES_FITNESS_ROUTE_HEATMAP_FILE_POINT_LIMIT` — the decimation now has a privacy consequence worth naming), the settings UI copy, and the `publicHeatmap.ts` doc comment. Grepped `*.md` and `docs/` for every other description of zone hiding.
- [x] No migrations changed, so neither schema dump needed regenerating

### Tests added

- `privacy.test.ts` — 56 tests. A `getPrivacyVisibleRange` table covering both trim directions, the length-vs-area distinction (a point that has left the circle but not covered the hide length stays hidden), overlapping zones taking the largest radius, a candidate skipped for sitting inside a *different* zone, zero-radius zones, empty/single-point input, an all-inside route, a stationary trace, and a standalone case for the head and tail trims crossing. Plus `annotatePointsWithPrivacy` asserting the exact boolean array for a re-entering route, `buildPrivacySegments` yielding exactly `[true, false, true]` for a route that dips into the zone three times mid-route, `getVisibleSegments` returning the caller's own point objects with no added flag (which is what protects the jobs' deep-equality assertions), and direct tests for the two renamed primitives.
- `route-data/route.test.ts` — a mid-route pass stays in one segment for other viewers, and only the head is trimmed when a re-entering route finishes outside every zone.
- `processFitnessFileJob.test.ts` — one unbroken `routeSegments` entry containing the mid-route pass. Also extracted a shared `setPrivacyZone`/`clearPrivacyZone` upsert helper: the file shares one actor and one `general` settings row, and adding a second creator exposed that the existing privacy test was order-dependent. Now verified green under `--sequence.shuffle.tests` across three seeds.
- `generateFitnessRouteHeatmapJob.test.ts` — two activities in one heatmap both get their own ends trimmed (`[true, false, true, false]`), proving annotation is per file rather than over the pooled list.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkqDAGBxAeMhAD4yzmf2d1)_


docs/fitness-file-storage.md
```diff
@@ -150,6 +150,6 @@ See [Maintenance Scripts](maintenance.md) for general script guidance.
 
 - Fitness files can contain sensitive location data.
 - Fitness posts default to private visibility in Strava flows unless the user chooses otherwise.
-- Privacy locations hide configured coordinate radii from route maps, route-data responses, and route heatmaps.
+- Privacy locations trim the head and tail of a route — not its middle — from route maps, route-data responses, and route heatmaps. A route that re-enters a zone part-way through is exposed there; see the Processing Pipeline above for why. A route whose exposed window collapses is not rendered at all. Route heatmaps annotate each activity separately, before pooling them, so every activity's own ends are trimmed rather than only the first and last of the pool.
```
llun: **This bullet overstates what heatmaps do, and it is the one place in the diff where the docs are wrong about privacy.**

Listing "route heatmaps" alongside route maps and route-data implies a public heatmap viewer sees the trimmed geometry. It does not. `flattenPrivacySegmentsForPublic` (`lib/services/fitness-files/publicHeatmap.ts:26`) maps each segment to `{ points: segment.points }` — it drops the `isHiddenByPrivacy` flag but **keeps every coordinate**, deliberately, because a missing donut around home pinpoints the location as well as the coordinates do. So on the heatmap the trim only changes which segments the **owner's authenticated view** recolours; the public embed renders the head and tail like any other geometry, before and after this change.

Claiming otherwise in the Security and Privacy section is the worst place to be imprecise — someone could set a zone expecting their heatmap embed to be trimmed.

Two smaller inaccuracies in the same area:

- Pipeline step 3 (line 109) says "or the two trims meet". The guard is `firstVisibleIndex > lastVisibleIndex`, which fires when they **cross**, not when they meet — `first === last` yields a one-point window, not `null`.
- "A route whose exposed window collapses is not rendered at all" is right for the map image, but should say so rather than reading as a claim about every surface.

---

lib/services/fitness-files/privacy.ts
```diff
@@ -248,23 +260,165 @@ export const isPointHiddenByPrivacyLocations = (
...
+    hideLengthMeters = Math.max(hideLengthMeters, privacyLocation.radiusMeters)
+  }
+
+  return hideLengthMeters
+}
+
+// Walks inward from one end of the track and returns the first index that has
+// both cleared every zone and travelled the anchor zone's radius from that end.
+// `step` is +1 walking from the first point, -1 walking back from the last.
+// Distance is cumulative along the route rather than straight-line, so looping
```
llun: **This comment argues the property backwards.**

> Distance is cumulative along the route rather than straight-line, so looping around the block near home does not buy the trim any credit.

Cumulative along-route distance is the **more permissive** of the two measures, not the stricter one: by the triangle inequality it is always `>=` the straight-line distance from the anchor, so it reaches the threshold *sooner*. Looping around the block near home is exactly what **does** buy the trim credit under this code — each lap adds metres to `travelledMeters` while the rider stays close to home, and once the points are outside the circle they qualify.

Straight-line-from-the-anchor is the reading that would deny that credit. The chosen semantics are still correct (they are what "length of route hidden" means, and it is what the settings copy now promises), but the justification needs to be the honest one, and the trade-off it implies — a circumferential route near a large zone clears the length sooner than a radial one — is worth stating rather than inverting.

---

lib/services/fitness-files/privacy.ts
```diff
@@ -248,23 +260,165 @@ export const isPointHiddenByPrivacyLocations = (
...
+  return { firstVisibleIndex, lastVisibleIndex }
+}
+
+/**
+ * `points` MUST be one activity's track, in recorded order. This is no longer a
+ * per-point predicate: the flag is derived from where each point sits relative to
+ * the route's own two ends, so handing it several routes concatenated together
+ * would trim only the outermost pair of ends and expose the rest.
+ *
+ * The three callers all satisfy that: `route-data` maps one file's `trackPoints`,
```
llun: "The three callers all satisfy that" then names two — `route-data` and `generateFitnessRouteHeatmapJob`. There are only two: `getVisibleSegments` used to route through `annotatePointsWithPrivacy` on `origin/main`, but this diff reimplements it on `getPrivacyVisibleRange` + `slice`, so it is no longer a caller.

The count matters more than a typo normally would, because the comment exists specifically to tell a future editor which call sites must keep passing a single ordered route. A reader who counts two and is told to expect three will go looking for a third and either invent one or distrust the comment.

Same fix needed at line 378: "The two trims met or crossed" — the guard is `>`, so only crossing reaches it.

---

lib/services/fitness-files/privacy.test.ts
```diff
@@ -68,10 +100,312 @@ describe('getDistanceMeters', () => {
...
+      zones: [zoneAtMetersNorth(0, 100)],
+      expected: { firstVisibleIndex: 2, lastVisibleIndex: 5 }
+    },
+    {
+      description:
+        'hides a point that has left the circle but not covered the hide length',
+      // Starts 90m from the centre of a 100m zone, so index 1 and 2 are already
+      // outside the circle while only 20m and 40m along the route. This is the
+      // radius acting as a LENGTH rather than only as an area.
+      offsets: [90, 110, 130, 200],
```
llun, line 239: **Nothing in this table pins the defining mechanic of the change: that the hide length is cumulative along the route rather than straight-line from the anchor.**

Verified by mutation. Replacing the accumulator at `privacy.ts:315`

```ts
travelledMeters += getDistanceMeters(points[index - step], points[index])
```

with a straight-line measure

```ts
travelledMeters = getDistanceMeters(points[anchorIndex], points[index])
```

leaves the **entire** fitness surface green — `privacy.test.ts` 56/56, and all of `lib/services/fitness-files`, the three consumer job/route suites, and the fitness components. The mutant is behaviourally real (crow-flies is always `<=` cumulative, so it trims strictly more), yet no assertion notices.

Every fixture is blind to it for a structural reason, which the file's own header comment (lines 36–41) concedes: the offsets sit on a single north–south line through the zone centre, so "outside the circle" and "R metres from the anchor" collapse into the same predicate. The two cases with an offset anchor (`[90, 110, 130, 200]` and `[95, 105, 98, 102]`) happen to give the same answer either way. The crossing test at line 319 is blind too — under the mutant the head scan returns `null`, so the function returns `null` for a different reason and `toBeNull()` still passes.

A separating fixture, arithmetic checked both ways:

```ts
{
  description: 'measures the hide length along the route, not from the anchor',
  offsets: [90, 10, 110, 300],
  zones: [zoneAtMetersNorth(0, 100)],
  expected: { firstVisibleIndex: 2, lastVisibleIndex: 3 }
}
```

Cumulative: 80 m at index 1 (inside the circle anyway), 180 m at index 2 whose point is 110 m out — so index 2 qualifies. Straight-line from the anchor: index 2 is only 20 m away, so it is skipped and the range becomes `{3, 3}`.

---

lib/services/fitness-files/privacy.test.ts
```diff
@@ -68,10 +100,312 @@ describe('getDistanceMeters', () => {
...
+    },
+    {
+      description: 'uses the largest radius among zones containing the start',
+      // With only the 100m zone the head would clear at index 1.
+      offsets: [0, 200, 400, 700, 1000],
+      zones: [zoneAtMetersNorth(0, 100), zoneAtMetersNorth(0, 500)],
+      expected: { firstVisibleIndex: 3, lastVisibleIndex: 4 }
+    },
+    {
+      description: 'treats a zero-radius zone as no zone at all',
```
llun, line 271: Three fixtures in this table do not discriminate the behaviour they are named for.

**`treats a zero-radius zone as no zone at all` (line 257) passes with the new filter deleted.** With `offsets: [0, 100, 200]` and one `radiusMeters: 0` zone, the unfiltered path reaches `getAnchorHideLengthMeters`, finds the start "inside" (`distance 0 <= 0`) but contributes a hide length of `0`, hits `hideLengthMeters <= 0` and returns the anchor index — the same answer. The filter in `normalizePrivacyLocations` only actually bites in the *scan* predicate, where a zero-radius zone sitting on a candidate point would mark it inside and skip it. That is the case worth pinning: a real zone containing the start plus a zero-radius zone parked on the first otherwise-qualifying point.

**`uses the largest radius among zones containing the start` (line 250) does not distinguish largest from summed.** With radii 100 and 500 over `[0, 200, 400, 700, 1000]`, `max = 500` and `sum = 600` both land on index 3. Radii 100 and 200 over `[0, 150, 250, 400]` separate them: max 200 gives index 2, sum 300 gives index 3.

**The `lastVisibleIndex === null` branch has no coverage at all.** In every fixture that returns `null` the *head* scan fails first and short-circuits, so a mutant that stopped honouring a finish inside a zone would survive. Reaching it needs the head to succeed and the tail to fail, which means a start outside every zone and a tail that runs out of points before covering `R_end` — e.g. `offsets: [1100, 800, 500]` against a 1000 m zone at the origin: the start is outside so `R_start = 0` and `firstVisibleIndex = 0`, while the backward scan accumulates only 300 m then 600 m and returns `null`.

---

lib/jobs/generateFitnessRouteHeatmapJob.test.ts
```diff
@@ -428,6 +428,99 @@ describe('generateFitnessRouteHeatmapJob', () => {
...
+    const reentryRoute = (lngOffset: number) => [
+      zoneCentre,
+      { lat: 52.3604, lng: 4.8794 },
+      { lat: 52.37, lng: 4.89 + lngOffset },
+      zoneCentre,
+      { lat: 52.365, lng: 4.895 + lngOffset },
+      { lat: 52.375, lng: 4.885 + lngOffset }
+    ]
+
+    await database.createFitnessSettings({
```
llun, line 445: `createFitnessSettings` sits outside the `try`, so only the job call is protected by the `finally` that deletes the row. Anything that throws between here and the `try` — `createCompletedFitnessFile`, the mock wiring — leaks a `general` settings row for the shared actor, and `createFitnessSettings` throws on a second insert for the same actor/serviceType. The next test in the file that wanted its own settings would then fail with a confusing "Fitness settings already exist" rather than its own assertion.

Low likelihood, but the fix is free: open the `try` immediately after the create so the `finally` covers everything that follows. `processFitnessFileJob.test.ts` in this same diff already had to grow an upsert helper for exactly this shared-row hazard, so it is worth not reintroducing it here.

---

llun (commented):
Sub-agent review, round 1 — five independent lenses (algorithm, consumer integration, privacy exposure, tests, project invariants), each finding then put to two independent skeptics told to refute it, one of them required to attempt a reproduction.

17 raw findings; 6 survived and are posted inline. What was refuted, so it is on the record rather than silently dropped:

- **"An end is trimmed based on a single GPS sample, so a route starting a few metres outside a zone gets no trim."** True, and intended — the radius is the area test for the *endpoint*, and the docs already say "an end that sits in no zone is not trimmed at all".
- **"The hidden head/tail segments starve the visible route of the 1500-point route-data budget."** `allocateSegmentTargets` prioritises by segment length, so the visible middle is served first and the small head/tail cap out at their own length. Strictly better than `origin/main`, which could face many segments each claiming the 2-point minimum.
- **"A one-point window survives the collapse guard, so `annotatePointsWithPrivacy` and `getVisibleSegments` disagree."** Real but harmless: `getVisibleSegments` correctly reports nothing drawable, and one point that is outside every zone *and* a full hide length along the route from both ends is not an exposure. Tightening the guard to reject every sub-2-point window would wrongly hide a legitimate single-point activity that never involved a zone at all. Kept as-is; the misleading *wording* around it is a real finding and is posted inline.

Nothing in the algorithm or the four consumer call sites survived scrutiny — the surviving findings are one genuine documentation error about heatmaps, two inverted/miscounted code comments, and three test-coverage gaps (one of them verified by a surviving mutant on the central mechanic). Addressing all six now.

---

docs/fitness-file-storage.md
llun: Fixed in e2be01f. This was the one finding with real user consequences, so the bullet is now split into two: route maps and route-data get the head/tail trim, and heatmaps get their own bullet saying plainly that the trim changes what the **owner's** view highlights while `flattenPrivacySegmentsForPublic` keeps every coordinate for public and shared embeds — with the reason (a gap where a route's ends should be pinpoints the location as precisely as the coordinates do) and a pointer to that function's own comment for the trade-off the instance owner accepted.

Pipeline step 3 now says "cross" and adds the parenthetical that trims meeting exactly leave a single point, which is not a drawable route either. The collapse sentence is scoped to the map image.

---

lib/services/fitness-files/privacy.ts
llun: Correct, and fixed in e2be01f. The comment now states which way the measure cuts instead of inverting it: cumulative is always `>=` crow-flies so it clears the threshold *sooner*, a route that winds near home banks metres while staying close to it, and a circumferential route around a large zone clears the length before a radial one. It also names what stops that being an exposure — a point still has to be outside every zone to qualify, so the circle carries the area guarantee and the length only decides how far past it to keep going.

---

lib/services/fitness-files/privacy.ts
llun: Both fixed in e2be01f. The contract comment now says "Both callers satisfy that" and names the two.

On line 378 I changed the wording rather than the guard, and it is worth recording why: tightening it to reject every sub-two-point window would also reject a legitimate single-point track that no zone ever touched (`R_start = R_end = 0` gives `first === last === 0`), which would hide an activity for a reason that has nothing to do with privacy. So the guard stays `>` and the comment now says the trims *crossed*, notes that meeting exactly leaves a one-point window, and says that `getVisibleSegments` reports it as nothing drawable — which is the answer every caller that renders a line actually needs.

---

lib/services/fitness-files/privacy.test.ts
llun: Added the fixture verbatim in e2be01f, and verified it kills the mutant rather than assuming it would:

| mutant | before | after |
| --- | --- | --- |
| `travelledMeters =` crow-flies from anchor | 56 passed | **1 failed** / 57 passed |

The comment on the fixture spells out both arithmetic paths so the next reader does not have to redo it, and says explicitly that this is the case standing between the accumulator and a "simplification" into a straight line.

---

lib/services/fitness-files/privacy.test.ts
llun: All three fixed in e2be01f, and each one is now mutation-verified rather than reasoned about:

| mutant | before | after |
| --- | --- | --- |
| drop the `radiusMeters > 0` filter in `normalizePrivacyLocations` | 56 passed | **1 failed** / 57 passed |
| `Math.max` → sum in `getAnchorHideLengthMeters` | 56 passed | **1 failed** / 57 passed |
| `findFirstUnhiddenIndex(…, -1) ?? points.length - 1` (ignore a failed tail scan) | 56 passed | **1 failed** / 57 passed |

- **Zero-radius**: the zone is now parked on index 2, the first point that would otherwise clear the real 100 m zone, so the unfiltered path skips a candidate for a zone that hides nothing. The comment notes why the anchor position proves nothing — a zero radius already falls out there via `hideLengthMeters <= 0`.
- **Largest radius**: switched to radii 100/200 over `[0, 150, 250, 400]` as suggested; max 200 lands on index 2, summing to 300 would land on index 3.
- **Failed tail scan**: added `[1100, 800, 500]` against a 1000 m zone, with the comment spelling out that this is the only fixture that reaches the guard because every other `null` short-circuits on the head.

---

lib/jobs/generateFitnessRouteHeatmapJob.test.ts
llun: Fixed in e2be01f — the `try` now opens immediately after the create and encloses the file creation, the mock wiring, the job call and the cache read, so the `finally` covers everything that could throw.

Chasing this turned up the real instance of the same hazard, in a different file: the route-data suite shares one settings row per actor too, and `serves route samples for public statuses without requiring a session` fails under `--sequence.shuffle.tests` when a zone is left behind — its two-point default route gets trimmed to one visible point. That one is **pre-existing**: I checked out `origin/main`'s `route-data/route.test.ts` and `privacy.ts` and reproduced the same failure at seed 11, so the old per-point rule hid that first point too. My two new tests only made it likelier to surface, so rather than leave it I added an `afterEach` that resets the row, with a comment recording that it predates these tests.

Verified across four shuffle seeds over all four affected files: 124 passed each time.

---

llun:
Sub-agent review, round 2 — three lenses over the round-1 fix commit (fixture arithmetic re-derived and mutation-tested, comment/doc truth-checked against the code, test hygiene plus shuffled-order proof), each finding put to a skeptic. One finding survived, and it was a direct sibling of round 1's: fixed in a198ae8.

**Round 1 fixed the heatmap claim in one doc and left it standing in the other.** `docs/features.md:64` — a line this PR had itself rewritten — still said privacy locations trim "route maps, route data, **and heatmaps**", which is exactly the claim round 1 removed from `docs/fitness-file-storage.md` because `flattenPrivacySegmentsForPublic` keeps every coordinate for public and shared embeds. Two documents in the same PR asserting opposite privacy guarantees for the same surface, with the wrong one being the top-level feature list linked from the README.

I also fixed the matching promise in the **Privacy Location card**, which said the trim applies "on shared maps" — that reads as covering shared heatmap embeds. A reviewer refuted it on the grounds that the round-1 commit did not touch that file, which is true but the wrong test: the copy arrived in this PR's first commit, and the settings page is where a user actually forms the expectation. It now names the surfaces that are trimmed and says plainly that heatmaps are not, with the reason.

Two findings were refuted with work worth recording:

- **"The rewritten largest-radius fixture still survives a min/first-wins mutation."** Accurate, and empirically demonstrated — but the pre-round-1 fixture was concentric too and survived the identical mutant, so nothing was introduced or left behind. The mutant round 1 targeted was sum-instead-of-max, and that one now dies on exactly one test where it previously survived. Surviving-mutant findings can always be extended with one more mutant; this one is a nice-to-have row, not a defect.
- **"The new tail-scan comment over-generalises about the sibling null cases."** The load-bearing half is correct and was verified by instrumenting every `return null` site and running all seven null fixtures through it: the new fixture is genuinely the only one that reaches the `lastVisibleIndex === null` guard. The trailing clause misattributes 2 of the 6 others (the empty-list row exits at the length guard, and the trims-cross case reaches its guard after both scans succeeded). No behavioural surface, and it does not license deleting anything — wording, not a defect.

Order-independence re-proved over six shuffle seeds across all four affected test files. Gate green: prettier, lint (0 errors), build, 746 files / 7910 tests.

Round 3 running now over this delta.

---

llun:
Sub-agent review, round 3 — **clean**. Two lenses (repo-wide copy/doc consistency after a198ae8, and a completeness critic asked to name what is *missing* from the whole diff rather than re-check what is present), five raw findings, all five refuted on verification. Per AGENTS.md → Code Review Loop that closes the loop at three rounds.

The refutations were all "real mechanics, but pre-existing on `origin/main` and untouched here", each demonstrated by showing the file byte-identical to main rather than argued:

- `regenerateFitnessMapsJob` has no test configuring a privacy zone — true, and equally true on main; it consumes the API with the same three lines as `processFitnessFileJob`, which is tested.
- `route-data` ships the power/HR/altitude/speed series covering the trimmed ends — the file is byte-identical to main; the series carry no coordinates.
- The stronger trim is not retroactive for already-attached and federated maps — pre-existing (stored maps have always been import-time snapshots), and the PR body already tells owners to run **Regenerate maps for old statuses**.
- A copy-emphasis opinion on the rewritten card.

**One refuted finding I acted on anyway, as documentation.** The critic pointed out that `GET /api/v1/fitness-files/:id` serves the original `.fit`/`.gpx`/`.tcx` bytes to any request — signed in or not — whenever the attached status is public or unlisted, and that the activity detail page links it as the source-file download for every viewer. That file carries the full track, including the ends the map and route data trim. The skeptic was right that this is entirely pre-existing and no part of this diff, and I have not changed any of it.

But this PR rewrote the **Security and Privacy** section, and that section now promises privacy locations trim the route. A reader would reasonably infer their location is protected on a public activity, and it is not — the trim covers what the server renders, not the file it stores. Leaving that unsaid would have made the section I wrote misleading, so `eb9efa9` states it plainly and points at the control that does work: visibility. No code change.

Worth an issue of its own if you want the raw download gated or the source link owner-only — happy to open one, but it is a product decision (the link is presumably there on purpose) rather than something to slip into this PR.

Final state: three review rounds, seven findings addressed, gate green on every commit (prettier, lint 0 errors, build, 746 files / 7910 tests).

---

PR #1365, opened 2026-07-30
fix: restrict the raw fitness file download to its owner

llun:
## Summary

Follow-up to #1363, which trimmed the head and tail of a route around privacy locations and documented — as an accepted limit — that the uploaded file itself was still public. This closes that.

`GET /api/v1/fitness-files/:id` served the **original** `.fit`/`.gpx`/`.tcx` to any request, signed in or not, whenever the attached status was public or unlisted. That file carries the whole track, including the ends the route map and route data now hide, so a viewer could take back exactly what every other surface had just hidden. Making an activity public does not have to mean handing out the athlete's raw telemetry.

**⚠️ This removes a capability.** A non-owner can no longer download the source file of a public activity. That is deliberate, and it is the point of the change.

### What changed

**The endpoint is owner-only.** The status lookup and visibility test are gone; a non-owner gets `404` rather than `403`, so the response does not confirm the id resolves to a real file. That also retires `public, max-age=31536000, immutable` — a year-long shared-cache TTL that outlived the visibility change meant to withdraw it. Responses are now `private, no-store`, plus `X-Content-Type-Options: nosniff` and `Content-Disposition: attachment`: the served `Content-Type` comes from the stored object, and on the S3 path that value originates from the presigned upload's client-supplied field, which has no allow-list. Owner-only already reduces that from stored to self-XSS; the two headers make it inert.

The filename in that header goes through a new `buildAttachmentContentDisposition`, because `sanitizeStoredFileName` deliberately leaves quotes alone — legal in a file name, but they close a quoted-string early. It emits a reduced quoted form plus an RFC 5987 `filename*`, so non-ASCII names survive intact.

**The media route was a way around all of it.** With no explicit `ACTIVITIES_FITNESS_STORAGE_TYPE` — the default — fitness storage falls back to media storage *inside the media root*: a `fitness/` key prefix in the same S3 bucket, or a `fitness` directory under `ACTIVITIES_MEDIA_STORAGE_PATH`. `GET /api/v1/files/*` has no access control at all and 308-redirects to the public CDN hostname when one is configured. I verified this rather than assuming it — against `origin/main`'s route, an anonymous request for the real stored path returned **200 and the full 12,076-byte GPX**. The route now refuses the reserved fitness prefixes, matched on a path-segment boundary so a legitimate `fitnessed/…` media key is unharmed. Map images and their JPEG email copies are written under `medias/`, so nothing legitimate is caught.

**Three UI surfaces linked the file, not the one I expected.** Alongside the fitness detail footer, the shared `Post` card and the DM bubble both build an anchor from `status.fitness.url` — which is how an anonymous visitor to a profile still had a working link until a browser check caught it. All three now link only for the owner.

The file **name** still shows to every viewer. It is the card's label and what the multi-file selector switches between; the decision here was about the download, not about hiding metadata. Only the anchor and the download affordance go.

Nothing federated is affected: the URL is stripped from every outbound payload (`getNoteFromStatus`, the AP note/replies payload, `getMastodonAttachment`), with existing regression tests locking that in, and it never appeared in an email. No job, script, backup or `lib/client.ts` function fetches it — every server-side reader goes through `getFitnessFileBuffer`/`getFitnessFile` directly. The owner's own Files page is unaffected.

### Operational note

Copies a CDN already cached under the old one-year `immutable` header stay reachable until they expire; there is no way to revoke them from the app.

## Screenshots

Anonymous view of a **public** fitness activity on a profile — the card keeps its filename label, type badge and stats, and the filename is no longer a link:

> `reentry-route.gpx` · gpx — Distance 7.05 km · Duration 40:00 · Avg speed 10.6 km/h · Elevation 103 m

Measured rather than eyeballed, on the same public activity:

| surface | owner | anonymous |
| --- | --- | --- |
| profile page | 1 link | **0 links** (was 1) |
| home timeline | 1 link | 0 links |
| `/fitness/files` | 1 link | n/a (owner-only page) |
| `GET /api/v1/fitness-files/:id` | `200`, `private, no-store`, `nosniff`, `attachment; filename="reentry-route.gpx"` | **`404`** |
| `GET /api/v1/files/fitness/<real stored path>` | — | **`404`** (was `200`, 12,076 bytes on `origin/main`) |
| `GET /api/v1/files/fitnessed/x.webp` | — | `200` — segment-boundary matching intact |

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — lint 0 errors, **746 test files / 7933 tests**
- [x] Docs updated: the endpoint list entry in `docs/fitness-file-storage.md` (which was doubly wrong — it is now owner-only, and no implementation redirects) and the **Security and Privacy** bullet added by #1363, which documented this exposure as accepted. Added a bullet recording that the media-route reservation is load-bearing rather than defensive. Grepped `*.md` and `docs/` for other claims of public downloadability.
- [x] No migrations changed, so neither schema dump needed regenerating

### Tests added

- `app/api/v1/fitness-files/[id]/route.test.ts` — the "serves files for public statuses without requiring a session" case is inverted to assert `404`; added a signed-in non-owner also getting `404`, the owner's response carrying all three headers, and a stored file name (`ru"n; attachment; filename="evil.html`) that cannot break out of the header.
- `app/api/v1/files/[...pathname]/route.test.ts` — a table over reserved fitness paths (nested, bare, differently-cased, reached by traversal) returning `404`, and a table proving `medias/…` and `fitnessed/…` still resolve.
- `lib/services/medias/fileName.test.ts` — `buildAttachmentContentDisposition` over quotes, semicolons, non-ASCII, the characters `encodeURIComponent` spares, and the fallback, plus an invariant that the header never carries a bare quote outside its two parameter values.
- Component tests on all three link surfaces: `post.test.tsx`, `FitnessStatusDetail.test.tsx` (including that a logged-out reader keeps the action row — the footer's gate had to be decoupled from the now-owner-gated href, or gating it would have taken the actions away too).

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkqDAGBxAeMhAD4yzmf2d1)_


llun:
Sub-agent review, round 1 — four lenses (bypass, header construction, UI regression, project invariants), each finding put to a skeptic told to refute it. Three survived; all fixed in fa11989.

### The guard was bypassable — three ways

This is the one that mattered, and it is my bug. The check compared the request path after `path.normalize`, but that string does not stay that string: with `ACTIVITIES_MEDIA_STORAGE_HOSTNAME` set, the value is interpolated into `https://<hostname>/<path>` and handed to `Response.redirect`, which runs the **WHATWG URL parser**. That parser normalises two things `path.normalize` does not — a backslash is a segment separator for special schemes, and `%2e` counts as a dot segment. A third form never collapses locally at all and is decoded at the origin instead.

Reproduced all three before fixing:

| request path | guard | `Location` emitted |
| --- | --- | --- |
| `medias\..\fitness/x.gpx` | **passes** | `https://cdn…/fitness/x.gpx` |
| `medias/%2e%2e/fitness/x.gpx` | **passes** | `https://cdn…/fitness/x.gpx` |
| `%66itness/x.gpx` | **passes** | `https://cdn…/%66itness/x.gpx` → origin decodes |
| `fitness/x.gpx` | 404 | — |

The last row is why this was easy to miss: the plain form kept 404ing, so the guard looked effective in its own tests while the owner-only gate was fully circumvented for anyone who knows or recovers the stored key.

The comparison now runs on a canonical form in a new `isReservedFitnessMediaPath` — backslashes become separators, escapes decode to a fixed point (Next already decodes once, so a double-encoded prefix is still one decode away), dot segments resolve, leading separators drop. The redirect branch **re-checks the parsed `Location`** before returning it, so the two checks fail independently. Mutation-tested: removing the decoding, the backslash handling, or the dot-segment resolution each fails tests that previously passed (4, 1 and 4 failures respectively).

### I took the owner's own download away on `/fitness`

The gate reads `props.currentActor`, and `RecentFitnessActivities` rendered `<Posts>` without it — on a page that shows nothing but the signed-in actor's own activities, loaded via `getFitnessFilesByActor({ actorId: currentActor.id })`. So the single viewer this PR is meant to keep serving was the only one denied, while the endpoint would have handed them the bytes. Now threaded through, deliberately **without** `showActions`, so the list stays read-only per the Status Posts & Actions rules.

Worth noting the reviewer's observation that my own component test (`withholds the download from a logged-out reader`, `currentActor: undefined`) asserted the correct component behaviour while masking that a signed-in owner reached that branch in production. Correct test, wrong call site.

### The docs described my first draft, not what shipped

The Security and Privacy bullet said the source-file row is hidden from non-owners and that the file name is a disclosure in its own right. That was true of my first pass; I then changed the design so the name stays as the card's label and only the link goes, and did not update the bullet. It now says what the code does, names all three surfaces, and gives the advice that actually follows — rename the file or drop below public visibility if the name itself matters.

Refuted, for the record: a claim that the reserved prefix strips a trailing slash the S3 driver never adds (accurate mechanically, but pre-existing and inert), and several scope-creep suggestions.

Gate green on the fix: prettier, lint 0 errors, build, **747 files / 7949 tests**. Round 2 running over this delta.

---

llun:
Sub-agent review, round 2 — three lenses over the round-1 fix commit, with the fuzzing lens pointed at the new canonicalisation since that is now the security boundary. Four findings survived; all fixed in eff71e8.

### 🔴 Round 1 leaked the actor's private key into the page HTML

This is the worst thing in this PR and I introduced it while closing a privacy hole. Stating it plainly.

Round 1 threaded the viewer into `RecentFitnessActivities` as `cleanJson(currentActor)`. `cleanJson` is `JSON.parse(JSON.stringify(x))` — it clones, it does not narrow — and `getActorFromSession` returns the full domain `Actor`, which extends `ActorProfile` with `privateKey`, `publicKey` and the entire `account` row: `email`, `passwordHash`, `verificationCode`, `passwordResetCode`, `emailChangeCode`, `role`. That prop crosses into a Client Component, so React serialised **all of it** into the RSC flight payload embedded in the `/fitness` HTML. The key there is the one that signs outbound ActivityPub HTTP signatures.

TypeScript could not catch it: the argument is a variable rather than an object literal, so no excess-property check applies, and `Actor` is structurally assignable to an `ActorProfile` prop. `yarn build` passed clean.

The `Actor` type's own comment says these fields are kept off `ActorProfile` *"so they never leak into public profile shapes"*, and every other page in the group (`search`, `bookmarks`, `explore`, `favorites`, `lists/[id]`, `messages`, `tags/[tag]`, `collections/[id]`, `[actor]`, `[actor]/[status]`) strips with `getActorProfile` first. This one now does too — and that helper is already covered by a test asserting it drops keys and account.

### The canonicalisation still had a hole

`decodeURIComponent` throws on the **first** malformed escape anywhere in its input, and I was decoding the whole joined path — so one bad segment protected every other one:

| request path | guard 1 | `new URL(...).pathname` | origin decodes to |
| --- | --- | --- | --- |
| `%66itness/x.gpx?%zz` | **passed** | `/%66itness/x.gpx` | `/fitness/x.gpx` |
| `%66itness/x.gpx#%zz` | **passed** | `/%66itness/x.gpx` | `/fitness/x.gpx` |
| `fit\tness/x.gpx` | **passed** | `/fitness/x.gpx` | — parser strips the tab |

Decoding is now **per segment**; a decoded segment is folded back through the same split so `%2Ffitness` cannot smuggle a separator; and tab/CR/LF are stripped up front because the URL parser removes them from a path.

Those rows also settle a question my own comment got wrong: the redirect re-check is **load-bearing, not belt-and-braces**. It is what caught the `?%zz` form while guard 1 was still fooled.

### Two things round 1 fixed had no test at all

- The **entire redirect branch** was unreached — every existing case mocked a buffer response, so the `new URL` parse, the malformed-URL 404 and the second reserved-path check were all dead code as far as `yarn test` was concerned.
- The `/fitness` regression round 1 fixed was likewise unpinned: `RecentFitnessActivities.test.tsx` had not been given the newly-required prop (test files are not type-checked, so it never surfaced) and its `Posts` stub structurally could not observe it.

Both covered now, including that the list stays read-only — `currentActor` without `showActions`.

### Mutation results

| mutant | failures |
| --- | --- |
| decode the whole path again (the round-1 bug) | 9 |
| stop stripping tab/CR/LF | 3 |
| delete the redirect re-check | 1 |
| drop `currentActor` from `Posts` on `/fitness` | 1 |

Gate green: prettier, lint 0 errors, build, **747 files / 7963 tests**. Round 3 running.

---

llun:
Sub-agent review, round 3 — three lenses (a serialization-leak sweep prompted by round 2's finding, another pass at the canonicalisation, and a whole-diff completeness sweep). **Two findings, both low severity, no high or medium.** Fixed in 1c95185.

**A local fitness directory under the media root was never reserved.** The reserved list only knew about `ACTIVITIES_FITNESS_STORAGE_PREFIX`, which is read in the S3/object branch. An operator setting `ACTIVITIES_FITNESS_STORAGE_TYPE=fs` names the directory with `ACTIVITIES_FITNESS_STORAGE_PATH` instead, and nothing stops that landing inside `ACTIVITIES_MEDIA_STORAGE_PATH` under some other name — `LocalFileStorage.getFile` serves anything below the media root and requires no `medias` row, so `/var/uploads/fit-files` beside `medias/` was readable through the media route with no access control. The same hole this reservation exists to close; it just would not have been called `fitness`.

The list now derives that segment when the configured local path resolves inside the media root, and reserves nothing when it is outside (unreachable anyway) or equal to the root — reserving `''` would refuse every media object on the instance.

**The DM bubble was the one restricted surface with no non-owner test.** The only case covering it asserted the sender's link; `post.tsx` and `FitnessStatusDetail` both got explicit signed-in-other/logged-out pairs. A recipient now has one proving the card keeps its name, type and metrics without an anchor.

Both mutation-tested — dropping the local-path reservation and restoring the DM anchor unconditionally each fail a test that passes now.

Worth recording what was **refuted**, since several were carefully argued and reproduced before being dismissed:

- **A configured prefix without a trailing slash makes the reservation a no-op** — matcher arithmetic real, claimed bypass not reproducible; the values tested all still match.
- **Trailing NUL / space / `?` / `#` are stripped by the URL parser but not the matcher** — the reviewer drove the real handler with all four inputs and got `404` and no `Location` on every one.
- **`buildAttachmentContentDisposition` throws `URIError` on an unpaired surrogate** — the mechanism is real (`isStrippedCharacter` does not cover U+D800–DFFF), but no path in this codebase produces one; latent, not live.
- **The endpoint gates on the account while the three UI surfaces gate on the actor** — confirmed the divergence can only fail safe: `isOwner` true implies the same actor, which implies the same account, so the UI can never offer a link the endpoint refuses. Also pre-existing on `origin/main`.
- **Round 2's `getActorProfile` narrowing is pinned by nothing** — there is no `page.test.tsx`, but `getActorProfile` itself is covered by a test asserting it drops keys and account.

Gate green: prettier, lint 0 errors, build, **747 files / 7970 tests**. Round 4 running to confirm a clean round.

---

llun:
Sub-agent review, rounds 4 and 5 — **the loop is closed.** Round 5 returned clean (zero confirmed findings).

**Round 4** found one thing: a stale comment of mine, and the mirror image of the error round 2 caught. The block above the redirect tests claimed the egress re-check was the *only* guard stopping `%66itness/x.gpx?%zz` — true before round 2, but round 2's per-segment decoding fixed the primary guard in the same commit. Two of my test files were contradicting each other. Corrected in f7590f8.

**Round 5** returned `clean: true`. Its single raw finding was refuted as a wording quibble — but the refutation noted the mechanism "reproduces", so I checked it myself rather than taking the clean result at face value. It was right, and my round-4 correction had swapped one overclaim for another:

| request path | ingress guard | egress guard |
| --- | --- | --- |
| `fitness?x/y.gpx` | ✗ misses | ✓ **catches** |
| `medias?/../fitness/y.gpx` | ✓ catches | ✗ misses |

Next decodes a catch-all segment once, so `%3F` arrives as a literal `?`. That keeps `fitness?x/y.gpx` off the `fitness/` prefix on the way in, while `new URL()` splits the query off and leaves a pathname of `/fitness` for the second check. It runs the other way too, because the parser stops the pathname at the `?` before a `..` can resolve.

So **neither check subsumes the other — they overlap, and that is the real reason both exist.** Fixed in f225738, with the egress-only case now pinned by a test that also asserts `getMedia` was reached, proving the first check genuinely let it through. That makes the guard non-deletable: removing it used to fail one test, now it fails two.

That is the third comment in this PR the code disproved — one per round for three rounds. The pattern is worth naming: each was an assertion about *which* guard catches *what*, written in prose next to code that then changed underneath it. The lesson went into the commit message rather than another comment, and the claim itself is now a test.

### Final state

Six commits, five review rounds, severity curve high → high → low → low → clean.

| round | findings | worst |
| --- | --- | --- |
| 1 | 3 | **high** — the media-route guard was bypassable three ways |
| 2 | 4 | **high** — round 1 leaked the actor's private key into the page HTML |
| 3 | 2 | low — a local fitness dir under the media root was unreserved; DM card untested |
| 4 | 1 | low — stale comment |
| 5 | 0 | — |

Gate green: prettier, lint 0 errors, build, **747 files / 7971 tests**.

---

PR #1367, opened 2026-07-31
fix: explain green route segments on hover instead of a bar over the map

llun:
## Summary

A fitness activity's route map draws the stretches a privacy location hides from other viewers in green, and a bar pinned to the map's bottom-left corner explained why. The bar sat on top of the map it was explaining.

This PR started by making that bar dismiss once per browser, which bounded how long it covered the map but did not stop it covering the map. So the approach is replaced outright: **the explanation now hangs off the green line itself.** Hover a hidden stretch on a pointer device, or tap it on a touch one, and a small chip reads "Hidden from other viewers" just above the pointer.

The map is never obstructed, the answer appears where the question does, and there is no dismissal state left to keep — `routePrivacyNotice.ts`, its per-browser `localStorage` key, and the `currentActorId` prop that existed only to scope it are all deleted.

## How each renderer answers "is the pointer on a green line?"

- **Mapbox GL / MapLibre GL** hit-test their own layers, so the hint hangs off a transparent 24px twin of the 4px green line, plus layer-scoped `mousemove` / `mouseleave` / `click`. The wide twin is not cosmetic: a line's hit test is its own `line-width/2` per side, making the painted line a ±2px target, and opacity is not consulted by the hit test at all — so a zero-opacity 24px layer is hoverable while drawing nothing. It is added before any listener names it, because a layer-scoped listener registered against a missing layer is dropped in silence.
- **Apple MapKit** has no pointer events on overlays and no hover event at all, so it answers geometrically. `getDistanceToHiddenSegments` (new, in the shared `mapGeometry`) measures the pointer against the hidden segments in *scaled degrees* — longitude narrowed by `cos(latitude)`, or an east-west tolerance would feel roughly twice as loose as a north-south one at European latitudes. The tolerance comes from projecting a second point one tolerance-width away through the same conversion, which keeps it a true pixel distance at any zoom without reading the region span or the element's box.

The chip is measured and clamped inside the map by `getRoutePrivacyHintPlacement`: centred above the pointer where there is room, flipped below when there is not, held inside both side edges. The panel is `overflow-hidden`, and the hidden stretches are the trimmed start and finish of the route — which `fitBounds` parks against the bounding box's rim — so an unclamped chip failed on the most likely hover.

Touch and mouse are told apart before the auto-retire timer is armed. A tap has no pointer-leave, so it needs one; but `click` and MapKit's `single-tap` fire for a mouse press too, and arming it there tore the hint away mid-hover. Where the device can hover, leave governs instead.

## Accessibility

A hover affordance is unreachable by keyboard and invisible to a screen reader, so `RoutePrivacyDescription` states the same thing outright whenever the route has hidden segments, and the chip is `aria-hidden` so the sentence is not announced twice. That also retires the focus drop [flagged in review earlier](https://github.com/llun/activities.next/pull/1367#discussion_r3689807950) — there is no longer a button to unmount out from under the user. The GL map container also picks up the `role="img"` + `aria-label` its MapKit sibling already had.

## Screenshots

Verified in Chromium against a local SQLite instance with a seeded GPX activity and a 1000 m privacy location over the route's start, so the trimmed ends really render green. (Screenshots live outside the repo — `docs/` is for durable reference documentation only — so the observed sequence is recorded here.)

| Step | Result |
| --- | --- |
| Page load | No overlay anywhere on the map — the whole route is visible |
| Pointer onto the green segment | Chip reads "Hidden from other viewers", anchored just above the pointer; cursor becomes `help` |
| Pointer away | Chip gone |
| Screen reader | The description is present whenever the route has hidden segments, with no hover required |

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (748 files / 8014 tests)
- [x] Docs updated: grepped `*.md` and `docs/` — nothing documents the notice or the storage key, so nothing went stale
- [x] If migrations changed: n/a, no migration in this PR

## Tests

- `mapGeometry.test.ts` — the distance helpers: latitude/longitude anisotropy, symmetry, a point on / beside / past the end of a segment, visible segments ignored even when nearer, a hidden stretch reduced to one sample, and the zero-length-segment guard (a repeated GPS fix would otherwise divide 0/0 and silently defeat every tolerance comparison).
- `RoutePrivacyHint.test.tsx` — the placement math at every edge, the hover-capability probe, and the chip's `pointer-events-none` / `aria-hidden` contracts.
- `FitnessStatusDetail.test.tsx` — a GL map stub that actually runs its `load` callback and records every layer-scoped handler, so the hover, the leave, the self-retiring tap, the mouse-click case that must *not* retire, the invisible hit layer's paint, the container's label, and the screen-reader description are all asserted against real registrations. The stub asserts each layer exists at registration time, pinning the ordering requirement.
- `ActivityRouteMapKit.test.tsx` — the MapKit path end to end through the existing test double: hover on the line, hover away from it, hover on a route with nothing hidden, pointer-leave, tap hit, tap miss, tap retire on a touch-only device, no retire on a hover-capable one, the anchor math, clearing the hint when the route is replaced underneath it, and listener teardown on unmount.

The six tests covering the removed bar's dismissal are deleted along with it.

https://claude.ai/code/session_01PvyJTyAfSoSkH3xHKdVe8X


lib/components/fitness/routePrivacyNotice.ts
```diff
@@ -0,0 +1,55 @@
...
+// Dismissal is therefore persisted per browser, alongside the theme mode and the
+// announcement banner's collapse preference, rather than held in component
+// state.
+//
+// There is deliberately no server-side counterpart: this is a UI hint, its scope
+// is the device that saw it, and persisting it per account would need a schema
+// change and a write on every dismissal to record something a reader can restore
+// by clearing site data.
+export const ROUTE_PRIVACY_NOTICE_STORAGE_KEY =
+  'fitness:route-privacy-notice-dismissed'
```
llun: The key is device-global, but the notice it gates is strictly account-specific. `app/api/v1/fitness-files/[id]/route-data/route.ts` passes `includeHidden: isOwnerAccount`, so hidden segments — and therefore this notice — only ever reach the owner of the activity. On a browser shared by two accounts (family device, a second account on the same laptop), account A's dismissal silently suppresses the explanation for account B's *first ever* look at their own redacted route, which is exactly the moment the notice exists for.

Privacy locations are themselves per-actor (`getFitnessSettings({ actorId, serviceType: 'general' })`), so scoping the stored key by actor id is the semantically right unit, not just a workaround. `FitnessStatusDetail` already has `currentActor` in hand, so this costs one prop.

Two independent review passes flagged this one.

---

lib/components/fitness/routePrivacyNotice.ts
```diff
@@ -0,0 +1,55 @@
...
+// the next activity file, not in the next section, and not after a reload.
+// Dismissal is therefore persisted per browser, alongside the theme mode and the
+// announcement banner's collapse preference, rather than held in component
+// state.
+//
+// There is deliberately no server-side counterpart: this is a UI hint, its scope
+// is the device that saw it, and persisting it per account would need a schema
+// change and a write on every dismissal to record something a reader can restore
+// by clearing site data.
+export const ROUTE_PRIVACY_NOTICE_STORAGE_KEY =
```
llun: `fitness:route-privacy-notice-dismissed` reads like it covers every fitness route surface, but it is specific to the **activity detail** map. `RouteHeatmapMap` (the fitness heatmap page, `RegionHeatmapDetail`, the shared-heatmap page and the public embed) draws the same `isHiddenByPrivacy` distinction in **blue**, not green, and carries no notice at all today. If one is ever added there it must not be born pre-dismissed by this key, and its copy would differ ("Green segments…" is wrong there).

Suggest naming the key for the surface it belongs to.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1862,10 +1873,10 @@ const ActivityMapPanel: FC<{
               </button>
             </div>
           ) : null}
-          {hasHiddenPrivacySegments && !isPrivacyNoticeDismissed ? (
+          {hasHiddenPrivacySegments && isPrivacyNoticeDismissed === false ? (
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: The strict `=== false` is the whole point of the tri-state — `null` means "not resolved yet", and treating it as truthy-falsy would show the notice on the first commit to a viewer who already dismissed it. But nothing in the suite pins that down: replacing this with `!isPrivacyNoticeDismissed` leaves all 65 tests in this file passing. It reads like a lint-style oddity, which is how a later "simplification" undoes it.

Worth at least a test that seeds the storage key **before** the first mount, so the component is exercised reading a value it did not itself write — today every test only covers the dismiss-then-remount path.

---

lib/components/fitness/routePrivacyNotice.test.ts
```diff
@@ -0,0 +1,56 @@
+/** @vitest-environment jsdom */
+import { afterEach, describe, expect, it, vi } from 'vitest'
```
llun: `vitest.config.ts` sets `globals: true`, and only 3 of ~740 test files import these names explicitly — none of them in this directory (`routeHighlightMarker.test.ts`, `mapGeometry.test.ts` both rely on the globals). Harmless, but it makes the new file look unlike its siblings.

Also: the single flat `describe('routePrivacyNotice')` names the module rather than the functions under test. The closest precedent, `lib/components/theme/theme-core.test.ts`, nests `describe('getStoredTheme')` / `describe('storeTheme')`, which is what AGENTS.md → Testing Guidelines asks for ("name the function or method under test directly").

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -1862,10 +1873,10 @@ const ActivityMapPanel: FC<{
...
             <button
               type="button"
-              onClick={() => setDismissedNoticeFileId(fitnessFileId)}
+              onClick={dismissPrivacyNotice}
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Dismissing unmounts the button that was just activated and nothing moves focus, so a keyboard or screen-reader user is dropped to `<body>` and back to the top of the tab order (WCAG 2.4.3).

Noting it rather than filing it as a defect of this PR: the pre-change code unmounted the same button on the same tap, so this is not a regression — and the impact shrinks here, since the tap now happens once per browser rather than once per activity file.

---

llun (commented):
Round 1 of the sub-agent review loop (AGENTS.md → Code Review Loop). Three agents reviewed the whole diff in parallel: correctness/React behavior, project conventions and test quality, and product/a11y/security semantics.

**No correctness defects found.** SSR/`window` guards, the hydration story, the tri-state gating, the effect-ordering race the old code documented, and the removal of the `fitnessFileId` prop all check out; the two rewritten component tests were confirmed non-vacuous by reverting the component and watching them fail. CI is green across all four test shards, Build, Lint/Prettier, Schema Dump Sync and CodeQL.

Five comments below. The substantive one is the storage key's scope, which two agents reached independently.

One finding I am **not** filing: that a permanent, unrecoverable dismissal is the wrong shape for a privacy-related notice, and it should be a reversible toggle or re-shown when the owner's privacy-location set changes. That is a reasonable product argument, but "show once, and after discard stay discarded for that browser" is the requested behavior of this change, so it stays as-is. If the notice should instead be recoverable, that is a separate decision and I would rather take it as one.

---

lib/components/fitness/routePrivacyNotice.ts
llun: Fixed in 9861074. The key is now scoped by actor — `fitness:activity-route-privacy-notice-dismissed:<actorId>` via `getRoutePrivacyNoticeStorageKey`, with both helpers taking the id. `ActivityMapPanel` gained a `currentActorId` prop, passed from `currentActor?.id` at both call sites; the mount effect keys on it, so switching actors re-resolves the acknowledgement rather than inheriting the previous one.

Agreed on the reasoning: privacy locations are per-actor, so that is the unit the acknowledgement belongs to, not the device.

Two tests cover it — `still shows the privacy notice to an actor who has not dismissed it on this browser` (seeds a *different* actor's dismissal, asserts the notice still renders) and `does not report another actor's dismissal` at the helper level. An actor-less viewer falls back to the bare prefix, which is only a defined default: they never receive hidden segments to begin with.

---

lib/components/fitness/routePrivacyNotice.ts
llun: Fixed in 9861074. The prefix is now `fitness:activity-route-privacy-notice-dismissed`, and the module header says why: the heatmap surfaces draw the same distinction in blue and carry no notice today, so one added there must not inherit this acknowledgement and could not reuse the "green" copy either. A test pins the `activity-route` segment of the name so a later rename cannot quietly widen the scope.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Partly addressed in 9861074, with a correction to the premise.

Added the test you asked for — `never renders the privacy notice when this browser already stored the dismissal` seeds the key before the first mount, so the component is now exercised reading a value it did not itself write. That was a real coverage gap and it is worth having.

But it does **not** kill the mutation, and neither would any test: I re-ran your experiment after adding it and `!isPrivacyNoticeDismissed` still passes the whole file. The reason is that the `null` frame is unreachable here. `shouldRenderMapPanel` turns on from `mapAttachment` / `hasMapData` / `isRouteDataLoading`, so the panel always mounts *before* its route data arrives; the `[]`-deps effect has committed a boolean long before `hasHiddenPrivacySegments` can become true. So the strict check is defensive, not load-bearing — it never observes `null` alongside hidden segments today.

I kept `=== false` (it is what makes the tri-state safe if route data ever arrives with the panel's first commit — server-rendered or cached) and put the whole explanation in a comment above the JSX, including the fact that no test can catch its removal. That comment is the guard here; claiming a test that does not actually pin it would have been worse than saying so.

---

lib/components/fitness/routePrivacyNotice.test.ts
llun: Both fixed in 9861074. The explicit `vitest` import is gone (the file now relies on the `globals: true` config like its siblings), and the flat module-level `describe` is split into three named for the functions under test — `getRoutePrivacyNoticeStorageKey`, `isRoutePrivacyNoticeDismissed`, `dismissRoutePrivacyNotice` — matching `theme-core.test.ts` and AGENTS.md → Testing Guidelines.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Confirmed and not changed here. The pre-change code unmounted the same button on the same tap, so the focus drop predates this PR — and now that the tap happens once per actor per browser instead of once per activity file, the change strictly reduces how often anyone hits it.

Fixing it properly means giving the map region a focus target (neither the map container nor the static-preview fallback is focusable today), which is a change to the panel's focus model rather than to the dismissal, so it does not belong in this diff. Leaving it visible here rather than silently resolving it.

---

llun (commented):
Round 2 of the sub-agent review loop, over commit 9861074 (the round-1 fixes), since fixes can introduce new problems.

**Verified clean:**
- The effect's deps went from `[]` to `[currentActorId]`, which is exactly the shape of the race the original code's comments warned about. It does not reintroduce it: `currentActorId` is a primitive, and the only thing that changes it — the actor switcher — does a hard `window.location.reload()`, so the panel remounts rather than re-running the effect over live state. At runtime the deps are effectively `[]`.
- Both new component tests are load-bearing, confirmed by mutation: dropping the actor id from the helper calls fails the seeded-dismissal test, and making `getRoutePrivacyNoticeStorageKey` ignore its argument fails the other-actor test. `beforeEach` clearing `localStorage` keeps them order-independent.
- The bare-prefix fallback is unreachable in practice (`isOwnerAccount` requires an authenticated account) and cannot collide with a scoped key.
- The two panel mount points stay mutually exclusive; prettier, eslint, both suites (80 tests) clean.

**One finding, fixed in a361e4a (comment-only):** my own justification comment was wrong. It cited `includeHidden: isOwnerAccount` to argue the actor is "the right unit" — but that gate compares **account** ids, and the privacy locations applied are the file owner's, so the cited code does not support the claim. `ActorProfile` carries no account id, so the actor remains the finest scope a Client Component can key on without new plumbing, and it errs the harmless way (an actor key can only re-show the notice, never hide it from someone who should see it). But the consequence is real and is now written down instead of glossed: an account with several actors is told what green means once per actor rather than once. Fixing that properly means plumbing an account id to the client, which is a bigger change than this PR should carry.

Also re-verified in the browser after the key change: notice shown → dismissed → still gone after reload → shown again in a fresh profile, with the stored key `fitness:activity-route-privacy-notice-dismissed:<actorId>`.

CI was green on 9861074 across all shards, Build, Lint/Prettier, Schema Dump Sync and CodeQL; a361e4a is a comment-only change on top.

Stopping the loop here: round 2 surfaced one documentation defect and no code defects, and the fix for it cannot itself regress behaviour. The one thread left open is the pre-existing focus drop on dismiss, which I deliberately did not fix in this diff.

---

llun (commented):
Round 3 of the sub-agent review loop, over the hover redesign (d6193b6). Four reviewers — GL path, MapKit path + geometry, test quality, product/a11y/deletion completeness — raised 12 findings, each then handed to two independent skeptics instructed to refute it. Fixes are in 91d0d85.

**Two real defects, both fixed:**

1. **The chip was clipped away at the map's edges.** Three of the four reviewers reached this independently. It was anchored at the raw pointer pixel and shifted up by its own height, inside a panel that is `overflow-hidden` — so a hover near the top or a side rendered it truncated or entirely invisible. Not a corner case: the hidden stretches *are* the trimmed start and finish of the route, and `fitBounds(…, { padding: 28 })` parks those against the bounding box's rim, so the most likely hover was the one that failed. `getRoutePrivacyHintPlacement` now measures the chip and clamps it — centred above the pointer where there is room, flipped below when there is not, held inside both side edges, and pinned left rather than centred if it is ever wider than the map. These are the same guards `ChartHoverMarker` already applies in this file, which the component's own comment cites as its model; adopting its styling without its clamping was the mistake.

2. **The touch retire-timer armed on a desktop mouse press.** Layer-scoped `click` and MapKit's `single-tap` both fire for a mouse, so clicking a green segment tore the hint away four seconds later while the pointer was still on it. Where the device can hover, `mouseleave` / `pointerleave` governs and the timer is not armed at all.

**Coverage the reviewers proved was missing** — each of these could be deleted from the implementation with the suite still green: the zero-length-segment guard (a repeated GPS fix would divide 0/0 and silently defeat every tolerance comparison, so the hint would never appear on that stretch), clearing the hint when the route is replaced under it, MapKit's tap retire and its anchor math, the chip's `aria-hidden` contract, and the GL container's new `role="img"`. One assertion was outright vacuous — the nothing-hidden case asserted the hint's absence without ever driving a hover, so it passed for the wrong reason; it now asserts the filtered hit layer and the registered handler.

The remaining findings did not survive verification.

Local gate green: prettier → lint → build → test (748 files / 8014 tests). Re-verified in Chromium before this round: clean map on load, chip on hover over the green segment, gone on leave, screen-reader description present without any hover.

---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Retired by the redesign in d6193b6, not fixed as such: the button this was about no longer exists. The explanation is no longer a dismissible control at all — it is a hover/tap chip on the green line itself, so there is nothing to activate and nothing to unmount out from under the user's focus.

The accessibility need the button was incidentally serving is now met deliberately instead. `RoutePrivacyDescription` renders an `sr-only` paragraph stating the explanation whenever the route has hidden segments, with no hover or focus required, and the chip is `aria-hidden` so it does not announce the same sentence twice. The GL map container also picked up the `role="img"` + `aria-label` its MapKit sibling already had.

Resolving now — the WCAG 2.4.3 concern is no longer reachable on this surface.

---

PR #1368, opened 2026-07-31
fix: group the post action row to the left with only ⋯ pulled right

llun:
## Summary

The post action row was `justify-between`, so reply / boost / like / bookmark / react were spread evenly across the whole width of a post. On the 940px content column that put ~135px between every control — five buttons reading as five unrelated things instead of one action set.

The design system groups them instead ([`ui_kits/web/Post.jsx`](https://claude.ai/design/p/a61009c5-7207-4e18-8d2b-a0cace8d672b?file=ui_kits%2Fweb%2Findex.html), whose `PostMenu` carries `ml-auto`): one `gap-1` cluster at the post's left edge, with only the `⋯` menu pushed to the far right. The row still spans the full width, so `⋯` stays pinned to the post's right edge on every surface.

The auto margin is what does the work — flexbox gives positive free space to auto margins *before* `justify-content` applies, so `ml-auto` beats `justify-between` (the kit keeps an inert one; this row just drops a class that describes the opposite layout).

## Changes

- `Actions` drops `justify-between` and passes `ml-auto` to `PostMenu`; `PostMenu` merges a caller `className` into its root.
- The edit-history panel now anchors to the **action row** rather than to its own trigger — the trigger's wrapper drops `relative`, so `Actions`' `relative` root is the containing block, and the panel keeps `right-0`. Anchored to the trigger, its 25rem width started wherever the engagement counts pushed that trigger, and the post's card clipped the overhang (see the review threads). The row spans the post exactly, so `right-0` is flush with the post's right edge at any width.
- `FitnessStatusDetail` inherits the new grouping for free — it already renders the shared `<Actions>` (`fullBleed={false}`).
- Docs that carried the old rule (`AGENTS.md` → Status Posts & Actions, `CLAUDE.md`, `REVIEW.md`) now describe the left cluster and what the auto margin / full width each buy. Tests pin `gap-1` (the entire spacing mechanism now), `ml-auto` on the `⋯` wrapper, and the panel's containing block — none of which jsdom can observe.

## Verification

Local dev server (SQLite + mock user/posts), Chromium at 1280×900, 768×900 and 420×800.

Measured geometry of the action row on the home timeline (940px column, row spanning x 343→1217):

| | before | after |
| --- | --- | --- |
| reply / boost / like / bookmark / react | x = 343, 511, 680, 848, 1017 | x = 343, 379, 415, 451, 487 |
| `⋯` | x = 1185 (right edge) | x = 1185 (right edge) |

- Home timeline, status detail (`StatusBox`), and 420px wide where the row goes compact: bookmark and react move into `⋯`, the remaining three stay packed left, `⋯` still pinned right.
- Edit-history panel on an edited post at 768px: panel 335→735 against a row of 105→735, inside a clipping ancestor ending at 752.
- Reaction chips still sit full-bleed above the row.

Not changed: the per-button padding. The design kit's `ActionButton` is `px-3 py-2` / `gap-2` against this repo's `px-2 py-1` / `gap-1.5`, about 10px wider per button. That is a separate change to the shared button treatment (it also shifts the 400px compact threshold), so it is out of scope here.

## Review

Five sub-agent review rounds (cap 20), converging on a clean round 5 — findings and resolutions in the comments below. Gate green on each round: `prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (748 files, 8014 tests).


llun:
Round 1 (sub-agent review) — one finding, fixed in 4fa78e0a1.

**`lib/components/posts/actions/edit-history-button.tsx` — the 25rem edit-history panel was clipped at `md`-ish widths.** With the actions now packed at the post's left edge, the trigger sits 180–274px in from that edge depending on the engagement counts beside it, and the panel was anchored to the trigger. On the status detail page at a 768px viewport the clipping ancestor ends at x=752, while a panel opened from a row with 3-digit counts reached x=779 — the card's `overflow-hidden` cut off its right edge (worse inside the fitness activity card, which is ~40px narrower again). Flipping back to `right-0` on the trigger was not an answer either: it put the panel's left edge off the post entirely.

Fix: anchor the panel to the **action row** rather than the trigger. The trigger's wrapper drops `relative`, so `Actions`' own `relative` root becomes the containing block, and the panel keeps `right-0`. The row spans the post exactly, so the panel is flush with the post's right edge at any width and whatever the counts do. Verified in a browser at 768px with an edited post: panel 335→735 against a row of 105→735, inside a clip ending at 752.

`post.test.tsx` pins all three invisible parts (`absolute right-0` on the panel, no `relative` on the trigger wrapper, `relative` on the row). Docs updated to match.

llun:
Round 2 (sub-agent review) — two findings, both addressed in c6321741d.

**1. `gap-1` was pinned nowhere (valid).** With no `justify-content` distributing the actions, `gap-1` is now the entire spacing mechanism between them, and deleting it from `actions.tsx` left the whole suite green with every button touching its neighbour. Now asserted on both surfaces (`post.test.tsx`, `FitnessStatusDetail.test.tsx`), replacing the `not.toHaveClass('justify-between')` assertions — which pinned a class that is inert either way, see below.

**2. "The design system uses `justify-between`, so the docs contradict it" (valid observation, wrong mechanism).** Two reviewers independently read `ui_kits/web/Post.jsx`, found `justify-between` on its row alongside the `ml-auto` on `PostMenu`, and concluded the kit renders a spread row and the `ml-auto` is inert. It is the other way round: flexbox distributes positive free space to **auto margins before `justify-content` sees it** ([CSS Flexbox §8.1](https://www.w3.org/TR/css-flexbox-1/#auto-margins)), so the auto margin wins and the kit renders the same left-packed row this PR ships — which is exactly what the design-system screenshot in the issue shows.

Measured it rather than arguing it. Adding `justify-between` back to the shipped row, live, with `ml-auto` present:

```
without justify-between: [105, 141, 177, 213, 249, 297, 703]
with    justify-between: [105, 141, 177, 213, 249, 297, 703]
```

Identical for all seven children. So no behavioural change is needed — but the docs were imprecise in implying the kit has no `justify-between` and that the class is forbidden. AGENTS.md / CLAUDE.md / REVIEW.md now say what actually carries the layout: the auto margin does the work, the kit's leftover `justify-between` is inert, and this row simply drops a class that describes the opposite layout.

Also noted by a reviewer and **not** changed here, as it predates this PR: `app/(timeline)/[actor]/[status]/page.tsx:335-338` has a comment reading "No `overflow-hidden`: this card contains posts…" directly above a class list containing `overflow-hidden`. That is the very ancestor behind the round-1 clipping, so the comment is actively misleading, but fixing it is out of scope for this PR.

llun:
Rounds 3–5 (sub-agent review).

**Round 3** — three docs and two test comments were justifying the layout with a mechanism that no longer exists. `CLAUDE.md`/`REVIEW.md` had been reworded to say dropping the `ml-auto` spreads the actions; it does not, now that the row carries no `justify-content` — it collapses `⋯` into the cluster, and only a `justify-between` added back *without* the margin spreads them (REVIEW.md managed both halves of that contradiction in one sentence). All three files also still justified the full-width rule with the spacing between actions, which was load-bearing under `justify-between` and is a flat width-independent `gap-1` now; what the full width buys today is `⋯` landing on the post's right edge and the edit-history panel's `right-0` having that edge to line up with. Fixed in e8b5c94bf.

**Round 4** — two more of the same, missed by round 3: the fitness test still credited the cluster + `ml-auto` for the inter-action spacing (a file that stubs `PostMenu` without a `className`, so no `ml-auto` renders in it at all), and `reaction-button.test.tsx` claimed a leftover wrapper's 4px is dead space "regardless" — with nothing after it the `ml-auto` absorbs the gap and the cluster is pixel-identical, which is precisely why it has to be asserted rather than looked at. Fixed in de768c9a6.

**Round 5** — clean. No new findings; a reviewer independently re-fetched `ui_kits/web/Post.jsx` through DesignSync and confirmed both claims the docs now make about the kit.

Loop converged at 5 rounds (cap 20). Gate green on every round: prettier → lint (0 errors) → build → test (748 files / 8014 tests).

PR #1369, opened 2026-08-01
fix: stop the status detail card clipping post overlays

llun:
## The contradiction

The conversation card on the status detail page carried `overflow-hidden` **directly beneath a comment saying it did not**:

```tsx
// No `overflow-hidden`: this card contains posts, and a post's
// non-portaled overlays (the reaction picker) would be clipped by it —
// the same reason `Posts` dropped it.
'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
```

`git log -L` shows the comment arrived in #1330 (the reaction-picker PR) as an addition to a line that already had the class — stale intent, never applied. This predates #1368, but that ancestor is exactly what clipped the edit-history panel there, so anyone debugging a clipped overlay would read the comment and rule out the element actually doing the clipping.

## Which side was right

**The conclusion, not the reason.**

The reason was false. `ReactionPicker` `createPortal`s to `document.body`, pinned by an assertion in `reaction-button.test.tsx`. The ⋯ menu's popover is a Radix portal too. Neither can be reached by an ancestor's overflow.

The conclusion was still correct, for a different overlay. `Posts` did drop the class, citing "edit-history panel, inline error bubbles" — and `EditHistoryButton` renders `absolute bottom-full`, ~360px tall (a 2.5rem header over a `max-h-80` list), opening **upward**. The only viewer who gets an action row at all is a signed-in one, whose focused post sits just below the back-button bar — so on a short post with a few edits the panel runs past the card's top edge.

## Changes

- **`page.tsx`** — drop `overflow-hidden` from the card; rewrite the comment to name the one overlay that actually reaches this edge, and to record explicitly what is *not* a reason (the picker and the menu popover portal; `PostMenu`'s non-portalled tooltip hangs mid-card).
- Children meeting the rounded corners now round themselves: the header wrapper when signed in, the first ancestor row or the focused post when logged out. That branch leads with an out-of-flow `sr-only` heading, so a `first:` selector would target the wrong element.
- **Sticky header.** Removing the clip also removed the scroll container pinning `Header`'s `sticky top-0`, which would have started detaching mid-scroll for the first time — carrying two transparent corner notches out over the posts, since `background` and `backdrop-filter` both clip to the radius. The clip therefore moves onto a **wrapper around the header** rather than being deleted outright: that subtree holds no overlays, so clipping costs nothing, and it restores the scrollport exactly. **This PR makes no sticky-behavior change.**
- **`edit-history-button.tsx` / `reaction-button.test.tsx`** — both claimed "every card that wraps a post clips it", which this change falsifies. Corrected.
- **`page.layout.test.tsx`** (new) — pins the class off and pins which child carries the radius in each branch.

## Verification

Measured, not assumed. Header top minus card top at `scrollY 900`, with `main` given the same `overflow-x-clip` as the real layout:

| variant | offset | |
|---|---|---|
| `overflow-hidden` on the card (main) | 1px | inert |
| radius on the header, no clip | 422px | detached, corner notches |
| clip on a header wrapper (adopted) | 1px | identical to main |

Corner rendering checked at 6× zoom: the fix is pixel-identical to main, and a control with the top child left square shows the white nub that appears otherwise — so the rounding additions are load-bearing.

Tests are mutation-checked. Four independent mutations are each caught: dropping either `!currentActorProfile` guard, dropping `previouses.length === 0`, and dropping `overflow-hidden` from the header wrapper.

Gates: prettier ✅ · lint ✅ (0 errors) · build ✅ · test ✅ 749 files / 8019 tests.

## Known-unfixed, on a different surface

The **fitness dashboard** card is *not* fixed here, and my earlier claim that it "doesn't need this change" was wrong. Review found the clipper there is not the outer card but an inner one at `FitnessStatusDetail.tsx:2752`, which wraps both `<Actions>` and its `EditHistoryButton`. Two overlays are clipped **today**:

- **Action-row errors are effectively invisible** — `absolute right-0 top-full mt-1` on a row that is the last child of the footer strip, so the box starts ~10px above the card's bottom border. A failed ⋯ → Delete on a fitness post shows the user nothing.
- **Edit history is cut off on desktop** — the ~360px panel opens upward into a ~230px body. `max-md:fixed` spares mobile.

Both are pre-existing and live in a different ~2900-line component with its own large test file, so they are tracked separately rather than folded in here.

### DIFF
--- a/app/(timeline)/[actor]/[status]/Header.tsx
+++ b/app/(timeline)/[actor]/[status]/Header.tsx
@@ -5,31 +5,16 @@ import { useRouter } from 'next/navigation'
 import { FC } from 'react'
 
 import { Button } from '@/lib/components/ui/button'
-import { cn } from '@/lib/utils'
 
 interface Props {
   isFitnessDashboard?: boolean
-  /**
-   * Extra classes for the header bar. The conversation card no longer clips its
-   * children (post overlays have to escape it), so that card rounds this — its
-   * first child — itself.
-   */
-  className?: string
 }
 
-export const Header: FC<Props> = ({
-  isFitnessDashboard = false,
-  className
-}) => {
+export const Header: FC<Props> = ({ isFitnessDashboard = false }) => {
   const router = useRouter()
 
   return (
-    <div
-      className={cn(
-        'sticky top-0 z-10 flex items-center gap-3 border-b bg-background/90 px-5 py-3 backdrop-blur',
-        className
-      )}
-    >
+    <div className="sticky top-0 z-10 flex items-center gap-3 border-b bg-background/90 px-5 py-3 backdrop-blur">
       <Button
         variant="ghost"
         size="icon"
--- a/app/(timeline)/[actor]/[status]/Header.tsx
+++ b/app/(timeline)/[actor]/[status]/Header.tsx
@@ -5,16 +5,31 @@ import { useRouter } from 'next/navigation'
 import { FC } from 'react'
 
 import { Button } from '@/lib/components/ui/button'
+import { cn } from '@/lib/utils'
 
 interface Props {
   isFitnessDashboard?: boolean
+  /**
+   * Extra classes for the header bar. The conversation card no longer clips its
+   * children (post overlays have to escape it), so that card rounds this — its
+   * first child — itself.
+   */
+  className?: string
 }
 
-export const Header: FC<Props> = ({ isFitnessDashboard = false }) => {
+export const Header: FC<Props> = ({
+  isFitnessDashboard = false,
+  className
+}) => {
   const router = useRouter()
 
   return (
-    <div className="sticky top-0 z-10 flex items-center gap-3 border-b bg-background/90 px-5 py-3 backdrop-blur">
+    <div
+      className={cn(
+        'sticky top-0 z-10 flex items-center gap-3 border-b bg-background/90 px-5 py-3 backdrop-blur',
+        className
+      )}
+    >
       <Button
         variant="ghost"
         size="icon"
--- a/app/(timeline)/[actor]/[status]/page.layout.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.layout.test.tsx
@@ -36,7 +36,10 @@ vi.mock('@/lib/database', async () => ({
   getDatabase: vi.fn(() => ({
     getStatus: mockGetStatus,
     getStatusReplies: mockGetStatusReplies,
-    getAcceptedOrRequestedFollow: mockGetAcceptedOrRequestedFollow
+    getAcceptedOrRequestedFollow: mockGetAcceptedOrRequestedFollow,
+    // Without this the page's settings read throws and every test logs an
+    // error while quietly exercising the env/default fallback path.
+    getAllServerSettings: vi.fn(async () => [])
   }))
 }))
 
@@ -151,17 +154,23 @@ describe('Conversation card chrome', () => {
     })
   })
 
-  // The card wraps posts, and a post's non-portalled overlays have to escape
-  // it: the edit-history panel opens upward from the action row and is taller
-  // than the space above the focused post, and the action-row error tooltips
-  // hang below their row. `Posts` dropped `overflow-hidden` for the same
+  // The card wraps posts, and the edit-history panel — which is not portalled
+  // — opens upward out of it. `Posts` dropped `overflow-hidden` for the same
   // reason. Clipping here was what a comment on this card wrongly claimed was
-  // already gone, so pin the class off rather than trusting the comment.
+  // already gone, so pin it off rather than trusting the comment.
   it('does not clip its children, so post overlays can escape it', async () => {
     const card = await renderPage()
 
     expect(card).toHaveClass('rounded-2xl')
-    expect(card).not.toHaveClass('overflow-hidden')
+    // Reject every clip, not just `overflow-hidden`: `overflow-x-hidden` would
+    // force the computed `overflow-y` to `auto`, re-clipping the panel
+    // vertically while an assertion naming only `overflow-hidden` passed.
+    // `overflow-x-clip` is the one form that does not clip the other axis, but
+    // it is also not needed here, so the simplest guard is to allow none.
+    const clipping = Array.from(card.classList).filter((name) =>
+      name.startsWith('overflow-')
+    )
+    expect(clipping).toEqual([])
   })
 
   // Nothing clips for the rounded corners any more, so the child that meets
@@ -213,6 +222,31 @@ describe('Conversation card chrome', () => {
     // The ancestor chain now sits above the focused post, so it takes the
     // corners and the post must not keep them.
     expect(rowFor('parent')).toHaveClass('rounded-t-2xl')
+    expect(rowFor('focused')).not.toHaveClass('rounded-t-2xl')
+  })
+
+  // With a single ancestor, `index === 0` and `index === previouses.length - 1`
+  // are the same row, so a chain of two is what distinguishes the topmost
+  // ancestor from the one nearest the focused post. The chain runs up to three.
+  it('rounds only the topmost ancestor when the chain is longer than one', async () => {
+    const focused = buildNote({ id: 'focused', reply: 'parent' })
+    mockResolveStatusFromPath.mockResolvedValue({
+      status: focused,
+      statusId: 'focused',
+      fullStatusId: focused.url,
+      isStatusHash: true
+    })
+    mockGetStatus.mockImplementation(
+      async ({ statusId }: { statusId: string }) =>
+        statusId === 'parent'
+          ? buildNote({ id: 'parent', reply: 'grandparent' })
+          : buildNote({ id: 'grandparent' })
+    )
+
+    await renderPage()
+
+    expect(rowFor('grandparent')).toHaveClass('rounded-t-2xl')
+    expect(rowFor('parent')).not.toHaveClass('rounded-t-2xl')
     expect(rowFor('focused')).not.toHaveClass('rounded-t-2xl')
   })
 
--- a/app/(timeline)/[actor]/[status]/page.layout.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.layout.test.tsx
@@ -154,23 +154,41 @@ describe('Conversation card chrome', () => {
     })
   })
 
+  // Collects every clip between the card and a post, inclusive of both. The
+  // panel opens upward *out of the post*, so a clip anywhere on that path cuts
+  // it off — checking only the card left "move the class down one level" as a
+  // green mutation. Matches on the prefix rather than `overflow-hidden` alone
+  // because `overflow-x-hidden` forces the computed `overflow-y` to `auto`,
+  // which re-clips the panel vertically. `overflow-visible`/`overflow-x-clip`
+  // would be rejected too; neither is needed here, so allowing none is the
+  // simplest honest guard.
+  const clipsBetween = (card: HTMLElement, statusId: string) => {
+    const found: string[] = []
+    let node: HTMLElement | null = screen.getByTestId(`status-${statusId}`)
+    while (node && node !== card.parentElement) {
+      found.push(
+        ...Array.from(node.classList).filter((name) =>
+          name.startsWith('overflow-')
+        )
+      )
+      node = node.parentElement
+    }
+    return found
+  }
+
   // The card wraps posts, and the edit-history panel — which is not portalled
   // — opens upward out of it. `Posts` dropped `overflow-hidden` for the same
   // reason. Clipping here was what a comment on this card wrongly claimed was
   // already gone, so pin it off rather than trusting the comment.
   it('does not clip its children, so post overlays can escape it', async () => {
+    // Signed in on purpose: a logged-out viewer gets no action row at all, so
+    // there is no overlay to clip and the assertion would be vacuous.
+    mockGetActorFromSession.mockResolvedValue(buildViewer())
+
     const card = await renderPage()
 
     expect(card).toHaveClass('rounded-2xl')
-    // Reject every clip, not just `overflow-hidden`: `overflow-x-hidden` would
-    // force the computed `overflow-y` to `auto`, re-clipping the panel
-    // vertically while an assertion naming only `overflow-hidden` passed.
-    // `overflow-x-clip` is the one form that does not clip the other axis, but
-    // it is also not needed here, so the simplest guard is to allow none.
-    const clipping = Array.from(card.classList).filter((name) =>
-      name.startsWith('overflow-')
-    )
-    expect(clipping).toEqual([])
+    expect(clipsBetween(card, 'focused')).toEqual([])
   })
 
   // Nothing clips for the rounded corners any more, so the child that meets
--- a/app/(timeline)/[actor]/[status]/page.layout.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.layout.test.tsx
@@ -2,7 +2,7 @@
  * @vitest-environment jsdom
  */
 import '@testing-library/jest-dom'
-import { render } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 
 import { getServerAuthSession } from '@/lib/services/auth/getSession'
 import { Actor } from '@/lib/types/domain/actor'
@@ -166,12 +166,27 @@ describe('Conversation card chrome', () => {
 
   // Nothing clips for the rounded corners any more, so the child that meets
   // them has to round itself or its square background bleeds past the border.
-  it('rounds the header, its topmost child, for a signed-in viewer', async () => {
+  // Exactly one child may do so — a second would notch a rounded row into the
+  // middle of the card.
+  // `StatusBox` is mocked to a bare testid div, so the row that carries the
+  // radius is its parent. Anchoring on the status rather than a child index
+  // keeps the assertion meaningful if the card gains another child.
+  const rowFor = (statusId: string) =>
+    screen.getByTestId(`status-${statusId}`).parentElement
+
+  it('rounds the header wrapper, its topmost child, for a signed-in viewer', async () => {
     mockGetActorFromSession.mockResolvedValue(buildViewer())
 
     const card = await renderPage()
 
+    // The wrapper, not the header itself: `Header` is `sticky top-0`, so it
+    // has to keep a clipping ancestor or it starts detaching mid-scroll and
+    // carries the corners out over the posts.
     expect(card.firstElementChild).toHaveClass('rounded-t-2xl')
+    expect(card.firstElementChild).toHaveClass('overflow-hidden')
+    // The header takes the corners, so the post below it must not — even
+    // though it is the topmost *post* and would take them when logged out.
+    expect(rowFor('focused')).not.toHaveClass('rounded-t-2xl')
   })
 
   it('rounds the focused post instead when logged out, which has no header', async () => {
@@ -180,7 +195,7 @@ describe('Conversation card chrome', () => {
     // The logged-out branch leads with an `sr-only` heading, which is out of
     // flow and paints nothing — the post below it is what meets the corners.
     expect(card.firstElementChild).toHaveClass('sr-only')
-    expect(card.children[1]).toHaveClass('rounded-t-2xl')
+    expect(rowFor('focused')).toHaveClass('rounded-t-2xl')
   })
 
   it('rounds the first ancestor row when logged out and the post is a reply', async () => {
@@ -193,14 +208,33 @@ describe('Conversation card chrome', () => {
     })
     mockGetStatus.mockResolvedValue(buildNote({ id: 'parent' }))
 
-    const card = await renderPage()
+    await renderPage()
 
     // The ancestor chain now sits above the focused post, so it takes the
     // corners and the post must not keep them.
-    expect(card.children[1]).toHaveClass('rounded-t-2xl')
-    expect(card.children[1]).toContainElement(
-      card.querySelector('[data-testid="status-parent"]')
-    )
-    expect(card.children[2]).not.toHaveClass('rounded-t-2xl')
+    expect(rowFor('parent')).toHaveClass('rounded-t-2xl')
+    expect(rowFor('focused')).not.toHaveClass('rounded-t-2xl')
+  })
+
+  // Both rounding rules are guarded on the viewer being logged out, because a
+  // signed-in viewer gets the header above everything. Without this, deleting
+  // either guard leaves the suite green while shipping a rounded row notched
+  // into the middle of the card.
+  it('rounds neither the ancestor row nor the post for a signed-in viewer', async () => {
+    mockGetActorFromSession.mockResolvedValue(buildViewer())
+    const focused = buildNote({ id: 'focused', reply: 'parent' })
+    mockResolveStatusFromPath.mockResolvedValue({
+      status: focused,
+      statusId: 'focused',
+      fullStatusId: focused.url,
+      isStatusHash: true
+    })
+    mockGetStatus.mockResolvedValue(buildNote({ id: 'parent' }))
+
+    const card = await renderPage()
+
+    expect(card.firstElementChild).toHaveClass('rounded-t-2xl')
+    expect(rowFor('parent')).not.toHaveClass('rounded-t-2xl')
+    expect(rowFor('focused')).not.toHaveClass('rounded-t-2xl')
   })
 })
--- a/app/(timeline)/[actor]/[status]/page.layout.test.tsx
+++ b/app/(timeline)/[actor]/[status]/page.layout.test.tsx
@@ -1,0 +1,206 @@
+/**
+ * @vitest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { render } from '@testing-library/react'
+
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+import { Actor } from '@/lib/types/domain/actor'
+import { Status } from '@/lib/types/domain/status'
+import { ACTIVITY_STREAM_PUBLIC } from '@/lib/utils/activitystream'
+import { getActorFromSession } from '@/lib/utils/getActorFromSession'
+
+import Page from './page'
+import { resolveStatusFromPath } from './resolveStatusFromPath'
+
+vi.mock('next/navigation', async () => ({
+  notFound: vi.fn(() => {
+    throw new Error('NEXT_NOT_FOUND')
+  }),
+  useRouter: vi.fn(() => ({ back: vi.fn(), push: vi.fn(), refresh: vi.fn() }))
+}))
+
+vi.mock('@/lib/config', async () => ({
+  getConfig: vi.fn(() => ({
+    host: 'activities.local',
+    fitnessStorage: undefined,
+    mediaStorage: undefined
+  }))
+}))
+
+const mockGetStatus = vi.fn()
+const mockGetStatusReplies = vi.fn()
+const mockGetAcceptedOrRequestedFollow = vi.fn()
+
+vi.mock('@/lib/database', async () => ({
+  getDatabase: vi.fn(() => ({
+    getStatus: mockGetStatus,
+    getStatusReplies: mockGetStatusReplies,
+    getAcceptedOrRequestedFollow: mockGetAcceptedOrRequestedFollow
+  }))
+}))
+
+vi.mock('@/lib/services/auth/getSession', async () => ({
+  getServerAuthSession: vi.fn()
+}))
+
+vi.mock('@/lib/services/queue', async () => ({
+  getQueue: vi.fn()
+}))
+
+vi.mock('@/lib/utils/getActorFromSession', async () => ({
+  getActorFromSession: vi.fn()
+}))
+
+vi.mock('@/lib/config/mapProvider', async () => ({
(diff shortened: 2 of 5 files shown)


app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
         // to the signed-in card to keep the spacing consistent across both.
         currentActorProfile && 'mt-4',
         // No `overflow-hidden`: this card contains posts, and a post's
-        // non-portaled overlays (the reaction picker) would be clipped by it —
-        // the same reason `Posts` dropped it.
-        'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
+        // non-portalled overlays would be clipped by it — the same reason
+        // `Posts` dropped it. The edit-history panel opens upward from the
+        // action row (`bottom-full`, up to 25rem tall) and the focused post
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx]: **`25rem` is the panel's WIDTH, not its height.**

`edit-history-button.tsx:64` sets `w-[25rem]`. The panel's *height* is its header (~2.5rem) plus a `max-h-80` (20rem) scrolling `<ul>` — about 22.5rem / ~360px. `25rem` appears nowhere in its vertical sizing.

This is a fabricated measurement in the one comment this PR exists to make trustworthy, and it now disagrees with the sibling comment at `edit-history-button.tsx:60-63`, which uses `25rem` correctly as a width. The same wrong figure is in the commit message body.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
...
+        // sits close enough to this card's top edge for it to run past; the
+        // action-row error tooltips hang below their row. The reaction picker
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx]: **The error-tooltip clause is not a reason for *this* card.** (Found independently by two reviewers.)

`post.tsx` gates the action row on `showActions && currentActor`, and `StatusBox.tsx:108` passes `showActions={variant === 'detail'}` — so within this card only the **focused** post has an action row. That post's wrapper is always followed by `SignInCallout` and/or the replies block / "No replies yet", so a `top-full mt-1` tooltip lands mid-card, hundreds of pixels above the bottom edge. Horizontally they're `right-0`/`left-0` inside the article's 16px padding.

The direction claim is true; the causal claim isn't. Someone who re-adds `overflow-hidden` and tests the tooltip will find no breakage and conclude the whole comment is unreliable — exactly the failure mode this PR is fixing.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
...
+        // and the ⋯ menu are *not* why — both portal to the document body, so
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx]: **"the ⋯ menu … portal[s] to the document body" is true of the popover, not the component.**

`DropdownMenuContent` does portal (`ui/dropdown-menu.tsx:40`), but `PostMenu` also renders its own **non-portalled** error tooltip at `post-menu.tsx:290-297` (`absolute right-0 top-full`) directly in the row. The comment tells the next reader the ⋯ menu is wholly portal-safe when part of it is an ordinary absolutely-positioned child.

Narrowing it to "the ⋯ menu's popover" fixes it. The reaction-picker half is fully verified — `reaction-picker.tsx:329-332` portals to `document.body`, pinned by `reaction-button.test.tsx:242-261`.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
...
+        // no ancestor's overflow can reach them. The children that meet the
+        // rounded corners round themselves instead.
+        'rounded-2xl border bg-background/80 shadow-sm'
       )}
     >
       {currentActorProfile ? (
-        <Header isFitnessDashboard={false} />
+        <Header isFitnessDashboard={false} className="rounded-t-2xl" />
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx]: **Removing `overflow-hidden` silently switches this header from inert to actually sticky — an unmentioned, untested behavior change.**

`Header` has always carried `sticky top-0 z-10`, but an ancestor with `overflow:hidden` is a scroll container, so the header's sticky viewport was the card itself, which never scrolls. The class was dead. Removing it makes the document the nearest scrollport and the header starts sticking for the first time. (`main` uses `overflow-x-clip` at `app/(timeline)/layout.tsx:122`, and `clip` does *not* create a scroll container, so it doesn't re-suppress it.)

Measured in a minimal repro:
- clipped card, `scrollTop 900`: `card.top = -700`, `header.top = -699` → scrolled away with the card, sticky inert.
- unclipped card, `scrollTop 2600`: `card.top = -355.5`, `header.top = 0` → pinned to the viewport.

**And it makes the `rounded-t-2xl` on this line wrong.** Once the header detaches it carries its 16px top corners into the middle of the card. `background` and `backdrop-filter` are both clipped to the radius, so while the bar floats over posts (`z-10`) there are two ~16px triangular cut-outs at its top corners showing crisp, unblurred post content through an otherwise 90%-opaque blurred bar.

Secondary: the fitness card at `page.tsx:288` keeps `overflow-hidden`, so the *same component* now sticks on a normal post and stays inert on a fitness post, decided by an unrelated class on an ancestor.

app/(timeline)/[actor]/[status]/Header.tsx
```diff
@@ -5,16 +5,31 @@ import { useRouter } from 'next/navigation'
 import { FC } from 'react'
 
 import { Button } from '@/lib/components/ui/button'
+import { cn } from '@/lib/utils'
 
 interface Props {
   isFitnessDashboard?: boolean
+  /**
+   * Extra classes for the header bar. The conversation card no longer clips its
+   * children (post overlays have to escape it), so that card rounds this — its
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/Header.tsx]: **This prop doc is false at one of `Header`'s two call sites.**

`Header` is rendered twice in `page.tsx`: the conversation card (line 348, now unclipped) and the **fitness** card (line 292), whose wrapper at line 288 still carries `overflow-hidden` and passes no `className`. The doc states unconditionally that "The conversation card no longer clips its children … so that card rounds this — its first child — itself."

A reader of `Header.tsx` can't tell that half its usage contradicts this, and it documents the *caller's* situation rather than the prop's contract, so it goes stale the moment either card changes.

app/(timeline)/[actor]/[status]/page.layout.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+  })
+
+  // Nothing clips for the rounded corners any more, so the child that meets
+  // them has to round itself or its square background bleeds past the border.
+  it('rounds the header, its topmost child, for a signed-in viewer', async () => {
+    mockGetActorFromSession.mockResolvedValue(buildViewer())
+
+    const card = await renderPage()
+
+    expect(card.firstElementChild).toHaveClass('rounded-t-2xl')
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.layout.test.tsx:212]: **Coverage gap: the signed-in + ancestor-chain combination isn't covered, and mutation testing proves it.**

Both new rounding rules are guarded by `!currentActorProfile`. Deleting that guard from either one leaves the whole suite green:
- drop `!currentActorProfile &&` at `page.tsx:362` → the first ancestor row rounds *while the signed-in header above it is also rounded*. Ran it: `Tests 4 passed`.
- drop `!currentActorProfile &&` at `page.tsx:380` → the focused post rounds under the signed-in header. Ran it: `Tests 4 passed`.

The regression that would ship is a rounded row notched into the middle of the card directly below a square-bottomed header — the exact class of bug this PR exists to prevent. One test (signed-in + a `parent` ancestor, asserting `children[1]` and `children[2]` are both **not** `rounded-t-2xl`) closes both.

app/(timeline)/[actor]/[status]/page.layout.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+  })
+
+  it('rounds the focused post instead when logged out, which has no header', async () => {
+    const card = await renderPage()
+
+    // The logged-out branch leads with an `sr-only` heading, which is out of
+    // flow and paints nothing — the post below it is what meets the corners.
+    expect(card.firstElementChild).toHaveClass('sr-only')
+    expect(card.children[1]).toHaveClass('rounded-t-2xl')
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.layout.test.tsx]: **Unanchored positional handle.**

Unlike line 201 — which anchors `children[1]` with `toContainElement(...status-parent)` — this never checks that `children[1]` is actually the focused post's wrapper. The file already mocks `StatusBox` to `data-testid="status-{id}"`, so `screen.getByTestId('status-focused').parentElement` is available and would both self-document and fail loudly if an element were inserted above.

Same applies to the negative at line 204: an unanchored negative assertion is the shape that silently stops testing anything. It does bite today (verified), but anchoring it keeps it that way.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
...
-        'overflow-hidden rounded-2xl border bg-background/80 shadow-sm'
+        // non-portalled overlays would be clipped by it — the same reason
+        // `Posts` dropped it. The edit-history panel opens upward from the
+        // action row (`bottom-full`, up to 25rem tall) and the focused post
+        // sits close enough to this card's top edge for it to run past; the
+        // action-row error tooltips hang below their row. The reaction picker
+        // and the ⋯ menu are *not* why — both portal to the document body, so
+        // no ancestor's overflow can reach them. The children that meet the
+        // rounded corners round themselves instead.
+        'rounded-2xl border bg-background/80 shadow-sm'
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:359]: **Asymmetry worth pinning: `posts.tsx:129` applies `first:rounded-t-xl` *and* `last:rounded-b-xl` defensively; this card only compensates the top.**

That is correct *today* — traced to the end of the component, the last child is always either the replies wrapper (no background) or the "No replies yet" div (no background/border), and the replies bottom out in `<article className="p-4 transition-colors">` with no background or hover fill. `SignInCallout`'s `bg-primary/5` is never last.

But it's safe only as long as no background-painting child is ever appended last — e.g. moving `SignInCallout` below the replies would break it silently, with nothing to catch it. Worth a note in the comment, since the whole premise of this PR is that undocumented assumptions here rot.

app/(timeline)/[actor]/[status]/page.tsx
```diff
@@ -333,23 +333,34 @@ const Page: FC<Props> = async ({ params }) => {
...
-      {previouses.reverse().map((item) => (
+      {previouses.reverse().map((item, index) => (
         <div
           key={item.id}
-          className="border-b border-l-4 border-l-primary/20 bg-muted/30"
+          className={cn(
+            'border-b border-l-4 border-l-primary/20 bg-muted/30',
+            // A logged-out view renders no `Header`, so the first ancestor row
+            // is what meets the card's rounded top corners.
+            !currentActorProfile && index === 0 && 'rounded-t-2xl'
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/page.tsx:396]: **Cosmetic, logged-out-with-ancestors branch only:** this combines `border-l-4 border-l-primary/20` with the new `rounded-t-2xl`. With top-border-width 0, CSS tapers the 4px left accent to nothing around the 16px arc, so the accent bar starts as a curved wedge rather than squarely at the corner. Previously the parent's clip produced the same arc but the accent kept full width right up to it.

Low severity — flagging so it's a decision rather than an accident.

llun (commented):
Sub-agent review round 1 — three reviewers (layout correctness, project invariants/comment accuracy, test quality). 9 findings, all verified against the code; two were confirmed independently by two reviewers, and the sticky-header one was confirmed with measured `getBoundingClientRect` values in a browser.

The headline: **removing `overflow-hidden` also un-breaks the header's `sticky top-0`**, which nothing in this PR mentions. An `overflow:hidden` ancestor is a scroll container, so the header's scrollport was the card — which never scrolls — making the class dead. Details in the inline threads.

llun:
### Two further findings, on files outside this diff (so they can't be inline)

**10. `edit-history-button.tsx:62-63` and `reaction-button.test.tsx:253-254` both assert "every card that wraps a post clips it" — this PR finishes falsifying that.**

`Posts` already dropped `overflow-hidden` (`posts.tsx:107`), and this PR removes it from the conversation card too. Only the fitness card still clips. Per AGENTS.md → Documentation Maintenance, and given this PR's own premise, leaving those two comments is the identical defect the PR was opened to fix, one file over.

**11. The PR description's claim about the fitness card is wrong, and there is a real pre-existing bug there.**

I wrote that the fitness dashboard card "does not need this change" because its `<Actions>` sits far down the page. Two reviewers independently found that reasoning is beside the point: the clipper on that surface is not the outer card at `page.tsx:288` but an **inner** one at `FitnessStatusDetail.tsx:2752` (`overflow-hidden rounded-xl border bg-card shadow-sm`), which closes at `:2945` and contains both `<Actions>` (`:2932`) and its `EditHistoryButton` in the footer strip at `:2910`. So changing the outer card alone would fix nothing there — and the overlays are clipped **today**:

- **Action-row errors are effectively invisible.** `ActionButtonError` and `PostMenu`'s own error are `absolute right-0 top-full mt-1`, and the action row is the last child of the footer's `px-4 py-2.5`, so the box starts ~10px above the card's bottom border. *Owner opens their fitness activity, ⋯ → Delete post, the request fails → the error renders into the clipped region and the user sees nothing happen.*
- **Edit history is cut off on desktop.** The ~360px panel opens upward from that footer; the card body above it is roughly 230px, so the oldest revisions are sliced off. `max-md:fixed` means mobile is unaffected — desktop only.

Both are pre-existing and in a different component (~2900 lines, with its own large test file), so I'm **not** expanding this PR to cover them — but the PR description will be corrected to stop asserting the surface is fine, and this needs its own change.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: Fixed in 020a598c7. The comment now reads "~360px — a 2.5rem header over a `max-h-80` list" and no longer cites 25rem at all, so it can no longer be confused with the width in `edit-history-button.tsx:60-63`. The stale figure in the first commit message stands as history; this commit body states the corrected one.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: Agreed and removed in 020a598c7 — confirmed the trace independently before cutting it. The rationale is now scoped to the single overlay that actually reaches this edge, plus an explicit note that `PostMenu`'s non-portalled tooltip hangs `top-full` mid-card and so is *not* a reason. Overstating the case is what made the original comment untrustworthy, so narrowing it is the fix rather than a concession.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: Fixed in 020a598c7 — narrowed to "the ⋯ menu's *popover*", with a parenthetical that `PostMenu`'s own error tooltip is not portalled but hangs mid-card. Good catch: the component/popover distinction is exactly the kind of imprecision that turns a comment into a trap.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx]: This was the most valuable finding in the round — I had missed it entirely, and it invalidated my original fix.

Reproduced the measurement, then used it to choose between options. Three cards, `main` given the same `overflow-x-clip` as the real layout, at `scrollY 900` (header top minus card top):

| | offset | |
|---|---|---|
| A: `overflow-hidden` on the card (main) | **1px** | inert |
| B: radius on the header, no clip (what I shipped) | **422px** | detached, notched |
| C: clip on a wrapper around the header | **1px** | identical to main |

Took C in 020a598c7. The header's subtree holds no overlays, so clipping it costs nothing, and it restores the scrollport exactly — so this PR now makes no sticky-behavior change at all, and the fitness/normal asymmetry you raised as a secondary point does not arise either. Whether the header *should* stick is a real question, but it is a deliberate non-change here.

app/(timeline)/[actor]/[status]/Header.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/Header.tsx]: Resolved by deletion in 020a598c7. Since the radius now goes on a wrapper, `Header` needs no `className` at all — the file is reverted to `main`, so `git diff main` no longer touches it. Your underlying point stands generally: that doc described the caller's situation rather than the prop's contract, which is what made it go stale across two call sites.

app/(timeline)/[actor]/[status]/page.layout.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.layout.test.tsx:212]: Fixed in 020a598c7, and your mutation result was reproducible — but the first fix was still insufficient, which the mutation harness caught.

Adding a signed-in + ancestor test killed mutation 1 but **not** mutation 2: with an ancestor present, `previouses.length === 0` is false, so the focused post never rounds regardless of the guard. Needed a signed-in test *without* an ancestor too. Now four mutations are covered:

- drop `!currentActorProfile` from the ancestor guard → caught
- drop `!currentActorProfile` from the focused-post guard → caught
- drop `previouses.length === 0` → caught
- drop `overflow-hidden` from the header wrapper → caught

Baseline 5/5 green between each.

app/(timeline)/[actor]/[status]/page.layout.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.layout.test.tsx]: Fixed in 020a598c7. Added a `rowFor(statusId)` helper resolving `screen.getByTestId("status-{id}").parentElement`, and every positional `card.children[n]` is gone — including the negative assertions, which is where you noted an unanchored form quietly stops testing anything. The remaining index-free assertions are on `card.firstElementChild`, which is meaningful because being *first* is the property under test.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:359]: Addressed as documentation rather than code in 020a598c7. The comment now states that only the top corners need compensating, why the bottom pair does not today (last child is always the replies wrapper or the "No replies yet" block, neither painting a background), and what would break the assumption — appending a background-painting child last.

Deliberately did not mirror `posts.tsx`'s defensive `last:rounded-b-xl`: there is no single stable last child to hang it on here, and adding it to all candidates would be four speculative classes. Recording the invariant is the honest version, since the premise of this PR is that unwritten assumptions here rot.

app/(timeline)/[actor]/[status]/page.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/page.tsx:396]: Acknowledged, keeping as-is. Verified the taper is real: with top-border-width 0, the 4px accent wedges into the 16px arc. But it is logged-out + has-ancestors only, affects the topmost row alone, and every alternative is worse — squaring that corner reintroduces the bleed, and dropping the accent on the first row makes the chain inconsistent. Flagging it as a decision rather than an accident, which was your ask.

llun:
### Findings 10 and 11 addressed

**10 — stale "every card that wraps a post clips it" comments: fixed in 020a598c7.**

Both now say what is actually true: `Posts` and this card have dropped the clip so the edit-history panel can escape upward, the fitness header card still clips, and therefore neither the panel's horizontal anchoring nor the picker's portalling may depend on the ancestor. Correcting these was non-optional — leaving them would have been the same defect this PR exists to fix, one file over.

**11 — fitness card: PR description corrected, bug not fixed here.**

The description no longer claims that surface is fine. It now names the real clipper (`FitnessStatusDetail.tsx:2752`, not the outer card) and states plainly that the edit-history panel and the action-row errors are clipped there today. Not fixing it in this PR: it is a pre-existing bug in a different ~2900-line component with its own large test file, and folding it in would bury a comment-accuracy fix inside an unrelated layout change. Tracked separately.

llun:
## Review round 2 — two reviewers (audit of the round-1 fixes + an independent full pass)

Both were asked to attack the premise the round-1 redesign rests on. Both re-derived it empirically and independently, at different scroll offsets:

| card variant | header top − card top | sticky engaged? |
|---|---|---|
| `overflow-hidden` on the card (main) | 1px | no |
| no clip anywhere | 401px / 422px | **yes** |
| clip on a header wrapper (adopted) | 1px | no |

So the wrapper stays. One reviewer added the mechanism I'd left implicit: the reason the viewport becomes the next scrollport is that `app/(timeline)/layout.tsx` uses `overflow-x-**clip**`, and `clip` is precisely the value that does *not* create a scroll container. It also noted the wrapper is *more* robust than main — being exactly the header's height, the containing-block clamp pins the sticky independently of the scrollport.

### Fixed in de6f54e59

**1. The guard test pinned the class, not the behaviour.** Adding `overflow-x-hidden` to the card left the suite green while re-clipping the panel vertically — `overflow-x: hidden` forces the computed `overflow-y` to `auto`. The test this PR added to prevent exactly this bug would have waved it through. Now rejects every `overflow-*` class on the card.

**2. The ancestor rule was untested past one ancestor.** With a chain of one, `index === 0` and `index === previouses.length - 1` are the same row; mutating to the wrong end shipped green. The chain runs to three, so added a two-ancestor case.

**3. "the focused post sits just below the back-button bar"** held only for a non-reply — up to three ancestor rows push it down. Qualified, and named the case that actually clips.

**4. The wrapper comment said the header's behaviour was "unchanged"**, which let a reader infer it sticks *within* the card. It does not, and never has, at either call site. Now stated outright.

**5. Mocked `getAllServerSettings`** so the tests stop logging an error and silently exercising the settings fallback.

Seven mutations are now caught: both `!currentActorProfile` guards, `previouses.length === 0`, the chain index, `overflow-hidden` on the wrapper, and `overflow-hidden` / `overflow-x-hidden` re-added to the card.

### Declined, with reasoning

**Delete `sticky top-0 z-10` from `Header.tsx` instead of keeping it inert behind a wrapper.**

The observation is correct and well made: that `sticky` has never engaged at either call site, so the wrapper exists to keep a dead class dead — and shipping "an inert `sticky` preserved by a wrapper" is arguably the same family of lie-in-code this PR is about.

I'm not taking it, because deleting it is a *design* decision, not a cleanup. `sticky top-0 z-10 backdrop-blur` reads as someone deliberately building a sticky bar; the honest conclusion is that it has been broken since it was written, and the plausible fix is to make it work, not to erase the intent. Choosing between those is @llun's call, not something to settle inside a PR about a stale comment.

What I did take is the reviewer's fallback: the comment now says plainly that the header does not stick on either call site and never has, and that whether it should is a live question this change deliberately leaves open. That converts the lie into a documented fact, which is this PR's actual thesis.

**Scope note, also accepted:** the reviewer checked and AGENTS.md's Documentation Maintenance rule enumerates `AGENTS.md`/`README.md`/`docs/` etc. — it does **not** cover source comments, so it doesn't compel the `edit-history-button.tsx` / `reaction-button.test.tsx` edits. They stand on their own merits (both asserted a universal this diff falsifies, both comment-only), but I shouldn't have cited the docs rule as the authority, and I'm not citing it now.

### Carried forward for the squash

Commit `a3e2a63` body contains my original wrong claim that the fitness card's "action row sits far enough down the page that nothing needs to escape". The PR description now corrects this, but since the repo squash-merges and commit bodies often get carried into the squash message, **that sentence should not survive into `main`** — a PR about a lying comment should not land a lying commit message.

Gates: prettier ✅ · lint ✅ (0 errors) · build ✅ · test ✅ 749 files / 8020 tests.

llun:
## Review round 3 — one reviewer, tasked with judging whether the loop had converged

Two findings, both on-thesis. Fixed in `a2f7319ca`.

**1. An eighth green mutation: the guard only inspected the card node.**

Moving the clip one level down — onto the focused post's own row wrapper, which is the *direct* ancestor of the action row — clips the upward-opening panel harder than the card ever did, and left all six tests passing. Verified, then fixed: the guard now walks the whole path from the card down to the post and rejects any `overflow-*` on it.

While fixing it I found the guard had a second weakness the reviewer didn't flag: it ran **logged out**, and a logged-out viewer gets no action row at all (`showActions && currentActor`), so there was no overlay to clip and the assertion was vacuous exactly where the bug lives. It now signs the viewer in.

**2. A round-2 parenthetical overstated the reply case.** I'd written that up to three ancestor rows was "room enough". It isn't once the list hits its `max-h-80` cap — a one-ancestor reply on a post edited five or more times still overruns the top edge. Softened to say the chain buys room without settling it.

### Not adopted

The reviewer extended finding 1 to the **ancestor** row wrapper. Ancestors render as `variant="comment"`, so `showActions` is false and `Actions` returns `null` — no action row, therefore no overlay and nothing to clip. Confirmed by mutation: clipping an ancestor row leaves the suite green, and that is the correct outcome, not a coverage gap. Adding a test for it would pin a non-property.

### Mutation coverage — 9 verified, green baseline between each

| mutation | caught |
|---|---|
| `overflow-hidden` re-added to the card | ✅ |
| `overflow-x-hidden` added to the card | ✅ |
| `overflow-hidden` moved onto the focused post's row wrapper | ✅ |
| `overflow-hidden` dropped from the header wrapper | ✅ |
| `rounded-t-2xl` dropped from the header wrapper | ✅ |
| `!currentActorProfile` dropped from the ancestor guard | ✅ |
| `!currentActorProfile` dropped from the focused-post guard | ✅ |
| `previouses.length === 0` dropped | ✅ |
| chain index flipped to the wrong end (`length - 1`) | ✅ |
| *`overflow-hidden` on an ancestor row* | *green — correct, no overlay there* |

Gates: prettier ✅ · lint ✅ (0 errors) · build ✅ · test ✅ 749 files / 8020 tests.

Round 4 running to confirm a clean round before I stop.

llun:
## Rebased onto #1368 — and it independently confirms this PR's premise

#1368 merged onto `main` while this was open, and it rewrote the same action row: `justify-between` → a left-packed `gap-1` cluster with `ml-auto` on the ⋯ menu, plus the edit-history panel re-anchored from the trigger to the action row.

**That invalidated a comment this branch had written.** Round 1 edited `edit-history-button.tsx` to say the row "is now `justify-between`" — true when I branched, false once #1368 landed. Rebased and resolved in favour of main's text, correcting only the clause this change falsifies: main says "a post's card clips whatever hangs outside it", which is no longer universal now that `Posts` and this card don't clip, though the fitness header card still does.

**The part worth flagging:** #1368's own commit message records measuring *this exact ancestor* —

> Measured on the status detail page at a 768px viewport, the clipping ancestor ends at x=752 while a panel opened from a 3-digit-count row reached x=779 — the card's `overflow-hidden` cut off its right edge.

So the clipping was independently observed and worked around **horizontally**, by anchoring the panel to the row. The **upward** overrun is the half that anchoring cannot reach — `bottom-full` needs the ancestor gone — which is exactly what this PR removes. The card comment now names which axis it governs, so the two fixes read as complementary rather than redundant.

### Re-verified against the new base

- panel is still `absolute bottom-full … w-[25rem]` over a `max-h-80` list → the "~360px, opens upward" claims hold
- action row is now `relative flex items-center gap-1` with `ml-auto` on the menu → my card comment never cited `justify-between`, so it needed no change
- fitness inner card still clips at `FitnessStatusDetail.tsx:2752` → the known-unfixed section stands
- all **seven** mutations re-run against the rebased tree, all still caught, green baseline between each
- prettier ✅ · lint ✅ (0 errors) · build ✅ · test ✅ 749 files / 8020 tests

Force-pushed with `--force-with-lease` (remote was at the expected `a2f7319ca`).

### Loop status: converged — 4 rounds of a 20-round cap

Round 4 returned clean: no tenth green mutation after six further attempts, every comment clause re-derived from source, and `clipsBetween` verified terminating and non-vacuous. The only remaining item is not a code change and is repeated below.

**At merge:** the squash message must be authored by hand. Commit `9dda6fd37`'s body still carries my original retracted claim that the fitness card's "action row sits far enough down the page that nothing needs to escape" (false — the clipper is the inner card), and describes the mechanism as `first:rounded-t-xl`, which is neither the approach nor the radius that shipped. GitHub pre-fills the squash body from commit messages, and a PR about a lying comment should not land a lying commit message on `main`.

llun:
### Correction to the record — the squash message carries one uncorrected false claim

Merged as `7946221ec`; the landed content is byte-identical to the reviewed branch. But the squash took the default concatenation of all five commit bodies, so the **first** commit's retracted statements went to `main` verbatim. Two of them are self-correcting in the same message, because the later review commits explicitly retract them in chronological order:

- "up to 25rem tall" → round 1's entry states it was the panel's **width** and gives the real ~360px.
- "`first:rounded-t-xl`" → rounds 1–3 describe the wrapper mechanism that actually shipped.

**One is not retracted anywhere in the message:**

> The fitness dashboard card keeps its clipping — its action row sits far enough down the page that nothing needs to escape.

That is false and was retracted in the PR description, which the squash body does not include. For anyone arriving from `git log`:

The clipper on the fitness surface is **not** the outer card, so its action row's distance down the page is irrelevant. It is an **inner** card at `FitnessStatusDetail.tsx:2752` (`overflow-hidden rounded-xl border bg-card shadow-sm`) that directly wraps `<Actions>` and its `EditHistoryButton`. Two overlays are clipped there **today**, both pre-existing and neither fixed by this PR:

- **Action-row errors are effectively invisible** — `absolute right-0 top-full mt-1` on a row that is the last child of the footer strip, so the box starts ~10px above the card's bottom border. A failed ⋯ → Delete on a fitness post shows the user nothing.
- **Edit history is cut off on desktop** — the ~360px panel opens upward into a ~230px body. `max-md:fixed` spares mobile.

Follow-up work on those two is already underway in a separate branch. No history rewrite here — `main` is shared, and a correction note is the right instrument.

PR #1393, opened 2026-08-08
feat: add UUIDv7 public ids for statuses and actors (accept-side)

llun:
## Summary

- Adds a nullable, uniquely-indexed `publicId` (UUIDv7) column to `statuses` and `actors`, backfilled from `createdAt` so lexicographic order == chronological order.
- Every status/actor insert now mints a `publicId`; new local status URIs use the publicId as their tail (`.../statuses/<uuidv7>`), so one id per newly-created status.
- Adds an accept-old resolver (`resolveStatusIdParam`/`resolveActorIdParam`, plus batched variants) used across the Mastodon API surface — every route that used to decode only the legacy colon/`apurl_` id form now also accepts a UUIDv7 publicId or a raw URI, falling back to the legacy decode. Pagination cursors accept publicIds the same way, resolved once at the API boundary so no query or join changed.
- AP content-negotiation routes resolve publicId tails via a single shared `database.getActorStatusFromPathSegment`, including the actor-scoping check that stops one actor's publicId resolving under another's URL space.
- **Zero client-visible change**: nothing is emitted differently yet. This PR is purely accept-side — the emission flip is the follow-up PR.

### Bugs fixed along the way

- `in_reply_to_id` on `POST /api/v1/statuses` was never decoded — a client sending the legacy colon form for a reply silently created a top-level post instead. Fixed, including the scheduled-status publish path.
- Admin-accounts pagination cursors (`GET /api/v1/admin/accounts`, `/api/v2/admin/accounts`) were emitted in colon form but never decoded anywhere, so pagination past page 1 had never worked.

## Deployment

**Order matters.** The runtime image ships no Knex CLI, so `yarn migrate` runs from a checkout while the previous build is still serving:

1. **`yarn migrate`** — adds the nullable columns + unique indexes, then backfills every existing row in batches (500/loop, transaction-free, resumable). The migration fails loudly if the backfill is genuinely stuck, so a re-run resumes; it deliberately does *not* fail when it converges and a concurrent insert lands afterwards, since failing a healthy online migration is worse than the residue.
2. **Roll out the new build.** Until the rollout completes, the old build keeps inserting rows with `publicId = NULL` — the migration cannot catch those by construction.
3. **`./scripts/maintenance/backfillPublicIds.ts`** — the post-rollout repair. Idempotent, safe against a live database, and its **exit code is the deploy gate**: 0 only when no NULL-publicId rows remain. See `docs/maintenance.md`.
4. Only once step 3 exits 0 should the follow-up emission PR ship, since it emits ids that must resolve.

Both reference schema dumps are regenerated in this PR.

## Test plan

- [x] `yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test` green (**757 files / 8250 tests**)
- [x] Migration: correct v7 backfill with `createdAt`-derived timestamps, idempotent, crash-recoverable (column/index added as independently idempotent steps), self-healing sweep for rows written behind the cursor, rollback clean
- [x] Accept-old resolver: raw URI / colon-form / `apurl_` / publicId / unknown-publicId all resolve as specified; single and batch resolvers proven element-by-element equivalent
- [x] Batch id routes resolve in a single query (not N), with explicit caps on previously-unbounded arrays
- [x] Cross-cutting integration tests: create → GET by publicId; quote by publicId; reply by publicId *and* legacy colon form; timeline paging identical with both cursor forms with the Link header still emitting the legacy form; poll GET/vote by publicId
- [x] AP content negotiation: backfilled (legacy-tail) statuses resolve via publicId, scoped to the correct path actor, non-public statuses still 404; the HTML fallback redirect is verified by feeding its own emitted `Location` back through the real `resolveStatusFromPath`
- [x] Repair script verified by running it against throwaway local SQLite (dry-run → exit 1, run → exit 0, re-run → clean no-op), with a parity test proving it mints ids identically to the migration

## Review

Six rounds of multi-agent review with adversarial verification of every finding (findings per round: 1, 0, 12, 7, 3, 1 — converging). All 9 review threads resolved. Rounds 4–6 specifically re-reviewed the *fixes*, which is what caught three regressions introduced by earlier fixes in this PR — including a migration change that would have failed a healthy production deploy.


migrations/20260808000000_add_public_ids.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const mintPublicId = (createdAt) => {
+  const time = toDate(createdAt).getTime()
+  const msecs = Number.isFinite(time) && time >= 0 ? time : 0
+  return v7({ msecs })
+}
+
+const addPublicIdColumn = async (knex, tableName, indexName) => {
+  const hasColumn = await knex.schema.hasColumn(tableName, 'publicId')
```
llun, line 63: **Migration idempotency gap: a crash between ADD COLUMN and ADD CONSTRAINT permanently skips the unique index.**

`addPublicIdColumn`'s only idempotency guard is `hasColumn(tableName, 'publicId')`. But the single `alterTable` call below (adding the nullable column + the unique index/constraint) compiles to **two separate, non-atomic SQL statements** on both PG and SQLite — confirmed by tracing knex's `Runner.queryArray`, which only wraps a statement group in `client.transaction(...)` on the single-statement `statementsProducer` path; a plain multi-statement array (this one) executes each statement independently with no atomicity, and this migration's own `config = { transaction: false }` (required for the batched backfill loop below to commit progressively) means there's no outer per-migration transaction either.

If the process is interrupted (pod eviction, deploy timeout, DB failover, OOM) between the ADD COLUMN and ADD CONSTRAINT/UNIQUE INDEX statements, the column exists but the unique index never gets created. Re-running the migration then sees `hasColumn` return `true`, logs "already exists, skipping schema change", and returns — silently and permanently leaving `statuses_publicid_unique`/`actors_publicid_unique` missing, with the deployed schema drifting from the committed `migrations/schema.sql`/`schema.sqlite.sql` and no DB-level uniqueness guard on `publicId` going forward.

Suggested fix: check column and index existence independently, and (re)create the index if the column exists but the index doesn't. Will push a fix.

llun: Fixed in c5620bc0: split the column-add and unique-index-add into two independently-checked idempotent steps, reusing the `hasIndex` helper already established in `20260519000000_add_status_reply_hash.js` for cross-dialect index-existence checking. `down()` got the matching treatment so rolling back from a partially-applied state doesn't try to drop a constraint that was never created.

Added a regression test simulating exactly the described crash window (column present, index absent, one row already carrying a placeholder value) — confirms a re-run creates only the missing index without touching the already-set value, and that the unique constraint is then genuinely enforced.

app/api/v1/accounts/route.ts
llun, line 1: **N+1 query amplification on batch id routes (0 → N round-trips), worst on this unauthenticated endpoint.**

`resolveStatusIdParam`/`resolveActorIdParam` issue a point query per id. At array call sites that replaced a pure-sync `ids.map(idToUrl)`, a request that previously cost **zero** decode queries now costs **one per id**:

- **`app/api/v1/accounts/route.ts:72`** — explicitly unauthenticated ("Intentionally public (no OAuthGuard)"), and `encodedIds` is neither deduped nor capped. ~380 uuidv7-shaped ids fit in a default 16KB header, all dispatched concurrently against a pool of 10. Sibling `familiar_followers` deliberately dedupes and caps at 40; this route does neither.
- **`app/api/v1/statuses/route.ts:134`** — capped at 100 but `OptionalOAuthGuard`, so reachable anonymously; ironic given the route's own comment says it exists to batch into a single query.
- **`notifications/requests/{accept,dismiss}`, `lists/[id]/accounts` DELETE, `reports.status_ids`** — authenticated, but the Zod arrays carry no `.max()`, and the downstream DB calls are already `whereIn`-chunked, so these were O(1) before this change.

Fix: add batched `getStatusIdsByPublicIds`/`getActorIdsByPublicIds` (a single `whereIn`, mirroring the existing `getStatusPublicIds`) behind `resolveStatusIdParams`/`resolveActorIdParams`, use them at every array site, and dedupe + cap the accounts route.

lib/services/collections/addMembers.ts
llun, line 1: **Unresolvable publicId → `new URL()` TypeError → 500 *after* the membership row is committed.** (high)

`POST /api/v1/collections/{id}/items` with `{"account_id":"<well-formed uuidv7 not in actors.publicId>"}` — a deleted actor, a stale client-cached id, or a status publicId sent by mistake:

1. `resolveActorIdParam` matches `isPublicId`, the lookup returns null, and the resolver returns the input **unchanged** (per its documented contract).
2. `addCollectionMembers` does no existence validation and commits a `collection_members` row with `targetActorId` = the bare UUID.
3. Line 47 then calls `new URL(bareUuid)`, which throws `TypeError: Invalid URL`.

`OAuthGuard` does not wrap `handle()`, so this surfaces as an unhandled 500 — with a phantom membership row already written. Two independent verifiers reproduced it against a real seeded SQLite DB, confirming the row is committed before the throw.

Pre-PR this could not happen: `idToUrl(uuid)` yielded the parseable `https://<uuid>/`.

Fix: drop ids that did not resolve to an `http(s)` URI **before** calling `addCollectionMembers`, so no phantom row is written and no unparseable id can reach `new URL`.

app/api/v1/admin/reports/route.ts
llun, line 1: **An unknown publicId silently DROPS the admin report filter instead of matching nothing — the moderator sees the entire queue believing it is filtered.**

`GET /api/v1/admin/reports?account_id=<well-formed uuidv7 not in actors.publicId>`:

- `resolveActorIdOrNull` takes the `isPublicId` branch → `getActorIdByPublicId` returns `null` → `?? undefined` erases it
- `getAdminReports`' `if (accountId) query.where('actorId', accountId)` guard is then skipped **entirely**, returning every report in the instance

Pre-PR, `safeIdToUrl` returned the non-matching `https://<uuid>/`, which was applied as a filter and correctly produced an empty list. A verifier confirmed empirically: a probe asserting `[]` for that query **fails** on this branch and passes against main's behavior.

This is a moderation surface, so "filter silently became no filter" is worse than a wrong-but-visible error.

Fix: on an unresolvable publicId, fall back to the raw value — matching `resolveActorIdParam`'s own "return the input unchanged" contract — so it filters to nothing rather than collapsing to `undefined`. Same for `target_account_id`. Add a negative test.

migrations/20260808000000_add_public_ids.js
llun, line 1: **Forward-only keyset strands rows inserted behind the cursor during an online migration — permanently NULL, with no repair path.**

The documented production flow runs `yarn migrate` against the live database while the currently-deployed (pre-PR) app keeps serving. On a multi-million-row `statuses` table the backfill walks `lastId` upward for many minutes. Meanwhile the old app inserts new statuses whose **v4 UUID tails sort uniformly across the id space**, so roughly half of them land *behind* the cursor. Every remaining batch selects `where('id', '>', lastId)`, so those rows are never revisited.

`up()` then returns successfully, knex records the migration as applied, and re-running `yarn migrate` is a no-op — the affected rows keep `publicId = NULL` forever.

NULL is a supported state for emission fallback, so nothing breaks visibly. But the PR body's own deploy gate — "confirm no NULL-publicId rows remain before the emission PR" — becomes unsatisfiable, with no repair path short of hand-written production SQL.

Fix: after the keyset walk, add a terminating self-healing sweep predicated on `whereNull('publicId')` rather than on the cursor. Because the predicate is the condition being repaired (not a position), it converges regardless of insert order, and it makes a plain re-run repair any stragglers — exactly what the PR-2 gate needs.

app/api/users/[username]/statuses/[statusId]/route.ts
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/route.ts:1]: **HTML fallback redirect echoes the raw `statusId` param instead of the status actually resolved, producing an unresolvable URL.**

This applies to every status whose URI tail is *not* its publicId — i.e. all statuses created before this PR (legacy v4 tail, publicId supplied by the backfill) and every Strava fallback note (which deliberately keeps a deterministic sha256 tail while the DB layer mints an unrelated v7 publicId).

A browser GETs `/users/alice/statuses/<publicId>`: the new fallback (lines 39-52) resolves the status correctly and the readability gate passes, but with an HTML `Accept` header `negotiateActivityPubContentType` returns null, so the handler falls through to the redirect — which is built from the raw request param, yielding `/@alice/<publicId>`. The web detail path does not resolve publicIds in this PR (that is follow-up work), so the redirect lands on a 404.

Not pre-existing: on main the handler could only resolve via `getLocalStatusId`, so a found status always had `statusId == its URI tail` and this same line was always correct.

Fix: redirect to the resolved status' canonical URL (`status.url`), keeping the `status.id`-derived fallback for Announce rows, which store `url: null`.

app/api/v1/statuses/route.ts
llun, line 1: **Accept-side gap: `POST /api/v2/filters/:filter_id/statuses` was never swept.**

`app/api/v2/filters/[id]/statuses/route.ts:78` takes a client-supplied `status_id`, and `parseStatusCreateInput` returns it verbatim (validated only as `z.string().max(2048)`, no decode) straight into `database.addFilterStatus`. A publicId is therefore persisted as-is into `filter_statuses.statusId`, where `matchFilter` only ever compares against `status.id` and `urlToId(status.id)` — so a filter created that way silently never matches anything.

Noting honestly: this file is byte-identical to main, and no client can obtain a bare publicId until emission flips in the follow-up PR, so it cannot bite in production *today*. But "accept a publicId wherever a status id is accepted" is precisely this PR's mandate, and leaving it unswept means the gap ships silently and only surfaces much later, in a surface (content filtering) where the failure is invisible — the filter just quietly does nothing.

Fix: `await resolveStatusIdParam(database, statusId)` before persisting, matching every other status-id-accepting write route in the sweep.

app/api/v1/accounts/[id]/statuses/route.ts
Comment (llun) [inline app/api/v1/accounts/[id]/statuses/route.ts:1]: **Self-referential `Link` pagination headers interpolate the raw path param un-encoded, for the raw-URI id form the resolver newly accepts.**

The resolver deliberately passes a raw `http(s)://` id through unchanged, so `GET /api/v1/accounts/https%3A%2F%2Fllun.test%2Fusers%2Fuser1/statuses` now returns 200. On main, `idToUrl` mangled that input to `https:////llun.test/…`, which matched nothing → 404, so a Link header was never built for it.

`pathBase` is then built from the decoded param and interpolated raw (neither `buildPaginationLinkHeader` nor the local `buildLink` percent-encode), emitting:

```
Link: <https://host/api/v1/accounts/https://llun.test/users/user1/statuses?limit=1&max_id=…>; rel="next"
```

…which does not route back to this endpoint. A verifier confirmed by running the real handler and observing exactly that header.

Low severity — it requires the unusual raw-URI id form — but the fix is a single call.

Fix: `encodeURIComponent` the id when building a self-referential path, or build the link from a canonical re-encoding of the resolved actor id rather than echoing the client's raw path param.

app/api/users/[username]/statuses/[statusId]/likes/route.ts
```diff
@@ -21,10 +22,24 @@ export const GET = traceApiRoute(
...
+      if (uriFromPublicId) {
+        const candidate = await database.getStatus({
+          statusId: uriFromPublicId,
+          withReplies: false
+        })
+        // Serve only this path-actor's own statuses — a publicId belonging to
+        // another actor must not resolve under this username's URL space.
+        if (candidate?.actorId === actor.id) status = candidate
+      }
+    }
```
Comment (llun) [inline app/api/users/[username]/statuses/[statusId]/likes/route.ts]: Can you put this in the database layer instead, so it's not repeated in every places.

app/api/v1/accounts/route.ts
llun: Fixed in b0f25800. Added batched `getStatusIdsByPublicIds`/`getActorIdsByPublicIds` (single `whereIn`, chunked via the existing `getWhereInBatchSize` so SQLite's 999-bind ceiling cannot be hit) behind new `resolveStatusIdParams`/`resolveActorIdParams`, and swapped every array call site onto them — so each batch route is one lookup regardless of array length. Batch semantics are per-element identical to the single-id resolver (raw-URL passthrough, publicId lookup, legacy fallback), with input order, length and duplicates preserved; a parity test asserts that element-by-element.

Bounds added, all 100 to match the existing `MAX_BATCH_STATUSES`/`MAX_COLLECTION_ACCOUNT_IDS` ceilings: `MAX_BATCH_ACCOUNTS` (dedupe + truncate, following the `familiar_followers` precedent), `MAX_LIST_ACCOUNT_IDS`, `MAX_BULK_NOTIFICATION_REQUEST_IDS`, `MAX_REPORT_STATUS_IDS` (the latter three rejected via each route's existing 422). The lists cap is enforced twice on purpose — `parseAccountIds` also reads urlencoded/multipart/query sources that never reach the Zod schema, so a schema-only `.max()` was bypassable. The two byte-identical notification `BulkBody` definitions were also deduplicated into a shared module so the cap cannot drift.

Tests assert single-lookup behaviour (spying on the DB method and asserting call count), cap enforcement, and mixed/duplicate/empty arrays.

lib/services/collections/addMembers.ts
llun: Fixed in 9ab8b3be. Unresolvable ids are now filtered out in `addMembers.ts` **before** `addCollectionMembers`, so no phantom row is written and nothing unparseable reaches `new URL`. The filter uses an actual `new URL` + protocol check rather than a regex, since `https://` alone satisfies a regex but still throws.

On the all-unresolvable case I checked the legacy behaviour first, as suggested: the colon form has no "account not found" rejection to be consistent with — an id for an actor this instance has never seen still resolves to a URI and *is* added, because a collection may legitimately reference a remote actor that gets fetched later. So an unresolvable id is a documented no-op rather than a new error class, and the single-`account_id` spec form still surfaces it through the route's existing 422 (no row written → the existing `getCollectionItemByAccount` lookup misses). `resolveCollectionItem.ts` was checked and needs no change (it only does a DB lookup, no `new URL`); the lists route already 404s on unresolvable ids.

app/api/v1/admin/reports/route.ts
llun: Fixed in 0ea31ec8. `resolveActorIdOrNull` is now `resolveActorIdFilter` returning `string` rather than `string | null`, falling back to the raw input so an unresolvable id filters to **nothing** instead of dropping the filter. The `?? undefined` is gone from both call sites (`account_id` and `target_account_id`).

Worth noting the fix also closes the same hole for an undecodable *legacy* id — e.g. a malformed `apurl_` value previously dropped the filter too, so this was slightly broader than the publicId case that surfaced it. Added the negative test asserting an empty list for an unknown publicId; it was verified to fail against the pre-fix code.

migrations/20260808000000_add_public_ids.js
llun: Fixed in e9a8e85c. Kept the forward keyset walk as the bulk O(N) pass and added a terminating `sweepMissingPublicIds` predicated on `whereNull(publicId)` rather than on a cursor, run at the end of the backfill for both tables — so it converges regardless of insert order and a plain re-run repairs stragglers, which is what the deploy gate needs.

Two independent termination guards, both logging rather than spinning: a per-pass no-progress break (nothing in this migration ever clears `publicId`, so a pass that selects NULL rows and updates none means the UPDATE is not taking effect), and a hard `MAX_SWEEP_PASSES` bound. The sweep SELECT deliberately has no `ORDER BY` — the unique index added by this same migration indexes NULLs on both dialects, so `WHERE publicId IS NULL LIMIT 500` stays an index scan that an ORDER BY would turn into a sort.

On testing: your suggested test (insert behind the cursor after `up()`, then re-run) does **not** actually fail against the pre-fix code, because a fresh `up()` restarts the walk from `lastId = ""` and re-walks everything. It is included anyway since it pins the re-run repair path, but the real race is reproduced deterministically by a second test using a SQLite `AFTER UPDATE OF publicId` trigger that inserts a lower-sorting row the instant the last walked row is updated — i.e. mid-run, behind the cursor. Verified: with the sweep call commented out that test fails and the other seven pass; with it restored all eight pass. Schema dumps are untouched, as expected for a DML-only change.

app/api/users/[username]/statuses/[statusId]/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/route.ts:1]: Fixed in 326e2954. The HTML fallback now redirects to the resolved `status.url`, with an `https://${actor.domain}/@${actor.username}/${uriTail}` fallback for Announce rows (which store `url: null`). Also switched that fallback's host from `status.actor?.domain` to `actor.domain` so an absent actor cannot emit the literal string "undefined" in the `Location` header.

Sibling routes checked: `likes/`, `replies/` and `shares/` always answer as ActivityPub and have no HTML redirect at all, so no change was needed there.

app/api/v1/statuses/route.ts
llun: Fixed in 4989fd5c. `POST /api/v2/filters/:id/statuses` now resolves the client-supplied `status_id` with `resolveStatusIdParam` before persisting, matching every other status-id-accepting write route in the sweep. Existing rows holding legacy colon-form values are untouched and `matchFilter` still tolerates both stored forms, so nothing is migrated or broken. Added a test that a publicId-supplied filter status is stored resolved and actually matches.

app/api/v1/accounts/[id]/statuses/route.ts
Comment (llun) [reply app/api/v1/accounts/[id]/statuses/route.ts:1]: Fixed in b055f5bb. Self-referential `Link` paths are now percent-encoded, so the raw-URI id form the resolver accepts produces a well-formed URL that round-trips back to the same route. Applied consistently across the sibling account subroutes that build Link headers the same way (`followers`, `following`, `endorsements`, `media`, `collections`, `in_collections`), with a test asserting the emitted header parses and routes back.

app/api/users/[username]/statuses/[statusId]/likes/route.ts
Comment (llun) [reply app/api/users/[username]/statuses/[statusId]/likes/route.ts]: Good call — done in e6a5dbaf.

Extracted into a single database method, `getActorStatusFromPathSegment({ actorId, pathSegment, withReplies })`, declared on `StatusDatabase` and implemented in the status mixin. All four AP routes (`route.ts`, `likes`, `replies`, `shares`) now collapse to one call each, so this route went from ~16 lines to 5.

Named it to fit the two existing families it sits between — the `getStatusFrom*` resolvers (`getStatusFromUrl`, `getStatusFromUrlHash`, `getStatusFromPublicId`) and the actor-scoped `getActor*` methods — rather than the `getLocalStatus*` I first considered, since it scopes by `actorId` and does not actually check locality.

Semantics are unchanged, including the security-critical actor scoping (a publicId belonging to a different actor must never resolve under this actor's URL space) — that comment moved with the code. One consequence worth noting: the route-level cross-actor test necessarily became "the database resolves the segment to nothing → 404", since the decision now lives in the database layer, so the security property is asserted directly in `lib/database/sql/status.test.ts` instead — where the new test first proves the same publicId *does* resolve under its owning actor before asserting null under another, so the null is genuinely scoping rather than a missing row.

Gate green: 756 files / 8211 tests.

llun:
## Review round 4 — 7 findings, all fixed

Round 3 (11 partitioned reviewers) found 12 issues, fixed in the previous batch. Round 4 then reviewed **those fixes** and found 7 more — including **two regressions introduced by my own round-3 fixes**, which is exactly why the loop re-reviews fixes rather than stopping at green tests.

| # | Finding | Fix |
|---|---|---|
| 1 | **Regression from `4989fd5c`**: resolving the filter `status_id` before persisting also changed what the API *emits* — `status_id` and `status_matches` started returning raw AP URIs, breaking this PR's accept-only invariant and mixing two id forms in one filter list | `ed0d4951` — keep the write-side resolution, normalize at the emission boundary with `urlToId` in `getMastodonFilter` and `applyFilters`; added a parity test asserting a legacy row and a new row emit the same form |
| 2 | **Regression from `326e2954`**: the HTML redirect still pointed somewhere unresolvable — `status.url` has a single-`@` actor segment (`/@test/123`) and `resolveStatusFromPath` requires `@user@domain`, so it 404'd; the test pinned the broken URL | `3a36feb0` — build the Location from `getStatusDetailPath`, the same helper the web UI links with; the test now feeds the emitted Location back through the real `resolveStatusFromPath` and asserts it resolves |
| 3 | Sweep hard-fails on NULL-`id` actors rows: knex compiles `.where("id", null)` to `WHERE id IS NULL`, matching every such row, so one publicId is assigned to all and the unique constraint aborts the migration | `edd130ee` — `.whereNotNull("id")` on the sweep, with skipped rows counted and reported |
| 4 | Both sweep exits let `up()` resolve, so knex records the migration and the logged remedy ("re-run the migration") is impossible — `migrate:latest` skips recorded migrations | `edd130ee` — re-count remaining addressable NULLs and **throw** if any remain, so the migration is genuinely resumable; corrected the false comments/log lines |
| 5 | The "re-run repairs stragglers" test passed with the sweep entirely removed — it guarded nothing (the forward walk reached the row on its own) | `edd130ee` — rewritten so it fails without the sweep, plus direct coverage for the no-progress break and the pass bound |
| 6 | Batch vs single resolver diverge on a case-insensitive collation (MySQL is a supported backend) | `3f9a3749` — normalize the lookup *parameter* on both paths. Note the originally-proposed map-key-only fix was proven insufficient: it diverges in the opposite direction with the same id in two casings in one array, no MySQL required |
| 7 | Repo-owner feedback: the AP publicId fallback was duplicated across four routes | `e6a5dbaf` — extracted to `database.getActorStatusFromPathSegment`; the four routes now make one call each |

Every fix was mutation-checked — the new tests were verified to fail with the corresponding production change reverted, which is what exposed finding 5.

**Gate:** prettier clean, lint 0 errors, build clean, full suite **756 files / 8229 tests** passing. Both PR-1 invariants verified: exactly the 6 documented follow-up emission-side joins remain, and `getClientStatusId`/`getClientActorId` are still unused (this PR emits nothing new).

PR #1396, opened 2026-08-09
minor: emit UUIDv7 public ids on the Mastodon API and web paths

llun:
Follow-up to #1393. That PR made every id-accepting surface accept a UUIDv7 `publicId` while still **emitting** the legacy colon/`apurl_` form. This PR flips emission: clients now receive publicIds, and web status URLs become `/@user@domain/<uuidv7>`.

The accept side from #1393 is what makes this safe — every legacy id form (colon, `apurl_`, sha256 web hashes, percent-encoded remote URLs) keeps working indefinitely, so cached and bookmarked ids do not break.

## What flips

- **Status serializer** — `id`, poll `id`, `in_reply_to_id`, `in_reply_to_account_id`, `quoted_status_id`, and mention ids (the last via one batched actor lookup per page).
- **Account serializer** — the single `getMastodonAccountFromSQLActor` emission point, so every embedded account across statuses, notifications, search, reactions and favourited/reblogged_by flips with it.
- **Entity serializers** — relationships, notification groups + envelope, notification requests, collections, admin accounts/reports, report entities.
- **Pagination cursors and Link headers** — via a shared `getClientStatusCursors` helper that answers from the page in memory and batches the rare off-page boundary into one lookup.
- **Web paths** — `getStatusDetailPath` / `getStatusDetailPathClient` prefer the publicId for local *and* remote statuses; `resolveStatusFromPath` gains a publicId branch with path-actor validation ahead of the existing sha256/percent-encoded/synthesized-URI chain, all of which keep working.

`uri` and `url` fields keep carrying ActivityPub URIs — they are not ids and must not flip. Federation is untouched.

## Also included

- **Carried forward from #1393's review**: `getStatusPublicIds`/`getActorPublicIds` issued a single unbounded `whereIn`. They had no production callers then; this PR is what starts feeding them full timeline pages, so they are now chunked like their forward counterparts, with a boundary test asserting the actual per-statement binding count.
- **Encoding-independent joins**: sites that reverse-decoded an emitted id back to a URI now join on `uri`/`url`. Note `getMastodonConversation` needed this even though the design notes said otherwise — that assumption only holds when an actor's `url` equals its `uri`, which is not true for actors serving `/@name`.
- **Filters** match across all three stored forms (raw URI, colon, publicId) and still emit a consistent id form.

## Deployment — read before merging

**This PR must not ship until #1393's backfill is complete in production**, because it emits ids that must resolve:

1. #1393 deployed and `yarn migrate` run.
2. New build fully rolled out.
3. `./scripts/maintenance/backfillPublicIds.ts` exits 0 — that exit code is the gate (it is 1 while any NULL-publicId row remains). Rows written by pre-publicId builds during the rollout are only repaired by this script.
4. Only then merge/deploy this PR.

Rolling-deploy safety: any row that still has a NULL publicId emits its legacy id via the `publicId ?? urlToId(id)` fallback, which is covered by tests at every flipped site.

## Test plan

- [x] `yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test` green (**759 files / 8307 tests**)
- [x] Every flipped emission keeps a legacy-fallback test (publicId null → colon form) so the rolling-deploy path stays covered
- [x] Every new test negative-controlled — verified to fail with its source change reverted
- [x] Request-construction tests deliberately left on legacy id forms: they are the permanent regression suite proving accept-old still works
- [ ] Browser verification (in progress — screenshots to follow)


llun:
## Review summary — multi-agent loop, 3 rounds

Same process used on #1393: reviewers each read a **bounded slice in full** (rather than skimming the whole diff), and every finding then went through **three different verification lenses** — reproduce it end to end / is it already handled / is it actually caused by this PR — with majority rule. Fix batches are themselves re-reviewed, which is what caught most of the real problems.

**Findings per round: 10 → 2 → (round 3 running).**

### Round 1 — 16 candidates, 10 confirmed

| Finding | Severity | Fix |
|---|---|---|
| **"Follow back" was broken** — `FollowNotification` passes `account.id` (now a publicId) into `follow()`, which ran `urlToId()` on it client-side, mangling it to `<uuid>:` that no resolver can decode | high | `6507e929` — treated as a class, not an instance: **all 39 client-side `urlToId` calls removed** from `lib/client.ts`, split by destination (path segment → new `toIdPathSegment`, query params and JSON bodies → no encoding at all, two `?id[]=` interpolations → `encodeURIComponent`) |
| **Filters never flipped** — `FilterStatus.status_id` and `status_matches` still emitted the colon form while `status.id` was a publicId, so one document carried two irreconcilable ids for the same post (found independently by 3 reviewers) | medium | `949b1c4f` — both emission points flipped, matching preserved across all three stored forms |
| **v2/search N+1** — `orderAccountsByIds` awaited a resolver per account in a sequential loop; post-flip each became a DB round trip on an abortable typeahead | low | `533b529f` — keyed off `account.uri`, zero awaits in the loop |
| **Notification N+1** — the single-status mention path issued one lookup per status, and notifications serialize one at a time (40 extra queries per page) | low | `533b529f` — both serializers routed through the batching path |
| Orphaned JSDoc + 3 stale id-contract comments | low | `533b529f`, `7a174277` |

### Round 2 — 2 candidates, both confirmed

| Finding | Severity | Fix |
|---|---|---|
| **The filters fix's agreement was coincidental, not structural** — the two emission points agreed only when `idToUrl(urlToId(id))` round-tripped losslessly. It does not for a remote status id with an uppercase host or an explicit `:443` (stored verbatim by `createNoteJob`), so one response could still name a status two ways | low | `4a73adda` — both points now read one hydrated row through a single `getEmittedFilterStatusId`, so they fall back **together**. Also narrowed `matchFilter`'s input from `Map<string,Status>` to `Set<string>`, removing the object it would need to reintroduce the bug |
| **The new notification join had zero coverage** — a verifier mutated the join key from `status.uri` to `status.id` and all 232 notification tests still passed, meaning v1 notifications shipping with no `status` would be invisible to CI | medium | `912a9834` — real seeded-DB coverage, explicitly mutation-verified (3 tests now fail under that mutation; previously it killed nothing) |

### What was checked and found clean

- **Federation untouched** — `git diff main...HEAD -- lib/activities lib/types/domain` is empty of behavior change; `uri`/`url` still carry ActivityPub URIs everywhere
- **Accept-old preserved** — request-construction tests deliberately still send legacy colon-form ids; those are the permanent regression suite for every client holding an old id, and no reviewer found one flipped
- **Rolling-deploy safety** — every flipped emission keeps the `publicId ?? urlToId(id)` fallback, each with its own test, since production carries NULL-publicId rows until the backfill script has run
- **Id-space coherence** — final audit classified all 26 remaining `urlToId(` call sites as deliberate fallback or accept-side decode; zero missed emissions
- **Browser-verified** end to end: new status URL is a UUIDv7, the same status still resolves at its legacy sha256 path, synthesized short-id and percent-encoded remote URLs resolve, and boost/like/react/reply all work

**Gate:** prettier clean, lint 0 errors, build clean, full suite **760 files / 8342 tests** passing.

llun:
## Round 3 — converged, 0 findings

Final round reviewed the round-2 fixes and did a whole-PR pass. **1 candidate, 0 confirmed** after three-lens verification (the one candidate — a coverage gap on the `NotificationRequest.id` emission — was refuted by majority: the id is taken directly off the already-tested `account.id`, so it cannot drift independently).

**Findings per round: 10 → 2 → 0.**

Reviewers verified this round by *mutation* rather than by reading:

- Reverting `applyFilters.ts` fails exactly the two new filter tests, and the failure diff is the disagreement itself (`status_matches` = uuid vs `status_id` = colon form) — so the regression tests genuinely pin the fix.
- Mutating the notification join key (`status.uri` → `status.id`) now fails exactly the three new assertions and nothing else. Before `912a9834` that mutation killed nothing.
- Probing the sibling join in `getNotificationGroupsEnvelope` kills 13 tests, confirming the v2 path was already well covered and only v1 was blind.
- The `Map` → `Set` narrowing in `getCandidateStatusIds` was proven behaviourally equivalent (the old `.get(a) ?? .get(b)` was truthy exactly when the key existed, since values were always `Status` objects), with all six candidate forms still matching.

One trade-off is recorded explicitly rather than hidden: when hydration misses on a status whose stored id is not a fixed point of `idToUrl(urlToId(id))` (a remote id with non-canonical host casing), `status_matches` now emits the colon form instead of the publicId, so it no longer string-equals the entity `id`. That is deliberate — both fields fall back **together**, so the payload stays self-consistent, and the emitted colon form is still accepted by `resolveStatusIdParam`. The alternative was round-1 behaviour, which shipped a document naming one status two different ways.

**Final state:** prettier clean, lint 0 errors, build clean, full suite **760 files / 8342 tests**, CI green. Working tree clean, all commits pushed.

Left unmerged for review, per request. **Do not deploy before** #1393's migration has run, the new build is fully rolled out, and `./scripts/maintenance/backfillPublicIds.ts` exits 0 — that exit code is the gate.

PR #1398, opened 2026-08-09
none: store full ActivityPub URI as status id in mock seeders

llun:
## Summary

The mock seeders stored a **bare uuid** as a local status `id`, with a comment calling it "the local convention". The app does not do that — `lib/actions/createNote.ts` stores the full ActivityPub URI (`https://<domain>/users/<username>/statuses/<publicId>`) as `statuses.id`, and `createLocalOnlyFitnessStatus` in `lib/jobs/importFitnessFilesJob.ts` does the same for fitness posts.

That made locally-seeded data behave unlike real data. `urlToId()` parses a bare uuid as a hostname and yields `<uuid>:`, which resolves to no status, so the status detail page's `favourited_by` lookup 404'd on every load of a seeded post. Statuses created through the real composer send `apurl_…` and return 200.

The call site is `StatusBox.tsx`, which passes the raw `actualStatus.id` into `getStatusFavouritedBy` → `urlToId()` — so it is the **stored id** that decides this, not the `publicId` column.

This is pre-existing and unrelated to the UUIDv7 publicId work (#1393 / #1396); it was surfaced while browser-verifying those.

## Changes

Both seeders now mirror their production counterparts:

- **`scripts/mock/createMockStatuses.ts`** — local statuses use `getLocalStatusId(...)` for the id (full AP URI whose tail is the client-facing publicId), the `@username` web form for the url, and pass `publicId` so URI tail, web url and stored publicId agree. Remote (external-actor) statuses keep their origin's canonical URI as both id and url and let the database mint their publicId, the way federation delivers them.
- **`scripts/mock/createMockFitnessData.ts`** — same shape, backdating the publicId to the activity start time so the URI tail sorts with the activity (matching `createLocalOnlyFitnessStatus`).

## Verification

`scripts/` is neither linted nor covered by CI, so this was verified by actually running the seeders against a **local throwaway SQLite** database per the "Local Manual / Browser Testing" section of `AGENTS.md`. Both the old and the fixed script were seeded into the same database so the two variants could be compared on the same page, signed in.

| Seeded by | `statuses.id` | request issued by `StatusLikes` | result |
| --- | --- | --- | --- |
| old script | `d86ff5dd-…` (bare uuid) | `/api/v1/statuses/d86ff5dd-…:/favourited_by?limit=5` | **404** |
| fixed script | `https://localhost:3000/users/test1/statuses/019fe59f-…` | `/api/v1/statuses/apurl_aHR0cHM6…/favourited_by?limit=5` | **200** |
| fixed fitness script | `https://localhost:3000/users/test1/statuses/019fe594-…` | `/api/v1/statuses/apurl_aHR0cHM6…/favourited_by?limit=5` | **200** |

The trailing colon in the 404 URL reproduces the reported symptom exactly.

## Notes

- No migration, no schema change, no runtime code touched — `scripts/mock/` only.
- Pre-existing bare-uuid rows in an existing local dev database keep 404ing; the fix only affects newly seeded rows.

## Test plan

- [x] `yarn run prettier --write .` — no drift
- [x] `yarn lint` — 0 errors
- [x] `yarn build` — passes
- [x] `yarn test` — 757 files, 8250 tests passed
- [x] Browser-verified against local SQLite as above

### DIFF
--- a/scripts/mock/createMockFitnessData.ts
+++ b/scripts/mock/createMockFitnessData.ts
@@ -14,7 +14,10 @@ import crypto from 'crypto'
 
 import { getConfig } from '@/lib/config'
 import { getDatabase, getKnex } from '@/lib/database'
+import { getMention } from '@/lib/types/domain/actor'
+import { getLocalStatusId } from '@/lib/utils/activitypubId'
 import { ACTIVITY_STREAM_PUBLIC } from '@/lib/utils/activitystream'
+import { generatePublicId } from '@/lib/utils/publicId'
 
 const DAY_MS = 24 * 60 * 60 * 1000
 
@@ -127,14 +130,22 @@ async function createMockFitnessData() {
     // "Recent activities" feed has content.
     let statusId: string | null = null
     if (index < 3) {
-      const id = crypto.randomUUID()
-      const url = `https://${domain}/users/${username}/statuses/${id}`
+      // Mirror `createLocalOnlyFitnessStatus` in
+      // `lib/jobs/importFitnessFilesJob.ts`: the id is the full ActivityPub URI
+      // whose tail is the client-facing publicId (backdated to the activity
+      // start so it sorts with it), and the url is the `@username` web form. A
+      // bare uuid here would make seeded posts behave unlike real ones —
+      // `urlToId` parses it as a hostname, so client calls such as
+      // `/api/v1/statuses/<id>/favourited_by` resolve to no status and 404.
+      const publicId = generatePublicId(startedAtMs)
+      const id = getLocalStatusId({ actorId: actor.id, statusId: publicId })
       await database.createNote({
         id,
+        publicId,
         actorId: actor.id,
         to: [ACTIVITY_STREAM_PUBLIC],
         cc: [actor.followersUrl],
-        url,
+        url: `https://${actor.domain}/${getMention(actor)}/${publicId}`,
         text: `<p>${template.text}</p>`,
         createdAt: startedAtMs
       })
--- a/scripts/mock/createMockStatuses.ts
+++ b/scripts/mock/createMockStatuses.ts
@@ -2,7 +2,10 @@
 import { getConfig } from '@/lib/config'
 import { getDatabase } from '@/lib/database'
 import { Timeline } from '@/lib/services/timelines/types'
+import { getMention } from '@/lib/types/domain/actor'
+import { getLocalStatusId } from '@/lib/utils/activitypubId'
 import { ACTIVITY_STREAM_PUBLIC } from '@/lib/utils/activitystream'
+import { generatePublicId } from '@/lib/utils/publicId'
 
 async function createMockStatuses() {
   const database = getDatabase()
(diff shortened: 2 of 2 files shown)


llun:
### Review round 1 — sub-agent findings

Three sub-agents reviewed the whole diff (behavioral fidelity vs. the production code paths, project invariants per `AGENTS.md`/`REVIEW.md`, and an adversarial runtime pass that actually ran both seeders against throwaway local SQLite databases).

**One actionable finding.**

---

#### 🔸 Finding 1 — commit/PR prefix should be `none:`, not `fix:`

`scripts/mock/` is dev-only tooling. Verified it ships nothing: the runtime stage of `Dockerfile` copies only `.next/standalone`, `public`, `.next/static`, `data.sqlite` and `sharp` — there is no `scripts/` in the image. It is also excluded from lint (`eslint.config.mjs` ignores `scripts/**`) and from the CI prettier gate.

Per the Version Bump Prefixes table in `AGENTS.md`, `none:` is for "internal-only changes that do not require a release" — the same reasoning `AGENTS.md` already applies to the schema dumps ("commit as `none:` … they ship nothing").

**Consequence:** merging with `fix:` cuts a patch version bump, an auto-merging version-bump PR, a tag, and a multi-arch Docker publish for a change that alters no shipped byte.

**Fix:** retitle the PR (squash-merge makes the title the commit subject) and amend the commit to `none:`.

---

#### Explicitly checked and NOT a problem

- **Fidelity to production** — the `id`/`url`/`publicId` triple matches `lib/actions/createNote.ts` and `createLocalOnlyFitnessStatus` in `lib/jobs/importFitnessFilesJob.ts` formula-for-formula.
- **Future-dated `generatePublicId(Date.now() + timeOffset)`** — harmless. `getPublicIdTimestamp` has no production call sites; pagination orders by the `createdAt`/`id` columns, never by decoding the uuid. The same `createdAt` feeds both `generatePublicId` and `createNote`, so they stay in lockstep.
- **`publicId` unique-index collisions** — `statuses_publicid_unique` exists, but the seeded timestamps are days apart and v7 mixes in random bits. Running each seeder twice produced no constraint failure; `origin/main` behaves identically, so re-runnability is unchanged.
- **Remote statuses passing `publicId: undefined`** — matches real remote ingestion (`lib/jobs/fetchRemoteStatusJob.ts` never passes one); `database.createNote` mints it.
- **Server/Client module boundary** — the three new imports pull in only `zod`/`pino`/`uuid`; no `'use client'` anywhere in the transitive graph, and both scripts run to completion under the `scripts/run.cjs` + `@swc-node/register` CJS bootstrap.
- **Import order / prettier** — `prettier --check` passes on both files despite `scripts/` being outside the CI gate.
- **Comment accuracy** — every factual claim in the added comments verified against the real code.
- **Docs** — nothing goes stale; `docs/maintenance.md` and `AGENTS.md` only show invocation commands, which are unchanged.
- **Referential integrity** — zero orphans across `attachments`, `fitness_files.statusId`, `timelines.statusId`, `recipients`; zero null `publicId`; all 14 local rows satisfy `id == actorId || '/statuses/' || publicId`.

#### Root-cause refinement

The reviewers pinned the emit path more precisely than the PR body does: `getClientStatusId` (`lib/utils/publicId.ts`) is currently **dead production code** (referenced only by its own test). The live serializer is `lib/services/mastodon/getMastodonStatus.ts` → `id: urlToId(status.id)`, and `StatusBox` likewise hands the raw `actualStatus.id` to `getStatusFavouritedBy`. So it is the **stored id** that decides this in every path today — which is exactly what this PR changes.

Measured before/after through that path, 2× each seeder:

| | statuses | attachments | fitness | timelines | orphans | `id == actorId+'/statuses/'+publicId` |
|---|---|---|---|---|---|---|
| pre-change | 18 | 44 | 80 | 12 | 0 | **0 / 14** |
| this PR | 18 | 44 | 80 | 12 | 0 | **14 / 14** |

#### Pre-existing, out of scope (not regressions — identical on `origin/main`)

- `createMockFitnessData` never calls `createTimelineStatus`, so its seeded posts don't land on the home timeline (the real `createLocalOnlyFitnessStatus` calls `addStatusToTimelines`).
- It also inserts into `fitness_files` with raw knex, so `CounterKey.totalFitness` is never incremented.
- `createMockFitnessData.ts` has never been listed in `docs/maintenance.md`.

llun:
**Finding 1 addressed** — commit amended and PR retitled to `none:` (`320c5c8c`).

Independently confirmed the reasoning before applying: the runtime stage of `Dockerfile` copies only `.next/standalone`, `public`, `.next/static`, `data.sqlite` and `sharp`, so `scripts/` genuinely never reaches the image and this change ships no byte. Squash-merge takes the PR title as the commit subject, so the retitle is what actually drives the workflow; the commit body was amended to match.

All other round-1 items were refuted with evidence and need no change. Resolved.

llun:
### Review round 2 — clean

Re-reviewed the fix itself plus the whole diff again (fixes can introduce new problems).

- Commit subject is now `none: …`; the message body is intact and the `Co-Authored-By` trailer survived the amend.
- The amend changed the message only — `git show --stat HEAD` is still exactly the two intended files (`scripts/mock/createMockStatuses.ts`, `scripts/mock/createMockFitnessData.ts`, +35/−8). No scope drift, no stray helper files committed, working tree clean.
- PR title matches the commit subject.
- Fresh re-read of the full diff against `lib/actions/createNote.ts` and `createLocalOnlyFitnessStatus` found nothing new; comments remain accurate to the code.

**No new actionable findings.** A clean round, so the review loop stops here at round 2 of a maximum 20.

PR #1402, opened 2026-08-11
minor: add gear tracking for fitness activities

llun:
Implements the **Gear tracking** spec from the design system (`ui_kits/web/gear.html` + `GearKit.jsx`), wired into the fitness section the way `ui_kits/web/index.html` shows.

Bikes and shoes, and the components fitted to a bike, with lifetime distance **derived from the activities they are used on rather than stored**.

## The model

A gear total is `SUM(fitness_files.totalDistanceMeters) WHERE gearId = ?`. A component total is that same sum restricted to activities inside its `[addedAt, removedAt)` install window — a null `addedAt` means "since the gear's beginning", a null `removedAt` means "still fitted". Nothing is cached, so a back-dated upload, an archive re-import, an activity edit and a delete all reconcile for free; a stored total would have to be reconciled on every one of them.

Both rollups reuse the same completed/primary/not-deleted predicate as `getFitnessActivitySummary`, so gear numbers agree with the fitness overview, and each answers a whole page in **one** grouped query. The component query keeps the install window in the JOIN condition so a component with no matching activity still returns a row and counts 0.

## Schema

New `fitness_gears` and `fitness_gear_components`, plus a plain indexed `fitness_files.gearId`. That column deliberately has **no database-level foreign key**: adding one through `alterTable` needs a table rebuild on SQLite, which no migration in this repo does. Ownership is enforced in `lib/database/sql/fitnessGear.ts`, and deleting gear soft-deletes it and nulls the column in the same transaction. Both reference schema dumps are regenerated.

## Surfaces

- **Fitness → Gear** — bikes and shoes with their derived distance, default sports, retired gear behind a toggle, and add/edit dialogs.
- **Gear detail** — stat tiles and the component service log. *Replace* closes the fitted part today and opens a fresh one at 0 km carrying the same service interval; components with a reminder show a wear bar that turns amber at 85% and red past the mark.
- **Activity detail** — an owner-only gear select, narrowed to the kind the activity's sport implies (unknown sport shows everything, so nothing is ever unassignable).
- **Post fitness chip** — the gear name follows the distance: `26.2 km · Moots`.

## Attribution

- **Default sports** assign gear automatically once a file finishes parsing. `activityType` is free-form across four vocabularies (FIT `cycling`, TCX `Biking`, Strava `GravelRide`, free-form GPX), so a new normalizer maps it onto canonical sport keys and returns null rather than guessing — a wrong mapping would silently attribute activities to the wrong bike.
- **Strava** imports carry their gear across, keyed on the Strava gear id rather than the name (names are editable on both sides, so name-keying would fork a duplicate on any rename). Archive imports have only the name and fall back to matching it against `bikes.csv` / `shoes.csv` / the `Activity Gear` column.
- Import jobs assign through a `whereNull('gearId')`-guarded update, so a re-run can never clobber a manual assignment.

## Service reminders

Compared against the shoes distance alert and each component's service interval, recording the distance they fired at so a crossing notifies exactly once and a raised threshold re-arms on its own. **Evaluated on write, not on a schedule** — the queue can delay a message but not repeat one, and this instance has no cron infrastructure. Adds a `gear_service_due` notification type and email template.

## API

All owner-scoped, answering `404` rather than `403` for anything the actor does not own.

| Route | Methods |
|---|---|
| `/api/v1/fitness/gear` | `GET` `POST` |
| `/api/v1/fitness/gear/:id` | `PATCH` `DELETE` |
| `/api/v1/fitness/gear/:id/retire` | `POST` `{ retired: boolean }` |
| `/api/v1/fitness/gear/:id/components` | `GET` `POST` |
| `/api/v1/fitness/gear/:id/components/:componentId` | `PATCH` `DELETE` |
| `/api/v1/fitness/gear/:id/components/:componentId/replace` | `POST` |
| `/api/v1/fitness-files/:id` | `PATCH` `{ gearId }` |

Two deliberate deviations from the handoff's route list: retire/unretire is one idempotent toggle rather than two verbs, and `replace` is its own endpoint because the two writes must be atomic and what the replacement inherits is a server-side rule.

## Verification

- `yarn run prettier --write .` → clean · `yarn lint` → 0 errors · `yarn build` → compiles, all gear routes registered · `yarn test` → **777 files / 8673 tests passing** (~330 new).
- The component window join was additionally run against a real **PostgreSQL 17**, since the shared pg test harness is broken for an unrelated reason (see below). Results match SQLite exactly, including the half-open boundary and the null-start-time case.
- Browser-verified end to end on local SQLite: created a bike, added a component with a 5,000 km reminder, replaced it (old part archived behind the toggle, fresh row inheriting the interval), assigned the bike to a ride from the activity page, and watched the gear list's derived distance go from `0.0 km` to `26.2 km` and the chip render `26.2 km · Moots Vamoots RSL disc`.
- Both email variants rendered through `scripts/mock/renderEmailPreviews.ts`.

## Notes for review

- **Zod ordering bug caught during development, worth knowing about:** `.nullish()` applied *before* a `.transform()` still runs the effect for a key absent from the body, yielding `null`. Since the update paths decide what to write from key presence, that would have made every `PATCH` clear the columns the caller never mentioned. `optionalText` now puts `.nullish()` last, and `gearRequests.test.ts` guards it.
- **Unverified assumptions about Strava's formats** (no real export or live account available), all documented in code comments and degrading to "no gear" rather than throwing: gear id prefixes (`b…` bike, `g…` shoes), `frame_type` being bike-only, the `Activity Gear` column name, and the gear-name column in `bikes.csv`/`shoes.csv` (matched as the first header containing "name", else the first column). Archive gear kind is keyed off *which file* a name came from, so a renamed header cannot mis-file a bike as shoes.
- **Deviation from the mock:** the design's tables pin the first column and snap-scroll below 480px. This uses the app's existing `overflow-x-auto` treatment instead, since the pinned-column pattern exists nowhere else in the codebase.
- **Pre-existing bug found, not fixed here** (out of scope, flagged separately): running the suite with `TEST_DATABASE_TYPE=pg` fails immediately because `applyPostgresSchema` loads a dump whose `set_config('search_path','',false)` is session-level, leaving the pooled connection unable to resolve unqualified table names. CI pins SQLite, which is why it went unnoticed.


llun:
## Sub-agent review round 1 — findings

Four reviewers audited the diff across disjoint slices (data layer + migration, API + validation, UI + integrations, jobs + Strava + reminders). Everything below was reproduced before acting on it; fixes are landing in a follow-up commit on this branch.

### Critical

**1. The service reminder could never fire for the activity that crosses the threshold.** `processFitnessFileJob` set `processingStatus = 'processing'`, evaluated reminders, and only wrote `'completed'` afterwards — but the rollups only count completed activities, so the evaluation read the total from *before* the ride that caused the crossing. Shoes at 640 km with a 650 km reminder + a 15 km run = no notification; the reminder was systematically one activity late, and never arrived at all if that was the last ride on the gear. Moved the evaluation after the `completed` write. Verified by mutation: restoring the old ordering fails the new test with `expected undefined to be 5200`.

**2. Every gear reminder in the notifications list read "This imported activity is no longer available."** Two reviewers found this independently. `gear_service_due` was registered as `kind: 'system'`, whose renderer is written for activity imports and treats a missing status as a *deleted* one — but a gear reminder carries no status by design. Added a `'plain'` notification kind for self-addressed notices; the row now shows the headline and links to `/fitness/gear`.

### Major

**3. Soft-deleting Strava-imported gear permanently broke re-attribution.** The unique index on `(actorId, stravaGearId)` covers soft-deleted rows while the lookup filters them out, so after a delete every future import of that bike hit the index, failed the recovery re-read, and silently left the activity unattributed — unrepairable, since hand-created gear has no `stravaGearId`. `deleteFitnessGear` now releases the Strava id in the same transaction.

**4. `gear_service_due` was missing from the notification-settings save schemas,** so the new toggle was stripped by `z.object`, the route returned 200, and the setting sprang back on reload — gear emails could never be turned off. Added to both request schemas and to `UpdateActorParams`.

**5. `mastodonTypeToInternal` was only wired one way.** Gear reminders serialize as Mastodon `status` but `exclude_types[]=status` did not exclude them (and `types[]=status` did not return them).

**6. `PATCH /api/v1/fitness-files/:id` with `gearId: ""` bypassed the ownership check** — `if (gearId)` is false for the empty string, so the "is this gear mine?" lookup was skipped and `''` was written. `assignFitnessFileGearIfUnset` guards on `whereNull('gearId')`, so that activity could never be auto-assigned again. The schema now normalizes empty to null rather than relying on the client to.

**7. One transient database blip could leave the rest of an archive import unattributed.** The name→gear cache stored `null` on failure, and `null` is only ever produced by a failed create — so a pool timeout on activity 12 of 3,000 short-circuited every remaining activity with that gear name, with a single log line. Only successes are cached now.

**8. The component rollup was not actor-scoped** (the gear rollup is), and `assignFitnessFileGearIfUnset` enforced no ownership at all. Not reachable through today's callers, but the two rollups disagreed, and a component total absorbing another actor's distance while the gear total stayed correct is not a discrepancy anyone would reproduce. Both now scope by actor.

**9. Dates rendered in local time.** `formatGearDate` had no `timeZone`, while `<input type="date">` values parse as UTC midnight — so a component added "Mar 1" showed as "Feb 29" west of UTC, and the suite genuinely failed under `TZ=America/Los_Angeles`.

### Also fixed

Unbounded epoch milliseconds reaching `new Date()` (a 500 on PostgreSQL, a silent "since the beginning" on SQLite); the component `PATCH` being able to invert an install window into an empty one; `/replace` coercing a malformed body to `{}` and mutating anyway; `normalizeActivityTypeToSportKey('constructor')` returning the `Object` constructor typed as a `SportKey`; a no-op retire toggle re-arming an already-fired reminder; concurrent evaluations double-firing one crossing (the claim is now a conditional write, so the database arbitrates); a delivery failure aborting the rest of the reminder batch; the `status.ts` join lacking a `deletedAt` guard (in the `ON` clause — as a `WHERE` it would have dropped the whole activity); an over-long archive gear name minting a duplicate per activity; and a reserved-binding count that could exceed the repo's SQLite guard.

### Deliberately not changed

- **Undated activities count toward a gear total but not toward a dated component window.** An activity with no timestamp cannot be placed inside `[addedAt, removedAt)`, so excluding it is the honest answer; documenting and testing it rather than guessing. The related doc claim that gear totals "reconcile" with `getFitnessActivitySummary` has been softened — that surface additionally requires a non-null `activityType` and `activityStartTime`.
- **Gear `kind` stays immutable.** A reviewer noted a wrong Strava kind-inference is therefore permanent; the comment claiming the owner can retype it was wrong and has been corrected. Making `kind` mutable is a larger change than this PR should carry.

llun:
## Review loop converged — round 4 clean

Four rounds of sub-agent review, each pointed at the *previous round's fixes* rather than the original diff. Round 4 found no production bug, so the loop stops here.

| round | outcome |
|---|---|
| 1 | Feature review across four slices. 2 critical, 7 major. |
| 2 | Reviewed round 1's fixes — found one had **introduced a regression**. |
| 3 | Reviewed round 2's fixes — no code bug, but proved two shipped with tests that could not catch a revert, and found a remaining race. |
| 4 | Reviewed round 3's changes — **clean**. One comment that overstated its scenario; corrected. |

### What round 4 verified rather than assumed

The one production change in round 3 rewrote a live `UPDATE` whose null return becomes a 404, so it was checked by execution, not reading: the statement compiled for both dialects, a 17-case matrix run against a real database (including a concurrent double-retire via `Promise.all`), and each of the five new tests killed by mutating the code it guards. Independently, I walked the same matrix myself before the review returned — retiring an already-retired gear returns the gear with its date **unmoved**, and nonexistent, foreign and soft-deleted gear all return null.

It also confirmed the fake-timer test cannot evict the shared in-memory SQLite connection (knex pins `pool {min:1, max:1}` for that dialect, and `toFake: ['Date']` leaves timers real), and that faking `Date` does not disturb knex's timestamp binding on either backend.

### The most useful thing the loop produced

Round 3's finding was not a bug — it was that **two fixes were guarded by assertions that would pass against the unfixed code**. One compared two timestamps taken ~0 ms apart; replayed 200× against the pre-fix code it passed 183 times, and CI only runs SQLite where the gap never opens. A test like that is worse than no test, because it advertises safety it does not provide. Both are now deterministic and verified by reverting the fix.

Every behavioural fix in rounds 1–3 is now mutation-verified: revert the fix, watch the specific test fail.

### Deliberately not changed

- An undated activity counts toward a gear total but not toward a dated component window — an activity that cannot be placed in time cannot be placed inside an install window. Documented and tested rather than guessed at.
- Gear `kind` stays immutable, so a wrong Strava kind-inference is permanent. The comment claiming the owner could retype it was wrong and has been corrected to state the real consequence.
- Archive and webhook imports can create two rows for one physical gear (name-keyed vs id-keyed). Known limitation, no merge affordance.
- Strava CSV headers and gear-id prefixes remain unverified against a real export; every path degrades to "no gear" rather than throwing, and archive gear kind is keyed off *which file* a name came from so a renamed header cannot mis-file a bike as shoes.

All gates green: 777 files / 8,754 tests, build compiles, lint 0 errors.

PR #1412, opened 2026-08-12
none: add fitness gear import and Strava export converter scripts

llun:
## Why

Gear tracking (#1402) arrived after most activities did, so anything imported before it carries no gear at all. Re-importing would duplicate the posts, and the Strava import paths only attribute activities they import themselves — which leaves no way to backfill years of history.

## What

**`scripts/fitness/importFitnessGear.ts`** — creates gear and its component history from a JSON file, then attributes existing activities by matching each entry's timestamp against `activityStartTime`.

- **Assignments** are the precise instrument (one `{time, gear}` entry per activity); **per-gear date windows** are the blunt one, applied afterwards to whatever is still unassigned. Windows are half-open `[from, to)` so back-to-back periods share a boundary date without both claiming the activity that starts on it.
- An entry that matches nothing, ties between two activities, or lands on one another entry already claimed is **reported and skipped, never guessed at**. The dry-run report includes how far the nearest activity was for each unmatched entry, so a systematic clock problem shows up as a uniform offset before anything is written.
- Activities that already carry gear are left alone unless `--overwrite` is given, so re-running is free and never undoes a manual correction. Gear is found by name then Strava id, and components are diffed on `(type, brand, model, addedAt)` — a bike carries several identical chains over its life, distinguished only by when they went on.
- Dates are a `YYYY-MM-DD` day (UTC midnight) or a datetime with an explicit `Z`/offset. A bare local datetime is **rejected**: JavaScript would read it in the running machine's zone and shift every activity by hours, which mostly manifests as a sea of unmatched entries with no obvious cause.

**`scripts/fitness/convertStravaExportToGearImport.ts`** — builds the assignments from a Strava export. `activities.csv` is the only record of which gear each historical activity used, and its `Activity Date` column is UTC (it matches the exported GPX `<time>` to the second). Component install dates exist only on Strava's gear pages, so the gear description stays hand-authored and is merged in.

Parsed in arrays mode because the CSV repeats five header names — the second `Distance` is the one in meters, and it is the only column that reproduces the totals Strava shows, which makes it the checksum for the whole conversion.

**`scripts/fitness/gearImportPlan.ts`** — the schema, matcher, window pass and component diff as pure functions, following the `storedImportPlan.ts` pattern so the interesting decisions are testable without a database.

## Verification

- `yarn lint` (0 errors), `yarn build`, `yarn test` — **8,824 tests pass**, 70 of them new.
- Ran end-to-end against a local SQLite database seeded with **real timestamps from a 1,745-row Strava export**: 40/41 activities matched (the 41st has no start time and is reported), re-run was fully idempotent (0 written, 40 already assigned), `--overwrite` corrected 40 deliberately wrong assignments, a duplicate-timestamp twin was correctly flagged ambiguous, and a windows-only file claimed exactly the 21 post-2018 bike rides.
- The converter reproduced every per-gear total from the Strava gear page (Moots 35,670.2 km, Brompton S6R 4,228.3 km, Rental bike 518.2 km).
- Browser: gear page shows derived distances and retired grouping; component install windows render with correctly apportioned distance (two chains sharing an Apr 1 2020 boundary, 431.4 km vs 0.0 km, no double-counting).

## Notes

- No schema change, no runtime change — scripts and docs only, hence `none:`.
- The import does not evaluate service reminders, so backfilling years of activities sends no alerts.
- `defaultSports` takes a sport from whatever gear holds it now; the script warns before doing so. A purely historical import should leave it empty.

### DIFF
--- a/docs/fitness-file-storage.md
+++ b/docs/fitness-file-storage.md
@@ -166,6 +166,8 @@ Fitness maintenance scripts live in `scripts/`:
 - `scripts/fitness/importStoredFitnessFile.ts` — rebuilds a post from the already-stored file with no Strava call (deleted-from-Strava case); merges same-ride files into one post
 - `scripts/fitness/repairStravaActivityFiles.ts` — pass `--delete-missing` to hard-delete activities Strava 404s (default: report only; deletion is irreversible)
 - `scripts/fitness/backfillFitnessMovingTime.ts` — recomputes `movingTimeSeconds` for already-stored files so their average pace/speed switches from elapsed-time to moving-time (matching Strava); idempotent, supports `--dry-run` and `--force`
+- `scripts/fitness/importFitnessGear.ts` — creates gear and component history from a JSON file and attributes existing activities to it by matching `activityStartTime`, for activities imported before gear tracking existed; idempotent, supports `--dry-run`, `--overwrite` and `--tolerance-seconds`
+- `scripts/fitness/convertStravaExportToGearImport.ts` — builds that import file's assignments from a Strava export's `activities.csv` (the only record of which gear each historical activity used), merged with a hand-authored gear description
 - `scripts/fitness/retrigerStravaActivities.ts`
 - `scripts/fitness/runImportStravaActivity.ts`
 - `scripts/fitness/listStravaWebhooks.ts`
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -357,6 +357,7 @@ NODE_ENV=production ./scripts/fitness/fixStuckFitnessProcessing.ts --actor-id ht
 NODE_ENV=production ./scripts/fitness/recreateFitnessRouteHeatmaps.ts --actor-id https://your-domain.tld/users/username --dry-run
 NODE_ENV=production ./scripts/fitness/repairStravaActivityFiles.ts --actor-id https://your-domain.tld/users/username --dry-run
 NODE_ENV=production ./scripts/fitness/backfillFitnessMovingTime.ts --actor-id https://your-domain.tld/users/username --dry-run
+NODE_ENV=production ./scripts/fitness/importFitnessGear.ts --actor-id https://your-domain.tld/users/username --input ./gear-import.json --dry-run
 NODE_ENV=production ./scripts/fitness/retrigerStravaActivities.ts --actor-id https://your-domain.tld/users/username --activity-id 123456789
 NODE_ENV=production ./scripts/fitness/listStravaWebhooks.ts @username@your-domain.tld
 ```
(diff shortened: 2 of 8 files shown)


llun:
## Review round 1 — findings addressed in 153adec1

Two sub-agent reviews (correctness; error handling / silent failures) found real defects. All are fixed; the matcher's binary search, tie-breaking, half-open window boundaries, component identity round-trip and `stravaGearId` uniqueness handling were independently verified as correct.

**Wrong attribution (correctness):** an assignment that tied between two activities was skipped as `ambiguous` but left both files unclaimed, so a window belonging to a *different* gear could then take them — putting a ride on the wrong bike precisely because the right answer was too uncertain to use. Files an assignment spoke for are now reserved whether or not it resolved (`reservedFileIds`), with a regression test.

**Silent failure:** `assignFitnessFileGearIfUnset` returns `false` for four different reasons and all four were counted as "already had gear". The matcher already recorded whether the activity carried gear at plan time and the script ignored it, so a row that changed under a running import was reported as benign and could never produce a non-zero exit code. In the window loop there is no benign reading at all, since only gearless activities get that far.

**New guards, both refusing before any write:**
- the actor has no activities but the file has assignments — creating gear there would put rows on the wrong actor and `stealDefaultSports` would strip sports off their real gear;
- two file entries resolving to one existing gear row, which would insert its component history twice.

Plus: a final non-zero exit when no assignment reached any activity; the gear, activity and reason named at every failure instead of only counted; the summary preserved when a write throws mid-run; `reuse` now says existing gear fields are left alone rather than reading as "already up to date"; kind mismatches listed per gear with an "every one of them" hint that catches a wrong `kind` while it is still fixable; and a note when matched activities are not `completed` yet and so will not count toward totals.

In the converter: a missing meters column was swallowed into a confident `0.0 km` for every gear — the one number the docs offer as the check that gear names were transcribed right — and a malformed row threw `Unparseable Strava activity date:` with no line number in a 1,700-row file. Both now say what happened; a non-empty CSV that yields zero assignments is an error rather than a green run.

`yarn lint` 0 errors, `yarn build`, `yarn test` 8,827 passing (95 in these scripts).

llun:
## Review round 2 — one finding, fixed in da47762c + 6a99da56

The second pass confirmed the round-1 fixes are correct (closure scope at the catch, the `remaining` off-by-one, guard reachability, the counter split, `reservedFileIds` semantics) and found one real defect in them.

**The CSV error's line number was a record index, not a physical line.** It drifted low after every blank line and every description containing a newline — both routine in a real export — so `line 640` would send you to an unrelated, valid row. A confidently wrong number is worse than none, because it gets trusted; it defeated the exact hunt the number was added to prevent. Now read from `csv-parse`'s `info`, which carries the physical line each record ends on, with a test that fails on the drift.

That fix then broke `yarn build`: `csv-parse`'s sync signature always declares `string[][]` and does not model how `info: true` changes the shape, so the cast has to go through `unknown`. Worth recording that **Vitest transpiles without type checking** — all 8,828 tests passed while `tsc` failed.

Verified after both commits: `yarn build` exit 0, `yarn test` 8,828 passing, converter reproduces the same 1,480 assignments and per-gear totals from the real 1,745-row export.

PR #1413, opened 2026-08-12
fix: size the fitness stat strips by container and inline the gear row

llun:
Aligns the fitness status **item** (the inline chip in a post) and the **activity detail page** with the design system — `FitnessKit.StatGrid`, `FitnessChip` and `FAGearRow` in `ui_kits/web/`.

## The stat strips measure themselves, not the viewport

The design's grids read their **own width** with a `ResizeObserver`, because a strip can sit in a narrow column on a wide window. Both strips on the activity page and the chip's grid were still on `sm:grid-cols-4`, a viewport query, so:

- On a **tablet** the header strip is a 565px column inside a 768px viewport — four `text-[28px]` tiles sat side by side at 132px each and "31.1 km/h" wrapped onto two lines.
- On a **phone** it stayed 2-up, wrapping "MOVING TIME" and the pace value.
- A post **chip** in a wide column stayed 2-up at any window under 640px, with 500px+ to fill.

The new shared `FitnessStatGrid` carries the design's two rules as Tailwind v4 container queries:

| variant | columns | used by |
| --- | --- | --- |
| `detail` | 1 → 2 at **420px** → 4 at **780px** | activity header strip, strip under the route map |
| `chip` | 2 → 4 at **424px** | inline fitness chip in a post |

Measured on the running page:

| viewport | header strip | strip under map | post chip |
| --- | --- | --- | --- |
| 1280px | 824px → 4-up | 866px → 4-up | 796px → 4-up |
| 768px | 580px → 2-up | 622px → 2-up | 552px → 4-up |
| 390px | 282px → 1-up | 324px → 1-up | 246px → 2-up |

The two page strips measure **separately** — the header one sits inside the card's `p-5` and is 42px narrower — so they can differ by one step in a narrow band of widths. That is the rule reading real available width, the same reasoning `useCompactActionBar` uses on the post action row.

`@container` goes on a wrapper, never on the grid: a container query styles a container's *descendants*, not the container itself, so a grid cannot both establish the container and read it.

The chip keeps its own rule rather than inheriting the detail one. Its values are `text-sm`, so four cells still fit in 424px (the kit derives that from 4×100px + 3×8px of gap), and a 4-row chip in a feed is a worse trade than a slightly tight cell.

## Gear moves into the metadata line

The design reads the recording metadata as one wrapping line — `date · visibility · gear` — where the app had gear as a labelled `GEAR` field with a `<select>` on a row of its own, between "Recorded with…" and the caption. Inline takes two rows off the header.

- **Owner** — a dropdown with the same chrome as the section sub-nav below it, listing each candidate with its lifetime distance (the design's `27.2 km` column, which is what tells two similar bikes apart) plus "No gear". Unchanged behaviour: disabled while the PATCH is in flight, optimistic with a per-file rollback, and the assigned gear is always in the list even when the kind filter or its retirement would drop it.
- **Everyone else** — the kind's icon and the name.
- **Nobody** when no gear is attributed, including an owner with an empty shed — a menu whose only entry is "No gear" is dead UI.

The trigger is named from its content rather than by `aria-label`, so a screen reader hears "Gear: Moots Routt YBB" instead of a bare "Gear" that says nothing about what is assigned. The kind's icon is `Bike`/`Footprints`, falling back to `Wrench` — the app's own Gear-section icon — when neither the assigned gear nor the activity type says which kind it is.

## Tests

- New `lib/components/fitness/FitnessStatGrid.test.tsx` covers what jsdom can actually protect — the `@container`-on-the-wrapper structure (inverting it is the failure with no symptom), the caller `className` landing on the container, and the deliberate difference between the two variants' floors. It does **not** try to restate the thresholds: jsdom never evaluates a container query, so that would only fail when someone edits the constant on purpose.
- The existing gear suite is rewritten against the dropdown (same nine behaviours, including the in-flight disable, the per-file rollback with a file list landing mid-flight, and clearing to `null`).
- Full gate green: prettier → lint (0 errors) → build → 8832 tests.

## Not migrated (known gap)

Two older fitness stat strips stay as they are, and are now named as such in `AGENTS.md` rather than implied to be done: the gear detail page's strip (`GearDetailView.tsx`, still a `sm:` **viewport** query — the same defect, on a page outside this ask) and the fitness overview's totals (`ActorFitnessDashboard.tsx`, container-queried but hand-rolled on its own `@2xl/fitness` threshold). Migrating both is a worthwhile follow-up.

## Verified

Locally against SQLite with seeded gear and a real GPX (timestamps + heart rate), at 1280 / 768 / 390px in light and dark, signed in as the owner and logged out.

## Docs

`AGENTS.md` gains a **Fitness Stat Strips** section and a gear-presentation bullet; `REVIEW.md` and `CLAUDE.md` get the matching checklist line.


app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,144 @@ const StatTile: FC<{
...
+        </DropdownMenuTrigger>
+        {/* Same menu chrome as the section sub-nav above. */}
+        <DropdownMenuContent align="start" className="rounded-xl shadow-lg">
+          {[...options, null].map((option) => {
+            const id = option?.id ?? ''
+            const isActive = (selectedGearId ?? '') === id
+            return (
+              <DropdownMenuItem
+                key={id || 'no-gear'}
+                onSelect={() => onSelect(id)}
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1007]: **Regression vs the `<select>`: re-picking the gear that is already assigned fires a redundant PATCH.**

`onSelect` fires for every row including the checked one, and `handleGearChange` has no equality guard. A native `<select>` does not fire `onChange` when the user re-picks the current option, so this is new behaviour.

Failure scenario: the owner opens the menu to see what is attributed and taps the checked row to dismiss it — the natural gesture — which writes `PATCH /api/v1/fitness-files/:id`, re-runs `evaluateGearServiceReminders`, disables the picker for the round trip, and on any network hiccup shows a red `role="alert"` "Failed to update gear." for a change they never made. Same for choosing "No gear" when nothing is assigned.

Fix: early-return in `handleGearChange` when `value === (selectedGearId ?? '')`.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,144 @@ const StatTile: FC<{
...
+        </span>
+      </>
+    )
+  }
+
+  return (
+    <>
+      <span aria-hidden="true">·</span>
+      <DropdownMenu>
+        <DropdownMenuTrigger asChild>
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: **The in-flight guard is on the `asChild` child, not on `DropdownMenuTrigger`, so Radix's own open handlers stay live.**

Radix reads its *own* `disabled` prop (`onKeyDown: (event) => { if (disabled) return; … }`); `@radix-ui/react-slot`'s `mergeProps` only lets the child's `disabled` win on the rendered DOM node, leaving Radix's internal `disabled === false`. Verified in jsdom: while the PATCH is in flight, `fireEvent.keyDown(trigger, { key: 'ArrowDown' })` still opens the menu — the exact second-change race the disable exists to prevent.

In a real browser a native disabled `<button>` is unfocusable and swallows pointer events, so the practical hole is small (confirmed by driving the running page), but the protection has moved from the component to native form semantics and the new test asserts only `toBeDisabled()`, which passes either way.

Fix: `<DropdownMenuTrigger asChild disabled={isSaving}>` (keep the attribute on the child for native semantics and styling), plus an assertion that the menu does not open while saving.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,144 @@ const StatTile: FC<{
...
+        {/* Same menu chrome as the section sub-nav above. */}
+        <DropdownMenuContent align="start" className="rounded-xl shadow-lg">
+          {[...options, null].map((option) => {
+            const id = option?.id ?? ''
+            const isActive = (selectedGearId ?? '') === id
+            return (
+              <DropdownMenuItem
+                key={id || 'no-gear'}
+                onSelect={() => onSelect(id)}
+                aria-current={isActive ? 'true' : undefined}
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: **The current selection is announced only by `aria-current` on a `role="menuitem"`; the `<select>` announced "selected".**

The `Check` icon is `aria-hidden`, so `aria-current` is the sole state signal — and support for it on `menuitem` is patchy. It was copied from the *navigation* dropdowns (the section sub-nav below, and `section-nav-dropdown.tsx`), where the state is also carried by the URL.

This menu is pick-one-of-N, so the right semantics are `role="menuitemradio"` + `aria-checked`.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,144 @@ const StatTile: FC<{
...
+            disabled={isSaving}
+            className="inline-flex min-w-0 items-center gap-1.5 rounded-md transition-colors hover:text-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary/50 disabled:cursor-not-allowed disabled:opacity-50"
+          >
+            <Icon className="size-3.5 shrink-0" aria-hidden="true" />
+            {/* Named from its content rather than by `aria-label`, so the
+                accessible name carries the current value too — "Gear: Moots
+                Routt YBB", not a bare "Gear" that says nothing about what is
+                assigned. */}
+            <span className="sr-only">Gear: </span>
+            <span className="truncate">{label ?? 'No gear'}</span>
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: **A long gear name truncates on the owner's trigger with no tooltip.**

The read-only branch above carries `title={`Gear: ${label}`}`, but the picker trigger has none, and `fitness_gears.name` is `varchar(255)`. An owner whose gear is "Specialized Tarmac SL7 Expert (winter build)" sees a cut name and has to open the menu to read it.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,144 @@ const StatTile: FC<{
...
+  isSaving,
+  errorId,
+  onSelect
+}) => {
+  const selected = options.find((option) => option.id === selectedGearId)
+  const label = selected?.name ?? gearName ?? null
+  const kind = selected?.kind ?? activityKind
+  const Icon = kind ? GEAR_KIND_ICON[kind] : Wrench
+
+  if (!isOwner || options.length === 0) {
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:947]: **The "empty shed" branch that AGENTS.md singles out has no test.**

This is what implements "nobody gets anything when no gear is attributed, **including** an owner whose shed is empty". Change it to `if (!isOwner)` and the whole suite still passes: the only `queryByText(/^Gear/)`-style assertion is in the *non-owner* test, and every owner test stubs a non-empty `getFitnessGearList`.

This PR rewrote the whole gear suite, so it is the moment to pin it — one `renderDetail()` with `mockGetFitnessGearList.mockResolvedValue([])` asserting no `/^Gear:/` button.

llun (commented):
Sub-agent review round 1 — three reviewers over the diff (layout/Tailwind correctness, gear-picker regressions, docs + test quality). Findings posted inline below. Tailwind v4 syntax was verified by compiling the changed files through the repo's own Tailwind 4.3.3 (`@container` + `(width >= 420px|424px|780px)` all emit correctly), and `container-type: inline-size` was checked for containing-block/clipping side effects — nothing inside either grid is positioned, so both are clean.

AGENTS.md
```diff
@@ -347,6 +347,36 @@ it; there is no legacy shape left to copy.
   Outlook's Word engine is the one that needs `mso-` properties and the ghost
   table, and none of that is observable in a browser.
 
+## Fitness Stat Strips
+
+- **Every labelled stat strip on a fitness surface renders through
```
llun: **This "Every …" claim is literally false of the tree after this PR, and one counter-example uses the exact viewport breakpoint the new rule forbids.**

- `app/(timeline)/fitness/gear/[id]/GearDetailView.tsx:231` renders three labelled `StatTile`s (Distance / Activities / Components installed) inside a hand-rolled `grid grid-cols-1 gap-3 sm:grid-cols-3`. That is a viewport query on a fitness surface — the same defect this PR's narrative says it fixes, and it hits the same ~565px tablet column.
- `app/(timeline)/fitness/ActorFitnessDashboard.tsx:261` hand-rolls `grid grid-cols-2 gap-2 @2xl/fitness:grid-cols-4`. Container-based, so not broken, but a third unrelated 4-up threshold (672px) for the same visual pattern.

Restated at `CLAUDE.md:51`. `REVIEW.md` is the one version that is already correct, because its parenthetical scopes it to the three strips.

Failure scenario: the next agent takes the invariant at face value, then either edits `GearDetailView.tsx` without noticing it contradicts a documented rule, or cites the rule in review and gets a contradiction with committed code. Either narrow the wording to the three strips actually covered, or migrate the gear detail strip.

lib/components/fitness/FitnessStatGrid.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+const renderGrid = (props: Parameters<typeof FitnessStatGrid>[0]) =>
+  render(<FitnessStatGrid {...props} />)
+
+describe('FitnessStatGrid', () => {
+  it('measures its own container rather than the viewport', () => {
+    renderGrid({ children: <div data-testid="cell" /> })
+
+    const grid = getGrid()
+    expect(grid.parentElement).toHaveClass('@container')
+    expect(grid.className).not.toMatch(/(^|\s|:)sm:grid-cols/)
```
llun: **Tautological assertion.** The class string comes from the module-level `VARIANT_CLASS_NAMES` literal, which the `toHaveClass` assertions already pin exactly, so this can never fail independently. It reads as protection against the viewport-breakpoint regression while protecting nothing — and that regression is still live one directory away (`GearDetailView.tsx:231`).

Same class of problem for the two threshold tests: jsdom never evaluates a container query, so restating `VARIANT_CLASS_NAMES` verbatim verifies nothing a reader of the constant could not, and changing 420→480 fails the test with no possible response but editing the test.

The two that *do* protect something real and non-obvious are `measures its own container rather than the viewport` and `puts the caller class on the container, not the grid` — the `@container`-on-the-wrapper structure is the one thing that silently produces zero styling if inverted. The repo's precedent for a claim as broad as the `sm:` one is a source-walking guard (`lib/components/tailwindCssVariableSyntax.test.ts`).

lib/components/fitness/FitnessStatGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * - `detail` — the activity page's header strip and the strip under its map.
+ *   Values are 21–28px, so they need ~200px a cell: 4-up from 780px, 2-up from
+ *   420px, 1-up on a phone.
+ * - `chip` — the inline card inside a timeline post. Values are `text-sm`, so
+ *   four cells still fit in 424px (the kit derives that from 4×100px + 3×8px of
+ *   gap) and it never drops to a single column — a 4-row chip in a feed is a
+ *   worse trade than a slightly tight cell.
+ */
+const VARIANT_CLASS_NAMES = {
+  detail: 'gap-3 grid-cols-1 @min-[420px]:grid-cols-2 @min-[780px]:grid-cols-4',
```
llun, line 30: **Note on the `detail` floor: no phone reaches the 2-up band.**

The header strip's container is the card body, so the chrome is layout `px-4` (32) + page `p-4` (32) + card border and `p-5` (42) = 106px. At a 430px viewport — the largest common phone — the container is ~324px, and 420px is not crossed until roughly a 528px viewport. So on every phone the four header tiles stack into four rows rather than two, which is ~340px more vertical scroll on the most-used viewport.

This is intended rather than a defect: it is the design system's own rule (`FitnessKit.StatGrid` — `w >= 780 ? 4 : w >= 420 ? 2 : 1`) and the explicit ask was 4 desktop / 2 tablet / 1 mobile. Verified in a browser at 390px: nothing wraps any more, where the previous 2-up broke "MOVING TIME" and the pace value onto two lines. Recording it here so the trade is written down — the `chip` variant deliberately reasons the opposite way, and the difference is the type size (28px vs `text-sm`).

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:1007]: Fixed in d7e7b3b0 — `handleGearChange` early-returns when `value === (selectedGearId ?? "")`.

Verified both ways: a new test (`ignores re-picking the gear that is already assigned`) asserts `updateFitnessFileGear` is never called, and removing the guard makes it fail. Also confirmed in a real browser — opening the menu and clicking the checked row produces no `PATCH /api/v1/fitness-files/:id` in the network log.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Confirmed and fixed in d7e7b3b0 — `disabled` now goes on `DropdownMenuTrigger` as well as on the button (the child keeps it for native semantics and the disabled styling).

I checked the claim before complying rather than taking it on trust: reverting just the trigger prop makes the new assertion fail with "expected document not to contain element, found `<div role=menu>`", so the keyboard path really was open while saving. The test now asserts both `toBeDisabled()` **and** that `ArrowDown` opens no menu.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Fixed in d7e7b3b0 — the rows are `role="menuitemradio"` + `aria-checked` now, and `aria-current` is gone. Kept the `DropdownMenuItem` chrome rather than `DropdownMenuRadioItem` so the design's check mark survives (shadcn's radio item hard-codes a circle indicator); Radix spreads `role` onto the item, and its roving focus and typeahead do not depend on the attribute.

Verified in the browser: three `menuitemradio` rows, `aria-checked` true on the assigned one, no `menuitem` left in the menu — and picking a different gear still persists (the `PATCH` landed and the row survived a reload).

Also noting the mobile trade you raised: the OS wheel picker is gone. That is the design system's own control for this line (`FAGearRow`), and the same Radix menu already drives the section sub-nav directly below it on the same page, so the page is not mixing paradigms — but it is a real difference from a native `<select>` and worth having on the record.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Fixed in d7e7b3b0 — the trigger carries `title={`Gear: ${label ?? "No gear"}`}`, matching the read-only branch.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:947]: Fixed in d7e7b3b0 — `shows an owner with no gear at all nothing to pick` renders with `mockGetFitnessGearList.mockResolvedValue([])` and asserts neither a `/^Gear:/` trigger nor any `/^Gear/` text. You were right that nothing covered it: the branch could be deleted with the suite still green.

AGENTS.md
llun: You are right, and this is the finding I am most glad came back — the sentence was false when I wrote it.

Fixed in d7e7b3b0 by narrowing rather than migrating: AGENTS.md now says **three** strips render through `FitnessStatGrid`, and names both hold-outs explicitly — `GearDetailView.tsx` (still a `sm:` viewport query, same defect) and `ActorFitnessDashboard.tsx` (container-queried but hand-rolled on `@2xl/fitness`) — as a worthwhile follow-up. CLAUDE.md carries the same correction and REVIEW.md points at the section.

Migrating them in this PR would take it outside the ask (the fitness status item and page), and the gear detail strip in particular deserves its own before/after check — so it is written down as a known gap instead of quietly implied to be done.

lib/components/fitness/FitnessStatGrid.test.tsx
llun: Agreed, and cut in d7e7b3b0. The tautological `sm:` assertion is gone, and so are the two tests that restated `VARIANT_CLASS_NAMES` verbatim.

What is left is three tests that can actually fail for a real reason: the `@container`-on-the-wrapper structure (asserted in both directions — the grid must NOT carry it), the caller `className` landing on the container rather than the grid, and the deliberate difference between the two variants' floors (`detail` 1-up vs `chip` 2-up), which is the tempting thing for someone to "simplify" into one rule. I did not add the source-walking guard, because it would fail today on `GearDetailView.tsx` — which is now documented as a known gap instead.

The PR body's "asserts no `sm:` viewport variant creeps back in" claim is dropped with it.

lib/components/fitness/FitnessStatGrid.tsx
llun: Intended, keeping as-is — recording the reasoning here rather than changing it.

The rule is the design system's own (`FitnessKit.StatGrid`: `w >= 780 ? 4 : w >= 420 ? 2 : 1`) and the ask was explicitly 4 desktop / 2 tablet / 1 mobile. Your arithmetic is right that no phone reaches the 2-up band, and that is the trade being made: the previous 2-up at 390px wrapped "MOVING TIME" onto two lines and broke "31.1 km/h" across a line, which is what the change set out to fix. Verified in a browser at 390px — every tile is now one clean line.

Your point about the chip reasoning the opposite way is exactly right, and that difference is now the one variant test that survived the trim.

lib/components/fitness/FitnessStatGrid.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+// purpose, and could not catch a hand-rolled strip elsewhere anyway.
+const getGrid = () => screen.getByTestId('cell').parentElement as HTMLElement
+
+const renderGrid = (props: Parameters<typeof FitnessStatGrid>[0]) =>
+  render(<FitnessStatGrid {...props} />)
+
+describe('FitnessStatGrid', () => {
+  it('establishes the container on the wrapper, not on the grid it sizes', () => {
+    renderGrid({ children: <div data-testid="cell" /> })
+
```
llun, line 20: **Round 1's trim went one assertion too far, and this reviewer is right.**

Dropping the two threshold tests was correct — jsdom cannot evaluate a container query, so restating `VARIANT_CLASS_NAMES` only re-asserts the constant. But the `not.toMatch(/sm:grid-cols/)` line that went with them was a *negative* invariant, and with the threshold tests gone it was no longer redundant: it was the only thing standing between this component and the viewport breakpoint the whole change exists to remove.

I verified before complying rather than taking it on trust — appending `sm:grid-cols-4` to `VARIANT_CLASS_NAMES.detail` passed all three remaining tests.

Fixed in f62c91b6: re-added as a table-driven case over both variants, widened past `sm:` to every breakpoint prefix.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2172,7 +2332,15 @@ export const FitnessStatusDetail: FC<Props> = ({
     null
   )
   const [gearOptions, setGearOptions] = useState<GearEntity[]>([])
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2334]: **The round-1 fix for the orphaned alert only covered the settled case — the in-flight one still mis-attributed the error.**

Clearing `gearUpdateError` on the file switcher's `onChange` does nothing when the PATCH is still open: `isSavingGear` disables only the gear trigger, the file `<select>` stays enabled, so the failure lands *after* the clear. `handleGearChange` scopes its rollback to the captured `fitnessFileId`, but the error state was component-global — so file 1's failure rendered as file 2's alert, with `aria-describedby` wiring file 2's picker to it.

Fixed in f62c91b6 by scoping the error to its file (`{ fitnessFileId, message }`) rather than by disabling the switcher — blocking a legitimate action for the length of a network round trip is the worse trade. The `onChange` clear is gone with it: one mechanism, and switching back now brings the error with the file it belongs to, which is what actually happened.

Covered by `keeps a failed change with its own file when the reader switches`, verified causal — unscoping the error fails it.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -885,6 +892,159 @@ const StatTile: FC<{
...
+            and the guard resting on native form semantics. */}
+        <DropdownMenuTrigger asChild disabled={isSaving}>
+          <button
+            type="button"
+            aria-describedby={errorId}
+            disabled={isSaving}
+            className="inline-flex min-w-0 items-center gap-1.5 rounded-md transition-colors hover:text-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary/50 disabled:cursor-not-allowed disabled:opacity-50"
+          >
+            <Icon className="size-3.5 shrink-0" aria-hidden="true" />
+            {/* Named from its content rather than by `aria-label`, so the
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:985]: **The `title` added in round 1 duplicated the button's own accessible name.**

The button is deliberately named from its contents (`sr-only` "Gear: " + the label), so per HTML-AAM a `title` on it becomes the accessible *description* — name and description the same string, which is the double-read `REVIEW.md` warns about in the read-only-stats rule.

Fixed in f62c91b6 by moving it to the `truncate` span, which is the element that actually clips. The hover tooltip survives, the description does not. Confirmed on the running page: the button has no `title`, its accessible name is "Gear: Moots Routt YBB", and the span carries the tooltip.

Left the read-only branch's `title` on its outer span: it is a non-interactive `<span>` with no role, so there is no name/description pair to duplicate there.

llun (commented):
Sub-agent review round 2 — two reviewers: one on whether the round-1 fixes introduced anything, one fresh-eyes over the whole diff. The fresh-eyes pass returned no findings (it independently confirmed the truncation mechanics inside the nested `inline-flex`, that the extra wrapper div is layout-equivalent, that the `·` separator never orphans in any branch, and that no dead code or stale comment survives the gear change). The fixes-focused pass found three, all fixed in f62c91b6 and posted inline. It also verified by reverting each in turn that all three round-1 gear tests are causal, and read the Radix 2.1.24 sources to confirm the `role` override and the double `disabled` are safe.

lib/components/fitness/FitnessStatGrid.test.tsx
llun: Fixed in f62c91b6 and verified causal: with the assertion back, appending `sm:grid-cols-4` to the constant fails it. Widened past `sm:` to every breakpoint prefix and made it table-driven over both variants.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2334]: Fixed in f62c91b6. Error state is now keyed to its file, the `onChange` clear is gone, and a new test (`keeps a failed change with its own file when the reader switches`) covers it — verified causal by unscoping the error.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:985]: Fixed in f62c91b6 — the `title` moved to the truncating span. Verified on the running page: no `title` on the button, accessible name still "Gear: Moots Routt YBB", tooltip intact.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2319,7 +2508,13 @@ export const FitnessStatusDetail: FC<Props> = ({
     const previousGearId = fitness?.gearId ?? null
     const previousGearName = fitness?.gearName ?? null
 
-    setGearUpdateError(null)
+    // Scoped like the render is: an unconditional reset here would drop a
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: **Round 2 scoped where the error RENDERS but left the reset unconditional, so a change to one file discarded another file's pending failure.**

The sequence the scoping exists to handle ended worse than before it: owner assigns gear on file 1 → switches to file 2 (the switcher stays enabled during the PATCH) → file 1's PATCH fails and is stored hidden → owner successfully assigns gear on file 2 → `setGearUpdateError(null)` drops file 1's error → switching back shows file 1 reading "No gear" with **nothing** saying why. The reader never saw the message at any point, and the optimistic value had already been rolled back, so the failure was completely silent.

Fixed in f1a57f5a — the clear is now keyed to the same file the render is:

```ts
setGearUpdateError((current) =>
  current?.fitnessFileId === fitnessFileId ? null : current
)
```

Covered by `does not let a change to one file discard another file's failure`, verified causal — reverting to the unconditional clear fails it.

lib/components/fitness/FitnessStatGrid.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    'sizes the %s variant without a viewport breakpoint',
+    (variant) => {
+      renderGrid({ variant, children: <div data-testid="cell" /> })
+
+      // The one NEGATIVE invariant worth pinning, and the whole point of the
+      // component: a viewport query cannot see a strip sitting in a narrow
+      // column on a wide window. Unlike the thresholds — which jsdom cannot
+      // evaluate and which restating here would only re-assert — this fails on
+      // a real regression, and it is one step away in the same file.
+      //
```
llun, line 39: **The reinstated guard enumerated breakpoint names, which misses the regression that actually threatens this file: losing one character.**

`min-[420px]:` is a perfectly valid Tailwind v4 **viewport** variant — it compiles to `@media (width >= 420px)` where `@min-[420px]:` compiles to `@container (width >= 420px)`. Dropping the `@` in a file where every threshold is written that way reverts the component to precisely the defect this PR removes, with no build error, no lint error, and nothing on screen until someone opens a tablet. `max-sm:grid-cols-1` slipped past the name list too (the `-` before `sm` fails the `(^|\s|:)` prefix).

Fixed in f1a57f5a — it now asserts every variant on a `grid-cols` is `@`-prefixed rather than naming breakpoints:

```ts
expect(getGrid().className).not.toMatch(/(^|\s)[^\s@]+:grid-cols/)
```

Verified by mutation: dropping the `@` from `detail` fails the detail case, and `max-sm:grid-cols-1` on `chip` fails the chip case. A legitimate `@sm:grid-cols-4` still passes.

llun (commented):
Sub-agent review round 3, on the round-2 fixes. Two findings, both real and both fixed in f1a57f5a. Everything else on the stuck-error checklist came back clear (successful retry clears it, a deleted/renumbered file id only leaves invisible state residue, `fitness === undefined` hides it, `aria-describedby` and the alert cannot disagree since both derive from `gearErrorMessage`), and the round-1 clear-on-switch behaviour is confirmed not to be missed.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx]: Fixed in f1a57f5a — the clear is keyed to the same file as the render, and `does not let a change to one file discard another file’s failure` covers it (verified causal by reverting to the unconditional clear).

lib/components/fitness/FitnessStatGrid.test.tsx
llun: Fixed in f1a57f5a — the guard now asserts every `grid-cols` variant is `@`-prefixed instead of naming breakpoints. Mutation-verified both ways: dropping the `@` fails the detail case, `max-sm:` fails the chip case, a real `@sm:` still passes.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2172,7 +2332,16 @@ export const FitnessStatusDetail: FC<Props> = ({
     null
   )
   const [gearOptions, setGearOptions] = useState<GearEntity[]>([])
-  const [gearUpdateError, setGearUpdateError] = useState<string | null>(null)
+  // Keyed by file, not a single slot. An activity can aggregate several files
+  // and the switcher stays enabled during a PATCH, so a failure that lands after
+  // the reader has moved on must not render as the NEXT file's error — that
+  // would wire `aria-describedby` from that file's picker to a message about a
+  // different one. One slot was not enough either: whichever file failed second
+  // would evict the first, and the first was invisible at the time precisely
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2340]: **One error slot still discarded a hidden failure — the exact harm the round-3 comment argues against, reached by a second failure instead of a second success.**

Round 3 closed the success path but the `catch` still wrote the single slot unconditionally, so whichever file failed *second* evicted the first — and the first was invisible at the time precisely because its reader had moved on. Two files, two failed changes, and the earlier one vanished with its gear rolled back and nothing anywhere saying why.

The reviewer offered a comment amendment or the full fix. Took the full fix in 80d94560: the state is a `Record<string, string>` keyed by file id, the clear omits one key instead of resetting, and `gearErrorMessage` reads `gearUpdateErrors[fitness.id]`. The comment now describes a guarantee that holds without a caveat, which is worth more than a documented gap in a component whose whole job here is not to lose a message.

Covered by `keeps both files' failures when each change fails in turn`, verified causal — collapsing the map back to a single entry fails it.

llun (commented):
Sub-agent review round 4, on the round-3 fixes. One finding, fixed in 80d94560. Everything else came back verified rather than merely reasoned about: the round-3 guards were mutation-tested (un-scoping the clear fails exactly one test; dropping the `@` fails the detail case), the new regex was probed against 17 class strings the component and tailwind-merge can produce — it passes `@sm:`, `@max-md:`, `@min-[420px]/strip:`, `@container/name` and even a stacked `dark:@min-[420px]:`, and flags `sm:`, `max-sm:`, `min-[420px]:`, `print:`, `group-hover:` — and the comment's central claim was checked by compiling both forms through the repo's own Tailwind 4.3.3 (`@min-[420px]:` → `@container (width >= 420px)`, `min-[420px]:` → `@media (width >= 420px)`). The doc arithmetic was re-derived from the markup and holds, and no stale reference to the old `<select>` gear row survives anywhere in code or docs.

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
Comment (llun) [reply app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2340]: Fixed in 80d94560 — errors are held per file id now, so a second failure cannot evict a first one that was never seen. Verified causal by collapsing the map back to a single entry.

llun:
Sub-agent code-review loop **closed after 5 rounds** — round 5 returned no findings.

Rounds 1–4 produced 14 findings, 12 fixed and 2 answered as intended (the phone 1-up floor, and the `title` left on the read-only branch). Every fix that could be guarded got a test, and each of those was **verified causal by reverting the fix and watching exactly the intended test fail** — the guard against the guard, since three of the five rounds found problems in the *previous* round's fix rather than in the original change.

Round 5 verified the final state rather than reasoning about it: mutation-tested both halves of the per-file error map, instrumented the two-failure test to confirm it cannot pass on a stale rejection (8 consecutive runs, 16/16), read the **built CSS** to confirm `@container (min-width: 420px|424px|780px)` and `container-type: inline-size` actually ship, and confirmed `tailwind-merge` drops none of the variant classes.

One non-finding it raised deliberately, recorded here rather than buried: `isSavingGear` is still a single component-wide boolean while errors are now per-file, so a PATCH on one file also disables a sibling file's picker. That line predates this PR and serializing writes is defensible, so it was left alone.

Local gate green on the final commit: prettier → lint (0 errors) → build → **8838 tests**. CI green on every push.

PR #1415, opened 2026-08-12
chore: update dependencies and close brace-expansion advisories

llun:
## Summary

Routine dependency refresh: every direct dependency that had a newer installable release is bumped, plus `yarn` 4.17.1 → 4.18.0. No application source changed — only `package.json` and `yarn.lock`.

## Security

- **brace-expansion** 2.1.2 → 2.1.4 and 5.0.8 → 5.0.9 (`yarn up -R`), closing two high-severity advisories, both transitive via `minimatch`:
  - [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) — DoS via unbounded expansion length
  - [GHSA-rgw5-rvv9-x895](https://github.com/advisories/GHSA-rgw5-rvv9-x895) — DoS via unbounded intermediate arrays

**nodemailer** 9.0.3 → 9.0.5 also carries header-injection hardening (header-key callbacks and DKIM tags, control characters in header values and Message-ID, `List-*` comment escaping, address normalization between headers and envelope; 9.0.4 adds unpaired-surrogate handling and escapes backslashes/quotes in the Content-Type `name` parameter). Defence-in-depth here rather than a fix for something reachable: `lib/services/email/smtp.ts:25` hand-builds `"name" <addr>` by interpolation, but every `from` is the admin-configured `config.email.serviceFromAddress` and every `to` is a bare account-email string, so no remote-actor display name reaches an address header — display names land in the rendered body, which goes through the templates' own escaping.

`yarn npm audit --all --recursive` is now down to two entries, both covered under **Held back** below.

## Notable changes

- **`@radix-ui/react-slot` un-pinned** from the exact `1.3.0` back to `^1.3.3`. The pin was added in #1308 because 1.3.1 introduced a module-scope `React.createContext` without re-adding its `use client` directive, which broke the RSC build of every page rendering `ui/button.tsx`. The published 1.3.3 bundle has no module-scope context at all, and `yarn build` is green.
- **postcss** bumped in *both* the direct devDependency (`^8.5.26`) and the `resolutions` entry (`8.5.26`), and this one was **forced, not tidy**: `vite@8.2.1` raised its postcss floor to `^8.5.25` (was `^8.5.17` on `main`), so the old `8.5.23` resolution is no longer a version the tree can satisfy. Worth naming the consequence — `next@16.3.0` pins postcss to an **exact** `8.5.23` (`yarn.lock:6557`), which the resolution matched exactly on `main` and now overrides for the first time, so Next's CSS pipeline runs on a postcss it did not pin. Risk is low (8.5.24–8.5.26 are a BOM-preservation fix, two `list.split()` regressions, a visitor regression and symlink handling in source-map path protection), and the alternative — dropping the resolution — re-forks the tree, since `next`'s exact pin cannot float on its own. Five distinct ranges are declared across the tree and all collapse onto the single `postcss@npm:8.5.26` entry.
- **The better-auth stack moves together** (`better-auth`, `@better-auth/core`, `@better-auth/oauth-provider`, `@better-auth/passkey` → 1.6.27), matching the dependabot group that exists precisely because piecemeal bumps have broken the build twice (#1003, #1011).
- **yarn 4.17.1 → 4.18.0** via the `packageManager` field; corepack picks it up in CI and Docker with no workflow change.

### Two duplications worth naming

**`lightningcss` 1 → 2 copies.** `vite@8.2.1` requires `^1.33.0` (`yarn.lock:8391`) while `@tailwindcss/node@4.3.3` exact-pins `1.32.0` (`yarn.lock:3092`); on `main`, vite 8.1.5's `^1.32.0` was satisfied by tailwind's pin, so there was one shared entry. Two now, plus a second copy of the platform binary (~8.7 MB here). **devDependency-only** — it costs CI/dev install time and a Docker build layer, never the runtime image, since the Dockerfile's output stage copies just `.next/standalone`, `public`, `.next/static`, `data.sqlite`, `sharp` and `@img` — and it un-forks when Tailwind moves its pin. A `resolutions: { lightningcss }` entry would collapse it today, but only by overriding Tailwind's deliberate exact pin for a disk-size win, so I left it alone.

**`rou3` 1 → 2 copies**, and this one touches a *runtime* dependency. The nested `better-call` moves to 1.4.0 with the better-auth bump and wants `rou3@^0.9.1` (`yarn.lock:4167`), while the root-hoisted `better-call@2.0.6` still wants `^0.7.12` (`yarn.lock:4184`); on `main` both wanted `^0.7.12`. Size is irrelevant (92K) and the *duplication* does not reach the image — hoisting puts 0.9.2 at `node_modules/rou3` where better-auth's router uses it, and strands 0.7.12 under `node_modules/better-call/`, serving only the root pin that no source file imports, so `@vercel/nft` will not trace it into `.next/standalone`. The real point is the **version jump**: better-auth's HTTP router goes 0.7.12 → 0.9.2, across rou3's one breaking release (0.8.0), as a side effect of what reads as a patch bump. Downstream of the root `better-call: 2.0.6` pin discussed above — as is the same-shaped `@better-auth/utils: 0.5.0` pin, whose only importer is a test — so one cleanup would fix all of it.

Everything else in this PR moves the other way: 44 packages went from 2–3 copies down to 1 (the whole `@aws-sdk/*`, `@smithy/*` and `@radix-ui/*` internals, `@csstools/css-calc`, `@types/node`, `@types/react`, `undici-types`), which is what the net −583 lockfile lines are.

Also bumped: the four `@aws-sdk/*` clients, eight Radix packages, `eslint`, `@typescript-eslint/*`, `vite`, `lucide-react`, `jose`, `pg`, `kysely`, `marked`, `nanostores`, `nodemailer`, `resend`, `lint-staged`, `globals`, and the `@types/*` and testing-library packages.

## Held back deliberately

- **typescript** stays on `6.0.3` — already the newest 6.x. TypeScript 7 ships no JavaScript Compiler API, which both `@typescript-eslint/*` (the `yarn lint` gate) and `@swc-node/register` (the `scripts/run.cjs` bootstrap) require. Per `AGENTS.md`, the 6.x line is a standing constraint.
- **@better-auth/oauth-provider** stays on stable `1.6.27`. [GHSA-p2fr-6hmx-4528](https://github.com/advisories/GHSA-p2fr-6hmx-4528) still has no patched stable release — the fix lands only in `1.7.0-beta.4+`/`1.7.0-rc.*`, a breaking change requiring a schema migration. This instance configures no `validAudiences`, so the allow-set defaults to this server's own base URL and its `/oauth2/userinfo` endpoint; both are our own endpoints, so the escalation described in the advisory is not reachable. Same assessment as #1308.
- **@aws-sdk/\* 3.1109.0** and **globals 17.10.0/17.11.0** were published inside yarn's default 24-hour npm age gate at the time of this run (aws-sdk 2h old, globals 17.11.0 10h and 17.10.0 ~24h-minus-a-few-minutes), so yarn resolved 3.1108.0 and 17.9.0. Dependabot will pick up the remainder.
- **better-call** stays at `2.0.6`. Its npm `latest` tag is `1.4.0`, so there is nothing newer to move the root pin to — but note what that pin actually is: `1.4.0` is the exact version `@better-auth/core@1.6.27` peer-requires, and `better-auth` pulls it as a real dependency, so a nested `better-call@1.4.0` is installed regardless. The root `better-call: 2.0.6` entry is imported by no source file in the repo (`grep -rn 'better-call' app lib scripts` is empty); its only effect is to hoist an incompatible version over what better-auth wants, which is where the `YN0060` peer warnings come from. Pre-existing on `main` (same six warnings, same shape — this PR adds none), and worth cleaning up separately.
- **crypto-js deprecation notice** (via `@upstash/qstash@2.11.3`, already the latest release) is an upstream deprecation, not a vulnerability, and there is no version to move to.

GitHub Actions were already on their latest major tags (`actions/checkout@v7`, `setup-node@v7`, `upload-artifact@v7`, `download-artifact@v8`, `github-script@v9`, `codeql-action@v4`, the `docker/*` set, `dependabot/fetch-metadata@v3`, `anthropics/claude-code-action@v1`); no workflow changes were needed.

## One follow-up commit

`docs: drop the stale pinned version from the SafeUrlSchema comment` — `app/api/v1/apps/createApplication.ts:3` named `@better-auth/core` **1.6.20** while the pin has since moved four times (1.6.20 → 1.6.23 → 1.6.25 → 1.6.26 → 1.6.27, the last of them in this PR). Since this is the PR that edits the pin, the number goes rather than gets re-stated; `lib/services/oauth/matchRedirectUri.ts:2` already says "exact-pinned" without one, which is the pattern to copy.

## Testing

Full gate, run in order, all green:

- `yarn run prettier --write .` — no changes
- `yarn lint` — 0 errors (warnings are the pre-existing `no-explicit-any` backlog; this change adds no code)
- `yarn build` — compiled successfully
- `yarn test` — **782 files / 8842 tests passed**

Browser-verified against a local SQLite instance (`ACTIVITIES_DATABASE_CLIENT=better-sqlite3`) with the sanctioned `scripts/mock/createMockUser.ts` + `createMockStatuses.ts` seed, since the Radix and lucide-react bumps touch UI. Signed in and walked:

- **Timeline** — Radix Avatar, lucide icons, composer, multi-image attachments all render.
- **Composer visibility menu** — Radix `DropdownMenu` opens, checkmarks and the "Who can quote" group intact.
- **Settings** — the shared `SectionNavDropdown` opens and the menu still matches its trigger width (the Tailwind v4 `w-(--radix-dropdown-menu-trigger-width)` path).
- **Fitness → Files** — Radix `Progress` (storage quota bar), buttons, and the section header render.

No browser console errors and no hydration warnings; the dev-server log had no errors either.


package.json
```diff
@@ -94,42 +94,42 @@
...
+    "lint-staged": "^17.3.0",
     "pino-pretty": "^13.1.3",
-    "postcss": "^8.5.23",
+    "postcss": "^8.5.26",
     "prettier": "^3.9.6",
     "tailwindcss": "^4.3.3",
     "tw-animate-css": "^1.4.0",
     "typescript": "^6.0.3",
-    "vite": "^8.1.5",
+    "vite": "^8.2.1",
```
llun, line 132: **MEDIUM — this bump forks `lightningcss` into two copies.**

`vite@8.2.1` requires `lightningcss@^1.33.0` (`yarn.lock:8391`), while `@tailwindcss/node@4.3.3` **exact**-pins `lightningcss@1.32.0` (`yarn.lock:3092`). On `main`, vite 8.1.5 asked for `^1.32.0` and tailwind's pin satisfied it, so the lockfile had a single shared entry. Now there are two (`yarn.lock:6077` and `:6120`) — plus two copies of each of the 12 `lightningcss-<platform>` binary packages, ~8.7 MB on disk here.

**Not fixing it, deliberately.** `vite` is a devDependency, so this costs CI/dev install time and a Docker build layer, never the runtime image, and it un-forks on its own when Tailwind moves its pin. A `resolutions: { "lightningcss": "1.33.0" }` entry would collapse it today, but only by overriding Tailwind's deliberate exact pin to win back disk space — a worse trade than the duplication. Unlike the postcss resolution, there is no advisory forcing the issue.

Now called out explicitly in the PR description so it isn't a silent regression in an otherwise heavily-deduping PR (45 packages went 2–3 copies → 1).

llun (commented):
Sub-agent review round 1 (per `AGENTS.md` → Code Review Loop). Nothing blocking: no downgrades, no major-version moves, no new packages, no new peer warnings, and both load-bearing `resolutions` (postcss, sharp) still collapse the tree to a single copy. The `@radix-ui/react-slot` un-pin was verified against the installed bundle rather than a changelog — `dist/index.mjs` has zero `createContext` hits, and 1.3.3 was already in `main`'s tree transitively through three other Radix packages with a green RSC build.

Three findings, all addressed in the PR description (commit unchanged — none of them require a code change):

1. **MEDIUM — unmentioned `lightningcss` duplication from the `vite` bump.** Inline comment below.
2. **LOW — factual error in the description:** it said `next` hard-pins postcss to an exact `8.4.31`. That was true of an older `next`; `next@16.3.0` declares `postcss: "npm:8.5.23"` (`yarn.lock:6557`). The mechanism the description describes is right, only the number was wrong. Corrected.
3. **LOW — misleading `better-call` note.** "npm `latest` is 1.4.0, so there is nothing newer" is true of the root pin but invites the wrong conclusion: `1.4.0` is the exact version `@better-auth/core@1.6.27` peer-requires, and `better-auth` pulls it as a real dependency, so a nested copy is installed anyway. The root `better-call: 2.0.6` pin is imported by no source file in the repo and its only effect is hoisting an incompatible version over what better-auth wants — which is where the `YN0060` warnings come from. Pre-existing on `main` (identical six warnings), so out of scope here, but the description now says so plainly and flags it as worth cleaning up separately.

package.json
llun: Addressed: no code change is warranted (see the rationale above), and the duplication is now documented in the PR description under **One regression worth naming**. Resolving.

llun (commented):
Sub-agent review round 2 (per `AGENTS.md` → Code Review Loop). Still nothing blocking, and round 1's three fixes were re-verified as correct. This round went after angles round 1 did not cover: changelogs for every bumped runtime dep, lucide icon renames, the yarn 4.18 `--immutable` implication, and the repo's own guard tests.

Four findings, all description-level again — the commit is unchanged (only rebased onto current `main`). Two are inline below; the other two:

- **LOW — the `globals` held-back note skipped a release.** `17.10.0` published 2026-08-11T21:16Z and was also inside the 24h gate at install time (by a few minutes), not just `17.11.0`. Conclusion was right, reason was incomplete. Corrected.
- **NIT — enumeration gap:** the Actions list omitted `anthropics/claude-code-action@v1`, which is also on its current major. Added. Also corrected "two copies of each platform binary" → only the one host binary installs.

Verified clean this round, recorded so a future round need not redo it: all 108 direct deps checked against npm `latest` (the only gaps are the four documented hold-backs); `yarn 4.18.0` has no lockfile-format implication (`__metadata.version: 10` / `cacheKey: 10c0` byte-identical across branches, so `--immutable` in CI and Docker is unaffected); all 115 lucide icon identifiers imported across `app/` and `lib/` resolve in the installed 1.31.0 build, with no dynamic or string-keyed lookup anywhere; `jose` 6.2.5–6.2.8 are verification/decryption-side strictness fixes and this repo only signs (`SignJWT` in `app/api/v1/fitness/apple-maps-token/route.ts`); `pg` 8.23.0's pipelining is opt-in; `kysely`, `marked`, `resend` changes are all inert here; and none of the four guard tests (`clientModuleBoundary`, `tailwindCssVariableSyntax`, `next.config`, `globals.contrast`) is made vacuous — they are source scanners, and `next`/`tailwindcss`/`typescript` did not move.

One adjacent, out-of-scope observation worth knowing: the Dockerfile runs `yarn install --immutable` and then `RUN yarn dedupe` (`Dockerfile:32-34`), so the image build can rewrite the lockfile and resolve a different tree than CI verified. Pre-existing, and neither duplication below is dedupe-collapsible, so it changes nothing here.

package.json
```diff
@@ -21,34 +21,34 @@
...
     "@radix-ui/react-switch": "^1.3.7",
     "@radix-ui/react-tabs": "^1.1.21",
-    "@radix-ui/react-toggle": "^1.1.16",
-    "@radix-ui/react-tooltip": "^1.2.14",
+    "@radix-ui/react-toggle": "^1.1.18",
+    "@radix-ui/react-tooltip": "^1.2.16",
     "@upstash/qstash": "^2.11.3",
     "bcrypt": "^6.0.0",
-    "better-auth": "1.6.26",
+    "better-auth": "1.6.27",
```
llun, line 51: **MEDIUM-LOW — this bump forks `rou3`, and unlike the `lightningcss` fork this one is a runtime dependency.**

The nested `better-call` moves to 1.4.0 with better-auth 1.6.27 and declares `rou3: "npm:^0.9.1"` (`yarn.lock:4167`), while the root-hoisted `better-call@2.0.6` still declares `^0.7.12` (`yarn.lock:4184`). On `main` both wanted `^0.7.12`, so there was a single `rou3@0.7.12` entry; there are now two (`yarn.lock:7621` and `:7628`, resolving 0.7.12 and 0.9.2). Both are on disk (`node_modules/rou3`, `node_modules/better-call/node_modules/rou3`).

Two reasons to name it rather than let it pass silently: `better-auth` is a prod dependency, so this can reach the standalone trace and the runtime image; and better-auth's HTTP router jumps two minor lines as a side effect of what reads as a patch-level bump. Size is irrelevant (92K).

**Not fixing it here.** It is downstream of the root `better-call: 2.0.6` pin — the same unused pin discussed in round 1, which hoists an incompatible version over what better-auth actually wants. Removing that pin would collapse this *and* the six `YN0060` peer warnings, but it is a pre-existing cleanup with its own blast radius, not part of a dependency refresh. Now called out in the PR description alongside the lightningcss fork; the earlier "one regression" framing was wrong.

package.json
```diff
@@ -59,24 +59,24 @@
...
-    "marked": "^18.0.7",
+    "lucide-react": "^1.31.0",
+    "marked": "^18.0.9",
     "mime-types": "^3.0.2",
-    "nanostores": "^1.4.1",
+    "nanostores": "^1.4.2",
     "next": "^16.3.0",
-    "nodemailer": "^9.0.3",
-    "pg": "^8.22.0",
+    "nodemailer": "^9.0.5",
```
llun, line 72: **LOW — 9.0.4/9.0.5 are security-hardening releases and belonged in the Security section, not buried in the bump list.**

9.0.5 fixes header injection via header-key callbacks and DKIM tags, control characters in header values and Message-ID, `List-*` comment escaping, and address normalization between headers and envelope; 9.0.4 fixes unpaired-surrogate handling and escapes backslashes/quotes in the Content-Type `name` parameter.

The address-parsing half is worth checking against this repo, because `lib/services/email/smtp.ts:25` hand-builds an RFC 5322 address by interpolation:

```ts
export const getAddressFromEmail = (email: Email) =>
  typeof email === 'string' ? email : `"${email.name}" <${email.email}>`
```

and feeds the result to `sendMail` as `from`/`to`/`replyTo`. I traced every caller: `from` is always the admin-configured `config.email.serviceFromAddress`, and every `to` is a bare account-email **string** (`deleteActorJob.ts:171`, `sendConfirmationEmail.ts:28`, `sendNotificationAlerts.ts:108`, plus the two account routes) — the object form with a `name` is never constructed from remote input. So no actor display name reaches an address header; display names land in the rendered body and go through the templates' own escaping. Defence-in-depth, not a reachable fix, and the description now says exactly that.

package.json
llun: Addressed in the PR description; no code change warranted for the reasons above. Resolving.

package.json
llun: Addressed in the PR description; no code change warranted for the reasons above. Resolving.

app/api/v1/apps/createApplication.ts
```diff
@@ -1,8 +1,8 @@
 // SafeUrlSchema is a PUBLIC export of @better-auth/core (its `exports` map
```
llun, line 1: **LOW — this PR makes a stale comment staler, so it gets fixed here.**

Line 3 read "`@better-auth/core` is a direct, exact-pinned (**1.6.20**) dependency". The pin has moved twice since; this PR takes it to 1.6.27. It was already stale on `main`, but `CLAUDE.md`/`AGENTS.md` require updating what a change makes stale, and this is the PR that edits the pin.

Fixed by dropping the number rather than restating it — every future core bump would re-stale a hardcoded version, and `lib/services/oauth/matchRedirectUri.ts:2` already says "exact-pinned" without one. The comment's actual claim (a direct exact pin can't drift on a minor/patch bump) is unchanged and still true.

Commit `6f9da6fa`. Full gate re-run on it: prettier clean, lint 0 errors, build compiled, **782 files / 8842 tests passed**.

llun (commented):
Sub-agent review round 3 (per `AGENTS.md` → Code Review Loop). **Verdict: go.** No blocking or high-severity finding; every claim in the twice-revised description was re-checked against the repo, the lockfile, npm and the GitHub advisory DB.

Five low-severity findings. One needed a repo change and got its own commit (inline below); the rest were description accuracy:

- **L1 — Dockerfile enumeration incomplete.** The output stage also copies `data.sqlite` (`Dockerfile:43`), so "copies only …" was wrong. The conclusion (no devDependency reaches the runtime image) is unaffected. Fixed.
- **L2 — the `rou3` paragraph overstated the impact.** Hoisting is the reverse of what I implied: `node_modules/rou3` is **0.9.2** (what better-auth's nested `better-call@1.4.0` actually uses) and `node_modules/better-call/node_modules/rou3` is **0.7.12**, serving only the root pin no source file imports — so `@vercel/nft` will not trace the duplicate into `.next/standalone`. The genuine point is the version jump 0.7.12 → 0.9.2 across rou3's one breaking release (0.8.0). Reworded. Also verified both copies were already ESM-only, so 0.8.0's ESM-only change is a no-op here.
- **L3 — "45 packages" is 44.** `lru-cache` went 3 → 2, not to 1, which is the off-by-one; `@csstools/css-calc` was missing from the list. The −583 net-lines figure is exact (9361 → 8778). Fixed.
- **L5 — optional, not taken:** un-pinning `@radix-ui/react-slot` removes the only in-repo trace of the #1308 hazard, and `^1.3.3` can float. Verified safe (zero `createContext` in both shipped bundles, all descriptors collapse to one 1.3.3 entry) and the `Build` job is the safety net that caught it originally, so I left it as a floating caret rather than re-pinning or adding a comment that would go stale.

Fresh angles this round, all clean: the standalone trace gets **smaller**, and the Dockerfile's `yarn dedupe` is a verified no-op (all 57 multi-version packages have mutually disjoint ranges, so the image tree cannot drift from the CI-tested lockfile); no bumped package changed `engines`, and nothing across 395 packages conflicts with `node: 24.x`; exports-map diffs across all 20 real version pairs show one addition (`set-cookie-parser` gains a `types` condition) and no path resolving to different code; no transitive pin silently holds anything back — every caret floor in `package.json` now equals the installed version exactly; the `sharp` resolution still shares one copy with `next@16.3.0`'s own `^0.35.3`; and no license changed. Lockfile integrity checked too — no new `patch:`/`git:`/`link:` protocols and no entry lost a checksum.

Residual risk, stated plainly: `package.yml` only builds the Docker image on push to `main`, so the image build isn't exercised by this PR. The yarn 4.18.0 bump *is* exercised (every CI job does `corepack enable` + `yarn install --immutable`); the only unverified surface is `node:24-alpine`'s bundled corepack fetching 4.18.0.

app/api/v1/apps/createApplication.ts
llun: Fixed in `6f9da6fa`. Resolving.

llun (commented):
Sub-agent review round 4 (per `AGENTS.md` → Code Review Loop) — confirmation pass over `6f9da6fa` and the third description revision. **Verdict: go.** No blocking or medium finding; nothing new at code level.

Two prose items:

- **LOW — "moved twice" understated it.** The comment was written at 1.6.20 (#1126, `67807bda`); walking `package.json` history the pin went 1.6.20 → 1.6.23 → 1.6.25 → 1.6.26 → 1.6.27 — four moves. Corrected in the description; it only strengthens the case for dropping the number, so the commit message isn't worth amending.
- **LOW, out of scope — a genuinely stale sibling comment.** `app/(nosidebar)/oauth/authorize/authorizeQuery.ts:3` says Better Auth signs `serializeAuthorizationQuery(ctx.query).toString()` and instructs the reader to "keep both the order and seconds-based exp units in sync when upgrading". Against the installed provider that order guidance is obsolete: `signParams` (`@better-auth/oauth-provider/dist/index.mjs:4059-4071`) signs `canonicalizeOAuthQueryParams(params).toString()`, which **sorts** entries over an explicit `ba_param` allow-list plus a `ba_iat` timestamp (`dist/version-C5TN23gF.mjs:5-24`), so param order is no longer signature-relevant. This is **not** a regression from this PR — `main` is already on 1.6.26, the canonicalized scheme predates the bump, and `AuthorizeCard.test.tsx:109-110` already asserts `ba_iat`/`ba_pl`. Left for a separate follow-up rather than widening a dependency refresh into an OAuth-comment audit. (Checked the other two version-stamped comments: `CredentialForm.tsx:23` is still accurate at 1.6.27, and `.github/dependabot.yml:10` is framed as an example and still shape-accurate.)

Re-verified accurate this round: the reworded comment still holds (`@better-auth/core` is caret-free at `package.json:29`; 1.6.27 still declares `"./utils/*"` in its exports map and still exports `SafeUrlSchema` from `dist/utils/redirect-uri.mjs`, so the load-bearing claim survives the bump); the rou3 hoisting description matches the installed tree exactly (better-call 1.4.0 lives under `node_modules/better-auth/` with no nested rou3, so it resolves the hoisted 0.9.2); all five lockfile line citations land on the cited lines; the Dockerfile enumeration now matches lines 40-43 and 51-52; `git ls-files | grep -cE '\.test\.tsx?$'` is exactly 782; the better-auth stack is internally coherent (`@better-fetch/fetch` 1.3.1 and `better-call` 1.4.0 both match core's exact peers, `kysely@0.29.5` satisfies its range); nothing hardcodes a yarn version outside `packageManager`; and no markdown went stale.

All 21 CI checks are green on this head. Closing the review loop here: four rounds, and this one produced no code-level finding.

PR #1417, opened 2026-08-12
minor: assign imported gear from settings, never from Strava

llun:
## Summary

Imported activities are now attributed to gear from **your own settings**, never from the value Strava sends — and no importer creates a gear row any more.

Removed:

- the webhook path's `activity.gear_id` lookup and its `GET /gear/{id}` call
- the archive path's `Activity Gear` column and its `bikes.csv` / `shoes.csv` reader
- `lib/services/fitness-gears/resolveImportedGear.ts` (the find-or-create resolver)
- `fitness_gears.stravaGearId` and its `(actorId, stravaGearId)` unique index (migration + both schema dumps)

**Why.** The failure mode was `kind`. It was guessed from an undocumented `b`/`g` Strava id prefix, or from an optional archive CSV that many exports lack, and `kind` is immutable by design. A bike filed as shoes shows no components card, refuses a frame type and a weight, and its service reminder offers shoe advice — repairable only by deleting the gear, which detaches every activity attributed to it. A decade-old archive additionally populated the shed with every bike the athlete had ever owned, none of which they asked for.

An imported activity is now attributed exactly like an uploaded one: `processFitnessFileJob` assigns the gear whose `defaultSports` claims the parsed sport. That path already existed and already ran for every fitness file; the Strava branch was only ever getting in front of it.

## Default gear editor

The Strava settings page gains a **Default gear** section, matching the design system (`ui_kits/web/FitnessPage.jsx` → `StravaGearDefaults`): one row per activity type, each with a gear picker and a remove control, plus an "Add activity type…" dropdown grouped into Cycling / Running & walking.

It is a **view over `fitness_gears.defaultSports`, not a table of its own** — the same mapping each gear's own form edits, listed the other way round because "which bike do my imported rides go to?" is the question someone connecting an import source is asking. Pointing a sport at a gear is one `PATCH` of that gear; the database's existing steal takes the sport off whoever held it inside the same transaction. That is also why the editor re-reads the whole list after every write: the response carries only the gear that was written, never the one it was taken from.

Both controls are Radix menus rather than `<select>`s, matching `ActivityGearMeta` on the activity page — the repo's existing instant-write gear picker. They commit the moment they change, and on Windows/Linux an arrow key on a *closed* `<select>` moves the selection and fires `change` per keystroke, so keyboard-navigating to the third bike would PATCH the second on the way past. A menu commits only on Enter or click.

Details worth flagging:

- An activity type is only offerable when there is an active gear of its kind to point it at — no dead menu entries.
- Whatever is currently assigned stays in its picker even when retirement or the kind filter would drop it; a picker that cannot represent its own value reads as the assignment having changed on its own.
- Every control is disabled while a write is in flight, and focus is returned to a sensible target once it settles — resolved *after* the re-render, since the control that started the write does not always survive it (removing the last mapped sport unmounts its row; adding the last addable one unmounts the add menu). The restore only fires when focus was actually dropped, so it never yanks a reader out of the credential form above.
- A failed write is announced (`role="alert"`, with every control that can cause it pointing at the message), and a failed initial load offers a retry rather than dead-ending until a page reload.
- `UpdateFitnessGearInput` in `lib/client.ts` is now `Partial<…>`, matching `UpdateGearRequest`'s presence semantics — a caller changing one field must not have to resend a stale copy of the rest.

## Migration

`migrations/20260812000000_drop_fitness_gear_strava_gear_id.js` drops the column and its unique index (index first, in its own `alterTable`, so SQLite's table rebuild does not carry an index naming a departing column). Gear rows a previous import created are left exactly where they are — names, components and attributed activities intact; only the Strava id linking them back is dropped, and nothing reads it. Both reference schema dumps are regenerated from a full migration run (PostgreSQL 17 and SQLite, 148 migrations each).

## Verification

- `yarn run prettier --write .` → `yarn lint` → `yarn build` → `yarn test` all green (781 files, 8795 tests).
- Both schema dumps regenerated canonically; `sqlite3 … < migrations/schema.sqlite.sql` loads clean.
- Browser-checked on a local SQLite instance at 1280px and 375px: rows render with the kind icon, type label, gear picker and remove control; changing a row's gear moves the sport off the previous holder (verified against `GET /api/v1/fitness/gear`); adding and removing a type persist; the "no defaults yet" and "no gear yet" empty states both render. At 375px the picker and its remove button wrap onto their own line so the activity type is not truncated to a single character.


app/(timeline)/fitness/strava/StravaGearDefaultsSection.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  )
+                  if (nextGear) void assignSport(nextGear, sportKey)
+                }}
+              >
+                {options.map((option) => (
+                  <option key={option.id} value={option.id}>
+                    {getOptionLabel(option)}
+                  </option>
+                ))}
+              </Select>
```
llun: **A native `<select>` that writes on `change`, plus `disabled` while saving, persists an assignment the user never chose and drops their keyboard focus.**

On Windows/Linux Chrome and Firefox, arrow keys on a *closed* `<select>` change the value and fire `change` per keystroke. A keyboard user tabs to the Ride picker (Moots), presses Down to reach the third bike: the first Down fires `change` -> PATCH persists Ride on a gear they were only passing through, and `stealDefaultSports` takes it off Moots. In the same tick `setIsSaving(true)` disables the focused element, so the browser blurs it to `<body>` and the remaining Down presses go nowhere.

The repo already has the answer for an instant-write gear picker: `FitnessStatusDetail.tsx:969` uses a Radix `DropdownMenu` with `onSelect`, and its comment calls out this exact difference from a `<select>`. Switching to that pattern removes the intermediate commits entirely and makes `disabled` legitimate again (AGENTS.md endorses disabling a gear picker while its PATCH is in flight).

app/(timeline)/fitness/strava/StravaGearDefaultsSection.tsx
```diff
@@ -0,0 +1,302 @@
...
+          </Link>{' '}
+          to assign it to imported activities.
+        </div>
+      )}
+
+      {gears !== null && gears.length > 0 && rows.length === 0 && (
+        <div className="rounded-lg border border-dashed p-4 text-center text-sm text-muted-foreground">
+          No defaults yet — add an activity type below.
+        </div>
+      )}
```
llun, line 322: **Dead-end empty state when every gear is retired.**

This message renders whenever `gears.length > 0 && rows.length === 0`, but the "Add activity type..." select below it only renders when `addableSportsByKind` is non-empty — and that requires an *active* gear of the kind. An actor whose only gear is retired is told "add an activity type below" with nothing below and no explanation. Reachable in two steps: retire your only bike, then remove its row.

lib/client.ts
```diff
@@ -2549,7 +2549,14 @@ export interface CreateFitnessGearInput {
   notes?: string | null
 }
 
-export type UpdateFitnessGearInput = Omit<CreateFitnessGearInput, 'kind'>
+// Every field is optional, matching `UpdateGearRequest`: the PATCH route uses
+// presence semantics, so a caller changing one field (the Strava page's default
+// gear editor sends only `defaultSports`) must not have to resend the rest —
+// re-sending a stale copy is how a concurrent edit gets silently reverted.
+// `kind` is immutable and absent by design.
```
llun: **The comment overstates what the caller achieves.**

`Partial<>` is right, and it does stop a caller clobbering `name`/`notes`. But this comment says re-sending a stale copy "is how a concurrent edit gets silently reverted", and the Strava editor still read-modify-writes the whole `defaultSports` array from a snapshot it only refreshes after its own writes — so editing the same gear in another tab and then adding a sport here restores what the other tab removed. Narrow, but it is precisely the failure the comment claims is prevented. Reword to what the type actually buys.

lib/jobs/importStravaActivityJob.ts
```diff
@@ -440,53 +438,12 @@ export const importStravaActivityJob = createJobHandle(
...
-      }
-    }
-
+    // Gear is deliberately NOT read from `activity.gear_id`. The athlete's own
+    // shed is the source of truth: `processFitnessFileJob` attributes the
+    // imported file from the gear whose `defaultSports` claims the parsed
+    // sport, which is what the Strava settings page edits. Importing Strava's
+    // value instead created a local gear per Strava id — with a guessed `kind`
+    // that is immutable once wrong — for a shed the owner never asked for.
     const batchFiles = await database.getFitnessFilesByBatchId({ batchId })
```
llun, line 448: The replacement comment is accurate but parked directly above `const batchFiles = ...`, which has nothing to do with gear. Move it where a reader looking for the gear decision would land.

llun (commented):
Sub-agent review round 1 — three reviewers (removal correctness, new UI, invariants/docs/schema). Both schema dumps were independently re-derived and the SQLite one is byte-identical; the migration was run up/down/up on a fresh 148-migration database; the `minor:` bump was checked against the AGENTS.md table and holds. Findings below.

lib/services/strava/archiveReader.test.ts
```diff
@@ -257,137 +198,73 @@ describe('StravaArchiveReader', () => {
...
+          // An empty cell is "absent", not an empty description.
+          activityDescription: undefined,
+          fitnessFilePath: 'activities/1002.tcx',
+          mediaPaths: []
+        }
+      ])
     })
 
-    it('leaves the gear unset when the export has no Activity Gear column', async () => {
+    it('skips rows whose file is not a supported fitness file', async () => {
```
llun: Two coverage points on the replacement tests:

1. **`maxCsvRows` threading is now untested.** The deleted `applies the archive row limit to the gear CSVs` was the only test asserting that `StravaArchiveReader` passes its configured `this._limits.maxCsvRows` into `parseStravaArchiveCsvRows`. `getActivities` still does (`archiveReader.ts:551-553`) but never had its own test, so a refactor could drop the option and the suite would stay green. One `openArchive([...], { maxCsvRows: 2 })` case closes it.

2. **This title overstates what it covers.** Row `1002,Manual entry,` has an *empty* Filename, which trips the `!fitnessFilePath` half of the guard; the `!isSupportedFitnessPath(...)` half is never reached. A third row with an unsupported extension would make the name true.

app/(timeline)/fitness/strava/StravaGearDefaultsSection.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    render(<StravaGearDefaultsSection />)
+
+    fireEvent.change(await screen.findByLabelText('Ride'), {
+      target: { value: 'gear-other' }
+    })
+
+    // Only the gear being pointed at is written: the database takes the sport
+    // off whoever held it inside the same transaction.
+    await waitFor(() =>
```
llun, line 189: **The refetch is asserted as a call count, never as an outcome.** Every test uses `mockResolvedValue`, so both list calls return the same array — the component's whole design premise ("the response carries only the gear that was written, never the one it was taken from", hence the re-read) is untested. A `mockResolvedValueOnce` pair showing the Ride row actually move off Moots and onto Giant is the missing test, and it covers the riskiest path here.

Also: the `gear-other` fixture inherits `defaultSports: ['ride']` from the factory, so the fixture has two gears claiming `ride` — a state the invariant forbids. The expected payload is the same with `defaultSports: []`, so assert against data the server could actually return.

llun (commented):
Round 1, part 2 — findings outside the diff, plus coverage gaps.

llun:
Two more round-1 findings that fall outside the diff, so they have no line to anchor to:

**1. `lib/jobs/processFitnessFileJob.ts:491-496` — stale comment naming the path this PR deleted.** Found independently by two reviewers. It reads:

> Only ever fills a hole: an import that already attributed the activity (Strava's own `gear_id`) and a manual assignment both win

Nothing reads `activity.gear_id` any more, so this block is now the *sole* automatic attribution path — and `AGENTS.md`, `CLAUDE.md`, `REVIEW.md` and `docs/fitness-file-storage.md` all point a reader here as the source of truth. Someone arriving from those docs finds a comment asserting the exact mechanism the docs just said was removed, which reads as the docs being wrong. The rest of the comment (manual assignment wins, jobs re-run) is still correct and stays.

**2. `docs/maintenance.md:374` — residual phrasing.** "the Strava import paths only attribute activities they import themselves" — after this PR they attribute nothing at all; `processFitnessFileJob` does, for imported and uploaded files alike. The sentence's conclusion (the backfill script is still needed) survives, so this is wording, but the parallel sentence in `docs/fitness-file-storage.md` was rewritten and this one was missed.

app/(timeline)/fitness/strava/StravaGearDefaultsSection.tsx
llun: Fixed in c6c4494. Both this row picker and the "Add activity type" control are now Radix `DropdownMenu`s, matching `ActivityGearMeta` in `FitnessStatusDetail.tsx` — the repo's existing instant-write gear picker. A menu commits only on Enter or click, so arrowing past an option no longer PATCHes it, and `disabled` while the write is in flight is now safe as well as correct (AGENTS.md endorses it for exactly this picker). Re-picking the gear already assigned is now a no-op instead of a PATCH, with the same reasoning the activity page spells out. The add control had the identical defect and got the same treatment.

Verified in a real browser: the menu opens, the checked row carries `aria-checked`, picking another bike moves the sport off its previous holder server-side, and the trigger's accessible name carries the value ("Ride: Giant Propel · 0.0 km").

app/(timeline)/fitness/strava/StravaGearDefaultsSection.tsx
llun: Fixed in c6c4494 — see the reply on the row picker thread. Both controls are menus now.

lib/client.ts
llun: Fixed in c6c4494. Reworded to what the type actually buys: `Partial<>` protects the fields a caller OMITS, and the comment now states the limit explicitly — a field built by read-modify-write, as the editor does with the `defaultSports` array, is still last-write-wins against a concurrent edit of that same field.

lib/jobs/importStravaActivityJob.ts
llun: Fixed in c6c4494. Moved and rewritten so it reads as a statement about this job rather than about the line under it, and it now ends with the instruction that matters to the next person here: do not add an attribution step to this job, there is no gear decision in it.

lib/services/strava/archiveReader.test.ts
llun: Both fixed in c6c4494. `applies the configured CSV row limit` drives `maxCsvRows` through `getActivities`, so dropping the option at the call site fails a test again. And the skip test grew a third row (`activities/1003.csv`) so it genuinely exercises the `!isSupportedFitnessPath(...)` half; renamed to match what it now covers.

app/(timeline)/fitness/strava/StravaGearDefaultsSection.test.tsx
llun: Both fixed in c6c4494. The refetch is asserted as an outcome now: the two list reads return different arrays (Moots loses `ride`, Giant gains it), so a component that never refetched — or that patched state locally instead — fails on the trigger text rather than passing on a call count. The `gear-other` fixture is explicitly `defaultSports: []`, so no fixture claims a sport twice. Also added the no-op case (re-picking the assigned gear must not PATCH) and the retired-only empty state, both new behaviour from this round.

llun:
## Round 2 — reviewed the round-1 fixes themselves

The picker rewrite was the risky part of round 1, so this round was pointed at it. Four real defects, all fixed in b717318.

- **A failed save was never announced.** The menu closes, the trigger still reads the old gear, and the message was appended at the bottom of the section with no live region and nothing pointing at it — a screen-reader user could not tell "failed" from "did nothing". Now `role="alert"` with an id, and both triggers carry `aria-describedby`, matching what `ActivityGearMeta` already does.
- **The add menu's group headings were invisible to assistive tech.** Radix's `MenuGroup` renders a bare `role="group"` and its `Label` a bare div with no association, so "Cycling" / "Running & walking" were announced by neither — a regression from the `<optgroup label>` they replaced. Each group now names itself via `aria-labelledby`. Verified in a browser: both groups resolve to their heading text.
- **Focus was dropped to `<body>` on every change, and my round-1 comment claimed the opposite.** `onSelect` runs inside Radix's `flushSync`, so `setIsSaving(true)` lands before the menu closes and Radix's own restore then targets a disabled button. The menu did fix the stray-PATCH-while-arrowing problem; it did not fix this one. The trigger that started the write is now captured and refocused once it settles, and the comment says what the menu does and does not fix. Verified in a browser: `document.activeElement` is the trigger again after the write.
- **A failed gear load collapsed to an empty array**, so the empty-state branch told an actor with a full shed that they had no gear and should go add some. `null` now means "not answered yet" and stays that way on failure.

Plus the `docs/maintenance.md` "never revisits activities already stored" overstatement — `retry-fitness` does re-enqueue processing for a completed file with a `mapError`.

**On the new focus test.** The first version passed with the fix disabled: with an instantly-resolving mock the trigger is re-enabled before Radix's `onCloseAutoFocus` runs, so Radix restores focus by itself and the test could not discriminate. It is rewritten against a deliberately-unresolved write, and I confirmed both directions — it fails with the restore disabled, passes with it.

Gate: prettier → lint → build → test green (781 files, 8801 tests).

llun:
## Round 3 — reviewed the round-2 fixes

Four real defects, all fixed in 0cefde5. The round-2 focus restore was the weak spot, as suspected.

- **Remove was excluded from the focus fix entirely.** `clearSport` passed no target, so removing an activity type — success *or* failure — still dropped focus to `<body>`: the exact regression round 2 said it was fixing. The failure case is the worse one, since the row is still there and the alert has just appeared, leaving the reader on `<body>` with no way to reach it short of tabbing from the top of the document. Same class, second instance: adding the *last* addable sport unmounts the add menu, so the captured trigger was detached and focus fell through the same way.
- **The restore stole focus the reader had moved.** It fired unconditionally, and the section disables itself during a write while the credential form above it does not — so someone who tabbed up and started typing got yanked back out of the field. It now only fires when focus was actually dropped to `<body>`.
- **`aria-describedby` did not cover the remove buttons**, and the round-2 test that claimed to check this drove the failure from Remove and then asserted on the *picker* — so it passed without covering the causing control at all. Both are wired now and the test asserts on both.
- **A comment described a detached-node case the Map it annotates cannot reach.**

Focus targets are now resolved *after* the re-render from a preference order rather than captured before it, so a caller can name a control that may not survive its own write: assigning lands on the sport's own row, removing lands on the add menu the removal just repopulated, falling back to the first remaining row. Verified in a browser — removing a row now lands on the add menu rather than `<body>`.

Also added a retry for a failed initial load. The reviewer flagged it as a judgement call rather than a defect, but it is a genuine dead end: a failed load renders no rows and no add control, so nothing that could clear the error is reachable and the only recovery was a page reload.

**On test rigour:** after round 2 shipped a vacuous test, every new test in this round was checked in both directions by reverting its own fix and confirming it fails. All four discriminate.

Gate: prettier → lint → build → test green (781 files, 8804 tests).

PR #1422, opened 2026-08-13
minor: model recorded devices as a first-class gear kind

llun:
## Summary

The head unit or watch that recorded an activity was rendered as an external manufacturer link on the post chip ("Via: garmin.com") and the activity detail page ("Recorded with"), while the gear page listed only bikes and shoes. Devices now get a `kind: 'device'` row of their own, a **Devices** card on the gear page, and a detail page with an activity history — matching the design system's `GearKit` `DeviceTable`/`DeviceDetailSurface`/`DeviceForm` and `FitnessActivity`'s `FADeviceLine`.

**A device is not a bike, and the model says so.** No components, no default sports, no distance total, no service reminder, no retiring. A head unit records rides and runs alike, so one summed distance would be a number with no meaning; its page reports an activity count and a first-used date instead. `SPORT_KIND` is now typed `Record<SportKey, UserCreatableGearKind>`, so `getSportKeysForKind('device')` answers `[]` with no special case.

**Identity and display are deliberately separate.** `fitness_gears.deviceKey` is derived only from what the file recorded (`name:<normalized deviceName>`, else `mfr:<brand key>`, else no row at all), is UNIQUE with `actorId`, and is never rewritten. `name`, `brand`, `model` and `productUrl` are the owner's to edit. Keying on the name instead would fork a duplicate row the first time someone renamed "Garmin Edge 840" to "the Edge", and every later ride would land on the new row while the earlier ones stayed on the old. `fitness_files.deviceName`/`deviceManufacturer` stay the immutable recorded facts; the new `deviceGearId` is the editable link.

**`resolveDeviceGear` is the sole creator** and never mutates a row it finds. The four import sites that write the device columns call it through a shared best-effort `linkFitnessFileDeviceGear` — attribution is metadata on an activity that has already parsed and stored, so losing it must never fail an import. Unlike `assignFitnessFileGearIfUnset`, that write is an unconditional overwrite: which bike a ride was on is a judgement the owner may correct, while which head unit recorded it is a fact the file states, so re-runs converge. `POST /api/v1/fitness/gear` rejects `device` with a 422 — a hand-made row would carry no `deviceKey` and match no upload.

**Deleting a device releases its `deviceKey`** in the same transaction that soft-deletes it and detaches `deviceGearId`. The unique index covers soft-deleted rows, so without that release the next upload from that device could never create one again.

**Link audience is the owner only.** `/fitness/gear/<id>` is owner-scoped, so everyone else keeps today's branded manufacturer link — labelled with the gear row's name where one exists, since a rename is a rename of the device rather than a private note about it. Both render sites gate on the gear name **or** the recorded label, so a device renamed to something the brand map cannot resolve does not vanish from the post.

**Bug fixed along the way:** devices are filtered out of the activity gear picker *before* its kind narrowing. An unrecognised activity type narrows to nothing and offers every active gear, so filtering afterwards would have put the head unit that recorded the ride in the list of things the ride could have been done on. They are also excluded from the Strava page's default-gear editor, whose empty states all key on `gears.length`.

### Notable implementation detail

`getFitnessGearActivities` orders timestamp-less activities last with a portable `CASE WHEN … IS NULL` expression rather than knex's `nulls: 'last'` option. That option is **broken on SQLite**: knex emulates it as `order by (col is null) desc`, which puts NULLs first *and* silently drops the column's own direction, so the list came back nulls-first and unsorted. Covered by a test that asserts the ordering on both backends.

### Migration

`migrations/20260813000000_add_fitness_device_gear.js` adds `fitness_gears.deviceKey` + `productUrl` with a `(actorId, deviceKey)` unique index, and `fitness_files.deviceGearId` + index. No FK on `deviceGearId`, same rationale as `gearId` beside it. NULLs compare as distinct on both backends, so bikes and shoes are unconstrained. Verified by applying, rolling back and re-applying against both a throwaway SQLite file and a local PostgreSQL 17; both reference dumps regenerated from a clean run.

### Backfill

`scripts/fitness/backfillFitnessDevices.ts` links existing history: it groups the whole history by device identity (so 5,000 rides on one head unit resolve it once), resolves each group to one row, and stamps the link. Dry run by default, `--apply` writes, and only NULL-`deviceGearId` files are selected so a re-run is a no-op.

## Screenshots

Verified in a real browser against a local SQLite database with a seeded bike, three devices and five activities (one deliberately carrying no device information). The images are attached in the session rather than committed — `docs/` is for durable reference documentation only.

- **Gear page** — Bikes / Shoes / **Devices** cards; the Devices card shows `3 recording`, the product-page hostname (`garmin.com`, `wahoofitness.com`) and an activity count per device. No Add button and no retired toggle.
- **Device detail page** — name, `{brand} {model} · recording since {first used}`, product link, **Edit** only, two stat tiles (Activities / First used), and a compact activity list whose rows link to the status page.
- **Edit dialog** — exactly Brand, Model, Nickname, Product page. Saving `the Edge` + a product URL left `deviceKey` at `name:garmin edge 840`, and the rename propagated to the gear list, the post chip and "Recorded with the Edge" on the activity page.
- **Profile feed** — `Via: Garmin Edge 840` / `Garmin Forerunner 265` / `Wahoo ELEMNT BOLT` on the activities that carry a device, nothing on the one that does not.
- **Devices card hidden** — after deleting all three devices via the API the card disappeared entirely, each `deviceKey` was released and every `deviceGearId` detached; re-running the backfill recreated them on the freed keys.

Link audiences and guards checked against the running app:

| Check | Result |
| --- | --- |
| Owner "Via:" / "Recorded with" | `/fitness/gear/<id>` |
| Logged-out viewer, same posts | `https://www.garmin.com` / `https://www.wahoofitness.com` |
| Logged-out viewer opening a device page | redirected to `/auth/signin` |
| Activity gear picker | `Moots Routt 45`, `No gear` — no device |
| `POST /api/v1/fitness/gear` `{kind:'device'}` | 422 |
| `POST …/retire` on a device | 422 `A recording device cannot be retired` |
| `GET`/`POST …/components` on a device | 422 `A recording device has no components` |
| `PATCH` `productUrl` on a bike | 422 `productUrl is only valid for a device` |
| `PATCH` `productUrl: 'javascript:…'` | 422 |
| `PATCH` `defaultSports` on a device | 422 `defaultSports is not valid for a device` |
| `GET …/activities` on a device | 200, rows carrying `statusPublicId` |

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (788 files, 8975 tests; the gear/file DB suites also run green against a local PostgreSQL 17 via `TEST_DATABASE_TYPE=pg`)
- [x] Docs updated: `docs/fitness-file-storage.md`, `docs/maintenance.md`, `AGENTS.md` (Fitness Gear) and `CLAUDE.md`

https://claude.ai/code/session_01Pv94FwR3ZXAUhjEA3JCEY9

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pv94FwR3ZXAUhjEA3JCEY9)_


llun:

Three reviewers fanned out over the data/service layer, the routes/UI, and the tests/script/docs. Everything actionable is addressed in e1c8287. Two of the findings were real bugs, and both were the same shape: **a device ending up outside every rollup.**

### Fixed — correctness

**1. `isPrimary` does not transfer to devices.** The shared countable-activity predicate exists so a ride recorded on two devices counts once toward the bike it was ridden on. But the two files are exactly what tells those two *devices* apart, and the non-primary one is the only trace the second device left. With `isPrimary` in place, the watch that recorded the secondary half of a merged same-ride post got a gear row — `importFitnessFilesJob` links every file *before* the group is collapsed — that reported `0` activities and "Nothing recorded on this device yet" forever. `applyCountableActivityFilter` now takes `includeSecondaryFiles`, set only for the two device queries so the count and the list on the same page can never disagree. The bike rollup still excludes it, which the new test asserts in the same case.

**2. `setFitnessFileGear` accepted a device.** `gearId` answers "what was this ride done on", which a head unit never is. The picker excluding devices was an affordance, not enforcement — `PATCH /api/v1/fitness-files/:id` takes a gear id straight from the body. An activity pointed at a device left *every* rollup at once (device rollups match `deviceGearId`; distance rollups skip devices), `assignFitnessFileGearIfUnset`'s `whereNull` guard then blocked auto-assign forever, and the status meta line named the head unit as the bike. Rejected at the data layer, where ownership already lives.

**3. An activity whose status predates the publicId backfill lost its link.** `statuses.publicId` is nullable and stays null on any instance that has not run the backfill (`docs/maintenance.md` → Public ID Backfill), so `statusPublicId` alone is not the test for "was this posted?" — those rows were silently reclassified as never posted. Now falls back to the percent-encoded URI, the same way `getFileStatusLink` does on the file-management screens.

**4. `getProductUrlHostname` was not a scheme check.** `new URL` parses an authority for non-special schemes, so `javascript://evil.example/%0aalert(1)` yields a perfectly good hostname — and both render sites gate their `href` on this helper. Not reachable through the API (the Zod schema does check the protocol), but `REVIEW.md` requires the allowlist at the sink. Added, with cases for `javascript:`, `data:` and `file:`.

**5. Three smaller ones:** the Strava default-gear editor's third `getFitnessGearList()` refetch missed `toAssignableGears`; `DeviceDetailView` did not reset its list when the gear changed (a "Load more" in that window pages from the wrong offset) and now dedupes on append, since offset pagination over a growing list repeats the boundary row.

### Fixed — tests that could not fail

These are the ones I'd have most wanted caught, since each was guarding something real while asserting nothing:

- The NULL-ordering test's comment named `nulls: 'last'` — the knex option the implementation **deliberately rejects** because it is broken on SQLite (emitted as `order by (col is null) desc`, which puts NULLs first *and* drops the column's direction). A maintainer trusting the comment would have "simplified" straight back into the bug. Comment corrected to say so; the assertion's NULL half only bites under `TEST_DATABASE_TYPE=pg`, which is now stated rather than implied.
- The pagination test seeded five distinct start times, so ordering was already total and the `id` tiebreak was untested. All five now share one timestamp.
- Nothing anywhere produced a non-null `statusPublicId` *from the database* — both other assertions fabricated it in a stub, so a wrong join or alias would have unlinked every device-page row silently. Now exercised with a real seeded status.
- "A viewer who is not the owner" only covered a logged-out reader, where `isOwner` is false for two independent reasons. Now a table with a signed-in non-owner — the case that would actually 404.
- `resolveDeviceGear`'s convergence test stubbed the lookup to succeed unconditionally, so it could not have failed. It now asserts the row is created exactly once across two runs.
- Added the missing test for devices being excluded from the Strava editor.

### Also

`planDeviceBackfill` folds each page into a shared group map instead of accumulating the whole history in memory (the map is keyed on the device, so grouping still spans pages), and `parseArgs` explains itself when handed `--dry-run` rather than answering "Missing value for --dry-run" or eating the next argument. Docs: `docs/features.md` and the `REVIEW.md` checklist were both stale, and the new `isPrimary` asymmetry is documented in `AGENTS.md`, `REVIEW.md` and `docs/fitness-file-storage.md` beside the rule it deviates from.

### Not changed

- **Cursor pagination for the activity list.** A reviewer noted `REVIEW.md` names cursors as the house pattern, and offset pagination can skip a row after a deletion between pages. The dedupe covers the insertion case, and this list is bounded by one gear's history behind an explicit "Load more"; a cursor on `(activityStartTime, id)` is the durable form if it ever becomes a real feed.
- **A test for `backfillFitnessDevicesScript` itself.** Only `parseArgs` and the pure planner are covered — the same shape as `backfillFitnessMovingTime.test.ts`. I verified by reading that the dry-run default is honoured on every path (the single write site sits behind `if (!input.apply) { …; continue }`) and that `printDatabaseBanner()` runs before `getDatabase()`.

### Gate

`prettier` → `lint` → `build` → `test` green: **788 files, 8984 tests**. The gear and fitness-file DB suites also pass against a local PostgreSQL 17. `lib/database/sql/status.test.ts` has 6 pre-existing PG-only failures in unrelated `updateNote` attachment tests (string ids against an integer `mediaId`); I confirmed they reproduce identically on `main` in a clean worktree, so they are not from this PR and CI is unaffected (it pins SQLite).

---

llun:

A reviewer audited the round-1 fixes themselves. It found **a regression in round 1's own fix**, which is exactly why the loop re-runs after every change. Addressed in 1a9cf64.

### Fixed — the round-1 regression

**Dropping `isPrimary` for devices was too broad.** What makes a file secondary is **time overlap, not device identity** — `importFitnessFilesJob` links every parsed file *before* `groupFilesByOverlap` collapses the group, and that grouper never looks at the device columns. So there are two shapes of merged post and round 1 only handled one:

| | round 1 | now |
| --- | --- | --- |
| Two devices, one ride (head unit + watch) | watch counts 1 ✅ | 1 ✅ |
| One device, two files (`.fit` + `.gpx`, or a manual upload beside the Strava sync) | **2 — one ride counted and listed twice** ❌ | 1 ✅ |

A secondary file now counts **only when the primary of its own merged post belongs to a different device** — a correlated `NOT EXISTS` on `(statusId, deviceGearId)`, which is portable and lets the rollup and the activity list share the identical predicate, so a device's count and its page cannot drift apart. A secondary whose primary was since deleted still counts, for the same "only trace left" reason.

The prose I added in round 1 to `AGENTS.md`, `REVIEW.md` and `docs/fitness-file-storage.md` asserted the old rationale ("the non-primary one is the only trace the second device left"), which is false for the same-device merge. All three are corrected to describe the narrower rule.

Verified on SQLite **and** PostgreSQL, and in a browser against both shapes — the two-device merge counts the watch (`Coros Apex 2 → 1`, which was 0 before round 1), the same-device merge counts the ride once (`Wahoo ELEMNT ROAM → 1` despite two linked files).

### Fixed — `DeviceDetailView` request handling

- **`loadMore` had no cancellation.** `GearDetailView` renders the device view at a fixed position with **no `key`**, so navigating between two device pages swaps `gear` on the same instance. A "Load more" fired on device A that settled after B's first page appended A's rows to B's list — and the id dedupe cannot catch it, since the ids are from a different device. Responses are now dropped when the gear id they were requested for is no longer current.
- **The offset now advances by the rows the server returned**, not the deduped list length. Paging from `activities.length` after a dedupe re-requested rows already consumed, and a page that came back entirely duplicates (≥20 activities inserted since the last fetch) left "Load more" stuck at the same offset forever.

Both have tests, along with the previously-missing one for `parseArgs` rejecting `--dry-run`.

### Verified clean by round 2

Worth recording what was checked rather than assumed: the `encodeURIComponent(statusId)` fallback genuinely resolves (`resolveStatusFromPath` decodes the segment, matches `isFullStatusUrl` and looks the URI up directly); bike/shoes rollups are byte-for-byte unchanged by the filter refactor; `setFitnessFileGear`'s device guard returns before any write so the transaction commits empty, and neither caller misreads null as anything but "not found"; `assignFitnessFileGearIfUnset` cannot receive a device because a device can never hold a default sport; `planDeviceBackfill`'s accumulator default is evaluated per call, so there is no shared-state leak; and every other doc claim added by this PR is still true of the code.

### Still open, deliberately

- **Cursor pagination.** Offset pagination can still skip a row if one is *deleted* between pages. The list is bounded by one gear's history behind an explicit "Load more"; a cursor on `(activityStartTime, id)` is the durable form if it ever becomes a real feed.
- **No test for `backfillFitnessDevicesScript` itself** — same shape as `backfillFitnessMovingTime.test.ts`. The dry-run default was verified by reading every path.

### Gate

`prettier` → `lint` → `build` → `test` green: **788 files, 8987 tests**. The gear DB suite also passes against local PostgreSQL 17, which matters here because the new predicate is a correlated subquery.

---

llun:

Reviewed the round-2 fixes. Three more real bugs, **two of them introduced by round 2 itself**. All addressed in b826e86. Every finding below was confirmed by executing the code, not by reading it.

### Fixed — the device predicate was still too weak

Round 2's `NOT EXISTS` asked *"does a primary of my device exist"*. Two holes:

- **A device with several non-primary files and no primary of its own suppresses none of them.** One ride, three files in a single overlap group (`groupFilesByOverlap` has no cardinality limit) — head unit `.fit` (primary), watch `.fit`, watch `.gpx` — and the watch reports **2 activities for one ride**, listing it twice. Exactly the failure round 2 claimed to have fixed.
- **It deferred to a primary that is not itself countable.** `mergeFitnessFilesIntoStatus` deliberately writes the primary `pending` and the secondaries `completed`, so a same-device merged ride counted **0** on the device page until processing finished — and permanently if it ended `failed`.

The rule is now per **ride** per **device** rather than per file: of the countable files sharing a `(statusId, deviceGearId)`, exactly one survives — the primary if that device owns it, otherwise the lowest id. Written as "nothing else beats me", so it needs no window function and reads identically on both backends, and the sibling it defers to must itself be countable. Both new cases have tests, green on SQLite **and** PostgreSQL.

### Fixed — two races I introduced in round 2

- **`isLoadingMore` stuck true.** The new cancellation guard skips `finally` for a superseded request, and the effect never reset the flag. Switching devices mid-`loadMore` left the *new* device's button permanently disabled reading "Loading…".
- **A→B→A leaked rows.** The guard compared gear **ids**, which cannot distinguish "still device A" from "went to B and came back". A request outstanding across that round trip was appended to a list that had since been reset, and its rows were double-counted in the offset, so the next page skipped rows.

Both are now a monotonic generation token, and the effect resets `isLoadingMore` itself.

### Also

`assignFitnessFileGearIfUnset` gained the device guard its sibling `setFitnessFileGear` got in round 1. Unreachable through auto-assign itself (a device can hold no default sport), but `scripts/fitness/importFitnessGear.ts` reaches it through a **name** lookup, and auto-created device rows now share that namespace. The device rollup's `getWhereInBatchSize` reserve went from 3 to 8, since the new predicate binds more values toward SQLite's 999-variable ceiling.

### On test quality

This is the second round where the most valuable finding was about my tests. `DeviceDetailView.test.tsx` **never changed the `gear` prop**, which is precisely why both races shipped. It does now — and I verified each new test genuinely fails by reintroducing each bug in turn:

| Bug reintroduced | Result |
| --- | --- |
| effect's `setIsLoadingMore(false)` removed | `× never leaves the new device stuck loading` |
| generation guard removed | `× drops the previous device's page instead of appending it` |
| both restored | `Tests 16 passed` |

### Verified clean by round 3

The correlated-subquery aliasing trap I asked it to check specifically: **not a bug** — `FROM fitness_files AS sibling` hides the bare name, so the outer reference binds correctly, confirmed by executing the generated SQL on SQLite 3 and PostgreSQL 17.10 with a fixture covering all the merge shapes. Also clean: rollup ≡ activity list (same helper, same flag, the `statuses` join is on a primary key and cannot fan out); NULL `statusId` counts as unmerged; bike/shoes behaviour byte-for-byte unchanged; `fitness_files_status_id_idx` serves the correlated lookup and the subplan is only reached for non-primary rows.

Doc prose in `AGENTS.md`, `REVIEW.md` and `docs/fitness-file-storage.md` was over-stated after round 2 (it described only the "different device owns the primary" case); all three now describe the per-ride-per-device rule and the deliberate non-deferral to an uncountable sibling.

### Gate

`prettier` → `lint` → `build` → `test` green: **788 files, 8992 tests**. Gear DB suite green on local PostgreSQL 17 as well, which matters because this round changed the correlated subquery again. CI was fully green on the previous commit (21/21).

---

llun:
## Sub-agent code review — round 4: clean

**No correctness bugs.** The loop has converged; this is the terminating round.

The reviewer re-verified all five round-3 changes and swept the whole PR independently. The device predicate — the part I got wrong three times — was verified by execution rather than reading:

> A harness drove the **real** `getFitnessGearDeviceRollups` and `getFitnessGearActivities` against in-memory SQLite (loaded from `schema.sqlite.sql`) and a fresh PostgreSQL 17.10 (loaded from `schema.sql`), with expected results computed from an **independently written** statement of the intent rather than a restatement of the SQL. Coverage: all 13 named shapes plus an exhaustive enumeration over a 24-variant space — **4177 shapes, 6996 per-device checks, 0 mismatches on each backend.**

Specific hazards checked and cleared:

- **Exactly one row survives per `(statusId, deviceGearId)`** in every shape, including two primaries (min id wins), all-uncountable (0 rows), and several NULL-`statusId` rows on one device (each counts — they were never merged).
- **Boolean equality is safe across backends.** `isPrimary` is `NOT NULL` in both dumps, so the NULL hazard that would let two rows both survive is unreachable.
- **`sibling.id < …` cannot tie** — `id` is the primary key, values are lowercase UUIDv4, SQLite compares BINARY and PostgreSQL's `varstr_cmp` falls back to `strcmp` on collation equality, so it is a strict total order under any locale.
- **The row the COUNT keeps is the same row the LIST returns** — asserted on row **ids**, not just cardinality.
- **Bike/shoes is byte-for-byte unchanged.**
- `generationRef` is bumped before any await and read synchronously in `loadMore`, so a stale generation can never pair with a fresh gear id; Strict Mode's double-invoke is harmless.
- `.whereNot('kind', 'device')` is NULL-safe because `kind` is `NOT NULL` in both dumps.
- `getWhereInBatchSize(database, 8)` — the generated SQL binds exactly **5** non-chunk values, so 8 is conservative but correct.

It also **regenerated both schema dumps** from the 149 migrations against fresh databases: the SQLite dump is byte-identical to the committed file and the PostgreSQL dump differs only by a leading blank line its own pipeline added. And it independently confirmed the 6 PG-only `status.test.ts` failures reproduce on `main` — unrelated to this PR.

The only findings were two comment inaccuracies, fixed in a3d0a68 (comments only): a cross-reference to `mergeFitnessFilesIntoStatus`, which does not exist — the function is `assignFitnessFilesToImportedStatus` — and a call-site comment still saying `isPrimary` was "relaxed" after round 3 replaced it outright.

---

### Review loop summary

| Round | Correctness bugs | Regressions from the previous round |
| --- | --- | --- |
| 1 | 5 (+ 6 tests that could not fail) | — |
| 2 | 3 | 1 |
| 3 | 3 | 2 |
| 4 | **0** | — |

Every regression landed in one place: the device rollup predicate, rewritten three times. Each version was right for the merge shapes I had in mind and wrong for one I hadn't — first the two-device merge, then the same-device merge, then a device owning several secondaries and no primary. The current rule is stated at the level where all three are the same case: **one row per ride per device**.

The other recurring lesson was about tests rather than code. Round 1 found several that would still pass with the behaviour they named removed; round 3 found that `DeviceDetailView.test.tsx` never changed the `gear` prop, which is exactly why two races shipped. New tests are now verified by reintroducing the bug and watching them fail.

### Final gate

`prettier` → `lint` → `build` → `test` green: **788 files, 8992 tests**. Gear DB suite green on local PostgreSQL 17 as well. CI green on every pushed commit.

---

PR #1426, opened 2026-08-13
none: install Bun ourselves in the oltp sync workflow

llun:
## Problem

`.github/workflows/sync-main-to-oltp.yml` failed twice on 2026-08-12 with `bun: command not found` / exit code 127 inside `anthropics/claude-code-action@v1`:

- https://github.com/llun/activities.next/actions/runs/31638242964
- https://github.com/llun/activities.next/actions/runs/31639748162

## Root cause — the 127 is a symptom, not the cause

The `Install Bun` step showing `outcome=skipped` is a red herring: that entry is setup-bun's cache-save **post-step**, logged under `Post job cleanup`. It is declared `post-if: success()`, so it skipped only because the job had already failed. The action's real install step **failed**:

```
##[start-action display=Install Bun;id=__anthropics_claude-code-action.setup-bun]
Downloading a new version of Bun: https://github.com/oven-sh/bun/releases/download/bun-v1.3.14/bun-linux-x64.zip
socket hang up
Waiting 18 seconds before trying again
socket hang up
Waiting 19 seconds before trying again
##[error]Error: socket hang up
##[end-action id=…setup-bun;outcome=failure;conclusion=failure;duration_ms=37261]
```

A transient failure fetching the Bun release archive from the GitHub release CDN. `@actions/tool-cache` retried 3 times over ~37s and gave up. Every remaining step of the composite action was then skipped, and the action's `Post buffered inline comments` step — an ordinary step carrying `if: always()` — invoked `bun`, which was never installed, producing `command not found` and **exit 127**. That is the failure the run reports, one step removed from the actual cause.

## Options evaluated

| Option | Verdict |
| --- | --- |
| Pin `claude-code-action` to a release/SHA | **Rejected as a fix for the flake.** Not a release regression: the action's step failed on a network error, it was not skipped, and both runs already resolved `@v1` to the same SHA `239e3a73…`. Pinning changes nothing about the download. (It *is* an independent option for the version-drift issue below — see Notes.) |
| Add `oven-sh/setup-bun` before the action | **Rejected, but not for the reason first written here.** An earlier version of this PR claimed the action's `no-cache: true` makes it re-download regardless; that is wrong — `noCache` only gates the Actions cache, and setup-bun short-circuits on an existing matching `~/.bun/bin/bun` first. The real reason: a pre-step reuses the same untunable `@actions/tool-cache` retry policy that failed here, relocating the flake instead of removing it. |
| Retry the whole action step | **Rejected.** Re-running a 30-minute Claude step to recover a 2-second download is the wrong granularity, and the step mutates the merge state it would be retried against. |
| Upstream issue | None exists for this failure mode. [#618](https://github.com/anthropics/claude-code-action/issues/618) tracks the setup-bun dependency generally. |

## Fix

Install Bun in the workflow and hand the action the path via its documented `path_to_bun_executable` input. That input is the off-switch for the fragile step — the action's install carries `if: inputs.path_to_bun_executable == ''`, so it is skipped entirely, and its `Setup Custom Bun Path` step puts our binary on `PATH` for the run *and* for the `if: always()` steps that produced the 127.

The download now runs under a retry budget we control:

- `--retry 5 --retry-delay 10 --retry-all-errors` — that last flag is load-bearing. A mid-transfer connection reset is `CURLE_RECV_ERROR`/`CURLE_PARTIAL_FILE`, **not** one of curl's default retryable errors, so without it the extra attempts would never fire for this exact failure.
- `--retry-max-time 120` + `--max-time 120` + step `timeout-minutes: 6`. Both bounds are needed: curl **resets `--max-time` on every retry**, and a server's `Retry-After` can push a wait past `--retry-delay`, so neither caps the loop on its own.
- The archive is SHA-256-verified against the pinned release before use, and `bun --version` runs so a bad install fails loudly *at the install step* rather than cascading into a confusing 127 later.
- A best-effort check warns when `claude-code-action@v1`'s pinned `bun-version` drifts from `BUN_VERSION`, since `@v1` is a moving tag and nothing else would signal the divergence.

## Verification

- Executed the step body verbatim (extracted from the YAML) under GitHub's exact shell flags (`bash --noprofile --norc -e -o pipefail`):
  - **Happy path** — download, checksum, unzip, and the `bun-linux-x64/bun` layout confirmed against the real 1.3.14 release.
  - **Bad checksum** — exits 1 before `unzip`; the binary can never be executed or handed to the action.
  - **Unusable binary** — exits 126 and never writes `GITHUB_OUTPUT`, so the job fails at the install rather than cascading.
  - **Drift check** — extraction returns `1.3.14` (matching, no warning today); the warning branch emits correct `::warning::` syntax; an unreachable host leaves the step at exit 0.
- **`--max-time` is per-attempt**: reproduced against a stalling local server — `--retry 2 --retry-delay 1 --max-time 3` took 11s, not 3s. `--retry-max-time` verified to bound it (22s vs ~86s), exiting as a clean curl error rather than a hard step kill.
- Confirmed `path_to_bun_executable` exists on the `v1` tag and on the exact SHA (`239e3a73…`) the failing runs resolved, and that no later step of the composite action loses access to `bun`.
- Local gate green: `prettier:check` clean, `yarn lint` 0 errors, `yarn build` passed, `yarn test` **788 files / 8992 tests passed**.

## Review

Three sub-agents reviewed this in parallel (shell/Actions correctness, supply-chain security, adversarial refutation of the diagnosis). 6 findings, all addressed in e0e34a0 and posted as inline comments. The security pass returned no findings and assessed the change as *improving* supply-chain posture: it replaces an unverified transitive download with a version-pinned, SHA-256-verified one that fails closed.

## Notes

- No version bump: `.github/`-only change, committed as `none:`.
- **Version drift is the one new failure mode this introduces.** `BUN_VERSION`/`BUN_SHA256` are pinned by hand against a moving `@v1` tag. It fails loudly inside the action rather than silently, and the new drift warning surfaces it, but pinning the action to a SHA so both move as one unit remains an option if this proves annoying.
- Not addressed here: a failed sync is still silent unless someone reads the Actions tab. It self-recovers on the next push to `main` (the next dispatch carries the newer sha and merges everything), so `oltp` does not fall behind. Worth a follow-up if silent failures become a recurring concern.

### DIFF
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -93,12 +93,64 @@ jobs:
           echo "git merge failed before conflict resolution" >&2
           exit 1
 
+      # anthropics/claude-code-action runs on Bun and installs it itself, in a
+      # bundled `Install Bun` step that pulls the binary straight off the GitHub
+      # release CDN with three closely-spaced retries and no fallback. Two syncs
+      # on 2026-08-12 (runs 31638242964 and 31639748162) died right there with
+      # `socket hang up`: the download gave up after ~37s, every later step of
+      # the composite action was skipped, and the action's always-run post-step
+      # then hit `bun: command not found` and reported exit code 127. That 127 is
+      # the reported failure but not the cause — the cause is the failed
+      # download, and the `Install Bun` entry showing `outcome=skipped` in those
+      # logs is setup-bun's *post*-job step, which always skips.
+      #
+      # Installing Bun here instead and handing the action the path via
+      # `path_to_bun_executable` disables its internal install outright (that
+      # step carries `if: inputs.path_to_bun_executable == ''`) and puts the
+      # download on a retry budget we control. `--retry-all-errors` is the part
+      # that matters: a mid-transfer reset is not one of curl's default
+      # retryable errors.
+      #
+      # BUN_VERSION tracks the `bun-version` that claude-code-action pins in its
+      # own action.yml — check it after an upstream bump, since the action pins
+      # deliberately to dodge Bun regressions. BUN_SHA256 is that release's
+      # `bun-linux-x64.zip` entry from its SHASUMS256.txt and must be updated
+      # together with the version. linux-x64 is hardcoded to match `runs-on`.
+      - name: Install Bun
+        id: bun
+        if: steps.synced.outputs.skip == 'false'
+        shell: bash
+        env:
+          BUN_VERSION: '1.3.14'
+          BUN_SHA256: '951ee2aee855f08595aeec6225226a298d3fea83a3dcd6465c09cbccdf7e848f'
+        run: |
+          bun_root="${RUNNER_TEMP}/bun"
+          mkdir -p "${bun_root}"
+
+          curl \
+            --fail --location --silent --show-error \
+            --retry 5 --retry-delay 10 --retry-all-errors \
+            --connect-timeout 15 --max-time 300 \
+            --output "${bun_root}/bun.zip" \
+            "https://github.com/oven-sh/bun/releases/download/bun-v${BUN_VERSION}/bun-linux-x64.zip"
+
+          echo "${BUN_SHA256}  ${bun_root}/bun.zip" | sha256sum --check --quiet
+
+          unzip -q "${bun_root}/bun.zip" -d "${bun_root}"
+
+          bun_bin="${bun_root}/bun-linux-x64/bun"
+          chmod +x "${bun_bin}"
+          "${bun_bin}" --version
+
+          echo "path=${bun_bin}" >> "${GITHUB_OUTPUT}"
+
       - name: Resolve and validate merge with Claude
         if: steps.synced.outputs.skip == 'false'
         uses: anthropics/claude-code-action@v1
         timeout-minutes: 30
         with:
           anthropic_api_key: ${{ env.CLAUDE_API_KEY }}
+          path_to_bun_executable: ${{ steps.bun.outputs.path }}
           allowed_bots: github-actions
           prompt: |
             Resolve and validate a sync from `main` into `oltp`.


.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
+      #
+      # Installing Bun here instead and handing the action the path via
+      # `path_to_bun_executable` disables its internal install outright (that
+      # step carries `if: inputs.path_to_bun_executable == ''`) and puts the
+      # download on a retry budget we control. `--retry-all-errors` is the part
+      # that matters: a mid-transfer reset is not one of curl's default
+      # retryable errors.
+      #
+      # BUN_VERSION tracks the `bun-version` that claude-code-action pins in its
+      # own action.yml — check it after an upstream bump, since the action pins
```
llun, line 115: **Finding 1 (CONFIRMED, factual error in the original rationale).** The first commit claimed a plain `oven-sh/setup-bun` pre-step "would not have helped" because the action passes `no-cache: true`. That is wrong. In setup-bun `src/action.ts` @ `0c5077e5`, `noCache` is consumed only by `isCacheEnabled()`, which gates `restoreCache`/`saveCache`. An existing `~/.bun/bin/bun` whose version matches short-circuits *before* that, unconditioned on `noCache`:

```js
if (!options.customUrl && existsSync(bunPath)) {
  const existingRevision = await getRevision(bunPath);
  if (existingRevision && isVersionMatch(existingRevision, options.version)) {
    cacheHit = true; // Treat as cache hit to avoid unnecessary network requests
```

The conclusion (don't take that option) still stands, but for a different reason: the pre-step would run the same untunable `@actions/tool-cache` retry policy that failed here. Rationale corrected in the comment and in the follow-up commit message.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
+      # bundled `Install Bun` step that pulls the binary straight off the GitHub
+      # release CDN with three closely-spaced retries and no fallback. Two syncs
+      # on 2026-08-12 (runs 31638242964 and 31639748162) died right there with
+      # `socket hang up`: the download gave up after ~37s, every later step of
+      # the composite action was skipped, and the action's always-run post-step
+      # then hit `bun: command not found` and reported exit code 127. That 127 is
+      # the reported failure but not the cause — the cause is the failed
+      # download, and the `Install Bun` entry showing `outcome=skipped` in those
+      # logs is setup-bun's *post*-job step, which always skips.
+      #
```
llun, line 106: **Finding 2 (CONFIRMED, ×2 agents).** The original comment said the `outcome=skipped` entry "always skips". setup-bun's `action.yml` declares `post: dist/cache-save/index.js` with **`post-if: success()`** — it skipped in those runs only because the job had already failed. Corrected.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
           exit 1
 
+      # anthropics/claude-code-action runs on Bun and installs it itself, in a
+      # bundled `Install Bun` step that pulls the binary straight off the GitHub
+      # release CDN with three closely-spaced retries and no fallback. Two syncs
+      # on 2026-08-12 (runs 31638242964 and 31639748162) died right there with
+      # `socket hang up`: the download gave up after ~37s, every later step of
+      # the composite action was skipped, and the action's always-run post-step
+      # then hit `bun: command not found` and reported exit code 127. That 127 is
+      # the reported failure but not the cause — the cause is the failed
```
llun, line 103: **Finding 3 (CONFIRMED).** The original comment called `Post buffered inline comments` a "post-step". It is an ordinary composite step carrying `if: always() && inputs.classify_inline_comments != 'false'` (action.yml L432), logged *before* `Post job cleanup`. Since the whole point of this comment is to help the next person read that log, naming the wrong entry defeats it. Corrected.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
       - name: Resolve and validate merge with Claude
         if: steps.synced.outputs.skip == 'false'
         uses: anthropics/claude-code-action@v1
         timeout-minutes: 30
         with:
           anthropic_api_key: ${{ env.CLAUDE_API_KEY }}
+          path_to_bun_executable: ${{ steps.bun.outputs.path }}
           allowed_bots: github-actions
           prompt: |
             Resolve and validate a sync from `main` into `oltp`.
```
llun, line 156: **Finding 4 (CONFIRMED, ×2 agents + my own repro).** The retry budget was not bounded. curl resets `--max-time` on each retry — man page: *"If you enable retrying the transfer (--retry) then the maximum time counter is reset each time the transfer is retried"* — which I reproduced against a stalling local server: `--retry 2 --retry-delay 1 --max-time 3` took **11s**, not 3s. One agent additionally showed a `Retry-After: 25` response overrides `--retry-delay 2` *upward*, so the delay is not capped either.

Worst case was 6 × 300s + 5 × 10s ≈ **31 minutes** of the job's 45-minute budget, which would leave `Resolve and validate merge with Claude` (`timeout-minutes: 30`) to be killed by the *job* timeout mid-merge — turning a fast, clear failure into a slow, confusing one.

Fixed with `--retry-max-time 120` (stops curl starting further attempts; verified: 22s instead of ~86s against a stalling server, exiting cleanly as curl error 28 rather than a hard step kill), `--max-time 120`, and `timeout-minutes: 6` as the backstop. Worst case is now ~240s, inside the step timeout.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
+
+          unzip -q "${bun_root}/bun.zip" -d "${bun_root}"
+
+          bun_bin="${bun_root}/bun-linux-x64/bun"
+          chmod +x "${bun_bin}"
+          "${bun_bin}" --version
+
+          echo "path=${bun_bin}" >> "${GITHUB_OUTPUT}"
+
       - name: Resolve and validate merge with Claude
```
llun, line 147: **Finding 5 (new failure mode introduced by this PR).** `BUN_VERSION` is pinned by hand against `@v1`, which is a moving tag. If upstream bumps its `bun-version`, nothing here signals it — and per this PR's own notes a failed sync is silent unless someone reads the Actions tab, so the observable effect is that oltp sync quietly stops working.

Added a best-effort check that reads `bun-version` from `claude-code-action@v1`'s `action.yml` and emits `::warning::` on mismatch. Verified end to end: extraction returns `1.3.14` (matching today, no warning), the warning branch emits correct annotation syntax, and an unreachable host leaves the step at exit 0 under `-e -o pipefail`.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,12 +93,64 @@ jobs:
...
+
+          curl \
+            --fail --location --silent --show-error \
+            --retry 5 --retry-delay 10 --retry-all-errors \
+            --connect-timeout 15 --max-time 300 \
+            --output "${bun_root}/bun.zip" \
+            "https://github.com/oven-sh/bun/releases/download/bun-v${BUN_VERSION}/bun-linux-x64.zip"
+
+          echo "${BUN_SHA256}  ${bun_root}/bun.zip" | sha256sum --check --quiet
+
```
llun, line 138: **Finding 6 (informational, CONFIRMED mechanism).** Hardcoding `bun-linux-x64.zip` drops setup-bun's AVX2 detection — `getAvx2()` reads `/proc/cpuinfo` and selects `-baseline` when AVX2 is absent. Today's runner fleet resolves to the non-baseline build (visible in the failing run's own log), and the failure mode would be loud (SIGILL at the `bun --version` probe, at the install step). Left as-is, but the comment now names the baseline dimension so a future runner migration is one lookup rather than a surprise.

llun (commented):
**Sub-agent code review — round 1** (3 agents in parallel: shell/Actions correctness, supply-chain security, adversarial refutation of the diagnosis).

6 findings, all addressed in e0e34a0. The security pass returned no findings — the pinned checksum was independently verified twice against the release `SHASUMS256.txt`, the checksum was confirmed to fail closed, and the new `run:` block introduces no `${{ }}` interpolation.

Two agents independently reached the same conclusion on the unbounded retry budget, and I reproduced the `--max-time` reset myself against a stalling local server (11s elapsed for 3 attempts at `--max-time 3`, proving it is per-attempt, not per-operation).

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in e0e34a0 — see the finding above for the fix and its verification.

PR #1427, opened 2026-08-13
none: correct the Bun install rationale and bound its retries

llun:
Follow-up to #1426. That PR was squash-merged carrying only its first commit, so the review fixes below never reached `main` — `main` today still has the three factual errors in the comment, `--max-time 300` with no bound, and no drift check.

This PR carries those fixes plus one more found afterwards. No behaviour change to the happy path: the Bun install, checksum, and `path_to_bun_executable` wiring that #1426 shipped are untouched.

## 1. Three factual errors in the comment and commit message

All verified against upstream source. The comment exists to help the next person read the failing run's log, so naming the wrong entries defeats its purpose.

- **A plain `oven-sh/setup-bun` pre-step *would* have avoided the second download.** #1426 claimed `no-cache: true` makes the action re-download unconditionally. Wrong — in setup-bun `src/action.ts` @ `0c5077e5`, `noCache` is consumed only by `isCacheEnabled()`, which gates `restoreCache`/`saveCache`. An existing `~/.bun/bin/bun` of a matching version short-circuits *before* that, unconditioned on `noCache`:

  ```js
  if (!options.customUrl && existsSync(bunPath)) {
    const existingRevision = await getRevision(bunPath);
    if (existingRevision && isVersionMatch(existingRevision, options.version)) {
      cacheHit = true; // Treat as cache hit to avoid unnecessary network requests
  ```

  The conclusion still holds, but the reason is different: a pre-step reuses the same untunable `@actions/tool-cache` retry policy that failed here, relocating the flake instead of removing it.
- **`Post buffered inline comments` is not a post-step.** It is an ordinary composite step carrying `if: always() && inputs.classify_inline_comments != 'false'` (action.yml L432), logged *before* `Post job cleanup`.
- **setup-bun's post-step does not "always skip".** Its `action.yml` declares `post: dist/cache-save/index.js` with **`post-if: success()`** — it skipped in those runs only because the job had already failed.

## 2. The retry budget was not actually bounded

curl **resets `--max-time` on every retry** — man page: *"If you enable retrying the transfer (--retry) then the maximum time counter is reset each time the transfer is retried."* Reproduced against a stalling local server: `--retry 2 --retry-delay 1 --max-time 3` took **11s**, not 3s. A `Retry-After` response header can also push a wait *above* `--retry-delay`.

Worst case was 6 × 300s + 5 × 10s ≈ **31 minutes** of the job's 45-minute budget, which would leave `Resolve and validate merge with Claude` (`timeout-minutes: 30`) to be killed by the *job* timeout mid-merge — turning the fast, clear failure this PR series is about into a slow, confusing one.

Fixed with `--retry-max-time 120` (verified to bound it: 22s vs ~86s against a stalling server, exiting as a clean curl error 28 rather than a hard step kill), `--max-time 120`, and `timeout-minutes: 6` as the backstop. Worst case is now ~240s, inside the step timeout.

## 3. Bun version drift had no signal

`BUN_VERSION`/`BUN_SHA256` are pinned by hand against `@v1`, a moving tag. If upstream bumps its `bun-version`, nothing here says so — and per #1426's own notes a failed sync is silent unless someone reads the Actions tab, so the observable effect is that oltp sync quietly stops working.

Added a best-effort check that reads `bun-version` from `claude-code-action@v1`'s `action.yml` and emits `::warning::` on mismatch. It can never fail the install it annotates.

## 4. …and that check discarded its own result (found reviewing the fix)

The check as first written piped curl into an `awk` that `exit`s on the first match — which sits ~200 lines into a 23,689-byte response. Leaving early closes the pipe while curl is still writing, curl exits 23, and `pipefail` hands the `||` fallback a failed pipeline, so the version it just extracted is thrown away.

This is **not** a "large file" edge case and it is not hypothetical. It is a race on how much curl has flushed before awk leaves, so it fails at a rate that varies run to run rather than never or always — three samples of 20 real fetches each:

| form | sample 1 | sample 2 | sample 3 |
| --- | --- | --- | --- |
| `awk '… { print $2; exit }'` | 6 kept / **14 discarded** | 11 kept / **9 discarded** | 8 kept / **12 discarded** |
| `awk '… && !found { print $2; found = 1 }'` | 20 kept / 0 | 20 kept / 0 | 20 kept / 0 |

Left as-is, the drift check would have shipped looking perfectly healthy and simply never warned — the exact silent-failure mode this PR series exists to remove.

Splitting fetch from parse turned out to be **worse**: the writer moves inside its own command substitution with no `||` of its own, so the same SIGPIPE aborts the step with **141** under `-e`, after the install had already succeeded. That arrangement was tried and discarded.

The fix is to have awk read to EOF. The file is 23KB, so reading all of it costs nothing and removes the class entirely.

## Verification

Every path executed verbatim (step body extracted from the YAML) under GitHub's exact shell flags, `bash --noprofile --norc -e -o pipefail`:

| Path | Result |
| --- | --- |
| Happy path | download + checksum + unzip + `bun-linux-x64/bun` layout, exit 0 |
| Bad checksum | exit 1 before `unzip` — binary can never be executed or handed to the action |
| Unusable binary | exit 126, `GITHUB_OUTPUT` never written |
| Drift check, real fetch | returns `1.3.14` (matching today, no warning), exit 0 |
| Drift check, 60 real fetches | 20/20 extracted `1.3.14` in all three samples (old form: 6, 11, 8 of 20) |
| Drift check, 9.5MB stream | extracts **and keeps** the value, exit 0 |
| Drift check, unreachable host | empty, step survives, exit 0 |
| `--max-time` per-attempt | 11s for 3 attempts at `--max-time 3` — confirms the reset |
| `--retry-max-time` bound | 22s vs ~86s unbounded, clean curl exit 28 |

Local gate green: `prettier:check` clean, `yarn lint` 0 errors, `yarn build` passed, `yarn test` **788 files / 8992 tests passed**.

## Notes

- No version bump: `.github/`-only change, both commits `none:`.
- The hardcoded `bun-linux-x64.zip` is the non-baseline AVX2 build. Today's runner fleet resolves to it (visible in the failing run's own log) and a mismatch would fail loudly at the `bun --version` probe, but the comment now names the baseline dimension so a future runner migration is one lookup rather than a surprise.

### DIFF
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -171,13 +171,14 @@ jobs:
           # look. Never let this check fail the install it is annotating.
           #
           # awk deliberately reads to EOF instead of `exit`ing on the first
-          # match. Leaving early closes the pipe under the writer, and with
-          # `pipefail` that turns a successful extraction into a failed
-          # pipeline the moment the response outgrows the pipe buffer — either
-          # silently discarding the version via the `||` below, or, with the
-          # writer inside its own substitution, killing this step outright on
-          # SIGPIPE (141) after the install had already succeeded. The file is
-          # a few KB; reading all of it costs nothing and removes the class.
+          # match, which sits ~200 lines into a 23KB response. Leaving early
+          # closes the pipe while curl is still writing, curl exits 23, and
+          # `pipefail` then hands the `||` below a failed pipeline — so the
+          # version it just extracted is thrown away. This is a race on how
+          # much curl has flushed, not a size threshold, and it already loses
+          # most of the time: 20 real fetches gave 6 extractions and 14
+          # discards, against 20/20 reading to EOF. Left as `exit` the check
+          # would have looked fine and simply never warned.
           action_bun_version="$(
             curl --fail --location --silent --show-error --max-time 30 \
               'https://raw.githubusercontent.com/anthropics/claude-code-action/v1/action.yml' |
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -169,10 +169,19 @@ jobs:
 
           # Best effort: a mismatch is not fatal, it just means someone should
           # look. Never let this check fail the install it is annotating.
+          #
+          # awk deliberately reads to EOF instead of `exit`ing on the first
+          # match. Leaving early closes the pipe under the writer, and with
+          # `pipefail` that turns a successful extraction into a failed
+          # pipeline the moment the response outgrows the pipe buffer — either
+          # silently discarding the version via the `||` below, or, with the
+          # writer inside its own substitution, killing this step outright on
+          # SIGPIPE (141) after the install had already succeeded. The file is
+          # a few KB; reading all of it costs nothing and removes the class.
           action_bun_version="$(
             curl --fail --location --silent --show-error --max-time 30 \
               'https://raw.githubusercontent.com/anthropics/claude-code-action/v1/action.yml' |
-              awk '/^[[:space:]]*bun-version:/ { print $2; exit }'
+              awk '/^[[:space:]]*bun-version:/ && !found { print $2; found = 1 }'
           )" || action_bun_version=''
 
           if [ -n "${action_bun_version}" ] &&
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -96,12 +96,16 @@ jobs:
       # anthropics/claude-code-action runs on Bun and installs it itself, through
       # a bundled `oven-sh/setup-bun` step that pulls the binary off the GitHub
       # release CDN. Two syncs on 2026-08-12 (runs 31638242964 and 31639748162)
-      # died right there with `socket hang up`: `@actions/tool-cache` gave up
-      # after three attempts in ~37s, every later step of the composite action
-      # was skipped, and its `Post buffered inline comments` step — which is an
-      # ordinary step carrying `if: always()`, not a post-step — then ran `bun`
-      # and hit `command not found`. That is the exit code 127 those runs report;
-      # the failed download is the cause. (The other `Install Bun` entry, logged
+      # died right there: `@actions/tool-cache` gave up after three attempts,
+      # in ~37s and ~31s respectively — the first run failed all three with
+      # `socket hang up`, the second mixed in an `Unexpected HTTP response:
+      # 503`, so this is CDN flakiness generally rather than one error. Every
+      # substantive step of the composite action was then skipped, and its
+      # `Post buffered inline comments` step — which still ran, being an
+      # ordinary step carrying `if: always()`, not a post-step — invoked `bun`
+      # and hit `command not found`. That is the exit code 127 those runs
+      # report; the failed download is the cause. (The other `Install Bun`
+      # entry, logged
       # `outcome=skipped` under `Post job cleanup`, is setup-bun's cache-save
       # post-step. It is declared `post-if: success()`, so it skipped only
       # because the job had already failed — it is not the step that failed.)
@@ -126,9 +130,12 @@ jobs:
       # neither caps the loop: `--retry-max-time` is what stops curl starting
       # further attempts, and `timeout-minutes` is the backstop. Without them a
       # CDN that accepts the connection and then stalls — a different failure
-      # from the `socket hang up` seen here — would burn ~31 minutes of the
-      # job's 45-minute budget and leave the merge step to be killed mid-run,
-      # turning a fast, clear failure into a slow, confusing one.
+      # from the ones seen here — could spend ~31 minutes of the job's
+      # 45-minute budget on this step alone. Whichever way that lands is worse
+      # than failing fast: exhaust the attempts and the sync has burnt the
+      # window to report what the old code reported in ~37s, or succeed on a
+      # late attempt and the merge step inherits too little of the budget to
+      # finish, dying on the job cap rather than its own.
       #
       # BUN_VERSION tracks the `bun-version` that claude-code-action pins in its
       # own action.yml; the drift check below warns when they diverge, because
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -175,10 +175,11 @@ jobs:
           # closes the pipe while curl is still writing, curl exits 23, and
           # `pipefail` then hands the `||` below a failed pipeline — so the
           # version it just extracted is thrown away. This is a race on how
-          # much curl has flushed, not a size threshold, and it already loses
-          # most of the time: 20 real fetches gave 6 extractions and 14
-          # discards, against 20/20 reading to EOF. Left as `exit` the check
-          # would have looked fine and simply never warned.
+          # much curl has flushed, not a size threshold, so it fails at a rate
+          # that varies run to run rather than never or always: three samples
+          # of 20 real fetches discarded 14, 9 and 12 of them, where reading to
+          # EOF gave 20/20 every time. Left as `exit` the check would have
+          # looked healthy and simply gone quiet for whole runs at a time.
           action_bun_version="$(
             curl --fail --location --silent --show-error --max-time 30 \
               'https://raw.githubusercontent.com/anthropics/claude-code-action/v1/action.yml' |
--- a/.github/workflows/sync-main-to-oltp.yml
+++ b/.github/workflows/sync-main-to-oltp.yml
@@ -93,32 +93,55 @@ jobs:
           echo "git merge failed before conflict resolution" >&2
           exit 1
 
-      # anthropics/claude-code-action runs on Bun and installs it itself, in a
-      # bundled `Install Bun` step that pulls the binary straight off the GitHub
-      # release CDN with three closely-spaced retries and no fallback. Two syncs
-      # on 2026-08-12 (runs 31638242964 and 31639748162) died right there with
-      # `socket hang up`: the download gave up after ~37s, every later step of
-      # the composite action was skipped, and the action's always-run post-step
-      # then hit `bun: command not found` and reported exit code 127. That 127 is
-      # the reported failure but not the cause — the cause is the failed
-      # download, and the `Install Bun` entry showing `outcome=skipped` in those
-      # logs is setup-bun's *post*-job step, which always skips.
+      # anthropics/claude-code-action runs on Bun and installs it itself, through
+      # a bundled `oven-sh/setup-bun` step that pulls the binary off the GitHub
+      # release CDN. Two syncs on 2026-08-12 (runs 31638242964 and 31639748162)
+      # died right there with `socket hang up`: `@actions/tool-cache` gave up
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-main-to-oltp.yml
```diff
@@ -144,6 +167,32 @@ jobs:
...
+          # version it just extracted is thrown away. This is a race on how
+          # much curl has flushed, not a size threshold, and it already loses
+          # most of the time: 20 real fetches gave 6 extractions and 14
+          # discards, against 20/20 reading to EOF. Left as `exit` the check
+          # would have looked fine and simply never warned.
+          action_bun_version="$(
+            curl --fail --location --silent --show-error --max-time 30 \
+              'https://raw.githubusercontent.com/anthropics/claude-code-action/v1/action.yml' |
+              awk '/^[[:space:]]*bun-version:/ && !found { print $2; found = 1 }'
+          )" || action_bun_version=''
```
llun, line 194: **Finding 7 (CONFIRMED, and worse than first described).** The drift check's original `awk '… { print $2; exit }'` does not fail only "once the response outgrows the pipe buffer" — upstream's `action.yml` is **23,689 bytes**, well under it, and the match sits ~200 lines in. It is a race on how much curl has flushed before awk leaves, and against the real URL it loses most of the time.

Measured, 20 real fetches each:

| form | result |
| --- | --- |
| `{ print $2; exit }` | **6** extracted, **14** silently discarded |
| `&& !found { … }` | **20** extracted, 0 discarded |

So the check would have shipped looking healthy and simply never warned — the exact silent-failure mode this PR series exists to remove. Fixed by reading to EOF; the 23KB read costs nothing.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -144,6 +167,32 @@ jobs:
 
           echo "path=${bun_bin}" >> "${GITHUB_OUTPUT}"
 
+          # Best effort: a mismatch is not fatal, it just means someone should
+          # look. Never let this check fail the install it is annotating.
+          #
+          # awk deliberately reads to EOF instead of `exit`ing on the first
```
llun, line 180: **Finding 8 (CONFIRMED, comment accuracy).** The fix commit's own comment overstated the failure mode on two counts, which matters because round 1 was specifically about comment accuracy:

1. "the moment the response outgrows the pipe buffer" — wrong model, as measured above.
2. The SIGPIPE-141 clause. With the writer inside `"$( … )" || action_bun_version=''` the `||` always catches a failed pipeline, so 141 was unreachable in the shape ever committed; it only appeared in a split fetch-and-parse arrangement that was tried and discarded.

Comment rewritten to state the measured behaviour instead.

llun (commented):
**Sub-agent review — round 2** (re-reviewing round 1's fixes, since fixes can introduce new problems).

2 findings, both addressed in 742fac4. Round 2 also independently re-verified every factual claim in the comment block against upstream source and found them accurate, and confirmed the timeout arithmetic (~4.6 min worst case against `timeout-minutes: 6`, job's 45 min untouched).

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 742fac4.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 742fac4.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -144,6 +174,33 @@ jobs:
...
+          # match, which sits ~200 lines into a 23KB response. Leaving early
+          # closes the pipe while curl is still writing, curl exits 23, and
+          # `pipefail` then hands the `||` below a failed pipeline — so the
+          # version it just extracted is thrown away. This is a race on how
+          # much curl has flushed, not a size threshold, so it fails at a rate
+          # that varies run to run rather than never or always: three samples
```
llun, line 186: **Finding 9 (CONFIRMED — caught independently by me and by round 3).** The comment cited one 20-fetch sample (6 kept / 14 discarded) and called it "loses most of the time". Re-running gave 11/9 and 8/12. Since the comment itself says this is a race on flush timing, a hard-coded split is network-dependent and a future reader who measures 11/9 reads the comment as wrong.

Fixed in 6ea66f8: it now reports all three samples (14, 9, 12 discards of 20) and describes a rate that varies run to run, against 20/20 for the EOF form in every sample.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,32 +93,62 @@ jobs:
...
-      # `socket hang up`: the download gave up after ~37s, every later step of
-      # the composite action was skipped, and the action's always-run post-step
-      # then hit `bun: command not found` and reported exit code 127. That 127 is
-      # the reported failure but not the cause — the cause is the failed
-      # download, and the `Install Bun` entry showing `outcome=skipped` in those
-      # logs is setup-bun's *post*-job step, which always skips.
+      # anthropics/claude-code-action runs on Bun and installs it itself, through
+      # a bundled `oven-sh/setup-bun` step that pulls the binary off the GitHub
+      # release CDN. Two syncs on 2026-08-12 (runs 31638242964 and 31639748162)
+      # died right there: `@actions/tool-cache` gave up after three attempts,
```
llun, line 99: **Finding 10 (CONFIRMED against the logs).** Two errors in one sentence:

1. `~37s` is true of run 31638242964 only (`duration_ms=37261`). Run 31639748162 was `duration_ms=30702`, ~31s.
2. It attributed `socket hang up` to every attempt of both runs. Run 31639748162's middle attempt was actually `Unexpected HTTP response: 503`:

```
20:51:59 socket hang up            -> Waiting 10 seconds
20:52:10 Unexpected HTTP response: 503 -> Waiting 20 seconds
20:52:30 ##[error]Error: socket hang up
```

Fixed in f422575, and it makes the point better: the flake is the CDN generally, not one specific error.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,32 +93,62 @@ jobs:
...
+      # in ~37s and ~31s respectively — the first run failed all three with
+      # `socket hang up`, the second mixed in an `Unexpected HTTP response:
+      # 503`, so this is CDN flakiness generally rather than one error. Every
+      # substantive step of the composite action was then skipped, and its
```
llun, line 103: **Finding 11 (CONFIRMED).** "every later step of the composite action was skipped" was contradicted by the clause immediately after it, which says `Post buffered inline comments` ran — that is the whole point of the sentence. The log shows `__run` … `__run_7` and `__run_9` skipped but `__run_8` running. Reworded to "every substantive step … was then skipped, and its `Post buffered inline comments` step — which still ran …" in f422575.

.github/workflows/sync-main-to-oltp.yml
```diff
@@ -93,32 +93,62 @@ jobs:
...
+      # retryable errors, so without it the extra attempts never fire for this
+      # exact failure.
+      #
+      # Both time bounds are deliberate. curl resets `--max-time` on every retry
+      # and a server's `Retry-After` can push a wait past `--retry-delay`, so
+      # neither caps the loop: `--retry-max-time` is what stops curl starting
+      # further attempts, and `timeout-minutes` is the backstop. Without them a
+      # CDN that accepts the connection and then stalls — a different failure
+      # from the ones seen here — could spend ~31 minutes of the job's
+      # 45-minute budget on this step alone. Whichever way that lands is worse
```
llun, line 134: **Finding 12 (PLAUSIBLE, accepted).** The stall scenario did not support the consequence drawn from it. If a stalling CDN exhausts all six 300s attempts, `Install Bun` *fails* at ~31 min and `Resolve and validate merge with Claude` never starts — so nothing is "killed mid-run". The merge step only dies on the job cap if the download *succeeds* after burning the budget.

The arithmetic was right, the consequence was stated for a case the sentence didn't describe. f422575 now spells out both landings and why each is worse than failing fast.

llun (commented):
**Sub-agent review — round 3.** 4 findings, all comment-accuracy, all addressed (6ea66f8 and f422575). No functional defects: round 3 executed the step end-to-end in an `ubuntu:24.04` container with mawk under GitHub's exact shell flags and confirmed the drift check exits 0 on all six paths (DNS failure, HTTP 404, connection refused, HTML-200-no-match, match, drift), that `[[:space:]]` works in mawk, that `found` is not an awk special variable, that the `::warning::` continuation emits exactly one line, and that the worst-case step time is ~4.7 min against `timeout-minutes: 6`.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 6ea66f8 / f422575.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 6ea66f8 / f422575.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 6ea66f8 / f422575.

.github/workflows/sync-main-to-oltp.yml
llun: Addressed in 6ea66f8 / f422575.

llun:
**Sub-agent review loop closed — round 4 is clean, no findings.**

Four rounds, 12 findings, all addressed and resolved. Rounds 1–2 found functional defects; rounds 3–4 found only comment-accuracy issues, which mattered here because the comment exists to help the next person read the failing run's log.

Round 4 verified independently, against primary sources:

- **Run-log claims** — durations `37261ms`/`30702ms`; run 31638242964 failed all three attempts with `socket hang up`, run 31639748162 went `socket hang up` → `503` → `socket hang up`. It enumerated the nine composite steps that logged `outcome=skipped` and confirmed `Post buffered inline comments` is the single one that ran, exactly as the comment's exception clause says.
- **`--max-time` resets per retry** — reproduced a third time: 11s for `--retry 2 --retry-delay 1 --max-time 3`, cut to 7s by adding `--retry-max-time 4`. The ~31-minute figure for the old config checks out (6×300 + 5×10 = 1850s), as does the second landing: ~31 min burnt leaves ~14 min of the 45-min job cap, under the merge step's own `timeout-minutes: 30`.
- **The drift-check race** — an independent 20-fetch sample gave 7/20 discards for the `exit` form, outside the 9–14 band the comment quotes, and 20/20 for read-to-EOF. That is the comment's point rather than a contradiction: the figures are given as samples of a rate that varies run to run, which is why the earlier "loses most of the time" wording was removed.
- **Step behaviour end-to-end** under `bash --noprofile --norc -e -o pipefail`: success → exit 0 with `path=` written; version mismatch → one well-formed `::warning::` and exit 0; bad checksum → exit 1; unreachable drift URL → exit 0 with the output still set.
- Checksum, pinned version, `-baseline`/`-aarch64` asset names, the `if: inputs.path_to_bun_executable == ''` gate, and setup-bun's pre-cache `~/.bun/bin/bun` reuse all re-confirmed.

Local gate green (prettier, lint 0 errors, build, 788 files / 8992 tests) and CI green.

PR #1428, opened 2026-08-13
fix: match the design system's gear tables on the fitness gear pages

llun:
## Summary

The gear tables rendered as three plain columns on the card, so the rows read as loose text with a rule between them rather than as a table — the reported "table color and table line look different from the design".

The design system's `ui_kits/web/GearKit.jsx` builds every gear table from a **pinned first column** (`useGKSnapCols`): the data columns scroll under it while the row's subject stays put, on its own opaque surface with a hairline down its right edge (`boxShadow: inset -1px 0 0`). That single pairing is both halves of the report — the column standing off the card is the "table color", the hairline is the "table line" — and it is what the tables were missing.

Alongside it, this picks up the rest of the kit's table treatment:

- `text-xs font-medium` column headers (were `text-sm font-normal`)
- `border-t` row rules instead of `border-b`, so the header carries no rule of its own and the last row no trailing one
- the retired-gear toggle, the components card's replaced toggle, and the product-page link in `GKLink` orange, the last gaining its `ExternalLink` icon
- the components table's Distance column right-aligned and weighted like the kit's, with its dates and action cell no longer wrapping

The chrome lives in two shared constants in `app/(timeline)/fitness/gear/gearUi.ts` (`STICKY_COLUMN`, `STICKY_CLICKABLE_COLUMN`) so the gear list's bikes/shoes/devices tables and the components table on a gear's page cannot drift apart. The gear detail page's components table uses the same `stick`/`dcol` helpers in the kit, so it is converted in the same pass rather than left looking unrelated. The pinned width stays at the call sites, since the kit pins the gear and device tables at 150px and the denser components table at 104px.

### Load-bearing details, all commented at the constants

- **`bg-background` does double duty** — it steps the column off the card, *and* it is the opacity a sticky cell requires, or the columns scrolling underneath show straight through it. It steps opposite ways in the two themes (lighter in light, darker in dark, where `--background` sits below `--card`), so dark reads as a recessed well — a deliberate exception to the dark ramp, because following it would mean `--muted`, which is also the hover colour.
- **The divider is an inset shadow, not a `border-r`** — `border-collapse: collapse` (Tailwind's preflight default) hands border painting to the table, which drops a sticky cell's own right border.
- **A dimmed row dims its cells, never the `<tr>` and never the pinned `<td>`** — `opacity` fades an element's background along with its text, so either takes the pinned column's surface down with it.
- **The hover is the opaque `bg-muted` on both the row and its pinned cell, never `bg-muted/50`** — a translucent hover replaces `bg-background` rather than layering over it, so the cell would turn 50% transparent exactly while the pointer is on the row.

### New token: `--primary-text`

The orange link treatment introduced a contrast regression — `--primary` is 3.37:1 on the card, under the 4.5:1 AA floor for the 12px text these controls use. The design kit doesn't actually ask for that: it keeps `GK_ORANGE` for icons and accents and a darker `GK_ORANGE_TEXT` for link text, and this had flattened the two together.

`--primary-text` restores the split, tuned per theme until it clears 4.5:1 on every surface such a link sits on — including the `--muted` row hover, which is stricter than the card:

| theme | value | vs `--card` | vs `--muted` (hover) |
| --- | --- | --- | --- |
| light | `hsl(24 95% 37%)` | 4.94:1 | 4.70:1 |
| dark | `hsl(24 95% 55%)` | 6.52:1 | 5.15:1 |

Dark goes *lighter*, not darker — on the dark ramp contrast comes from going up. `app/globals.contrast.test.ts` recomputes both tokens from the live values, so collapsing them back together fails the suite. The rest of the app's orange text predates this and should move over when touched.

### Deliberately not carried over

The kit's below-480px **snap scrolling** (`useGKSnapCols` turns the data columns into full-width right-aligned snap panels) is not implemented — it needs a `ResizeObserver` and inline per-column styles, and free horizontal scrolling with a pinned first column is usable as-is. Noting it rather than leaving it implied.

The pattern is documented in AGENTS.md → Fitness Gear.

## Verification

Checked in a real browser (Chromium) against a local SQLite database with mock gear — bikes, shoes, devices, retired gear, and gear components:

- **`/fitness/gear`, light and dark** — pinned first column with the hairline divider, orange toggles and product link. In dark the pinned column reads as a recessed surface and the divider stays visible.
- **`/fitness/gear/<id>`** — components table pins its `Type` column, Distance right-aligned with the wear bar under it.
- **420px wide, retired rows expanded, scrolled horizontally** — the pinned column holds and the data columns pass behind it.
- **Hover, measured not eyeballed** — pinned cell and row both compute to `rgb(245, 245, 245)`, opaque and identical, so there is no seam and nothing ghosts through.
- **Row rules, measured not eyeballed** — a 1px strip sampled along a row boundary is `rgb(229, 229, 229)` (= `--border`) across both the pinned column and the data columns.

Regression tests cover the invariants that fail silently — no type error, no lint error, no visual diff in the other tests. All were confirmed by mutation: dropping `bg-background`, dropping `group`, making the hover translucent, moving a dim onto the row or the pinned cell, and collapsing `--primary-text` back into `--primary` each fail.

Screenshots are in the originating session rather than inline here — the API used to open this PR can't upload images. Happy to attach them if wanted.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (lint: 0 errors; tests: 788 files / 9017 passing)
- [x] Docs updated: the pinned-column pattern and the `text-primary-text` rule are both documented in AGENTS.md → Fitness Gear

https://claude.ai/code/session_01JnPCV9Xss3NV4g7DQn5ehb


app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -200,3 +200,37 @@ export const COMPONENT_TYPE_OPTIONS = [
...
+  'sticky left-0 z-1 min-w-[150px] bg-background shadow-[inset_-1px_0_0_var(--border)]'
+
+/**
+ * Pinned first cell of a body row whose row is clickable. The row's `hover:` has
+ * to be repeated here because this cell paints its own background over the
+ * row's — without it the pinned column stays unlit while the rest of the row
+ * highlights. A table with inert rows uses `STICKY_COLUMN` instead; pairing this
+ * with a row that has no hover of its own lights the first column alone.
+ */
+export const STICKY_CLICKABLE_COLUMN = `${STICKY_COLUMN} group-hover:bg-muted/50`
```
llun: **The pinned cell loses its opacity in exactly the state it needs it.**

`bg-muted/50` is a single `background-color` declaration, so on hover it *replaces* `bg-background` rather than layering over it. Confirmed in the browser — the computed background of the pinned cell on a hovered row is `oklab(0.970081 … / 0.5)`, i.e. 50% alpha.

Two consequences, both real:

1. **Transparency during scroll.** With the table scrolled right and the pointer on a row, the data columns sliding under the pinned cell ghost through it — the precise failure the comment at lines 213–215 says `bg-background` exists to prevent.
2. **Shade mismatch with no scrolling at all.** The `<tr>` paints `muted/50`, then the cell paints another `muted/50` on top, so the first column composites two layers against one for every other column — a visible seam down the pinned column on every hovered row, which is the inverse of the "unlit first column" this constant exists to prevent.

The design kit avoids both by using one **opaque** colour on both surfaces (`GearKit.jsx`: `hover:bg-[color:hsl(0_0%_97%)]` on the row, `group-hover:!bg-[…]` on the cell). Fix is an opaque hover token on both, not a `/50` alpha.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -318,54 +320,87 @@ export const GearComponentsCard: FC<Props> = ({
...
+                  <tr key={component.id} className="border-t">
+                    {/* A replaced component dims its contents rather than the
+                        row: fading the row would take the pinned column's own
+                        background down with it and let the data columns scroll
+                        through. */}
+                    <td
+                      className={cn(
+                        STICKY_COLUMN,
+                        'px-4 py-2.5 align-top font-medium',
+                        isReplaced && 'opacity-60'
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **This does the exact thing the comment three lines above forbids.**

`opacity` applies to the element as a whole, background included. Putting `opacity-60` on the `<td>` that carries `bg-background` is identical in effect to putting it on the `<tr>` — which the comment right above, and the new AGENTS.md paragraph, both rule out. `GearListView.tsx` gets this right with an inner `<div>`; this file does not.

Failure scenario: a bike with a replaced component (`removedAt` set). The components table is `min-w-[720px]`, so any window where the content column is under ~720px puts it in horizontal scroll. Scroll right → the replaced row's pinned "Type" cell is 60% opaque, so Brand/Model/Distance slide visibly under the word "Chain", while the rows above and below stay clean. The `shadow-[inset_-1px_0_0_var(--border)]` divider fades on those rows too, so the table's single vertical rule breaks into gaps wherever a component was replaced.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -318,54 +320,87 @@ export const GearComponentsCard: FC<Props> = ({
...
-                <th className="px-4 py-2 font-normal" />
+              <tr className="text-left text-xs font-medium text-muted-foreground">
+                <th
+                  className={cn(STICKY_HEADER_COLUMN, 'px-4 pb-2 font-medium')}
+                >
+                  Type
+                </th>
+                <th className="px-3 pb-2 font-medium">Brand</th>
+                <th className="px-3 pb-2 font-medium">Model</th>
+                <th className="px-3 pb-2 font-medium">Distance</th>
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **The components table's Distance column is still left-aligned and unweighted.**

The design right-aligns it in both the header and the body and sets the number `font-semibold` (`GearKit.jsx`: `text-right` on the `th`, `whitespace-nowrap text-right` on the `td`, `font-semibold` on the value). Here the `th` and `td` carry no alignment and the span is bare `tabular-nums`.

Pre-existing rather than introduced by this PR, but this is the one numeric column on the page and the sibling gear-list table *did* get `text-right … font-semibold` in this same diff — so the two gear tables now disagree with each other as well as with the kit. Fixing it also means right-aligning `WearBar`, which the design's `GKWearBar` does with `justify-end` for exactly this reason.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -200,3 +200,37 @@ export const COMPONENT_TYPE_OPTIONS = [
...
+// as plain columns on the card, as these were, the rows read as loose text.
+//
+// `bg-background` is load-bearing twice over: it is the design's solid surface
+// standing off the card behind it, and a sticky cell has to be opaque or the
+// columns scrolling underneath show straight through it. The divider is an inset
+// shadow rather than a `border-r` because `border-collapse: collapse` (Tailwind's
+// preflight default for tables) hands border painting to the table, which drops
+// a sticky cell's own right border.
+export const STICKY_COLUMN =
+  'sticky left-0 z-1 min-w-[150px] bg-background shadow-[inset_-1px_0_0_var(--border)]'
```
llun: **One hardcoded `min-w-[150px]` is applied to two tables the design pins at different widths.**

The kit calls `useGKSnapCols(150)` for the gear and device tables but `useGKSnapCols(104)` for the components table. Folding both into a single constant gives the 7-column components table — the most cramped on the page — a 150px "Type" column holding values like "Chain" and "Cassette", taking 46px from columns that need it. The pinned width should be a per-table parameter, not baked in.

Separately, the design's `dcol()` per-column minimums (96/132/108/112/88/84) and its `whitespace-nowrap` on the two date cells were dropped. The table-level `min-w-[720px]` guarantees a *total* but not a *distribution*, so under `table-layout: auto` a long Model value takes width from Added/Removed and "Since beginning" / "Mar 3, 2025" wrap to two lines. The `whitespace-nowrap` at least is worth restoring.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -200,3 +200,37 @@ export const COMPONENT_TYPE_OPTIONS = [
...
+ * highlights. A table with inert rows uses `STICKY_COLUMN` instead; pairing this
+ * with a row that has no hover of its own lights the first column alone.
+ */
+export const STICKY_CLICKABLE_COLUMN = `${STICKY_COLUMN} group-hover:bg-muted/50`
+
+/**
+ * Pinned first cell of the header row. One layer above the body's, so a data
+ * column scrolling underneath passes behind both.
+ */
+export const STICKY_HEADER_COLUMN = `${STICKY_COLUMN} z-2`
```
llun: **This JSDoc asserts a mechanism the code doesn't have, and the `z-2` is inert.**

The header `<th>` and the body `<td>`s are in different rows and never overlap, so their relative z-index is unobservable. What puts each pinned cell above the cells sliding under it is its own positive z-index within its own row, which `z-1` already supplies. (`thead` is not sticky here — if it ever became sticky, the ordering would start to matter and this would earn its keep.)

Also worth noting: the string is `z-1 z-2`, and every call site routes it through `cn` → `twMerge`, which collapses it to `z-2`. Harmless, but a load-bearing comment that misstates the code is worse than no comment, because the next reader preserves it. Either drop the constant and use `STICKY_COLUMN` for headers, or state the real reason.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -200,3 +200,37 @@ export const COMPONENT_TYPE_OPTIONS = [
...
+// on a gear's page so they cannot drift apart.
+//
+// The design system's gear tables (`GearKit.jsx`) pin the first column: the data
+// columns scroll under it while the gear's name stays put, on its own opaque
+// surface with a hairline down its right edge. That pairing is the table's
+// structure — the lighter pinned column is what separates each row's subject
+// from its numbers, and the hairline is the table's only vertical rule. Rendered
+// as plain columns on the card, as these were, the rows read as loose text.
+//
+// `bg-background` is load-bearing twice over: it is the design's solid surface
```
llun: **"the design's lighter surface" is only true in light mode.**

Every one of these tables sits inside a `<Card>`, i.e. on `--card`. Light theme: `--card` 98%, `--background` 100% → the pinned column is lighter, as documented. Dark theme: `--card` 9%, `--background` 3.9% → the pinned column is *darker* and reads as a well punched through the card.

The separation still works in both themes, so this is a wording defect rather than a visual one — but the same sentence is mirrored into AGENTS.md, and under Documentation Maintenance the prose has to describe what the code does. Worth noting too that `app/globals.css` documents the dark ramp as "muted insets/tracks sit above the card", so a dark inset stepping *down* to `--background` is a deliberate exception that should be called out rather than left implied.

---

app/(timeline)/fitness/gear/GearListView.tsx
```diff
@@ -171,7 +197,7 @@ const GearSection: FC<SectionProps> = ({ kind, gears, onAdd }) => {
         <div className="px-4">
           <button
             type="button"
-            className="cursor-pointer text-sm text-muted-foreground hover:text-foreground"
+            className="cursor-pointer text-xs font-medium text-primary hover:underline"
```
llun: **Contrast: this control moved from passing WCAG AA to failing it.**

`--primary` is `hsl(24 95% 46%)` ≈ `#E55F06`. Against the card (`--card: hsl(0 0% 98%)`) that is **3.36:1**; the `text-muted-foreground` it replaced was **4.90:1**. Both this toggle and the product-page link below are 12px normal-weight text, so the 4.5:1 threshold applies — the 3:1 large-text exemption needs ≥18.66px bold / 24px. The toggle also shrank from `text-sm` to `text-xs` in the same edit. Dark mode is fine (5.1:1).

Notably the design kit makes a distinction this implementation flattens: `GK_ORANGE` (`hsl(24 95% 46%)`) for icons and accents, `GK_ORANGE_TEXT` (`hsl(24 95% 40%)`) for link text — the darker shade exists precisely for this. Even that only reaches ~4.31:1 on a 98% card, so matching the kit exactly still lands just under AA.

Flagging rather than fixing unilaterally: the honest fix is a darker link-text token, which is an app-wide design-system change (`text-primary` has ~141 usages, and `Button variant="link"` already ships this colour at these sizes). Raising with the repo owner.

---

llun (commented):

Three reviewers fanned out over the diff (rendering/a11y, project invariants, design fidelity + tests). Findings are inline; two more that couldn't anchor to a diff line, plus what was checked and cleared:

**Additional findings**

- **`GearComponentsCard.tsx:450`** — the "Show N replaced components" toggle was not restyled. It is the same `GKLink` in the design as the gear list's retired toggle, which this PR *did* convert to `text-xs font-medium text-primary hover:underline`. The same control now renders in two styles across two pages.
- **Below-480px snap scrolling was dropped.** `useGKSnapCols` turns the data columns into full-width right-aligned snap panels on narrow viewports; this degrades to plain `overflow-x-auto`. Defensible on its own — it needs a `ResizeObserver` and inline styles, and free scrolling with a pinned first column is still usable — but it should be stated rather than left implied, since it compounds the over-wide pinned column on the components table.

**Verified and NOT a defect** (recorded so it isn't re-litigated)

- **Row rules are not covered by the pinned cell.** The concern was that a collapsed-model border is painted by the table, below positioned descendants, so the hairline's segment across the sticky first column would be hidden. Checked in the browser with a 3× crop across a row boundary: the rule runs continuously across the pinned column. No change needed.
- **`shadow-[inset_-1px_0_0_var(--border)]` is correct Tailwind v4** and emits a real declaration — compiled with the repo's own `@tailwindcss/postcss` 4.3.3, producing `--tw-shadow: inset -1px 0 0 var(--tw-shadow-color, var(--border))`. The `tailwindCssVariableSyntax` guard bans only a bracket whose *entire* content is a bare variable, so this is correctly not an offender.
- **`z-1` / `z-2` are real utilities here** — `.z-1{z-index:1}` and `.z-2{z-index:2}` are present in the compiled output.
- **The `group` ↔ `STICKY_CLICKABLE_COLUMN` pairing is correct** at every call site; no row lights only its first column. Nothing collides with page chrome (`PageHeader` is `z-20`, Radix overlays `z-50`), and no ancestor sets an `overflow` that would neutralise the sticky.
- **`border-t` vs the old `border-b`/`last:border-b-0` is rule-for-rule equivalent**, including the single-row and empty-table cases (both files early-return before rendering a lone bordered header).
- **No existing test became vacuous** — the gear tests are entirely role- and text-based, with no class assertions.
- **`gearUi.ts` is the right home** for these constants (already the shared gear-presentation module, no `'use client'`, so no server/client value-import problem), and there is no table primitive in `lib/components/ui/` that was bypassed.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Fixed in `3112ddd`. Both surfaces now use the opaque `bg-muted` — `hover:bg-muted` on the `<tr>`, `group-hover:bg-muted` on the pinned cell.

Re-measured in the browser after the fix: pinned cell `rgb(245, 245, 245)`, row `rgb(245, 245, 245)` — fully opaque and identical, so both the ghost-through and the seam are gone.

The JSDoc now spells out why the alpha variant is wrong, since `bg-muted/50` is the obvious thing to reach for and reads as harmless. There's also a regression test that fails if the `/50` comes back.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: Fixed in `3112ddd` — correct, and the more embarrassing of the two opacity bugs, since the comment sitting directly above it said not to do this.

The cell now carries only the pinned-column classes and wraps its contents in `<div className={cn(isReplaced && 'opacity-60')}>`, matching `GearListView`. A regression test covers it: moving the dim back onto the row (or the cell) fails.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: Fixed in `3112ddd`. The `th` and `td` are `text-right`, the value is `font-semibold tabular-nums`, and the cell is `whitespace-nowrap` like the design's.

`WearBar` needed `ml-auto` on the bar rather than `justify-end` — it renders a fragment of two block children into the cell, not a flex row, so the caption right-aligns from the cell's own `text-right` and only the fixed-width `w-20` bar needed pushing over. Confirmed in the browser: bar and "replace due" caption both sit under the right-aligned number.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Fixed in `3112ddd`. The width is out of the constant and at the call sites — `min-w-[150px]` on the gear and device tables, `min-w-[104px]` on the components table, matching the kit's `useGKSnapCols(150)` / `useGKSnapCols(104)`. The comment now says why it isn't baked in.

Restored `whitespace-nowrap` on both date cells and the action cell. Left the per-column `dcol()` minimums out deliberately: they were a distribution guarantee for the design's own column set, and with the dates and the action column no longer wrapping the practical failure they prevented is gone. Easy to add later if a long Model value turns out to squeeze something in practice.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Fixed in `3112ddd` by taking the first option — `STICKY_HEADER_COLUMN` is deleted and headers use `STICKY_COLUMN`. You're right that the rationale was fictional: `thead` isn't sticky, so the header and body cells never overlap and the layering was unobservable. Two constants instead of three.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Fixed in `3112ddd`, in both the comment and the mirrored AGENTS.md paragraph. They now say the column "steps off the card" and spell out that it steps opposite ways in the two themes — lighter in light, darker in dark, where `--background` sits below `--card`, so dark reads as a recessed well.

Kept `bg-background` rather than following the dark ramp with `--muted`, and the comment now gives the reason: `--muted` is also the hover colour, so the pinned column would stop responding to hover in dark mode. Called out explicitly as the deliberate exception it is rather than left implied.

---

app/(timeline)/fitness/gear/GearListView.tsx
llun: Fixed in `dc70888`, taking the option this comment pointed at — the design kit's own `GK_ORANGE` / `GK_ORANGE_TEXT` split now exists in the app as `--primary` / `--primary-text`.

`--primary-text` is tuned per theme until it clears 4.5:1 on **every** surface one of these links sits on, which includes the `--muted` row hover — stricter than the card, and the reason the kit's own `hsl(24 95% 40%)` wasn't enough:

| theme | value | vs `--card` | vs `--muted` (hover) |
| --- | --- | --- | --- |
| light | `hsl(24 95% 37%)` | 4.94:1 | 4.70:1 |
| dark | `hsl(24 95% 55%)` | 6.52:1 | 5.15:1 |

Dark goes *lighter*, not darker — on the dark ramp contrast comes from going up, and `--primary` there is only 4.03:1 on the hover, so it was failing in dark on hover too.

`app/globals.contrast.test.ts` now recomputes both tokens' ratios from the live values across all six surfaces × both themes. Mutation-checked: setting `--primary-text` back to `--primary`'s lightness fails six cases, and it reports `3.369:1` on the card — matching the figure in this comment exactly. A second case pins the token to the brand hue at ≥0.8 saturation so contrast can't be "fixed" later by desaturating toward black.

Applied to the retired toggle, the product-page link and the components card's replaced toggle. The rest of the app's orange text still predates this; AGENTS.md says to move it over when touched rather than sweeping it here.

---

llun:

Reviewed the round-1 fixes for regressions plus the whole diff again. Three findings, all now addressed in `dc70888`:

- **Contrast** — the open thread from round 1; resolved with the `--primary-text` token.
- **The components card had no regression test for its half of the round-1 opacity fix**, despite being the file that had the bug. Fair hit: the guards went into `GearListView.test.tsx` only, so folding `isReplaced && 'opacity-60'` back into the `cn(STICKY_COLUMN, …)` call would have passed all 22 existing tests. Mirrored the guards there — pinned cell and header opaque, no hover variant on an inert row, replaced dim on a descendant — and extended the gear list's own assertions to cover `left-0` and the header cell. Mutation-checked.
- **The row-rule-under-the-sticky-cell claim, raised again.** Settled by sampling rather than by eye this time. Screenshotting a 1px strip along a row boundary and decoding it: every pixel across the pinned column (x 327–565) and across the data columns is `rgb(229, 229, 229)` — exactly `--border` (`hsl(0 0% 89.8%)`). The rule paints continuously; no change needed. The reviewer flagged it as an unverified mechanism, having no browser available, and the mechanism is plausible — it just isn't what Chromium does.

**Verified clean in round 2** (no defect): the opaque hover fix, including that `.group:hover .group-hover\:bg-muted` reliably out-specifies `.bg-background` and that `hover:bg-muted` is already the dominant hover token in the codebase (~30 call sites); `ml-auto` on `WearBar` (module-local, one call site, correct for a fixed-width block in a `text-right` cell); deleting `STICKY_HEADER_COLUMN` (no `<thead>` in the repo is sticky, page chrome is `z-20` and still passes above); and the new tests' correctness — the `it.each` titles render cleanly, the tuple typing is right, and none of them can pass vacuously, since jest-dom throws on a null receiver rather than skipping.

Round 3 is running now over the full three-commit diff.

---

llun:

Two real findings, both now fixed in `ee7211f`:

- **The components table's per-row `Replace` button still set `text-primary`** — the exact 3.37:1 foreground the new token was added to replace. `Button variant="ghost"` declares no resting text colour, so `--primary` was it, at `text-sm` (14px), which is normal text under SC 1.4.3. The PR had restyled the `<td>` around this button without touching the button, so the same table was shipping its Replace action at 3.37:1 and its replaced-components toggle at 4.94:1, twenty lines apart. Now `text-primary-text`. Every remaining `text-primary` in these files is on an icon, which is what that token is for.
- **The contrast suite's companion guard didn't guard what its comment claimed.** It asserted hue and HSL *saturation* — but saturation and lightness are independent in HSL, so `hsl(24 95% 15%)`, a near-black brown, keeps the hue, keeps `S` at 0.95, clears 4.5:1 on every light surface, and would have passed the whole suite while no longer reading as the brand colour. Good catch. The bound is on lightness now (within 0.15 of `--primary`), and I mutation-checked it against both that near-black and a near-white dark value — each fails.

One prose correction: the JSDoc, AGENTS.md and a test comment all said pairing `STICKY_CLICKABLE_COLUMN` with an inert row "lights the first column alone". That state isn't reachable from the current code — `group-hover:` on a row with neither `group` nor `hover:` matches nothing, so the wrong constant there is dead weight rather than a visible bug. Lighting the column alone needs a row carrying `group` but no `hover:`. Reworded in all three places; the rule itself is unchanged.

**Verified clean in round 3** (by compiling and recomputing, not by inspection): `text-primary-text` emits `color: var(--primary-text)` with `.text-primary` left intact, and resolves at the use site so the `.dark` override applies; all twelve surface/theme contrast pairs clear 4.5:1, and every ratio quoted in the comments and the PR body was independently recomputed and matched; the comment-stripping in `parseBlock` handles the new multi-line comments; the test's surface list does cover where these links actually render; and no dimming or stacking regression was introduced by moving `opacity-60` onto non-pinned cells.

Round 4 running now.

---

llun:
## Sub-agent code review — rounds 4 and 5

**The code has been clean for two consecutive rounds.** Both rounds found only documentation gaps, now fixed.

**Round 4** — one finding: two new conventions went into AGENTS.md without the matching `REVIEW.md` checklist rows, which AGENTS.md → Documentation Maintenance requires. That matters most for the orange-text rule, which is deliberately forward-looking, and `REVIEW.md` is the checklist this very review loop reads — so without a row there, nothing would catch the next one. Added both, beside the existing `FitnessStatGrid` entry. Fixed in `3c0ddae`, along with a 3.36 vs 3.37 rounding that differed between the token comment and three other mentions (the true value is 3.3695).

**Round 5** — one finding: `CLAUDE.md` needed a key reminder for the orange-text rule too. AGENTS.md makes that entry conditional — "(when agents need it at task start)" — and the reviewer drew the line convincingly:

- **The orange-text rule earns one.** It applies repo-wide, to code nowhere near this PR. An agent adding an orange link in settings reaches for the obvious `text-primary`, nothing in the file being edited hints otherwise, and **no gate catches it** — `app/globals.contrast.test.ts` validates the token *values*, never their *usage*, so a stray `text-primary` on a text node passes lint, build and the entire suite.
- **The gear-table rule does not.** It is confined to two files, both of which import the constants whose own doc comments carry the full rationale — it can't be violated without reading it.

Fixed in `182a063`.

Both rounds independently recomputed every contrast figure quoted in the code comments, the docs and this PR (`--primary` 3.3695:1 on the card, 4.0260:1 on the dark hover; `--primary-text` 4.728–5.155 light, 5.151–7.202 dark) and confirmed all twelve surface/theme pairs clear 4.5:1. Round 4 also verified the design-kit fidelity line by line against `GearKit.jsx` — the pinned-column recipe, the 150/104 split, `border-t` rows, per-cell dimming, the link recipe and the icon-before-hostname product link all match — and confirmed the token's 37% is a justified deviation from the kit's `GK_ORANGE_TEXT` 40%, which is 4.13:1 on `--muted` and would not have cleared AA.

Round 6 is running as the confirming pass.

---

llun:
## Sub-agent code review — round 6: clean

**No new findings.** Six rounds; the last three found only documentation gaps, and this one found nothing. Ending the loop here per AGENTS.md → Code Review Loop ("stop early the moment a clean round produces nothing actionable").

Round 6 re-verified from scratch rather than trusting earlier rounds:

- **Gates** — `prettier --check` clean, `lint` 0 errors, `build` succeeded, full suite 788 files / 9017 tests green. `tsc --noEmit` reports the repo's known pre-existing backlog with **zero** errors in any file this PR touches.
- **Every contrast figure recomputed** from the live tokens: `--primary` 3.37:1 on the light card and 4.03:1 on the dark hover, matching all four places they're quoted; `--primary-text` worst cases 4.73 (light `--muted`) and 5.15 (dark `--muted`). The lightness guard has real headroom (Δ0.09 against a 0.15 bound) and still fails on a collapse back to `--primary`.
- **The built CSS was inspected directly** — `.text-primary-text{color:var(--primary-text)}`, the inset-shadow divider, `z-1`, `sticky` and `group-hover:bg-muted` are all present in `.next/static/chunks/`, independently confirming what earlier rounds settled by compiling.
- **AGENTS.md / REVIEW.md / CLAUDE.md are mutually consistent** on both new rules, with no contradiction between them, and every claim in the CLAUDE.md bullet checks out against the code — including that every surviving `text-primary` in the gear tree is on a Lucide icon.
- **Stacking re-checked**: the `opacity-60` cells create stacking contexts at `z-index: auto`, so the `z-1` pinned cell still paints above them during horizontal scroll — the dim-the-cells approach is load-bearing, not cosmetic.

### CI

Green on `d2f39c8` — all 9 CI jobs, both event runs, plus CodeQL; `mergeable_state: clean`.

The `All Tests` / `CI Success` failures reported against `182a063` were **cancellations, not test failures**: the job log reads `Test shards result: cancelled`, and the two cancelled shards stopped seconds after the `d2f39c8` push created its own runs — concurrency-group supersession. Both shards allowed to finish passed. No test has failed at any point on this branch.

---

PR #1429, opened 2026-08-13
fix: snap the gear components table one column per swipe on a phone

llun:
## Summary

**Rewritten after #1428 landed.** That PR fixed the same report's *colour and border* half — the pinned first column, `STICKY_COLUMN`, the `--primary-text` contrast token, the row spacing — and this branch has been reconciled onto it. What #1428 does **not** change is the mobile scroll behaviour: it keeps `min-w-[720px]`, so a phone still gets one long horizontal scroll. That is all this PR does now.

`min-w-[720px]` forced the wide layout at every width. The new `useGearTableColumns` hook — a port of the design system's `useGKSnapCols` in `ui_kits/web/GearKit.jsx` — removes it and, below 480px, sizes each data column to exactly the width the pinned column leaves over, with `scroll-snap-type: x mandatory` and a `scroll-padding-left` clear of the pin. A swipe then lands on one whole column instead of stranding a row's values halfway across the viewport, which is the behaviour in the screen recording on the original request.

Notes on the mechanism:

- It measures the table's **own scroll container** with a `ResizeObserver`, not the viewport, for the same reason `useCompactActionBar` and `FitnessStatGrid` do: a gear table can sit in a narrow column on a wide window.
- It attaches through a **callback** ref. The table is conditional — a bike with nothing installed renders the empty state — and a ref object assigned later re-runs no effect, so the observer would never reach the table that appears when the first component is added.
- Every cell carries `wrap-anywhere`. A `<td>`'s width is advisory and type/brand/model are free text to 255 chars; a long unbroken value widens its column past the snap interval, and under `x mandatory` a column wider than its interval has a tail the scroller can never rest on. `break-words` is not a substitute — `overflow-wrap: break-word` does not reduce min-content contribution, and the difference is invisible in Chromium while breaking elsewhere.
- The threshold does not promise no horizontal scrolling. The per-cell minimums add up to ~724px, so between 480px and that width the table still scrolls as one block — now with the type column pinned through it, which is the half that was broken. Both docs say so.

**Reconciled away from my earlier version, in favour of what #1428 shipped:** the card stays a `Card` on `bg-card` (its pinned column deliberately steps off that surface), the header keeps its "N installed" count, replaced rows keep their wear bar, and row actions stay ghost buttons on `text-primary-text` rather than the bare links I had used. Two of those were judgement calls I had flagged as debatable at the time; #1428 decided them, so they follow it.

## Screenshots

Verified in Chromium against a local SQLite instance with a seeded bike, eight installed components and one replaced chain, in light and dark at 1280px and 390px. I can't attach binaries to a PR body from this environment; measured at 390px after the merge:

| | |
| --- | --- |
| scroller width | 356px |
| pinned "Type" column | 104px, `position: sticky` |
| each data column | 252px (= 356 − 104) |
| `scroll-snap-type` | `x mandatory` |
| `scroll-padding-left` | 104px |
| snap positions after three swipes | 252 / 504 / 756 — exact column multiples |
| `scrollWidth` with a 49-char unbroken model | 1616 = 104 + 6 × 252, i.e. still exactly on snap boundaries |

## Checklist

- [x] Docs updated: AGENTS.md → Fitness Gear and CLAUDE.md carry the snap rule, scoped so they don't restate or contradict #1428's pinning rules

https://claude.ai/code/session_01DMveDSCvjYwGgTK1ZBGoZD


app/(timeline)/fitness/gear/useGearTableColumns.ts
```diff
@@ -0,0 +1,122 @@
...
+ * jsdom): the full table is the honest default, and the first client layout
+ * corrects it.
+ */
+export const useGearTableColumns = (
+  pinnedColumnWidth: number
+): GearTableColumns => {
+  const ref = useRef<HTMLDivElement>(null)
+  const [width, setWidth] = useState(0)
+
+  useIsomorphicLayoutEffect(() => {
```
llun, line 77: The layout effect has `[]` deps, so it only ever sees the `ref.current` that existed at the first commit — but the node it wants is conditional. `GearComponentsCard` renders the scroller only when `visible.length > 0`, so a bike with nothing installed (a brand-new bike, or one whose every component has been replaced) mounts the empty-state paragraph instead, `ref.current` is null, and the effect returns early.

When the table then appears — the first component is added, or "Show N replaced components" is clicked — no `ResizeObserver` is ever attached. `width` stays 0, `isSnapping` stays false, and the table renders unpinned and unsnapped until a page reload, which is exactly the behaviour this PR exists to remove.

A ref object can't fix this, because assigning `ref.current` doesn't re-run anything. Use a callback ref (`useState<HTMLDivElement | null>` + the node in the dep array) so attach/detach follows the node's mount.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -310,68 +333,131 @@ export const GearComponentsCard: FC<Props> = ({
...
+                      className={cn(CELL, 'whitespace-nowrap text-right')}
+                      style={dataColumnStyle(84)}
+                    >
                       {isReplaced ? (
                         <Button
                           size="sm"
                           type="button"
-                          variant="ghost"
-                          className="text-destructive"
+                          variant="link"
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: Replace/Delete went from `variant="ghost"` `size="sm"` (`h-8`, so a 32px target) to `variant="link"` with `h-auto p-0`, which collapses the tap target to the text box — about 16px tall with no padding at all. That is under the WCAG 2.2 SC 2.5.8 minimum of 24×24 CSS px, and it lands hardest on exactly the phone layout this PR is about.

It matters more than usual for Delete specifically: a missed tap that lands outside the 16px strip also blurs the button, and the `onBlur` handler disarms the pending confirmation — so the fumble silently resets the two-click delete rather than doing nothing.

The design's `GKLink` is genuinely a bare text button, so the *look* is right; the fix is to buy back the hit area without moving the text — padding plus a compensating negative margin, which is invisible but real to the pointer.

---

AGENTS.md
```diff
@@ -417,6 +417,30 @@ it; there is no legacy shape left to copy.
   component whose window is open on that side, since it cannot be placed inside
   `[addedAt, removedAt)`. A gear total may therefore exceed the sum of its
   components.
+- **Every gear table pins its first column and snaps one data column per swipe
```
llun: This says **every** gear table pins its first column and snaps below 480px, and cites "the bike's name" as a pinned label — but `GearListView.tsx:113` and `:213` still render `min-w-[520px]` tables with no hook, which is the same forced-horizontal-scroll pattern the rule bans two sentences later. The mirrored bullet in `CLAUDE.md:47` overclaims the same way.

Documenting an aspiration as if it were the state of the code is the drift these files exist to prevent: the next agent reads the rule, sees the list tables, and can't tell whether they're an oversight or an exception.

The repo already has the honest pattern for a partial migration — the **Fitness Stat Strips** section names which strips went through `FitnessStatGrid` and then says outright which two have *not* been migrated yet. Either migrate the list tables in this PR or scope the rule the same way.

---

llun (commented):
Sub-agent review, round 1 (high effort). Three findings, all confirmed against the code rather than inferred from the diff: a conditional-mount bug in the new hook that silently disables the whole feature on an empty bike, a tap target that drops below the WCAG minimum on the layout this PR targets, and a docs bullet that overclaims a partial migration. Details inline.

---

app/(timeline)/fitness/gear/useGearTableColumns.ts
llun: Fixed in 862657d — the hook now holds the node in state and hands back a callback ref, so attach and detach follow the node's own mount rather than the card's.

Reproduced and confirmed in a browser, not just in a test: on an empty bike at 390px the card starts with zero `<table>` elements, and after adding the first component through the UI — no reload — the table comes up with `scroll-snap-type: x mandatory`, the Type column `position: sticky` at 104px, and data columns at 252px. Before the fix that path produced an unpinned, unsnapped table.

Two regression tests cover it: one mounts the scroller after an empty state and asserts it snaps, one unmounts it and asserts the observer is dropped.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: Fixed in 862657d. The actions keep the design's bare text link, but now carry `-mx-2 -my-1.5 px-2 py-1.5`: the padding is the hit area, the negative margin cancels it visually, so the label sits exactly where it did and the row height is unchanged.

Measured at 390px after the change: **63×28** for the Replace button, up from ~16px tall. Row heights are the same as before (45px for a plain row, 66–69px for one carrying a wear bar), so nothing moved to pay for it.

---

AGENTS.md
llun: Agreed — the rule was written as if the migration were finished. Scoped it in 862657d rather than migrating the list tables, since this PR is about the components page and porting three more tables would widen it well past what was asked for.

AGENTS.md now says "the components table pins its first column…" and carries a second bullet saying outright that the gear list's bikes/shoes/device tables are **not** migrated, that they still carry `min-w-[520px]`, and that adopting the hook there is a straight port of `useGKSnapCols(150)` when someone wants it. CLAUDE.md's bullet was narrowed the same way. That follows the Fitness Stat Strips precedent you point at.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -182,31 +215,36 @@ export const GearComponentsCard: FC<Props> = ({
...
-          onClick={() => setIsFormOpen(true)}
-        >
+    // The design system's gear surfaces sit on the page background rather than
+    // the card grey, which is what the stat tiles above this one use — the two
+    // are meant to read as different depths, not the same slab twice.
+    // `overflow-hidden` is what keeps the table inside the rounded corners: with
+    // no replaced components the scroller is the section's last child, and the
+    // pinned cell's opaque square background painted straight over the
+    // bottom-left arc (as did the horizontal scrollbar, across both corners).
+    <section className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 2 — fixed in 060b132.** The `rounded-2xl` section had no `overflow-hidden`, and when a bike has no replaced components the `overflow-x-auto` scroller is its last child with only `pb-1` beneath it. The pinned cell's opaque square `bg-background` painted straight over the bottom-left corner arc, and a classic horizontal scrollbar crossed both bottom corners. The old `Card`'s `py-4` had hidden this by keeping the table clear of the radius.

Reproduced at 390px on a bike with one component and nothing replaced — the corner rendered square inside its own border — and confirmed fixed on the same page at 4× zoom, arc and border intact. Clipping the section does not affect the pinned column: the sticky cell resolves against the inner `overflow-x-auto` scrollport, which is nearer.

---

app/(timeline)/fitness/gear/useGearTableColumns.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+  })
+
+  it('floors a snapped column so a long distance still fits on one line', () => {
+    render(<Probe />)
+    act(() => deliver?.(240))
+
+    expect(styleOf('data').width).toBe('160px')
+  })
+
+  it('keeps the last measured width when the table reports zero', () => {
```
llun, line 127: **Round 2 — fixed in 060b132.** The previous version of this test (`deliver(700)` then `deliver(0)`, assert "wide") asserted nothing: `isSnapping` is false at width 0 whether or not the guard exists, so deleting the guard left all tests green.

It now delivers 390 first, asserts the table is snapping, *then* delivers 0 and asserts the snapped layout survives — which is the case the guard exists for, since a `display: none` ancestor reports 0 for everything. Verified the other way too: with `if (measured === 0) return` deleted, this test fails and the other eight still pass.

The comment above the guard also claimed it was there to stop columns pinning at the 160px floor. That was never its job — it retains the last good width — so it says that instead.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -310,68 +348,135 @@ export const GearComponentsCard: FC<Props> = ({
...
+                    <td
+                      className={cn(CELL, PINNED_CELL, 'z-[1]')}
+                      style={pinnedColumnStyle}
+                    >
+                      <span className={cn('text-[13px] font-medium', dim)}>
+                        {component.componentType}
+                      </span>
                     </td>
-                    <td className="px-4 py-2 text-muted-foreground">
+                    {/* Brand keeps the row's own colour and model steps back
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 2 — comment added in 060b132.** Brand losing `text-muted-foreground` while the adjacent Model kept it was deliberate, not a class dropped in the rewrite: `GearKit.jsx`'s row leaves Brand on the inherited colour and steps Model back (`hsl(0 0% 30%)` there, `text-muted-foreground` here, since there is no token at 30% and inventing one for a single cell is worse than the small shift).

That said, "deliberate but unexplained" and "accidental" look identical to the next reader, which is the actual defect — so the pair now carries a comment saying why the two columns differ.

---

llun (commented):
Sub-agent review, round 2 (high effort) on `15ff725`+`862657d`. No correctness bugs in the hook this time — the callback-ref fix holds, the snap arithmetic lines up at both ends of the scroll range, and hiding the wear bar on replaced components matches what `serviceReminders.ts` already does with `removedAt`. Three lower-severity issues, all fixed in 060b132: a rounded-corner bleed that only appears when a bike has no replaced components, a test that asserted nothing, and an unexplained colour step. Details inline.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -32,6 +32,31 @@ interface Props {
...
+  'sticky left-0 bg-background shadow-[inset_-1px_0_0_var(--border)]'
+
+const CELL = 'px-3 py-2.5 align-top'
+
+/** Width the pinned "Type" column keeps, and the snap offset that follows it. */
+const TYPE_COLUMN_WIDTH = 104
+
+/**
+ * A row action is the design system's bare text link, but a 16px line of text
+ * is a target well under the WCAG 2.2 minimum of 24x24 on the phone layout this
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 3 — fixed in the latest push.** Type, Distance and the action cell set `text-[13px]` on an inner span/button, so the `<td>` itself stayed at the inherited 16px and kept a 24px strut in its line box, while Brand/Model/Added/Removed (sized on the `<td>`) got 19.5px. With `align-top` that puts the same row's text on two different lines.

Measured on the first row at 1280px before the fix: Brand/Model/Added glyph tops at y=505, but "Frame" and "5,932.0 km" at y=508 — a 3px step within one row — and the row 45px tall instead of 40.5px.

After moving the size onto the shared `CELL`: every cell reports 13px/19.5px, all five tops at y=505, row height 40.5px. The wear bar (`text-[11px]`) and the row actions (`text-xs`) still set their own sizes on the elements that own them, which is fine — they are not what establishes the cell's strut.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -32,6 +32,31 @@ interface Props {
 
 const SERVICE_REMINDER_KM_OPTIONS = [1000, 3000, 5000, 8000, 12000]
 
+/**
+ * The pinned column has to paint an opaque background of its own — the rows are
+ * transparent over the card, so without one the data columns scroll visibly
+ * underneath it — and draws its right-hand rule as an inset shadow rather than
+ * a border, because a border would scroll with the cell's box on some engines.
+ */
+const PINNED_CELL =
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 3 — fixed in the latest push.** `TYPE_COLUMN_WIDTH` is used for two things that have to agree: the pin's width and the scroller's `scroll-padding-left`. The picker only offers `COMPONENT_TYPE_OPTIONS`, but the API doesn't — `gearRequests.ts:133` validates `componentType` as `z.string().trim().min(1).max(VARCHAR_MAX)` — so an API client can store a long unbroken type. `max-width` on a `<td>` is advisory under `table-layout: auto`, so such a word widens the real column past 104 and every snap then lands with the incoming column's left edge tucked under the pin.

`break-words` on the pinned cell makes the constant authoritative again. Verified by inserting a 37-character single-word type: every pinned cell still measures exactly 104px and the long value wraps over four lines.

---

app/(timeline)/fitness/gear/useGearTableColumns.ts
```diff
@@ -0,0 +1,128 @@
...
+      if (measured === 0) return
+      setWidth(measured)
+    }
+
+    // Measure here rather than leaving it to the observer's first delivery,
+    // which lands after the frame has painted — one flash of the wide table on
+    // every phone.
+    measure(element.clientWidth)
+
+    const observer = new ResizeObserver(([entry]) => {
```
llun: **Round 3 — fixed in the latest push.** The pre-paint measure read `element.clientWidth` (integer, padding box) while the observer read `entry.contentRect.width` (fractional, content box). Two consequences: a table sitting a fraction either side of 480 renders wide on first paint and reflows a frame later, and any future reuse on a scroller with horizontal padding disagrees with itself by that padding.

Both paths now read `clientWidth`, which is also the more honest measure here — it is the scrollport the columns are snapped against. The test's `ResizeObserver` stub sets `clientWidth` on the observed element before firing, since jsdom lays nothing out and reports 0.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx
```diff
@@ -151,8 +151,8 @@ describe('GearDetailView', () => {
     expect(screen.getByText('Activities')).toBeInTheDocument()
     expect(screen.getByText('1204')).toBeInTheDocument()
     expect(screen.getByText('Components installed')).toBeInTheDocument()
-    // The replaced component counts toward neither the tile nor the header.
-    expect(screen.getByText('2 installed')).toBeInTheDocument()
+    // The replaced component does not count toward the tile.
+    expect(screen.getByText('2')).toBeInTheDocument()
     // "Distance" is both a stat tile and the components table's column header.
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx:156]: **Round 3 — fixed in the latest push.** Dropping the header's "2 installed" left this assertion as a bare `getByText('2')`, which matches any exact-text "2" anywhere in the tree. It would pass on a coincidence and fail on an unrelated change, which is worse than not asserting at all — the point of the test is that a replaced component is excluded from the tile.

It now resolves the tile from its own label and asserts on that subtree.

---

llun (commented):
Sub-agent review, round 3 (high effort) on `060b132`. Nothing wrong with the mechanism this round — the reviewer independently checked that the per-cell `minWidth` values sum to 724 and so replace the deleted `min-w-[720px]` exactly, that `min-width` (not the advisory `max-width`) is what forces the snapped table's width, and that the dimmed data cells paint below the `z-[1]` pinned cell as the "dim contents, not the row" rationale claims. Four findings, all fixed in 00212fb: a 3px baseline step within each row, an unbounded pinned column that can break the snap offset, two disagreeing width measures, and a brittle assertion I left behind in round 1. Details inline.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -32,6 +32,38 @@ interface Props {
...
+  'sticky left-0 break-words bg-background shadow-[inset_-1px_0_0_var(--border)]'
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 4 — fixed in the latest push.** Round 3's `break-words` was the wrong tool. Tailwind's `break-words` is `overflow-wrap: break-word`, which per CSS Text 3 does *not* reduce a box's min-content contribution — only `anywhere` (and `word-break: break-all`) does. So the fix rested on Chromium happening to honour the explicit `width` over min-content, which is not something to depend on across engines, and the phone browsers this table is for are exactly where it would break.

Now `wrap-anywhere` (`overflow-wrap: anywhere`). Re-verified with the same 37-character single-word type: computed `overflow-wrap: anywhere`, every pinned cell still exactly 104px.

Good catch on my own round-3 fix — the browser check I did passed for the wrong reason.

---

app/(timeline)/fitness/gear/useGearTableColumns.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A snapped data column never gets narrower than this even on a very small
+ * screen: "35,670.2 km" plus its wear bar has to fit on one line.
+ */
+const MIN_SNAP_COLUMN_WIDTH = 160
+
+// The measurement has to land before the browser paints, or the table renders
+// wide for a frame and then reflows. `useLayoutEffect` warns when React renders
+// on the server, where there is nothing to measure anyway.
```
llun, line 32: **Round 4 — fixed in the latest push.** The floor was 160 with a comment claiming the distance and its wear bar fit in it, which was not true. Measured the wear line in the browser: 158px (80px bar + 8px gap + a caption as long as "of 12,000 km"), plus the cell's 24px of horizontal padding — so 182px is what the content actually needs.

It matters because the line is `whitespace-nowrap` in a `justify-end` flex row: it does not clip, it spills out of the *start* edge and under the pinned column.

Raised to 184 with the measurement written down. It only engages below a 288px scroller — at a 320px viewport the scroller is 286px and the column comes out at 182px on its own — so this changes nothing on any real phone and stops the very narrow case from overlapping.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -182,31 +222,36 @@ export const GearComponentsCard: FC<Props> = ({
...
+    // `overflow-hidden` is what keeps the table inside the rounded corners: with
+    // no replaced components the scroller is the section's last child, and the
+    // pinned cell's opaque square background painted straight over the
+    // bottom-left arc (as did the horizontal scrollbar, across both corners).
+    <section className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
+      <div className="flex items-center justify-between gap-3 px-5 py-4">
+        <div className="flex min-w-0 items-center gap-2">
+          <Wrench className="size-4 shrink-0 text-primary" />
+          {/* The installed count is not repeated here: the stat grid above the
+              card already carries it as "Components installed". */}
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 4 — fixed in the latest push.** The section was `bg-background/80` while `PINNED_CELL` has to be opaque `bg-background`, so wherever `body`'s fixed radial tints (`globals.css:183`) fall behind the card, 20% of the tint reaches every column *except* the pinned one.

I sampled pixels either side of the boundary in both themes at 1280px and could not make it visible on this page — the card sits outside both gradients' 60%/55% stops, so page background, pinned cell and data cell all read identical (255,255,255 light / 9,9,9 vs 10,10,10 dark). So this is latent rather than a bug you can see today.

Fixing it anyway, because the mismatch is structural: the pinned cell's opacity is not negotiable, so the surface around it has to match. The section is opaque now, with a comment saying why it departs from the `bg-background/80` the other design-system sections use. The stat tiles above stay `bg-card`, so the depth relationship the design asks for is unchanged.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:1]: **Round 4 — fixed in the latest push.** Fair, and it was the gap that mattered most: this suite never stubbed `ResizeObserver`, so in every one of its tests the hook early-returned and not one assertion touched the pinning or snapping this PR is about. The hook's own suite tested the hook in isolation; nothing tested that its output reached the right cells.

Added a `responsive columns` block that stubs the observer and covers the wiring: that every cell in a column (header included) gets the same width — a header and body that disagree is how a pinned column ends up straddling the boundary it exists to hold — that below the threshold the type column is `sticky`, the data columns carry `scroll-snap-align: start`, and the scroller gets `x mandatory` with a 104px `scroll-padding-left`; and that a wide table keeps the pin but drops the snap styles entirely.

---

llun (commented):
Sub-agent review, round 4 (high effort) on `00212fb`. This round caught a bad fix from round 3 — `break-words` doesn't do what that commit claimed, and my browser check had passed for the wrong reason — plus a snap floor whose comment didn't match its number, a translucent/opaque mismatch around the pinned column, and the fact that this card's own suite never stubbed `ResizeObserver`, so none of its tests exercised the pinning at all. All four addressed in the latest push. Details inline.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -32,6 +32,38 @@ interface Props {
...
+ * the wear bar and the row actions set their own smaller sizes on the elements
+ * that own them.
+ */
+const CELL = 'px-3 py-2.5 align-top text-[13px]'
+
+/** Width the pinned "Type" column keeps, and the snap offset that follows it. */
+const TYPE_COLUMN_WIDTH = 104
+
+/**
+ * A row action is the design system's bare text link, but a 16px line of text
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx]: **Round 5 — fixed in the latest push.** Round 4 gave `wrap-anywhere` to the pinned cell and stopped there, but the reasoning applies to every column: a `<td>`'s width is advisory, and brand and model are free text up to 255 chars (`optionalText` in `gearRequests.ts`). A long unbroken model name widens its column past the snap interval, and under `scroll-snap-type: x mandatory` the scroller cannot come to rest between snap points — so the tail of that cell is not reachable on a phone at all. Worse than the pinned-column case, which merely misaligned the offset.

`wrap-anywhere` moved onto the shared `CELL`, so every body cell gets it. The `whitespace-nowrap` cells (distance, added, removed) are unaffected — their content is formatted and bounded.

Verified at 390px with a 49-character unbroken model: every Model cell still exactly 252px, and the table's `scrollWidth` is 1616 = 104 + 6 × 252, so every column boundary is still exactly on a snap point.

---

AGENTS.md
```diff
@@ -417,6 +417,41 @@ it; there is no legacy shape left to copy.
...
+  other columns scroll visibly through it. A row action stays the design's bare
+  text link but keeps padding with a compensating negative margin, so its tap
+  target clears the WCAG 2.2 24x24 minimum without moving the label.
+- **The gear list's own tables are NOT migrated to that hook yet.** The bikes,
+  shoes and device tables in `GearListView.tsx` still carry `min-w-[520px]` and
+  scroll as one block; the design system drives them from the same hook
+  (`useGKSnapCols(150)`), so adopting it there is a straight port, but until
+  someone does it the rule above describes the components table only. Do not
+  read it as a claim about the whole tree.
+- **The components card is a `bg-background/80` section, not a `Card`.** The
```
llun: **Round 5 — fixed in the latest push.** Round 4 made the section opaque and left both docs saying `bg-background/80`. CLAUDE.md was worse than stale — it contradicted itself inside one bullet, saying the pinned cell "paints its own opaque `bg-background`" and then that the card "is a `rounded-2xl border bg-background/80` section". An agent following that restores the `/80` and reintroduces exactly the mismatch the commit removed.

Both now say `bg-background`, and both give the reason (the pinned column's opacity is not negotiable, so the surface around it has to match) plus the `overflow-hidden` requirement, which was also undocumented.

Worth noting this is the second docs finding in five rounds, both from the same cause: updating code in a round and not re-reading the rule it was written from.

---

llun (commented):
Sub-agent review, round 5 (high effort) on `6fe8dcf`. The hook is clean this round — the reviewer verified the callback-ref attach, the zero-width guard (fails without it), the stable ref identity across re-renders, the snap targets landing exactly on `maxScrollLeft`, the absence of a ResizeObserver feedback loop, and that `wrap-anywhere` and the inset-shadow arbitrary value both compile on Tailwind 4.3. Two findings, both fixed in the latest push: round 4's wrap fix stopped one column short of where the same reasoning applies, and round 4's opacity change left both docs stale. Details inline.

---

app/(timeline)/fitness/gear/useGearTableColumns.ts
```diff
@@ -0,0 +1,137 @@
...
+} from 'react'
+
+/**
+ * Below this the gear tables stop laying their data columns out side by side.
+ * Each one instead takes the whole width the pinned first column leaves over
+ * and becomes a scroll-snap panel, so a swipe moves exactly one column into
+ * view — the alternative on a phone is a table wide enough that every column is
+ * half-legible and the row it belongs to has scrolled off the left edge.
+ */
+export const GEAR_TABLE_SNAP_WIDTH = 480
```
llun, line 18: **Round 6 — documented in the latest push, deliberately not "fixed".** The finding is correct: the per-cell minimums sum to ~724px, so a container between the 480px threshold and 724px gets no snapping *and* still overflows — an ordinary one-block horizontal scroll. My AGENTS.md wording implied that band didn't exist.

Not changing the threshold, though. 480 is the design system's own number (`useGKSnapCols`), and raising it to 724 would put a 600px column — a tablet, or this page in a narrow window — into one-column-per-swipe, which is a worse trade than a 124px scroll that shows six columns at once. And the band is not what the bug report was about: before this PR the type column scrolled away at *every* width, including that one. It is pinned throughout now, so the row keeps its label while you scroll.

So both docs now state the three regimes explicitly — snap below 480, pinned-but-scrolling between 480 and ~724, no scroll above — rather than implying horizontal scrolling was eliminated. Happy to raise the threshold instead if you'd rather; it is a one-constant change.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -310,68 +366,128 @@ export const GearComponentsCard: FC<Props> = ({
...
-                      isReplaced && 'opacity-60'
-                    )}
-                  >
-                    <td className="px-4 py-2 font-medium">
-                      {component.componentType}
+                  <tr key={component.id} className="border-t">
+                    <td
+                      className={cn(CELL, PINNED_CELL, 'z-[1]')}
+                      style={pinnedColumnStyle}
+                    >
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:402]: **Round 6 — partly a misread, partly fair; addressed in the latest push.**

AGENTS.md scopes the rule to the pinned cell — "A pinned cell paints its own opaque `bg-background` **and** takes the dim of a replaced row on its contents, never on the `<tr>`" — and that is exactly what the code does: the pinned cell dims its inner span, and no `<tr>` is dimmed anywhere. The data cells dimming themselves does not violate it.

But CLAUDE.md's mirror said "a replaced row dims its cell **contents**, never the `<tr>`", which does read as a claim about every cell, and you're right that the invariant's real precondition — *a cell that paints a background* — was left implicit in both. That is the part worth fixing, since the hook is now shared and a second pinned column is a plausible next step.

CLAUDE.md now names the pinned cell and says outright that cells painting no background may dim themselves, and the `dim` constant carries a comment stating the condition and what to do when a cell gains a background.

Not restructuring the five data cells to wrap their text in spans: that is markup noise for no behavioural difference today, and the comment is what stops the next person getting it wrong.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx
```diff
@@ -150,9 +150,12 @@ describe('GearDetailView', () => {
...
-    // The replaced component counts toward neither the tile nor the header.
-    expect(screen.getByText('2 installed')).toBeInTheDocument()
+    // Scoped to the tile: a bare getByText('2') matches any exact "2" in the
+    // page and would pass, or start failing, for reasons that have nothing to
+    // do with the installed count.
+    const installedTile = screen.getByText('Components installed').parentElement
+    // The replaced component does not count toward the tile.
+    expect(installedTile).toHaveTextContent('2')
     // "Distance" is both a stat tile and the components table's column header.
     expect(screen.getAllByText('Distance')).toHaveLength(2)
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx:157]: **Round 6 — fixed in the latest push.** Right, and this is the third time this one assertion has been wrong: `toHaveTextContent('2')` against the tile is a substring match on "Components installed2", so 12, 20, 21 and 25 all satisfy it. Round 3 scoped it to the tile and I called it fixed without noticing the matcher was still loose.

It now targets the value element rather than the tile and matches it whole (`/^2$/`), so only the count itself can satisfy it.

---

llun (commented):
Sub-agent review, round 6 (high effort) on `22739f9`. The mechanism is settled — this round the reviewer confirmed that `MIN_SNAP_COLUMN_WIDTH` is precisely what triggers the CSS Scroll Snap "snap area larger than snapport" escape, so the last column stays reachable under `x mandatory`, and found nothing wrong with the hook itself. Three findings, all addressed: one real gap in what the docs promised about the 480–724px band (documented rather than changed, reasoning inline), one half-misread about the dim rule that still exposed a genuinely implicit precondition, and a test matcher that was loose for the third round running. Details inline.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -310,68 +366,134 @@ export const GearComponentsCard: FC<Props> = ({
...
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:402]: **Round 7 — investigated and rejected; no change.** The claim was that under `border-collapse: collapse` the row separators paint with the table while the sticky pinned cells paint their opaque background above them, so every horizontal divider is interrupted across the pinned column's 104px, and that the same paint order makes the pinned cell's inset-shadow rule unreliable. It was flagged plausible because the reviewer had no browser; I do, so I measured it.

Sampling a 6px strip centred on a row boundary, the divider reads exactly `rgb(229, 229, 229)` — the `--border` token — at every x, both inside the pinned column and outside it, at rest and with the table scrolled two columns in. Not interrupted anywhere.

A 4× zoom of the pin boundary on a scrolled table confirms it visually: the vertical rule runs the full height of the row, and the dividers above and below run straight through the boundary in both directions.

Chromium paints a collapsed table border above the sticky cell's background here, so the premise doesn't hold. Not adding `inset 0 1px 0` or moving to `border-separate` — both would be a fix for something that isn't happening, and `border-separate` in particular would change the row rhythm.

Two earlier readings that looked anomalous while I was checking this turned out to be my own measurement error: the clip had landed under the fixed bottom nav.

---

llun (commented):
Sub-agent review, round 7 (high effort) on `cef71db`. **Clean round** — one finding, explicitly marked plausible-not-confirmed, and it does not reproduce in a browser (evidence inline). Nothing else surfaced.

The reviewer verified a good deal along the way with no findings: the snap positions tile exactly and stay reachable at both ends of the scroll range, the `MIN_SNAP_COLUMN_WIDTH` floor is what triggers the Scroll Snap "snap area larger than snapport" escape so the last column can still be reached under `x mandatory`, there is no `ResizeObserver` feedback loop (a horizontal scrollbar changes `clientHeight`, not `clientWidth`), the `z-[1]`/`z-[2]` values are what an `opacity < 1` sibling requires to stay below the pinned cell, and dropping the header's installed count is safe because `GearComponentsCard` renders only for `kind === 'bike'` — exactly when the "Components installed" tile renders.

That closes the loop at 7 rounds of a 20-round cap. Across rounds 1–6, 17 findings were raised and addressed; every thread is replied to and resolved. Two of them were defects in my own earlier fixes, which is the loop working as intended.

---

PR #1430, opened 2026-08-13
minor: customize navigation

llun:
## Summary

Implements **Customize navigation** from the design system project (`ui_kits/web/nav-customize.html`, `NavCustomizeKit.jsx`, `NavCustomizeShowcase.jsx`).

Every account uses a different slice of the product — a runner lives in Fitness, plenty of people never open it. Any nav item that is not pinned can now be hidden in place, tucked under a **More** group at the end of the navigation, and restored with one click. All items can be reordered. Admins can additionally turn optional features off for the whole instance.

Precedence, as in the design: **admin feature off** (gone for everyone) > **user hidden** (under More) > **defaults**. An uncustomized account sees exactly the navigation it saw before.

### What it looks like

- **Sidebar** — hovering a row reveals a `⋯` menu: *Move up · Move down · Hide from navigation*. Pinned rows (Timeline, Search, Lists, Notifications, Account, Settings) offer the moves but say *Always shown* instead. Hidden items collapse into a **More** group with a count; hovering a row there reveals an eye that puts it back. The group renders only when something is hidden. The unread badge steps aside while a row is hovered so it never sits under the `⋯`.
- **Rail** — same order and hidden set, reachable through a More flyout with the same restore control.
- **Mobile** — the bar takes the first four shown items; the rest, the Profile entry, and a muted **Hidden** section live in the More sheet. Per the decision on the open question, the unread badge moves onto the **More** tab whenever Notifications is not in the bar, so the count is never lost. This does change the default bar from Home/Search/Messages/Alerts to Home/Search/Explore/Messages, which anyone can undo by moving Notifications up.
- **Settings → Navigation** (new tab, after Preferences) — one list of every item with drag or arrow-key reordering, a switch per row, chips for *under More* and *off for this server*, and a reset. It saves as you change it, and shares one store with the sidebar, so on a wide screen the sidebar beside it updates live.
- **Admin → Instance → Optional features** — a switch each for Fitness, Explore and Messages.

### How it is built

- **`lib/services/navigation/navPreferences.ts`** is the whole domain model: the nav item ids, which are pinned, which belong to an optional feature, and the pure order/split/move algebra. It is dependency-free so the client surfaces and the server route that validates writes share it without crossing the server/client module boundary.
- **`lib/components/layout/nav-items.ts`** stays the visual registry (icon, label, blurb, href). A new nav item is one registry entry plus its id.
- Preferences are two lists on the existing `actors.settings` JSON column — **no migration**, so no schema dumps to regenerate. Empty lists mean the shipped defaults, which is how a reset is stored too, so the account follows the navigation as it changes in later releases.
- Writes send a **full snapshot**, single-flight with one trailing save, so rapid edits resolve to last-write-wins and a whole drag costs one request. The server drops ids it does not know and refuses to hide a pinned item, so a hand-edited preference cannot strand someone.
- Feature switches reuse the existing `server_settings` registry (group `instance`), so no new API surface. They deliberately carry no environment variable, since an env pin would lock the switch.

### Deliberately out of scope

Turning a feature off **only removes it from navigation**. The routes and API keep working, per the decision on that question — an existing link or bookmark still resolves, and no saved layout is deleted, so re-enabling restores each account's layout. (The design doc will be updated to match.)

## Screenshots

Verified in Chromium against a local SQLite instance with a mock admin user. Rather than attach images, here is what each surface actually did:

| Surface | Result |
| --- | --- |
| Sidebar row menu | `⋯` appears on hover; Move up/down and Hide from navigation present |
| Hide Favorites | Row leaves the list, **More 1** appears, hover reveals the restore eye |
| Reload | Hidden item still under More — server-rendered, no flash of the old nav |
| Pinned row | *Always shown* with a lock, no hide option |
| Settings → Navigation | 12 rows with grips, icon tiles, blurbs, Pinned captions, footer + Reset |
| Toggle in the manager | Sidebar beside it updates live, no reload |
| Reset | Back to the shipped order, More group gone |
| Admin turns Fitness off | Item leaves the sidebar; `GET /fitness` still **200**; manager row greys out with *off for this server* + *Admin* |
| Re-enable | Layout returns unchanged |
| Mobile (390px) | Bar reads Home · Search · Explore · Messages · More; More sheet carries the rest plus the Hidden section |
| Rail (900px) | Order matches; More flyout restores |
| Dark mode | Reads correctly; no hydration warnings in the console on any route |

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (792 files, 9071 tests)
- [x] Docs updated: `docs/features.md` gains the feature entry, `docs/environment-variables.md` documents the `features.*` settings and what "off" does and does not do
- [x] No migrations in this PR (preferences ride on the existing `actors.settings` JSON column), so no schema dumps to regenerate

https://claude.ai/code/session_01CR92RT8wJYXmJjziJmymnW

---
_Generated by [Claude Code](https://claude.ai/code/session_01CR92RT8wJYXmJjziJmymnW)_


llun:
## Sub-agent review rounds

Three rounds so far, each fixed in its own commit. Recording them here rather than as inline threads I would open and resolve against myself — the interesting part is what was wrong and how it is now prevented.

Everything below was **reproduced** before it was fixed, and every item has a regression test.

### Round 1 → `dd8e093`

The save queue decided whether a write was needed by comparing **what the navigation looks like**, while the request carries **the lists to store**. Those differ — Reset deliberately stores empty lists so the account keeps following the shipped navigation — and three bugs fell out of that one mistake:

- **Reset did nothing** whenever the visible order already matched the defaults, which is exactly when an account most needs its stored lists cleared.
- **Try again re-sent a change the user had taken back**: undoing a failed edit left the failed lists queued.
- **A failure while a newer edit was queued** corrupted the bookkeeping so the next edit was silently skipped.

Also: the tablet rail's More flyout was a **keyboard dead end** (Radix swallows Tab inside a menu and only moves focus between registered items, so hand-built rows are unreachable); the settings manager's grip **did nothing on a hidden row** while dragging the same row worked, and it announced moves that never happened; drag never started in **Firefox** (empty `dataTransfer`).

### Round 2 → `eea8c9b`

The round-1 fix was **incomplete** — the payload key it added was written and never read, so the decision still came down to visible state. A reset queued or in flight could be replaced by a gesture that changed nothing (picking a row up and dropping it back), and a failed save was discarded when a later edit happened to land back on the account's state, taking its error with it.

Also: rail restore items were all announced as the same bare "Show in navigation"; every hover-revealed control was **invisible but still tappable on a touch device** — and the rail is exactly tablet width; a drag cancelled with Escape still committed; the live region wrapped the Reset button, so `role="status"` announced the button on every save.

### Round 3 → `cd9585b`

Drag refs were set after an early return and never cleared, so a stray `dragend` **replayed an older drag's order**, reverting on screen what the account had already saved. An abandoned drag reverted the screen but not the account if something saved mid-drag. `dragover` re-spliced on every event, so the dragged row oscillated under a still cursor (masked only because rows are one line tall). Inside a list, **two links claimed `aria-current="page"`**. And an edit that superseded a reset and was then undone left the user looking at the reset navigation while the account kept the lists the reset existed to clear.

### Where that leaves the state machine

A save now records three things: the lists it sends, the navigation they leave behind, and whether its point is invisible (Reset). An ordinary edit that changes nothing on screen leaves a queued save alone; one that undoes itself drops the save it undid; a deliberate one is remembered until the account accepts it. Dedupe is on the payload, never on appearance.

### Verified

`prettier`/`lint`/`build`/`test` green (792 files, 9088 tests). Browser-verified against a local SQLite instance: hide/restore/reorder on all three surfaces, persistence across reload, live sync between the manager and the sidebar beside it, reset, the admin switch removing an item while `/fitness` still returns 200, mobile bar and More sheet, dark mode, and no hydration warnings. Keyboard-only pass through the rail flyout, and a touch-emulated pass confirming the controls are visible when the pointer cannot hover (opacity 1 on `hover: none`, 0 with a mouse).

---

llun:
## Sub-agent review rounds 4–10

Ten rounds in total; the comment above covers 1–3. **Round 10 found nothing**, which closes the loop.

Same rule as before: everything below was reproduced before it was fixed, and every fix has a regression test that I confirmed **fails without it** — a guard that passes either way is not a guard.

### Rounds 4–5 → `33f6da5`, `5692eb4`, `2a68194`

The once-per-row drag guard from round 3 had the opposite failure: after a move the dragged row sits under the pointer, and its own `dragover` was ignored without updating the guard, so dragging back over the row just passed did nothing — and the drop then saved the swap the user was taking back.

Reordering also **could not be done at all on a phone or tablet**: dragging needs a mouse and the grip's arrow keys need a keyboard. Each row now has Move up/down buttons, which makes the keyboard path discoverable at the same time; the grip is hidden where a pointer cannot hover, giving the label its width back. Pressing one of those buttons to the end of the list then dropped focus to the top of the document — a browser blurs a focused element the moment it becomes `disabled`, and that is exactly the button being pressed — so they are `aria-disabled` and say "already first/last" instead. Also: `/lists/new` and a list's edit page had no current-page marker at all, and a reset was cancelled by any save that landed, including an unrelated one queued ahead of it.

### Round 6 → `a2481fd`

Two more in the save store, both around Reset. A failed Reset was dropped by the next *successful* save, so the account silently kept the explicit lists Reset exists to clear while the screen showed the reset navigation. And pressing **Reset to defaults** again after its save failed did nothing — the identical-payload early return fired with nothing in flight — leaving only the small "Try again" link as recovery.

Two accessibility ones: the announcement region only spoke when the wording *changed*, so holding Move down against the bottom of the list said "already last" once and then went silent; and the save caption was itself a live region that changes twice per save, burying each row's "moved up, position 2 of 10" under "Saving…"/"Saved…" on every keystroke of a keyboard reorder.

### Round 7 → `bd4c06f`

The worst one. **"Try again" wrote the wrong thing.** A failed Reset stayed queued holding the snapshot it was created with, so: press Reset → it fails → rebuild your sidebar → press Try again, and the store sent Reset's empty lists while the caption flipped to "Saved". The next reload threw the layout away. Reset's empty lists stop describing the navigation the moment the user builds something else, so the queued save is now dropped like any other; the intent survives separately and is sent only if the navigation returns to what Reset produced.

Pressing that button also unmounted it, dropping focus to `<body>`, and — because round 6 had made the caption non-live — a successful retry reached a screen reader as silence.

### Round 8 → `82ecd11`

The store was **fuzzed clean**: 250 seeds × 60 steps mixing Reset, failures, edits during a save and drag commits, against three invariants — the store never reports a state the account does not hold, it converges once every flight drains, and it never runs two requests at once.

The findings were elsewhere: restoring an item from **More** dropped focus to `<body>` (the button unmounts *because* it succeeded); "Your changes are saved." stuck permanently, replacing the line that tells you there is no Save button; and the optional-feature list was written out four times across three files.

### Round 9 → `3ee0188`

The mirror image of round 8, on the gesture people actually use: **hiding** a row from the `⋯` menu dropped focus to `<body>`, with a worse cause — the row leaves the navigation taking its trigger with it, and Radix then restores focus to that detached trigger. Focus now follows the row wherever the menu said it was going; the menu stands aside via `onCloseAutoFocus` only when hiding, so Escape, an outside click and the moves keep the default restore. The rail's icon links also had no accessible name at all — the tooltip beside them only *describes* — so the rail read as a list of unnamed links, including the one focus is now handed.

### Where the loop converged

- **Saves are keyed on the payload, never on appearance**, and a save whose point is invisible (Reset) is owed until the account accepts *it* — not until some other save lands. Fuzz-verified.
- **A control that unmounts because it succeeded must hand focus on.** Three separate rounds found this: the disabled Move button, the More restore, the `⋯` hide. All three now name where focus goes.
- **One registry.** Adding a nav item is its id plus its presentation; adding an optional feature is now a *type error* until it has a switch, its copy, and a server-settings field, with a test asserting the last link.

### Verified

`prettier` / `lint` / `build` / `test` green — **792 files, 9107 tests**. Browser pass against a local SQLite instance for the focus work, since jsdom only approximates the real restore ordering:

| gesture | focus lands on |
| --- | --- |
| hide, More collapsed | the More toggle |
| hide, More open | the row's link in the group |
| restore from More | its place in the navigation |
| Escape from the `⋯` menu | the `⋯` trigger |
| Move up | the `⋯` trigger |
| restore the last item from the rail flyout | the rail button it became |

No page errors. Earlier rounds' browser passes (hide/restore/reorder on all surfaces, persistence, live sync between the manager and the sidebar beside it, the admin switch removing an item while `/fitness` still returns 200, mobile bar and More sheet, dark mode, no hydration warnings) still stand.

---

PR #1433, opened 2026-08-14
fix: paint the pinned gear table column in the card surface, and stop its type breaking mid-word

llun:
## Summary

Two fixes to the pinned first column of the gear tables, both checked against the design handoff kit rather than inferred.

**1. The pinned column was a white stripe down a grey card.** `STICKY_COLUMN` painted the sticky cell `bg-background`; it is now `bg-card`, so the column matches the card behind it and the hairline down its right edge does the separating on its own.

That is what the design does. `useGKSnapCols` in `ui_kits/web/GearKit.jsx` pins the cell with:

```js
const stick = { position: 'sticky', left: 0, zIndex: 1, background: 'white', boxShadow: `inset -1px 0 0 ${GK_BORDER}`, minWidth: typeW, … }
```

and every card holding one of these tables is `rounded-2xl border bg-white/80 shadow-sm backdrop-blur` over a white page. The lane is painted the **card's own colour** — the literal white is there to make the sticky cell opaque, not to step the column off anything. There is no recessed lane anywhere in the kit, and the kit's canonical `components/Card/Card.jsx` is `background: var(--card)`.

So `bg-background` copied a colour rather than a relationship. The kit is a static prototype that hardcodes white instead of reading `--card`, while its `app/globals.css` carries the same tokens this app has (light `--background` 100% / `--card` 98%; dark 3.9% / 9%) — so against a `bg-card` table it broke the design's relationship in **both** themes: a bright stripe in light, a well sunk below the card in dark. `bg-card` restores it in both.

**2. "Handlebar" rendered as "Handleba / r".** The pinned type column carried the design's 104px, but our pinned cell is `px-4` where the design runs a flat `px-3` across all seven columns — we spend the extra 8px so the column lines up with the card header above it. That left 72px of content, and "Handlebar" measures 72.6px at `text-sm font-medium`, so the most ordinary value in `COMPONENT_TYPE_OPTIONS` missed by half a pixel and `wrap-anywhere` did what it is there to do.

The column is now 120px rather than the wrap being dropped — `wrap-anywhere` is load-bearing for the free-text 255-character case, where a long unbroken value would otherwise widen the column past its snap interval and leave a tail `x mandatory` can never rest on. 88px of content clears "Chainrings" (74.7px), the widest single word in the option list, with headroom for the wider system fonts other platforms substitute into the same stack. Multi-word values still wrap at their spaces: "Front brake pads" would need a 149px pin to sit on one line, 38% of a 390px phone.

### Changes

- `app/(timeline)/fitness/gear/gearUi.ts` — `STICKY_COLUMN` uses `bg-card`; the comment explains the relationship the kit encodes and why the literal token does not transfer.
- `app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx` — `TYPE_COLUMN_WIDTH` 104 → 120, with the measurements it was derived from.
- `app/(timeline)/fitness/gear/gearUi.test.ts` — guards on the constants: the surface is the card's and not the page background, the hairline is still there, neither variant carries an opacity modifier (`bg-…/50` would let the data columns scroll through the sticky cell), and the clickable variant still repeats the row hover.
- `GearListView.test.tsx` / `GearComponentsCard.test.tsx` — pinned-column assertions follow the token and the width, plus a guard on the content budget the width was derived from.
- `AGENTS.md`, `CLAUDE.md`, `REVIEW.md` — the three places that documented `bg-background` as required now document `bg-card` with the kit's reasoning, and the width departure is recorded with its measurements.

## Screenshots

Verified in Chromium against a local SQLite instance, at 390px and 1280px, in both themes, with every worst-case standard component type seeded. Measured from the rendered page with `getComputedStyle`:

| | pinned cell | card behind it | hairline |
|---|---|---|---|
| before, light | `rgb(255, 255, 255)` | `rgb(250, 250, 250)` | present |
| after, light | `rgb(250, 250, 250)` | `rgb(250, 250, 250)` | `rgb(229, 229, 229)` inset |
| after, dark | `rgb(23, 23, 23)` | `rgb(23, 23, 23)` | `rgb(46, 46, 46)` inset |

At 120px, "Frame", "Handlebar", "Chain", "Cassette", "Front tire", "Chainrings" and "Seatpost" each render on one line; "Rear Derailleur", "Bottom bracket" and "Front brake pads" wrap at their spaces. No mid-word break.

## Not in this PR

A full conformance pass over the gear tables against `GearKit.jsx` turned up two divergences that are **out of scope here** and are recorded in `AGENTS.md` / `CLAUDE.md` as known gaps rather than decisions:

1. **The bikes/shoes/devices tables never snap below 480px** and still carry `min-w-[520px]`, where the kit runs all four tables through `useGKSnapCols`. On a phone the gear list scrolls as one block — the same failure `min-w-[720px]` used to cause on the components table before it was removed.
2. **The pinned header cell is `z-1` where the kit overrides `stick` with `zIndex: 2`.** No symptom today because no `thead` is vertically sticky, but it becomes real the moment one is.

Everything else matched: hairline mechanism, side, width and colour; hover lit on both the row and the pinned cell, opaque, on the two clickable tables and absent on the inert components table; the snap threshold, column formula, `scroll-snap-type` and `scroll-padding-left`; `border-t` row separators; dimming applied to cells and never the `<tr>` or the pinned `<td>`, with the components table's action cell deliberately undimmed. The gutter differs at 16px against the kit's 20px, tracking each one's own card header — which is also where the 120px type column comes from.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — re-run after rebasing onto `main` (793 files, 9150 tests)
- [x] Rebased onto `main` at 377b0ab; the one conflict (`CLAUDE.md`, adjacent to the new navigation bullet) is resolved keeping both
- [x] Docs updated: grepped `*.md` and `docs/` for `bg-background` and the pinned-column rules; `AGENTS.md`, `CLAUDE.md` and `REVIEW.md` updated
- [x] No migrations in this PR

https://claude.ai/code/session_01NPieDqoHcJQ9hT6ynKRhoP


app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -212,15 +212,22 @@ export const COMPONENT_TYPE_OPTIONS = [
 // Rendered as plain columns on the card, as these were, the rows read as loose
 // text.
```
llun: Round 1 (sub-agent review) — **stale opening paragraph contradicts the new one directly beneath it.**

This paragraph was left untouched, so the comment now says both things at once. It still claims the pinned column stands "on its own surface", that "the column standing off the card" is what separates each row's subject from its numbers, and that "rendered as plain columns on the card … the rows read as loose text" — which is precisely the layout the change below now adopts. Lines 215+ then say the surface is `bg-card` and the hairline is what separates it.

`AGENTS.md` was updated to the new framing ("that hairline is the table's structure … rendered as plain columns **with no rule**"), so the canonical doc and the code comment it points at disagree, and the next reader of this constant finds a written rationale for reverting the fix.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -212,15 +212,22 @@ export const COMPONENT_TYPE_OPTIONS = [
...
-// card"): following that ramp would mean `--muted`, which is also the hover
-// colour, so the pinned column would stop responding to hover.
+// The column's surface is `bg-card` — the same grey as the card it sits on —
+// and the hairline is what separates it. It used to be `bg-background`, which
+// is how the design draws it, but only because the design's surfaces are
+// stacked the other way up: there the page is the grey and a card is white, so
+// the pinned lane reads as a slightly recessed strip inside a white card. This
+// app inverts that pair (`--background` is pure white, `--card` is 98%), so the
+// same token rendered the lane as a bright white stripe down a grey card —
+// louder than the design's, and the first thing anyone noticed about these
```
llun: Round 1 (sub-agent review) — **the rationale is stated theme-independently, but it is only true in light mode.**

"This app inverts that pair (`--background` is pure white, `--card` is 98%)" quotes the light tokens only. In dark mode `app/globals.css` has `--background: hsl(0 0% 3.9%)` and `--card: hsl(0 0% 9%)`, so the card is the *lighter* of the two — stacked the same way up as the design. `bg-background` there was a genuinely recessed dark lane, not "a bright white stripe".

The prose this PR replaced got that right ("it steps opposite ways in the two themes"), so as written this trades accurate two-theme documentation for a one-theme generalisation, and it silently drops the dark-mode consequence of the change: the lane is now identical to the card there too, leaving only the hairline. Same wording is repeated in `AGENTS.md`, `CLAUDE.md` and `REVIEW.md`, so all four need the correction — and the dark-mode trade should be stated as a deliberate choice (one rule across themes) rather than left unmentioned.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -239,7 +246,8 @@ export const STICKY_COLUMN =
  * highlights.
  *
  * Both surfaces use the OPAQUE `bg-muted`, never `bg-muted/50`. A translucent
- * hover does not layer over `bg-background`, it replaces it, so the cell would
+ * hover does not layer over the cell's own `bg-card`, it replaces it, so the
+ * cell would
```
llun: Round 1 (sub-agent review) — **ragged mid-edit line breaks.** "…it replaces it, so the / cell would / be 50% transparent…" is an un-reflowed artifact of the edit. `.prettierrc.yml` sets no `proseWrap`, so the default `preserve` means `yarn prettier --write .` will never clean this up. Same in two spots in `AGENTS.md` (the "The divider is an inset / shadow" and "exactly / while the / pointer is on the row" breaks).

---

app/(timeline)/fitness/gear/gearUi.test.ts
```diff
@@ -316,3 +318,31 @@ describe('getProductUrlHostname', () => {
     expect(getProductUrlHostname(url)).toBe(expected)
   })
 })
+
+describe('STICKY_COLUMN', () => {
```
llun, line 322: Round 1 (sub-agent review) — **the block is named for `STICKY_COLUMN` but two of its three cases are about `STICKY_CLICKABLE_COLUMN`.** `AGENTS.md` → Testing Guidelines says to name the thing under test directly. Split them, or name the block for what it actually covers (the pinned-column constants).

Also, `expect(STICKY_CLICKABLE_COLUMN).toContain(STICKY_COLUMN)` is near-tautological — the clickable variant is defined as a template literal over `STICKY_COLUMN`, so it can only fail if someone re-literalises the constant. It adds little next to the `group-hover:bg-muted` assertion on the following line.

(For the record, the opacity regex was checked and is **not** vacuous: `/bg-[a-z-]+\/\d+/` is `false` on both constants today and `true` on `bg-card/50` and `group-hover:bg-muted/50`.)

---

llun (commented):
Sub-agent code-review, round 1 of up to 20. Two agents reviewed the whole diff — one on correctness/CSS, one on docs/tests.

**Correctness/CSS agent: no defects found.** It verified that every consumer of `STICKY_COLUMN`/`STICKY_CLICKABLE_COLUMN` renders inside a `Card` (which paints `bg-card` unconditionally and is never overridden at these three call sites), that the cell stays fully opaque and that dimming is applied to inner elements rather than the `<tr>` or the pinned `<td>`, that `group-hover:bg-muted` still wins on specificity — Tailwind v4 compiles it to `.group-hover\:bg-muted:is(:where(.group):hover *)`, (0,2,0) against `.bg-card`'s (0,1,0), so it is not order-dependent — and that `useGearTableColumns` reads no colour. It also computed the hairline: `--border` on `--card` is 1.21:1 in light and 1.32:1 in dark, both slightly weaker than before but exactly the strength of the `border-t` row rules already in these same tables, and in dark the 18%-on-9% pairing the ramp in `app/globals.css` documents.

**Docs/tests agent: four findings**, filed inline below. It separately confirmed no stale `bg-background` survives in any `*.md`, that `GEMINI.md` and `.github/copilot-instructions.md` don't mirror this rule, that `app/globals.css` never carried a gear-table exception comment, and that the opacity regex is not vacuous.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Fixed in cb694ab. The opening paragraph is rewritten around the hairline: it now says the pinned column scrolls the data columns under it "with a hairline down its right edge", that the hairline "is the table's structure — its only vertical rule, and what separates each row's subject from its numbers", and that the failure mode is "rendered as plain columns **with no rule**". No sentence claims the column stands off the card any more, so the comment and `AGENTS.md` say the same thing.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Correct, and fixed in cb694ab. All four places now split the two themes: light mode inverts the design's pair (`--background` white, `--card` 98%) and is where the white stripe came from; dark mode runs the design's way round (`--background` 3.9% below `--card` 9%), so `bg-background` genuinely read as a recessed lane there. The dark-mode flattening is now stated as a deliberate trade — one rule that behaves the same in both themes beats a lane that is right in one and wrong in the other — rather than left unmentioned. `gearUi.ts`, `AGENTS.md`, `CLAUDE.md`, `REVIEW.md` and the test comment all updated.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Reflowed in cb694ab — the JSDoc line here and both `AGENTS.md` spots are back to full 80-column lines. (Confirmed `yarn prettier --write .` reports all three files unchanged afterwards, so this was indeed only fixable by hand.)

---

app/(timeline)/fitness/gear/gearUi.test.ts
llun: Split in cb694ab: `describe('STICKY_COLUMN')` now holds only the two cases about that constant (card surface, opaque surface), and a separate `describe('STICKY_CLICKABLE_COLUMN')` holds the hover ones. The `toContain(STICKY_COLUMN)` assertion is dropped for the reason you give; the `group-hover:bg-muted` and opacity assertions carry the block. The `it.each` went with the split — the two remaining opacity cases now sit in different blocks and each has its own reason to exist, so a shared table would have obscured rather than compressed them.

---

app/(timeline)/fitness/gear/gearUi.test.ts
```diff
@@ -316,3 +318,36 @@ describe('getProductUrlHostname', () => {
...
+  it('paints the pinned column in the card surface, not the page background', () => {
+    expect(STICKY_COLUMN).toContain('bg-card')
+    expect(STICKY_COLUMN).not.toContain('bg-background')
+  })
+
+  // A sticky cell with a see-through background lets the data columns scroll
+  // straight under it, so the surface may not carry an opacity modifier.
+  it('keeps the surface opaque', () => {
+    expect(STICKY_COLUMN).not.toMatch(/bg-[a-z-]+\/\d+/)
+  })
```
llun, line 338: Round 2 (sub-agent review) — **nothing guards the hairline, and this PR is what makes that matter.** Verified by mutation: deleting `shadow-[inset_-1px_0_0_var(--border)]` from `STICKY_COLUMN` leaves all 166 gear tests passing.

Before this change the surface step did the separating and the inset shadow was reinforcement; now the comment above the constant says the hairline "does that separating on its own", so losing it silently returns the table to the loose text the constant exists to prevent. That is exactly the silent-failure class this `describe` block was added for. One assertion closes it.

---

AGENTS.md
```diff
@@ -548,28 +548,35 @@ it; there is no legacy shape left to copy.
...
+  `border-r`, because `border-collapse: collapse` (Tailwind's preflight default)
+  hands border painting to the table and drops a sticky cell's own right border.
+  A dimmed row (retired gear, a replaced component) dims its **cells**, never the
+  `<tr>` and never the pinned `<td>` itself — `opacity` fades an element's
+  background along with its text, so either one takes the pinned column's surface
+  down with it. And the hover colour is the OPAQUE `bg-muted` on both the row and
+  its pinned cell, never `bg-muted/50`: a translucent hover replaces the cell's
+  own surface rather than layering over it, so the cell would turn 50%
+  transparent exactly while the pointer is on the row. Rows separate with
+  `border-t`, so the header carries no
```
llun: Round 2 (sub-agent review) — **one ragged line survived the round-1 reflow.** "`border-t`, so the header carries no" ends at column 38 in a paragraph whose every other line wraps at 65–82. Cosmetic, but it's the visible seam of the edit; rewrap from here to the end of the sentence.

---

app/(timeline)/fitness/gear/gearUi.ts
```diff
@@ -205,22 +205,30 @@ export const COMPONENT_TYPE_OPTIONS = [
...
-// `bg-card` behind it, and a sticky cell has to be OPAQUE or the columns
-// scrolling underneath show straight through it. It steps opposite ways in the
-// two themes — lighter than the card in light mode, the way the design has it,
-// and darker in dark mode, where `--background` sits below `--card`. Dark
-// therefore reads as a recessed well rather than a raised surface, a deliberate
-// exception to the ramp documented in `app/globals.css` ("insets sit above the
-// card"): following that ramp would mean `--muted`, which is also the hover
-// colour, so the pinned column would stop responding to hover.
+// The column's surface is `bg-card`, the same grey as the card behind it, so
+// the hairline does that separating on its own. The design paints the lane in
```
llun: Round 2 (sub-agent review) — minor: the hairline's job is now stated three times in 25 lines ("that hairline is the table's structure … what separates each row's subject from its numbers" at 209-210, "so the hairline does that separating on its own" here, "and the hairline is doing the work in either case" at 226). The first states it and the last carries the trade-off's conclusion; this middle one is pure repetition and can go.

---

llun (commented):
Sub-agent code-review, round 2 of up to 20 — a pass over the round-1 fixes plus anything round 1 missed.

**Verified clean:** every token value quoted in `gearUi.ts`, `AGENTS.md`, `CLAUDE.md` and `REVIEW.md` matches `app/globals.css` (light `100%`/`98%`, dark `3.9%`/`9%`), and the light-inverts / dark-follows-the-design split is the correct reading of them. No stale sentence survives anywhere (greps for `bg-background`, "stands off", "steps the column off", "its own surface", "recessed well" across `*.md`/`*.ts`/`*.tsx`). Dropping `expect(STICKY_CLICKABLE_COLUMN).toContain(STICKY_COLUMN)` in round 1 lost no coverage — mutating the constant to lose its base still fails three tests in `GearListView.test.tsx`. "Four details are load-bearing" still parses as four, with the opacity sentence reading as a rider on the surface detail ("Whatever the colour…").

**Three findings**, filed inline. The first is the one that matters: mutation testing showed the hairline is now the pinned column's only separator and **nothing asserts it exists**.

---

app/(timeline)/fitness/gear/gearUi.test.ts
llun: Good catch — closed in d84a777 with `it('keeps the hairline down its right edge')`, asserting `STICKY_COLUMN` still contains `shadow-[inset_-1px_0_0_var(--border)]`. The comment above it records why the guard is new: the shadow was reinforcement while the column had its own surface, and became the only separator when it stopped having one. Gear suites now 167 tests.

---

AGENTS.md
llun: Rewrapped in d84a777, along with a second short line the same edit left two sentences later ("Four details are / load-bearing."). Checked the whole bullet by line length afterwards — the only lines under 60 columns now are paragraph-final ones.

---

app/(timeline)/fitness/gear/gearUi.ts
llun: Dropped in d84a777. The paragraph now opens straight on the surface ("The column's surface is `bg-card`, the same grey as the card behind it. The design paints the lane in its page background…"), leaving the statement at the top and the conclusion at the bottom.

---

llun:
Sub-agent code-review loop closed after **round 3 of up to 20** — a full round with nothing actionable, which is the termination condition in `AGENTS.md` → Code Review Loop.

| Round | Findings | Outcome |
|---|---|---|
| 1 | 4 (stale comment paragraph, light-mode-only rationale, ragged prose, test block naming) | all fixed in cb694ab, replied and resolved |
| 2 | 3 (**unguarded hairline**, one ragged line, a repeated sentence) | all fixed in d84a777, replied and resolved |
| 3 | 0 | clean |

Round 3 re-verified the earlier rounds rather than trusting them, and mutation-tested every assertion this PR adds — each mutation fails the suite it should:

| mutation of `STICKY_COLUMN` / `STICKY_CLICKABLE_COLUMN` | failures |
|---|---|
| `bg-card` → `bg-background` | 6 |
| `bg-card` → `bg-card/50` | 7 |
| drop `shadow-[inset_-1px_0_0_var(--border)]` | 1 (the guard added in round 2; nothing caught this before) |
| drop `group-hover:bg-muted` | 4 |
| `group-hover:bg-muted` → `group-hover:bg-muted/50` | 4 |

It also confirmed all four render sites sit inside a `<Card>`, that the specificity of `group-hover:bg-muted` (0,2,0) still beats the cell's own `.bg-card` (0,1,0) so hover is unchanged, that every token value quoted in the four docs matches `app/globals.css`, and that no stale `bg-background` claim survives anywhere in the repo. CI is green on d84a777 (Build, all four test shards, Lint and Prettier, Schema Dump Sync, CodeQL).

---

llun:
**Verified against the design handoff** (`ui_kits/web/GearKit.jsx`), which was not readable when this PR was opened. The change stands; the rationale in the docs was wrong and is corrected in 50eb934.

What the kit actually does:

```js
// useGKSnapCols
const stick = { position: 'sticky', left: 0, zIndex: 1, background: 'white', boxShadow: `inset -1px 0 0 ${GK_BORDER}`, minWidth: typeW, … }
```
```jsx
<section className="rounded-2xl border bg-white/80 shadow-sm backdrop-blur" …>   // all three tables
```

The lane and the card are **the same colour**. The `background: 'white'` is there to keep the sticky cell opaque, and the hairline is the only separator — there is no recessed strip anywhere in the kit, in either theme. The kit's own canonical surface component is `background: var(--card)`, and its `app/globals.css` carries the same tokens as this app (light 100%/98%, dark 3.9%/9%), so the prototype's hardcoded white is a shortcut, not a spec.

That means `bg-background` was copying a colour where it should have copied a relationship, and it broke that relationship in **both** themes rather than only in light. Two corrections to what this PR said earlier:

- ~~"the design stacks its surfaces the other way up (grey page, white card)"~~ — it does not; the design system's tokens are identical to this app's.
- ~~"in dark mode `bg-background` genuinely read as a recessed lane, and flattening it is a deliberate trade"~~ — there was no fidelity to trade. The dark well was a divergence too, and `bg-card` fixes both themes rather than one.

The conformance pass also turned up two divergences outside this PR's scope, now recorded in `AGENTS.md`/`CLAUDE.md` as known gaps and listed in the description: the gear list's three tables never snap below 480px (the kit runs all four through `useGKSnapCols`), and the pinned header cell is `z-1` where the kit sets `zIndex: 2`. Everything else — hairline, hover on both halves, 150/150/150/104 pin widths, the snap formula, row separators, dimming placement including the undimmed action cell — matches.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -39,8 +39,24 @@ const SERVICE_REMINDER_KM_OPTIONS = [1000, 3000, 5000, 8000, 12000]
...
+ *
+ * 120px leaves 88px of content, clear of the widest single word in the option
+ * list — "Chainrings" at 74.7px — with enough headroom for the wider system
+ * fonts other platforms substitute into the same stack. Multi-word values still
+ * wrap, but at their spaces: "Front brake pads" needs 117px on one line, which
+ * would be 149px of pinned column, 38% of a 390px phone for a column of short
+ * labels. The design does not spend that either.
  */
-const TYPE_COLUMN_WIDTH = 104
+const TYPE_COLUMN_WIDTH = 120
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:59]: Round 4 (sub-agent review) — **widening the pin to 120px pushes the snapped distance off the scrollport on a 320px phone, and `x mandatory` means you cannot scroll to it.** This one is a regression introduced by the width fix, not pre-existing.

`MIN_SNAP_COLUMN_WIDTH` (184) floors the data column, so when the pin leaves less than that the column is deliberately wider than the space available and hangs off the scroller's right edge. Its content is `textAlign: 'right'`, so the overhang eats the value first.

Measured in the browser at a 320px viewport (scroller 286px, pin 120px):

```
dataColumnWidth: 184   cellRight: 321   valueRight: 309   scrollerRight: 303
valueClippedBy: 6      valueText: "0.0 km"
```

6px of the distance is off the edge at every rest position. At 104px the same scroller gave a 2px overhang, inside the cell's own 12px right padding, so nothing was hidden — this band (viewport ~310–325px: an SE, or any phone in Display Zoom) is newly broken.

The floor exists so the wear line still fits, and its own comment says the degradation below it is the line spilling leftwards **under the pinned column** — which is strictly better than a distance nobody can reach. So the floor should be allowed to overhang only as far as the cell's own right padding can absorb, and fall back to the available width past that.

---

llun (commented):
Sub-agent code-review, round 4 of up to 20 — over the newest commit (the 120px type column) and over the rebase itself.

**One finding**, inline below, and it is a real regression from this PR: the wider pin pushes the floored data column past the scrollport at ~320px viewports, hiding 6px of the distance where 104px hid nothing. Reproduced in the browser, not just computed.

**Verified clean otherwise:**

- **The new tests are not vacuous.** Mutating `TYPE_COLUMN_WIDTH` back to 104 fails three tests in `GearComponentsCard.test.tsx`, including the new content-budget one (`expected 72 to be greater than or equal to 75`). The 120px / 270px / `scrollPaddingLeft: 120px` numbers all follow from the formula at the delivered width of 390.
- **The rebase lost nothing.** `git diff origin/main...HEAD -- app/ REVIEW.md` is byte-identical to the pre-rebase `git diff 3198228...66acb30`, and the `AGENTS.md` diff is identical modulo hunk offsets. `CLAUDE.md` carries main's navigation bullet exactly once, byte-identical to `origin/main`'s, alongside this branch's gear edits byte-identical to the pre-rebase commit's. No conflict markers anywhere; main's `NAV_ITEM_IDS` / `LOCKED_NAV_IDS` / "Navigation Customization" content all present.
- **The design claims in the docs check out against the kit**, re-verified from the files rather than from the earlier round: `background: 'white'` at `GearKit.jsx:126` inside `bg-white/80` cards (231/429/496), `useGKSnapCols(150)` at 223/492 and `useGKSnapCols(104)` at 365, the kit's flat `px-3` on its components table's pinned header (457) and body cells (382) — so 104−24=80 there against our 104−32=72, exactly the stated cause of the mid-word break. Token values in `app/globals.css` match; `min-w-[520px]` really is still on both gear-list tables, so the "known gap" wording is honest.

---

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:59]: Fixed in 80f199b, exactly as suggested — the floor may now overhang the scrollport only by `SNAP_OVERHANG_ALLOWANCE` (12px, the data cell's own `px-3` right padding), and past that the column falls back to the width actually available:

```ts
const availableWidth = width - pinnedColumnWidth
const columnWidth = Math.max(
  availableWidth,
  Math.min(MIN_SNAP_COLUMN_WIDTH, availableWidth + SNAP_OVERHANG_ALLOWANCE)
)
```

Re-measured in the browser at the same 320px viewport:

| | before | after |
|---|---|---|
| data column | 184px | 178px |
| value right edge | 309 | 303 |
| scroller right edge | 303 | 303 |
| **clipped by** | **6px** | **0** |

The value now ends flush with the scroller's edge — tight, but whole — and the wear line spills 4px under the pin instead of 12px of the distance being unreachable.

The floor still applies wherever it costs nothing: at a 280px container (176 available) it takes 184 and overhangs by 8, inside the padding. Two tests in `useGearTableColumns.test.tsx` pin both sides of that boundary, and the existing floor test moved to widths where the floor genuinely binds. `AGENTS.md`'s snap bullet now states the cap rather than "exactly the width the pinned column leaves over".

---

llun:
Sub-agent code-review, round 5 of up to 20 — over the round-4 fix.

The fix itself verified correct: the formula was walked across `width` 0–480 for pins 104/120/150 (`columnWidth` never below `availableWidth`, overhang never above 12, and equal to `availableWidth` exactly whenever it exceeds the floor), the last column's snap point stays reachable, the trailing `pr-4` cell absorbs the 12px with 4px to spare, and mutating `SNAP_OVERHANG_ALLOWANCE` to 0 and to 40 fails tests in both directions.

**Two findings, both fixed:**

1. **The new comment contradicted the code it documents.** It said the column "falls back to the width actually available" past the cap; it does not — it keeps exactly 12px of overhang all the way down, which the existing test already pinned (`240 − 104 = 136` available → `148px`, not `136px`). The code's behaviour is the better one: the value lands flush with the scroller's edge at every narrow width instead of 12px short of it at some. Prose corrected to match, since a future reader "fixing" the code to match the prose would take 12px of content off every column below 292px.

2. **The regression's own scenario wasn't covered.** The test comment cited a 320px viewport, but the probe pins at 104px, where that width is fine — the overhang only bit at the components table's production 120px pin. The probe now takes a `pinnedWidth`, and a new case asserts the reported scenario directly: 286px scroller at a 120px pin → 178px, not the floor's 184.

Also corrected `CLAUDE.md`'s mirrored bullet, which still claimed a data column "takes **exactly** the width the pinned column leaves over" — untrue since the floor was introduced and doubly so now.

**Noted, not changed** (pre-existing, and out of this PR's scope): `MIN_SNAP_COLUMN_WIDTH`'s 184 is justified in its comment by a 158px wear line — "an 80px bar, an 8px gap and a caption" — which describes the kit's `GKWearBar`, a single `justify-end` row. Ours stacks the caption below the bar, so the distance cell's real min-content is nearer 108px, which is exactly what the same cell's off-snap `dataColumnStyle(108)` already claims. The kit's own floor (`max(160, w − 104)`) never overhangs at all. So the right long-term fix is either lowering the floor to our layout or bringing `WearBar` in line with the kit's inline row — a design-fidelity gap of its own, alongside the gear-list snapping one.

---

llun:
Sub-agent code-review, round 6 of up to 20 — over the round-5 corrections.

The round-5 prose is confirmed accurate against the formula, which the round walked itself: with `a = width − pin`, `a ≥ 184` → `a` (no floor, no overhang); `172 < a < 184` → `184` (floor binds, overhang always under 12); `a ≤ 172` → `a + 12` (cap binds, constant 12px overhang, value flush with the scroller edge, since `column − 12px padding = a` exactly). The 320px anecdote re-derives: 286 − 120 = 166 available, floor 184 hangs 18px off, minus 12px of padding = the 6px of glyph that was hidden. The new test's 178 at a 120px pin follows, the `pinnedWidth` default leaves every older case byte-identical in meaning, and the one older case that changed had to.

**Two findings, both prose, both fixed:**

1. **`MIN_SNAP_COLUMN_WIDTH`'s own comment still opened "A snapped data column never gets narrower than this"** — which the cap added in this PR makes false (240px scroller at a 104px pin gives 148px, asserted three lines from the claim). It sat directly above the block round 5 was spent correcting and said the opposite, so a reader trusting it would have removed the cap and restored the 320px regression. It now reads as the width the column *aims* for, subject to the allowance below it.

2. **The "~724px" figure in `AGENTS.md` and `CLAUDE.md` was stale.** It is the sum of the components table's seven per-cell minimums, and this PR moved the first of them: 120 + 96 + 132 + 108 + 112 + 88 + 84 = **740**. That number is doing real work in both docs — it is the stated reason `min-w-[720px]` is redundant *and* the top of the "still scrolls as one block" band — so both now say ~740px and show the arithmetic.

**Verified clean otherwise:** `bg-card` consistent across `gearUi.ts`, both test files and all three docs; the token values quoted in `gearUi.ts` match `app/globals.css` exactly; `globals.css` never named the removed "recessed well" exception, so nothing dangles; `docs/` says nothing about these tables; no hand-rolled sticky cell anywhere; commit prefixes conventional and the `docs:` commit genuinely comment/test/markdown only.

---

llun:
Sub-agent code-review, round 7 of up to 20.

Round 6's arithmetic verified independently — the seven minimums read straight out of `GearComponentsCard.tsx` (120 pinned + 96 + 132 + 108 + 112 + 88 + 84) do sum to 740, both docs quote that and the same working, and a sweep of every `104`/`720`/`724`/`740`/`min-w-[` occurrence in the gear tree and all `*.md` found no other stale figure: the remaining `104`s are all explicitly *the design's* number or the deliberately-not-120 probe constant. `GearComponentsCard.tsx`'s looser "already add up to it" still holds, since 740 ≥ 720.

**One finding, fixed:** the comment on the new `COMPONENTS_PINNED_WIDTH` test constant said the overhang cap "only bites at the wider pin" — contradicted by the test 90 lines below it, which uses the **default 104px** pin at a 240px scroller and asserts 148px, i.e. the cap biting. The cap applies at either pin; only the width where it starts moves (below 276px at 104, below 292px at 120). What is specific to the wider pin is the 320px-viewport regression, and the comment now says that instead. As written, a future editor could have read it as "the cap is inert at 104px" and stripped those assertions.

**Not taken:** the round also flagged a few short wrapped lines the doc edits left in `AGENTS.md`. Reflowing those paragraphs properly rewrote 58 lines of a shared file for something prettier accepts and no reader relies on, so I reverted it — the churn is worse than the raggedness.

Otherwise clean: the `bg-card` change re-checked end to end (both call sites inside a `Card`, hover still landing on rows that carry `group` and their own `hover:`, the opacity guard unable to false-pass), no leftover `bg-background` anywhere, no project-invariant violations, commit prefixes conventional.

---

llun:
Sub-agent code-review loop closed after **round 8 of up to 20** — a full round with nothing actionable, which is the termination condition in `AGENTS.md` → Code Review Loop.

| Round | Findings | Outcome |
|---|---|---|
| 1 | 4 — stale comment paragraph, light-mode-only rationale, ragged prose, test block naming | fixed, replied, resolved |
| 2 | 3 — **unguarded hairline**, ragged line, repeated sentence | fixed, replied, resolved |
| 3 | 0 | clean |
| — | *design handoff arrived; rationale rewritten from the kit, conformance pass run* | |
| 4 | 1 — **the 120px pin pushed the distance off a 320px scrollport** | fixed, replied, resolved |
| 5 | 1 — the new comment described a fallback the code does not have | fixed |
| 6 | 2 — `MIN_SNAP_COLUMN_WIDTH`'s own comment contradicted the cap; stale ~724px total | fixed |
| 7 | 1 — the test constant's comment named the wrong pin | fixed |
| 8 | 0 | clean |

Round 8 re-derived the cap's two thresholds from the formula rather than reading them (`width < pin + 172` → 276px at the 104px default, 292px at the 120px components pin, matching what the comment claims), re-checked that every load-bearing number in the docs matches the code it describes (`TYPE_COLUMN_WIDTH = 120` against the `120px`/`270px`/`scrollPaddingLeft` assertions, `120 − 32 = 88 ≥ 75` for the wrap budget, the seven minimums summing to 740), and confirmed `bg-card` is consistent across the code, both test files and all three docs with no `bg-background` left anywhere.

Two substantive defects were caught across the loop and neither was in the original diff: the hairline had become the pinned column's only separator with nothing asserting it existed, and widening the pin moved a pre-existing sharp edge in the snap floor onto a real device width. Both are fixed and guarded.

State: rebased onto `main` at 377b0ab, `mergeable_state: clean`, CI green.

---

PR #1436, opened 2026-08-14
minor: show a gear's activities as posts, on the gear and device pages

llun:
## Summary

A gear's page now carries its activities the way the design system does. A bike switches between **Components** and **Activities** with an in-page dropdown; a device — which has no components to switch to — goes straight to the feed. Both render the shared `Posts` feed the timelines use, so a ride reads identically wherever it appears: same body, same fitness stat chip, same action row (reply / boost / like / bookmark / react / `⋯`).

**What changed**

- **`GearActivitiesFeed`** (`app/(timeline)/fitness/gear/[id]/`) is the single implementation, used by a bike's Activities view and by the whole of a device's page. It replaces the device page's bespoke compact row list, which offered none of a post's actions.
- **`GET /api/v1/fitness/gear/:id/activities`** now answers `{ statuses, hasMore, nextOffset }`. The page's posts are loaded in one batched `getStatusesByIds` read, hydrated for the caller so the like / bookmark / reaction state is the viewer's own. The ownership 404, the limit clamp and the one-row lookahead behind `hasMore` are unchanged.
- **`nextOffset` counts activity ROWS, not the statuses returned.** Deleting a status only nulls `fitness_files.statusId`, so a row with no post left still occupies an offset and still counts toward the gear's totals; paging from `statuses.length` would re-request everything in between forever. A page that comes back entirely postless is walked past rather than handed back as a "Load more" that adds nothing — capped the way the bookmarks timeline caps its own continuations.
- **`SectionNavSelect`** (`@/lib/components/section-nav-select`) is new, extracted from the private copy the fitness activity detail was already carrying. The state-driven twin of `SectionNavDropdown` now has one implementation instead of two, and the gear switcher is its second consumer rather than a third copy. Its active row moves to `text-primary-text` (the contrast-safe orange the design kit itself uses there) from `text-primary`.
- **Shoes and devices render no switcher** — neither has a components card, and a menu with one entry is dead UI, the same rule that keeps "No gear" off an empty shed's picker.

**One asymmetry this introduces, deliberately:** the Activities stat tile counts activities while the feed shows the ones still posted, so an activity whose post was deleted counts but cannot appear. `AGENTS.md`, `REVIEW.md` and `docs/fitness-file-storage.md` previously said a device's count and its page "can never disagree"; all three now say what actually holds.

Docs updated in the same PR: `AGENTS.md` (Fitness Gear), `CLAUDE.md`, `REVIEW.md`, `docs/fitness-file-storage.md` (gear tracking + the endpoint's shape), `docs/features.md`.

## Screenshots

Verified in a real browser against a local SQLite database seeded with a bike (4 components), a pair of shoes, a recording device and 40 activities, 17 of them posted. Desktop 1280px and mobile 390px.

**Bike → Components (default), switcher open** — the dropdown lists Components / Activities with the active row in the orange wash, exactly as the design kit has it; the components table is unchanged.

**Bike → Activities** — the components card is replaced by the shared feed: each ride is its own post with the fitness stat chip (`22.0 km · Moots`, `Via: Garmin Edge 840`) and the full action row.

**Device page** — header, the Activities / First used tiles, then the same feed, no switcher.

**Mobile (390px)** — the dropdown goes full width, the stat tiles stack, and the posts collapse bookmark/react into `⋯` through the existing `useCompactActionBar` rule.

Screenshots are attached in the session rather than committed — `docs/` is for durable reference documentation only (AGENTS.md → Project Structure).

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — 795 test files, 9159 tests, all green

https://claude.ai/code/session_018ifxGVQHGt22wJRxowMQ8G

---
_Generated by [Claude Code](https://claude.ai/code/session_018ifxGVQHGt22wJRxowMQ8G)_


app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
```diff
@@ -0,0 +1,246 @@
...
+    <div className="space-y-4">
+      {error && (
+        <p className="text-sm text-destructive" role="alert">
+          {error}
+        </p>
+      )}
+
+      {isLoading ? (
+        <Card className="p-6 text-sm text-muted-foreground">Loading...</Card>
+      ) : statuses.length === 0 ? (
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx]: **bug — `statuses.length === 0` no longer means "no activities", so the empty message can render above an enabled "Load more"** *(found independently by the correctness and design/a11y reviewers)*

The empty card is gated on `statuses.length === 0` alone. That test was correct in the code this replaces — `DeviceDetailView` on `main` listed activity **rows**, so an empty page really did mean an empty history — but it is not correct now that the page carries statuses. `loadMore` has `MAX_EMPTY_CONTINUATIONS` precisely because a page of 20 activity rows can yield zero statuses: deleting a post only nulls `fitness_files.statusId` (`ON DELETE SET NULL`), and `applyCountableActivityFilter` has no `statusId` predicate, so the row survives and keeps counting. The initial effect has no equivalent walk, and nothing gates the empty card on `hasMore`.

**Concrete failure.** A device with 30 activities whose 20 newest lost their posts. The route answers `{ statuses: [], hasMore: true, nextOffset: 20 }`, and the page renders, top to bottom: the `Activities: 30` stat tile, then `No recent activities recorded with this device.`, then an enabled `Load more` — which, when clicked, fills the feed and proves the message false. The `useLoadMoreOnVisible` sentinel only papers over this when it is already in the viewport at mount; on a phone the back link, `PageHeader` and the stat tiles push it below the fold.

It is also reachable with no deleted post at all: `onPostDeleted` → `removeStatus` empties `statuses` while `hasMore` stays true, so deleting the last visible post on page 1 of a long history swaps the feed for "no activities".

Worth noting the AGENTS.md text added in this same commit asserts the property that is missing — "a page that comes back entirely postless is walked past rather than handed to the reader as a 'Load more' that adds nothing" is true of `loadMore` and false of the initial load.

**Fix:** run the same empty-continuation walk from the initial load so the first render is a settled answer, and gate the card on `statuses.length === 0 && !hasMore` so no later state can contradict it either.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
```diff
@@ -0,0 +1,246 @@
...
+    // the new gear's "Load more" disabled and reading "Loading..." forever.
+    isLoadingMoreRef.current = false
+    setIsLoadingMore(false)
+    setIsLoading(true)
+    // Reset before fetching, not after: the previous gear's posts would
+    // otherwise stay on screen with the previous gear's `hasMore`, and a
+    // "Load more" clicked in that window would page from the wrong offset and
+    // skip the rows in between.
+    setStatuses([])
+    setHasMore(false)
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:157]: **quality — the gear-change effect resets six pieces of state but leaves the previous gear's error banner up**

The effect deliberately resets everything *before* fetching — `generationRef`, `offsetRef`, `isLoadingMoreRef`, `isLoadingMore`, `isLoading`, `statuses`, `hasMore` — with comments explaining why each has to happen before the round trip rather than after. `error` is the one it omits: it is only ever cleared inside the initial `.then()` and inside `loadMore`'s loop.

**Concrete failure**, in exactly the scenario this file's own `describe('switching to another gear on the same mounted instance')` block sets up: gear A's load rejects, so `Failed to load activities.` is showing in the `role="alert"` paragraph. `gearId` then swaps to gear B on the same instance (both parents render the feed at a fixed position with no `key`). The next render shows gear B's `Loading...` card with gear A's error above it, and it stays for the whole round trip — clearing only if B resolves successfully. If B is a device, the reader sees a device page reporting a failure that belongs to a bike they navigated away from.

**Fix:** `setError(null)` alongside the other resets at the top of the effect.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.tsx
```diff
@@ -279,12 +314,28 @@ export const GearDetailView: FC<Props> = ({ gearId, actorHandle }) => {
         )}
       </div>
 
+      {/* Only a bike has a second view to reach. */}
       {gear.kind === 'bike' && (
+        <SectionNavSelect
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.tsx:331]: **invariant — a nested sub-nav inside a section is a second dropdown, which "Page Header & Sub-Navigation" currently forbids**

`AGENTS.md` → **Page Header & Sub-Navigation**, last bullet of the dropdown sub-nav subsection:

> A **nested** sub-nav inside a section renders as a small **in-content segmented control**, not a second dropdown or rail. Hand it to the closest section-mode `PageHeader` via `PageSubnavProvider` so it sits directly **below the per-page title** … rather than above it.

`/fitness/gear/[id]` is inside that section — `app/(timeline)/fitness/layout.tsx` already renders `<SectionNavDropdown label="Fitness" …>` immediately above `{children}` — so a bike's page now stacks two outline dropdown triggers in the same column ("Gear ▾", then "Components ▾"), and the second sits below the stat tiles rather than under the section-mode title.

The precedent this PR leans on does not cover the case: `FitnessStatusDetail`'s `SectionNav` lives at `/[actor]/[status]`, which is not a settings-style section and has no ancestor `SectionNavDropdown`, so it never met this rule. This is the repo's first nested sub-nav inside a section.

The PR also leaves `AGENTS.md` self-contradictory: the new Fitness Gear bullet says the switcher is `SectionNavSelect`, while the Page Header section still says a nested sub-nav is "not a second dropdown". A future agent grepping either section gets the opposite answer.

**Fix — pick one, don't leave both rules standing:** either render it as the in-content segmented control via `PageSubnavProvider`, or, if the dropdown is the deliberate design call, amend the Page Header & Sub-Navigation section in this same PR to record the carve-out so the two sections agree.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.tsx
```diff
@@ -279,12 +314,28 @@ export const GearDetailView: FC<Props> = ({ gearId, actorHandle }) => {
...
+      )}
+
+      {gear.kind === 'bike' && view === 'components' ? (
         <GearComponentsCard
           gearId={gear.id}
           components={components}
           onChanged={reload}
         />
+      ) : (
+        <GearActivitiesFeed
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.tsx]: **quality — every Components/Activities toggle unmounts the feed and refetches from offset 0**

The feed is the *else* branch of a ternary, so switching a bike's view to Components unmounts it, destroying `statuses`, `hasMore` and `offsetRef`; switching back mounts a fresh instance whose effect runs at offset 0.

**Concrete failure:** a rider scrolls Activities down five pages (100 posts, 5 requests), clicks **Components** to check a chain's wear, clicks **Activities** again — the 100 posts and the scroll position are gone and the feed re-requests page 1. Those requests are not cheap any more: the route calls `getStatusesByIds` for 20 ids and each status fans out its own hydration (recipients, attachments, tags, actor, counters, `status_history`, the fitness join …), so one toggle round-trip rebuilds state the client had a moment earlier.

The rest of this very component avoids exactly this — `GearDetailView` holds `components` in its own state, and its comment says tearing the card down "would close its add form and collapse its 'Show N replaced' toggle".

**Fix:** mount the feed lazily on first use and then keep it mounted, toggling visibility, so its loaded pages and offset survive a switch. (Rendering it eagerly would be worse — it would fetch for every reader who never opens the tab.)

---

app/api/v1/fitness/gear/[id]/activities/route.ts
```diff
@@ -66,11 +76,30 @@ export const GET = traceApiRoute(
...
     })
     const hasMore = rows.length > limit
+    const page = rows.slice(0, limit)
+
+    // One batched read, in the activity order the rows came back in —
+    // `getStatusesByIds` preserves the input order and drops the ids it cannot
+    // find, which is also how a post deleted between the two queries leaves.
+    // No `visibleToActorId`: every one of these is the caller's own post,
+    // reached through their own `fitness_files` rows.
+    const statusIds = page
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/activities/route.ts:86]: **quality — `getFitnessGearActivities` still LEFT JOINs `statuses` and selects six columns this route no longer reads** *(and the database test pinning that join now guards nothing)*

This route is the only consumer of `database.getFitnessGearActivities`, and after this change it reads exactly one field per row — `activity.statusId` — plus `page.length` for `nextOffset`. `lib/database/sql/fitnessGear.ts` still does the work the old compact rows needed:

- `.leftJoin('statuses', 'statuses.id', 'fitness_files.statusId')`, whose only purpose is `statuses.publicId as statusPublicId` — and `getActivityStatusHref`, the sole reader of `statusPublicId`, was deleted from `DeviceDetailView` in this diff.
- `select(… fileName, description, activityType, activityStartTime, totalDistanceMeters)` plus the per-row `getCompatibleTime` / `normalizeOptionalNumber` mapping, none of which reach the response.

So every page — including each infinite-scroll page the sentinel now fires automatically — pays a join and materialises a `description` TEXT column per row for nothing.

Two things are stale alongside it: the doc comment on `FitnessGearActivity` still says "`statusPublicId` is null for an activity that was never posted, and such a row renders unlinked", and `lib/database/sql/fitnessGear.test.ts`'s `carries the posted status through the join` describes itself as "the only assertion anywhere that a NON-null `statusPublicId` comes out of the database. Without it a wrong join or a wrong alias would render every row on a device page unlinked" — the device page renders no rows and no links any more, so that test can no longer fail for anything a user would notice, while holding the dead join in place.

**Fix:** narrow the query to `fitness_files.id` + `fitness_files.statusId`, drop the `statuses` join, and trim `FitnessGearActivity`, its doc comment and the two database tests to match.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+    )
+    fireEvent.click(loadMore)
+
+    await waitFor(() =>
+      expect(screen.getByText('Evening ride')).toBeInTheDocument()
+    )
+    expect(screen.getAllByText('Morning ride')).toHaveLength(1)
+  })
+
+  it('walks past a page whose activities have all lost their posts', async () => {
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx:251]: **test-gap — `MAX_EMPTY_CONTINUATIONS = 5` has no test; the cap can be deleted with a green suite**

`walks past a page whose activities have all lost their posts` walks exactly one empty page, and `stops walking empty pages once the server says there are no more` exits via `!page.hasMore`. Neither reaches the loop bound, so the cap — the thing the new AGENTS.md text calls load-bearing ("capped the way the bookmarks timeline caps its own continuations") — is unguarded.

**Verified:** setting `MAX_EMPTY_CONTINUATIONS` to `Number.POSITIVE_INFINITY` leaves the file 9/9 green. With the cap gone, a gear whose older history is all postless rows — the exact case this feed is built around — turns one "Load more" click into an unbounded serial walk of a five-figure history with the button stuck on "Loading...".

**Fix:** queue eight consecutive postless `hasMore: true` pages, click "Load more", and assert exactly six calls (first load + five continuations), that "Load more" is back and enabled, and that the offset advanced to the fifth page's `nextOffset`.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
```diff
@@ -0,0 +1,246 @@
...
+      )
+    } finally {
+      if (generationRef.current === generation) {
+        isLoadingMoreRef.current = false
+        setIsLoadingMore(false)
+      }
+    }
+  }, [gearId])
+
+  const { loadMoreRef } = useLoadMoreOnVisible({
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:235]: **test-gap — the IntersectionObserver auto-load path and its in-flight guard are entirely untested**

`useLoadMoreOnVisible` is new on this surface; the component it replaces had a "Load more" button and nothing else. The test file never stubs `IntersectionObserver`, and the hook short-circuits on `typeof IntersectionObserver === 'undefined'`, so under jsdom the sentinel is never observed and `onLoadMore` is only ever reached through a click. Two new behaviours therefore have zero coverage: `enabled: hasMore && !isLoading` (the sentinel must not be armed during the initial load), and the `if (isLoadingMoreRef.current) return` guard, whose own comment says it exists because the hook "re-invokes on every intersection, and state has not settled yet when it does" — a path `fireEvent.click` cannot reproduce, since RTL flushes `setIsLoadingMore(true)` (and the `disabled` attribute) between clicks.

**Verified:** deleting the `if (isLoadingMoreRef.current) return` line leaves the file 9/9 green. In production, two intersections in one tick then fire two requests at the same `offsetRef.current`, both advance the ref, and the second page's posts are skipped from the feed for good.

**Fix:** stub `IntersectionObserver` the way `BookmarksTimeline.test.tsx` does (capture the callback, define it on `globalThis` in `beforeEach`, delete in `afterEach`), then assert (a) two intersections while a page is in flight produce one request, and (b) nothing is observed while `isLoading` is true.

---

llun (commented):
Round 1 of the sub-agent code-review loop (AGENTS.md → Code Review Loop). Five reviewers ran over correctness, project invariants, tests, design/a11y and security/performance; each finding was then handed to an adversarial verifier told to refute it. 14 raw findings, 9 survived, posted here as 7 threads (two reviewers found the empty-state bug independently).

Five were refuted and are **not** posted: a duplicate of the dead-join finding aimed at code outside this diff, a complaint that the `Posts` test stub is thin (it matches the house pattern in `RecentFitnessActivities.test.tsx`), a claim that `text-primary` should have been changed in two sibling components (both pre-existing on `main` and outside this diff), a claim that a failed "Load more" reports too far from its button (pre-existing pattern, and better than the alternative), and a documentation-placement preference.

One is a genuine fork rather than a defect — the nested-dropdown thread — and I'll resolve it by amending the rule rather than the UI, since the dropdown is what the design handoff specifies. Working through all of them now.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx]: Fixed in 752991e — both halves, because each covers a case the other does not.

The walk now lives in one place, `loadPageWithPosts`, and the initial load goes through it exactly as "Load more" does: it pages forward until a page carries a post, the server says there is nothing more, or it hits the five-request cap. So the 30-activity device whose 20 newest lost their posts renders the ten that survive on first paint rather than a message contradicted by the button under it.

The `!hasMore` gate went on as well, since the walk cannot cover the `onPostDeleted` case you named — deleting the last visible post of a long history empties the list with `hasMore` still true, and no amount of walking at load time helps there.

Two new tests, plus one behavioural note. `walks the same empty pages on the FIRST load, not only on load more` fails against the old initial effect, and `never reports an empty history while the server says there is more` pins the gate. The note: the walk moved the loop out of `loadMore`, so `MAX_EMPTY_CONTINUATIONS` became `MAX_PAGES_PER_LOAD` — it now bounds requests per load rather than empty continuations after the first, which is the same five requests but says what it counts.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:157]: Fixed in 752991e. `setError(null)` now sits with the other resets at the top of the effect, and the comment above them says why it belongs there rather than in the `.then()` — a failure left behind reports a bike's outage on a device's page.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearDetailView.tsx:331]: You're right that both rules were left standing, and that is the actual defect here. Taking your option 2, because the dropdown is the design's call rather than mine: `ui_kits/web/GearKit.jsx` puts a `GKViewDropdown` on a gear's page below the stat tiles, with the section's own "Gear ▾" still above it — the two stacked triggers in the handoff screenshots are deliberate.

`AGENTS.md` in 752991e now splits the rule by what the control *does* rather than where it sits:

- a nested sub-nav that **navigates** to another route in the section stays the in-content segmented control via `PageSubnavProvider`;
- a nested sub-nav that switches a **view of the page you are already on** is `SectionNavSelect`.

The reasoning is written down with it: a segmented control reads as "more of this page", which is wrong for a control that replaces the page's whole body, and it carries no per-entry icon, which both of these designs do. There is also a new **Section sub-nav in local state** subsection documenting `SectionNavSelect` — its two consumers, the boolean `aria-current`, and the `text-primary-text` active row — so the next person writing one finds it instead of re-inlining a third copy. `REVIEW.md` picked up the matching checklist line.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearDetailView.tsx]: Fixed in 752991e, taking exactly the shape you describe: the feed mounts on first use, latched by `hasOpenedActivities`, and after that it is hidden rather than unmounted. Lazy for the reason you flag — mounting it eagerly would fetch a page of posts for every reader who only ever looks at the components.

Verified in the browser rather than only in tests, counting requests to `/activities`: 0 on the Components view, 2 after opening Activities (React's dev-mode double effect), and still 2 after switching to Components and back — where it was 4 before.

Two tests: `does not mount the feed until Activities is first opened` and `keeps the feed mounted when the reader switches back to Components`, the latter asserting the feed is in the document but not visible.

---

app/api/v1/fitness/gear/[id]/activities/route.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/activities/route.ts:86]: Fixed in 752991e, all of it. `getFitnessGearActivities` lost the `statuses` LEFT JOIN and now selects `fitness_files.id` and `fitness_files.statusId` only, with the `getCompatibleTime` / `normalizeOptionalNumber` mapping gone with the columns it was mapping. `FitnessGearActivity` is those two fields.

Its doc comment now says what the type is for — the identity of the activity and of the post it was published as — and records why the rest went, so the next person does not re-add a column "while they're in there". A new comment on the query itself explains the one thing the narrowing must not tempt anyone into: there is deliberately no `statusId` predicate, because the caller pages on rows, and filtering postless ones out would make the offsets skip them and re-serve everything behind them.

The database test is now `carries the id of the post the activity was published as`, guarding `statusId` — which is the field a wrong column or alias would silently empty every gear's feed with — and the sibling `toMatchObject` is down to `statusId: null` with a note that an unposted activity still occupies a row.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx:251]: Fixed in 752991e. `caps one load at five requests and hands back a Load more` queues ten postless `hasMore: true` pages and asserts the walk stops at five, that the button comes back enabled rather than stuck on "Loading...", that the empty message is *not* shown, and that the next click resumes at the fifth page's `nextOffset` (100) rather than restarting.

Checked it the way you checked the gap: raising `MAX_PAGES_PER_LOAD` to 500 fails exactly that test and nothing else.

The count is five per load rather than your six, because the walk now covers the initial load too, so there is no "first load plus five continuations" any more — one load is at most five requests, wherever it starts.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:235]: Fixed in 752991e with the stub pattern you pointed at, in its own `describe('the infinite-scroll sentinel')` so the `IntersectionObserver` definition is scoped to the tests that want it.

Three tests: `loads the next page when the sentinel scrolls into view`, `fires one request when the sentinel re-intersects mid-flight` (two intersections in one tick against a page still in flight, asserting one request and that the page still lands when it resolves), and `is not armed while the first page is still loading`, which asserts `observe` is never called until the initial load settles — the `enabled: hasMore && !isLoading` half.

Verified the same way you verified the gap: deleting `if (isLoadingMoreRef.current) return` fails exactly the mid-flight test and nothing else.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
```diff
@@ -0,0 +1,315 @@
...
+    )
+
+  // Three separate reasons the list can be empty, and only one of them is "this
+  // gear has no posted activities". `!hasMore` rules out a page that simply had
+  // no posts on it — the walk above makes that rare but cannot make it
+  // impossible, since it caps out and `onPostDeleted` can empty a page-one list
+  // at any time — and `!error` rules out a load that never landed, which resets
+  // both pieces of state on its way to failing and would otherwise render
+  // "no activities" under its own error message.
+  const reportsEmptyHistory = statuses.length === 0 && !hasMore && !error
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:255]: **bug — a failed first load rendered the "no activities" empty state, with no way to retry** *(found independently by the correctness and security/perf reviewers; fixed in 01de73a)*

The `!hasMore` gate added in round 1 did not cover the error case. The effect resets `statuses` to `[]` and `hasMore` to `false` *before* fetching, and the `.catch` only calls `setError` — so a 500, a timeout or any blip on the initial load rendered the alert and, directly under it, `No recent activities on this gear.` asserting something false about a gear whose Activities tile said otherwise.

Worse, there was no recovery. With `hasMore` false there is no "Load more" and no sentinel, and the effect re-runs only on a `gearId` change — so returning to the same gear did not retry it. Round 1's own fix made this reachable in a new place: the feed now stays mounted behind the switcher, so bouncing to Components and back does not remount it either. Only a full page reload cleared it.

Two changes: `reportsEmptyHistory` now requires `!error`, and a failed load with nothing on screen offers **Try again**, which bumps a `retryToken` in the effect's deps. Both mutation-checked — dropping `!error` fails `does not claim the gear has no activities` and nothing else.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+    return () => {
+      cancelled = true
+      // Unmount supersedes an in-flight `loadMore` walk too. That walk's
+      // liveness check reads `generationRef`, which only ever moves at the top
+      // of this effect — so without this it stays true for a component that no
+      // longer exists, and the walk runs its remaining requests out, each one
+      // a hydrated page read nobody will see. The token is monotonic, so the
+      // extra bump costs the next run nothing.
+      generationRef.current += 1
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.tsx:188]: **bug — unmounting did not stop a "Load more" walk; it ran its remaining four requests out** *(fixed in 01de73a)*

The two walks used different liveness signals and only one was unmount-aware. The initial load passes `() => !cancelled`, and the cleanup flips `cancelled` — correct. `loadMore` passes `() => generationRef.current === generation`, and nothing bumped `generationRef` on unmount: it only ever moved at the top of the effect body, which never runs again once the component is gone.

So a walk in flight when the reader navigated away kept going, `isCurrent()` staying true for a component that no longer existed, until it had spent its whole `MAX_PAGES_PER_LOAD` budget. Reproduced at **6** calls where 2 were expected — four `GET /api/v1/fitness/gear/:id/activities` after unmount, each running the gear query plus a batched `getStatusesByIds` hydration. Worst exactly where the walk is longest, which is the case the walk exists for. Nothing surfaced it: the trailing `setState`s are silent no-ops.

It also broke the guarantee `loadPageWithPosts` documents two lines above — "a walk superseded by a gear change stops at the next page instead of running its remaining requests out" — for unmount, which is the commoner exit.

The cleanup now bumps the token the load-more walk reads. Mutation-checked: removing that line fails `stops an in-flight walk when the feed unmounts` and nothing else.

---

app/(timeline)/fitness/gear/[id]/GearDetailView.tsx
```diff
@@ -279,14 +326,47 @@ export const GearDetailView: FC<Props> = ({ gearId, actorHandle }) => {
...
+          onChange={handleViewChange}
         />
       )}
 
+      {/* Hidden rather than unmounted, for the same reason the refetch above
+          keeps it mounted: the card holds its add form, its typed-in values,
+          its "Show N replaced" toggle and its save error in local state, and a
+          glance at Activities mid-form would otherwise throw all of it away. */}
+      {gear.kind === 'bike' && (
+        <div hidden={!showsComponents}>
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.tsx:344]: **quality — the switcher preserved the feed across a view change but destroyed the components card's form state** *(fixed in 01de73a)*

Round 1's fix kept the feed mounted and hidden; the card three lines above got the opposite treatment and was unmounted outright on every switch to Activities. `GearComponentsCard` holds all of its UI in local state — `isFormOpen`, `componentType`, `brand`, `model`, `addedMode`, `addedDate`, `serviceKm`, `showReplaced`, `error`.

Concrete failure: press **Add component**, choose "Chain", type `Shimano` / `CN-HG901-11`, glance at Activities for the ride you're servicing after, switch back — the form is closed and every typed value is gone, with no warning. The same switch collapsed "Show N replaced components" and silently discarded a save error the card was displaying. None of it was reachable before this PR, because there was no switcher.

That is precisely the loss `GearDetailView` already documents a refetch must not cause ("tearing the card down would close its add form and collapse its 'Show N replaced' toggle"), so the card now gets the same `hidden` treatment — no laziness needed on that side, since it is mounted at page load anyway. `keeps the components card mounted while the reader is on Activities` expands the replaced rows, switches away and back, and asserts they are still expanded.

---

lib/components/section-nav-dropdown.tsx
```diff
@@ -103,8 +103,14 @@ export const SectionNavDropdown: FC<SectionNavDropdownProps> = ({
...
-                    // current row doesn't drop it to the plain hover grey.
-                    'bg-primary/10 text-primary focus:bg-primary/10 focus:text-primary',
+                    // current row doesn't drop it to the plain hover grey. The
+                    // text takes `--primary-text`, not `--primary`: the latter
+                    // is the icon/fill orange and sits under the AA floor as a
+                    // foreground. Its state-driven twin `SectionNavSelect`
+                    // paints the identical row, and the pair now sits on one
+                    // screen on a gear's page, where two shades of orange
+                    // would read as a mistake.
+                    'bg-primary/10 text-primary-text focus:bg-primary/10 focus:text-primary-text',
```
llun, line 113: **quality — the two "twin" dropdowns painted their active row in different oranges, on the same screen** *(found independently by the invariants and design/a11y reviewers; fixed in 01de73a)*

Round 1's new AGENTS.md section says the pair "differ only in what a row does" and then, two bullets later, that the active row takes `text-primary-text`. Both could not be true: `SectionNavSelect` used `text-primary-text` while `SectionNavDropdown` — the sub-nav for settings, fitness *and* admin — still used `text-primary`, so the doc asserted a rule its own primary implementation broke.

That is not only a doc problem now. A bike's page carries the fitness section's "Gear ▾" above its own "Components ▾", so the two rows sit on one screen and a mismatch reads as a mistake rather than a subtlety. And the failure mode the reviewer named is real: `section-nav-select.tsx` says "keep the two in step", so an agent aligning the pair would read the twin, see `text-primary`, and copy it back — silently returning the active row to 3.37:1 on the card, under the AA floor, with lint, build and the whole suite green, because `app/globals.contrast.test.ts` guards token values and never their usage.

Flipped the twin instead of reverting mine, per AGENTS.md's own instruction to move stale orange text over when you touch it. AGENTS.md now says *both* rows take it, and why.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx
```diff
@@ -0,0 +1,697 @@
...
+import { ActorProfile } from '@/lib/types/domain/actor'
+import { Status, StatusType } from '@/lib/types/domain/status'
+
+import { GearActivitiesFeed } from './GearActivitiesFeed'
+
+vi.mock('@/lib/client', () => ({
+  getFitnessGearActivities: vi.fn()
+}))
+
+// The stub renders one delete and one edit control per post, because the feed's
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx:20]: **test-gap — `removeStatus` / `updateStatus` had no coverage, and neither did the `error` reset** *(two findings, fixed in 01de73a)*

The `Posts` stub took six props and neither data-sync callback, so the two handlers the feed passes were never executed by any test. Verified the way the reviewer did: swapping them (`onPostDeleted={updateStatus} onPostUpdated={removeStatus}`) left the file green — shipped that way, deleting a post from a gear's feed would leave it on screen and editing one would remove it.

That also left the round-1 empty-state gate half-proven, since its own comment justifies itself with "`onPostDeleted` can empty a page-one list at any time" and nothing exercised that sequence.

The stub now renders a delete and an edit control per post. Three tests: `drops a post deleted from the feed` (which also asserts the empty state then appears, closing the gate's own argument), `replaces an edited post in place rather than dropping it`, and — for the other finding — `clears the previous gear's error`, which asserts the alert is gone *while the new gear is still loading*, not merely once it resolves. Both mutations caught: swapping the callbacks fails two tests, dropping `setError(null)` fails one.

---

app/api/v1/fitness/gear/[id]/activities/route.test.ts
```diff
@@ -56,19 +58,23 @@ const activity = (
 ): FitnessGearActivity => ({
   id: 'file-1',
   statusId: 'https://llun.test/users/test1/statuses/1',
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/activities/route.test.ts:60]: **quality — the `activity()` fixture still built the columns the query dropped, and two source comments still carried the sentence this PR corrected in the docs** *(two findings, fixed in 01de73a)*

The fixture declared `statusPublicId`, `fileName`, `description`, `activityType`, `activityStartTime` and `totalDistanceMeters` after round 1 narrowed `FitnessGearActivity` to `{ id, statusId }` — stale, and no longer type-correct (test files are outside `yarn build`'s checking, so nothing caught it). Trimmed to the two fields, with the `statusPublicId: null` override in the postless-activity test gone with them.

The other half is in `lib/database/sql/fitnessGear.ts`, which kept two copies of "a device's count and its page can never disagree" — the exact claim this PR corrected in `AGENTS.md`, `REVIEW.md` and `docs/fitness-file-storage.md`, and now false, since the page renders posts and a row whose status was deleted still counts. One of them sits inside `getFitnessGearActivities` itself, on a line this PR had already rewritten around, so it was a straight miss: someone triaging a device showing "12 activities" over 11 posts would read it in the very function they were debugging and conclude the query had drifted from the rollup. Both now say what the docs say.

---

llun (commented):
Round 2, run against the round-1 fixes with the reviewers told what had already changed and asked to look specifically for what those fixes broke or newly exposed. 21 agents, 12 findings survived verification, deduplicating to 6 threads — all fixed in 01de73a, all posted here with the fix rather than left for a third round, since none of them were judgement calls.

Two were real bugs, and both were **created or newly exposed by round 1**, which is the argument for the second round existing:

- The `!hasMore` empty-state gate did not cover a failed load, so an error rendered "no activities" under its own alert — and because round 1 also made the feed survive the switcher, bouncing to Components and back no longer remounted it, leaving a page reload as the only recovery.
- The shared walk gave `loadMore` a liveness check that unmount never invalidated, so an abandoned walk spent its remaining four requests on hydrated page reads nobody would see.

Four were quality and consistency: the switcher preserved the feed but tore down the components card's form state; the two twin dropdowns painted their active row in different oranges now that they share a screen; the route's test fixture still built the columns the query dropped; and two source comments still carried the "can never disagree" sentence this PR had corrected in three docs — one of them inside the very function someone would read while debugging that discrepancy.

Four findings were refuted and are not posted: two about the `feed` context and `postLineLimit` being unasserted (real in the literal sense, but the chain is prop-threading with nothing to break), one duplicate of the components-card finding aimed at the wrong mechanism, and one about `MAX_LIMIT=100` whose premises did not hold.

Every new guard is mutation-checked: dropping the unmount bump, the `!error` exclusion, the `setError(null)` reset, or swapping the two data-sync callbacks each fails exactly the test that covers it and nothing else. Gate green at 795 files / 9174 tests, and re-verified in the browser — 0 activity requests on Components, 2 after opening Activities, still 2 after switching back and forth.

Round 3 is running.

---

app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx
```diff
@@ -0,0 +1,779 @@
...
+      await renderWithOnePost()
+
+      fireEvent.click(screen.getByRole('button', { name: 'edit status-1' }))
+
+      expect(screen.getByText('Morning ride (edited)')).toBeInTheDocument()
+      expect(screen.getByTestId('posts')).toBeInTheDocument()
+    })
+  })
+
+  describe('a load that fails', () => {
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearActivitiesFeed.test.tsx:412]: **test-gap — the whole `catch` in `loadMore` was unexercised; a failed "Load more" had no test** *(fixed in cf0c62f)*

Every `mockRejected*` in this file rejected the **first** request, which trips the initial-load catch in the effect and leaves `hasMore` false — so no button renders, no sentinel arms, and `loadMore` is never entered. Verified by replacing its catch body with a no-op: 21/21 still green.

Two behaviours went unguarded, and the second is the more serious:

1. A rejected "Load more" must surface the message and hand the button back. Swallowed, the reader clicks, watches `Loading...`, and gets the button back with nothing added and no explanation — the "Load more that adds nothing" outcome the walk exists to avoid, only silent, and with `hasMore` still true the sentinel keeps re-firing it.
2. The stale-generation guard in that catch had no test at all. Delete it and a gear-A walk that rejects *after* the reader switched to gear B calls `setError` on gear B's feed — a bike's outage reported on a device's page, which is exactly the failure the effect's `setError(null)` reset was added to prevent in round 1. It is the failure twin of the two success-path tests round 2 added, and both of those exist while this one did not.

Both now covered: `surfaces a failed Load more and hands the button back` (which also asserts the page that did land survives — a failed continuation is not a reason to discard what the reader was already reading) and `does not report the previous gear's failure over the new one`, built on the existing deferred-promise switch harness but rejecting instead of resolving. Mutation-checked: swallowing the catch fails the first and nothing else; dropping the guard fails the second and nothing else.

---

lib/components/section-nav-select.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    // Nothing navigates here, so the boolean form rather than "page".
+    expect(
+      within(menu).getByRole('menuitem', { name: 'Activities' })
+    ).toHaveAttribute('aria-current', 'true')
+    expect(
+      within(menu).getByRole('menuitem', { name: 'Components' })
+    ).not.toHaveAttribute('aria-current')
+  })
+
+  it('keeps a focus indicator on the active row, which pins its own colours', async () => {
```
llun, line 74: **test-gap — the extracted `SectionNavSelect` dropped the active-row focus-indicator assertion its twin's test carries** *(fixed in cf0c62f)*

The new suite covered the landmark name, the unmatched-id fallback, `aria-current`, `onChange` and the empty-tabs case, but never asserted anything about the active row's classes. Deleting `focus:ring-2 focus:ring-primary/50` left the whole suite green.

What would have shipped: the active row pins its own colours on focus (`focus:bg-primary/10 focus:text-primary-text`) and the shared `DropdownMenuItem` base sets `outline-hidden`, so without the ring the focused active row is pixel-identical to its resting state and has no focus affordance at all — a keyboard user arrowing down the menu watches the highlight vanish on exactly one row (WCAG 2.4.7). The source comment asserts that reasoning; nothing enforced it.

The twin already has this test, with the rationale for asserting on the utility class rather than a computed style (jsdom does not evaluate Tailwind). The point of the extraction was to keep the two in step, and the shared component now backs **two** surfaces where the private copy backed one — so the drift costs twice and was guarded on only one side.

Ported it, and pinned `text-primary-text` in the same match while there: this PR moved both rows onto that token, and per AGENTS.md nothing else in the repo catches a `text-primary` regression on a text node — `app/globals.contrast.test.ts` guards the token values, never their usage.

---

llun (commented):
Round 3. **Correctness, project invariants, design/a11y and security/performance all came back empty** — the two findings that survived were both test gaps, both from the tests reviewer, and both fixed in cf0c62f.

They are the same shape, which is worth noting as the round's actual signal: each is a guard that rounds 1 and 2 *added*, sitting there unexercised. `loadMore`'s catch — including the stale-generation guard that stops a failed gear-A walk reporting its outage over gear B — could be deleted wholesale with the suite green, because every failing test in the file rejects the first request and never reaches it. And `SectionNavSelect` shipped without the focus-ring assertion its twin has, so the extraction that was supposed to keep the pair in step dropped the one test that would have caught them drifting apart.

Both mutation-checked, as everything in this PR has been: swallowing the catch, dropping the guard, or removing the ring each fails exactly the test that now covers it and nothing else.

Gate green at 795 files / 9177 tests. Running one more round to confirm a clean one, per the loop's stop condition.

---

lib/components/section-nav-dropdown.test.tsx
```diff
@@ -93,6 +93,13 @@ describe('SectionNavDropdown', () => {
     const active = within(menu).getByRole('menuitem', { name: 'Strava' })
     expect(active.className).toMatch(/focus:bg-primary\/10/)
     expect(active.className).toMatch(/focus:ring-2/)
+    // `text-primary-text`, never `text-primary`: `--primary` is the icon
+    // orange and is under the AA floor as a foreground. This assertion and the
+    // identical one in `section-nav-select.test.tsx` are the only two things in
+    // the repo that catch that regression — `app/globals.contrast.test.ts`
+    // guards the token values, never their usage — and the pair has to hold on
+    // BOTH, since a gear's page renders the two dropdowns one above the other.
+    expect(active.className).toMatch(/text-primary-text/)
```
llun, line 102: **test-gap — the dropdown's `text-primary` → `text-primary-text` fix had no test, while its new twin pinned exactly that** *(fixed in 45ed4fd)*

Round 2 changed `SectionNavDropdown`'s active row to `text-primary-text`, and round 3 added the assertion pinning it — **to the twin**, `section-nav-select.test.tsx`, with the note that nothing else in the repo catches that regression on a text node. The note was right; the asymmetry was the bug. The file this PR actually changed the token in is the one that went unguarded, and its test is titled `keeps a focus indicator on the active row, which pins its own colours` while pinning the wash and the ring but not the colour.

Verified by mutation: reverting that line to `text-primary` leaves both nav suites and `app/globals.contrast.test.ts` green, 39/39. That is exactly as documented — the contrast test recomputes ratios from token **values**, never their usage — so lint, build and the whole suite would pass while shipping 3.37:1 orange on the current row of every settings, account, fitness and admin sub-nav, plus two visibly different oranges stacked on the gear page this PR builds. The identical regression on `SectionNavSelect` fails instantly, which is what made the gap worth closing rather than shrugging at.

One assertion, in the test that already claimed to make it. Now both halves of the pairing AGENTS.md documents are guarded.

---

llun (commented):
Round 4 — the confirmation pass, run with the reviewers told that an empty result was the expected and most useful answer. **One finding, from one of five dimensions; nothing was refuted because nothing else was raised.** Fixed in 45ed4fd.

It is the mirror of round 3's own fix, which is a fair catch: round 3 pinned `text-primary-text` on the twin and left the component this PR actually changed the token in unguarded. Reverting that line leaves 39/39 green across both nav suites and the contrast test, because the contrast test recomputes ratios from token values and never their usage — so the regression would ship silently. One assertion closes it.

Running a fifth round to get the clean one the loop stops on.

---

llun:
## Review loop complete — round 5 clean, CI green

**Round 5 returned nothing on all five dimensions**, which is the loop's stop condition (AGENTS.md → Code Review Loop: "a full sub-agent review round yields no new actionable comments"). Five rounds, well inside the 20-round cap.

| Round | Raw | Survived verification | Threads | Outcome |
| 1 | 14 | 9 | 7 | 2 bugs, 1 invariant fork, 4 quality/coverage |
| 2 | — | 12 | 6 | 2 bugs, both created or newly exposed by round 1's fixes |
| 3 | — | 2 | 2 | test gaps only; 4 of 5 dimensions empty |
| 4 | — | 1 | 1 | test gap only; 4 of 5 dimensions empty |
| 5 | 0 | 0 | 0 | **clean** |

All 16 threads addressed, replied to and resolved. Nothing is open.

### What the loop actually caught

Four user-visible bugs, and the shape of them is the argument for running more than one round — **two were introduced by the previous round's own fixes**:

1. **R1** — "No activities" rendered above an enabled "Load more" whenever a page carried no posts. Once the page carried statuses rather than rows, an empty page stopped meaning an empty history: deleting a post only nulls `fitness_files.statusId`, leaving a row that still counts.
2. **R2** — the `!hasMore` gate from R1 did not cover a *failed* load, so an error rendered "no activities" under its own alert; and because R1 also made the feed survive the switcher, there was no remount left to recover through. Only a page reload cleared it.
3. **R2** — the shared walk from R1 gave `loadMore` a liveness check that unmount never invalidated, so an abandoned walk spent its remaining four requests on hydrated page reads nobody would see.
4. **R1** — the Components/Activities toggle unmounted the feed and re-requested page one; **R2** found the same treatment owed to the components card, whose add-form state a glance at Activities was silently discarding.

Plus one genuine fork rather than a defect: `AGENTS.md` forbade a second dropdown for a nested sub-nav, and the design handoff specifies exactly that. Resolved by splitting the rule on what the control *does* — navigating to another route stays a segmented control, switching a view of the page you are on is `SectionNavSelect` — with the reasoning recorded and the shared component documented.

### Verification

Every guard this PR adds is **mutation-checked**: removing it fails one specific test and nothing else. That covers the page cap, the in-flight sentinel guard, the unmount bump, the `!error` exclusion, the `setError(null)` reset, the two data-sync callbacks, the stale-generation guard in `loadMore`'s catch, and `text-primary-text` on both nav dropdowns.

Local gate green in the required order (795 files / 9177 tests), CI green — Lint and Prettier, Build, All Tests, Schema Dump Sync, CodeQL. Browser-verified on a seeded local SQLite database: 0 activity requests on the Components view, 2 after opening Activities, still 2 after switching back and forth.

Ready for review.

---

PR #1440, opened 2026-08-15
minor: redirect Better Auth OAuth errors to an in-app /auth/error page

llun:
## Summary

Better Auth's default error page at `/api/auth/error` is a development affordance. In production it does not render at all — it 302s straight to the home timeline:

```js
// better-auth/dist/api/routes/error.mjs
if (isProduction && !options.onAPIError?.customizeDefaultErrorPage)
  return new Response(null, { status: 302, headers: { Location: `/?${queryParams}` } })
```

`AGENTS.md` already documented the symptom without naming the cause: a Mastodon client presenting a `client_id` this server no longer knows "just lands on the timeline and its login silently does nothing."

This sets `onAPIError.errorURL` to a new `/auth/error` page and renders a real explanation there.

### What now lands on `/auth/error`

Every authorize-time rejection `@better-auth/oauth-provider` raises — `invalid_client`, `client_disabled`, `unauthorized_client`, `invalid_redirect`, `invalid_request`, `invalid_request_uri`, `unsupported_response_type`, `unsupported_prompt_select_account`.

Two codes deliberately do **not** come here: `access_denied` and `invalid_scope` are both reported to the client's `redirect_uri` per RFC 6749 §4.1.2.1 (`formatErrorURL(query.redirect_uri, …)`, with no `getErrorURL` call site for either). They are mapped on the page only for a client that hands the code back by hand.

Core's `INVALID_TOKEN` can reach the page, but only narrowly: `/reset-password/:token` requires `callbackURL` (a bare hit 400s) and sends **both** failing branches to the *client's* `callbackURL`, so only an empty `callbackURL` falls back to the error page — and this instance runs its own reset flow and never links that endpoint. The entry is kept so the code is not bare if that changes.

### Three decisions worth reviewing

**The path is root-relative on purpose.** Better Auth copies `errorURL` straight into the `Location` header — `ctx.redirect` and `formatErrorURL` do no resolution, and there is no `new URL(errorURL)` anywhere in either package. A relative path keeps the visitor on the host the request arrived on. An absolute URL built from `getBaseURL()` would bounce a login started on a trusted alias domain over to `ACTIVITIES_HOST` mid-flow, the trap `/oauth/authorize` already avoids by building its sign-in redirects from the request host.

**The page never renders `error_description`, and renders the `error` code only when it is allow-listed.** Both are free text anyone can put in a hand-crafted link, and prose we did not write on our own auth card is a ready-made phishing surface. Token-shaped is *not* sufficient — `?error=Account-locked-please-call-1-800-555-0100` clears the character class and the 64-character cap while reading as ordinary prose — so the technical-detail line is gated on `isKnownAuthErrorCode`, never on `sanitizeAuthErrorCode` alone. Codes are looked up with `Object.hasOwn`, since a bare lookup answers `constructor`/`toString` with a truthy inherited function.

That allow-list gates **rendering only, never logging**: both fields are logged whatever the code is. Allow-listing bounds nothing in the log (any description can ride a mapped code — the 200/64-character caps are the real bounds), and an unmapped code is what most needs the description, since Better Auth rewrites codes it cannot classify to `UNKNOWN` while forwarding the real description verbatim.

**`onAPIError.onError` is deliberately not set.** Better Auth's router short-circuits `onError` for anything it redirects (`if (isAPIError(e) && e.status === 'FOUND') return`), which is exactly this class of failure — so it would never fire here. Setting it at all also suppresses Better Auth's own built-in logging. The error page logs what it renders instead.

### Not included: `errorCallbackURL`

`errorCallbackURL` only exists on `signIn.social` / generic-OAuth / SSO — flows that consume a provider callback. This instance configures no `socialProviders`, no `sso()` and no `genericOAuth()` (the `ACTIVITIES_AUTH_GITHUB_*` vars are explicitly deprecated in `lib/config/auth.ts`); it is an OAuth *provider*, not a consumer. Adding it now would be dead code, so it is documented in `AGENTS.md` as a requirement for whoever adds a social provider later.

### Re-verified against better-auth 1.6.29

`main` upgraded better-auth 1.6.27 → 1.6.29 (#1439) while this PR was open. Since the whole change rests on that library's internals, every load-bearing claim was re-checked against the new version after merging: the production `/?error=` fallback, the `Location`-header string concatenation with no `new URL(errorURL)`, the `status === 'FOUND'` short-circuit, the `UNKNOWN` rewrite with verbatim description forwarding, all 8 `getErrorURL` codes (and that all 10 call sites still pass a description), the `redirect_uri` routing of `access_denied`/`invalid_scope`, and the `INVALID_TOKEN` path. All unchanged, and `errorURL.test.ts` drives the real handler so it pins this behaviourally rather than by inspection.

## Screenshots

Verified in Chromium at 900×700 in both light and dark, across four cases: known code, unmapped code, injected `error_description`, and no code at all. The card matches the sibling auth pages (`/auth/signin`, `/auth/forgot-password`) — title, description, a monospace code line when the code is allow-listed, and a "Back to sign in" footer link. The `Account-locked-please-call-1-800-555-0100` lure was re-checked in the browser after the fix: generic copy, no code line.

Screenshots could not be attached from this sandboxed session (no image host reachable, and `AGENTS.md` forbids committing PR screenshots to `docs/`). The rendered text for every case is asserted in `app/(nosidebar)/auth/error/page.test.tsx` instead, including the negative cases — that the injected description, a non-token-shaped code, and a token-shaped-but-unmapped code never reach the DOM.

Live redirect behaviour against a local SQLite dev server:

```
GET /api/auth/error?error=invalid_client&error_description=client_id+is+required
  → 302 location: /auth/error?error=invalid_client&error_description=client_id+is+required

GET /api/auth/oauth2/authorize?client_id=does-not-exist&response_type=code&redirect_uri=…
  → 302 location: /auth/error?error=invalid_client&error_description=client_id+is+required
```

And the accompanying server log, confirming the description is captured for operators but never shown to visitors:

```
WARN: message: "Auth request failed and was redirected to the error page"
      errorCode: "client_disabled"
      errorDescription: "client is disabled"
```

## Review

A 4-round sub-agent review loop ran on this PR (see the review comments): 13 findings, all fixed, replied to and resolved, closing on a clean round. Notably it caught a phishing vector still open in the first commit, a complete absence of coverage on `onAPIError.errorURL`, and a regression test of mine that passed with the feature removed via a substring collision.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — 798 test files, 9224 tests, all passing; lint clean (0 errors)
- [x] Docs updated: new **Auth Error Page** section in `AGENTS.md` and a matching **Auth error page** section in `REVIEW.md`; the stale "bounces through `/api/auth/error` to the home page" notes in the OAuth Client Registrations section and in `app/(nosidebar)/oauth/authorize/page.tsx` + its test are corrected. Grepped `*.md` and `docs/` for `/api/auth/error` — no other references
- [x] If migrations changed: no migrations in this PR, so both schema dumps are untouched

https://claude.ai/code/session_01NzXk56MQYNfCTRhqJSBiHS


app/(nosidebar)/auth/error/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+      ) ?? null
+  })
+
+  return (
+    <Card>
+      <CardHeader className="text-center">
+        <CardTitle className="text-2xl">{content.title}</CardTitle>
+        <CardDescription>{content.body}</CardDescription>
+      </CardHeader>
+      {code && (
```
llun: **The phishing vector this PR exists to close survives here, just capped at 64 characters.**

The technical-detail block is gated on `code` being *token-shaped*, not on it being a *known* code. So any string matching `/^[A-Za-z0-9_-]{1,64}$/` is rendered verbatim — including codes that fall through to `AUTH_ERROR_FALLBACK`.

Failure scenario: an attacker sends a victim

```
https://instance.example/auth/error?error=Account-locked-please-call-1-800-555-0100
```

That is 41 characters, all in the allowed class. `sanitizeAuthErrorCode` returns it, `resolveAuthErrorContent` returns the generic fallback, and the page renders the instance's own auth card reading *"We couldn't complete that sign-in … contact the administrator of this instance"* with **`Account-locked-please-call-1-800-555-0100`** directly underneath. Hyphens make it read as ordinary prose; underscores work too.

That is precisely the "prose we did not write, shown on our own auth card" threat the comment at lines 31-36 cites as the reason not to render `error_description`.

The test at `page.test.tsx:54-61` only proves a code containing a *space* is dropped — it does not cover a token-shaped unknown code.

Fix: render the `<code>` block only when the code is in the allow-list, so an unmapped code shows the fallback copy and nothing else.

---

app/(nosidebar)/auth/error/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  // better-auth's own router never sees these failures — it short-circuits
+  // onError for anything it redirects — so this is the only place they are
+  // observable. Warn, not error: every one of them is a bad request from a
+  // client, not a fault in this instance.
+  logger.warn({
+    message: 'Auth request failed and was redirected to the error page',
+    errorCode: code ?? 'unknown',
+    errorDescription:
+      firstValue(params.error_description)?.slice(
```
llun, line 74: **Unauthenticated log-volume amplifier: 200 bytes of caller-chosen text per request, with no rate limit in the path.**

`dynamic = 'force-dynamic'` plus an unauthenticated route means every `GET /auth/error` runs the handler and writes a `logger.warn` record carrying up to `MAX_LOGGED_DESCRIPTION_LENGTH` characters the caller picked. Production log level defaults to `info` (`lib/utils/logger.ts`), so `warn` is always emitted, and on GCP these are billable ingested bytes. There is no `middleware.ts` rate limit in this repo.

100k unauthenticated requests → 100k `WARNING` records with ~20 MB of attacker-chosen payload.

Not an injection issue — pino JSON-escapes the value — purely a volume/cost amplifier, so this is low severity and arguably defensible as-is. But it is worth making deliberate rather than incidental, and the `error_description` earns very little here: across the `getErrorURL` call sites in `@better-auth/oauth-provider`, only `invalid_request_uri` has two distinct descriptions, while `invalid_client` uses the *same* description ("client_id is required") for two different causes — the exact ambiguity `AGENTS.md` already complains about.

Suggest gating the description on the code being allow-listed, which composes with the fix above and bounds the amplifier to a known code set.

---

lib/services/auth/auth.ts
```diff
@@ -41,6 +41,29 @@ const buildAuth = (baseURL: string) => {
...
+    // does not: a user declining consent is reported to the client's
+    // redirect_uri, per RFC 6749 §4.1.2.1.
+    //
+    // Note this cannot be paired with `onAPIError.onError` for correlation:
+    // better-auth's router skips onError for anything it redirects
+    // (`if (isAPIError(e) && e.status === 'FOUND') return`), which is exactly
+    // this class of failure, and setting onError at all suppresses better-auth's
+    // own built-in logging. The error page logs what it renders instead.
+    onAPIError: {
+      errorURL: AUTH_ERROR_PATH
```
llun, line 70: **The PR's central invariant has zero test coverage — delete these three lines and the whole suite stays green.**

Grepping for `onAPIError` / `errorURL` / `AUTH_ERROR_PATH` outside comments hits only `auth.ts` and `constants.ts`. There is no `lib/services/auth/auth.test.ts` and no existing test reaches this config. Removing it silently reverts to better-auth's built-in `/api/auth/error` — i.e. straight back to the production behaviour this PR exists to fix — with nothing failing. Nothing pins `AUTH_ERROR_PATH` to the route directory either, so renaming `app/(nosidebar)/auth/error/` also breaks the flow with a green suite.

This is testable today: `lib/services/auth/jwks.test.ts:49-67` already stands up the real better-auth instance against in-memory SQLite via `getAuth('https://test.example.com')`, and `lib/services/wellknown/index.test.ts:133` sets the precedent for pinning an auth-config constant this way (it does exactly this for `AUTH_BASE_PATH`).

Prefer an end-to-end assertion over reading `auth.options` back: drive `auth.handler` at `/api/auth/error?error=…` and assert the `Location` header. That pins the actual redirect behaviour rather than the config shape, and would also catch a future better-auth release changing how `errorURL` is consumed.

---

lib/services/auth/auth.ts
```diff
@@ -41,6 +41,29 @@ const buildAuth = (baseURL: string) => {
...
+    // error_description=...`, so a Mastodon client presenting a client_id this
+    // server no longer knows just lands on the home timeline and its sign-in
+    // looks like it silently did nothing.
+    //
+    // Every authorize-time rejection @better-auth/oauth-provider raises
+    // (invalid_client, client_disabled, unauthorized_client, invalid_redirect,
+    // invalid_request, invalid_request_uri, unsupported_response_type,
+    // unsupported_prompt_select_account) and core's expired password-reset
+    // token (INVALID_TOKEN) route through here. `access_denied` deliberately
+    // does not: a user declining consent is reported to the client's
```
llun: **This comment and the error-code map contradict each other about which codes reach the error page.**

The comment singles out `access_denied` as deliberately *not* routed here (correctly — `@better-auth/oauth-provider/dist/index.mjs:42` sends it to `query.redirect_uri`). But `invalid_scope` is in the same category and is silently mapped in `errorPage.ts:54` without explanation:

```js
// index.mjs:3914 — goes to redirect_uri, not errorURL
if (invalidScopes.length) return handleRedirect(ctx, formatErrorURL(query.redirect_uri, "invalid_scope", …))
```

Verified there is no `getErrorURL(ctx, "invalid_scope", …)` call site anywhere in the plugin, so it can never arrive via `errorURL`.

Both entries are harmless (they cover a client that forwards the code back by hand) but the two should be documented as one category, otherwise the next reader has to re-derive which of the mapped codes are actually reachable.

---

lib/services/auth/errorPage.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    { description: 'rejects a missing parameter', raw: undefined },
+    { description: 'rejects an empty value', raw: '' },
+    { description: 'rejects an empty repeated value', raw: [] },
+    {
+      description: 'rejects prose with spaces',
+      raw: 'your account is suspended'
+    },
+    { description: 'rejects markup', raw: '<img src=x onerror=alert(1)>' },
+    { description: 'rejects a value carrying its own query', raw: 'a&b=c' },
+    { description: 'rejects a code over the length cap', raw: 'a'.repeat(65) }
```
llun, line 39: **The length cap is only tested on the reject side, so an off-by-one passes.**

The table tests `'a'.repeat(65)` → `null` but never a 64-character code → kept. Mutation-proven: flipping `errorPage.ts:94` from `>` to `>=` leaves all 27 tests green. A boundary the code explicitly names deserves both sides — the cap silently shrinking would start dropping legitimate codes from the technical-detail line.

---

app/(nosidebar)/auth/error/page.tsx
```diff
@@ -0,0 +1,90 @@
...
+// `error_description` is free text better-auth puts in the query string, and
+// anyone can hand-craft a link to this page with any value in it. Log a bounded
+// slice of it so an operator can still tell two failures apart, but never render
+// it: prose we did not write, shown on our own signed-in-looking card, is a
+// ready-made phishing surface ("Your account was suspended, call ..."). The
+// visitor gets our copy for the code instead.
+const MAX_LOGGED_DESCRIPTION_LENGTH = 200
+
+const firstValue = (raw: string | string[] | undefined) =>
+  Array.isArray(raw) ? raw[0] : raw
```
llun, line 47: **This array branch is unreachable from any test.**

No test passes an array `error_description`. Mutation-proven: changing `Array.isArray(raw) ? raw[0] : raw` to return `undefined` for arrays passes the whole suite.

`sanitizeAuthErrorCode`'s array handling *is* covered in `errorPage.test.ts:16`, but that is a separate duplicate of the same logic — this copy is untested, and `?error_description=a&error_description=b` is trivially reachable by anyone crafting a link.

---

app/(nosidebar)/auth/error/page.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    await renderPage({ error: 'contact support at evil.example' })
+
+    expect(screen.queryByText(/evil\.example/)).not.toBeInTheDocument()
+    expect(
+      screen.getByText("We couldn't complete that sign-in")
+    ).toBeInTheDocument()
+  })
+
+  it('always offers a way back to sign in', async () => {
+    await renderPage({})
```
llun, line 109: **The no-params case asserts only the link, leaving both logging fallbacks and the code-block absence uncovered.**

`renderPage({})` never checks the `logger.warn` payload, so mutating `errorCode: code ?? 'unknown'` (`page.tsx:53`) and `?? null` (`page.tsx:58`) to arbitrary sentinels passes — mutation-proven.

The test also never asserts the `{code && <CardContent>}` block is *absent*, so the technical-detail line rendering `unknown` to a visitor would go unnoticed. Given the PR's stated goals are operator correlation and not echoing untrusted input, both are worth pinning here.

---

lib/services/auth/errorPage.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  ])('$description', ({ raw }) => {
+    expect(sanitizeAuthErrorCode(raw)).toBeNull()
+  })
+})
+
+describe('resolveAuthErrorContent', () => {
+  it.each([
+    { code: 'invalid_client', title: "This app isn't registered here" },
```
llun, line 47: **5 of the 11 mapped codes have no row in this table.**

`invalid_request`, `invalid_request_uri`, `invalid_scope`, `unsupported_response_type` and `unsupported_prompt_select_account` have entries in `AUTH_ERROR_CONTENT` but are never asserted. A typo in any of those keys degrades silently to the generic fallback with a green suite.

Low severity, but the table is already there — adding five rows is free, and it closes the gap for exactly the codes a reader is least likely to eyeball.

---

AGENTS.md
```diff
@@ -886,7 +886,16 @@ consistency is enforced by keeping the wiring in one place rather than per page.
...
+- **An unknown `client_id` must fail at `/oauth/authorize`, not be forwarded to Better Auth.** Better Auth's authorize endpoint answers an unregistered client with `invalid_client` / **`client_id is required`** — the same message it uses for a genuinely absent `client_id`, which makes the failure very hard to read — and then redirects to the error page, so a failed login used to look like it silently did nothing (before `onAPIError.errorURL`, better-auth's own `/api/auth/error` 302'd straight on to the home timeline in production — see **Auth Error Page** below). `app/(nosidebar)/oauth/authorize/page.tsx` validates the client (and its `redirect_uri`) up front and returns `notFound()`; keep that check ahead of the Better Auth delegation. Per RFC 6749 §4.1.2.1 an invalid `client_id`/`redirect_uri` must be reported to the user rather than redirected to the requested `redirect_uri`.
+
+## Auth Error Page
```
llun, line 891: **New convention section in `AGENTS.md` with no matching `REVIEW.md` checklist entry.**

This adds a whole `## Auth Error Page` section with five hard rules, but `REVIEW.md` is untouched and still has no auth/OAuth section at all.

That violates `AGENTS.md` → Documentation Maintenance — *"New or changed coding conventions and patterns → the matching `AGENTS.md` section, the `REVIEW.md` checklist, and (when agents need it at task start) the `CLAUDE.md` key reminders"* — restated in `REVIEW.md:404` itself: *"conventions → `AGENTS.md` + this checklist."*

The two rules most worth a checklist line are the ones a future diff can silently break: rendering `error_description`, and making `AUTH_ERROR_PATH` absolute.

---

llun (commented):
## Sub-agent review — round 1

Three sub-agents reviewed the whole diff in parallel (correctness/security, project invariants against `AGENTS.md`/`CLAUDE.md`/`REVIEW.md`, and test quality). Nine findings posted inline. The test findings are each mutation-proven — the reviewer mutated the source and confirmed the suite stayed green.

**The one that matters:** the page still renders attacker-chosen prose on the instance's own auth card (`?error=Account-locked-please-call-1-800-555-0100`), because the technical-detail line is gated on the code being *token-shaped* rather than *known*. That is the same vector this PR closes for `error_description`, surviving at 64 characters.

**Second-most important:** the PR's central invariant — `onAPIError.errorURL` — has zero test coverage. Deleting those three lines leaves the suite green and silently restores the production behaviour the PR exists to fix.

### Verified sound, no change needed

The highest-risk claim in the PR description — that a **relative** `errorURL` is safe — holds up under tracing every consumer in the installed packages. `error.mjs:371-375`, `formatErrorURL`/`getErrorURL` in the oauth-provider, `better-call/context.mjs:64-67` and `to-response.mjs` are all pure string concatenation into the `Location` header, and the state schema is `z.string()`, not `.url()`. **There is no `new URL(errorURL)` anywhere in either dist.** The two plugins that do call `new URL` on redirect-ish values (`oauth-proxy`, `mcp`) are not installed here.

Also confirmed sound and left alone:

- `sanitizeAuthErrorCode` itself — array → `[0]`, empty array → `null`, length check precedes the regex, the regex is anchored and non-global, and JS `$` does not match before a trailing `\n` so `"abc\n"` is rejected.
- The `error_description` non-render test is real, not tautological — mutating the page to render the description makes it fail as intended.
- Using a bespoke `Card` rather than the shared `ErrorPage` component is **correct**: `ErrorPage`'s `ErrorCode` union is HTTP statuses, and it renders its own full-page `min-h-dvh` wrapper that would double-center inside the `(nosidebar)` layout's already-centered `max-w-md` container.
- `text-primary-text` (not `text-primary`), server/client module boundary, config-read placement, `AUTH_ERROR_PATH` placement, test naming conventions, and the `minor:` prefix all check out.

Working through these now.

---

app/(nosidebar)/auth/error/page.tsx
llun: Fixed in `adb63cb`. Agreed — this was the real hole, and the lure reads convincingly because hyphens are in the allowed class.

Added `isKnownAuthErrorCode` to `errorPage.ts` and gated the technical-detail block on it, so an unmapped code now shows fallback copy and nothing else. `sanitizeAuthErrorCode` still runs first (it bounds what can be *logged*); the allow-list is what decides what can be *rendered*.

Covered from both directions: `page.test.tsx` now asserts the exact `Account-locked-please-call-1-800-555-0100` lure never reaches the DOM, and `errorPage.test.ts` has an `isKnownAuthErrorCode` table with the lure, an unmapped code, `null`, and the two inherited object keys.

Re-verified in a real browser at `/auth/error?error=Account-locked-please-call-1-800-555-0100` — the card renders the generic copy with no code line.

---

app/(nosidebar)/auth/error/page.tsx
llun: Fixed in `adb63cb`, taking the suggested approach — the description is now gated on the same `isKnownAuthErrorCode` allow-list.

The code itself is still logged even when unmapped, deliberately: a code we don't recognise is worth seeing, since a better-auth upgrade can introduce one, and it is bounded at 64 characters by `sanitizeAuthErrorCode`. It just never reaches the rendered card.

Agreed on the description's low diagnostic value, which is why gating rather than dropping it felt right: `invalid_request_uri` is the one code with two distinct descriptions, so keeping it for known codes retains the only case where it disambiguates anything.

Pinned by a new test asserting `errorDescription: null` for `?error=made_up_code&error_description=<500 chars>`.

---

lib/services/auth/auth.ts
llun: Fixed in `adb63cb` — added `lib/services/auth/errorURL.test.ts`, using the end-to-end approach you suggested rather than reading `auth.options` back.

It reuses the `jwks.test.ts` in-memory SQLite harness, drives the real `auth.handler`, and asserts the `Location` header for three cases: core's `/api/auth/error` endpoint, an oauth-provider authorize rejection (which builds its redirect from the same option through a different code path), and that the target is root-relative and not protocol-relative.

Mutation-verified: deleting the `onAPIError` block fails all three, restoring it passes all three.

One thing worth flagging — my first draft of the authorize-rejection assertion used `toContain` and **passed with the option removed**, because better-auth's untouched default `/api/auth/error?error=invalid_client` contains `/auth/error?error=invalid_client` as a substring. Switched to `startsWith` and noted the trap in `AGENTS.md` so the next person doesn't reintroduce it.

Also pinned `AUTH_ERROR_PATH` to the real route: the test imports the constant and the assertions run against the live handler, so renaming `app/(nosidebar)/auth/error/` without updating the constant now fails.

---

lib/services/auth/auth.ts
llun: Fixed in `adb63cb`. Confirmed your reading independently — `grep 'getErrorURL(ctx, "invalid_scope"'` over the plugin returns nothing, so it can only arrive via `redirect_uri`.

The `auth.ts` comment now names both codes as one category with the mechanism (`formatErrorURL(query.redirect_uri, …)` with no `getErrorURL` call site), and `errorPage.ts` carries a matching note on each of the two map entries pointing at the other. `AGENTS.md`'s bullet was `access_denied`-only too; it now covers both and says to keep the comment and the map agreeing.

---

lib/services/auth/errorPage.test.ts
llun: Fixed in `adb63cb` — added a `'keeps a code exactly at the length cap'` row with `'a'.repeat(64)` to the accept table, so both sides of the boundary are now pinned and the `>` → `>=` mutation fails.

---

app/(nosidebar)/auth/error/page.tsx
llun: Fixed in `adb63cb` — added `'takes the first value when error_description is repeated'`, which renders with `error_description: ['first cause', 'second cause']` and asserts the logged value is `'first cause'`. That exercises the page's own `firstValue` rather than the `sanitizeAuthErrorCode` copy.

---

app/(nosidebar)/auth/error/page.test.tsx
llun: Fixed in `adb63cb` — added `'logs the missing-parameter fallbacks and shows no technical detail'`, which asserts `{ errorCode: 'unknown', errorDescription: null }` on the log payload and that `document.querySelector('code')` is `null`.

The code-block assertion became more load-bearing than originally scoped: the block is now gated on `isKnownAuthErrorCode` rather than on `code` being truthy, so this test also guards the no-code path of that new gate.

---

lib/services/auth/errorPage.test.ts
llun: Fixed in `adb63cb` — all five added to the table, so every mapped code is now asserted.

Two of them (`invalid_request`/`invalid_request_uri` and `unsupported_response_type`/`unsupported_prompt_select_account`) deliberately share a title, so the rows also document that the map has intentional duplicates rather than a copy-paste slip.

---

AGENTS.md
llun: Fixed in `adb63cb` — added an `## Auth error page` section to `REVIEW.md`, placed between Logging and Emails.

It carries five checklist bullets covering the two you called out (never render `error_description`, keep `AUTH_ERROR_PATH` root-relative) plus the `onAPIError.errorURL` invariant itself, the `onError` trap, and the `errorCallbackURL` requirement if a social provider is ever added.

The `error_description` bullet is now stronger than the original `AGENTS.md` wording, because round 1 showed the rule as written was insufficient: it also requires the `error` **code** to be allow-listed rather than merely token-shaped, with the `Account-locked-please-call-1-800-555-0100` case named inline so the reason is not re-derived.

Left `CLAUDE.md` alone — that file's key reminders are for rules an agent needs loaded at task start, and this one is specific enough that `AGENTS.md` + `REVIEW.md` is the right pair.

---

llun (commented):
## Sub-agent review — round 2

Re-reviewed the round 1 fix commit adversarially, since fixes introduce problems. **One real finding, and it was against the round 1 fix itself.** Fixed in `43c491c`.

### The finding: the log gate was justified by a claim that doesn't hold

Round 1 suggested gating the logged `error_description` on the allow-list to "bound the amplifier to a known code set," and I took it. That reasoning is wrong, and this PR's own test proved it — a caller just appends a mapped code:

```
/auth/error?error=invalid_client&error_description=<200 bytes>   → all 200 bytes logged
```

The 200-character description cap and the 64-character code cap were always the real bounds. Allow-listing bought nothing.

Worse, it cost a real diagnostic in exactly the case that needs it most. Better Auth's own `/error` endpoint rewrites any code it cannot classify to `UNKNOWN` **while forwarding the genuine description verbatim**:

```js
// better-auth/dist/api/routes/error.mjs:363-369
const safeCode = /^['A-Za-z0-9_-]+$/.test(unsanitizedCode) ? unsanitizedCode : "UNKNOWN"
if (unsanitizedDescription) queryParams.set("error_description", unsanitizedDescription)
```

`UNKNOWN` is not in the allow-list, so the gate blanked the description for a genuine Better Auth failure — and for any rejection a future release adds that we have no copy for. It also destroyed the tell separating a real redirect (always carries a description) from a hand-crafted link (usually doesn't).

**Resolution:** log both fields for any code. Rendering stays gated on the allow-list — that is the security-relevant half and is untouched. The wrong rationale is corrected in `AGENTS.md`, `REVIEW.md`, and both source comments.

That is my second reasoning error in this PR that only mutation-testing caught, after the `toContain`/`startsWith` substring collision. Both were in the tests-and-justification layer rather than the shipped behaviour, but the pattern is worth naming.

### Also applied

- `isKnownAuthErrorCode` is now a type predicate, so `resolveAuthErrorContent` indexes the map without the `code as string` cast. Sound today, but the cast was the only thing that would have hidden an undefined title under a later `Partial<Record<…>>` retype or `noUncheckedIndexedAccess`.
- `errorURL.test.ts` now reuses `getTestSQLDatabaseWithInstance` instead of carrying a second hardcoded copy of the schema-dump path. Re-ran the mutations after the swap — still caught.

### Checked and clean

- **`AUTH_ERROR_CONTENT[code]` cannot return `undefined`** — `isKnownAuthErrorCode` requires a non-null own key, every own key maps to a defined value, and `__proto__` is not an own property of an object literal.
- **No input gets caller-controlled text onto the card.** The card renders only allow-listed `title`/`body`/`code`; metadata and the footer link are static; no other route echoes `error_description`.
- **`errorURL.test.ts` is not flaky and does not leak.** Knex pins the sqlite pool to `{min:1,max:1}`, so the schema loads into the same connection Better Auth queries; `getAuth`'s module-level LRU cannot cross files because vitest gives each file a fresh module registry; ran 3× consecutively, green each time.
- **Every new test was mutation-tested and none passes for the wrong reason** — removing `onAPIError` fails all 3 `errorURL` tests, reverting the render gate fails exactly 2 page tests, the `>`/`>=` flip fails the 64-char row, and a bare lookup for `Object.hasOwn` fails the three prototype-key rows.
- Doc claims re-verified against `node_modules`: `invalid_scope` and `access_denied` really are `redirect_uri`-routed with no `getErrorURL` call site.

`798 test files, 9224 tests` passing; lint clean; build green.

Round 3 next — it should be the clean round that closes the loop.

---

llun (commented):
## Sub-agent review — round 3

**No code defects.** Three comment/doc accuracy findings, all verified by driving the real handler rather than reading source. Fixed in `8d297be`.

Every one was prose I wrote to justify correct code — which is the kind that misleads the next reader into changing it.

### 1. "a genuine redirect always carries a description" — false

Core's `INVALID_TOKEN` redirect carries none. `redirectError` (`password.mjs:10-14`) sets only `error`, and `/error` appends the description only `if (unsanitizedDescription)`. Verified end-to-end: `GET /api/auth/reset-password/<token>?callbackURL=` → `302 …/api/auth/error?error=INVALID_TOKEN` → `302 /auth/error?error=INVALID_TOKEN`, no description on either hop.

An operator applying the documented tell would classify a genuine Better Auth redirect as a hand-crafted probe. The tell does hold for oauth-provider rejections — all eight `getErrorURL` call sites pass a description — so it now says that, scoped.

### 2. `INVALID_TOKEN` reachability was wrong

I described it as reaching this page on "a missing or expired token". It does not. `/reset-password/:token` requires `callbackURL` (`z.string()`, not optional — a bare hit 400s), and **both** failing branches call `redirectError(ctx.context, callbackURL, …)`, which redirects to the *client's* `callbackURL`. Only an **empty** `callbackURL` falls back to `${baseURL}/error` and reaches us. Compounding it, this instance runs its own reset flow (`/auth/reset-password?code=`, `app/api/v1/accounts/password/reset/`) and never links Better Auth's endpoint.

`AGENTS.md` already requires `auth.ts`'s comment and `errorPage.ts`'s map to agree on which codes are reachable — this is that rule applied to itself. Kept the map entry (the code can still arrive) and corrected the note in both places.

### 3. The `/oauth/authorize` comment argued against its own check

My rewrite claimed delegating would give "a generic 'this app isn't registered' card instead of the 404 that names the real problem." Backwards: the 404 is `ErrorPage code="404"` — *"We couldn't find that page"* — which names neither the client nor the registration, while `/auth/error`'s `invalid_client` copy is the most specific in this PR.

As written, a maintainer would conclude the check exists for user-facing diagnosis, find `/auth/error` better at it, and delete the check. The real reason is RFC 6749 §4.1.2.1: reject the client **before involving the resource owner** and never redirect to the requested `redirect_uri`. Both the source comment and the test comment now say that, with an explicit "do not remove this on the grounds that /auth/error explains invalid_client well."

### Independently verified clean this round

- The type predicate is sound; `tsc --noEmit` clean on every file this PR touches (`strict: true`, no `noUncheckedIndexedAccess`).
- The log-gate removal rationale holds: `error.mjs:363-378` confirms the `UNKNOWN` rewrite with verbatim description forwarding, and the `MAX_LOGGED_DESCRIPTION_LENGTH` comment's per-request bound is accurate.
- The harness swap introduces no coupling or ordering hazard — `getTestSQLDatabaseWithInstance` mints a fresh `:memory:` knex per call with no module-level state, ignores `TEST_DATABASE_TYPE`, and `migrate()` is the `withSchemaDumpMigrate` override that only `exec`s the SQLite dump. Commenting out `await database.migrate()` fails the authorize test with a 500, so the schema load is genuinely load-bearing.
- `errorURL.test.ts` costs ~2.6s more in isolation (it now pulls `pg` transitively); at directory scope ~0.4s, at suite scope noise. Not worth reverting.
- `AGENTS.md` and `REVIEW.md` do not contradict each other or the code — REVIEW.md is a strict subset.

`798 test files, 9224 tests` passing; lint clean; build green.

Round 4 next to confirm the loop closes.

---

llun (commented):
## Sub-agent review — round 4: CLEAN. Review loop closed.

A full round surfaced **no new actionable findings**, so the loop closes here at 4 of the 20-round maximum. All 9 threads from round 1 are replied-to and resolved; rounds 2 and 3 were self-review summaries with their fixes pushed.

### What round 4 independently re-verified

- **The narrowed "tell" claim is true.** All ten `getErrorURL` call sites in `@better-auth/oauth-provider` (lines 3876–3907, eight distinct codes) pass a non-empty description, and `getErrorURL` → `formatErrorURL` sets `error_description` unconditionally. The counter-case is equally real — `redirectError` sets only `error` — so the scoped wording is right in both directions.
- **All four sub-claims of the corrected `INVALID_TOKEN` note hold**, including that better-call turns the non-optional `callbackURL` query schema into a 400 (`endpoint.mjs:40`), and that `originCheck`'s `validateURL` returns early on a falsy URL so an empty value isn't blocked ahead of the fallback.
- **`8d297be` is provably comment/doc-only** — filtering its `.ts`/`.tsx` diff for changed lines not beginning with `//` returns nothing.
- The `onError` short-circuit, the root-relative `Location` behaviour (plus the `startsWith('/')` / `!startsWith('//')` pins closing protocol-relative), and the forward-looking `errorCallbackURL` note (`callback.mjs:55` — `errorURL ?? defaultErrorURL`, per-flow state wins) were all re-checked against `node_modules`.

### One observation, deliberately not actioned

`page.tsx:54` says an unknown client "must be reported without involving the resource owner", where RFC 6749 §4.1.2.1 literally says the server SHOULD *inform* the resource owner. Read against line 48's established use of "involving" (= putting them through sign-in/consent) it is consistent, and the comment's operative instruction is "do not remove this check" — the opposite of the round-3 failure mode. Recording it rather than spending a round on it.

### Final state

`798 test files, 9224 tests` passing. Lint 0 errors, build green, and **all 21 CI checks green** on `8d297be` — Build, Lint and Prettier, all 4 test shards, CodeQL, Schema Dump Sync, CI Success.

### Loop summary

| Round | Findings | Nature |
|---|---|---|
| 1 | 9 | Phishing vector via unmapped-but-token-shaped codes; zero coverage of `onAPIError.errorURL`; 6 test gaps; missing `REVIEW.md` section |
| 2 | 1 | The round 1 log-gate rationale did not hold — gate removed, render gate kept |
| 3 | 3 | Inaccurate comment rationales, incl. one that argued for deleting the check it documented |
| 4 | 0 | Clean |

Two of my own reasoning errors were caught only by mutation-testing (a `toContain` substring collision that passed with the feature removed, and the unsound log-gate justification). Both were in the tests-and-justification layer rather than shipped behaviour, but that is the pattern worth carrying forward: assertions and rationales need the same adversarial treatment as the code.

Ready for review.

---

app/(nosidebar)/auth/error/page.tsx
llun: Correction for the record: **this decision was reversed in round 2 (`43c491c`)** — the reply above no longer describes the shipped code.

Gating the logged description on the allow-list bounds nothing, because a caller pairs any 200-byte description with a mapped code. It also blanked the description for `UNKNOWN`, which Better Auth emits for codes it cannot classify *while forwarding the real description verbatim* — the case an operator most needs it.

Shipped behaviour: the allow-list gates **rendering only**. Both fields are logged for any code; the 200/64-character caps are the real bounds. The test cited above was replaced with one asserting the description **is** still logged for an unmapped code.

Leaving this thread resolved — flagging only so the collapsed reply isn't read as current.

---

PR #1444, opened 2026-08-15
minor: move fitness gear changing into the post menu and wire edit

llun:
The fitness activity detail page's metadata line now **reports** the gear rather than editing it, matching the design system's `FAGearRow` / `FAMoreMenu` split in `ui_kits/web/FitnessActivity.jsx`.

## What changed

**The gear on the metadata line is a link, not a control.** The owner gets a link to `/fitness/gear/<id>`; everyone else gets plain text with the kind's icon; nobody gets anything when no gear is attributed. That route is owner-scoped, so a link offered to a viewer would only ever 404 — the same constraint `BrandedDeviceLink` resolves the same way on the line below it. The lifetime distance rides in the link's `title` rather than on a line already carrying the date and the visibility.

**Changing the assignment is a "Change gear" submenu in the post's own ⋯**, listing each candidate with its lifetime total plus a "No gear" row that clears it. It reaches the shared `PostMenu` through a new `extraMenuItems` prop on `Actions`, which can only **add** to the action set — there is deliberately no way to remove or replace one of the menu's own items — and it renders after the items a compact row has displaced into the menu, so those stay nearest the row they came from. Absent entirely when the owner has nothing to pick (a submenu whose only entry is "No gear" is dead UI), and trigger-and-rows disabled while the PATCH is in flight.

**Edit and Quote now work here**, through the same `useInlineComposer` / `InlineStatusComposer` every other surface uses — closing half the gap `AGENTS.md` recorded, so a fitness activity is no longer the one post its own author cannot edit from its own page. Reply still jumps to the always-on composer in the Comments section rather than opening a second one.

## Drive-by fix

`DropdownMenuSubTrigger` was the one item primitive missing `data-[disabled]:opacity-50` / `pointer-events-none`, and this is the first `SubTrigger` in the repo ever given `disabled` — so mid-PATCH the row painted identically to the enabled one beside it while being inert. Added the classes its three sibling item primitives have always carried, with a regression assertion.

## Deliberately out of scope

`onPostDeleted` / `onLikeChanged` / `onBookmarkChanged` stay unwired on this page. `StatusBox`, the non-fitness detail page, wires none of them either; fixing only one would make the two detail surfaces disagree about what deleting a post does. `AGENTS.md` now records it as a gap they share.

## Verification

`prettier` → `lint` (0 errors) → `build` → `test` (800 files, 9257 tests, all passing).

Browser-checked against a local SQLite database with seeded gear:

- The gear link resolves to `/fitness/gear/gear-moots` with `title="Gear: Moots Routt YBB · 529.1 km"`.
- The ⋯ reads *Change gear ▸ · Quote post · Edit post · Change visibility ▸ · Change who can quote ▸ · Copy link · Delete post*.
- The submenu lists each bike with its km, `aria-checked` on the assigned one.
- A ride offers only bikes; a run offers only shoes; devices never appear.
- Changing the gear persisted to the database; an edit round-tripped and left an edit-history button.
- A logged-out fetch renders the gear as a bare `<span title="Gear: …">` — no link, no action row.
- At 390px, bookmark and react displace into the menu and sit above "Change gear".
- With the PATCH held open, "Change gear" renders at 0.5 opacity and `pointer-events: none` beside a full-opacity "Change visibility".

## Review

The change was put through a 24-agent adversarial review before this PR: five independent lenses (correctness/state, shared-component regression, a11y + design fidelity, test quality, docs/conventions) each fanned out to refuters. Exactly one finding survived — the `DropdownMenuSubTrigger` affordance above, which is fixed here.


lib/components/ui/dropdown-menu.tsx
```diff
@@ -211,7 +211,12 @@ function DropdownMenuSubTrigger({
       data-slot="dropdown-menu-sub-trigger"
       data-inset={inset}
       className={cn(
-        "focus:bg-accent focus:text-accent-foreground data-[state=open]:bg-accent data-[state=open]:text-accent-foreground [&_svg:not([class*='text-'])]:text-muted-foreground flex cursor-default items-center gap-2 rounded-sm px-2 py-1.5 text-sm outline-hidden select-none data-[inset]:pl-8 [&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg:not([class*='size-'])]:size-4",
+        // `data-[disabled]:*` matches the three sibling item primitives above.
+        // Radix marks a disabled SubTrigger inert (its onClick/onKeyDown return
+        // early) but only reports it as `aria-disabled`, so without the dim it
```
llun: This comment is factually wrong about Radix. `MenuItemImpl` emits **both** attributes for a disabled item — `node_modules/@radix-ui/react-menu/dist/index.mjs:448-449`:

```js
"aria-disabled": disabled || void 0,
"data-disabled": disabled ? "" : void 0,
```

So `data-disabled` is present on a disabled `SubTrigger` already; what was missing is any *styling* keyed to it, which is exactly what this line adds. Saying Radix "only reports it as `aria-disabled`" would send the next reader looking for a missing attribute instead of a missing rule.

The same wording is in `post-menu.test.tsx:499`.

AGENTS.md
```diff
@@ -867,15 +880,29 @@ consistency is enforced by keeping the wiring in one place rather than per page.
...
+  are wired together or not at all. Reply is the one action it routes
+  differently: this page has an always-on composer in its Comments section, and
+  the reply action jumps to that rather than opening a second one.
+  What is still unwired is `onPostDeleted`/`onLikeChanged`/`onBookmarkChanged`,
+  so a delete from the menu leaves the page showing a status that no longer
+  exists. That is a **known gap it shares with `StatusBox`**, the non-fitness
+  detail page, which wires none of them either — fix it in both or in neither,
+  or the two detail surfaces disagree about what deleting a post does.
+- **A surface may ADD an item to the `⋯`, never remove or replace one.**
+  `Actions` takes `extraMenuItems: PostMenuExtraItem[]`, forwarded to
```
llun, line 896: `REVIEW.md` is the checklist a reviewer actually works from, and it was not updated alongside this. Its "Status posts" bullet (`REVIEW.md:147-154`) still reads as a flat prohibition — "it must not pass per-status action callbacks (`onReply`/`onQuote`/`onEdit`), hide individual actions, or build a bespoke post/action row" — with no mention of `extraMenuItems` or of the detail-surface carve-out.

Half of that gap predates this PR (`StatusBox` already passed `onEdit`/`onQuote` on `main`, so the flat wording was already imprecise), but `extraMenuItems` is new here, and `AGENTS.md` → Documentation Maintenance names `REVIEW.md` among the files to keep in step. Left as-is, the next reviewer reading only the checklist would flag a sanctioned pattern.

llun (commented):
Round 1 of the sub-agent review loop: 5 lenses over the whole diff against `REVIEW.md`, 12 findings raised, 12 refuted by adversarial verification. Two of the refutations nonetheless established facts worth acting on, filed below — a comment that misstates what Radix emits, and a checklist that this PR left behind.

lib/components/ui/dropdown-menu.tsx
llun: Fixed in e80e0e8. The comment now says Radix already emits both `data-disabled` and `aria-disabled` and makes the trigger inert, and that what was missing is styling keyed to that attribute. Same correction applied to the matching comment in `post-menu.test.tsx`.

AGENTS.md
llun: Fixed in e80e0e8. `REVIEW.md`'s "Status posts" bullet now carries two sub-bullets: one naming `StatusBox` and `FitnessStatusDetail` as the detail-surface exception that legitimately passes `editable`/`onEdit`/`onQuote` (while a *feed* passing them is still the defect the rule is about), and one describing `extraMenuItems` — that it can only add, where an extra item renders relative to a compact row's displaced actions, and that there is no prop for removing or replacing one of the menu's own items.

lib/components/posts/actions/post-menu.tsx
```diff
@@ -335,22 +379,67 @@ export const PostMenu: FC<Props> = ({
...
-                  }}
-                >
-                  {item.icon}
-                  {item.label}
-                </DropdownMenuItem>
-              ))}
+              {extraItems.map((item) =>
+                item.items ? (
+                  <DropdownMenuSub key={item.key}>
+                    {/* Radix disables a *trigger* through `data-disabled`
```
llun: Round 1 fixed two copies of a false claim about Radix and missed this third one, which now asserts the negation of what `dropdown-menu.tsx:215` and `post-menu.test.tsx:498` say — and misdescribes the line directly beneath it.

"Radix disables a *trigger* through `data-disabled` rather than the `disabled` prop the items use" is backwards. `MenuSubTrigger` spreads straight into `MenuItemImpl` (`@radix-ui/react-menu` index.mjs:618), and it is `MenuItemImpl` that turns the `disabled` **prop** into the `data-disabled` **attribute** (:448-449). There is no trigger-vs-item distinction — same primitive. All three inertness gates read `props.disabled`: `onClick` (:650), the hover-open timer (:660), `onKeyDown` (:706). Line 390 passes `disabled`, i.e. exactly the prop the comment says Radix does not use here.

It is a comment, so nothing misbehaves today — but the mis-follow it invites compiles: `<DropdownMenuSubTrigger data-disabled="">` type-checks, and since `MenuItemImpl` writes its own `data-disabled` *before* spreading `...itemProps`, a hand-passed one wins. That yields a trigger the new `data-[disabled]:opacity-50` rule paints dim while none of the `props.disabled` gates fire — precisely the "menu of choices none of which respond" this comment's own second half says the disabling exists to prevent.

AGENTS.md
llun, line 896: Unrelated deletion: this branch removes the committed `<!-- BEGIN:nextjs-agent-rules -->` block from the end of `AGENTS.md`.

The block is on `main` — `git show main:AGENTS.md | grep -c nextjs-agent-rules` → 2 — and the only commit that touches it is this branch's own 3ad6b4705, which deletes it. It is vendor boilerplate that `next dev` re-appends on every run (`node_modules/next/dist/server/lib/generate-agent-files.js`), and its own text says so: *"Removing it from a diff only re-creates the uncommitted change; committing it with your work keeps the tree clean."* Someone committed it deliberately for that reason.

I stripped it after running `yarn dev` for the browser verification, on a stale belief that it was an uncommitted artifact to revert. It isn't any more. Restoring it byte-identically so this PR's diff contains only the change it is about.

(Three verifier agents refuted this as "vendor boilerplate, process preference, not a defect". I disagree on the merits: whatever the block's provenance, deleting content `main` deliberately carries — in a PR about a gear menu — is scope the diff should not have.)

llun (commented):
Round 2: 5 lenses (one aimed squarely at the round-1 fixes), 6 findings raised, 1 survived verification. Filing that one plus one the verifiers refuted three times but which I think they got wrong on the merits.

lib/components/posts/actions/post-menu.tsx
llun: Fixed in c9afed4. The comment now reads that the trigger takes the same `disabled` prop its rows do — Radix routes `SubTrigger` through the very same `MenuItemImpl` — and keeps the reason it needs it (a submenu whose every row is disabled is still openable on its own). All three comments in the PR now say the same true thing.

AGENTS.md
llun: Fixed in c9afed4. The block is restored byte-identically — `git diff main -- AGENTS.md` shows no change in that region. The PR diff is now confined to the ten files this change is actually about.

llun:
## Sub-agent review loop: complete after 3 rounds

Stopped on a clean round, well inside the 20-round cap. All threads replied to and resolved.

| Round | Lenses | Raised | Survived verification | Outcome |
|---|---|---|---|---|
| Pre-PR | 5 | 12 | 1 | `DropdownMenuSubTrigger` had no disabled affordance — fixed before opening |
| 1 | 5 | 12 | 0 | Two refuted findings nonetheless established facts worth acting on → e80e0e8 |
| 2 | 4 | 6 | 1 | A third copy of the same false Radix comment; plus an unrelated deletion I disagreed with the verifiers about → c9afed4 |
| 3 | 3 | 0 | 0 | Clean |

**What the rounds actually caught**

1. `DropdownMenuSubTrigger` was the only item primitive in the repo without `data-[disabled]:opacity-50` / `pointer-events-none`, and this PR is the first ever to pass it `disabled` — so mid-PATCH the row painted as enabled while being inert.
2. Three separate comments claimed Radix disables a `SubTrigger` via `data-disabled` rather than the `disabled` prop. Backwards: `MenuSubTrigger` spreads into the same `MenuItemImpl` that *produces* that attribute from the prop, and every inertness gate reads `props.disabled`. Round 1 fixed two, round 2 caught the third — which by then asserted the negation of its own neighbours.
3. `REVIEW.md` — the checklist reviewers actually work from — still read as a flat ban on `onEdit`/`onQuote` with no mention of `extraMenuItems`. Now documents both the detail-surface carve-out and the extension point, including that it can only add.
4. An earlier commit here deleted the committed `nextjs-agent-rules` block from `AGENTS.md` after a `yarn dev` run. Restored byte-identically; the diff is now confined to its subject.

On (4) I overrode the verifiers: three of them refuted it as vendor boilerplate and a process preference. Whatever its provenance, deleting content `main` deliberately carries, in a PR about a gear menu, is scope this diff should not have.

Gates green locally (lint 0 errors, build, 800 files / 9257 tests) and on CI (Build, all 4 test shards, Lint and Prettier, Schema Dump Sync).

PR #1446, opened 2026-08-16
minor: add Mastodon-style remote follow in both directions

llun:
## Summary

Following an account on this instance from another server dead-ended on a 404. Entering a local handle in Mastodon's "Sign in to continue" box redirects the browser to `https://your-instance/authorize_interaction?uri=…`, which this server did not serve — and neither half of the fediverse remote-follow handshake existed here.

Three changes, one per direction plus discovery:

**Inbound — `/authorize_interaction` (fixes the 404).** A new page at `app/(nosidebar)/authorize_interaction/` resolves the `uri` another server sends and renders a confirmation card with the shared `FollowAction`. It accepts every form senders use: `user@domain`, `@user@domain`, `acct:user@domain`, a profile URL, or a bare ActivityPub actor URI. Signed-out visitors bounce through `/auth/signin` and back — **before** any resolution runs, so an anonymous visitor can never drive an outbound fetch through the page. Resolution lives in a framework-free service (`lib/services/actors/resolveAccountTarget.ts`) that reuses the existing primitives (`getWebfingerSelf`, `getActorPerson`, `recordRemoteActorBestEffort`, `canFederateWithDomain`, the headless signing actor) and returns a closed set of outcomes, so the page renders distinct copy for "that isn't an address", "we can't reach that server" and "no such account" instead of one generic failure. A local target short-circuits with no network call; a status permalink (Mastodon also sends those for reply/boost intents) falls out as not-found because `getActorPerson`'s schema rejects it.

**Discovery — WebFinger.** The document now advertises the standard `http://ostatus.org/schema/1.0/subscribe` template. Mastodon guesses the path when it's absent; other implementations don't. Built from `getBaseURL()` (instance host, scheme-aware) by plain concatenation — `new URL`/`URLSearchParams` would percent-encode `{uri}` to `%7Buri%7D` and break consumers doing a literal `.replace('{uri}', …)`.

**Outbound — logged-out Follow.** A logged-out visitor on a local profile previously saw no Follow affordance at all (`FollowAction` returns `null` without a session). They now get a Follow button opening a dialog for their own handle, mirroring Mastodon's. `GET /api/v1/remote-follow` resolves their home server's subscribe template server-side — WebFinger isn't readable from the browser — and falls back to Mastodon's conventional `/authorize_interaction?uri={uri}` when that server advertises none.

### Notes for review

- **A profile URL carries exactly one `@`.** `parseAccountTarget` therefore decides URL-ness *before* handle parsing: otherwise `parseAccountHandle` splits `https://d/@user` into the nonsense handle `https://d/` @ `user` and silently sends the most common Mastodon input form to WebFinger as garbage. This was caught by a test, and the ordering is commented at the call site.
- **`GET /api/v1/remote-follow` is deliberately unauthenticated** — the whole feature exists for visitors with no account here — and read-only. The abuse surface is one outbound WebFinger GET to a user-supplied domain, the same primitive search already exposes, bounded by: strict handle/domain syntax validation (username charset, bare-hostname rule rejecting credentials/paths), the instance's federation policy, retry-disabled requests with the shared helper's timeouts and body cap, an https-only check on the constructed URL, and the requirement that `target` name an actor this server actually hosts — so it can't be repurposed into a general-purpose URL builder or redirector. The shared `request` helper delegates to `safeRemoteFetch`, which rejects private/loopback/link-local addresses (v4 and v6), pins the resolved address, and is HTTPS-only outside development — verified empirically. The repo has no inbound rate-limit layer; operators should rate-limit at the reverse proxy. Flagging the unauthenticated decision for an explicit second opinion.
- `getProfileUrlAccountHandle` moved out of `app/api/v2/search/route.ts` into `lib/utils/accountHandle.ts` as `parseProfileUrlAccountHandle` (verbatim, now shared and tested). The rest of the search route's resolution was deliberately left alone — it's interleaved with search-only concerns and reshaping it would risk a hot endpoint for little gain.

## Review

Five rounds of adversarial sub-agent review, ending clean. **Sixteen findings raised, four real, twelve refuted with evidence.** Full round-by-round detail is in the review threads and the summary comment below; the four fixes were:

1. The federation gate ran on a profile URL's own host rather than the handle's domain it actually resolved, so `https://allowed.test/@user@blocked.test` slipped past a check that correctly refused the same account written as a bare handle.
2. Dismissing the follow dialog mid-lookup still navigated the visitor off-site seconds later.
3. That guard was reopen-unsafe — a shared boolean cleared on open, so retrying revived the abandoned lookup.
4. The replacement per-attempt token missed unmount, i.e. browser and Android system Back.

Every fix carries a regression test **verified to fail against the unfixed code**, which mattered here: two of my tests initially passed against the very bugs they were written for (a mock matching `blocked.test` as a substring of the whole URL, and a dialog test that reopened only after resolving). Reverting the source and re-running is the only thing that established either was worth keeping.

## Screenshots

Verified in Chromium against a local SQLite instance with mock users. Transcribed here since they were produced by an automated run rather than uploaded through the web UI.

**Logged-out profile → follow dialog (new).** The profile now shows a **Follow** button where a logged-out visitor previously saw nothing. It opens a dialog titled *"Follow @testuser@localhost:3000"* with the body *"To follow this account, sign in to your account on whatever fediverse server you use. Enter your address and we'll send you there."*, a **Your address** field placeheld `E.g. username@mastodon.social`, and **Cancel** / **Go** — the same shape as the Mastodon box in the report. Typing `visitor@example.com` and clicking **Go** navigated the browser to:

```
https://example.com/authorize_interaction?uri=testuser%40localhost%3A3000
```

**`/authorize_interaction` confirmation card.** For another account it renders *"Follow this account | You are about to follow this account from your account on this server"* above the avatar, `@seconduser@localhost:3000`, a **Follow** button and **View profile**. For your own account it renders *"This is you | You are signed in as this account"* with the Follow button suppressed.

Error states all render at HTTP 200 with the offending `uri` shown as inert text, never a link:

| `uri` | Rendered |
| --- | --- |
| `banana` | "That doesn't look like an account" |
| `ghost@localhost:3000` | "Account not found … a link to a single post won't work" |
| `https://remote.invalid/@user/12345` | same not-found copy (status permalinks are out of scope for v1) |
| *(absent)* | "Nothing to follow" |

### Manual verification performed

```
# WebFinger advertises the template (http:// here proves getBaseURL() honours
# ACTIVITIES_INSECURE_AUTH, unlike the file's older hardcoded https:// links)
$ curl -s '.../.well-known/webfinger?resource=acct:testuser@localhost:3000' | jq '.links[-1]'
{ "rel": "http://ostatus.org/schema/1.0/subscribe",
  "template": "http://localhost:3000/authorize_interaction?uri={uri}" }

# remote-follow: fallback path, and the guards
$ curl -s '.../api/v1/remote-follow?account=example.com&target=testuser@localhost:3000'
{"url":"https://example.com/authorize_interaction?uri=testuser%40localhost%3A3000"}
$ ...?account=user:pass@evil.test&...   -> 400
$ ...?target=ghost@localhost:3000       -> 404

# signed-out bounce preserves an ampersand-bearing uri through the round trip
$ curl -s -o /dev/null -w '%{redirect_url}' '.../authorize_interaction?uri=https%3A%2F%2Fremote.test%2Fusers%2Fa%3Fb%3D1%26c%3D2'
.../auth/signin?redirectBack=%2Fauthorize_interaction%3Furi%3Dhttps%253A%252F%252F...%253Fb%253D1%2526c%253D2
```

Also driven in a browser with a lookup held open to imitate a dead instance: dismissing mid-flight leaves the visitor on the profile, and the reopened form is usable rather than stuck.

One honest gap: clicking **Follow** on the card returned 404 in local dev. That is **pre-existing and unrelated** — the ordinary profile page's own Follow button returns the identical 404 for the same local target, because the follow route fetches `https://localhost:3000/users/…` over HTTPS, which an http dev server can't serve. The card wires the same shared component, so it behaves identically to every other surface. Following a genuinely remote account needs outbound network and was not exercised here.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order — lint 0 errors, build compiled with `/authorize_interaction` and `/api/v1/remote-follow` both registered, **807 test files / 9373 tests passing**
- [x] Docs updated: `docs/features.md`, `docs/architecture.md`, and `docs/mastodon-api-compatibility.md` (Extensions); grepped `*.md` and `docs/` for the new route names
- [x] If migrations changed: n/a — no schema change, so no dump regeneration

https://claude.ai/code/session_0166m459D8LVfVEz8N7m4HLY

### DIFF
--- a/lib/components/remote-follow/remote-follow-dialog.tsx
+++ b/lib/components/remote-follow/remote-follow-dialog.tsx
@@ -1,0 +1,143 @@
+'use client'
+
+import { FC, FormEvent, useState } from 'react'
+
+import { getRemoteFollowUrl } from '@/lib/client'
+import { Button } from '@/lib/components/ui/button'
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle
+} from '@/lib/components/ui/dialog'
+import { Input } from '@/lib/components/ui/input'
+import { Label } from '@/lib/components/ui/label'
+
+const REMOTE_FOLLOW_ACCOUNT_KEY = 'activities.remote-follow-account'
+const MAX_ACCOUNT_LENGTH = 320
+
+const readRememberedAccount = () => {
+  if (typeof window === 'undefined') return ''
+  try {
+    return window.localStorage.getItem(REMOTE_FOLLOW_ACCOUNT_KEY) ?? ''
+  } catch {
+    return ''
+  }
+}
+
+const rememberAccount = (account: string) => {
+  try {
+    window.localStorage.setItem(REMOTE_FOLLOW_ACCOUNT_KEY, account)
+  } catch {
+    // Storage can be unavailable (private mode, blocked cookies); remembering
+    // the handle is a convenience, never a requirement.
+  }
+}
+
+interface RemoteFollowDialogProps {
+  targetHandle: string
+}
+
+/**
+ * The logged-out Follow affordance: a visitor with an account on another
+ * fediverse server types their own handle, and we send them to their home
+ * server's remote-follow page with this account pre-filled.
+ */
+export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
+  targetHandle
+}) => {
+  const [open, setOpen] = useState(false)
+  const [account, setAccount] = useState(readRememberedAccount)
+  const [isLoading, setIsLoading] = useState(false)
+  const [error, setError] = useState<string | null>(null)
+
+  const handleOpenChange = (nextOpen: boolean) => {
+    if (!nextOpen) setError(null)
+    setOpen(nextOpen)
+  }
+
+  const handleSubmit = async (event: FormEvent) => {
+    event.preventDefault()
+    setError(null)
+
+    const trimmedAccount = account.trim()
+    if (!trimmedAccount || /\s/.test(trimmedAccount)) {
+      setError('Enter your address, for example username@mastodon.social')
+      return
+    }
+
+    setIsLoading(true)
+    try {
+      const url = await getRemoteFollowUrl({
+        account: trimmedAccount,
+        target: targetHandle
+      })
+      rememberAccount(trimmedAccount)
+      window.location.assign(url)
+    } catch (submitError) {
+      setError(
+        submitError instanceof Error
+          ? submitError.message
+          : 'Unable to reach that server'
+      )
+      setIsLoading(false)
+    }
+  }
+
+  return (
+    <>
+      <div className="flex-shrink-0">
+        <Button type="button" onClick={() => handleOpenChange(true)}>
+          Follow
+        </Button>
+      </div>
+      <Dialog open={open} onOpenChange={handleOpenChange}>
+        <DialogContent>
+          <DialogHeader>
+            <DialogTitle>Follow @{targetHandle}</DialogTitle>
+            <DialogDescription>
+              To follow this account, sign in to your account on whatever
+              fediverse server you use. Enter your address and we&apos;ll send
+              you there.
+            </DialogDescription>
+          </DialogHeader>
+          <form onSubmit={handleSubmit}>
+            <div className="space-y-2 py-4">
+              <Label htmlFor="remote-follow-account">Your address</Label>
+              <Input
+                id="remote-follow-account"
+                autoFocus
+                autoCapitalize="none"
+                autoCorrect="off"
+                spellCheck={false}
+                maxLength={MAX_ACCOUNT_LENGTH}
+                placeholder="E.g. username@mastodon.social"
+                value={account}
+                disabled={isLoading}
+                onChange={(event) => setAccount(event.target.value)}
+              />
+              {error ? (
+                <p className="text-sm text-destructive">{error}</p>
+              ) : null}
+            </div>
+            <DialogFooter>
+              <Button
+                type="button"
+                variant="outline"
+                onClick={() => handleOpenChange(false)}
+                disabled={isLoading}
+              >
+                Cancel
+              </Button>
+              <Button type="submit" disabled={isLoading || !account.trim()}>
+                {isLoading ? 'Redirecting...' : 'Go'}
+              </Button>
+            </DialogFooter>
+          </form>
+        </DialogContent>
+      </Dialog>
+    </>
+  )
+}
--- a/lib/components/remote-follow/remote-follow-dialog.tsx
+++ b/lib/components/remote-follow/remote-follow-dialog.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { FC, FormEvent, useRef, useState } from 'react'
+import { FC, FormEvent, useEffect, useRef, useState } from 'react'
 
 import { getRemoteFollowUrl } from '@/lib/client'
 import { Button } from '@/lib/components/ui/button'
@@ -64,6 +64,17 @@ export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
   // on open is un-set by the visitor simply reopening the dialog to retry,
   // which put the abandoned navigation right back.
   const attemptRef = useRef(0)
+
+  // Leaving the page abandons the lookup too. Radix does not push a history
+  // entry, so a browser or Android system Back unmounts this without ever
+  // closing the dialog — an abandonment the visitor means exactly as much as
+  // Escape, and the only one that could still navigate them.
+  useEffect(
+    () => () => {
+      attemptRef.current += 1
+    },
+    []
+  )
 
   const handleOpenChange = (nextOpen: boolean) => {
     if (!nextOpen) {
--- a/lib/components/remote-follow/remote-follow-dialog.tsx
+++ b/lib/components/remote-follow/remote-follow-dialog.tsx
@@ -52,16 +52,22 @@ export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
   const [account, setAccount] = useState(readRememberedAccount)
   const [isLoading, setIsLoading] = useState(false)
   const [error, setError] = useState<string | null>(null)
-  // Closing the dialog abandons whatever lookup is in flight. The lookup can
-  // take as long as the outbound WebFinger request's timeout, and it resolves
-  // to a URL even when that request fails (the route degrades to the
-  // conventional path), so without this a visitor who gave up and dismissed
-  // the dialog was navigated off the page seconds later anyway.
-  const abandonedRef = useRef(false)
+  // Identifies the one lookup whose result still matters. Closing the dialog
+  // or starting another lookup bumps it, and a settling request compares the
+  // token it captured before acting — so a result the visitor walked away from
+  // can neither navigate them nor write an error into the dialog.
+  //
+  // This has to be a token rather than a boolean flag: the lookup can run for
+  // as long as the outbound WebFinger timeout and it RESOLVES even when that
+  // request fails (the route degrades to the conventional path), so an
+  // abandoned attempt always settles into a usable URL later. A flag cleared
+  // on open is un-set by the visitor simply reopening the dialog to retry,
+  // which put the abandoned navigation right back.
+  const attemptRef = useRef(0)
 
   const handleOpenChange = (nextOpen: boolean) => {
-    abandonedRef.current = !nextOpen
     if (!nextOpen) {
+      attemptRef.current += 1
       setError(null)
       setIsLoading(false)
     }
@@ -78,18 +84,18 @@ export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
       return
     }
 
-    abandonedRef.current = false
+    const attempt = ++attemptRef.current
     setIsLoading(true)
     try {
       const url = await getRemoteFollowUrl({
         account: trimmedAccount,
         target: targetHandle
       })
-      if (abandonedRef.current) return
+      if (attempt !== attemptRef.current) return
       rememberAccount(trimmedAccount)
       window.location.assign(url)
     } catch (submitError) {
-      if (abandonedRef.current) return
+      if (attempt !== attemptRef.current) return
       setError(
         submitError instanceof Error
           ? submitError.message
--- a/lib/components/remote-follow/remote-follow-dialog.tsx
+++ b/lib/components/remote-follow/remote-follow-dialog.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { FC, FormEvent, useState } from 'react'
+import { FC, FormEvent, useRef, useState } from 'react'
 
 import { getRemoteFollowUrl } from '@/lib/client'
 import { Button } from '@/lib/components/ui/button'
@@ -52,9 +52,19 @@ export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
   const [account, setAccount] = useState(readRememberedAccount)
   const [isLoading, setIsLoading] = useState(false)
   const [error, setError] = useState<string | null>(null)
+  // Closing the dialog abandons whatever lookup is in flight. The lookup can
+  // take as long as the outbound WebFinger request's timeout, and it resolves
+  // to a URL even when that request fails (the route degrades to the
+  // conventional path), so without this a visitor who gave up and dismissed
+  // the dialog was navigated off the page seconds later anyway.
+  const abandonedRef = useRef(false)
 
   const handleOpenChange = (nextOpen: boolean) => {
-    if (!nextOpen) setError(null)
+    abandonedRef.current = !nextOpen
+    if (!nextOpen) {
+      setError(null)
+      setIsLoading(false)
+    }
     setOpen(nextOpen)
   }
 
@@ -68,15 +78,18 @@ export const RemoteFollowDialog: FC<RemoteFollowDialogProps> = ({
       return
     }
 
+    abandonedRef.current = false
     setIsLoading(true)
     try {
       const url = await getRemoteFollowUrl({
         account: trimmedAccount,
         target: targetHandle
       })
+      if (abandonedRef.current) return
       rememberAccount(trimmedAccount)
       window.location.assign(url)
     } catch (submitError) {
+      if (abandonedRef.current) return
       setError(
         submitError instanceof Error
           ? submitError.message
--- a/lib/services/actors/resolveAccountTarget.ts
+++ b/lib/services/actors/resolveAccountTarget.ts
@@ -1,0 +1,185 @@
+import { getActorPerson } from '@/lib/activities/getActorPerson'
+import { getWebfingerSelf } from '@/lib/activities/getWebfingerSelf'
+import { Database } from '@/lib/database/types'
+import {
+  canFederateWithDomain,
+  isLocalFederationDomain
+} from '@/lib/services/federation/domainPolicy'
+import { getFederationSigningActorSafe } from '@/lib/services/federation/getFederationSigningActor'
+import { Actor } from '@/lib/types/domain/actor'
+import {
+  parseAccountHandle,
+  parseProfileUrlAccountHandle,
+  stripAcctPrefix
+} from '@/lib/utils/accountHandle'
+import { normalizeActorId } from '@/lib/utils/activitypub'
+
+import { recordRemoteActorBestEffort } from './refreshRemoteActor'
+
+// Anything longer than this is not an account address. The cap runs before any
+// parsing so a hostile `uri` cannot push work into the URL parser or a log
+// line; it is generous enough for the longest realistic actor URI.
+export const ACCOUNT_TARGET_MAX_LENGTH = 2048
+
+export type AccountTarget =
+  | { type: 'handle'; username: string; domain: string; acct: string }
+  | { type: 'url'; url: string }
+
+export type ResolveAccountTargetResult =
+  | { type: 'resolved'; actor: Actor }
+  | { type: 'invalid' }
+  | { type: 'forbidden' }
+  | { type: 'not-found' }
+
+/**
+ * Normalizes every form a remote server may hand to `/authorize_interaction`
+ * into either an account handle or an http(s) URL. Accepts `user@domain`,
+ * `@user@domain`, `acct:user@domain`, a profile page URL and a bare
+ * ActivityPub actor URI; rejects everything else (bare domains, non-http
+ * schemes, over-long input) rather than guessing.
+ */
+export const parseAccountTarget = (input: string): AccountTarget | null => {
+  const trimmed = input.trim()
+  if (!trimmed || trimmed.length > ACCOUNT_TARGET_MAX_LENGTH) return null
+
+  const withoutAcct = stripAcctPrefix(trimmed)
+  if (!withoutAcct) return null
+
+  // URL-ness is decided FIRST. A profile URL carries exactly one `@`
+  // (`https://d/@user`), which `parseAccountHandle` would otherwise split into
+  // the nonsense handle `https://d/` @ `user` — silently sending the most
+  // common Mastodon input form to WebFinger as garbage. Anything that parses
+  // as a URL is therefore never reconsidered as a handle; a non-http scheme is
+  // rejected outright rather than falling through.
+  try {
+    const url = new URL(withoutAcct)
+    if (url.protocol !== 'https:' && url.protocol !== 'http:') return null
+    return { type: 'url', url: withoutAcct }
+  } catch {
+    // Not a URL — fall through to handle parsing.
+  }
+
+  const handle = parseAccountHandle(withoutAcct)
+  if (!handle) return null
+
+  return {
+    type: 'handle',
+    username: handle.username,
+    domain: handle.domain,
+    acct: `${handle.username}@${handle.domain}`
+  }
+}
+
+const resolveHandle = async ({
+  database,
+  username,
+  domain,
+  acct,
+  signingActor
+}: {
+  database: Database
+  username: string
+  domain: string
+  acct: string
+  signingActor?: Actor
+}): Promise<ResolveAccountTargetResult> => {
+  const persistedActor = await database.getActorFromUsername({
+    username,
+    domain
+  })
+  if (persistedActor) return { type: 'resolved', actor: persistedActor }
+
+  // Never WebFinger ourselves: an unknown username on a domain this instance
+  // serves is simply not an account here.
+  if (await isLocalFederationDomain(database, domain)) {
+    return { type: 'not-found' }
+  }
+
+  const actorId = await getWebfingerSelf({ account: acct })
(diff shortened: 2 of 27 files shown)


lib/services/actors/resolveAccountTarget.ts
```diff
@@ -0,0 +1,185 @@
...
+      acct: target.acct,
+      signingActor
+    })
+  }
+
+  // A known id (local actors included) resolves with no outbound request.
+  const knownActor = await database.getActorFromId({ id: target.url })
+  if (knownActor) return { type: 'resolved', actor: knownActor }
+
+  const profileHandle = parseProfileUrlAccountHandle(target.url)
```
llun, line 173: **Federation-policy bypass via the profile-URL branch.**

The URL branch gates on the URL's *own host* (line 135), but `parseProfileUrlAccountHandle` returns the **embedded** domain when the path segment itself contains an `@` — so the handle handed to `resolveHandle` here can be on a completely different domain that was never gated.

Concretely, with `blocked.test` domain-blocked (or an allowlist that admits only `mastodon.social`):

- `uri=spammer@blocked.test` → correctly `forbidden` at line 135. ✅
- `uri=https://mastodon.social/@spammer@blocked.test` → line 135 gates `mastodon.social` and **passes**; line 158 then extracts `{username: 'spammer', domain: 'blocked.test'}` and `resolveHandle` proceeds. ❌

Two consequences: an actor on a blocked domain that was cached before the block gets rendered with a Follow button, and an uncached one causes an outbound `GET https://blocked.test/.well-known/webfinger?...` from the instance's IP. Any allowlisted hostname works as the wrapper, since the URL host is only ever used for the gate.

The same-account asymmetry between the two input forms is the tell. Note the file already re-gates in the two sibling paths — the WebFinger result at line 103 and the canonical actor id at line 175 — so this is the one path that breaks its own invariant.

Fix: gate inside `resolveHandle` itself, so the check always guards the domain about to be contacted regardless of which call site got there.

---

llun (commented):
Sub-agent code review, round 1 (three reviewers: security, correctness, project conventions; every finding then put through two independent adversarial verifiers).

**8 findings raised, 1 survived verification** — the federation-gate bypass in the profile-URL branch, inline below. Its refuter could not refute it and reproduced it by execution; the impact judge rated it low-severity (every downstream state change is gated again — `POST /api/v1/accounts/:id/follow` 403s, and `recordActorIfNeeded` gates before persisting) but still called for the re-gate. Fixing it, since the input asymmetry breaks an invariant the file otherwise holds.

The 7 refuted findings, recorded so they are not re-litigated in later rounds:

- **SSRF via the unauthenticated WebFinger lookup** — refuted, and this corrects the PR description. `request()` delegates to `safeRemoteFetch`, which resolves DNS and rejects unsafe addresses (`isUnsafeIpv4`/`isUnsafeIpv6` cover `10/8`, `127/8`, `169.254/16`, `172.16/12`, `192.168/16`, CGNAT, multicast and the IPv6 equivalents), pins the resolved address, and is HTTPS-only outside development. I verified separately that `user@169.254.169.254`, `user@10.0.0.1`, `user@192.168.1.1` and `user@127.0.0.1` all return null.
- **Open redirect via a hostile subscribe template** — refuted: the destination is the domain the visitor typed themselves, and pointing elsewhere requires compromising that server, which is strictly less useful to an attacker than the fallback they already control.
- **Port smuggling through `parseBareDomain`** — refuted: a non-default port is preserved deliberately, and the capability is not new.
- **XSS/phishing via the rendered `uri`** — refuted: React escapes it into a `<code>` text node, never a link or `dangerouslySetInnerHTML`.
- **Remote profiles get an unusable dialog** — refuted: `getProfileData` returns null for remote actors when logged out, so a logged-out visitor never reaches a remote profile page.
- **Duplicated subscribe-rel constant** — refuted: the two constants encode different protocol facts (what this instance emits vs. what it reads from others), and drift is caught by existing tests.
- **Duplicated length cap on the page** — refuted: the page imports the same exported `ACCOUNT_TARGET_MAX_LENGTH`.

---

lib/services/actors/resolveAccountTarget.ts
llun: Fixed in d0888bd.

The gate moved **into `resolveHandle`**, so it now covers the domain about to be contacted regardless of which call site got there — matching what the two sibling paths in the same file already did. A URL is still gated on its own host before anything is fetched from it, so the wrapper is checked too; the difference is that the embedded domain no longer inherits the wrapper's verdict.

Worth recording how close this came to shipping with a green test: **my first regression test passed against the unfixed code.** The mock matched `canFederateWithDomain` on the raw string, and `https://mastodon.social/@spammer@blocked.test` *contains* `blocked.test` as a substring — so the buggy path returned `forbidden` for entirely the wrong reason. The mock now resolves the value to a host the way the real gate does. Verified the honest way, by reverting the fix and re-running:

```
=== against the BUGGY version ===
 × refuses a profile url carrying a handle on a domain it cannot federate with
 × refuses both spellings of an account on a domain it cannot federate with
   Tests  2 failed | 29 passed (31)
=== with the fix ===
   Tests  31 passed (31)
```

Three cases added: the wrapped-profile-URL refusal, an equivalence assertion that both spellings of the same account get the same verdict (the asymmetry was the actual defect), and one confirming a profile URL whose embedded domain *is* allowed still resolves.

Also folded in the one thing both verifiers suggested while refuting their own finding: the visitor's domain in `/api/v1/remote-follow` now goes through `parseBareDomain` for both input forms. That applies the bare-hostname rule to a handle's domain as well, drops an equality check that rejected every internationalized handle, and spells the outbound `acct:` resource in punycode (`bücher.test` → `xn--bcher-kva.test`) the way the remote server actually serves it.

Full gate green: lint 0 errors, build compiled, 807 test files / 9367 tests passing.

---

lib/components/remote-follow/remote-follow-dialog.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  const handleSubmit = async (event: FormEvent) => {
+    event.preventDefault()
+    setError(null)
+
+    const trimmedAccount = account.trim()
+    if (!trimmedAccount || /\s/.test(trimmedAccount)) {
+      setError('Enter your address, for example username@mastodon.social')
+      return
```
llun, line 95: **Dismissing the dialog does not stop the navigation.**

`Cancel` is disabled while the lookup is in flight, but that is not the only way out: `DialogContent` defaults to `showCloseButton`, and Radix also closes on Escape — neither is guarded. `handleOpenChange(false)` clears only `error`, so when the promise settles, `window.location.assign(url)` runs regardless of whether the dialog is still open.

The window this is open for is not small, and the slow path ends in a *navigation* rather than an error:

- the lookup is bounded by the outbound WebFinger request timeout (`network.requestTimeoutMs`, default 10s), and
- a failed lookup still **resolves** — the route degrades to the conventional `/authorize_interaction` path and returns 200 — so the dead-instance case ends in `assign`, not in the `catch`.

Concretely: a logged-out visitor types a handle at an instance that resolves but never answers (routine on the fediverse), waits, gives up, presses Escape, and carries on reading the profile. Up to ten seconds later the browser navigates them off to that dead host. Same thing in the sub-second case if they dismiss because they spotted a typo.

Fix: mark the attempt abandoned on close, and check that before navigating or setting an error. Clearing `isLoading` there too keeps a reopened dialog usable.

---

llun (commented):
Sub-agent code review, round 2. Two reviewers this time: one aimed squarely at whether the round-1 fix regressed anything, one fresh-eyes pass over the whole diff with the seven refuted round-1 findings ruled out so they could not be re-litigated.

**The fix-regression reviewer returned zero findings.** It enumerated every route into `resolveHandle` and specifically checked the case I was most worried about — that gating a *local* handle could now refuse it, which would break the page's most common use — and confirmed `canFederateWithDomain` short-circuits to `true` for local domains. I verified that end to end in a browser too: following a local account, viewing yourself, and the local actor-URI form all still render the card.

**Fresh eyes raised 4, of which 1 survived: the in-flight dismissal bug, inline below.** Its two verifiers split, and the disagreement is worth recording. The refuter argued the residue is a UX preference and correctly demolished half the original claim — the form is *not* permanently stuck, since `isLoading` clears once the bounded request settles. The impact verifier reproduced the other half precisely and rejected it as a real defect: the X button is never disabled, Escape is unguarded, and the slow path *resolves* rather than erroring, so the visitor really is navigated away after dismissing. I side with the impact verifier — dismissing a dialog should cancel what it started — and fixed it in 5cc5dcc.

The 3 refuted findings, recorded so later rounds skip them:

- **`throwApiError` surfaces bare HTTP reason phrases like "Bad Request" instead of friendly copy** — refuted: the premise that the human-readable fallback is unreachable is false (it fires for any non-JSON error body), and preferring the API's own `error` field is the documented Mastodon-shaped convention every other caller in `lib/client.ts` follows. What remains is error-copy polish.
- **The error paragraph lacks `role="alert"`** — refuted as pre-existing house style, and the premise was wrong: 55 sibling error paragraphs across `app/` and `lib/` use the identical roleless form, including the near-identical `AddActorDialog` on `main`. Worth doing repo-wide, not here.
- **Unguarded `response.json()` on the success path** — refuted: `lib/client.ts` does this in ~100 places on `main`, the route only ever emits JSON, and a `SyntaxError` is still caught and rendered rather than crashing.

Two rounds have now produced two confirmed findings, both fixed with regression tests that I verified fail against the unfixed code. CI green at every step; full suite 807 files / 9369 tests.

---

lib/components/remote-follow/remote-follow-dialog.tsx
llun: Fixed in 5cc5dcc.

Closing now marks the attempt abandoned, and both settle paths check that before doing anything: no navigation, and no error written back into a dialog the visitor already dismissed. Closing also clears `isLoading`, so reopening gives a usable form rather than one waiting on a request nobody is watching.

Verified by reverting the fix and re-running, same as round 1:

```
=== without cancellation ===
 × does not navigate when the dialog is dismissed while the lookup is in flight
 × leaves the form usable after a dismissed lookup resolves
   Tests  2 failed | 6 passed (8)
=== with the fix ===
   Tests  8 passed (8)
```

Both tests hold the lookup open with an unresolved promise, dismiss via Escape, *then* resolve it — the step whose absence made the original claim's second half ("permanently disabled") wrong.

On the verifiers' disagreement: the refuter was right that `isLoading` is not a leaked latch and the form is not permanently stuck, and I kept that correction rather than the reviewer's stronger claim. But "the request is bounded, so it eventually settles" is an argument about the form recovering, not about where the browser ends up — and it settles into the navigation, which is the part that matters. A dialog that keeps acting after you dismiss it is a bug regardless of how long it takes.

---

lib/components/remote-follow/remote-follow-dialog.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const [error, setError] = useState<string | null>(null)
+  // Identifies the one lookup whose result still matters. Closing the dialog
+  // or starting another lookup bumps it, and a settling request compares the
+  // token it captured before acting — so a result the visitor walked away from
+  // can neither navigate them nor write an error into the dialog.
+  //
+  // This has to be a token rather than a boolean flag: the lookup can run for
+  // as long as the outbound WebFinger timeout and it RESOLVES even when that
+  // request fails (the route degrades to the conventional path), so an
+  // abandoned attempt always settles into a usable URL later. A flag cleared
```
llun, line 63: **The round-2 abandon guard does not survive a reopen — the fix was incomplete.**

`abandonedRef.current = !nextOpen` sets the flag back to **false** on open, and the Follow button reopens through this same handler. So the sequence a visitor actually performs after giving up — Escape, then click Follow again to retry — clears the abandonment of the lookup still in flight. It settles, the guard passes, and `window.location.assign` fires. **No second submit required.**

The same shared flag breaks the two-attempt case: `abandonedRef.current = false` at submit un-abandons the *earlier* attempt too, so an abandoned lookup can beat the one the visitor actually wants and send them to the server they gave up on — with `rememberAccount` persisting that wrong handle for next time. And a superseded attempt that rejects writes a stale error into a dialog that has moved on.

Round 2's regression test misses all of this because it reopens the dialog only *after* resolving the lookup, so the flag is never observed in its cleared state while a request is live.

Fix: a per-attempt token rather than a shared boolean — capture an id at submit, compare it after the await. Closing bumps it; reopening must deliberately **not** reset it.

---

llun (commented):
Sub-agent code review, round 3 — and it caught that **my round-2 fix was incomplete**, which is the best argument for running the loop past the first clean-looking pass.

Both reviewers independently landed on the same defect, and unlike rounds 1 and 2 the verifiers did **not** split: all four returned `refuted: false`, each having reproduced it against the real component. The round-2 guard was a single boolean cleared on open, so reopening the dialog to retry — the natural next move after giving up — re-armed the very lookup that had just been abandoned. It then settled and navigated the visitor away *without a second submit at all*. The same flag also let an abandoned attempt beat the retried one and persist the wrong handle to `localStorage`.

What makes this worth dwelling on: **round 2's regression test passed against the broken code**, because it reopened the dialog only *after* resolving the lookup, never observing the flag cleared while a request was live. That is the second time in this PR a test agreed with a bug — the first was the substring-matching federation mock in round 1. Both times the test was written from the fix's point of view rather than the failure's.

Fixed in cc9a06b with a per-attempt token: closing bumps it, each submit captures a fresh one, and a settling request acts only if it is still current. Reopening deliberately does not reset it, which is exactly what the boolean got wrong. Three orderings now covered, each verified to fail against the round-2 implementation:

```
=== against the round-2 shared boolean ===
 × does not navigate when a dismissed lookup settles after the dialog is reopened
 × sends the visitor to the retried address, not the abandoned one
 × does not show an error from a superseded lookup
   Tests  3 failed | 8 passed (11)
=== with the token ===
   Tests  11 passed (11)
```

Also confirmed in a real browser against the running dev server, with the lookup held open to simulate a dead instance: dismiss mid-flight → the visitor stays on the profile, and the reopened form is usable rather than stuck.

Round 3's other reviewer swept the whole diff for anything the first two rounds missed and found nothing beyond this same defect — no findings on the WebFinger document (I separately confirmed this instance's own JRD round-trips through the Zod schema it uses to parse *remote* documents, subscribe link included), the moved search-route parser, the docs, or the test suites.

Full gate: lint 0 errors, build compiled, 807 test files / 9372 tests. Running one more round to confirm a clean pass.

---

lib/components/remote-follow/remote-follow-dialog.tsx
llun: Fixed in cc9a06b — replaced the boolean with a per-attempt token, exactly as suggested.

`attemptRef` identifies the one lookup whose result still matters. Closing bumps it, each submit takes a fresh one, and a settling request compares the token it captured before doing anything. Reopening deliberately does **not** touch it, which is the specific thing the boolean got wrong.

That covers all three orderings in one mechanism:

| sequence | before | now |
| --- | --- | --- |
| submit → Escape → reopen → first settles | navigates away | ignored |
| submit → Escape → reopen → submit again → first settles first | abandoned server wins, wrong handle remembered | retried address wins |
| superseded attempt rejects | stale error over a live attempt | ignored |

```
=== against the round-2 shared boolean ===
 × does not navigate when a dismissed lookup settles after the dialog is reopened
 × sends the visitor to the retried address, not the abandoned one
 × does not show an error from a superseded lookup
   Tests  3 failed | 8 passed (11)
=== with the token ===
   Tests  11 passed (11)
```

The point that stings and is worth keeping in the record: round 2's test passed against round 2's own bug, because it reopened the dialog only *after* resolving the lookup — so it never observed the flag in its cleared state while a request was live. Same failure mode as the round-1 mock that matched a substring of the whole URL. Both times I wrote the test to describe the fix rather than to reproduce the failure, which is why the revert-and-rerun check is the only thing that actually established either test was worth anything.

Also re-verified in a real browser with the lookup held open to imitate a dead instance: dismiss mid-flight, and the visitor stays on the profile with a usable form on reopen.

---

lib/components/remote-follow/remote-follow-dialog.tsx
```diff
@@ -0,0 +1,173 @@
...
+  // on open is un-set by the visitor simply reopening the dialog to retry,
+  // which put the abandoned navigation right back.
+  const attemptRef = useRef(0)
```
llun, line 66: **The token is bumped on close and on submit, but not on unmount.**

Radix pushes no history entry and does not intercept popstate, so browser Back, Alt+Left, swipe-back and **Android system back** all unmount this component without `onOpenChange(false)` ever firing. Nothing bumps `attemptRef`, the captured token still matches, and the settled lookup navigates the visitor away from the page they went back to — with `rememberAccount` writing their address on the way out.

Reproduced with two probes differing only in the abandonment gesture: Escape → `assign` calls `[]`; `unmount()` → `assign` called with the abandoned URL.

It matters more than it looks because of who this dialog is for: it is the logged-out Follow affordance, so its audience arrives from external links and skews mobile, where system back *is* how you dismiss a modal.

Fix: bump the token from an unmount cleanup. It can only ever run when there is no component left to serve, so it cannot cancel a live attempt.

---

llun (commented):
Sub-agent code review, round 4. One reviewer tasked purely with breaking the round-3 token (told to assume it was wrong, since the two previous attempts at that guard were), one closing sweep over everything else.

**2 findings, 1 survived** — the unmount gap, inline below, fixed in 271b5d5.

The verifiers split, and the dissent is worth recording because it is a good argument. The refuter showed the token still *discriminates correctly* across an unmount: it ran the round-3 ordering plus unmount and confirmed the abandoned/typo'd lookup produced no navigation and no localStorage write, while only the un-superseded attempt acted. So what completes after Back is the address the visitor typed and pressed Go on — the requested action arriving late, not the wrong server winning. It also noted no component in this repo guards unmount this way. Fixed regardless: Back is an abandonment signal exactly like Escape, and the guard's own comment already claims to cover "a result the visitor walked away from". A three-line cleanup that can only run when there is no component left to serve is a cheap way to make the comment true.

The other finding — that the URL-form federation gate in `resolveAccountTarget` has no test pinning it — was refuted by both verifiers, and the reasoning is one I want on the record because it looks superficially like the round-1 situation and is not. Both verifiers confirmed the gate *works* (input `https://blocked.test/users/spammer` returns `forbidden` and never calls `getActorPerson`) and that deleting it leaves 351 tests green. But round 1 was a test passing against **broken** behavior; this is a test suite not pinning **correct** behavior. Untested-and-correct and tested-and-broken are different classes of problem. I added the coverage anyway as part of the round-1 work — the two profile-URL cases go through the `resolveHandle` gate — so what is unpinned is specifically the actor-URI path's own host check, whose removal would still be caught downstream by the canonical-id re-gate. Noted as a hardening nit rather than a defect.

I also ran my own probe set for the failure mode that would be worst here — a *missed* navigation, where the visitor clicks Go and nothing happens — covering plain submit, retry after a dismissed attempt, close/reopen without submitting, recovery after a failed attempt, and a validation rejection mid-session. All five navigate correctly; the token breaks nothing legitimate.

Full gate: lint 0 errors, build compiled, 807 test files / 9373 tests. Running round 5 to confirm a clean pass.

---

lib/components/remote-follow/remote-follow-dialog.tsx
llun: Fixed in 271b5d5 — the token is now bumped from an unmount cleanup, so leaving the page abandons the lookup the same way closing the dialog does.

```
=== without the unmount cleanup ===
 × does not navigate when the page is left while the lookup is in flight
   Tests  1 failed | 11 passed (12)
=== with it ===
   Tests  12 passed (12)
```

Fixed despite the split verdict, and the dissent deserves credit rather than dismissal: the refuter demonstrated that after an unmount the token still picks the *right* attempt — an abandoned or superseded lookup stays suppressed, so what completes is the address the visitor typed and pressed Go on, not the one they gave up on. That is a materially weaker failure than rounds 2 and 3, and the observation that nothing else in this repo guards unmount is accurate.

What decided it: the guard's own comment says a result the visitor walked away from can neither navigate them nor write an error, and Back is walking away. Either the comment or the code had to change, and three lines that only run when there is no component left to serve is the cheaper of the two. It also cannot cancel a live attempt by construction, so there is no missed-navigation risk to trade against — I checked that separately across five orderings (plain submit, retry after dismissal, close/reopen without submitting, recovery after failure, validation rejection mid-session), all of which still navigate.

That closes the whole abandonment surface: Escape, the X button, Cancel, a superseding submit, and now leaving the page.

---

llun:
## Sub-agent code review: round 5 clean, loop complete

Two independent final passes — one over the server side (page, resolution service, public route, WebFinger emission, moved parser), one over the client side (dialog, client function, profile wiring, card) — each driving the real code rather than reading it. **Both returned zero findings.** Loop terminates here per AGENTS.md, at 5 of the 20 permitted rounds.

### What the five rounds found

| Round | Raised | Survived | Defect |
| --- | --- | --- | --- |
| 1 | 8 | 1 | Federation gate ran on a profile URL's own host, not the handle's domain it actually resolved |
| 2 | 4 | 1 | Dismissing the dialog mid-lookup still navigated the visitor off-site |
| 3 | 2 | 2 (same root cause) | That guard was reopen-unsafe — a shared boolean cleared on open |
| 4 | 2 | 1 | The replacement token missed unmount (browser/Android Back) |
| 5 | 0 | 0 | — |

Sixteen findings raised, four real, twelve refuted with evidence. Each fix carries a regression test **verified to fail against the unfixed code**, because that check turned out to matter more than anything else here:

- **Round 1's test passed against the bug.** The mock matched `canFederateWithDomain` on the raw string, and `https://mastodon.social/@spammer@blocked.test` contains `blocked.test` as a substring — so the broken path returned `forbidden` for the wrong reason.
- **Round 2's test passed against round 2's own bug.** It reopened the dialog only after resolving the lookup, so it never observed the flag cleared while a request was live — which is exactly the state round 3 found.

Both times the test described the fix rather than reproducing the failure. Reverting the source and re-running is the only thing that established either test was worth keeping.

The dissents were useful too, not just noise. Round 2's refuter correctly killed half of that finding (the form is not permanently stuck), and I kept the correction rather than the reviewer's stronger claim. Round 4's refuter showed the token still picks the right attempt across an unmount, which is why that fix is described as closing a stated-intent gap rather than a wrong-destination bug.

### Where it landed

The abandonment surface is now closed on every gesture: Escape, the X button, Cancel, a superseding submit, and leaving the page. The federation gate sits on the domain being resolved, wherever the caller came from. And both spellings of the same account — bare handle and wrapped profile URL — get the same verdict, which is the invariant round 1 broke.

Final state: lint 0 errors, build compiled with `/authorize_interaction` and `/api/v1/remote-follow` registered, **807 test files / 9373 tests passing**, CI green. Verified end to end in a real browser at each stage, including the dead-instance dismissal case with the lookup held open.

---

PR #1448, opened 2026-08-16
minor: give every gear a product page link, not just devices

llun:
## Summary

`fitness_gears.productUrl` was device-only: `POST /api/v1/fitness/gear` never accepted it, `PATCH` answered `422 productUrl is only valid for a device`, and only `DeviceDetailView` rendered it. A bike and a pair of shoes have a manufacturer's page worth linking to just as much, and the design's gear page shows one under the default-sports line. This lifts the restriction and gives every kind the same field and the same link.

**API / service**

- `CreateGearRequest` now takes `productUrl` and `POST /api/v1/fitness/gear` stores it, so a bike can be added with its product page in one step.
- `getGearKindFieldError` no longer rejects `productUrl` for a bike or shoes, and the field is gone from its `fields` type since nothing reads it there. `bikeType`/`weightKilograms` (bike-only) and `alertDistanceMeters` (shoes-only) are untouched.
- The http(s)-only refinement is unchanged and now covers all three kinds: the value becomes an `href`, so `javascript:`/`data:` there would be a script injection, and a bare `garmin.com` would render as a link back to this instance.
- `UpdateFitnessGearInput` drops its device-only `productUrl` intersection and inherits the field from `CreateFitnessGearInput`.

**UI**

- New shared `GearProductLink` (`app/(timeline)/fitness/gear/`) is the single implementation everywhere a product page is shown — a bike's page, a pair of shoes', a device's, and the gear list's device table, which previously carried its own copy of the same anchor. Hostname only (`moots.com`, not the whole URL), leading `ExternalLink`, `target="_blank"` + `rel="noopener noreferrer"`, and an `aria-label` so the link is not announced as a bare domain.
  - Two props let one component serve every surface: `onEdit` makes the empty state the "No product page — add one" prompt that opens the gear form, and its absence makes that state an em dash (the list has no form to open); `onClick` is where a clickable row passes `stopPropagation`, without which the anchor would open the vendor's page *and* navigate the row behind it.
  - The anchor is gated on `getProductUrlHostname`, not on the string being non-empty — it returns null for anything that is not an http(s) URL, which is what stops a row written before the column was validated from rendering a `javascript:` href.
  - The hostname takes `text-primary-text`, not `text-primary` (3.37:1 on the card, below the AA floor for text).
- `GearDetailView` renders the link under the default-sports line, as the design has it.
- `GearFormDialog` shows the "Product page" field for every kind, placed between Type/Weight and Default sports to match the design, capped at `maxLength={255}` so a long pasted URL is stopped at the input rather than returning the bare `ERROR_422` body. The device's brand-map wording ("Starts as the manufacturer's site…") is kept only for a device; bikes and shoes get "Optional link to the manufacturer's product page."

No migration: the `productUrl` column already exists on `fitness_gears` for every kind, so both schema dumps are unchanged.

**Tests** — `GearProductLink.test.tsx` (hostname rendering, a `javascript:` URL and a bare hostname both falling back rather than becoming an `href`, the em-dash fallback, `onClick` forwarding); the row-click guard in `GearListView.test.tsx`; gear-page link and prompt-opens-the-dialog cases in `GearDetailView.test.tsx`; product-page create *and* edit round-trip cases per kind, plus the per-kind helper copy, in `GearFormDialog.test.tsx`; `CreateGearRequest` cases in `gearRequests.test.ts`; accept/reject cases on both gear routes.

**Docs** — `AGENTS.md` (new Fitness Gear bullet), `CLAUDE.md`, `docs/fitness-file-storage.md` (the `fitness_gears` column notes and both gear endpoint entries, which stated the device-only rule), and `docs/features.md` (which still presented the product page as a device trait).

## Screenshots

Verified in Chromium against a local SQLite dev database with the sanctioned mock user, in light and dark:

- **Gear page (bike)** — the header reads `Moots Vamoots RSL disc` / `Moots Vamoots RSL disc · Road bike · 8.1 kg · added Aug 16, 2026` / `Default for Ride, Gravel ride` / `⧉ moots.com`, the last line an orange external link to `https://moots.com/pages/vamoots-rsl` with `target="_blank"` and `rel="noopener noreferrer"`.
- **Add a bike / Edit bike dialog** — "Product page" sits between Type/Weight and Default sports with the helper line "Optional link to the manufacturer's product page.", round-trips the stored value on reopen, and clearing it saves `null`.
- **Gear page with no product page** — the line becomes "No product page — add one", which opens the edit dialog with the field empty.
- **Device page and gear list** — unchanged in content, both now rendering through the shared component; clicking the list's product link opens the vendor page without navigating the row.

(Images are attached in the originating conversation; this environment cannot upload them to GitHub directly.)

## Review

Three rounds of sub-agent review, 8 findings, all addressed and resolved: the duplicated anchor in the gear list, the missing `aria-label`, the missing `maxLength`, two test-coverage gaps, stale `docs/features.md`, a test that asserted its own handler rather than the production wiring, and an `as never` cast. Round 3 surfaced nothing new. Each new assertion was mutation-tested — reverting the behaviour it guards fails it.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (808 files, 9391 tests)
- [x] Docs updated: grepped `*.md` and `docs/` for `productUrl` and the device-only wording; `AGENTS.md`, `CLAUDE.md`, `docs/fitness-file-storage.md` and `docs/features.md` updated
- [x] If migrations changed: n/a — no migration, the column already exists

https://claude.ai/code/session_01PKoenoaJrH5Caxm9exJgto


app/(timeline)/fitness/gear/[id]/GearProductLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <a
+      href={productUrl}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="inline-flex items-center gap-1 font-medium text-primary-text hover:underline"
+    >
+      <ExternalLink className="size-3 shrink-0" />
+      {hostname}
+    </a>
+  )
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearProductLink.tsx]: **A duplicate of this anchor was left behind in the gear list.** `GearListView.tsx:293-306` still inlines the device table's product link with byte-identical markup — same `inline-flex items-center gap-1 font-medium text-primary-text hover:underline`, same leading `<ExternalLink className="size-3 shrink-0" />`, same `hostname && productUrl` gate. This PR's own new `AGENTS.md` bullet asserts "one component renders it", so the tree now contradicts the rule the PR added.

It cannot be dropped in as-is: the table cell needs `onClick={(event) => event.stopPropagation()}` (the row has its own navigation `onClick`) and an `'—'` empty state rather than the edit prompt. Both are a prop away.

Failure scenario: someone adds `referrerPolicy` or an `aria-label` to `GearProductLink` per the new rule, and the gear list's device rows silently keep the old markup — the two renderings drift with nothing to catch it.

---

app/(timeline)/fitness/gear/[id]/GearProductLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </button>
+    )
+  }
+
+  return (
+    <a
+      href={productUrl}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="inline-flex items-center gap-1 font-medium text-primary-text hover:underline"
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearProductLink.tsx]: **The link's accessible name is the bare hostname**, with nothing tying it to what it is. The empty state one branch up is explicitly labelled "No product page — add one", so the two states are unevenly served: a screen-reader user listing the links on a bike's page hears "moots.com, link" and has no way to tell it apart from any other outbound link on the page.

`aria-label={`Product page: ${hostname}`}` fixes it with no visual change, and keeps the visible text inside the accessible name (WCAG 2.5.3).

---

app/(timeline)/fitness/gear/GearFormDialog.tsx
```diff
@@ -284,6 +266,29 @@ export const GearFormDialog: FC<Props> = ({
...
+            <Label htmlFor="gear-product-url">Product page</Label>
+            <Input
+              id="gear-product-url"
+              type="url"
+              inputMode="url"
+              placeholder="https://"
+              value={productUrl}
+              onChange={(event) => setProductUrl(event.target.value)}
+              disabled={isSaving}
+            />
```
llun, line 290: **No `maxLength` on a field whose column caps at 255**, and the route's over-length rejection is opaque. `optionalProductUrl` applies `.max(VARCHAR_MAX)` (`gearRequests.ts:83`), and a schema failure returns the generic `apiErrorResponse(HTTP_STATUS.UNPROCESSABLE_ENTITY)` — a fixed `ERROR_422` body naming no field and no reason, unlike the kind-mismatch path which returns `getGearKindFieldError`'s message.

Failure scenario: the owner pastes a 280-character product URL (routine for Shopify/Amazon links, in a way brand and model never are) into a bike's Product page and clicks "Save changes" → the dialog shows the literal string "Unprocessable entity". This is newly reachable for bikes and shoes, which is what makes it worth fixing here.

`maxLength={255}` stops it at the input. Native `type="url"` already rejects a bare `moots.com`; a non-http scheme like `ftp://…` still lands on the same opaque 422, which is a smaller and rarer case.

---

app/(timeline)/fitness/gear/GearFormDialog.tsx
```diff
@@ -284,6 +266,29 @@ export const GearFormDialog: FC<Props> = ({
...
+            <p className="text-xs text-muted-foreground">
+              {isDevice
+                ? "Linked from this device's page. Starts as the manufacturer's site; point it at the product if you like."
+                : "Optional link to the manufacturer's product page."}
```
llun, line 294: **This ternary is only half covered.** `GearFormDialog.test.tsx:376-378` asserts the device copy, but nothing asserts that a bike or shoes dialog shows "Optional link to the manufacturer's product page." and *not* the device wording. Inverting the branches fails exactly one test, not two.

---

app/(timeline)/fitness/gear/GearFormDialog.test.tsx
```diff
@@ -412,17 +419,47 @@ describe('GearFormDialog', () => {
...
+    fireEvent.change(screen.getByLabelText('Product page'), {
+      target: { value: '  https://moots.com/pages/vamoots-rsl  ' }
+    })
     fireEvent.click(
       screen.getByRole('button', {
         name: kind === 'bike' ? 'Save bike' : 'Save shoes'
       })
     )
 
     await waitFor(() => expect(mockCreateFitnessGear).toHaveBeenCalled())
```
llun, line 480: **The bike/shoes EDIT round-trip is untested** — the exact path this PR added. Both new cases go through `createFitnessGear`; the update round-trip is covered for `device` only (lines 382, 403). The one bike/shoes edit test, `seeds the fields from the gear being edited and updates it` (line 288), asserts Brand/Model/Nickname/Type/Weight/Notes but never Product page, and the `createGear()` fixture at line 45 has `productUrl: null`.

So nothing asserts that a bike with a **stored** `productUrl` pre-fills the field from `GearFormDialog.tsx:104` and sends it back. Delete that `setProductUrl` seed for non-devices and the whole suite still passes, while every edit of a bike would silently blank a saved link.

---

lib/services/fitness-gears/gearRequests.test.ts
```diff
@@ -360,6 +376,18 @@ describe('getGearKindFieldError', () => {
       description: 'a bike explicitly clearing a product page it never had',
       kind: 'bike' as const,
       fields: { productUrl: null }
```
llun: **This case is now vacuous, and it documents a parameter that no longer exists.** `getGearKindFieldError` stopped reading `productUrl` entirely when the field was dropped from its `fields` type (`gearRequests.ts:203-207`), so "a bike explicitly clearing a product page it never had" asserts the same single fact as the two new cases below it — and it would have passed under the old device-only code too, which is the definition of a test that guards nothing.

The `fields: { productUrl: … }` objects in all three cases also declare a property the function's signature no longer has. It type-checks only because the excess-property check doesn't reach through the `it.each` destructure.

---

llun (commented):
## Sub-agent review — round 1

Three reviewers over the whole diff: server/API + service layer, client components, and tests + docs. Six inline findings above, plus two that can't be anchored inline because their files aren't in the diff:

**`docs/features.md:63-64` — stale by omission.** The "Gear tracking" bullet (bikes and shoes) lists no product page, while the "Recording devices" bullet still advertises "an editable name and product page" as a device trait. Read together, the user-facing feature list still says the product page is a device feature. This is the one doc file that needed a change and didn't get one.

**`GearListView.tsx:233-306` — scope note, no change intended.** Now that every kind carries `productUrl`, the gear list's devices table has a "Product page" column while the bikes and shoes tables don't. That's the last place the old device-only shape survives, but adding a fourth column to those tables is a design-system decision about the list, not this PR's subject (the gear *page*). Flagging it rather than fixing it.

**Clean:** the server-side pass found no defect. Presence-vs-null semantics hold (`.nullish()` stays outermost, so an absent key never clears a column); the http(s) refinement is reachable on both create and update, and was probed against `javascript:`, `data:`, protocol-relative, bare-hostname, mixed-case `JavaScript:` and tab-smuggled `java\tscript:` inputs; the varchar cap still applies; and every device invariant is intact (`POST` still 422s a device, `deviceKey` still create-only, `resolveDeviceGear` still the sole creator, no cached distance). No migration or schema-dump regeneration was needed — `productUrl` has been on `fitness_gears` since `20260813000000_add_fitness_device_gear.js`, in both dumps. The one test asserting the old device-only rule was removed, and `"only valid for a device"` no longer appears anywhere in the repo.

---

app/(timeline)/fitness/gear/[id]/GearProductLink.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearProductLink.tsx]: Fixed in 44ebfa0. `GearProductLink` moved up to `app/(timeline)/fitness/gear/` (it is no longer a detail-page component) and `GearListView`'s device cell renders through it — its inline anchor, the `getProductUrlHostname` call and the now-unused `ExternalLink` import are all gone.

Two props carry the differences you named: `onClick` is where the row passes `stopPropagation`, and `onEdit` is what makes the empty state the prompt — omit it, as the list does, and the empty state is the `'—'`. Both are documented on the component and in the `AGENTS.md` bullet, so the next surface knows which to pass.

The pre-existing `GearListView.test.tsx` cases for the device link and the em dash both still pass unchanged, which is the regression guard for the swap. Verified in the browser too: the list's link is still `https://www.garmin.com` and clicking it leaves the page on `/fitness/gear` rather than navigating the row.

---

app/(timeline)/fitness/gear/[id]/GearProductLink.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearProductLink.tsx]: Fixed in 44ebfa0 — `aria-label={`Product page: ${hostname}`}`, exactly as suggested, with the visible hostname kept inside the accessible name. Because it now renders in the gear list too, the label lands on both surfaces from one place.

The existing queries that matched the link by its old bare-hostname name were updated (`GearListView.test.tsx`, `GearProductLink.test.tsx`); `DeviceDetailView.test.tsx` already matched on a regex and needed no change.

---

app/(timeline)/fitness/gear/GearFormDialog.tsx
llun: Fixed in 44ebfa0 — `maxLength={255}` on the input, with the reasoning in a comment beside it so it isn't mistaken for a stray cap.

Taking the narrow fix only, as you framed it: the length case is the one that gets hit routinely, and stopping it at the input costs nothing. The non-http scheme still lands on the opaque 422 — making that legible means changing what the route returns for a schema failure, which is a shared behaviour across every gear field and doesn't belong in this PR.

---

app/(timeline)/fitness/gear/GearFormDialog.tsx
llun: Fixed in 44ebfa0 — added `describes the product page plainly for $description` (bike and shoes), which asserts the non-device copy is present *and* that the brand-map wording is absent. Inverting the ternary now fails both sides.

---

app/(timeline)/fitness/gear/GearFormDialog.test.tsx
llun: Fixed in 44ebfa0 — added `seeds the stored product page for $description and sends it back` (bike and shoes), which opens the dialog on a gear whose `productUrl` is stored, asserts the field is pre-filled, saves without touching it, and asserts `updateFitnessGear` still carries the URL.

Confirmed it fails for the right reason: deleting the `setProductUrl` seed at `GearFormDialog.tsx:104` now breaks it, where before the whole suite stayed green.

---

lib/services/fitness-gears/gearRequests.test.ts
llun: Fixed in 44ebfa0 — dropped all three, not just the vacuous one. You're right that the function no longer reads `productUrl` at all, so *every* `fields: { productUrl: … }` case there was asserting a fact about a parameter the signature doesn't have. The device case that carried it alongside `defaultSports: []` is now just `a device with an empty sport list`.

The behaviour those cases looked like they covered is tested where it actually lives: the schema (`CreateGearRequest`/`UpdateGearRequest` accept/trim/reject cases) and the routes (`accepts a product page on a bike`, and the create-path 422).

---

app/(timeline)/fitness/gear/GearProductLink.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+          productUrl="https://moots.com"
+          onClick={onClick as never}
+        />
+      </div>
+    )
+
+    fireEvent.click(screen.getByRole('link'))
+    expect(onClick).toHaveBeenCalled()
+    expect(onRowClick).not.toHaveBeenCalled()
+  })
```
llun, line 67: **This test doesn't guard the wiring it was written for.** It supplies its own `stopPropagation` handler, so all it proves is that React's event system works and that `GearProductLink` forwards `onClick` to the anchor. Nothing asserts that `GearListView` actually passes one.

Mutation-verified: deleting `onClick={(event) => event.stopPropagation()}` from `GearListView.tsx:294-297` leaves both `GearListView.test.tsx` and `GearProductLink.test.tsx` fully green (33/33) — while in a browser, clicking a device's product link would open the vendor page *and* push `/fitness/gear/device-1` behind it. That is precisely the bug the prop exists to prevent, with nothing failing.

The missing case belongs in `GearListView.test.tsx`, which already has `mockPush` and the `navigates to the device page when a row is clicked` counterpart at line 380: click the product link, assert `mockPush` was **not** called.

---

app/(timeline)/fitness/gear/GearProductLink.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+  it('stops a click from reaching the row it sits in', () => {
+    // A gear list row navigates on click; without this the anchor would open
+    // the vendor's page and push the gear route behind it.
+    const onRowClick = vi.fn()
+    const onClick = vi.fn((event: MouseEvent) => event.stopPropagation())
+    render(
+      <div onClick={onRowClick}>
+        <GearProductLink
+          productUrl="https://moots.com"
+          onClick={onClick as never}
```
llun: **`onClick as never` disables type checking at the one place the prop's contract is exercised.** It isn't hiding a production problem — `Props.onClick` is correctly typed `MouseEventHandler<HTMLAnchorElement>`, and the real call site in `GearListView.tsx:296` type-checks with an inferred parameter. The cast exists only because the handler here was annotated with the DOM `MouseEvent` instead of React's synthetic one.

`never` is assignable to everything, so it also silences any future signature change: narrow `Props.onClick` later and this test keeps compiling on a contract it no longer satisfies. Annotate the parameter `React.MouseEvent<HTMLAnchorElement>` and drop the cast.

---

llun (commented):
## Sub-agent review — round 2

Re-reviewed round 1's fixes (`44ebfa0`) for regressions. Two findings, both in the test I added for the row-click guard, one of them mutation-verified. Everything else about the fixes checks out.

**Clean, each verified rather than assumed:** no stale import of the old `[id]/GearProductLink` path anywhere, and the move turns nothing into a route (only `page`/`layout`/`route` files do that). The device cell is byte-identical to the anchor it replaced — `href`, `target`, `rel`, the full `className`, the icon — with only the `aria-label` added, and the old `hostname && gear.productUrl` guard is the same predicate as the component's. The `—` fragment renders the same text node the string literal did and still inherits the cell's styling. Both surfaces that should pass `onEdit` do, and each has a test covering the prompt branch, so dropping it fails a test rather than silently degrading to an em dash. The `aria-label` broke only the two accessible-name queries that were already updated. `maxLength={255}` cannot truncate a stored value on edit — `maxlength` constrains only dirty user input, so a controlled value seeded from a row is rendered and resubmitted in full — and the server cap still rejects over-length input independently, so it is defence in depth rather than the only guard.

---

app/(timeline)/fitness/gear/GearProductLink.test.tsx
llun: Fixed in f7dd940 — you were right that the test was checking its own handler rather than the wiring.

The guard moved to `GearListView.test.tsx`, beside the `navigates to the device page when a row is clicked` case it's the counterpart to: click the `Product page: garmin.com` link, assert `mockPush` was not called. Mutation-verified in both directions — with `onClick` removed from the cell it fails (1 failed | 26 passed), with it restored the file is green.

The component's own test keeps the narrower claim it can honestly make and is renamed to match it: `forwards onClick to the anchor`, with a comment saying why the row behaviour is asserted in the list's file instead.

---

app/(timeline)/fitness/gear/GearProductLink.test.tsx
llun: Fixed in f7dd940 — cast dropped, handler now takes React's synthetic `MouseEvent<HTMLAnchorElement>` (imported as a type). Agreed on the reasoning: it was silencing nothing real today, but `never` would have kept it compiling against a narrowed signature tomorrow.

---

llun (commented):
## Sub-agent review — round 3 (clean, loop closed)

No new findings. Round 2's test-only diff was re-reviewed by mutation rather than by reading, and every claim held:

- **The new guard is genuine.** Removing `onClick={(event) => event.stopPropagation()}` from `GearListView.tsx:294-297` fails exactly one test — the new `does not navigate the row when the product link is clicked` — with `mockPush` called once with `/fitness/gear/device-1`. The other 26 in the file still pass, so nothing else was silently covering it.
- **`mockPush` isolation is real, not assumed.** `vi.clearAllMocks()` in the outer `beforeEach` reaches the nested describe; the mutated run reported 1 call, not 2, even though the immediately preceding test pushes once. The assertion is order-independent.
- **The propagation mechanism is sound.** The same mutation showed the click genuinely reaching the ancestor `<tr onClick>`, so the passing state is caused by `stopPropagation` and not by a click that never propagated.
- **The narrowed component test still earns its place** — deleting `onClick={onClick}` from the anchor fails it.
- `import type { MouseEvent } from 'react'` has no runtime binding and shadows nothing; `tsc --noEmit` is clean across the gear files, so the handler is really checked against `MouseEventHandler<HTMLAnchorElement>`.

**Loop summary:** 3 rounds, 8 findings, all addressed and resolved — 6 in round 1 (the duplicated anchor, the missing `aria-label`, the missing `maxLength`, two test-coverage gaps, and stale `docs/features.md`), 2 in round 2 (a test that asserted its own handler instead of the production wiring, and an `as never` cast). Round 3 surfaced nothing, so the loop stops here well inside the 20-round cap.

---

PR #1452, opened 2026-08-17
minor: add route heatmap tile pyramid storage and database accessors

llun:
## Summary

**Stacked PR 1 of a series.** This is the data layer only — three tables and their database accessors. It is inert: nothing reads or writes them yet, and `fitness_route_heatmaps` keeps serving every existing heatmap exactly as before.

### The problem this series fixes

The fitness route heatmap stores its whole geometry as **one JSON blob fit to a single global vertex budget** (80k points, reached by doubling the Douglas–Peucker tolerance from 1 m up to 256 m). Precision therefore falls off as coverage grows: a Singapore-sized heatmap fits the budget at the 1 m floor and looks sharp, while a world-spanning one lands at 128–256 m and cuts across roads. Zooming in never reveals more, because the map is simplified once and reused at every zoom, and the client re-flattens it again to a 60k render budget with no zoom awareness.

Generation is monolithic too: every run re-downloads and re-parses every FIT/GPX/TCX from object storage (`fitness_files` holds no geometry), and every 20 s checkpoint rewrites the entire accumulated blob.

The replacement is a **per-actor zoom-tiled pyramid** whose precision is fixed per screen pixel at every zoom — Strava's model — built incrementally route-by-route with bounded memory. This PR lays its storage.

### What is here

**`fitness_file_routes`** — caches each activity's simplified polyline so a rebuild stops re-downloading and re-parsing every source file. An activity with no GPS gets a row with no points and null bounds: that row is the cached answer "this has no route", and it is what stops every rebuild from re-reading treadmill files to rediscover the same thing.

**`fitness_route_heatmap_pyramids`** — one build-state row per actor: a monotonic `version`, a `(createdAt, id)` keyset resume cursor, and progress counters.

**`fitness_route_heatmap_tiles`** — the geometry, one row per `(actorId, z, x, y)`, so a viewport reads only the tiles it draws and a build flushes bounded batches instead of rewriting a growing blob.

### Two decisions worth reviewing

**Build ownership is a compare-and-swap on `version`.** The row is read, then updated only if its version has not moved. That one mechanism covers every case, and each outcome is reported distinctly rather than as a bare boolean: a completed pyramid newer than the request is *already fresh* (so N queued region generations cost one build plus N cheap completions); a recently-heartbeating build is left alone; an abandoned build is *resumed on its own version* when it got far enough to checkpoint — so the tiles it wrote are added to rather than replaced — and restarted on a fresh version when it did not; a caller that loses the swap is told, instead of writing tiles the winner would have to reconcile. Every later write is guarded on the same version, so a pass presumed dead that wakes up late writes nothing. Checkpointing doubles as the heartbeat, so there is no separate keepalive to forget.

**The keyset cursor is not a style preference.** A whole-history scan cannot page by `offset`: delete a row behind the cursor and everything after it shifts, so the next page skips an activity — and a skipped or double-counted activity in a *heat value* is permanently wrong rather than merely missing. The test pins that difference directly, asserting the offset read loses the row the cursor read keeps. The comparison is a nested OR rather than a row-value tuple, which the three backends do not implement alike.

Also included: `applyCountableActivityFilter` moves out of `fitnessGear.ts` into a shared util (pure move, no behaviour change). Its whole purpose is that every fitness rollup agrees on which activities count, so the new route aggregates reuse it rather than growing a second copy that could drift from the gear rollups and the fitness overview.

### Commits

Seven, each independently reviewable: the shared-filter extraction, the migration + both schema dumps, each mixin with its tests, the additions to the two existing mixins, a summary-column fix, and the docs.

### Still to come in later PRs

The tiler core (zoom ladder, 1-pixel simplification tolerance, quantized edge-count merge), the generation job rework, the viewport tile endpoints, the client's zoom-aware fetching with the Strava-style visit-count heat ramp, and the static image path.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn build`, and `yarn test` all pass locally, run in that order (full suite: 811 files, 9469 tests, 0 failures)
- [x] Docs updated: `docs/architecture.md` and `docs/fitness-file-storage.md` both enumerate the fitness tables, so both describe the three new ones
- [x] If migrations changed: BOTH `migrations/schema.sql` and `migrations/schema.sqlite.sql` are regenerated in this PR (PostgreSQL 17 and SQLite, via the documented pipelines; verified by first reproducing both committed files byte-for-byte from the pre-change migration set)


lib/database/sql/fitnessRouteHeatmapTile.ts
llun, line 1: **[round 1 · high · concurrency] The compare-and-swap was vacuous on the resume path.** — *fixed in 8a1e9a4*

`version` was doing two jobs at once, and they conflict.

It stamps the tile generation, so a **resumed** build must keep it — otherwise `deleteStaleFitnessRouteHeatmapTiles(beforeVersion)` at completion would sweep the very tiles the interrupted pass had already written. But it was also the column the claim's CAS guarded on, and the resume branch wrote only `status`/`error`/`updatedAt`. Since nothing the guard reads ever moved, `.where('version', pyramid.version)` still matched afterwards.

Two failures, both reproduced on the SQLite harness before the fix:

1. **Two owners.** Workers A and B both read an abandoned, cursor-bearing row, both compute `resumed=true`, both run the guarded UPDATE, both affect one row, and both receive `{claimed: true, resumed: true}`. They then scan from the same cursor and both flush tiles at the same version — and `upsertFitnessRouteHeatmapTiles` merges by overwriting, so whichever flush lands second silently discards the other's work.
2. **Zombie write.** A stalls past the stale window, B reclaims, then A wakes and calls `updateFitnessRouteHeatmapPyramid({version, status: 'completed', completedAt, cursor: <older>})` — and it returns `true`. The row is marked completed with the cursor rewound, and every later claim short-circuits on `already-fresh`, serving a half-built pyramid as finished.

Both are precisely what the docblocks here (and the PR description) claimed could not happen.

**Fix:** split the two responsibilities. `version` keeps its tile-generation meaning; a new `claimSeq` column is the ownership token, bumped by *every* successful claim including a resume, and both the claim CAS and `updateFitnessRouteHeatmapPyramid` guard on it. Simply bumping `version` on resume would not have worked — it would reintroduce the stale-sweep problem.

Two regression tests added, both mutation-checked (they fail against the old behaviour and pass against the fix): `fences a presumed-dead pass whose build was RESUMED by another worker`, and a table-driven `lets exactly one of two concurrent claimers win over $description` covering the fresh and resumable rows.

---

lib/database/sql/fitnessFile.test.ts
llun, line 1: **[round 1 · high · tautological test] The createdAt-tiebreak test could never run its assertion.** — *fixed in 8a1e9a4*

`it('separates rows sharing a createdAt by id')` wrapped its only tiebreak assertion in `if (sameTime.length > 1)`. The fixture creates its files with four sequential awaited `createFitnessFile` calls, each a transaction — they land milliseconds apart, so `sameTime` was always length 1 and the block never executed. The reviewer confirmed it by replacing the cursor predicate in the implementation and watching the test still pass.

Rewritten to force the collision instead of hoping for one: `vi.useFakeTimers()` freezes the clock (the pattern this file already uses at `uses a deterministic id tiebreaker when files share createdAt`), three files are created at an identical `createdAt`, the collision itself is asserted, and the test then pages **one row at a time** so every page boundary falls inside the group — returning each row exactly once and in id order.

---

lib/database/sql/fitnessFileRoute.test.ts
llun, line 1: **[round 1 · medium · tautological test] The chunking tests passed with chunking removed entirely.** — *fixed in 8a1e9a4*

Both "reads more ids/keys than SQLite allows bound variables in one statement" tests (here and in `fitnessRouteHeatmapTile.test.ts`) sent ~1200 bindings on the assumption that 999 is a hard ceiling. It is not: `better-sqlite3` ships a `SQLITE_MAX_VARIABLE_NUMBER` of 32766, so the unchunked query succeeds and the tests prove nothing. The reviewer verified this by replacing `chunkArray(...)` with `[uniqueIds]` and watching every test still pass — and I reproduced the same result before rewriting them.

999 is the project's own conservative `SQLITE_MAX_BINDINGS` floor, which matters for stricter SQLite builds and for PostgreSQL's 65535-parameter cap — not for this build. So a result-shape assertion structurally cannot detect the loss.

Rewritten to assert what actually distinguishes the two implementations: the **number of statements issued**. Each now builds its own database via `getTestSQLDatabaseWithInstance()`, attaches a knex `query` listener, requests 2500 ids, and asserts exactly 3 statements. Both fail when the chunking is removed. The original result-shape checks are kept alongside as companions (real rows placed in the first, second and last chunk) with corrected comments.

---

lib/database/sql/fitnessRouteHeatmap.ts
llun, line 1: **[round 1 · medium · test coverage] The subscriber filter's soft-delete clause had no coverage.** — *fixed in 8a1e9a4*

`applyPyramidSubscriberFilter` promises in its docblock that "terminal and deleted rows are left alone". The terminal half was well covered (removing the `whereIn('status', …)` fails five tests), but removing `.whereNull('deletedAt')` left all 36 tests in the file passing — so half a documented contract was unverified.

Added `excludes a soft-deleted run and never mirrors progress onto it`. Because `getFitnessRouteHeatmap` itself filters deleted rows, the deleted row cannot be read back to inspect directly; the test instead asserts it is absent from `getFitnessRouteHeatmapSubscribers` and that `mirrorFitnessRouteHeatmapGenerationProgress` returns exactly the number of live subscribers — so a mirror that touched the deleted row would report one too many. Mutation-checked: it fails with the clause removed.

Also in this file, from my own pre-push pass: `getFitnessRouteHeatmapSubscribers` was selecting `*`, which drags the whole `segments` payload off disk on every checkpoint just to count who is waiting. Both summary reads now share one explicit `SUMMARY_COLUMNS` list (f434246).

---

llun (commented):

Four reviewers fanned out across concurrency correctness, SQL/dialect portability, schema-dump and project invariants, and test quality. Each finding was then handed to an independent agent instructed to **refute** it: **21 raw findings, 6 survived**, and the 6 collapse to 4 distinct issues (the concurrency bug was found independently by two dimensions).

All four are fixed in `8a1e9a4`, each with a regression test that was mutation-checked — reverted the fix, confirmed the new test fails, restored it, confirmed it passes.

| # | Severity | Issue | Status |
|---|---|---|---|
| 1 | high | Claim compare-and-swap vacuous on the resume path | fixed |
| 2 | high | createdAt-tiebreak test's assertion was unreachable | fixed |
| 3 | medium | Chunking tests passed with chunking removed | fixed |
| 4 | medium | Subscriber filter's soft-delete clause untested | fixed |

The headline one is a real concurrency defect I introduced: `version` was serving as both the tile-generation stamp and the ownership fence, and those two requirements conflict precisely on the resume path. It is now split, with `claimSeq` as a dedicated ownership token.

The two test findings are the more instructive ones — both tests were green and both were worthless. One guarded its only assertion behind a condition that never held; the other assumed a 999-binding SQLite ceiling that this build does not have (`better-sqlite3` allows 32766), so it passed with the chunking deleted outright. That one is now asserted on the number of SQL statements issued, which is the only thing that can actually distinguish the two implementations.

15 findings were refuted, among them: that `getFitnessFileRouteBounds` needs a `GROUP BY` (a bare aggregate over a join is valid on all three backends and returns one all-NULL row, which the code handles); that the `.onConflict(...).merge([...])` array form is unsupported (it is the documented column-list form); and several notes about the tables having no readers, which is the intended shape of this PR.

Round 2 follows against the updated branch.

---

llun:
## CI shard 2 failure on `8a1e9a4` — fixed in `3b9e124`

CI was right and the **tests** were wrong, not the implementation.

The sort is `(createdAt, id)`. Two fitness files written in the same millisecond — ordinary on a fast machine — tie-break on their random uuid, so the sorted order is **not** the order the fixtures were created in. Three tests in `getFitnessFilesByActor keyset pagination` hardcoded `keyset-1.fit` … `keyset-4.fit` as the expected sequence, which made them latent flakes from the moment I wrote them. CI got `[1, 3, 2, 4]`.

This is a flake I introduced, not pre-existing: the full local suite reproduces exactly the same two failures, so it was luck that the first run happened to be green.

Each test now reads the canonical ascending order first and asserts the paged walk reproduces **that**, since what is under test is that cursor paging preserves the order — never what the order happens to be. The newest-first check became "the default read is the ascending read reversed" rather than naming a file.

Verified against the worst case the fixtures can produce: with the clock frozen so all four rows share one `createdAt` and ordering falls entirely to random uuids, the file passes repeatedly. `lib/database/sql/` is green (71 files, 1299 tests).

Round 2 of the sub-agent review is still running against the branch; I'll fold its findings in when it reports.

---

PR #1454, opened 2026-08-18
minor: cache each activity's route so heatmap generation stops re-reading object storage

llun:
## Summary

Phase 2 of the fitness route heatmap work, stacked on #1452 (merged). That PR added `fitness_file_routes` and its accessors but left them inert; this one gives the table a writer and a reader.

Today `fitness_files` stores no geometry, so **every** heatmap generation re-downloads and re-parses every FIT/GPX/TCX from object storage — the same bytes, the same parse, every run. This caches each activity's simplified polyline as a side effect of the parse that already happens during processing, and teaches the generator to read it.

**The heatmap itself does not change.** All 22 existing generation tests, which pin the produced segments, bounds and counts, pass unchanged. This is purely a speed and load change:

- A history uploaded since this lands regenerates **without touching object storage at all**.
- An older history pays the downloads once, writing through, and is free from the second run on.
- A cached empty row counts as a hit, so a treadmill-heavy history stops paying a download per run to rediscover an activity has no GPS.

It also fills the table Phase 4's tiled build reads, so that phase is not blocked on a backfill.

### How it works

`lib/services/fitness-files/fileRouteCache.ts` owns the one pipeline that turns parsed coordinates into cached points — cap to `filePointLimit`, simplify at the configured floor tolerance, round to six decimals. It is lifted verbatim from what the generator already did per file, and `downsampleRoutePoints` moves here out of the job so there is one copy rather than two that could drift. Both the cache-hit and cache-miss paths end in it, which is the point: a heatmap must not come out differently depending on whether a row happened to exist.

Points are cached **pre-privacy**. Privacy zones are a live setting, and a cache that had already trimmed the route could never grow those ends back without re-downloading the file, so trimming stays at build time from the settings in force then. These are the same coordinates object storage already holds, and the rows are never serialized to a client.

A `sourceVersion` mismatch is a miss, not a stale hit, so changing how routes are built re-derives them rather than serving geometry the old rules produced.

### Best-effort, deliberately without a persisted signal

Both writers swallow failures and log. Unlike `mapError`, a missing route row is invisible to the owner — the generator just parses the file as it does today — so there is nothing to act on and no repair key worth persisting. This is a cache of a cache.

In the generator specifically, the write-through failure must not escape: the route is already in hand, so a failed write costs this run nothing, while letting it throw would drop a perfectly good activity out of the heatmap (the caller reads a throw there as an unreadable file). Pinned by a test.

### Deletion

- Deleting an activity drops its cached route in the same transaction. Every aggregate over the table already joins through the soft delete so a leftover row was never *counted* — but it held the whole polyline forever, and that blob is most of what the cache costs.
- Clearing the route heatmaps now clears the tile pyramid too. It is the same geometry in a different shape, so taking only the region rows would leave an owner's routes behind for a later build to serve. Its row counts stay out of the reported `deleted`, which answers for the region rows the caller asked about.
- That endpoint deliberately leaves `fitness_file_routes` alone: it caches parses of files the owner still has, not heatmap output, and dropping it would only buy back the slow path on the next generate.

### The one behavioural difference

Six-decimal rounding now happens before privacy trimming instead of after it, because that is what makes hit and miss identical. That shifts a coordinate by at most ~0.11 m, which can in principle flip a point sitting exactly on a privacy-zone or region boundary. The existing generation tests pin the resulting geometry and are unchanged.

### Also in here

`docs/fitness-file-storage.md` said build ownership is a compare-and-swap on `version`. It is `claimSeq` — a resume keeps its version by definition, so guarding on the version would leave a resumed build with no fence at all, which is exactly why the two were split during #1452's review. The doc missed the correction.

### Commits

| | |
|---|---|
| `b9dfb42` | the shared route-building pipeline |
| `4c9e930` | write the cache from `processFitnessFileJob` |
| `035cd45` | deletion paths |
| `197ffda` | the generator reads the cache |
| `561a776` | docs |

### Tests

New coverage for: the route row written after a parse; a write failure not failing processing; the zero-GPS negative cache; pre-privacy caching; delete removing the row; write-through on a first generation run; **`getFitnessFile` never called on a second run**, producing byte-identical segments/bounds/counts; no download for a cached GPS-less activity; re-derivation at an older `sourceVersion`; and the activity still counted when its route cannot be cached.

Each of those was mutation-tested — reverting the behaviour it describes fails exactly that test and no others.

Full suite: 812 files / 9495 tests.

### Next

Phase 3 (the pure tiler) and Phase 4 (the tiled generation job) follow as further stacked PRs; those are where the precision actually changes.

## Checklist

- [x] If migrations changed: no migrations in this PR, so both schema dumps are unchanged

https://claude.ai/code/session_01ELw2T3kDEwCRZjdoveDefw

---
_Generated by [Claude Code](https://claude.ai/code/session_01ELw2T3kDEwCRZjdoveDefw)_


llun:
## Review round 1 — 10 findings, all fixed in `4462e8c`, `b5e8cd5`, `bf1b8d0`

Three sub-agent reviews: Phase 2 correctness, test power, and the **carried-over Phase 1 modules** (#1452 merged with only one review round completed, so the `claimSeq` fix that replaced its high-severity bug had had exactly one pass). Everything below was verified by reading the code, and every fix was mutation-tested — reverting it fails exactly the intended test and no others.

### The build-claim protocol (merged in #1452, latent until the tiler lands) — `4462e8c`

**1 · HIGH — the compare-and-swap guarded the token, but not the state the decision was made on.** `claimFitnessRouteHeatmapPyramidBuild` decided on `status`, `completedAt` and `updatedAt`, then CAS'd on `claimSeq` alone. `updateFitnessRouteHeatmapPyramid` writes all three of those and never touches the token, so an incumbent could invalidate every input to the decision while leaving the guard exactly where the CAS expected it:

- *a live build stolen* — claimer reads an expired heartbeat; owner checkpoints (refreshing `updatedAt`, not `claimSeq`); claimer's CAS still matches and it is told it owns a demonstrably running build.
- *a finished build discarded* — same read, but the owner completes in the window. The CAS still matches, resets the row to `generating`, bumps the version out from under the tiles on disk and rebuilds from scratch — exactly what the `already-fresh` branch exists to prevent.

Fixed by repeating the decision's predicate as a WHERE clause on the CAS. Written as a De Morgan expansion rather than two `whereNot` groups, because `completedAt` is nullable and `NOT (status = 'completed' AND completedAt >= ?)` is NULL for a completed build that never stamped a time — SQL treats that as not-matching, which would wedge the actor's pyramid permanently. A failed CAS now re-reads and re-classifies, so losing to a build that *completed* reports `already-fresh` and the caller serves from the finished pyramid instead of giving up.

**2 · HIGH — the tile flush had no fence at all.** `upsertFitnessRouteHeatmapTiles` took `version`, not `claimSeq`. A pass presumed dead and superseded kept writing, and its `.merge()` **replaces** `segments`/`pointCount` — so it overwrote tiles the successor had already merged. Worse on a resumed build, where the successor keeps the same version: the zombie's rows carry that same version, so `deleteStaleFitnessRouteHeatmapTiles` (`version < beforeVersion`) can never remove them. Permanently wrong heat, and nothing reports it because the counters come from the successor.

Now fenced on the claim token, checked with a guarded UPDATE inside the same transaction — an UPDATE rather than a SELECT because it takes the pyramid row's write lock, so a concurrent claim blocks rather than landing between the check and the writes. That check doubles as a heartbeat, so flushing tiles is itself what keeps a working pass from being reclaimed.

**3 · MEDIUM — the claim adopted whatever token its confirming re-read found**, not the one it wrote. A claim landing in that gap handed this pass the *thief's* token, and both would then write with a `claimSeq` the fence accepts — the fence failing open, which is worse than not having one. It now reports its own token, so a superseded pass is rejected by its first write. I removed the early lost-race branch I first wrote for this rather than ship a defensive branch whose interleaving the suite cannot stage deterministically.

Also: clearing an actor's heatmaps now drops the pyramid **before** its tiles, for the same reason — removing the row removes the token, so a running build is rejected from that moment instead of writing into the gap between the two statements.

### The cache read this branch added — `b5e8cd5`

**4 · HIGH — the prefetch was unbounded, and the failure would not have healed.** It read a whole page of cached routes (100 activities) before processing any, and held all of it. `filePointLimit` caps one row at 80,000 points, so a history of dense recordings puts on the order of 400 MB live — past the job's own 512 MB reduce threshold, on activities that completed fine when the job held a single parse at a time. `shouldReduceAccumulation` can't rescue it: it downsamples the accumulated segments, a *different* object, so it fires on every remaining file and frees nothing. And the checkpoint cursor points back at the same page, so every retry rebuilds the same batch and dies again.

Now read ten at a time, dropping each row once folded in. That keeps the worst case at the order of the single parse it replaced, while still collapsing a page's storage reads into ten indexed lookups.

**5 · HIGH — an unreadable cache row was served as a permanent hit.** `parseRoutePoints` answers `[]` for a payload it can't decode — documented as "a corrupt payload is a cache miss" — but by the time it reached the generator that was byte-identical to the negative-cache row a GPS-less activity legitimately writes. The activity silently vanished from every heatmap until its file was reprocessed. `pointCount` is stored beside the payload and is exactly the disambiguator; a row whose points don't match its own count is now re-derived and repaired on the way through.

### Tests that passed against mutants of what they claim — `b5e8cd5`, `bf1b8d0`

**6** — the version test seeded `SOURCE_VERSION - 1`, which is **0 and falsy**, so `if (cachedRoute?.sourceVersion)` passed it. It pinned "a version-0 row is re-derived", not the comparison — meaning the day the constant is bumped, every row written under the old rules would be served as a hit. Now seeds a version *ahead* of the current one, which also covers the rolling-deploy case.

**7** — the miss path could have skipped the entire build pipeline undetected (no cap, no simplification, no rounding). Every route test mocked a **two-point** route, where all three stages are the identity. Now covered by a 900-point route asserted against `buildFitnessFileRoutePoints` of the same input.

**8** — cached rows could have been matched to activities **positionally**. Every test had exactly one file, while the batch read has no `ORDER BY` — same class as the `(createdAt, id)`-ties-on-a-random-uuid failure this work hit in CI earlier. Now covered by a mixed page where only one activity has a cached row.

**9** — the delete test couldn't tell a scoped delete from a table-wide one: dropping the `where` clause, so one user's delete wipes the entire instance's cache, passed all 212 related tests. Now keeps a second activity and asserts it survives.

**10** — the cap-before-simplify test asserted only output length, and by its own premise a staircase yields exactly the cap in **either** order — so the ordering it exists to pin was satisfied by the fixture, not the code. It now observes what Douglas–Peucker is handed. (Plus the pre-privacy test, which asserted the same literal as the zone-less test above it and never verified the zone was in force.)

### Explicitly checked and refuted

Cache-hit vs cache-miss geometry (identical — both end in the same pipeline, and `FitnessCoordinate` is only `{lat, lng}`); checkpoint/resume interaction; `actorId` provenance on the write (the job throws on a mismatch before reaching it, and imports route through the same handler); lock ordering on the delete transaction; the create-then-claim race (a real `UNIQUE(actorId)` makes the losing insert a no-op); MySQL `affectedRows` semantics; SQLite bind limits; per-backend date/numeric typing on the new keyset; `applyCountableActivityFilter`'s extraction being byte-identical; index coverage; and that the `requestedAt: Date.now()` the new job tests pass is load-bearing rather than papering over a behaviour, with no same-millisecond tie hazard.

Full suite green: 812 files / 9504 tests.

Round 2 starting now over the same diff plus these fixes.

---

llun:
## Review round 2 — 6 findings, all fixed in `d9614df`, `206b879`

Two sub-agents: one attacking round 1's fixes, one sweeping the full diff for what a first round would miss. Three of the six are defects **in round 1's own fixes**. All mutation-verified.

### Round 1's claim fix was incomplete — `d9614df`

**1 · HIGH — the CAS asserts what a state is NOT, but `resumed`/`version` rests on what it IS.** Round 1 folded the freshness and liveness verdict into the compare-and-swap. But `resumed` and `nextVersion` are decided by `status === 'generating' && cursor !== undefined`, and "claimable" is satisfied by *every* terminal status equally — so that half of the decision was still an unguarded snapshot.

Reproduced by the reviewer against a real database: a stalled owner wakes inside the claim's window and writes its terminal state (`status: 'failed'`, `cursor: null`) — guarded only on a token it still holds, so **accepted** — and the CAS still matches, because `'failed'` satisfies "not generating". The winner is told `resumed: true`, keeps the previous build's version, and finds no cursor. It rescans from the beginning **into that build's own tiles**, doubling every count it revisits — the failure the migration's own cursor comment calls out as the worse one ("a double-counted activity is a permanently wrong heat value rather than a missing row") — and completion can't sweep the leftovers, because the version never moved.

Fixed with `applyResumePremiseFilter`: a resume asserts the row is still `generating` with a cursor, a fresh claim asserts it is not, so a claim whose ground shifted falls into the existing re-read path instead of succeeding on a premise that no longer exists.

**2 · MEDIUM — reordering the delete closed one window and opened its mirror.** Round 1 made the DELETE handler drop the pyramid before its tiles, so a running build is fenced off rather than writing orphan tiles. But between the two statements a build can now claim the freshly-absent row, flush legitimately, and have the second statement delete exactly what it just wrote — leaving a pyramid that reports itself `completed` with a `tileCount` over tiles that are gone, and answering `already-fresh` to every claim until a later `requestedAt`. Now one transactional call, `deleteFitnessRouteHeatmapPyramidAndTilesForActor`, which keeps what the reorder bought and closes the other side.

**3 · MEDIUM — the heartbeat test I added in round 1 could not fail.** Both assertions were tautologies: `after.updatedAt >= before.updatedAt` holds because `updatedAt` only moves forward, and the thief's staleness cutoff was read back *off the row it was testing*, making it `x >= x`. Verified: replacing the flush's heartbeat payload with a no-op field kept 33/33 green. Now waits for real elapsed time and gives the thief a cutoff the row clears only because the flush wrote one.

### Round 1's memory fix reverted silently — `206b879`

**4 · MEDIUM — nothing pinned the prefetch bound.** Setting `ROUTE_CACHE_PREFETCH_SIZE = ROUTE_HEATMAP_PAGE_SIZE` — byte-for-byte the ~400 MB regression round 1 fixed — passed all 30 tests, since the route tests use one to three activities and the outcome is identical either way. My first attempt at a test for this *also* failed to catch it: asserting each batch is `<= ROUTE_CACHE_PREFETCH_SIZE` moves with the constant under mutation. It is now pinned as a value.

### Docs described what had just been replaced — `206b879`

**5 · The `docs/` update in `561a776` predates both fixes**, so the one durable reference for this subsystem prescribed exactly the two shapes that turned out to be wrong: a page-at-a-time route read, and a build claim fenced on the ownership token alone. Someone implementing the tiler from it reproduces the race. Corrected, along with the stale "nothing writes to or reads from them yet" (this branch's delete handler does), and extended with the batch arithmetic and the `pointCount` cross-check.

**6 · Minor** — the per-file parse-failure log in the restructured path still used `error: nodeError.message`, so Error Reporting got no `stack_trace`. Pre-existing, but AGENTS.md § Logging is explicit and the file was open.

### Explicitly checked and refuted

The reviewers executed rather than reasoned where it mattered: **`applyClaimableFilter` vs `classifyPyramidForClaim` across all 60 states** (`status` × `completedAt` {null, <, ==, >} × `updatedAt` {<, ==, >}) driven through real query execution on both SQLite **and** a PostgreSQL 16 cluster — 0 mismatches on each, `>=`/`<` a correct complement, the nullable-`completedAt` disjunct behaving as documented; knex's nesting printed via `toSQL()` and parenthesised as intended on both dialects; cross-backend date binding (better-sqlite3 binds `Date` as epoch-ms, PG keeps full precision, and comparing a timestamp column against a `Date` is already the established pattern here); the flush transaction's lock verified against real PostgreSQL in **both** interleavings (flush-first → the claim blocked 559 ms then matched 0 rows; claim-first → the zombie wrote nothing), with no lock-order inversion anywhere in the repo; `getInsertBatchSize` through a transaction; the prefetch batch boundaries for a short page, an exact multiple and a partial tail; route rows outliving their activity (every deletion path goes through `deleteFitnessFile`, and there is no hard delete or un-delete); survival of actor deletion (built a SQLite DB from the committed dump and confirmed better-sqlite3 enables `foreign_keys` itself, so the CASCADE fires); and every reprocess path that might need to invalidate a row.

One note carried forward, not a defect: re-classification makes `lost-race` nearly unreachable, since losing to another claimer now reports `build-in-progress`.

Full suite green: 812 files / 9507 tests. Round 3 running.

---

llun:
## Review round 3 — 5 findings, all fixed in `ee8dfa3`

Attacking round 2's fixes. Three were defects **in those fixes**; the reviewer verified everything by execution, against in-memory SQLite and a real PostgreSQL 17 cluster.

**1 · HIGH — the resume premise wedges an actor permanently where JS and SQL disagree.** Round 2's `applyResumePremiseFilter` tests `cursorId` alone; the JS `resumed` decision it mirrors requires **both** cursor columns to be truthy. On a row where those disagree, *neither* branch of the filter can match — the CAS finds zero rows and the claim answers `lost-race` on every attempt, forever, with no escape because clearing the cursor needs a token nobody holds. Executed state matrix:

| row state | sqlite | pg |
|---|---|---|
| every state reachable through the mixin today | claimed ✓ | claimed ✓ |
| generating, `cursorId` set, `cursorCreatedAt` NULL | `lost-race` ✗ | `lost-race` ✗ |
| generating, `cursorId` = `''` | `lost-race` ✗ | `lost-race` ✗ |
| generating, `cursorCreatedAt` = epoch 0 | `lost-race` ✗ | claimed ✓ |

That last row is a second bug behind the first: knex/better-sqlite3 stores a `Date` as an epoch-ms **integer**, so an epoch-0 cursor reads back falsy on SQLite and as a truthy `Date` on PostgreSQL — the same row, classified differently per backend. Both sides now test both columns for **null** rather than truthiness.

Fair to note what the reviewer could not find: a reachable trigger in shipped code. The only writer sets both columns in one statement. But the claim has no production caller yet — the whole protocol exists for a tiler that lands later — the schema has no CHECK tying the columns, and the parser's own "both halves or neither" comment shows a half-cursor is a state the design already contemplates. A permanent wedge is not an acceptable response to one.

**2 · MEDIUM — the transactional delete's guarantee was false for the case it was added to cover.** The comment claimed a build "cannot start until this commits", but `DELETE … WHERE actorId = ?` matching **no** row takes no lock, so for an actor who has never built there is nothing to serialise against. Reproduced on PostgreSQL 17:

```
T1 deleted 0 pyramid row(s); holding tx open
T2 claim -> claimed=true claimSeq=1
T2 flushed tiles -> true
T1 deleted 2 tile row(s), committing
RESULT pyramid=generating claimSeq=1  tilesRemaining=0
```

The build holds a valid token, believes its flush succeeded, and its tiles are gone — it completes reporting a `tileCount` over rows that do not exist. The transaction now creates the row before deleting it, so there is always something to lock; a concurrent claim's own insert then waits for the commit and starts cleanly. (When the row *was* present the fix already worked — the reviewer measured the concurrent flush blocking 305 ms and then correctly returning `false`.)

**3 · The behavioural resume test I added in round 2 was a false witness.** Patching `applyResumePremiseFilter` out of the CAS failed exactly one test — the structural one — while the behavioural test stayed green. It writes the terminal state *before* calling the claim, and the claim does its own read, so `resumed` comes out false with or without the fix; its comment describes an interleaving it does not stage. Renamed to what it genuinely pins, with a new test that walks the half-cursor rows directly.

**4 · Dead code.** `deleteFitnessRouteHeatmapTilesForActor` and `deleteFitnessRouteHeatmapPyramidForActor` lost their only caller to the transactional pair — two public `Database` methods whose sole remaining consumer was their own test. Removed.

**5 · Docs corrected to their own parent, not to HEAD.** `206b879` still described the two-statement delete that `d9614df` had replaced one commit earlier, and omitted the cursor premise entirely. Now describes the transaction, why order alone is insufficient in either direction, why the row is created before being deleted, and the both-columns cursor rule.

### Explicitly refuted

**No deadlock** between the delete transaction and `upsertFitnessRouteHeatmapTiles` — both take `fitness_route_heatmap_pyramids` then `fitness_route_heatmap_tiles`, no path in the repo inverts that, and 20 rounds of four-way concurrent delete/upsert against real PostgreSQL produced zero deadlocks. **The rewritten heartbeat test is not flaky** — SQLite stores these as epoch-ms integers and PG as `timestamptz`, the 5 ms wait plus two round trips guarantees a strictly greater value; 20/20 green on SQLite, 35/35 for the whole file on PG. (MySQL's `TIMESTAMP(0)` truncation would break it, but the test matrix only constructs sqlite and pg.) And the CAS matches for **every** state reachable through the public API when the row genuinely does not change — only the half-cursor rows in finding 1 broke that.

Full suite green: 812 files / 9510 tests. Round 4 running against these fixes.

---

llun:
## Review round 4 — 2 findings, both fixed in `8e7ace1`

Converging: round 3 found 5, round 4 found 2, both latent. The reviewer's main job this round was to verify round 3's transactional-delete fix actually delivers what it claims, on a real database — it does.

### Verified, not just reasoned

A PostgreSQL 17 cluster loaded from `migrations/schema.sql`, running the exact race against the unmodified mixin:

- **Without** the insert (pre-`ee8dfa3` body): the concurrent claim + flush ran straight through the window and the tile delete removed them — final state a `generating` pyramid over **zero tiles**. Bug reproduced.
- **With** it: the claim **blocked** at its own `INSERT … ON CONFLICT DO NOTHING` on PostgreSQL's speculative-insertion wait, resumed after commit, and its tiles survived. The "sees no conflict, both proceed" hypothesis is refuted — an inserted-then-deleted tuple is still visible to a dirty snapshot as an in-progress insert, so the conflicting inserter waits.
- 120-iteration randomised stress: **0 orphan states**, no deadlock, and clean behaviour both when a clear arrives during a flush and when two clears race on an actor with no row.

### Findings

**1 · The insert closes the delete→claim ordering; the mirror ordering threw instead of losing cleanly.** The claim inserts its row and reads it back in a *separate* statement, under a comment asserting the insert guarantees a row. It doesn't — `ON CONFLICT DO NOTHING` takes no lock on the row it conflicts with, so a clear can delete it in between, and `readPyramidRow` then returns null into `throw new Error('Failed to create route heatmap pyramid row…')`. Staged deterministically, and also hit **4 times in 120** unstaged runs.

Attribution is worth stating: the throwing code came from #1452 and is already on `main`, but on `main` the DELETE route never touched these tables — **this PR is what makes it reachable**. It now retries, because a vanished row means it was just cleared and a claim arriving after a clear should build. The throw survives only for a clear winning three attempts in a row, which is the owner clearing repeatedly rather than anything being broken.

**2 · The test I renamed in round 3 still carried a name and comment about a mechanism it never exercises.** `status` decides `resumed` on its own and short-circuits before the cursor is read, so the setup's `cursor: null` was inert — every assertion holds with the cursor left in place, and dropping the cursor half of the `resumed` decision entirely fails four other tests but leaves this one green. It does earn its keep as the unique detector for the fresh branch's `scannedCount: 0` reset, so it is renamed to that, and the cursor is now deliberately left in place so nothing suggests otherwise. Cursor-driven freshness is covered by `starts a fresh version when an abandoned build never checkpointed`.

### Refuted

`cursorId: ''` under round 3's new null-check semantics is **not** a defect: it would make the keyset emit `createdAt = T AND id > ''`, an over-scan rather than a skip, and it is unreachable anyway since the only writer sets both columns or neither. Consistent-and-slightly-over-inclusive strictly beats the permanent wedge it replaced. The removed methods have zero remaining references. The insert leaves nothing behind on any failure path and does not change the returned count. And all three of round 3's test changes were re-verified by mutation — including that `new Date(0)` really is bound as integer `0` by better-sqlite3, so the half-cursor test genuinely exercises the SQLite-falsy row (3 of its 4 rows wedge under the old code; the 4th is the control).

### Pre-existing, not from this PR

`lets exactly one of two concurrent claimers win over …` is **flaky on PostgreSQL** (2/5 runs) — it assumes both claimers read before either writes, which only SQLite's max-1 pool guarantees. Production behaviour is correct; the test's premise isn't. It fails identically at the parent commit, was added in #1452 and is already on `main`, and CI never sees it because `ci.yml` pins `TEST_DATABASE_TYPE: sqlite`. Flagged only so it isn't attributed here; happy to fix it in a separate `fix:` PR rather than widen this one.

Full suite green: 812 files / 9511 tests. Round 5 running.

---

PR #1457, opened 2026-08-18
minor: add the route heatmap tiler — zoom-independent precision, as pure functions

llun:
## Summary

Phase 3 of the fitness route heatmap work, stacked on #1452 and #1454 (both merged). **This is where the precision actually changes** — but it ships **inert**: four pure modules with no database, no job, no UI. Nothing calls them until the generation job is reworked in Phase 4, which is exactly what makes this reviewable on its own.

### The problem it solves

Today the whole heatmap is one blob fitted to a single global vertex budget (80,000 points), by doubling the simplification tolerance from 1 m up to 256 m until it fits. So precision degrades as coverage grows: a Singapore-sized heatmap lands at the 1 m floor and looks sharp, while a world-spanning one is coarsened to hundreds of meters and cuts across roads. Zooming in never reveals more, because the geometry was simplified once and is reused at every zoom.

The pyramid inverts that. Each zoom is simplified to **its own pixel**, so detail per screen pixel is constant by construction — 9.8 km at z4, 611 m at z8, 38 m at z12, 2.39 m at z16 — and no longer depends on how far an actor has ridden.

### What's here

| Module | |
|---|---|
| `constants.ts` | the ladder, the extent, the per-zoom tolerance, the heat ramp |
| `tileCodec.ts` | the canonical tile encoding + the projection helpers, including the inverse the repo lacked |
| `tiler.ts` | route → per-zoom tile edges |
| `edgeMerge.ts` | edges ↔ stored polylines, and the version-aware merge |

**The codec owns the format outright.** The database layer stores a tile payload as an opaque string and never parses it (its own comment: *"the build path hands it straight to the tile codec"*), so nothing else had to change to adopt this.

Some decisions worth surfacing:

- **`TILE_EXTENT` equals the Web Mercator tile size on purpose**, so one stored unit is one screen pixel. That identity is what collapses GPS jitter from repeat rides onto *shared* edges rather than near-miss polylines — it's why a street ridden fifty times stores one edge with a count of fifty.
- **No maximum-points constant anywhere.** Per your earlier call ("as much as possible"), tile size is bounded by physical density — 1-pixel tolerance plus quantized-edge dedup means a tile holds only as much as the road network inside it.
- **Counts are per activity, not per traversal.** An out-and-back street or a repeated lap counts once: heat should say how often a street is *used*, not how a ride was routed, and a per-traversal count would be unstable under jitter. Documented trade-off.
- **`tileLocalToLngLat` is new**, not reused — `projectWebMercator` has no inverse and clipping a tile to a shared region needs one.

### The parts that fail silently if wrong

An adversarial design review before implementation found seven concrete holes in the algorithm; all are fixed here and each is pinned by a test that fails when the fix is reverted:

1. **Vertex packing uses base 257, not 256.** The coordinate space is inclusive at both ends (0..256 is 257 values, since a boundary point is 256 here and 0 in the neighbour), so base 256 collides `(1,0)` with `(0,256)` — welding two unrelated streets together.
2. **Tile index clamped at both ends.** The projection's own edges fall outside the grid on both sides: `lng = +180` and the southern Mercator limit land one *past* the last tile, and the northern limit floors to `-1`. (My own test caught the `-1` case during implementation — I'd only clamped the top.)
3. **Assembly partitions by `(count, hidden)`** — a stored polyline carries one count for all its points, so mixing a twice-ridden edge with a once-ridden one has to pick a count, and either choice corrupts the other.
4. **Assembly sweeps for cycles.** A closed loop — roundabout, lap, criterium — has *no odd-degree vertex*, so walking only from odd degrees never enters it and every edge in it disappears without trace.
5. **Boundary crossings are solved once in the continuous frame**, then localized by subtracting the integer origin (exact). Re-solving inside each tile's shifted frame is equal in maths but differs in the last bit — which is how an intermittent one-pixel seam appears.
6. **Pairs subdivide at every crossing on both axes**, so a long diagonal contributes to every tile it passes through instead of just its endpoints'.
7. **Cascade bound is 4/3, not 1.25** (tolerances grow ~4× per step; the series sums to 4/3) — corrected in the docs.

Plus: a non-finite point ends the run rather than being filtered later, because simplification anchors its projection at the **first** point's latitude — a leading `NaN` makes every projected coordinate `NaN` and collapses the whole run, losing the valid points after it.

### Tests

122 new tests. Every invariant above was mutation-tested — reverting the behaviour fails exactly the test that names it and no others. Two mutants survived my first pass (y-axis subdivision, and the non-finite skip, which is only distinguishable via the leading-`NaN` case) and I added tests for both rather than leaving them unguarded.

Also pinned: the heat ramp reproduces today's additive-alpha stacking exactly (`1-(1-0.55)^n`, computed from the formula rather than copied literals), tolerant decoding (one bad segment is dropped, not the tile), and that `lib/clientModuleBoundary.test.ts` stays green — the dir is under `lib/services/` and must be importable from client code, so it carries no `'use client'` and reads no environment (the tolerance floor arrives as a plain number).

Full suite: 816 files / 9633 tests. No migration, so both schema dumps are untouched. No UI surface yet — that lands in Phase 6.

### Next

Phase 4 wires this into the generation job. One integration constraint is already recorded in the plan: the tiler returns per-activity **deltas** rather than encoded tiles precisely so the job can fold them into an in-memory per-tile map and merge to storage only at each 20 s checkpoint — merging per activity would re-explode a hot downtown tile's whole growing edge set every time.

## Checklist

- [x] Docs updated: no command, env var, route, script or convention changed by this PR (the storage doc already describes the pyramid; Phase 4 updates it when the tables go live)
- [x] If migrations changed: none in this PR, so both dumps are unchanged

https://claude.ai/code/session_01ELw2T3kDEwCRZjdoveDefw

---
_Generated by [Claude Code](https://claude.ai/code/session_01ELw2T3kDEwCRZjdoveDefw)_

### DIFF
--- a/lib/services/fitness-files/heatmapTiles/constants.test.ts
+++ b/lib/services/fitness-files/heatmapTiles/constants.test.ts
@@ -1,0 +1,193 @@
+import {
+  HEAT_COUNT_SATURATION,
+  HEAT_HIDDEN_BASE_OPACITY,
+  HEAT_VISIBLE_BASE_OPACITY,
+  TILE_EXTENT,
+  TILE_LADDER_ZOOMS,
+  TILE_MAX_ZOOM,
+  TILE_MIN_ZOOM,
+  TILE_SIMPLIFY_TOLERANCE_PX,
+  heatOpacityForCount,
+  metersPerPixelAtZoom,
+  tileToleranceMeters
+} from '@/lib/services/fitness-files/heatmapTiles/constants'
+import { TILE_SIZE } from '@/lib/utils/webMercator'
+
+describe('metersPerPixelAtZoom', () => {
+  // Measured ground resolution, the figure the whole ladder is chosen from.
+  it.each([
+    {
+      description: 'z4 at the equator',
+      zoom: 4,
+      latitude: 0,
+      expected: 9783.94
+    },
+    { description: 'z8 at the equator', zoom: 8, latitude: 0, expected: 611.5 },
+    {
+      description: 'z12 at the equator',
+      zoom: 12,
+      latitude: 0,
+      expected: 38.22
+    },
+    {
+      description: 'z16 at the equator',
+      zoom: 16,
+      latitude: 0,
+      expected: 2.39
+    },
+    // cos(60) is exactly 0.5, so this pins the latitude scaling itself.
+    {
+      description: 'z16 at 60N, half the equator figure',
+      zoom: 16,
+      latitude: 60,
+      expected: 1.19
+    }
+  ])('reports $description', ({ zoom, latitude, expected }) => {
+    expect(metersPerPixelAtZoom(zoom, latitude)).toBeCloseTo(expected, 2)
+  })
+
+  it('halves for every zoom level gained', () => {
+    for (let zoom = 1; zoom <= 18; zoom += 1) {
+      expect(metersPerPixelAtZoom(zoom, 0)).toBeCloseTo(
+        metersPerPixelAtZoom(zoom - 1, 0) / 2,
+        6
+      )
+    }
+  })
+})
+
+describe('tileToleranceMeters', () => {
+  it.each([
+    { description: 'z8 at the equator', zoom: 8, latitude: 0, expected: 611.5 },
+    {
+      description: 'z12 at the equator',
+      zoom: 12,
+      latitude: 0,
+      expected: 38.22
+    },
+    { description: 'z16 at the equator', zoom: 16, latitude: 0, expected: 2.39 }
+  ])(
+    'passes $description through, being above the floor',
+    ({ zoom, latitude, expected }) => {
+      expect(tileToleranceMeters(zoom, latitude, 1)).toBeCloseTo(expected, 2)
+    }
+  )
+
+  it('clamps to the floor where a pixel is finer than GPS noise', () => {
+    // z16 at 80N is 0.41m/px — below the 1m floor, so simplifying to it would
+    // preserve jitter as though it were shape.
+    expect(metersPerPixelAtZoom(16, 80)).toBeLessThan(1)
+    expect(tileToleranceMeters(16, 80, 1)).toBe(1)
+  })
+
+  it('is the per-pixel figure times the pixel tolerance, not a fixed table', () => {
+    // Pinned as the relation rather than copied numbers: change the pixel
+    // tolerance and every level must move with it.
+    for (const zoom of TILE_LADDER_ZOOMS) {
+      expect(tileToleranceMeters(zoom, 0, 0)).toBeCloseTo(
+        metersPerPixelAtZoom(zoom, 0) * TILE_SIMPLIFY_TOLERANCE_PX,
+        6
+      )
+    }
+  })
+
+  it('never returns less than the floor at any ladder zoom or latitude', () => {
+    for (const zoom of TILE_LADDER_ZOOMS) {
+      for (const latitude of [0, 45, 60, 80, 85]) {
+        expect(tileToleranceMeters(zoom, latitude, 1)).toBeGreaterThanOrEqual(1)
+      }
+    }
+  })
+})
+
+describe('heatOpacityForCount', () => {
+  it.each([
+    { count: 1, expected: 0.55 },
+    { count: 2, expected: 0.7975 },
+    { count: 3, expected: 0.908875 },
+    { count: 4, expected: 0.95899375 }
+  ])(
+    'stacks $count visits the way overdrawing $count translucent lines did',
+    ({ count, expected }) => {
+      expect(heatOpacityForCount(count, HEAT_VISIBLE_BASE_OPACITY)).toBeCloseTo(
+        expected,
+        8
+      )
+    }
+  )
+
+  it('reproduces the additive-alpha formula exactly, at both bases', () => {
+    // The compatibility-bearing part: computed from the formula rather than
+    // compared against copied literals, so the two can never drift.
+    for (const base of [HEAT_VISIBLE_BASE_OPACITY, HEAT_HIDDEN_BASE_OPACITY]) {
+      for (let count = 1; count <= HEAT_COUNT_SATURATION; count += 1) {
+        expect(heatOpacityForCount(count, base)).toBeCloseTo(
+          1 - (1 - base) ** count,
+          10
+        )
+      }
+    }
+  })
+
+  it('saturates past the clamp instead of climbing forever', () => {
+    const saturated = heatOpacityForCount(
+      HEAT_COUNT_SATURATION,
+      HEAT_VISIBLE_BASE_OPACITY
+    )
+    for (const count of [HEAT_COUNT_SATURATION + 1, 50, 10_000]) {
+      expect(heatOpacityForCount(count, HEAT_VISIBLE_BASE_OPACITY)).toBe(
+        saturated
+      )
+    }
+  })
+
+  it.each([
+    { description: 'zero', count: 0 },
+    { description: 'a negative count', count: -3 }
+  ])('floors $description at a single visit', ({ count }) => {
+    expect(heatOpacityForCount(count, HEAT_VISIBLE_BASE_OPACITY)).toBeCloseTo(
+      HEAT_VISIBLE_BASE_OPACITY,
+      10
+    )
+  })
+
+  it('never leaves the drawable range', () => {
+    for (const base of [HEAT_VISIBLE_BASE_OPACITY, HEAT_HIDDEN_BASE_OPACITY]) {
+      for (const count of [0, 1, 3, 6, 99]) {
+        const opacity = heatOpacityForCount(count, base)
+        expect(opacity).toBeGreaterThan(0)
+        expect(opacity).toBeLessThanOrEqual(1)
+      }
+    }
+  })
+})
+
+describe('format invariants', () => {
+  it('keeps one stored unit equal to one screen pixel', () => {
+    // The identity the whole design rests on: quantizing to this collapses GPS
+    // jitter onto shared edges instead of near-miss polylines.
+    expect(TILE_EXTENT).toBe(TILE_SIZE)
+  })
+
+  it('stops the ladder at the GPS-noise floor', () => {
+    // Finer than ~2.4m/px would encode jitter rather than road.
+    expect(metersPerPixelAtZoom(TILE_MAX_ZOOM, 0)).toBeCloseTo(2.39, 2)
+    expect(metersPerPixelAtZoom(TILE_MAX_ZOOM + 2, 0)).toBeLessThan(1)
+  })
+
+  it('is ascending, unique, and spans its own min and max', () => {
+    expect([...TILE_LADDER_ZOOMS]).toEqual(
+      [...TILE_LADDER_ZOOMS].sort((a, b) => a - b)
+    )
+    expect(new Set(TILE_LADDER_ZOOMS).size).toBe(TILE_LADDER_ZOOMS.length)
+    expect(TILE_MIN_ZOOM).toBe(Math.min(...TILE_LADDER_ZOOMS))
+    expect(TILE_MAX_ZOOM).toBe(Math.max(...TILE_LADDER_ZOOMS))
+  })
+
+  it('steps evenly, so each level costs a quarter of the one below', () => {
+    const steps = TILE_LADDER_ZOOMS.slice(1).map(
+      (zoom, index) => zoom - TILE_LADDER_ZOOMS[index]
+    )
+    expect(new Set(steps)).toEqual(new Set([2]))
+  })
+})
--- a/lib/services/fitness-files/heatmapTiles/constants.test.ts
+++ b/lib/services/fitness-files/heatmapTiles/constants.test.ts
@@ -129,6 +129,48 @@ describe('heatOpacityForCount', () => {
     }
   })
 
+  it('pins the saturation point itself, not just its own consequences', () => {
+    // Every other assertion here derives from the constant, so they hold for
+    // any value it takes. Six is the point where the ramp reaches 0.99 and
+    // further visits stop being visible.
+    expect(HEAT_COUNT_SATURATION).toBe(6)
+    expect(heatOpacityForCount(7, HEAT_VISIBLE_BASE_OPACITY)).toBeCloseTo(
+      1 - 0.45 ** 6,
+      10
+    )
+    expect(heatOpacityForCount(5, HEAT_VISIBLE_BASE_OPACITY)).toBeCloseTo(
+      1 - 0.45 ** 5,
+      10
+    )
+  })
+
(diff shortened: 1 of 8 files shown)


llun:
## Review round 1 — 20 findings, all fixed in `27a2f8a`

Two sub-agents: tiler/codec correctness, and test power across all four files. `edgeMerge`'s trail decomposition came back **clean** — verified against 25,000+ randomized graphs plus hand-built adversarial shapes (flowers, K4–K8, cycle-touching-path, 500-edge paths), zero edges lost or duplicated, and measured linear (8k/16k/32k/64k edges → 21/38/69/131 ms).

### The one that matters — a bad GPS fix was 200× write amplification

Not a crash, not wrong pixels — a **cost** bug, and it only exists because of the tiled design. A single dropout inside the legal coordinate range draws a straight line from the route to that point and back. In the untiled heatmap that cost one wrong stripe; here **every tile the stripe crosses is a stored row**. Measured on a 500-point Amsterdam ride:

| input | tiles | edges |
|---|---:|---:|
| clean ride | 84 | 84 |
| + one `{0, 0}` "null island" fix | **16,563** | 31,650 |
| + one `{52.4, −175}` fix (legal lng, under the antimeridian threshold) | **44,918** | 83,968 |

`{0,0}` is the single most common GPS artifact, and `parseFitnessFile` has no reason to reject it — it only bounds `|lat|>90` / `|lng|>180`.

Fixed with a runaway guard: a pair spanning more than **1,024** tiles is dropped, leaving the run either side of the bad fix intact. The bound is measured, not guessed — the longest genuinely straight road on earth (~150 km of the Eyre Highway) simplifies to a single pair and stores **395 tiles at this cap and 395 with the cap four times higher**, so nothing real is being clipped, while the glitched ride falls to 1,090. At 256 that road starts losing a quarter of itself (103 tiles). The guard is pinned from *both* directions: tests fail if it's removed or widened to 65,536, **and** fail if it's tightened to 256.

### Also fixed

- **`parseTileKey` coerced rather than rejected.** `Number('')` is 0, so a truncated `"12:2103:"` resolved to a real but ~1,346-rows-away tile instead of being turned away; `"0x4:1:1"` → z4, `"1e1:5:5"` → z10, `" 4 :1:1"` accepted. And above zoom 1023 `2 ** z` is `Infinity`, making the index bounds vacuous. Now digit-matched per part and bounded by the ladder.
- **Unreachable count branch deleted** — the per-activity dedup guarantees an activity contributes each edge exactly once, so `existing.count += 1` could only ever have double-counted *within* one activity, which the comment beside it forbids. Confirmed dead by instrumentation over 3,000 randomized retraced tracks.
- `u`/`v` are now clamped like the index, since the projection's edges land a hair outside the space the docstring promises.

### Tests that named a behaviour they couldn't fail on

Six of these, and they're the interesting half of the round:

- **The cascade's direction was completely unpinned.** Reversing the sort passes all 122 tests while building z16 from the 611 m z8 simplification — **83% of the finest detail gone**. Both cascade tests missed it because their fixtures collapsed to two endpoints at every zoom, and the one named "stores *strictly* less" asserted `>=`, which three identical levels satisfy. Now uses a multi-scale fixture and strict `>`.
- The non-finite test left **one** point either side of the `NaN`, so both runs were dropped and every assertion ran over an empty map — it passed with the whole function stubbed to `() => new Map()`.
- Two discontinuity tests had upper bounds only (`<= 2`), which **zero** satisfies. One of them was *also* silently crossing the antimeridian itself.
- The odd-length payload fixture was already rejected by the minimum-length guard, so deleting the parity check passed — and let `countTilePoints` return **4.5** into an integer column.
- The dense-tile test's fixture split every edge into its own count class: max degree **1**, so the walker never took a second step and the cursor was never consulted. Its wall-clock assertion was also a genuine CI flake vector.
- `HEAT_COUNT_SATURATION`, `HEAT_HIDDEN_BASE_OPACITY` and the `Math.round` on counts were all tautologically "pinned" against themselves.

Eleven of 22 tiler tests passed against a stubbed-empty implementation; they now assert non-emptiness first.

Two comments were also overstating the code: the odd-degree pass is a **storage** optimisation (the sweep after it is what carries correctness — it would be complete on its own), and anchoring simplification at a run's first latitude errs toward detail only when travelling *poleward*.

### Verification

Every fix mutation-tested — 9 of 9 mutants killed, including the two that survived my first attempt. Full suite green on the rebased branch: 816 files / 9,657 tests. Round 2 running against these fixes.

---

llun:
## Review round 2 — 5 findings, all fixed in `ba21474`

Attacking round 1's own fixes. The headline finding is that **round 1's glitch guard was itself broken**, in a way that only shows up when you think about what a tile count means.

### 1 · HIGH — the guard was zoom-relative, so it couldn't fire at half the ladder

`MAX_TILE_CROSSINGS_PER_PAIR = 1024` was a tile count applied per zoom — but a tile count only means anything relative to how many tiles exist. The world is **16** tiles wide at z4 and **65,536** at z16, so the maximum possible Manhattan span is 30 at z4, 126 at z6, 510 at z8 — all below the bound. The guard was mathematically unreachable at three of the seven levels, and the same pair was "travel" at z12 and "a glitch" at z16.

Measured, same Amsterdam ride with one `{0,0}` fix:

| | z4 | z6 | z8 | z10 | z12 | z14 | z16 | total |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| clean | 1 | 2 | 2 | 3 | 7 | 24 | 92 | **131** |
| glitched | 3 | 12 | 48 | 192 | **768** | 24 | 92 | **1,139** |

So ~89% of what round 1 "saved" was still the glitch, just relocated to coarser zooms — and the user-visible result is worse than a uniform bug: the line from Amsterdam to the Gulf of Guinea is drawn at every zoom someone actually looks at, then **disappears when they zoom in**. The pyramid contradicting itself.

Fixed by reaching the verdict **once**, against `TILE_MAX_ZOOM`, in the run split — so every level builds from the same runs and no zoom can disagree about whether a stretch of road exists.

That also let the antimeridian rule go: a date-line-straddling pair projects to opposite edges of the world, which is the largest span there is, so the separate longitude test had become a branch nothing could reach. One rule instead of two.

### 2 · The anchor comment I rewrote in round 1 had both directions backwards

I wrote "over-detailed running poleward, under-detailed running equatorward". It is the reverse, and the reviewer proved it by execution on an L-shaped fixture: starting at the equator keeps **2** vertices on the far leg, starting at lat 60 keeps **237**. Tolerance scales with `cos(latitude)`, so a run *starting* nearer the equator carries the **coarser** tolerance into a finer-pixelled far end.

This mattered: the original wording was at least conservative ("errs towards keeping shape"), and my rewrite told a reader the error only ever *adds* detail — when poleward is precisely the direction that silently loses it.

### 3–5 · Three tests pinned less than they read as

- **Both cases covering the tile index bound (`x >= 2 ** z`) sat below the ladder**, so the zoom bound I added in round 1 turned them away first and they never reached the check. Deleting the index bound entirely stayed green — and `parseTileKey('4:16:0')` would have returned a tile that does not exist. (Round 1 spotted exactly this for the *positive* case and moved `'2:3:3'` → `'4:15:15'`, but left the two negatives behind.)
- **The non-finite count case used `null`**, which the type test catches one condition earlier — so `!Number.isFinite(c)` was never evaluated, and `{"c":1e999}` would decode as `count: Infinity` straight into the heat ramp. Now pinned with `1e999`.
- **The `u`/`v` clamp had no assertion at either edge it was added for.**

Plus a duplicate `it.each` description that made two different guards look like the same case.

### Verified sound (no action)

The reviewer confirmed by execution that the Manhattan span measure is a tight proxy (worst `tilesStored/(span+1)` ratio over 3,000 random z16 pairs was exactly **1.0000**; perfect 45° diagonals stay under it), that dropping a pair does **not** produce a wrong polyline (both pairs adjacent to a bad fix go, the real neighbours survive as endpoints, and the chain rejoins if they quantize together), that the Eyre Highway is genuinely not clipped (334 tiles at z16), and that round 1's cascade-direction and glitch-guard tests really do fail on the mutations they name.

### Verification

7 of 7 mutants killed, including the cap bounded from **both** sides (widening to 65,536 fails; tightening to 64 fails) and the verdict-at-max-zoom pinned (judging per zoom fails). Full suite green: 816 files / 9,660 tests. Round 3 running.

---

llun:
## Review round 3 — 3 findings, all fixed in `a5290ca`

**The tiling maths came back clean this round** — everything found was calibration and documentation. Converging: 20 findings in round 1, 5 in round 2, 3 here.

### 1 · The cap wasn't actually pinned — including by the test that claimed to pin it

Round 2's test comment said it checked the threshold "from both sides". It didn't: the fixtures sat at **182** tiles on the keep side and **32,585** on the split side, nowhere near the boundary. Measured by mutation, `MAX_TILE_SPAN_AT_FINEST_ZOOM` could be set anywhere in **[364, 12124]** with all 149 tests green — free to move 2.8× tighter or 11.8× looser undetected.

That's not academic. Tightening to 384 — a plausible "be safer about write amplification" edit — passes the whole suite while splitting a dead-straight 150 km road at 52°N (398 tiles), which is *the exact case the constant's own doc cites as kept whole*.

Now straddled at the equator, where 5.6° spans 1,019 tiles and 5.7° spans 1,037. The passing band is **[1019, 1036]** — ±1.7%, verified by sweeping the cap across 512/900/1000/1019/1020/1024/1037/1038/1100/2048/4096.

This is the same defect class round 2 was fixing, which is a fair thing to notice about my own work: I keep writing tests whose *comment* describes a stronger check than the fixture delivers.

### 2 · Two of the constant's worked examples were wrong

The docblock is effectively the specification for a constant nothing else pins, so a wrong figure there is a real defect:

- **"a 370 km train-journey gap spans 1,510 and is split"** — measured, it spans **1,029** due north and **983** due east from 52°N. So it sits *on* the boundary, not safely past it. The honest example is 400 km at 1,116; the doc now also says the 370 km case is marginal and direction-dependent, rather than implying a clean edge.
- **"deliberately a tile count rather than a distance"** — that stopped being true the moment round 2 pinned the verdict to a single zoom. 1,024 z16 tiles *is* a distance scaled by `cos(latitude)`: **626 km** at the equator, **386 km** at 52°N, **313 km** at 60°, **55 km** at 85°. A tile count is still the right invariant (stored rows are what's being bounded, not kilometres), but the guard is genuinely far tighter near the poles and the doc now says so.

### 3 · Test comments described the mechanism this branch replaced two commits ago

`describe('the glitch guard')` and "the pair is dropped, not the run" date from round 1's per-pair skip. Round 2 replaced it with a run split, so a maintainer reading them would not expect the reference latitude to be re-derived from the second half, nor a one-point remainder to be discarded. Renamed and corrected. Also fixed a comment giving the projection's edge offset as `-4e-7` — that's the continuous-frame figure; in local units at z16 it's `-1e-4`, and a test added last commit already said so right next to it.

### Refuted by execution (worth recording)

- **The antimeridian deletion is safe.** Swept `lng1 × lng2` over [-180, 180] at 0.25° across 15 latitudes including ±85.05 and ±89.9: the **minimum** span for `|Δlng| ≥ 180` is **32,768**, 32× the cap. Pole compression can't help — Mercator x is latitude-independent. The deleted rule was also *wrong in both directions*: it split `-179.9 → 180.1` (which normalizes to the same longitude) and missed `179.9 → 180.1` (a real wrap).
- **Splitting the run loses nothing it shouldn't** — traced through the real code for a mid-ride glitch (9 tiles, identical to clean), glitch-first, glitch-last, two consecutive glitches, and alternating good/bad. And it's strictly better than round 1 on short routes: a 100 m street with one dropout stored **757 tiles at z12** under the old per-pair skip versus **1** now.
- **Cost is negligible** — 18.5 ms of a 129.5 ms full-ladder build for an 80,000-point activity.
- **Pole aggressiveness is by design** — the guard can only ever remove the interpolated line across an *unrecorded* gap, never a recorded point.

### Verification

Full suite green: 816 files / 9,662 tests; lint clean; build OK. Round 4 running.

---

llun:
## Review round 4 — 1 low-severity finding, fixed in `1b5b0a1`. Loop converged.

Findings by round: **20 → 5 → 3 → 1**, and this last one is documentation. Round 3 already found nothing in the tiling maths; round 4 found nothing there either, and additionally re-verified every numeric claim in the codebase by execution.

### The finding

The span cap's docblock offered a 370 km gap as a worked example of a split, as `(1,029 north, 983 east)`. Those two figures **cannot be one measurement**: at the start latitude that yields 983 east, the same distance north is **1,024** — exactly the cap, therefore *kept*. So the "direction-dependent split" the sentence asserted does not happen there.

It's worse than an arithmetic slip, and that's the interesting part: 370 km northward reads as **1,022** tiles against a WGS84 meridian degree and **1,029** against an equatorial one — landing on *either side of the cap* depending only on which earth model measured the gap. A plausible real-world distance sits inside the measurement noise of my own threshold.

That's worth stating outright rather than papering over with a number, so the docblock now gives the boundary as the range it is (~370–390 km at 52°N) and keeps only the 400 km example, which is unambiguous at 1,116. The constant is otherwise specified *only* by its own comment, so a Phase 4 integration test written from the old wording would have asserted a split that doesn't occur.

### Verified by execution this round

Every figure now in the docblocks was re-measured against the shipped projection:

| claim | measured |
|---|---|
| `[0, 5.6]` = 1,019 / `[0, 5.7]` = 1,037 | exact |
| old fixtures sat at 182 and 32,585 | exact |
| 1,024 z16 tiles = 626 km equator / 386 km 52°N / 55 km 85°N | 626.2 / 385.5 / 54.6 |
| Eyre Highway = 335 tiles | 335 |
| 400 km at 52°N = 1,116 | 1,109–1,119 across readings; splits at every one |
| antimeridian minimum 32,768 tiles at every latitude | infimum exactly 32,768 |
| north Mercator limit = −1e-4 local units at z16 | −1.0448e-4 |
| "cannot fire below z10" | z9 max span 1,022 ≤ cap; z10 = 2,046 > cap |

And the pins hold: the cap's passing band is **exactly [1019, 1036]** (1018 and 1037 both fail), and the straddle test is *not* over-fitted — `EQUATOR_METERS_PER_PIXEL_AT_ZOOM_0` isn't even in the span's dependency chain (mutating it to 158,000 leaves all tests green), and the fixtures sit 0.31 and 0.51 tiles clear of a floor boundary against float error of ~1e-12°.

Fresh eyes on the parts four rounds hadn't attacked — `metersPerPixelAtZoom`'s formula, the `as const` ladder typing, the `tilesForBounds` ↔ tiler ↔ DB column mapping, and the boundary-vertex sharing in both float directions — found nothing. The unused exports are all deliberate Phase 4 surface.

### Status

Local gate green: 816 files / **9,662 tests**, lint clean, build OK. CI green on the previous head (`a5290ca`) across CI Success, All Tests, all four shards, Build, Lint, Schema Dump Sync, CodeQL and Analyze; re-running now on `1b5b0a1`, which changes one comment and one test fixture literal.

I consider the loop converged and this ready for your review.

---

PR #1463, opened 2026-08-19
minor: build the route heatmap tile pyramid beside the existing heatmap

llun:
## Summary

Phase 4 of the zoom-independent heatmap work: the generation job now **builds the tile pyramid**, so the tables Phase 1 created and the tiler Phase 3 wrote finally have a writer.

This is deliberately **additive**. The live heatmap still renders from the `segments` blob on `fitness_route_heatmaps`, and the variant it uses is exactly the one the pyramid covers — so a job that stopped writing the blob would blank the map until the serving API and client LOD land. Instead the legacy path is untouched and tiles are folded beside it. Nothing reads the pyramid yet; nothing visual changes.

Relates to #1452, #1454, #1457.

> **Note:** this description matches the code as it stands at `e06d3c9`. The resume and completion designs have been through several review-driven rewrites; the PR comments record what each earlier version got wrong and why, and are worth reading in order if you want the reasoning rather than the result.

### What the job does now

`isPyramidVariant = activityType === null && periodType === 'all_time'` — the only variant the heatmap UI asks for. Region is orthogonal: tiles are stored unclipped and a region is applied when they are served, so whichever region row wins the claim builds the same per-actor pyramid. Every other variant stays entirely on the blob path.

1. **Claim** the pyramid — but only if this pass could ever own the result. A pass carries either an offset of zero or the build's own token; anything else (the heatmap API's retry of a failed region row, which sets `resume: true` with that row's offset and no token) never touches the row, because the claim's compare-and-swap is destructive.
2. **Fold** each activity's privacy segments through `buildTileDeltasForActivity`, *before* the region filter, into an in-memory `tileKey → edges` map. Merging per activity into that map rather than into tile rows avoids the O(E²) re-merge on a downtown tile that thousands of activities cross.
3. **Flush** at the existing 20 s checkpoint, at `TILE_FLUSH_PENDING_LIMIT` pending tiles, and when the accumulation guard trips.
4. **Complete** by stamping the build's own counters — including `tileCount`/`pointCount` counted off the rows on disk — then sweeping `version < buildVersion`.
5. **Fail / cancel / decline** all release the build rather than sitting on it.

### The hard part: folding each activity exactly once

Tile visit counts **accumulate**. Folding an activity twice inflates its heat permanently — a wrong number rather than a missing one, and unlike a missing tile nothing downstream can detect it. Skipping one is the mirror hazard: a completed build certifies the hole as fine.

**A build folds an activity only if its own cursor says it has not already.** The scan runs `(createdAt, id)` descending and the build records the last file it took, so anything sorting at or after that cursor is skipped. Asking *where a file sits* rather than *how many came before it* is what makes the fold idempotent. The cursor advances for every file the pass finished with — including one whose download or parse threw, because a build with no cursor cannot be resumed at all — and it advances **with** the fold rather than after it, since a flush can fire part-way through an activity and writes its tiles and cursor in one statement.

**Resuming a build requires presenting its token, not declaring an intent.** Keeping a build's version means adding to its tiles, which is safe only for a pass carrying on from where that build left off. A claimer without a matching token gets a fresh version: a full rebuild that sweeps the old build away, which is merely wasteful where the alternative is silently wrong.

**Every fence names the pyramid row as well as `claimSeq`** — the sequence counts from zero per row and clearing an actor's heatmaps deletes the row, so token 1 before a clear and token 1 after it are different builds. That covers the claim, the tile flush, the progress/status write and the completion sweep.

**The claim is atomic.** Its compare-and-swap and the read that confirms it share one transaction, so a claim either happens and is reported or does not happen — and the row it reports is provably the row it wrote.

Tiles and the progress describing them land in **one guarded statement**, so the cursor a flush commits always covers exactly the tiles committed with it.

### What a build refuses to certify, and what it records

**A build only completes over a history it actually scanned**, counted again at the moment of the decision rather than trusted from the start of the pass. `completedAt` is what makes the next claim answer `already-fresh`, so certifying a short scan would refuse the very regenerate that would heal it. That catches an activity *added* while the build runs; it cannot catch one *deleted* from the already-scanned part, where the shift and the recount cancel out.

**A build that could not READ every file still completes, and says so.** An unreadable activity is skipped by the legacy blob too, and refusing to complete would wedge any actor with one permanently missing object — so the row records `Completed without N activities that could not be read` instead of `error: null`, carried across continuations because the pass that finishes a long history is rarely the one that hit the outage.

### Isolation from the legacy blob

Nothing reads the pyramid yet, so losing a build costs a rebuild while failing the run costs the user the heatmap they can actually see. Every tile-path error — the claim, the tiler, a flush, the completion — abandons the build, records why on the pyramid row, and lets the run finish the legacy heatmap. A build the pass was merely *carrying* is released from one place, the handler's `finally`, which also covers a throw between reading the token and making the claim.

### Known and deliberately deferred

- The scan is still the legacy integer paging, so an activity deleted from the already-scanned part is skipped for that build and healed by the next full generate. Keyset paging retires this in Phase 8.
- A build discarded because an activity arrived mid-run is not re-enqueued by anything: route heatmaps are generated on a user action or by the maintenance script.
- `getFitnessRouteHeatmapTilesInRange` has no `version` predicate. Harmless while nothing reads the pyramid, but whoever writes the Phase 5 serving route must filter by the pyramid's version or a viewport during a rebuild renders two builds at once.

### Tests

The first coverage of the `all_time` variant in this job's test file at all — the full ladder with counts, a street ridden twice reading 2, resume without double-count, the fold gate driven through a three-pass chain whose rows all share one millisecond, an already-folded activity ignored when an upload shifts the offsets, storage outages total and partial, and release on every exit — plus the legacy blob asserted byte-identical for an `all_time` run, and the mixin's claim guards staged against real read-vs-write interleavings.

**Nine sub-agent review rounds**, each over what the previous round changed; every finding is recorded in the comments below. Both touched suites run on SQLite and PostgreSQL 17.


llun:
## Sub-agent review — round 1

Four agents in parallel: the job's pyramid integration, the claim mixin's JS/SQL parity under concurrency, the new tests (mutation-tested, plus a PostgreSQL 17 run), and project invariants + docs. **17 findings, and one of them was a data-corruption bug that this PR shipped.** All are addressed below; details of each fix are in the follow-up commits.

### The critical one: a fresh generate silently doubled every count

**Confirmed by reproduction, not reasoning.** A pass dies mid-build, leaving the pyramid `generating` with a cursor. Two minutes later the owner hits Generate again — a *fresh* run, scanning from offset 0. The claim decided `resumed` from the row alone (`status === 'generating' && cursor !== undefined`), so that run inherited the dead build's **version**, and `mergeTileDelta` saw `existingVersion === buildVersion` and **added**. Two rides in two different countries, no repeats:

```
after crash:          {"status":"generating","version":1,"scanned":1,"hasCursor":true}
after fresh generate: {"status":"completed","version":1,"activityCount":2}
distinct visit counts (expected [1]): [1,2]
```

Permanent, and self-certifying: the version never moved, so completion's sweep could never clear it. The resume guard was gated on `isResume` and never ran. **Cancel-then-regenerate is a plain user action that lands directly on this.**

### The rework

Three findings (a tile-less checkpoint orphaning the build, retry-after-failure never being able to rebuild, an upload between two passes re-folding an activity) all traced to the same root: comparing an integer `scannedCount` against the legacy `cursorOffset` as the resume premise. Counting scanned files cannot distinguish honest progress from a page you have already seen. So the premise is now **positional**:

- **A build folds an activity only if its own cursor says it has not already.** The scan runs `(createdAt, id)` descending; anything sorting at or after the build's cursor is skipped. That makes the fold idempotent against a redelivered page, a shifted offset, and a lost progress write alike. The cursor advances for every file scanned, GPS or not.
- **A build is either fresh or continuing, and the caller says which** — `allowResume` on the claim, defaulting to `false`. A fresh run always gets a new version. A pass that can cover neither the whole history nor the build's cursor hands the build back rather than sitting on it.
- **Tiles and the progress describing them land in one guarded statement**, closing the window where tiles were on disk and the cursor still pointed before them.

The continuation token now names the pyramid **row** as well as `claimSeq`: a reviewer proved that clearing an actor's heatmaps deletes the row and the next build restarts at `claimSeq: 1`, so a continuation stranded across a clear could evict the live build that replaced it — and the collision lands on the likeliest value there is.

### Also fixed

- **Tile work no longer fails the user-visible heatmap.** A tile-store error used to throw into the shared catch and mark the legacy row `failed`. It now abandons the build, records why on the pyramid row, and lets the run finish the heatmap the user can actually see — which is what the PR body already claimed.
- **The build is released on cancellation and on every decline**, instead of leaving a `generating` row with a fresh heartbeat and no writer, blocking other claimants for 120 s.
- **`activityCount` is the pyramid's own**, not the legacy run's region-clipped one — tiles are stored unclipped, so the same tiles used to report different totals depending on which region row won the claim.
- **`tileCount`/`pointCount` are written at completion** (they were inserted as 0 and stayed 0 forever), counted from the rows on disk via a new `(actorId, version)` aggregate.
- `err: toLoggableError(...)` added to two catches; the ordering rationale in the docs and two code comments was stated backwards and is corrected; `docs/environment-variables.md` notes the tolerance floor's new second role.

### Test gaps closed

The tests reviewer built 30 mutants. Survivors, all now killed: `isPyramidVariant`'s activity-type half (load-bearing — `recreateFitnessRouteHeatmapJobs` enqueues an all-time variant *per activity type*, which would have folded only running rides into the shared pyramid and swept the rest); the cross-flush merge wiring (every fixture kept its routes too far apart to share a tile, so the accumulation path the comments were about was never entered); the mid-page flush path and both new constants; the flush ordering; "tiles are stored unclipped"; the page-limit guard.

**11 of 11 mutants killed** on the reworked code, including both halves of the resume token (JS and SQL separately) and the `(actorId, version)` SQL clause, which is pinned structurally because the JS verdict returns before the CAS whenever they agree.

### Verification

Gate green — prettier, lint 0 errors, build, full suite. End to end against SQLite with a checkpoint forced after *every* activity, so the whole build ran as a chain of continuations:

```
passes 4
pyramid {"status":"completed","version":1,"activityCount":4,"scannedCount":4,
         "tileCount":162,"pointCount":324}
total tiles 162 · distinct visit counts [1,2]
legacy blob completed activities=4
```

Version stays at 1 across three resumes, the one repeated route reads 2, nothing else inflated.

### Out of scope, worth its own issue

`generateFitnessRouteHeatmapJob.test.ts` is **already order-dependent at the base commit** — three pre-existing tests fail on roughly half of `--sequence.shuffle` seeds at `975950d`, because all tests share one actor and leak fitness files. This PR neither causes nor worsens it; the new block passes alone, in the whole file, and across 16 shuffle seeds on SQLite plus 5 shuffled runs on PostgreSQL 17.

Round 2 next.

---

llun:
## Sub-agent review — round 2

Run as an orchestrated workflow: four review dimensions → adversarial refutation of everything nobody reproduced → a completeness critic. 13 agents, 17 findings (6 reproduced, 6 survived refutation, 2 refuted, 3 unverified).

**Round 2 found that round 1's own fix reopened the hole it closed.** That is now three consecutive rounds each finding a defect in the previous round's fix.

### Critical — a retry could adopt a build it had no claim to

Round 1 replaced the counter-based resume premise with a positional one, and gated tile work on `pyramidClaim.resumed`. But the claim computed `resumed` from the caller's own `allowResume` flag — and **`resume: true` does not mean "I am this build's continuation"**. `app/api/v1/accounts/[id]/fitness-route-heatmap/route.ts` sets it for *any* retry of a failed or partial region row, carrying that row's offset and no pyramid token at all.

Reproduced end to end: three activities in three countries, a pass dies after checkpointing, the API retries at offset 2 with no token. The claim answers `resumed: true`, the version stays at 1, the run scans from offset 2, **activity #2 is never folded**, and the build finishes `status: 'completed', scannedCount: 2, activityCount: 2` against `totalCount: 3` — then sweeps every older version, so nothing is left to fall back on. A hole, certified as complete.

Fixed: resuming now **requires presenting the build's own token** (row id + `claimSeq`), which is the only evidence a caller can offer that it will carry on from where that build left off. `allowResume` is gone. A claimer without one gets a fresh version — a full rebuild that sweeps the old build away, which is merely wasteful where the alternative is silently wrong.

### High — a mid-activity flush committed tiles the cursor didn't cover

The cursor advanced *after* the two flush sites that can fire inside a single activity, so such a flush wrote that activity's tiles alongside a cursor still naming the previous file. A crash there left the tiles on disk with the build claiming not to have reached them, and the resume folded the activity again — measured histogram `{1: 221, 2: 201}` for three distinct rides.

The critic showed this is **routine, not exotic**: `shouldReduceAccumulation` returns true on a plain point counter (`pointCount >= accumulationPointLimit`), with no memory condition, so every large build enters that state repeatedly. Fixed by advancing the cursor with the fold, before any flush can observe it.

### High — the write fences didn't name the row

The claim was taught to name the pyramid row (because `claimSeq` restarts at zero when a clear deletes the row), but the *writes* still fenced on `(actorId, claimSeq)` alone. Reproduced: a pass stranded across a clear wrote its tiles **and** its cursor into the replacement build, rewinding a live build's progress. Both guarded writes now name the row.

### Also fixed

- The tiler fold ran inside the per-file `try` the legacy accumulation shares, so an exception there would have dropped the activity from the user-visible heatmap too — contradicting this PR's own isolation claim. Now contained like every other piece of tile work.
- `getFitnessRouteHeatmapTileTotals` gained a direct test for its version filter (it is what keeps about-to-be-swept tiles out of a completing build's summary).
- Three false claims in the rewritten docs and comments corrected — including a `progress` docblock asserting a gap-detection safeguard that no longer exists.
- The accumulate-once rule recorded in `AGENTS.md` (new **Fitness Route Heatmap Pyramid** section) and `REVIEW.md`, per Documentation Maintenance.
- A z4 fixture that cleared the degeneracy check by 1.46 px widened, and a duplicate variant test folded into its `it.each` table.

### Verification

**8/8 new mutants killed** (28 across all rounds), including `isOwnContinuation` on both the JS and SQL sides, the cursor-advance placement, each write fence independently, the totals version filter, fold containment and completion-failure release.

Gate green: prettier, lint 0 errors, build, and the full suite at **816 files / 9,696 tests**.

Round 3 is running now, over what round 2 changed.

---

llun:
## Sub-agent review — round 3

Six review dimensions → adversarial refutation of everything nobody reproduced → a completeness critic. 11 agents, **28 findings + 2 from the critic; 26 reproduced on both SQLite and PostgreSQL 17**, none refuted.

**Round 3 found the opposite failure from rounds 1 and 2.** Those found the permissive direction — an illegitimate pass adopting a build it had no right to, inflating counts. This round hunted the direction `8ba586c` opened by closing that: a **legitimate continuation being refused**. Two independent instances, both reproduced.

### High — a build refused its own continuation, and the whole chain then built nothing

Five of the six dimensions found this independently. `advancePyramidPast(file)` sat at the end of the per-file `try`, so a file whose download or parse threw was **scanned but never passed**. A first pass whose readable files all threw — a storage outage, which is also how the 20s budget gets consumed — therefore checkpointed with no cursor at all, and the claim reports `resumed` only for a build that *has* one.

That pass still published a continuation carrying the build's own valid token. The continuation presented it, was told it was not that build's successor, and — arriving at a non-zero offset — released the build it was sent to finish. Every later pass in the chain carried a non-zero offset too, so none of them could start a build either:

```
PASS 1 {"status":"generating","version":1,"claimSeq":1,"scanned":0,"tiles":0,"published":1}
PASS 2 {"status":"failed","version":2,"claimSeq":2,"scanned":0,"tiles":0,"published":1}
PASS 3 {"status":"failed","version":3,"claimSeq":3,"scanned":0,"tiles":0,"published":1}
PASS 4 {"status":"failed","version":4,"claimSeq":4,"scanned":0,"tiles":0,"published":0}
```

The legacy blob completed normally throughout, so nothing surfaced the loss. Fixed by moving the advance into a `finally`: a file the pass is finished with moves the cursor whether it folded geometry, held none, or could not be read. This also makes true the claim AGENTS.md already made — "the cursor advances for every file scanned, GPS or not" — which was false as written.

Deliberately **not** fixed by relaxing the resume premise to allow a null cursor: the continuation scans from a non-zero offset and would skip every file before it, which is the permissive hole rounds 1 and 2 closed.

### High — an ordinary retry demoted a completed pyramid and rebuilt nothing

The claim ran unconditionally; the coverage decision came after it. But the claim's compare-and-swap is destructive — it bumps `version`, stamps `generating`, and clears the counters, the cursor and `completedAt`. A run with `resume: true`, a non-zero offset and no pyramid token is exactly what the heatmap endpoint publishes when retrying a failed or partial region row, and such a pass can never legitimately own a build:

```
before: { status: 'completed', version: 1, activityCount: 2, tileCount: 426, completedAt: … }
after:  { status: 'failed',    version: 2, activityCount: 0, tileCount: 426, completedAt: undefined,
          error: 'Claimed by a pass that could not cover the build' }
tilesStillOnDisk: 426
```

A healthy pyramid reduced to a failed, empty row over tiles it no longer describes, with nothing rebuilt and the version churning on every pass of that retry chain. This was new damage from `8ba586c`: before it, that path resumed (wrongly); after it, it demoted.

Whether a pass could ever own a build is knowable from its job data, so it no longer claims at all — the decision moved in front of the claim rather than behind it.

### Medium — three more

- **A dropped continuation stranded its build.** The region-row staleness guard returns before the claim, and a continuation holds the only copy of its build's token. A second Generate resets the region row's offset, which makes the in-flight continuation stale; the continuation is then dropped and publishes nothing, while the second Generate is itself refused `build-in-progress`. The pyramid sat `generating` with a partial tile set, no writer and no queued job. It hands the build back now, fenced on the token it carries.
- **The completion sweep was the one unfenced pyramid write** — and it is the one that *deletes*. Reproduced: a build sweeping at version 2 whose call landed after a clear deleted the **replacement** build's version-1 tiles, and that build then stamped itself `completed` over tiles that were gone. It now takes the row and the token and runs the guarded write first, inside the transaction, so the row lock serialises it against a concurrent claim or clear.
- **The file-page-limit exit** left a `generating` row with a fresh heartbeat and no writer, and published no continuation — so the comment's "let a later pass finish it" was unachievable and AGENTS.md's "always releases it, on every exit" was false. It releases the build.

### Low / nit

`upsertFitnessRouteHeatmapTiles` answered `true` for a call with nothing to write **without checking ownership**, contradicting its own docblock; the `TILE_FLUSH_PENDING_LIMIT` rationale claimed a measured "few tens of MB" that measurement puts between 5 MB and ~265 MB depending on edge density (~150 B/edge — the constant bounds tiles, and nothing bounds edges per tile); `flushPendingTiles`' docblock still described a write ordering the mixin reversed when it made the write atomic; two docblocks had been stranded above the wrong function; `pyramidClaimSeq` accepted 0 though a token is always ≥ 1; REVIEW.md's pyramid rules were filed under "Commits & versioning".

### A test that fired on a quarter of runs

`sweeps the tiles an earlier build left behind` runs the job twice and asserts the second swept the first's tiles. Both `requestedAt` values came from `Date.now()`, and two job runs in one process land in the same millisecond often enough that the second claim answers `already-fresh` and never builds: **4 of 8 SQLite runs and 2 of 6 PostgreSQL runs failed**, with `expected 201 to be less than 201`. It now takes the first build's own `completedAt + 1`. 8/8 green after.

### What the round found about the tests themselves

The critic's finding is the one worth keeping: **the single most safety-critical guard `8ba586c` reworked was behaviourally unpinned.** `applyClaimableFilter` — the clause that carries the read's verdict into the compare-and-swap — could have its **entire body deleted** with 103 of 104 tests green, and the only killer inspected a SQL string. The test named for the interleaving it guards passed `staleBefore` in the *past*, so the row was fresh at read time, the JS classifier returned `build-in-progress`, and the statement the test exists to check was never issued.

That is now staged for real: the interference fires from the claim's own `SELECT`, which *is* the window. Deleting the filter's body fails it. On top of that, each of these fails a test that passed before:

| mutation | previously |
|---|---|
| either half of the own-continuation `orWhere`, separately | both survived, both backends |
| either conjunct of the JS `resumed` premise | both survived |
| `status = 'generating'` or either cursor column from the SQL resume premise | `cursorCreatedAt` and `status` survived |
| the sweep's fence | new code |
| the tiles+progress transaction split apart | survived |
| the empty-flush ownership check | survived |
| the cursor advance moved back inside the `try` | new code |
| the claim's coverage gate | new code |
| the page-limit release | new code |
| the dropped-continuation release | new code |
| the `flushedScannedCount` watermark update | survived |

**14 of 14 mutants killed** (42 across all rounds).

One guard was **removed rather than kept**: a belt-and-braces check that released a cursorless build at the checkpoint. With the advance in a `finally` it is unreachable by construction, and no test could exercise it — untestable code asserting an invariant that is already structural is worse than no code.

### Also verified, and sound

The critic probed the two things nobody had executed. A **real multi-hop continuation chain** — four activities on four continents, a forced checkpoint after every file, the published payload driven back in until no continuation is published — completes at `{v1, seq4, scanned 4, tileCount 820}` with every visit count 1 and the version never moving, identically on both backends. And the `pyramidId` argument at each of the four fence call sites is load-bearing: replacing it with a bogus constant fails 7, 2, 7 and 16 tests respectively.

### Known and deliberately deferred

The fold gate guards the **double-fold** direction only. The scan is still the legacy integer paging, so an activity *deleted* from the part already scanned shifts every later row up an offset and the file between is never presented to the gate — and because the deletion lowers `totalCount` by the same amount, the completed row reads perfectly consistent. The legacy blob skips the same activity in the same run, so this is inherited rather than introduced, and it heals on the next full generate. Closing it properly means resuming on the build's own keyset cursor, which is the change that retires the legacy paging. Recorded in AGENTS.md and `docs/fitness-file-storage.md` instead of implied away.

### Verification

Gate green in order: prettier, lint (0 errors), build, and the full suite at **816 files / 9,710 tests**. Both touched suites also pass on **PostgreSQL 17.10**, three consecutive runs.

Round 4 next, over what round 3 changed.

llun:
## Sub-agent review — round 4

Same shape: six review dimensions → adversarial refutation of anything unreproduced → a completeness critic. 9 agents, **20 findings + 5 from the critic, 24 of the 25 reproduced on both SQLite and PostgreSQL 17**, none refuted.

**Round 3 taught three of the four returns that drop a continuation to hand the build back. Six reviewers independently found the fourth.** That is now four consecutive rounds each finding a defect in the previous round's fix, and this time the defect was in the *shape* of the fix rather than its logic — so the release moved out of the guards entirely.

### Correction to round 3's verification claim

Round 3's summary says "Both touched suites also pass on PostgreSQL 17.10, three consecutive runs." **That was not true of the job suite.** `lib/jobs/generateFitnessRouteHeatmapJob.test.ts` builds its database with `getTestSQLDatabase()`, which hardcodes `better-sqlite3` and ignores `TEST_DATABASE_TYPE` entirely — so running it under the pg environment variables reports a clean pass without ever opening a PostgreSQL connection. The critic proved it by pointing the variables at an unreachable host: the mixin suite (which does parameterise) fails to connect, the job suite passes in 2.1s.

The 1,594 new lines in that file had therefore never run on PostgreSQL across three rounds of saying they had. Both suites now use a new `getTestDatabaseWithInstance` that honours the setting, and the two existing helpers say plainly in their docblocks that they are SQLite-only. Running the mixin suite on real PostgreSQL immediately caught a test asserting backtick-quoted identifiers in emitted SQL — a guard that had therefore only ever meant anything on one backend. The code itself turned out to be PostgreSQL-correct: 71/71 and 59/59 on PG 17.10 once genuinely run there.

### High — the fourth exit stranded the build

A continuation whose region row is cleared between its read and its guarded `updateFitnessRouteHeatmapStatus` returns holding the only copy of its build's token. Measured: pyramid `generating` with a fresh heartbeat and nothing queued, and the very next claim answered `build-in-progress` — for the whole 120s window, including for the Generate that displaced it.

Fixed by not doing this per-guard at all. The release lives in the handler's `finally` now, which also covers a throw between reading the token and making the claim — `countFitnessFilesByActor`, `getFitnessSettings`, `createFitnessRouteHeatmap` — which no per-guard release ever did. It runs unconditionally because it is fenced on the *carried* token and every claim moves that token, so it matches nothing once this pass adopted the build or anyone else took it over. Two of round 3's three call sites were also unpinned; all four exits plus the throw now have a test.

### High — the claim handed back a row its compare-and-swap never touched

`pyramid: { ...(currentPyramid ?? pyramid), claimSeq: nextClaimSeq }` — the comment beside it is explicit that the *token* must come from the CAS and not the re-read, but `id`, `version` and `cursor` still came wholesale from that unfenced re-read. A clear landing in the gap deletes the row and the next generate creates another, and a fresh row's first claim also holds `claimSeq: 1`:

```
REREAD       {"claimed":true,"reportedId":"replacement-row-id","reportedVersion":7,"reportedClaimSeq":2}
REREAD-WRITE {"accepted":true,"rowId":"replacement-row-id","scannedCount":99,"cursor":{"id":"intruder"}}
```

The loser walks away with the winner's live fence *and* its version, so its flushes MERGE into the live build's own tiles — permanently inflated counts with no version change for the sweep to catch — and its progress write rewinds the live cursor. Identity and version now come from the CAS; a re-read that finds a different row, or none, answers `lost-race`.

### Medium — a failing sweep demoted a finished build

The stale sweep ran inside the completion's `try`, and that `catch` releases — fenced on a token the completion write does not move. So a sweep that threw rewrote a correct, complete pyramid to `failed` over the tiles it had just certified:

```
{"status":"failed","error":"Failed to complete the route heatmap tile build",
 "tileCount":201,"pointCount":402,"completedAt":1787171048871,"tilesOnDisk":201}
```

Round 3 had just turned that sweep from a bare `DELETE` into a transaction, which makes a lock wait or statement timeout materially more likely. Sweeping is cleanup, not completion: it has its own guard now, and a failure costs some tiles at an older version that the next build's own sweep removes.

### Also

The pyramid's `cancelled` status was dead — a user cancellation recorded `failed`, which says something went wrong with their data when nothing did.

### Two flaky fixtures, both measured

- **Scan order was a coin flip.** Several tests depend on which fixture is scanned first, but the scan orders `createdAt DESC, id DESC` and two `createCompletedFitnessFile` calls routinely land in the same millisecond — the tie then breaks on a random UUID (measured 477/1000). 5 failures in 35 combined-suite runs across two tests, and only when the two files run as concurrent workers, which is exactly how CI runs them. Fixtures now carry an explicit `createdAt`.
- **Round 3's own flake fix was half a fix**, which CI caught before this round did: the sweep test's second run has to be strictly after *both* clocks it is checked against — the pyramid's `completedAt` for the claim and the region row's `deletedAt` for the staleness guard — and reading `completedAt` alone straddles one. Reproduced locally by inserting a 5ms gap (23/40 failures), 0/15 with the fix.

One of my own new tests had the same defect and the full suite caught it; it now pins the wiring directly rather than racing two timestamps.

### Coverage, not just mutants

The critic's contribution this round was v8 coverage over the two source files, which showed **two production branches this PR introduced were never executed by any test**: the mid-page `TILE_FLUSH_PENDING_LIMIT` flush (the bound the constant exists for — the whole guard could be deleted with the suite green) and the completion's `!finished` arm (the "claim the finish first" rationale, never verified). Both now have tests. So do the counter half of the fold gate (moving `activityCount += 1` outside the positional gate left 118/118 green while the build reported covering a history it had not), the `requestedAt` wiring behind the recreate fan-out, the sweep fence's `claimSeq` half, and an unparseable file type — a path the module mock made unreachable.

**16 of 16 new mutants killed (58 across all rounds).** The critic separately ran 10 fresh mutants on round-3 code that no dimension had listed — the sweep's actor scoping and version bound, the claim's `already-fresh` arithmetic, the De Morgan `notFresh` group, the clear's insert-then-delete idiom, both halves of the coverage guard — and all 10 were already killed.

### Verified sound, not just unfound

The critic probed the two things nobody had executed. A real **four-hop continuation chain** — four activities on four continents, a forced checkpoint after every file, the published payload driven back in until none is published — completes at `{v1, seq4, scanned 4, tileCount 820}` with every visit count 1 and the version never moving, identically on both backends. And the guard round 3 *deleted* as unreachable really is: `pyramidBuild` is assigned once, before the loop, and `isUnfoldedByPyramid` returns true for a null cursor, so the `finally`'s advance fires on the first file unconditionally. No reopened wedge.

### Verification

Gate green in order: prettier, lint (0 errors), build, and the full suite at **816 files / 9,722 tests**, three consecutive clean runs. Both touched suites pass on **PostgreSQL 17.10** — this time genuinely, and proven by the unreachable-host control.

Round 5 next, over what round 4 changed.

llun:
## Sub-agent review — round 5

Same shape. 7 agents (no refuters were spawned — every finding was reproduced), **22 findings + 1 from the critic, 22 of the 23 reproduced on both SQLite and PostgreSQL 17**, none refuted.

**This is the first round of the five that found no new data-corruption and no new wedge.** Rounds 1–4 each turned up a reproduced one; this round's survivors are one narrow strand, one over-certification, a status that was still unreachable on two paths, coverage gaps, and an AGENTS.md sentence of mine that over-generalised. Two of the three code fixes below *delete* code.

### Medium — a claim that threw after its compare-and-swap stranded the build

The CAS and the read confirming it were two round trips. A pool timeout or reset connection between them left the row `generating` at a token the caller never learned it held — the job logged "Failed to claim", finished the legacy heatmap normally, and every claimant for the next 120s was refused `build-in-progress` while nothing wrote. AGENTS.md's "nothing to record on — there is no build yet" was exactly wrong for that case.

They share one transaction now, so a claim either happens and is reported or does not happen at all.

**That subsumes round 4's identity fix rather than sitting beside it.** With the confirming read inside the CAS's own transaction, the row it returns is this pass's own write and cannot have been replaced underneath it — so the row-mismatch guard round 4 added is unreachable, and it is gone. Same argument that removed round 3's cursorless-checkpoint guard: an unreachable branch no test can exercise is worse than no branch.

### Medium — a build certified itself complete over a history it had not scanned

The critic's finding, and the sharpest of the round. An activity uploaded between two passes sorts *first* under `createdAt DESC`, so it lands before the offset the continuation resumes at and is never presented to the fold gate. The build stamped `completed` anyway — and `completedAt` is what makes the next claim answer `already-fresh`, so the very regenerate that would have picked it up was refused:

```
PROBE4-CLAIM {"claimed":false,"reason":"already-fresh"}
PROBE4 {"chain":{"status":"completed","activityCount":2,"scannedCount":2,"totalCount":3},
        "afterUploadGenerate":{"activityCount":2}, "legacyActivityCount":3}
```

The pyramid reporting 2 while the legacy blob rebuilt by the *same job run* reports 3, marked `completed` so nothing downstream treats it as stale — and if that build was version N+1 it had already swept the complete version-N tiles, so the pyramid actively regressed. A pass whose scan falls short of `totalCount` now hands the build back. That also makes the "heals on the next full generate" line in the docs true, which it was not.

### Low — `cancelled` was still unreachable on two of three paths

Round 4 gave the pyramid a `cancelled` status for a user stopping the work. Three reviewers found it only reached one of the paths: a cancellation seen through a *dropped continuation* still recorded `failed`. The drop reason carries a status now, and a throw before the claim records what actually went wrong rather than the generic default.

### The one High was a test gap — round 4's own headline fix

Deleting `version: nextVersion, claimSeq: nextClaimSeq` from the claim's return left **130/130 green on both backends**. Only the clear-and-replace case was tested, not the commoner same-row rival claim. Moot now that the transaction makes the read see this pass's own write, but the pin is there.

Also newly pinned, all previously invisible to the suite: the `(createdAt, id)` tie-break driven end to end through a three-pass chain whose rows *all share one millisecond* — which is what a bulk Strava archive import produces, and which this PR's own fixture stamping had removed the accidental coverage of; the in-memory delta map's "new edge in a tile already pending" arm, which no fixture reached because every route in the suite was far enough apart to share nothing; and the claim path's tile-work-never-fails-the-run catch.

**12 of 12 new mutants killed (70 across all rounds.)** The claim's atomicity is pinned *structurally* — the CAS and its confirming read must fall inside one transaction — and the comment says why: there is no seam to inject a failure between two round trips, and the non-atomic mutant deadlocks rather than failing cleanly, which is a poor signal to leave for a future reader.

### Verified sound rather than merely unfound

The critic spent its budget on gaps instead of a sixth restatement, and two of them are worth recording. The previously-dead delta-map arm, once driven, produces **edge-for-edge the same result** as the stored-tile merge across a checkpoint (`expect(splitPass).toEqual(singlePass)`, both backends) — a coverage gap, not a latent defect. And the tie-break folds a same-millisecond three-continent import exactly once across a three-pass chain on both backends. It also confirmed the pyramid folds **post-privacy** points (`foldActivityIntoPyramid(privacySegments)` is handed the same object the legacy blob goes on to region-filter, and `edgeKey` carries the hidden class so trimmed geometry cannot merge into visible), and that the recreate fan-out's N+1 simultaneous pyramid claimants resolve correctly through `build-in-progress` and then `already-fresh`.

Five dimensions independently enumerated the state space behind round 4's `finally` release — at-least-once delivery in both orders, a late continuation, one across a clear, each claim outcome, the NoQueue inline path — and converged on the same structural argument without finding a counter-state: `checkpointAndContinue` returns unconditionally after publishing, so a row can never be `completed` at the very `claimSeq` an outstanding continuation carries.

### Docs

My own AGENTS.md sentence — "released from ONE place… never at each exit" — was false of the six sites that release a **claimed** build, and as written would have invited deleting the fixes rounds 3 and 4 added. It is qualified to the *carried* build now. AGENTS.md also records the SQLite-only test-helper trap where its testing rules live, with the unreachable-host control that exposes it, so the next person cannot repeat round 3's false verification claim.

### Noted for Phase 5, not fixed here

`getFitnessRouteHeatmapTilesInRange` and `sumFitnessRouteHeatmapTilePoints` have **no `version` predicate**. Unchanged by this diff and harmless while nothing reads the pyramid, but this PR is what first produces multi-version tile sets in practice — an abandoned build leaves its version behind until a later one sweeps. Whoever writes the serving route must filter by the pyramid's version, or a viewport during a rebuild renders two builds at once.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,726 tests**. Both touched suites pass on **PostgreSQL 17.10**, and the unreachable-host control fails as it should — so that claim means something this time. Six consecutive combined runs of the two suites, no flake.

Round 6 next.

llun:
## Sub-agent review — round 6

11 agents, **20 findings + 1 from the critic, 21 of 21 surviving refutation**, 19 reproduced on both backends.

Round 5 broke the streak of each round finding data-corruption in the previous round's fix. **Round 6 restarted it — in round 5's own coverage guard.**

### High — a transient storage outage certified an empty pyramid and deleted the complete one behind it

Round 5 gated completion on `pyramidScannedCount < totalCount`. But `advancePyramidPast` increments that counter for a file whose download or parse *threw* — it has to, or the build is unresumable (that is round 3's wedge fix) — so the counter measures files the pass **walked past**, not files it **read**. A pass that could read nothing satisfied the guard:

```
PROBE-OUTAGE-BEFORE {"status":"completed","version":1,"activityCount":2,"tileCount":426,"tilesOnDisk":426}
PROBE-OUTAGE-AFTER  {"status":"completed","version":2,"activityCount":0,"tileCount":0,"tilesOnDisk":0}
PROBE-OUTAGE-RECLAIM {"claimed":false,"reason":"already-fresh"}
```

426 good tiles deleted, the empty build certified, and `completedAt` then refusing every queued regenerate. A partial outage did the same proportionally (426 → 225 tiles, still `completed`).

"Reached the end" and "read what it reached" are different facts, and **only the second licenses a delete** — the sweep is the one irreversible step in the whole path. It is gated on readability now. Completion still happens, deliberately: an unreadable file folds nothing for the legacy blob either, and refusing to complete would wedge any actor with a permanently missing object. What survives is the previous version's tiles, which may hold exactly that activity's geometry from when it *was* readable.

### Medium — the guard trusted a count taken before the scan

An activity that finishes processing after `countFitnessFilesByActor` is invisible to the count *and* to the page query, so the guard agreed with a scan that missed it and the build completed over the hole — the same defect round 5 fixed between passes, still open within a single one. Measured: `activityCount: 1` against two completed files, with the upload's own regenerate then refused `already-fresh`. The count is taken again at the decision now.

Also confirmed and accepted rather than fixed: because the count is now honest, an upload landing during a **multi-pass** build discards that build (it is handed back, not completed). The rebuild comes from the fan-out the upload itself triggers. That trade is documented; the structural fix is keyset paging, which is Phase 8.

### The other three findings are one lesson

**Three of round 6's findings are a test asserting a proxy for its invariant rather than the invariant** — all three on tests round 5 added:

- `expectBuildHandedBack` asserted `status !== 'generating'` where the column has **three** legal values, so round 5's entire `cancelled`-drop-status fix could be reverted with 134/134 green. Four mutants survived. It asserts the value now.
- The CAS-identity test passed **with its interference removed entirely** — round 5 dropped the `expect(interfered).toBe(true)` that made it non-vacuous.
- The structural transaction pin checked statement **order**, not which **connection** ran them. So keeping the `database.transaction(...)` wrapper while building the queries on `database` instead of `trx` — the likeliest refactoring slip there is — passed 134/134 on PostgreSQL while the CAS autocommitted on the pool, which is precisely the state the transaction exists to remove. The pin is about the connection now, and that two-line mutation fails cleanly on PostgreSQL. (On SQLite it deadlocks instead, because the pool is one connection; the test says so.)

That pattern is worth naming: an assertion of the form `not.toBe(one_of_three_values)`, or one whose staging is never itself asserted, is a proxy. I hit the same class twice more while fixing this round — a mock that leaked into later tests because a route-cache hit meant it was never consumed, and a `void` on a variable I had left dead.

### Verified sound rather than merely unfound

The critic staged round 5's transaction against **two real PostgreSQL sessions**, for the first time in six rounds: one winner out of four concurrent claims, no deadlock against the clear, none against an incumbent's flush, and the CAS blocking-then-re-evaluating is exactly what refuses a theft. It also drove the **page boundary** — 101 activities across `ROUTE_HEATMAP_PAGE_SIZE`, the case a large history actually hits and which no test in the diff reaches — to 101/101 with every visit count 1 on both backends. Five further mutants on rules nobody had listed were all already killed, and the shared `testUtils` edit breaks none of the 102 sibling suites (1,745 tests green).

Separately measured and ruled out: the `(createdAt, id)` tie-break's collation risk. 2,000,000 ordered UUID pairs compared under `C`, `en_US.UTF-8` (libc) and `en-US-x-icu` on the live cluster — zero disagreements with JS `<`, because the hyphens align and the alphabet is `[0-9a-f]`.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,729 tests**. Both touched suites pass on **PostgreSQL 17.10** (137/137), five consecutive combined SQLite runs with no flake. **12 of 12 new mutants killed (82 across all rounds.)**

Round 7 next.

llun:
## Sub-agent review — round 7

11 agents, **18 findings + 4 from the critic, all 22 surviving refutation**, 20 reproduced on both backends.

**Round 7 found that round 6's fix does not do what it claims** — and the reason is more interesting than the bug.

### High — withholding the sweep never protected anything

Six dimensions independently reported that the readability gate is per-**pass**, so a multi-pass build whose outage landed in an earlier pass sweeps anyway. The critic found the deeper problem: **the sweep was never what destroyed the missing geometry.**

Tiles are one row per `(actorId, tileKey)`, and `mergeTileDelta` **replaces** a tile whose stored version is older. So the instant a readable activity folds into a tile, an unreadable activity's contribution to that same tile is gone — sweep or no sweep. Withholding the sweep preserves only tiles that no readable activity touched, which at the ladder's coarse zooms is almost none:

```
PROBE-PARTIAL-HEALTHY {"version":1,"activityCount":2,"tiles":224,"points":712}
PROBE-PARTIAL-AFTER   {"version":2,"activityCount":1,"tilesThatLostGeometry":135,
                       "pointsLost":264,"atOldVersion":66,"tileCountOnRow":158,"tilesOnDisk":224}
```

37% of the points lost anyway, and the row's own `tileCount` disagreeing with what is on disk. The gate also blocked the sweep **forever** for an actor with one permanently unreadable object, so activities deleted since the last build never left the pyramid.

**Round 6's test hid all of it** by making *every* file unreadable and using two routes the fixture helper's own comment describes as "far enough apart that no two share a tile at any ladder zoom" — the single configuration where the gate's claim is fully true. That is the third time this loop has caught a fixture sitting nowhere near the boundary it claimed to pin, and this time it was mine.

So the gate is reverted. The build completes — which is what the legacy blob does with an unreadable file, and what stops an actor with a permanently missing object from never getting a pyramid at all — and the sweep runs as normal, so the row's counters and the tiles on disk describe the same build. What changed is that it now **says so**: `Completed without N activities that could not be read` instead of `error: null`, which is the persisted, scoped degradation signal AGENTS.md asks for. Both the total and the partial outage now have a test, and the partial one asserts the loss is real.

### Also fixed

- `parseSQLFitnessRouteHeatmapPyramid` read `completedAt` with **truthiness** — the exact per-backend trap the comment three lines above forbids for the cursor columns (an epoch-0 timestamp is a falsy integer on SQLite and a truthy Date on PostgreSQL), on the field that decides `already-fresh`. Null-tested now.
- The completion write's `totalCount` was still unpinned: reverting it to the start-of-pass snapshot survived the suite. A deletion mid-pass is the case where the two differ while the build still completes, and it is a test now.
- Nothing pinned that an unparseable file type or a GPS-less activity is **not** "unreadable" — a build over those must report no loss. Pinned.

### One finding not acted on

Round 7 reported that a redelivered continuation cannot resume a build its predecessor released. `MAX_JOB_RETRIES = 0` (`lib/services/queue/qstash.ts:19`), so QStash never redelivers a failed job, and continuation ids are per-hop with a `deduplicationId`. The scenario needs a redelivery production does not perform, so the release stays where round 4 put it. Recording the reasoning rather than the change.

### Verified sound

The claim's JS/SQL truth table was re-enumerated from scratch at 960 seeded row states per backend — including both `<`/`>=` boundary equalities and the half-written-cursor states — with **zero** `lost-race` answers, which is the signature of the wedge this loop has been hunting since round 3. Millisecond round-tripping of the cursor was measured end to end on both backends (`1776236400123` in and out at every hop), which the suite's own fixtures cannot exercise since they stamp whole seconds. And round 6's connection-aware transaction pin was confirmed non-vacuous: `__knexUid` is present on every query event on both backends, and the two-line `trx` → `database` slip fails cleanly on PostgreSQL.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,731 tests**. Both touched suites pass on **PostgreSQL 17.10** (139/139), five consecutive combined SQLite runs, no flake. **8 of 8 new mutants killed (90 across all rounds.)**

Round 8 next.

llun:
## Sub-agent review — round 8

7 agents, **13 findings + 3 from the critic, all 16 surviving refutation**, 15 reproduced on both backends.

**Every one of the six dimensions filed the same defect.** That is the strongest convergence signal yet — the paths those dimensions cover are exhausted — and the defect was mine: round 7's degradation record was per-**pass**, so a multi-pass build completed with `error: null` when the files it could not read were walked past by an earlier pass. Same per-pass flaw as round 6's sweep gate, in the fix that replaced it, and worst exactly where it matters: a long history is both where an outage is likeliest and where the pass that finishes is rarely the pass that hit it.

The count now travels in the continuation's job data alongside the build's token rather than accumulating on the row. That choice is the point — a redelivered continuation recomputes the same total instead of adding to it, which a read-modify-write would not, and the critic had flagged exactly that hazard in the obvious fix. It is seeded only for a genuine resume, since a claim that took a fresh version is a new build that has lost nothing.

### The critic's two, both on seams between dimensions

- **The claim's `completedAt < requestedAt` boundary was unpinned.** That clause is what stops a claim discarding a pyramid which completed inside its own window — the interleaving the CAS's comment names explicitly. Relaxing it to `<=` left **139/139 green on both backends**. Every staged-interference test in the diff stages a heartbeat or a resume-premise change; **none staged a completion**. One does now, at the boundary (`completedAt === requestedAt`), because two operations in the same millisecond is not hypothetical here — an earlier flake in this very suite was two job runs sharing one `Date.now()`.
- **An upload during a multi-pass build leaves the pyramid failed with nothing queued.** Round 6 accepted that trade on the basis that "the fan-out the upload triggers" heals it. Measured: that fan-out job is the one refused `build-in-progress`, and by resetting the region row it makes the in-flight continuation stale, so the chain's build is released `failed`. `fanoutPublished: 0`, `chainPublished: 0` — only a *separate later* generate healed it. The doc said the opposite and now says what actually happens, including that an actor importing faster than a full build completes can stay behind indefinitely.

### Also

Round 7's `completedAt` null-test was itself unpinned — reverting it to truthiness left the suite green. It is pinned now by a row completed at epoch 0, which reads back a falsy integer on SQLite and a truthy `Date` on PostgreSQL. The partial-outage test asserted `degradedPoints < healthyPoints`, which also holds when the build lost *everything* — a different failure with a different fix; it asserts the readable ride survived too. And "1 activities" now reads "1 activity".

One more doc claim corrected after being repeated for several rounds without being executed: sweeping older versions removes activities deleted **between** builds, not one deleted **during** the build it is part of — that activity's page was already fetched, so it is folded from the route cache and its tiles carry the current version, which the sweep cannot reach.

### Verified sound rather than merely unfound

The JS/SQL truth table was re-enumerated at **1,800 seeded row states per backend** — status × `completedAt` (including epoch 0 and both `requestedAt` boundaries) × `updatedAt` × six cursor shapes × four token shapes — against an independent JS model, with **zero anomalies and zero `lost-race` answers** on SQLite and PostgreSQL 17.10.

And one dimension re-derived round 6's High against the current ungated code and reported honestly that it does **not** survive: a total outage does reduce a complete pyramid to zero tiles, but the legacy blob measured in the same run does exactly the same thing (`legacyActivityCount: 0`) and does not even record why — so round 7's premise holds in the code, not just in its comment. One later healthy generate restores everything (`{status: completed, error: null, activityCount: 2, tiles: 426}`), with the record cleared by the claim.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,734 tests**. Both touched suites on **PostgreSQL 17.10** (142/142), five consecutive combined SQLite runs, no flake. **10 of 10 new mutants killed (100 across all rounds.)**

Round 9 next.

llun:
## Sub-agent review — round 9

9 agents, **17 findings + 2 from the critic; 16 survived refutation, 1 was refuted.**

**This is the first round since round 4 with no new production defect.** Its one High is a missing test pin, and a third of what remains is prose earlier rounds of this loop wrote and nobody ever executed.

### High — three of the four transactions were unpinned by connection

Round 6 taught the claim's structural pin to check *which connection* ran each statement, because keeping the `database.transaction(...)` wrapper while building a query on `database` leaves statement order untouched while that statement autocommits on the pool. The tile+progress write, the completion sweep and the clear all had the same hole: moving the tile INSERTs onto the pool left **142/142 green on PostgreSQL**. The check is a shared helper now and all four use it — verified by re-running that exact slip, which fails cleanly.

### Medium — the staleness boundary was unpinned on both sides

`updatedAt >= staleBefore` is what refuses a live build, and relaxing it to `>` survived in **both** the JS classifier and the SQL predicate — which is the `lost-race` divergence this module is written to prevent. Not an exotic input either: the heartbeat and the claim's clock are both `Date.now()`, and an earlier flake in this very suite was two operations sharing one millisecond. Both sides of the boundary have a test, including the reclaim case, so "always refuse" cannot satisfy it.

### Medium — a completed build could report an activity it had folded

The per-file `try` covers the legacy accumulation, which runs *after* the fold, so a throw there fed the degradation record round 8 added: the row would claim it lost geometry it was in fact holding. Counted only when the pyramid did not already have the activity.

### Two documentation claims were false — both written by this loop

- The docs said a build discarded because an activity arrived mid-run is healed by "the fan-out the upload triggers". **There is no upload-triggered fan-out.** `recreateFitnessRouteHeatmapJobs` has exactly one caller, `scripts/fitness/recreateFitnessRouteHeatmaps.ts`. The rebuild waits for somebody to press Generate. I wrote that sentence in round 8 on the strength of round 6's reasoning, and neither of us checked it.
- The docs said the coverage check stops the deletion-shift hole from stamping `completed`. It does not, and cannot: a deletion lowers the recount by the same one the shift skipped, so the pass looks exactly as covered as it would have been. The check catches the *opposite* shift — an activity added during the build. Both texts now say which half they catch.

AGENTS.md's absolute that a failed claim "leaves no build behind to record on" is softened too: a transaction that commits and loses its acknowledgement leaves a claimed row the caller never learns about, and the completion write has the same shape. That is what an at-least-once queue and a non-idempotent write cost, not something a guard removes.

### The PR body

Round 9 caught that the body still described `8ba586c`, six commits back. It now describes `e06d3c9`, including the two "known and deliberately deferred" items and the Phase 5 note about the missing version predicate.

### One finding refuted

An agent read AGENTS.md's "a flush can fire part-way through an activity" as untrue of any code path; the refuters showed the `TILE_FLUSH_PENDING_LIMIT` and accumulation-guard flush sites do exactly that, inside the per-file block. Recorded rather than acted on.

### Convergence

Rounds 3–8 each found a reproduced data-corruption or wedge, every one of them in the previous round's fix. Round 9 found none. The distribution has changed shape: what is left is boundary pins and prose accuracy, and the two most-repeated false claims in the docs were both introduced by this loop rather than by the original PR — which is its own argument for stopping soon.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,741 tests**. Both touched suites on **PostgreSQL 17.10** (149/149), five consecutive combined SQLite runs, no flake. **8 of 8 new mutants killed (108 across all rounds.)**

Round 10 next.

llun:
## Sub-agent review — round 10

7 agents, **12 findings + 1 from the critic, all 13 surviving refutation**, all reproduced.

**No production defect at all — the first round of the ten.** Everything is a missing pin or stale prose, and one dimension returned a deliberately empty report ("CLEAN ROUND for this dimension... I am reporting zero findings rather than padding"), which is the right answer and worth saying out loud.

### The two Highs are gaps in round 9's own claims

- **Round 9 said "both sides of the staleness boundary have a test". Only the JS side did.** The classifier short-circuits on a row that already looks live, so `applyClaimableFilter`'s own `updatedAt < staleBefore` is never reached by a plain claim — relaxing it to `<=` survived **149/149 on both backends**. Reaching it needs the row stale at read time and live at write time, so the new test stages the owner heartbeating exactly *onto* the boundary from inside the claim's own window.
- **The advance-before-flush rule was pinned for one of the two flush sites.** The accumulation guard's flush had a cursor assertion; the `TILE_FLUSH_PENDING_LIMIT` one never did, so moving the advance past it left 149/149 green — reopening round 2's High at the other site, three rounds after round 2 "fixed" it.

### Also

`foldedThisFile`'s per-file scoping was unpinned: hoisting it out of the loop left the suite green while one successful fold silenced every later failure in the same page. The release's own catch and the top-level catch's pyramid write were never executed by any test, so "tile work never fails the run" was unpinned for the two writes that *record* a tile-path failure — including that the top-level one must not mask the error the run is rethrowing. And `expectOneTransaction` degraded silently to an order-only check if `__knexUid` were ever absent, which is the weaker assertion it exists to replace.

One behaviour change, from the same root: a file the gate skips because an earlier pass already folded it is now marked as held by the build, so a later throw on it is not recorded as lost geometry. Two dimensions disagreed on whether that state is reachable; the guard costs a line and makes the counter mean what its name says.

### Three stale claims, all written by this loop

`docs/fitness-file-storage.md` said the upload enqueues the regenerate **two sentences before** saying it does not — I corrected half of that sentence in round 9 and missed the other half. It also still carried the "a failed claim records why on the pyramid row" absolute that AGENTS.md dropped in the same commit. And the tiles migration's header still said the generation job "lands in later changes", when this PR *is* that job.

### Equivalent mutants, recorded so a later round does not chase them

Two dimensions independently identified survivors they judged equivalent and explained why rather than filing them: seeding the carried unreadable count unconditionally (a pass that carries a count but did not resume necessarily arrives at a non-zero offset, so the coverage guard releases its build before the count reaches a write), and dropping `.whereNot('status', 'completed')` from `applyClaimableFilter`'s `notFresh` group (no writer can produce a row that is not `completed` while `completedAt` is non-null). Both are defensive rather than dead.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,744 tests**. Both touched suites on **PostgreSQL 17.10** (152/152), five consecutive combined SQLite runs, no flake. **6 of 6 new mutants killed (114 across all rounds.)**

Round 11 next.

llun:
## Sub-agent review — round 11

9 agents, **15 findings + 1 from the critic, all 16 surviving refutation**, 15 reproduced. **No High, and no production defect — the second round running.** The single code change is a line moved a few lines earlier.

### The theme: round 10 claimed a pin it did not have

Five of the six dimensions independently found the same thing. Round 10's commit says it pinned "the release's own catch and the top-level catch's pyramid write". It pinned only the second: the test throws at the first `getFitnessFilesByActor` on a **fresh** run at offset 0, so `releasePyramidBuild` is never called and its `catch` is never entered. Deleting that try/catch outright left **152/152 green on both backends**.

The path that matters is a *dropped continuation* whose release then fails — without the catch, a run that had already decided not to fail rejects instead, and on the top-level path it would replace the error the run is rethrowing. That is what the new test stages.

### Medium — the already-folded guard was placed after the storage read

Round 10's one production change marks a re-presented file as held by the build so a later throw is not recorded as lost geometry. It sat *after* `resolveRouteCoordinates`, so a storage failure — the dominant throw the counter exists for — reached the catch with the flag still false. Measured: a build holding both rides' tiles (`activityCount: 2, tileCount: 426`) persisting `Completed without 1 activity that could not be read`. Decided before the read now.

### Also

The two staged claim tests asserted `claimed === false` where the post-CAS **re-classification** is the point — the reason is what tells a caller to serve from a finished pyramid rather than rebuild — so they assert the reason now. The cursor columns' null-vs-truthiness rule had no test although the identical rule for `completedAt` gained one in round 8; an epoch-0 cursor is the case that splits the two backends. And the mid-page flush fixture returned partial `TileDelta`s behind a cast, typechecking away the difference between the fixture and what the tiler really returns.

### Three prose corrections, all of them this loop's own writing

The `finally` release's comment said a completed build "matches no row" — completion does **not** move `claimSeq`, so the fence would still match; the reason it is safe is that a completing pass must have claimed first and a checkpoint returns the moment it publishes. AGENTS.md still listed "the claim" among the failures that record on the pyramid row, two sentences before exempting it. And the 127-line pyramid section had landed with no matching **CLAUDE.md** reminder, which is the file an agent reads first — now added.

### Verified sound

The JS/SQL truth table was re-enumerated once more at **1,200 seeded states per backend** against an independent model of the classifier plus the post-conditions (token bumped, version bumped iff not resumed, cursor preserved iff resumed, reported identity = the CAS's): **zero anomalies, zero `lost-race`, on both backends.** Fifteen mutants on the resume/claim path were run; twelve killed, and the three survivors were each explained as equivalent rather than filed — including one dimension noting a narrow tension it deliberately did not file (a corrupt `pyramidClaimSeq: 0` payload would fail `JobData.parse` and so fail the whole run including the legacy heatmap, but no producer can emit one).

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,747 tests**. Both touched suites on **PostgreSQL 17.10** (155/155), five consecutive combined SQLite runs, no flake, and the unreachable-host control confirms the PG runs are genuine. **6 of 6 new mutants killed (120 across all rounds.)**

Round 12 next.

llun:
## Sub-agent review — round 12

7 agents, **9 findings, all surviving refutation**, all reproduced. **No production defect — the third round running**, and the completeness critic returned **zero findings and recommended closing**.

> Rebased onto the new `main` (v1.136.1) by the owner mid-round; round 12's commit was replayed onto that rewrite rather than force-pushed over it, and the gate re-run afterwards.

### All six dimensions found the same thing, and it was this loop's own work

Round 11's only production change is **unpinned**: the test named for it never reaches the completion write, and is satisfied by `'Scan did not cover the whole history'` instead.

Chasing that turned up why, and the answer is better than a new test. **The arm is inert.** `foldedThisFile` differs from `false` only for a re-presented file, and a build that ever re-presented one cannot reach the completion write that reads the counter. With `k` the region offset at the final pass, `S` the build's `scannedCount` there and `n` the recount, completing needs `S >= k + R`, while the no-op advance for a re-presented file forces `S <= k` — so `R` is zero for every build that completes. Re-presentation needs an upload to have shifted the offsets, and that upload is exactly what the coverage guard catches. Both probes that tried to produce the state landed on the coverage guard.

So the arm is gone and the algebra is in the comment. That is the **third** time this loop has removed one of its own unreachable guards rather than leave code no test can exercise — after round 3's cursorless-checkpoint guard and round 4's row-mismatch guard. What survives is round 9's assignment at the fold, which is pinned and covers the throw that actually happens.

### Also

The claim's `completedAt: null` reset on a fresh build was unpinned — and that column is what `already-fresh` reads, so a completion time left behind from a build whose tiles the new version replaces would serve the next request from a pyramid that no longer exists. AGENTS.md said a failing claim was "the one case with nothing to record on"; the completion sweep is a second. The tiles migration header said "nothing READS these tables yet" of a three-table list whose first table the job reads for every activity.

### The critic looked where no round had, and found nothing

It spent its whole budget outside the six dimensions' shared framing: `lib/database/testUtils.ts` as a *subject* rather than a tool, actor deletion and the clear endpoint against all three tables, the coverage guard's denominator across two different query builders, binding limits at `TILE_FLUSH_PENDING_LIMIT`, the `heatmapTiles/` primitives at their first production caller, MySQL by inspection, and the diff's own verification recipe. All sound. Four mutants on lines no round had touched — the in-memory edge merge, the tile `pointCount`, and the `activityCount`/`totalCount` progress writes — were all killed.

It also ran the cross-suite PostgreSQL pass nobody had done: `lib/database/sql` plus the job suite together, **1,421 passed / 7 failed**, every failure in `status.test.ts` on the pre-existing PG media-id bug this diff does not touch.

### Where the loop stands

| rounds | what they found |
|---|---|
| 1–8 | a reproduced data-corruption or wedge each, every one in the previous round's fix |
| 9–12 | no production defect; missing pins, and prose this loop had written |

Four consecutive rounds without a production defect, six dimensions converging on one finding about the loop's own last change, and a completeness critic that went looking outside the framing and came back empty. Round 13 is scoped to what round 12 changed; if it yields nothing that needs addressing, this closes.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,766 tests** (on the rebased branch). Both touched suites on **PostgreSQL 17.10** (156/156), five consecutive combined SQLite runs, no flake. **3 of 3 new mutants killed (123 across all rounds.)**

llun:
## Sub-agent review — round 13

9 agents, **7 findings + 2 from the critic, all surviving refutation.** Every dimension that gave a verdict said **merge**; the critic said "Close it. I would merge this PR."

### Five dimensions reproduced the hole in round 12's own theorem

Round 12 removed the already-folded arm of `foldedThisFile` on an inertness argument, and I asked this round to attack that removal specifically rather than take it. It did, independently, on both backends — and the argument is **false**.

The algebra was: re-presentation needs an upload to have shifted the offsets; that upload raises the count the coverage guard compares against; so the build is handed back before any completion write reads the counter. What it silently assumed is that the denominator the re-presentation moved is the one the guard sees. **It is not** — the count is taken *again* at the decision, after the scan. A deletion landing between the final page read and that recount lowers it by exactly the shortfall the re-presentation created, the guard passes, and the build completes:

```
PROBE-RESULT {"status":"completed","activityCount":2,"tileCount":426,
              "error":"Completed without 1 activity that could not be read"}
```

Both rides' tiles on disk, and the one column an operator reads saying one was lost. The window is not a millisecond — it spans the whole of the last page's per-file work and the final flush.

So the arm is restored, and the comment states the window rather than a theorem with a hole. Nothing serves the pyramid yet, so there was no user-visible effect.

The removal also rested on a test that **passed with its entire staged interference deleted**. It now pins the state it actually reaches (`failed` / `Scan did not cover the whole history`), and a new test drives the window itself — upload, re-presentation, deletion after the page read — and fails when the arm is removed again.

### Two smaller things, and one recorded rather than fixed

The fixture in `tracks what it folded per file, not per page` had `readableId` naming the file that **throws** and `unreadableId` the one that folds. And `docs/fitness-file-storage.md` still called the failing claim "the exception with nothing to record on" after AGENTS.md was corrected in the same commit to name two such writes.

Recorded, not fixed: `canOwnPyramidBuild` trusts a carried token's *presence* rather than its *currency*, so a superseded continuation still takes the destructive compare-and-swap before learning it cannot use the build. The critic reproduced it, but it is only reachable if `requestedAt` were absent from the job data — and both production publishers always set it. A hardening note for whoever touches the claim next, not a defect today.

### What the critic probed and found sound

Seven gap areas, three of which no round had touched: `lib/database/testUtils.ts` as a subject rather than a tool, **privacy parity between the pyramid fold and the legacy blob** (the fold is handed the same post-privacy segments the blob goes on to region-filter, and `edgeKey` carries the hidden class so trimmed geometry cannot merge into visible), and the PostgreSQL × `--sequence.shuffle` cross-product.

### Closing the loop

| rounds | what they found |
|---|---|
| 1–8 | a reproduced data-corruption or wedge each, every one in the previous round's fix |
| 9–13 | no production defect; missing pins, and prose or reasoning this loop had written |

Thirteen rounds, **126 mutants killed**, and the last five found nothing that would be wrong in production. The final finding is a fitting one to end on: the loop's own argument for deleting a guard was wrong, five reviewers proved it, and the guard is back with the counter-example written down.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,767 tests**. Both touched suites on **PostgreSQL 17.10** (157/157), five consecutive combined SQLite runs, no flake.

This is ready to merge.

llun:
## Sub-agent review — round 14 (closing)

7 agents. **No High, no Medium in the code, and no production defect** — the sixth round running. The completeness critic returned **zero findings** and said: *"I would merge this PR as it stands."*

> Two agents died mid-round on an expired OAuth token; the run was resumed, the five cached dimensions replayed, and the tests dimension and critic re-ran for real. This summary is the complete round.

### Everything the dimensions found was in the test round 13 added

Six of them converged on it, and they were right on all three counts:

- **The staged upload was unasserted.** Deleting the activity whose arrival causes the re-presentation left the test green — so the window it exists to drive was never checked to have opened. It now asserts storage was asked for the re-presented file at all, which only happens if it really was handed back, and pins the counts that show the recount caught up. Removing the staging now fails it.
- **It leaked its fixture.** The third activity was deleted only from inside the page spy, so any earlier failure would have cascaded one failure into nine.
- The sibling test still called the arm "removed as inert" one commit after it was restored.

Also fixed: the claim's transaction pin was an inline copy of `expectOneTransaction` written before that helper existed, so it lacked the undefined-connection guard round 10 added. It uses the helper now — verified still to kill the `trx` → `database` slip on PostgreSQL.

### The one finding that outlived that fix, and it is prose

AGENTS.md, REVIEW.md and `docs/fitness-file-storage.md` all stated the coverage check catches an activity **added** during a build, unqualified — while the code's own comment fifty lines away records the window where it does not. The recount runs after the scan, so a deletion landing between the final page read and it cancels the addition's shortfall exactly.

That window is precisely why the fold keeps its own already-folded guard, so stating the check as unconditional is what would invite someone to remove that guard again as redundant — which this loop did once and had to undo. All three now say what they catch and what they do not.

### What the critic added, having looked where nobody had

- **It reviewed `eaa5fac0d`, which no dimension saw** — the commit that rewrites the very test round 13 added. All three of its claims verified on both backends.
- **Scale.** Every test in this diff is 1–3 activities over 1–2 passes. The critic ran **120 activities on one shared route across the page boundary**, checkpointing after every file: 120 consecutive claims, version never bumped, 201 tiles, **every edge count exactly 120**, identical on SQLite and PostgreSQL 17.10.
- **Recovery at that scale.** A fenced write failing at pass 60 of 120 leaves `failed`/version 1/60; the rest of the chain runs tokenless and touches nothing; a fresh Generate rebuilds to version 2 with counts of exactly **120, not 180**. That is the cross-dimension interaction — the release path × the version fence × the tile merge — nobody had assembled, and it is exactly where silent inflation would have shown.

### Closing the loop

| rounds | what they found |
|---|---|
| 1–8 | a reproduced data-corruption or wedge each, every one of them in the previous round's fix |
| 9–14 | no production defect; missing pins, and prose or reasoning this loop had written |

Fourteen rounds, **~130 mutants killed**, six consecutive rounds with nothing that would be wrong in production, and a critic that drove it at 120 activities rather than fixture scale before saying merge.

### Verification

Gate green in order: prettier, lint (0 errors), build, full suite at **816 files / 9,767 tests**. Both touched suites on **PostgreSQL 17.10** (157/157), five consecutive combined SQLite runs, no flake. CI green on the branch.

Merging.

PR #1464, opened 2026-08-19
fix: attribute gear by Strava activity id, and report what it misses

llun:
## What's wrong

A gear's distance is derived, never stored — `SUM(fitness_files.totalDistanceMeters)` over the rows attributed to it. So an activity the backfill never reached is a kilometre the gear page never shows.

On the instance this was found on, the Moots reads **34,569.4 km** against **35,767.0 km** on Strava. That is not a broken pipeline: today's ride was imported, parsed, auto-assigned and counted within hours. It is a *fixed* deficit — ~20 rides (~1,500 km) carrying no gear at all — and a constant shortfall reads exactly like a total stuck in the past.

The arithmetic closes: all-time countable cycling is 1,424 activities / 47,852.4 km, of which 1,404 / 46,356.6 km are attributed to a bike. The 1,495.9 km left over matches the three gear pages' combined shortfall against Strava (1,197.6 + 270.1 + 8.6 km).

## Why the backfill missed them

`matchAssignmentsToFiles` only ever matched an entry to the activity **nearest its timestamp, within 60 s**. The plan's `stravaActivityId` and `filename` were parsed and carried into the reports, but never used to match — so an entry whose two clocks had drifted further than the tolerance found nothing, and an activity with no start time at all was unreachable by any entry.

The script also only reported one direction: which *entries* found no activity. Nothing reported which *activities* no entry found — which is the direction the gear total is actually made of, and the reason this went unnoticed.

## What changed

**Identity decides first.** `stravaActivityId` is matched against the activity's `sourceUrl`, or against the `strava-<id>.tcx` name the webhook importer writes; then `filename` against the stored name, compared as basenames with any `.gz` stripped (the archive importer strips it before storing). That names the very row, so it holds however far the clocks have drifted, and it reaches an activity carrying no start time. Only when neither side names the row does the timestamp decide, unchanged.

Two rows sharing one Strava activity id tie and are skipped — that is a genuine duplicate, and picking either would attribute half a ride. A repeated *file name* does not tie: it is weaker evidence than the timestamp (two hand-uploaded `activity.gpx` are different rides), so it falls through rather than blocking the entry.

**The run reports the other direction.** `findUnattributedActivities` lists the completed, distance-carrying activities that no assignment, window, or auto-assign ever reached — totalled, broken down by year, with an unrecognised sport called out, since nothing can attribute that automatically. This is the diagnostic that answers "why is this gear short against Strava", and it prints in `--dry-run` too.

**Silent attribution failures are now logged.** `processFitnessFileJob` discarded `assignFitnessFileGearIfUnset`'s return value and skipped past an unmapped sport without a word. It now logs which of the three silent paths was taken and why.

Also corrects a stale comment in `createFitnessFile` claiming attribution can come from "Strava's own gear id" — imports deliberately never read it (AGENTS.md → Fitness Gear).

## Verification

`prettier` → `lint` (0 errors) → `build` → `test` (9,680 passing). 18 new unit test cases cover both identity paths, the fallbacks, the tie rules and the leftover report.

Because `scripts/` is lint-ignored and thinly tested, the script was also run end-to-end against a local SQLite database. Two entries — one an hour past the tolerance matched by Strava id, one an archive `.gz` path matched by name — both attributed, where time-only matching would have found neither; the two leftovers were reported and left untouched:

```
Unattributed activities (2, 67.5 km):
  These count toward no gear. A gear page short against Strava is short by these.
    2023: 2 activities, 67.5 km
  2023-06-03T08:00:00.000Z    55.5 km  ride  orphan.tcx
  2023-06-04T08:00:00.000Z    12.0 km  sport not recognised from "Kayaking"  kayak.tcx
```

## Operator follow-up

This fixes the tool; the data still needs a run. Take a fresh Strava export, rebuild the plan with `convertStravaExportToGearImport.ts`, then `importFitnessGear.ts --dry-run` and read the new unattributed report before the real run (no `--overwrite` — `assignFitnessFileGearIfUnset` only ever fills a hole). Separately, that instance has an empty retired duplicate gear row, "Moots Vamoots RSL disc" with 0 activities, left over from the original backfill and safe to delete.


llun:
### Review round 1 — sub-agent fan-out

Four reviewers (correctness, regression-vs-main, project rules, report logic) produced 13 findings; each was then handed to an independent agent asked to *refute* it. **11 were refuted**, 2 survived and are confirmed against the code.

Notable refutations, recorded so they are not re-raised:
- *"`reservedFileIds` hides ambiguous/conflicting activities from the unattributed report"* (raised independently by 3 reviewers) — refuted: reserving them is the deliberate, tested contract from #1412, and ambiguous/conflicting entries are already printed in their own section immediately above.
- *"A filename match at index 0 can pre-empt a strava-id match at index 1"* — refuted: real mechanically, but both outcomes are correct attributions of the same plan, and no documented invariant orders the two identity keys against each other.
- *"`deltaMilliseconds` becoming optional breaks a consumer"* — refuted: no consumer reads it outside the tests, which assert it only for time matches.

#### Confirmed 1 — `scripts/fitness/importFitnessGear.ts:448` (low)
The header still prints `N activities have no start time and cannot be matched.` That is no longer true: the identity pass indexes the **unfiltered** file list, so a file with no `activityStartTime` is now reachable by `sourceUrl` or file name — this PR adds a test asserting exactly that (`matches a file that has no start time at all`). Only the time pass filters those rows out of `candidates`.

#### Confirmed 2 — `scripts/fitness/gearImportPlan.ts:127` (low)
The comment above `stravaActivityId`/`filename` still reads *"Diagnostics only — carried through to the reports…"*. Both fields are now the primary match keys, consumed by pass 1 before any timestamp is considered. As written it tells a plan author these fields are optional decoration, when omitting them is what drops an entry back onto the 60 s time match this PR exists to stop relying on.

Fixing both now.

llun:
### Review round 2 — clean

Three reviewers over the updated diff (an audit of the round-1 fixes themselves, a second correctness pass, and one on the new logging): **9 findings, 0 survived refutation**. Two rounds, 22 findings, 2 confirmed and fixed — the loop terminates here.

The round-1 fixes were audited specifically and held up. Refutations worth recording:
- *"The `logger.info` fires on files that DO have gear"* — refuted: the guarded-UPDATE-returned-false branch is a deliberate, commented case ("the row or the gear moved mid-job"), and it is a real event worth a line either way.
- *"The reworded schema comment overclaims that an identity key always survives clock drift"* — refuted: the claim is conditional on the file naming itself, and the population it worried about (archive imports with no `sourceUrl`) is covered by the *filename* key, which has a test.
- *"The plan-file JSONC reference omits the identity keys"* — refuted as a defect, since the `assignments` half is machine-generated by the converter and never hand-authored. Applied anyway as an improvement (a6db2782c): the example showed only `{time, gear}`, which is now the degraded shape, so it carries both keys and a note that the converter writes them.

Final gate on the merged result: `prettier` → `lint` (0 errors) → `build` → `test`, **816 files / 9,680 tests passing**.

PR #1467, opened 2026-08-20
minor: serve the route heatmap pyramid over an owner and a share-token API

llun:
Phase 5 of the route heatmap precision work, on top of #1463. Two routes now read the pyramid that change started building. **Nothing renders from them yet** — the client's zoom-aware fetching is the next step — so every map still draws the legacy blob.

## What this adds

- **Owner tiles** — `GET /api/v1/accounts/[id]/fitness-route-heatmap/tiles`, the same auth chain as its sibling GET, answering a batch of `x:y` indices at one ladder zoom.
- **Public tiles** — `GET /embed/heatmap/[token]/tiles`, the same response shape for a share token, with no session at all.
- **`tileSource`** on the heatmap payload (`{version, minZoom, maxZoom, ladder, extent}`), on the owner API, the embed page and the public shared page. Null unless the row is the all-activities/all-time variant **and** the actor's pyramid has completed a build — the pyramid is per-actor and covers everything, so no filtered row can be answered from it.
- **`lib/client.ts`** — `getFitnessRouteHeatmapTiles` and `getPublicHeatmapTiles`, both abortable.

## The security boundary

The pyramid covers an actor's whole history; a share covers one rectangle of it. The caller sends tile indices and nothing else, so **the region is taken from the shared row and never from the request**. Without that, a rect share is a lookup oracle for everywhere its owner has ever been.

- Out-of-region tiles are settled from their coordinates **before any read**.
- A boundary tile is clipped **vertex by vertex** through the untiled heatmap's own `splitSegmentByBounds`, so the result cannot contain a point the region does not. A run left with a single vertex is dropped — it draws no line, and its presence alone would leak that something was there.
- The classification may be pessimistic (a union of rects covering a tile reads as straddling and clips to the same geometry) but never optimistic.

**The resolver fails closed**, which is the sharp edge. `getRegionBounds` answers `[]` both for the whole world and for a region string it could not parse a rectangle out of, and `[]` means "clip nothing" everywhere downstream — so a rect share whose stored token failed to parse would have served the world pyramid it was cut from. Only the empty string, the world sentinel `serializeRegions` emits, reaches the unclipped path; anything else resolving to no bounds is a 404. That refuses a value no writer produces, and it stays because the cost of being wrong is asymmetric: a false refusal drops the share back to the untiled geometry the page already carries, a false pass publishes an entire history.

The public path also strips the privacy flag while keeping the geometry — `flattenTilePrivacyForPublic` sits beside the untiled doctrine it mirrors, not in the route — and **re-encodes every byte it returns** rather than forwarding a stored payload.

## Version handling

Only a `completed` pyramid serves tiles, and only at its own `version`. A build in flight has two versions in the table at once, and drawing them together shows heat no build ever produced; the filter also catches what a thrown completion sweep leaves behind. The response's keys are the ones the **request** named, never the ones the read returned.

The `v` parameter is a cache-buster and never a filter. A request naming a version that has since moved is answered with the current tiles and the current version — refusing it would blank a map the instant a rebuild finished underneath a client still holding the previous `tileSource`.

## Also

Extracts `normalizeRegionParam`, which three routes had a copy of, so the rectangle tiles are clipped to is canonicalised exactly like the one the stored heatmap was generated for.

## Verification

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (821 files / 9,877 tests).

Seven mutants run against the new tests, all killed:

| Mutation | Tests failed |
| --- | --- |
| Region resolver stops failing closed | 2 |
| Out-of-region tiles read anyway | 4 |
| Straddling tile classified `inside` (served whole) | 6 |
| Privacy flag kept on the public path | 2 |
| Stale-version tiles served | 1 |
| Unfinished pyramid serves tiles | 5 |
| Owner route refuses a stale `v` | 1 |


llun:
## Review round 1 — 26 findings, 11 survived refutation, all addressed

Six of the seven dimensions independently reproduced **the same defect**, and it is the one this PR exists to prevent.

### The leak

The public tile route never asked whether the shared row is the variant the pyramid describes. A share token resolves to a heatmap **row** carrying three axes of scope — sport, period, region — and the pyramid answers only the third. So a share cut down to "runs in 2019" was served from a store covering every sport and every year, and region clipping could not catch it because such a share is usually world-wide.

The gate already existed in the codebase. It was on the pages that **advertise** tiles and missing from the route that **serves** them — the exact drift the phase was supposed to design out. The route now derives its go/no-go from the same `buildHeatmapTileSource` the pages call, rather than re-deriving the conditions, so one predicate cannot disagree with itself; that also folds in the completed/version checks it had been doing by hand.

`isPyramidVariantHeatmap` now tests `activityType` for null rather than falsiness. The job's gate is `activityType === null`, and a stored empty string would filter to nothing there while reading as "no filter" here — the one disagreement that serves a whole history behind a row showing none of it. The API cannot store one today; a predicate this load-bearing should not rest on that staying true.

### Also fixed

| Finding | Fix |
| --- | --- |
| Every tile request read and `JSON.parse`d the entire untiled heatmap blob to use the actor, status and region — before the 304 short-circuit | New `getFitnessRouteHeatmapSummaryByShareToken` using the existing `SUMMARY_COLUMNS` |
| The region resolver ran after the `If-None-Match` check | Moved ahead of it: no response, 200 or 304, without the security boundary having succeeded |
| The two routes disagreed on repeated parameters (`?z=4&z=16` → 16 vs 4) and on a malformed `v` | One shared `parseTileBatchQuery`; a malformed `v` is now a 400 on both |
| `getPublicHeatmapTiles` could never return the `version: 0` its own type documents | 404 translated to `version: 0`, so both fetchers honour one contract |
| Docs over-claimed: "every client-facing region uses the shared normalizer" and "`v` never refuses a request" | Both corrected in AGENTS.md / REVIEW.md / CLAUDE.md |

### Test gaps closed

Every region fixture padded longitude on both sides and clipped only in latitude — the longitude half of both the classifier and the clipper could be **deleted with the suite green**. The out-of-region fixtures differed on both axes, hiding a one-axis break in the pre-read guard. And nothing placed a region edge near a tile edge, leaving `contains` free to carry kilometres of slack.

Added longitude mirrors, single-axis out-of-region fixtures, exact-boundary classification cases, and pins for each fix above.

### Verification

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (821 files / **9,914** tests, up from 9,877).

Six new mutants against the fixes, all killed:

| Mutation | Tests failed |
| --- | --- |
| Public route drops the variant gate (the round-1 leak) | 3 |
| Variant predicate reads a blank sport as "no filter" | 2 |
| Public route reads the full share row again | 16 |
| Region resolver moved back after the 304 check | 1 |
| Query parser back to last-wins on repeats | 1 |
| Malformed `v` silently accepted | 4 |

15 findings were refuted with reasons — notably that the 304 path cannot leak (a client only holds a validator for a body a checked 200 gave it), that duplicate tile indices collapse in the response map, and that `activityType === ''` is unstorable. Round 2 follows, and it reviews these fixes.

llun:
## Review round 2 — 21 findings, **0 survived refutation**; the completeness critic found 4

The six dimensions raised 21 findings against round 1's fixes and the adversarial pass refuted every one — several by reproducing the mechanic and then showing the failure it implied cannot occur. The critic, which looks for what nobody was assigned, found the round's real defect.

### A privacy leak that predates this PR

`serializeRegions` validated a rectangle **before** rounding it to two decimal places. A box thinner than 0.01° in either axis passed the check and then collapsed, so the token it emitted deserialized to nothing:

```
rect:52.002,5.001,52.001,5.002  →  "rect:52.00,5.00,52.00,5.00"  →  []  →  world scope
```

Every consumer reads no bounds as world scope. So the generation job built the actor's **entire unclipped history** under a region string that reads as a small rectangle, and the share page captioned that with the rectangle's own bounding box. Someone who drew a tiny box and shared it published everywhere they have ever been, under a label saying otherwise.

Serialization now validates the **rounded** box, so such a scope becomes the world sentinel and is labelled "Whole world" wherever it is shown. Only boxes that were already broken change behaviour: every rectangle that survives rounding serializes to the byte-identical token, so no existing cache key moves.

This PR's fail-closed resolver already refused those rows — but its comment, and the docs, justified the refusal as being "for a value no writer produces". That was the opposite of the truth. Corrected in all three places; the refusal stays regardless, since rows written before this still hold such a token.

### Also fixed

| Finding | Fix |
| --- | --- |
| The region refusal was silent end to end — no server log, and round 1 removed the last client-side signal | Logged: it is the one 404 there meaning "this server cannot read its own stored data", and it is permanent for that share |
| `docs/fitness-file-storage.md` still called `v` a cache-buster that never refuses | Aligned with the three shorter docs, which the previous commit had already corrected |
| The new summary share-token read had **no database test** — the sole read behind an unauthenticated route, soft-delete guard unpinned | Added, including that the row carries no geometry |
| Neither tile fetcher in `lib/client.ts` was tested | Added: request shapes, the public route's absent region parameter, the 404 → `version: 0` translation, and that every other status still throws |
| The owner route's `normalizeRegionParam` call was unpinned — its fixture region was already canonical | Pinned with the degenerate rectangle above, where raw and canonical differ sharply |

### Verification

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (821 files / **9,928** tests).

Five new mutants, all killed:

| Mutation | Tests failed |
| --- | --- |
| `serializeRegions` validates before rounding again | 4 |
| Summary read drops its soft-delete guard | 1 |
| Public fetcher stops translating 404 | 1 |
| Public fetcher synthesises the wrong keys | 1 |
| Owner route stops canonicalizing the region | 1 |

Round 3 follows, and it reviews these.

llun:
## Review round 3 — 29 findings, 11 survived, +6 from the critic

The critic named one it would have blocked the merge on, and it was a regression from **round 2's own fix** — the third round running where the defect lived in the previous round's fix.

### The regression

Round 2 made `serializeRegions` resolve a rectangle collapsed by rounding to the world sentinel. But `HeatmapRegionPicker` still gated Save on `isValidRect` applied to the **raw** box — and a box under 0.01° is a perfectly well-formed rectangle. That is about a kilometre: an ordinary "scope this to my neighbourhood" drag, not an exotic input.

Saved, it took the **world's** key, and `FitnessHeatmapView` keys every per-region action on that — generate, cancel, delete, rename and **Share** all addressed the actor's whole-world heatmap while the row was labelled as a small area. With the world row also present it was instead dropped silently, with no error.

The rule is now one exported predicate, `isSerializableRect`, which `serializeRegions` itself applies and which anything *producing* a rectangle gates on. The two questions were never the same — `isValidRect` asks whether the box as drawn is well formed; this asks whether it has a canonical key of its own — and keeping them in one function is what stops a producer and the serializer drifting again.

It now also validates **both** before and after rounding, so the rule can only ever *drop* a rectangle. Checking only the rounded box admitted coordinates just outside range (latitude 90.004 rounds to 90.00), which would move a cache key that resolves to the world today and orphan the heatmap stored under it. Differential over ~40k generated rectangles: **5,063 dropped, 0 newly admitted, 0 otherwise changed**. Round 2's commit message claimed exactly this; it was not quite true until now.

### Existing rows

For a row already stored with a collapsed token the job baked the whole world into the untiled `segments`, so there is nothing left to clip. All four public surfaces — both tile routes, the share page, the embed page, the embed image — now refuse through one `resolveSharedHeatmapRegionBounds`, which also records each refusal. **No backfill is performed**: such a row keeps its data and stops being publishable.

### Also fixed

| Finding | Fix |
| --- | --- |
| The tile route's refusals carried no `Access-Control-Allow-Origin` while its answers carried `*`, so the client's documented "404 means no tiles" branch was unreachable off-origin | Refusals carry the same allow-origin, and `no-store` |
| The owner route still read `region` last-wins via `Object.fromEntries` — on the one scope-bearing parameter, the exact divergence class the shared parser closed | Read first-wins like everything else on the route |
| Three copies of round 2's rationale said the share page captioned a collapsed rectangle with its bounding box. **It never did** — `bboxLabel` needs exactly one rect and a collapsed token yields none, so the page titled it "Map area" and nothing a viewer could see said world | Corrected; the bounding box appears on the *owner's* page, which formats the box it holds rather than the token it serializes to |
| `AGENTS.md` still said "Nothing reads the pyramid yet" in the section this PR rewrote | Corrected |

### Coverage closed, each verified by a killed mutant

- **`roundTripRect` had only two of four coordinates pinned.** Every fixture collapsed both edges of an axis onto a shared value, so leaving `nw.lng` or `se.lat` unrounded passed the whole suite while re-emitting the collapsed token. Added fixtures crossing a rounding midpoint one axis at a time.
- **Nothing pinned that the public route ignores a caller-supplied `region`** — the boundary the route exists for. Adding `region` to the shared parser and honouring it left the entire suite green.
- The database test's "carries no geometry" assertion was a proxy the row parser guarantees for *any* query shape, `select('*')` included. Replaced with an assertion on the emitted SQL.
- The public fetcher's URL was half-asserted (`z` and `v` unchecked); the refusal log and the image route's refusal were unpinned.

### Verification

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (821 files / **9,950** tests).

Nine mutants across the two batches, all killed:

| Mutation | Tests failed |
| --- | --- |
| Picker gate back to `isValidRect` (the blocker) | 1 |
| `isSerializableRect` drops the rounded check | 7 |
| Image route stops refusing an unresolvable region | 2 |
| Shared resolver stops failing closed | 10 |
| Shared resolver stops recording the refusal | 2 |
| `roundTripRect` un-rounds `nw.lat` / `nw.lng` / `se.lat` / `se.lng` | 2 / 1 / 1 / 3 |

Round 4 follows.

llun:
## Review round 4 — 22 findings, 9 survived, 1 from the critic

**First round with nothing critical or high-severity as a defect.** The substantive finding is that round 3's own picker guard is inert — and it comes with a correction I owe.

### Correcting round 3's claim

Round 3's summary said a ~1 km drawn box could be saved and take the world's cache key. That is **not reachable through the UI**. Every box the composer can hold already carries 2-decimal-place corners — the map hands back `boxFromPoints`, which rounds both corners, and the coordinate fields commit through `toFixed(2)` — so `isSerializableRect` reduces to `isValidRect` inside the composer. Round 3's reproduction reached the branch only by injecting a 4-dp rectangle through `value`, which no production path creates.

What a user dragging a sub-0.01° box actually gets is a **collapsed** box: oriented correctly, no extent. It fails `isValidRect`, so they were told *"Top-left must be north-west of bottom-right"* — which is not what they did, and is the only case they can reach.

The two messages now branch on two named predicates instead of one that folds them together: `isOrientedRect` (corners the right way round, zero extent allowed) separates "you dragged it backwards" from "you dragged it too small". A second test covers the shape the app can actually produce; the 4-dp test stays, because that *is* the shape the API's `region` parameter accepts by design.

The guard itself stays — it costs nothing and keeps a future producer that doesn't pre-round from silently taking the world's key. But it is defence, not a live-bug fix, and round 3's summary over-claimed.

### Also fixed

| Finding | Fix |
| --- | --- |
| The refusal on the two share **pages** was the only one of four with no test — and they are the surfaces that render the unclipped geometry | `page.test.tsx` for both, covering refusal *and* the two shapes that must still render |
| Refusal headers pinned on 2 of the route's 5 refusal sites, not including the security one | Table now covers every refusal it can emit |
| Docs said "**both** tile routes" refuse through `resolveSharedHeatmapRegionBounds` — the owner route does not, and must not: it clips to the region its own authenticated caller sent | Corrected in four files |
| Docs said a false refusal costs only zoom detail — untrue since the refusal moved onto the pages | Corrected: over-refusing now costs the entire share |

### Verification

Rebased onto current `main` (zero file overlap with the landing-page fix that landed meanwhile). Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (**825 files / 9,972 tests**).

Four new mutants, all killed:

| Mutation | Tests failed |
| --- | --- |
| Composer message branches on the wrong predicate | 2 |
| Share page stops refusing an unresolvable region | 2 |
| Embed page stops refusing an unresolvable region | 2 |
| Refusals lose their allow-origin header | 4 |

13 findings were refuted, notably that the share/generate APIs "silently widen" a sub-precision rectangle (the resulting row is honestly the world heatmap, labelled as such), and that `v` is an unbounded cache-buster.

Round 5 follows.

llun:
## Review round 5 — 10 findings, 1 survived; critic verdict: **merge**

The loop has converged: nothing critical or high, and no defect introduced by round 4. The critic's verdict, quoted:

> Yes, merge. Round 4 introduced nothing; the loop has converged. […] I found no path by which a scoped share reads beyond its scope […] the new refusals fire only for rows that are already broken […] with no tile consumer shipped the live blast radius of the whole PR is the additive `tileSource` field plus four fail-closed refusals.

I've still fixed everything it raised, because two items were real.

### A refutation from round 2 that was wrong

The pyramid read this PR added to the **owner's** heatmap GET was unguarded, so a failure of the tile-pyramid query cost the owner the untiled heatmap they can actually see. That is the exact trade this feature's rules forbid — *"tile work never fails the run"* — and the exact trade both public pages already avoid with `.catch(() => null)` on the same call. Round 2 raised this and the adversarial pass refuted it; the critic reinstated it and was right. Fixed, pinned, and the rule is now written down for every surface that reads the pyramid only to publish a `tileSource`.

### One message cell still wrong

Round 4 separated "too small" from "wrong way round", but a box whose corners are **off the globe** still landed in the direction branch — and that is reachable: a map panned onto the next copy of the world reports unwrapped longitudes, so a drag at 2°E arrives as **362** and a perfectly well-drawn box was told to reverse corners it had not reversed. `isRectInRange` now names that case, and collapses the range checks `isValidRect` and `isOrientedRect` had been repeating.

### Coverage

- **`classifyTileAgainstRegion` reads a scope as a union**, but the only multi-rect test used two rects that *both* straddled the tile — so `some` → `every` on the intersection line survived the whole suite. That mutant answers `outside` for a tile meeting only one rect of a scope, and `serveHeatmapTiles` settles an `outside` tile *without reading it*, so every multi-rect scope would go blank at tiled zooms. Pinned in both bound orders, and for the `contains` line too.
- The three rectangle predicates now have direct tests, including that `isSerializableRect` agrees with what `serializeRegions` actually emits — the invariant the composer's gate depends on.
- `AGENTS.md` said "the tile routes read the share row"; only the public one does.

### Verification

Rebased onto current `main`. Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (**825 files / 9,997 tests**).

Three new mutants, all killed:

| Mutation | Tests failed |
| --- | --- |
| Owner GET stops degrading a failed pyramid read | 1 |
| Picker drops the out-of-range message | 1 |
| `isOrientedRect` requires a non-empty box | 4 |

Plus the two union mutants above (2 and 1).

Round 6 will confirm this round introduced nothing; if it is clean, this is ready.

llun:
## Review round 6 — **clean**. All four dimensions empty; critic clears the merge.

No dimension raised a finding, and the critic verified round 5's fixes independently rather than taking them on trust:

- **238,196-case differential** comparing verbatim `HEAD~1` copies of `isValidRect`/`isOrientedRect` against HEAD's — full cross-product of edge coordinates (±Infinity, NaN, −0, ±90.0001, ±180.0001, 362, 1e308) plus 200,000 random quadruples. **Zero disagreements.** The `isRectInRange` extraction is provably behaviour-preserving.
- **Load-bearing check**: reverting only the composer's three-branch message makes exactly the new antimeridian test fail (1 failed / 26 passed), then byte-identical restore.
- **Gate reproduced**: 825 files / 9,997 tests, eslint 0 errors, prettier clean, `next build` exit 0.
- **Unlisted sub-change audited**: the `AGENTS.md` correction from "the tile routes read the share row" to "the public tile route" is factual — `getFitnessRouteHeatmapSummaryByShareToken` has exactly one caller.
- **The new best-effort rule is applied exactly where claimed**: all three non-tile pyramid reads carry `.catch(() => null)`; both tile routes read it unguarded, which is correct.

Verdict, quoted:

> Round 5's fixes introduced nothing. All 8 changed paths are accounted for, the `isRectInRange` extraction is provably behaviour-preserving for `isValidRect` and `isOrientedRect`, and the full local gate is clean. The verdict stands: merge.

### One non-gating follow-up, deliberately not taken here

The three best-effort pyramid reads swallow their rejection silently, where the project's Logging Guidelines want `logger.warn({ err: toLoggableError(error) })`. All three sites are new in this PR, so it is mine to fix — but adding unreviewed edits immediately after a clean round is exactly how the last five rounds each found a defect in the previous round's fix. It lands as the first commit of Phase 6, where it gets reviewed properly.

---

## The loop, end to end

| Round | Raised → survived | Worst finding |
| --- | --- | --- |
| 1 | 26 → 11 | **critical** — a sport/period-scoped share served the whole history |
| 2 | 21 → 0 (+4 critic) | **high** — a sub-0.01° box became world scope under a rectangle's label |
| 3 | 29 → 11 (+6 critic) | **high** — round 2's fix let a drawn box take the world's cache key |
| 4 | 22 → 9 (+1 critic) | **medium** — dead error branch; a too-small box told the wrong reason |
| 5 | 10 → 1 (+2 critic) | **medium** — owner GET lost the heatmap to a failed pyramid read |
| 6 | 0 → 0 | — |

Three of the six rounds found their defect inside the previous round's fix, and round 5's came from a finding round 2 had **refuted** — a reminder that a refutation is an argument, not a verdict.

~40 mutants were run across the rounds; every one reported is a mutant that was applied, run, and killed.

Merging.

PR #1468, opened 2026-08-20
fix: stop the landing page's public-post gate scanning the whole timeline

llun:
## Problem

Cloud SQL Query Insights had this at a **27.05 s average over 83 calls**:

```sql
SELECT DISTINCT statuses.id FROM recipients
INNER JOIN statuses ON recipients."statusId" = statuses.id
INNER JOIN actors   ON statuses."actorId"  = actors.id
WHERE recipients.type = 'to' AND recipients."actorId" = '…#Public'
  AND actors."privateKey" IS NOT NULL AND statuses.reply = ''
LIMIT 100
```

It is `getLocalPublicStatusesCount`, which the logged-out landing page runs on **every anonymous request** — every crawler included — to answer one yes/no question: does this server hold at least 100 public posts (`LANDING_FEED_MIN_POSTS`)?

### The `LIMIT 100` was doing nothing

A `Sort` feeding `DISTINCT` must consume the **entire** join before it emits its first row; `LIMIT` only truncates the final result. So the query always did the full unbounded work the limit was added to avoid. With no limit to push down, the planner also optimised for total cost rather than first-100-rows cost and drove from the least selective side.

Reading the sampled plan bottom-up (inner-node latencies are per-loop averages, and multiplying them out reproduces the parent totals):

| Node | Rows | Latency | Multiplied out |
|---|---|---|---|
| Seq Scan on `actors` | 73 | 3.12 s | full scan, no index for `privateKey IS NOT NULL` |
| Index Scan on `statuses` | 122/loop | 1.1 s | × 73 loops = **80 s** |
| **Nested Loop** actors × statuses | 9,030 | 1.34 min | flagged highest latency / rows / cost |
| Index Scan on `recipients` | ~0/loop | 7.2 ms | × 9,030 loops = **65 s** — a heap fetch per probe |
| **Nested Loop** × recipients | 2,406 | 1.09 min | ~73% of probes found nothing |
| Sort → Unique → Gather Merge | 465 → 100 | 2.96 s + 6.62 s | |

## Fix

**Express `recipients` as the semi-join it always was.** A status qualifies when it has at least one public `'to'` recipient, and must be produced once however many it has — exactly what `DISTINCT` was for. `whereExists` gets that without materialising the product first, so the `LIMIT` genuinely stops the scan.

Measured on a seeded local PostgreSQL 17 (4,040 actors / 8,000 statuses / 4,800 recipients), `EXPLAIN (ANALYZE, BUFFERS)`:

| | before | after |
|---|---|---|
| Plan top | `Limit → Unique → Sort` | `Limit → Nested Loop Semi Join` |
| Statuses scanned | 4,800 (all) | **163** |
| Join rows produced | 3,068 | **100** (exactly the limit) |
| Buffers | 14,554 | **504** |
| Sort | quicksort, 97 kB | **none** |

The `LOCAL_PUBLIC` feed branch (landing preview + `GET /api/v1/timelines/public?local=true`) now shares the same builder, so the gate can never disagree with the feed it gates. That feed had **no** `DISTINCT` at all, so this also fixes a latent duplicate: a status addressed to the public collection twice appeared in the timeline twice. There is a regression test, and it fails against the old code with `expected […] to have a length of 1 but got 2`.

### Indexes

Three, all confirmed used against a local PostgreSQL 17:

- `actors_local_idx` — partial on `(id) WHERE "privateKey" IS NOT NULL`. Replaces the 3.12 s sequential scan; **Index Only Scan, Heap Fetches: 0**. Indexing `privateKey` itself isn't an option (it holds a PEM key).
- `recipients_status_type_actor_idx` on `("statusId", type, "actorId")`. The existing index omits `actorId`, so each probe needed a random heap fetch to check the recipient. Now **Index Only Scan, Heap Fetches: 0**.
- `statuses_reply_created_idx` on `(reply, "createdAt", id)`. No index *led* with `statuses.createdAt` (`statuses_actorId_idx` carries it, but behind `actorId`, so it cannot serve a global ordering). Led by `reply` as an equality column, a backward scan produces the feed's exact ordering — verified to give `Index Scan Backward`, no `Sort`, 41 statuses read instead of 4,800, **9,751 → 167 buffers**.

Deliberately **not** a partial index on `(createdAt, id) WHERE reply = ''`: SQLite rejects the bound parameter knex emits for it, and on PostgreSQL a partial index is only usable when the planner can prove the query predicate implies the index predicate — which it cannot for `reply = $n` under a generic plan. As a leading equality column, `reply` has neither problem.

### Caching

The gate is the unauthenticated front door and its answer changes at most once in an instance's life. Cached per database instance, following the TTL pattern in `lib/services/serverSettings`: a result that **reached** the threshold is memoized for the process lifetime, and only a below-threshold one expires (60 s) so a young server's landing unlocks promptly. Failures aren't cached.

## Note on the one changed assertion

`app/api/v1/statuses/[id]/route.test.ts` pinned the order of an Announce's `cc` array. The recipients read behind it has no `ORDER BY`, so the new index changes which index SQLite picks and therefore the array order. An ActivityPub audience is a set — every consumer already treats it as one (`lib/services/federation/statusDelivery.ts` uses `.some()` / `Set` / `.filter()`) — so it now compares as a set. No production code indexes into `to`/`cc` by position.

## Verification

- `yarn run prettier --write .` → `yarn lint` (0 errors) → `yarn build` → `yarn test` (**817 files / 9,777 tests pass**)
- Migration applied against throwaway SQLite and local PostgreSQL 17; `knex_migrations` = 152 = migration file count on both
- Both reference schema dumps regenerated; diffs are purely additive (the three indexes)
- Plan shapes above measured with `EXPLAIN (ANALYZE, BUFFERS)` on seeded local PostgreSQL 17

## Follow-up (ops, not this PR)

The sampled plan chose `statuses_announce_actor_original_idx` (`type, actorId, originalStatusId`) for a lookup on `actorId` while `statuses_actorId_idx` (`actorId, …`) — a textbook fit — sat unused. That usually means stale statistics or an invalid index; worth checking `pg_index.indisvalid` and `last_autoanalyze` on prod. Index-only scans also depend on the visibility map, so the two new covering indexes reach full effect once autovacuum has run.

### DIFF
--- a/app/api/v1/statuses/[id]/route.test.ts
+++ b/app/api/v1/statuses/[id]/route.test.ts
@@ -657,7 +657,15 @@ describe('GET /api/v1/statuses/[id]', () => {
       })
       expect(announce).not.toBeNull()
       expect(announce?.to).toEqual([`${ACTOR2_ID}/followers`])
-      expect(announce?.cc).toEqual([ACTIVITY_STREAM_PUBLIC, ACTOR2_ID])
+      // Compared as a set: `cc` is an ActivityPub audience, which is unordered
+      // by definition, and the recipients are read back without an ORDER BY —
+      // so which index the backend picks decides the array order. Every
+      // consumer treats these as sets (see lib/services/federation/
+      // statusDelivery.ts); asserting a sequence here only pinned an
+      // implementation detail.
+      expect(new Set(announce?.cc)).toEqual(
+        new Set([ACTIVITY_STREAM_PUBLIC, ACTOR2_ID])
+      )
     })
 
     it('defaults to a public boost when no visibility is sent', async () => {
--- a/app/api/v1/statuses/[id]/route.test.ts
+++ b/app/api/v1/statuses/[id]/route.test.ts
@@ -657,14 +657,15 @@ describe('GET /api/v1/statuses/[id]', () => {
       })
       expect(announce).not.toBeNull()
       expect(announce?.to).toEqual([`${ACTOR2_ID}/followers`])
-      // Compared as a set: `cc` is an ActivityPub audience, which is unordered
-      // by definition, and the recipients are read back without an ORDER BY —
-      // so which index the backend picks decides the array order. Every
-      // consumer treats these as sets (see lib/services/federation/
-      // statusDelivery.ts); asserting a sequence here only pinned an
-      // implementation detail.
-      expect(new Set(announce?.cc)).toEqual(
-        new Set([ACTIVITY_STREAM_PUBLIC, ACTOR2_ID])
+      // Sorted before comparing: `cc` is an ActivityPub audience, which is
+      // unordered by definition, and the recipients are read back without an
+      // ORDER BY — so which index the backend picks decides the array order.
+      // Every consumer treats these as sets (see lib/services/federation/
+      // statusDelivery.ts), so asserting a sequence only pinned an
+      // implementation detail. Sorted arrays rather than Sets, because a Set
+      // would also stop this catching a duplicated recipient.
+      expect([...(announce?.cc ?? [])].sort()).toEqual(
+        [ACTIVITY_STREAM_PUBLIC, ACTOR2_ID].sort()
       )
     })
 
@@ -701,7 +702,12 @@ describe('GET /api/v1/statuses/[id]', () => {
       })
       expect(announce).not.toBeNull()
       expect(announce?.to).toEqual([ACTIVITY_STREAM_PUBLIC])
-      expect(announce?.cc).toEqual([ACTOR2_ID, `${ACTOR2_ID}/followers`])
+      // Sorted for the same reason as the unlisted-boost case above: the
+      // recipients read has no ORDER BY. This one happened to keep passing only
+      // because `ACTOR2_ID` sorts before its own `/followers` suffix.
+      expect([...(announce?.cc ?? [])].sort()).toEqual(
+        [ACTOR2_ID, `${ACTOR2_ID}/followers`].sort()
+      )
     })
 
     it('rejects an invalid reblog visibility with 422', async () => {
--- a/lib/database/sql/localPublicQueryShape.test.ts
+++ b/lib/database/sql/localPublicQueryShape.test.ts
@@ -1,4 +1,4 @@
-import { getTestSQLDatabaseWithInstance } from '@/lib/database/testUtils'
+import { getTestDatabaseWithInstance } from '@/lib/database/testUtils'
 import { Timeline } from '@/lib/services/timelines/types'
 
 // Guards the SHAPE of the local public timeline query, not its results.
@@ -11,16 +11,34 @@ import { Timeline } from '@/lib/services/timelines/types'
 // the plan that had this query at ~27s on the landing page. Without a test that
 // looks at the SQL itself, that revert lands silently and green.
 //
-// Asserting on `toSQL()` follows the precedent in `search.test.ts`.
+// Capturing knex's `query` event follows the precedent in `status.test.ts`
+// (the batched-hydration tests). It is preferred over asserting on a builder's
+// `toSQL()` here because it observes what the production path actually sends,
+// rather than a builder the test constructed for itself.
+//
+// Built on `getTestDatabaseWithInstance` rather than the SQLite-only
+// `getTestSQLDatabaseWithInstance`, because the query this pins is a
+// PostgreSQL fix and `testUtils` is explicit that the SQLite-only helpers
+// report a clean run under the pg environment variables without ever opening a
+// PostgreSQL connection.
 describe('local public timeline query shape', () => {
-  const { database, instance } = getTestSQLDatabaseWithInstance()
+  const { database, instance, prepare } = getTestDatabaseWithInstance()
+
+  // Identifier quoting is per-dialect — backticks on SQLite, double quotes on
+  // PostgreSQL — so strip it and let the assertions describe the query shape
+  // rather than whichever backend the suite is running against. Bound values
+  // can never appear here: knex emits `?`/`$n` placeholders and passes the
+  // values separately.
+  const normalize = (sql: string) => sql.replaceAll('`', '').replaceAll('"', '')
 
   // Capture what actually reaches the driver, so this tracks the query the
   // production path builds rather than a helper exported for the test's sake.
-  const capture = async (run: () => Promise<unknown>): Promise<string[]> => {
+  const captureLocalPublicQuery = async (
+    run: () => Promise<unknown>
+  ): Promise<string> => {
     const statements: string[] = []
     const listener = ({ sql }: { sql: string }) => {
-      statements.push(sql)
+      statements.push(normalize(sql))
     }
     instance.on('query', listener)
     try {
@@ -28,21 +46,24 @@ describe('local public timeline query shape', () => {
     } finally {
       instance.off('query', listener)
     }
-    return statements
-  }
 
-  // Identifier quoting is per-dialect (backticks here, double quotes on
-  // PostgreSQL); strip it so the assertions describe the query shape rather
-  // than the backend the harness happens to use.
-  const recipientsStatement = (statements: string[]) => {
-    const matches = statements
-      .map((sql) => sql.replaceAll('`', '').replaceAll('"', ''))
-      .filter((sql) => sql.includes('recipients'))
+    // Selected rather than assumed to be the only one: hydrating a status
+    // issues its own `recipients` reads, so a future seeded row here would
+    // otherwise turn a shape regression into a confusing "expected 1, got 3".
+    // Deliberately matched on both table names and not on which one is in the
+    // FROM — a revert puts `recipients` there — so that a regression fails on
+    // the assertion that names it rather than on finding no statement at all.
+    // The hydration reads mention only `recipients` (their column is
+    // `statusId`, not `statuses`), so they never match.
+    const matches = statements.filter(
+      (sql) => sql.includes('recipients') && sql.includes('statuses')
+    )
     expect(matches).toHaveLength(1)
     return matches[0]
   }
 
   beforeAll(async () => {
+    await prepare()
     await database.migrate()
   })
 
@@ -63,7 +84,7 @@ describe('local public timeline query shape', () => {
   ])(
     'reads recipients as a semi-join with no DISTINCT for $description',
     async ({ run }) => {
-      const sql = recipientsStatement(await capture(run))
+      const sql = await captureLocalPublicQuery(run)
 
       // A semi-join: correlated EXISTS rather than a row-multiplying join.
       expect(sql).toContain('exists (')
--- a/lib/database/sql/localPublicQueryShape.test.ts
+++ b/lib/database/sql/localPublicQueryShape.test.ts
@@ -1,0 +1,81 @@
+import { getTestSQLDatabaseWithInstance } from '@/lib/database/testUtils'
+import { Timeline } from '@/lib/services/timelines/types'
+
+// Guards the SHAPE of the local public timeline query, not its results.
+//
+// The semi-join in `localPublicStatusesQuery` is the entire performance fix:
+// it is what lets `LIMIT` bound the scan. Rewriting it back to
+// `innerJoin('recipients', …).distinct()` is behaviourally identical for every
+// other assertion in the suite — the counts stay right and the duplicate
+// regression test still passes, because DISTINCT dedupes too — while restoring
+// the plan that had this query at ~27s on the landing page. Without a test that
+// looks at the SQL itself, that revert lands silently and green.
+//
+// Asserting on `toSQL()` follows the precedent in `search.test.ts`.
+describe('local public timeline query shape', () => {
+  const { database, instance } = getTestSQLDatabaseWithInstance()
+
+  // Capture what actually reaches the driver, so this tracks the query the
+  // production path builds rather than a helper exported for the test's sake.
+  const capture = async (run: () => Promise<unknown>): Promise<string[]> => {
+    const statements: string[] = []
+    const listener = ({ sql }: { sql: string }) => {
+      statements.push(sql)
+    }
+    instance.on('query', listener)
+    try {
+      await run()
+    } finally {
+      instance.off('query', listener)
+    }
+    return statements
+  }
+
+  // Identifier quoting is per-dialect (backticks here, double quotes on
+  // PostgreSQL); strip it so the assertions describe the query shape rather
+  // than the backend the harness happens to use.
+  const recipientsStatement = (statements: string[]) => {
+    const matches = statements
+      .map((sql) => sql.replaceAll('`', '').replaceAll('"', ''))
+      .filter((sql) => sql.includes('recipients'))
+    expect(matches).toHaveLength(1)
+    return matches[0]
+  }
+
+  beforeAll(async () => {
+    await database.migrate()
+  })
+
+  afterAll(async () => {
+    await database.destroy()
+  })
+
+  it.each([
+    {
+      description: 'the LOCAL_PUBLIC feed',
+      run: () =>
+        database.getTimeline({ timeline: Timeline.LOCAL_PUBLIC, limit: 20 })
+    },
+    {
+      description: 'the landing page threshold count',
+      run: () => database.getLocalPublicStatusesCount(100)
+    }
+  ])(
+    'reads recipients as a semi-join with no DISTINCT for $description',
+    async ({ run }) => {
+      const sql = recipientsStatement(await capture(run))
+
+      // A semi-join: correlated EXISTS rather than a row-multiplying join.
+      expect(sql).toContain('exists (')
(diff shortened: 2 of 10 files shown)


lib/services/timelines/localPublicCount.ts
```diff
@@ -0,0 +1,67 @@
...
+    count,
+    ...(count >= limit
+      ? {}
+      : { expiresAt: Date.now() + BELOW_THRESHOLD_CACHE_TTL_MS })
+  })
+  return count
+}
+
+/** Test-only: drop a database's cached entry so cases do not leak into each other. */
+export const clearLocalPublicStatusesCountCache = (database: Database) => {
```
llun: **Dead code: `clearLocalPublicStatusesCountCache` is not needed.**

Its stated rationale is "so cases do not leak into each other", but `mockDatabase()` in the test file returns a **new object literal on every call**, so every test already gets a distinct `WeakMap` key. Cross-test leakage is structurally impossible.

The only caller is `localPublicCount.test.ts:111`, in a test that exists solely to exercise the helper that exists solely for that test — a closed loop shipping in the production bundle.

If it were to stay, it is also inconsistent with the repo's own pattern for test-only resets, two of which refuse to run outside the test runner (`lib/config/trustProxyIpHeaders.ts:15`, `lib/utils/http-headers/csp.ts:122` both throw unless `process.env.VITEST`) and use a `…ForTests` / `…ForTesting` suffix.

Removing the export and its test.

llun: Dead code: clearLocalPublicStatusesCountCache is not needed. mockDatabase() in the test file returns a new object literal on every call, so every test already gets a distinct WeakMap key — cross-test leakage is structurally impossible. Its only caller is localPublicCount.test.ts:111, a test that exists solely to exercise the helper that exists solely for that test, shipping in the production bundle. It is also inconsistent with the repo pattern for test-only resets (lib/config/trustProxyIpHeaders.ts:15 and lib/utils/http-headers/csp.ts:122 both throw unless process.env.VITEST, and use a ForTests/ForTesting suffix). Removing the export and its test.

app/api/v1/statuses/[id]/route.test.ts
```diff
@@ -657,7 +657,15 @@ describe('GET /api/v1/statuses/[id]', () => {
...
       expect(announce).not.toBeNull()
       expect(announce?.to).toEqual([`${ACTOR2_ID}/followers`])
-      expect(announce?.cc).toEqual([ACTIVITY_STREAM_PUBLIC, ACTOR2_ID])
+      // Compared as a set: `cc` is an ActivityPub audience, which is unordered
+      // by definition, and the recipients are read back without an ORDER BY —
+      // so which index the backend picks decides the array order. Every
+      // consumer treats these as sets (see lib/services/federation/
+      // statusDelivery.ts); asserting a sequence here only pinned an
+      // implementation detail.
+      expect(new Set(announce?.cc)).toEqual(
```
Comment (llun) [inline app/api/v1/statuses/[id]/route.test.ts]: Set comparison is order-insensitive but also swallows duplicates, which is a blind spot this PR in particular should not introduce: its own localPublicStatusesQuery comment notes duplicate recipients rows are reachable (remote to/cc is not de-duplicated on insert), and one of the bugs this PR fixes is a status appearing twice for exactly that reason. Verified: new Set([PUBLIC, ACTOR2, ACTOR2]) deep-equals new Set([PUBLIC, ACTOR2]), so a duplicated cc would pass. Comparing sorted arrays keeps the order-insensitivity and still fails on a duplicate. Switching to expect([...(announce?.cc ?? [])].sort()).toEqual([ACTIVITY_STREAM_PUBLIC, ACTOR2_ID].sort()).

migrations/20260820000000_add_local_public_timeline_indexes.js
```diff
@@ -0,0 +1,92 @@
...
+    )
+  })
+
+  await knex.schema.alterTable('statuses', function (table) {
+    // Nothing indexed `statuses.createdAt` at all, so the public feed's
+    // `ORDER BY createdAt DESC, id DESC` sorted every candidate status on every
+    // page. Led by `reply` — which both public surfaces pin to '' for top-level
+    // posts — the remaining columns are in the feed's exact sort order, so a
+    // backward scan of the `reply = ''` prefix produces the page directly and
+    // the LIMIT stops it after a page's worth of rows.
```
llun: Factually wrong comment: "Nothing indexed statuses.createdAt at all" is not true — statuses_actorId_idx is ("actorId", "createdAt", "updatedAt") (schema.sql:1827). The substantive point survives, since createdAt is not a LEADING column there and so cannot serve the feed global ORDER BY createdAt DESC, id DESC, but the sentence as written is checkably false and will mislead the next person auditing index coverage. Rewording to "No index LED with statuses.createdAt". The same wrong sentence is in the PR description and will be fixed there too.

migrations/20260820000000_add_local_public_timeline_indexes.js
```diff
@@ -0,0 +1,92 @@
...
+// each of those a scan for their statuses, then ~9,000 probes into `recipients`
+// that each needed a heap fetch. `lib/database/sql/timeline.ts` fixes the query
+// shape (a semi-join so a LIMIT can bound the scan); these three indexes give
+// that shape something to run on.
+//
+// Only `actors_local_idx` is partial, and only because its predicate is a bare
+// `IS NOT NULL` that carries no bound parameter — see the note on the statuses
+// index for why a parameterised predicate is the wrong tool here. Knex emits
+// the `where` clause on PostgreSQL and SQLite; MySQL-compatible clients ignore
+// it and get the equivalent full index, which is larger but still correct.
```
llun: The MySQL claim overstates what MySQL gets. On MySQL the predicate is dropped, so actors_local_idx becomes a plain index on (id) — but actors.id already carries actors_id_unique UNIQUE (id) (schema.sql:1238), making it a wholly redundant duplicate of an existing unique key with zero local-actor selectivity, not an "equivalent" index that is merely "larger". Nothing breaks, so this is comment accuracy only, but the current wording implies MySQL still gets the optimisation. Rewording.

llun:
Review finding (could not anchor inline — the line is unchanged context, outside the diff):

**`app/api/v1/statuses/[id]/route.test.ts:705` — the same order-dependence this PR just fixed, left in place one screen away.**

```
expect(announce?.cc).toEqual([ACTOR2_ID, `${ACTOR2_ID}/followers`])
```

This is the identical construct to the assertion at ~657 that the PR de-orders. It passes only by luck: under the new `recipients_status_type_actor_idx` the rows come back in `actorId` order, and `ACTOR2_ID` is a prefix of `${ACTOR2_ID}/followers`, so it happens to sort first. It will break the same way on the next index or planner change.

Applying the same sorted-array treatment for consistency. The single-element `to`/`cc` assertions elsewhere (lines 2586-2587, `lib/jobs/createPollJob.test.ts:460-461`) are genuinely order-independent and need no change.

lib/database/sql/timeline.ts
```diff
@@ -11,6 +11,38 @@ import {
...
+// to consume the *entire* join before LIMIT sees a single row — so
+// `SELECT DISTINCT … LIMIT n` never stopped early, and the planner, with no
+// limit to push down, optimised for total cost and drove from the least
+// selective side (every local actor, then all of their statuses). That is what
+// put this query at ~27s on the logged-out landing page.
+//
+// It also removes a latent duplicate: the feed had no DISTINCT at all, so a
+// status carrying the public collection twice in `to` (remote input is not
+// de-duplicated on insert) appeared twice in the timeline.
+const localPublicStatusesQuery = (database: Knex) =>
```
llun, line 33: Nothing pins the query SHAPE, so the whole performance fix can be reverted invisibly. localPublicStatusesQuery IS the fix — the semi-join is what lets LIMIT bound the scan. Rewriting it back to innerJoin(recipients).distinct() is behaviourally identical for every assertion in the suite (counts stay right, and the duplicate regression test passes too because DISTINCT dedupes as well), while restoring the pathological plan. Confirmed with EXPLAIN QUERY PLAN on the committed SQLite dump: the innerJoin form adds USE TEMP B-TREE FOR DISTINCT, which must materialise everything before LIMIT. There is repo precedent for guarding this — search.test.ts asserts on query.toSQL().sql in five places. Adding lib/database/sql/localPublicQueryShape.test.ts, which captures the SQL actually sent to the driver (via the knex query event, so it tracks the production path rather than a helper exported for the test) and asserts it contains a correlated exists and no distinct, for both the feed and the count.

lib/database/sql/timeline.test.ts
```diff
@@ -842,6 +842,37 @@ describe('TimelineDatabase', () => {
...
+        // recipient rows for one status. Joining `recipients` would then
+        // produce that status twice; the semi-join must not.
+        const before = await database.getLocalPublicStatusesCount()
+        await database.createNote({
+          actorId: COUNT_ACTOR,
+          cc: [],
+          to: [ACTIVITY_STREAM_PUBLIC, ACTIVITY_STREAM_PUBLIC],
+          id: `${COUNT_ACTOR}/statuses/count-duplicate-recipient`,
+          text: 'Addressed to the public collection twice',
+          url: `${COUNT_ACTOR}/statuses/count-duplicate-recipient`,
```
llun, line 858: The bounded count branch — the one production actually calls — is not covered for the duplicate case. This test asserts through getLocalPublicStatusesCount() with NO argument, i.e. the unbounded count() branch. The only test of the limit !== undefined branch is "stops counting at the given limit" with limit: 1, which cannot distinguish a correct count from a duplicate-inflated one since rows.length caps at 1 either way. app/(timeline)/page.tsx only ever calls the bounded branch (LANDING_FEED_MIN_POSTS = 100). Concrete failure: if localPublicStatusesQuery regressed to an innerJoin without DISTINCT, a server with 60 public posts each carrying the public collection twice would report 100 and show the preview the gate exists to withhold — with the whole suite still green. Adding an assertion that the bounded branch agrees with the unbounded one at an unreachable limit.

lib/services/timelines/localPublicCount.ts
llun: Fixed in d295fd341 — export and its test removed. Confirmed the reasoning: mockDatabase() builds a new object literal per call, so every case already had its own WeakMap key. Added a note to that effect where the mock is defined, so the next person does not re-add a reset for the same imagined reason.

app/api/v1/statuses/[id]/route.test.ts
Comment (llun) [reply app/api/v1/statuses/[id]/route.test.ts]: Fixed in d295fd341. Verified the blind spot concretely before changing it: new Set([A,B,B]) deep-equals new Set([A,B]), while [A,B,B].sort() does not equal [A,B].sort(). Now sorted arrays, which keep the order-insensitivity and still fail on a duplicate. Good catch — I had traded away duplicate detection I did not need to give up.

migrations/20260820000000_add_local_public_timeline_indexes.js
llun: Fixed in d295fd341 — reworded to "No index LED with statuses.createdAt", naming statuses_actorId_idx and why its trailing createdAt cannot serve a global ordering. The same wrong sentence in the PR description has been corrected too.

migrations/20260820000000_add_local_public_timeline_indexes.js
llun: Fixed in d295fd341. Reworded to say MySQL drops the predicate and the index degrades to a plain one over actors.id, which actors_id_unique already covers — redundant rather than merely larger, buying no local-actor selectivity — and that PostgreSQL and SQLite, the supported backends, both get the real thing.

lib/database/sql/timeline.ts
llun: Fixed in d295fd341 — added lib/database/sql/localPublicQueryShape.test.ts. It captures the SQL that actually reaches the driver via knex query event, so it tracks the production path rather than a helper exported for the test, and normalises identifier quoting so the assertions describe shape rather than dialect. Covers both the feed and the count. Verified it fails against exactly the revert you describe: simulating innerJoin + distinct fails both cases on "expected select distinct statuses.id... to contain exists (".

lib/database/sql/timeline.test.ts
llun: Fixed in d295fd341. The duplicate test now also asserts the bounded branch agrees with the unbounded one at an unreachable limit (1000), so the path the landing page actually calls is covered. You are right that limit: 1 could not distinguish — rows.length caps at 1 either way.

lib/services/timelines/localPublicCount.ts
llun: Duplicate of the thread just below — I posted this finding twice by accident. Same issue, fixed in d295fd341.

lib/database/sql/localPublicQueryShape.test.ts
```diff
@@ -0,0 +1,81 @@
...
+//
+// The semi-join in `localPublicStatusesQuery` is the entire performance fix:
+// it is what lets `LIMIT` bound the scan. Rewriting it back to
+// `innerJoin('recipients', …).distinct()` is behaviourally identical for every
+// other assertion in the suite — the counts stay right and the duplicate
+// regression test still passes, because DISTINCT dedupes too — while restoring
+// the plan that had this query at ~27s on the landing page. Without a test that
+// looks at the SQL itself, that revert lands silently and green.
+//
+// Asserting on `toSQL()` follows the precedent in `search.test.ts`.
```
llun: Round 2: the cited precedent is wrong. This test never calls toSQL() — it captures knex query events. search.test.ts really does use toSQL() on builders it constructs itself, which is a different technique, so a maintainer following this comment lands in the wrong file. The actual precedent for query-event capture is status.test.ts:1349-1392, which uses the same on/off pair. Comment corrected to point there, and to say why observing the driver beats asserting on a self-constructed builder.

lib/database/sql/localPublicQueryShape.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const { database, instance } = getTestSQLDatabaseWithInstance()
+
+  // Capture what actually reaches the driver, so this tracks the query the
+  // production path builds rather than a helper exported for the test's sake.
+  const capture = async (run: () => Promise<unknown>): Promise<string[]> => {
+    const statements: string[] = []
+    const listener = ({ sql }: { sql: string }) => {
+      statements.push(sql)
+    }
+    instance.on('query', listener)
```
llun, line 43: Round 2: SQLite-only harness under a comment that reasons about PostgreSQL. getTestSQLDatabaseWithInstance is hardcoded SQLite and ignores TEST_DATABASE_TYPE entirely — testUtils.ts flags this as a trap, since such a suite reports a clean run under the pg environment variables without ever opening a PostgreSQL connection. That made the double-quote strip dead code and the comment overclaim coverage of the very backend this fix targets. Switched to getTestDatabaseWithInstance (+ prepare()), matching fitnessRouteHeatmapTile.test.ts, so the query actually gets pinned on whichever backend the suite runs.

lib/database/sql/localPublicQueryShape.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  beforeAll(async () => {
+    await database.migrate()
+  })
+
+  afterAll(async () => {
+    await database.destroy()
+  })
+
```
llun, line 73: Round 2: toHaveLength(1) silently depended on an empty database — it held only because getStatusesByIds short-circuits on an empty id list, so seeding one local public status here would fire two more recipients reads per status during hydration and turn a shape regression into a confusing "expected 1, received 3". Now selects the statement instead of assuming it is the only one. Worth noting my first attempt at this fix matched on "from statuses", which was WORSE: a revert moves recipients into the FROM, so the primary regression started failing with "expected [] to have a length of 1" instead of naming the defect. Matching on both table names keeps the revert failing on the assertion that names it ("expected select distinct... to contain exists ("), while still excluding the hydration reads, whose column is statusId rather than statuses.

lib/database/sql/localPublicQueryShape.test.ts
llun: Fixed in 1c14142ed — comment now points at status.test.ts and explains why observing the driver beats asserting on a self-constructed builder.

lib/database/sql/localPublicQueryShape.test.ts
llun: Fixed in 1c14142ed — switched to getTestDatabaseWithInstance with prepare(), matching fitnessRouteHeatmapTile.test.ts. The quote-stripping is now live rather than dead, and the guard pins the query on whichever backend the suite runs against.

lib/database/sql/localPublicQueryShape.test.ts
llun: Fixed in 1c14142ed. Worth recording that my first attempt matched on "from statuses" and was worse: a revert moves recipients into the FROM, so the primary regression failed with "expected [] to have a length of 1" rather than naming itself. Matching both table names keeps the revert failing on "to contain exists (" while still excluding hydration reads, whose column is statusId rather than statuses. Verified both directions.

PR #1471, opened 2026-08-20
fix: return 404 instead of 500 for a malformed media id on PostgreSQL

llun:
## Summary

Fixes seven test failures that only appear under `TEST_DATABASE_TYPE=pg`. CI pins
SQLite, so both survived. They are two different problems that share one cause:
`medias.id` is a real `integer` column on PostgreSQL, while SQLite's dynamic
typing accepts strings.

### 1. A real production bug: 500 where a 404 belongs (`fix:`)

`getMediaByIdForAccount` passed `mediaId` straight into `where('medias.id', mediaId)`.
On PostgreSQL a malformed id is therefore an **error, not a miss**, so
`GET /api/v1/media/abc` answered **500** instead of **404**. The same path is
reachable from status creation via `getAttachmentsFromMediaIds`.

The sibling `getMediaByIdsForAccount` already handled the non-numeric case. That
idea is now a shared `toMediaRowId` helper — but review showed a bare `Number()`
guard, which is what the sibling used, is **not** sufficient. Three distinct input
classes still produced a 500 or a wrong row, each verified against a local
PostgreSQL 17 through the real knex/pg binding path:

| client sends | bare `Number()` guard | actual result |
| --- | --- | --- |
| `2147483648` | passes it through | `value out of range for type integer` → 500 |
| `1e21` | passes `1e+21` | `invalid input syntax for type integer: "1e+21"` → 500 |
| `0x10` | resolves to `16` | returns **media 16** — a different row than the id names, on *both* backends |

The last one is the sharp one: it would have been a regression introduced by this
PR, and it is the only member of the class observable on SQLite. So the helper is
shape-checked and range-bounded, not a bare coercion.

**This is partly a deliberate tightening on PostgreSQL, not purely a bug fix**, and
that is worth stating plainly. Measured against PostgreSQL 17 and SQLite through
this app's own drivers:

| spelling | PostgreSQL 17 | SQLite | after |
| --- | --- | --- | --- |
| `0012` | row 12 | row 12 | row 12 — unchanged |
| `12.0` | `invalid input syntax` | row 12 | row 12 — 500 → hit |
| `+12`, `' 12 '` | row 12 | row 12 | **404 — tightened** |
| `0x10` | row 16 | no match | **404 — tightened** |
| `1e3` | `invalid input syntax` | row 1000 | 404 |
| `2147483648` | `value out of range` | no match | 404 |
| `abc` | `invalid input syntax` | no match | 404 |

PostgreSQL has accepted non-decimal integer literals since 16, so `GET
/api/v1/media/0x10` really did return media 16 there. A client asking for `0x10`
did not ask for media 16, and Mastodon answers 404 for anything that is not a row
id — so the tightening is intended. `'12.0'` is the one spelling kept rather than
tightened away, because SQLite's `varchar` `attachments.mediaId` can hold ids in
that form.

Auditing the callers turned up two more methods directly reachable with a raw
client id that the brief didn't name:

| Method | Reachable from | Prior read guard? |
| --- | --- | --- |
| `getMediaByIdForAccount` | `GET`/`PUT` `/api/v1/media/[id]`, `DELETE /api/v1/accounts/media/[mediaId]`, `getAttachmentsFromMediaIds`, presigned completion | — |
| `updateMedia` | `PUT /api/v1/media/[id]` (description/focus only — that branch skips the early ownership read), `PUT /api/v1/statuses/[id]` via `media_attributes[][id]` | **no** |
| `deleteMediaForAccount` | `DELETE /api/v1/media/[id]` | **no** |
| `markMediaUploadVerified`, `deleteMedia` | presigned completion / `DELETE /api/v1/accounts/media/[mediaId]` | yes — guarded anyway so a future caller without one can't reintroduce the 500 |

`createMedia` also returned the driver's raw **number** as `Media.id` despite the
field being typed `string`. Handed to `createAttachment`, that lands in SQLite's
`varchar` `attachments.mediaId` as `'1.0'` via REAL→TEXT conversion, where
PostgreSQL stores a plain `1`. It now stringifies, matching `parseMediaRow` and the
declared type. This is behaviourally a no-op in production (every consumer already
stringified before the value went anywhere) — it is a type-lie fix, and it is what
made the strict guard viable.

### 2. Unrealistic fixtures (`test:`)

Six `updateNote` cases inserted `attachments.mediaId` values like `'old-media'`.
That column is deliberately `integer` on PostgreSQL and `varchar(255)` on SQLite —
`20260207223000_fix_attachments_media_id_type.js` returns early unless the client is
`pg` — so the two committed schema dumps **legitimately** disagree here. That
divergence is left alone; the fixtures now use per-test named constants holding
realistic auto-increment ids. The same sweep fixed the last two non-numeric media
fixtures that reach a real insert, in `getMastodonStatusEdits.test.ts`.

**No migration is added, so no schema dump is regenerated.** CI's Schema Dump Sync
passes.

### 3. Docs (`docs:`)

Both facts above were undocumented and both read as something else — a
numeric-column comparison looks safe until PostgreSQL sees it, and a per-column
difference between the dumps looks like drift. Recorded in `AGENTS.md`, the
`REVIEW.md` checklist, `CLAUDE.md`, and `.github/copilot-instructions.md`.

## Known bug found but deliberately NOT fixed here

`POST /api/v1/accounts/outbox` takes `PostBoxAttachment.id` as a bare `z.string()`
and passes it through `createNoteFromUserInput` to `createAttachment` unvalidated.
On PostgreSQL a malformed id fails that insert (`invalid input syntax for type
integer`, verified) **after** `createNote` has already committed the status row —
so the request 500s and leaves a status with no attachments.

It is real, but it is an input-validation fix on a different endpoint rather than
another instance of this lookup guard, and it causes none of the failures this PR
is scoped to. It is recorded in the code at `createAttachment` and in all three
docs rather than silently expanding this PR.

## Testing

TDD throughout: `media.test.ts > getMediaByIdForAccount > returns null when media
does not exist` was confirmed failing under pg first, with
`invalid input syntax for type integer: "non-existent-id"`, before any production
code changed. Every guard added since was watched failing for its own reason first
— including reverting the coercion to prove the 12 new route cases catch the 500,
and reverting the shape check to prove the alias test catches the wrong-row bug on
SQLite.

Against a throwaway **local** PostgreSQL 17 on `127.0.0.1:5432`:

| Run | Result |
| --- | --- |
| `TEST_DATABASE_TYPE=pg yarn test` | 818 files, 9805 tests passed (was 7 failing) |
| `yarn test` (SQLite, what CI runs) | 818 files, 9805 tests passed |
| `yarn run prettier --write .` → `yarn lint` → `yarn build` | clean (lint: 0 errors, 45 pre-existing warnings) |

`app/api/v1/media/[id]/route.test.ts` moved off the SQLite-only `getTestSQLDatabase`
onto `getTestDatabaseWithInstance`, which honours `TEST_DATABASE_TYPE`. It had been
passing 27/27 under `TEST_DATABASE_HOST=no-such-pg-host.invalid` without opening a
connection — exactly the trap AGENTS.md documents. It now fails with `ENOTFOUND`
against that host, which is how we know the 12 new route cases are real.

**One caveat worth stating plainly:** CI pins `TEST_DATABASE_TYPE=sqlite`, so apart
from the alias case, none of these guards can fail in CI. A revert of `toMediaRowId`
would ship green. Closing that needs a pg job in CI, which is an infra decision
outside this PR.

## Checklist

- [x] If migrations changed: BOTH schema dumps regenerated — n/a, no migration changed


lib/database/sql/media.ts
```diff
@@ -36,9 +36,44 @@ import { Attachment } from '@/lib/types/domain/attachment'
...
+// for type integer` this guard exists to prevent.
+//
+// A trailing all-zero fraction ('12.0') IS accepted, deliberately: it names row
+// 12 unambiguously, and it is the form existing SQLite deployments already hold
+// in `attachments.mediaId`. That column is `varchar` on SQLite, and a media id
+// bound as a JS number lands in it through SQLite's REAL->TEXT conversion as
+// '1.0'. Those rows resolve today, so rejecting the form would turn editing an
+// affected status into a 422. `createMedia` no longer produces such ids, but
+// the ones already written have to keep working.
+const toMediaRowId = (mediaId: string): number | null => {
```
llun, line 85: **The guard was too loose in two ways that each reproduce the very 500 this PR fixes.**

1. **No upper bound.** `medias.id` is PostgreSQL `int4`. Verified end-to-end through the real knex/pg binding path against a local PostgreSQL 17:

| client sends | old guard returned | PostgreSQL |
| --- | --- | --- |
| `2147483648` | `2147483648` | `value out of range for type integer` |
| `1e21` | `1e+21` | `invalid input syntax for type integer: "1e+21"` |

The second is the *identical* error string this guard exists to prevent — `pg` serializes numbers with `toString()`, and JS renders values at or above 1e21 exponentially.

2. **`Number()` accepts alternate spellings**, which was a regression introduced by the first commit rather than a pre-existing hole. `Number('0x10')` is 16, so `GET /api/v1/media/0x10` would have answered with **media 16** where every backend previously answered 404. Confirmed against SQLite directly: `WHERE id='0x10'` matches nothing. Same for `'0b101'`, `'0o17'`, `'1e3'`, `'+12'`, `' 12 '`.

Now shape-checked and range-bounded. This is also the only part of the class observable on SQLite, so it is the one piece CI can actually catch — the new `returns null for alternate spellings of a real media id` test builds its aliases from a freshly created row and fails on SQLite against the old helper.

lib/database/sql/media.ts
```diff
@@ -215,7 +254,12 @@ export const MediaSQLDatabaseMixin = (database: Knex): MediaDatabase => ({
       }
 
       return {
-        id: ids[0].id,
+        // `Media.id` is a string everywhere else (`parseMediaRow` stringifies
+        // it), and callers hand this straight to `createAttachment`. Returning
+        // the driver's raw number wrote it into SQLite's `varchar`
+        // `attachments.mediaId` as '1.0' via REAL->TEXT conversion, where
+        // PostgreSQL's integer column stored a plain 1.
+        id: String(ids[0].id),
```
llun, line 279: **Latent bug surfaced by tightening the guard: `createMedia` returned the driver's raw number as `Media.id` despite the field being typed `string`.**

Callers hand that straight to `createAttachment`. On SQLite `attachments.mediaId` is `varchar`, so a bound JS number lands there as **`'1.0'`** through REAL-to-TEXT conversion, where PostgreSQL's `integer` column stores a plain `1` — a silent cross-backend divergence. Confirmed:

```
sqlite> CREATE TABLE a (mediaId varchar(255)); INSERT INTO a VALUES (1.0);
sqlite> SELECT quote(mediaId), typeof(mediaId) FROM a;  -->  '1.0'|text
```

Those rows are read back and re-resolved on **every** status edit, so a strict digits-only guard turned editing an affected status into a 422 — it broke 8 existing tests. Two-part fix: `createMedia` now stringifies so no new ones are written, and the guard deliberately still accepts a trailing all-zero fraction so ids already in existing SQLite deployments keep resolving. `'12.0'` names row 12 unambiguously, so unlike the radix aliases it is safe to accept.

app/api/v1/media/[id]/route.test.ts
```diff
@@ -52,9 +53,16 @@ vi.mock('@/lib/config', () => ({
...
 
 describe('/api/v1/media/[id]', () => {
-  const database = getTestSQLDatabase()
+  // `getTestDatabaseWithInstance` honours TEST_DATABASE_TYPE; the SQLite-only
+  // `getTestSQLDatabase` does not, so under `TEST_DATABASE_TYPE=pg` this suite
+  // used to report a clean pass having never opened a PostgreSQL connection.
+  // It has to run on PostgreSQL for the malformed-id cases below to mean
+  // anything: `medias.id` is an integer column there, so a bad id is an error
+  // rather than a miss unless the query layer rejects it first.
+  const { database, prepare } = getTestDatabaseWithInstance()
```
Comment (llun) [inline app/api/v1/media/[id]/route.test.ts:62]: **The user-visible half of this bug — the 500 to 404 — had zero route-level coverage, and could not have any.**

This suite used `getTestSQLDatabase`, which AGENTS.md documents as SQLite-only and ignoring `TEST_DATABASE_TYPE`. Proven: it passed 27/27 under `TEST_DATABASE_TYPE=pg TEST_DATABASE_HOST=no-such-pg-host.invalid` without ever opening a connection — exactly the trap AGENTS.md warns about.

Switched to `getTestDatabaseWithInstance`, which honours the env var. Verified it now genuinely connects: the same unreachable-host run fails the suite with `ENOTFOUND` instead of passing it. The 12 new cases below (GET/PUT/DELETE across four malformed id shapes) then became a real guard — with the coercion removed they all fail with 500, and they pass on both backends with it.

lib/database/sql/media.test.ts
```diff
@@ -482,6 +575,79 @@ describe('MediaDatabase', () => {
       })
     })
 
+    describe('markMediaUploadVerified', () => {
```
llun, line 584: **A negative-only describe is satisfied by `() => null`.** `markMediaUploadVerified` had no SQL-level coverage anywhere — the only other references are in `S3StorageFile.test.ts`, which mocks the whole database and asserts call/not-call, so the ownership join, the `upload.state` rewrite and the persist were entirely unexercised.

Added a positive case that asserts the returned object *and* re-reads the row (the return value alone would pass against a method that never wrote), plus a cross-account case to pin the join.

lib/database/sql/media.test.ts
```diff
@@ -322,18 +322,111 @@ describe('MediaDatabase', () => {
     })
 
     describe('getMediaByIdForAccount', () => {
+      // A well-formed id that simply has no row — this must stay numeric so it
+      // exercises an actual database miss rather than short-circuiting in
+      // toMediaRowId, which is what the malformed ids below cover.
       it('returns null when media does not exist', async () => {
         const actor = await database.getActorFromId({ id: actors.primary.id })
         expect(actor).toBeDefined()
```
llun, line 331: **This test stopped testing what its name says.** It passed `'non-existent-id'`, which after the fix short-circuits in `toMediaRowId` before a query is ever built — so `returns null when media does not exist` no longer exercised a database miss at all, and had become a duplicate of the new non-numeric row.

Restored to a numeric sentinel so it exercises a real miss again; the malformed shapes are covered separately below. (The sibling describes for `deleteMedia`, `updateMedia` and `deleteMediaForAccount` already used numeric sentinels.)

AGENTS.md
```diff
@@ -1463,6 +1465,7 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
...very method that **compares** a `mediaId` against `medias.id` runs it through `toMediaRowId` first, so the caller reports "not found" without touching the database. (`createAttachment` **writes** `mediaId` rather than comparing it, and is not guarded — every production caller hands it an id read back out of `medias`.) The guard is shape-checked and range-bounded on purpose, not a bare `Number()`: `Number()` also accepts `'0x10'` (16), `'0b101'` (5), `'1e3'` (1000), `'+12'` and `' 12 '`, each of which resolves a _different_ row than the id names, while an id above `integer`'s 2147483647 raises `value out of range for type integer` — the same 500 in a new costume. A trailing all-zero fraction (`'12.0'`) is the one accepted alias: it names row 12 unambiguously and is the form existing **SQLite** deployments already hold in `attachments.mediaId`, because that column is `varchar` there and a media id bound as a JS number reaches it through SQLite's REAL→TEXT conversion. Rejecting it would 422 an edit of any affected status. Apply the same treatment to any new query that compares a caller-supplied value against a numeric column, and give test fixtures values the column can actually hold.
```
llun: **This claimed more than the code does.** The original wording said *every* method taking a `mediaId` string runs it through `toMediaRowId`. `createAttachment` takes `mediaId?: string` and inserts it raw — and it is the only place `attachments.mediaId` receives a string, which is precisely the column named as an integer PG column two sentences earlier, and precisely what made the six `status.test.ts` fixtures fail.

Not a bug today (every production caller hands it an id read back out of `medias`), but the doc named the column while the guard skipped its only string-write path. Reworded to "every method that **compares** a `mediaId` against `medias.id`", with `createAttachment` called out explicitly. Corrected in REVIEW.md and CLAUDE.md too, along with the "digits-only" wording once `'12.0'` became a deliberate exception.

lib/services/mastodon/getMastodonStatusEdits.test.ts
```diff
@@ -48,7 +52,7 @@ describe('getMastodonStatusEdits', () => {
       width: 320,
       height: 240,
       name: 'old alt',
-      mediaId: 'edits-old-media'
+      mediaId: '9701'
```
llun, line 55: **Documentation Maintenance sweep.** This PR adds a rule that fixtures use values the column can actually hold, so the rule needed applying repo-wide, not just to the six cases that happened to fail. These two were the only remaining non-numeric `mediaId` values that reach a real database insert.

This file is immune today only because it builds its own hardcoded better-sqlite3 instance instead of going through `testUtils` — the same latent trap, one harness switch away from breaking. (The other non-numeric `mediaId` literals in the repo — `post-box.test.tsx`, `S3StorageFile.test.ts`, `medias/presigned/route.test.ts` — are plain objects or fully-mocked databases and never reach SQL.)

llun (commented):
Round 1 of the AGENTS.md sub-agent review loop — three reviewers across correctness, tests/fixtures, and invariants/docs/security. Every finding below was verified against a local PostgreSQL 17 before acting on it, and all are now addressed on the branch.

Also verified clean and worth recording: no authorization decision changes (every guarded query keeps its `actors.accountId` predicate, and `getAttachmentsFromMediaIds` returning null still 422s rather than silently dropping an attachment); the `getMediaByIdsForAccount` refactor is input-for-input equivalent to the code it replaced; no caller depended on the old throw; and no other route compares a raw request value against a numeric column (poll `choice` is Zod-bounded, everything else is a server-derived surrogate).

lib/database/sql/media.ts
```diff
@@ -275,6 +329,15 @@ export const MediaSQLDatabaseMixin = (database: Knex): MediaDatabase => ({
...
+  // NOTE: `mediaId` is WRITTEN here, not compared, so it does not go through
+  // `toMediaRowId` — coercing would silently drop the link rather than surface
+  // the caller's bad id. Almost every caller hands over an id read back out of
+  // `medias`, but `POST /api/v1/accounts/outbox` does not: its
+  // `PostBoxAttachment.id` is a bare `z.string()` that reaches
+  // `lib/actions/createNote.ts` unvalidated, so a malformed id fails this
+  // insert on PostgreSQL (`attachments.mediaId` is `integer` there) AFTER the
+  // status row is already committed. That endpoint needs its own validation —
+  // it is a separate bug from the lookup guard above, tracked separately.
   async createAttachment({
```
llun, line 351: **Round 2 finding — the claim the previous commit added here was false, and there is a real 500 behind it.**

The docs asserted `createAttachment` is safe because "every production caller hands it an id read back out of `medias`". Not true of one live caller:

- `app/api/v1/accounts/outbox/types.ts:18` — `attachments: PostBoxAttachment.array()`, and `UploadedAttachment.id` is a bare `z.string()` (`lib/types/domain/attachment.ts:19`). No validation against `medias`.
- `app/api/v1/accounts/outbox/route.ts:109` → `createNoteFromUserInput`
- `lib/actions/createNote.ts:566` → `mediaId: attachment.id` → this insert.

Verified against local PG 17 that a parameterised insert into an `integer` column fails:

```
mediaId="abc" -> ERROR: invalid input syntax for type integer
mediaId="1.0" -> ERROR: invalid input syntax for type integer
mediaId="12"  -> inserted
```

Worse than a plain 500: `createNote.ts` commits the status row before the attachment `Promise.all`, so the request fails having already persisted a status with no attachments. `POST /api/v1/statuses` and `PUT /api/v1/statuses/:id` are genuinely safe — both route through `getAttachmentsFromMediaIds`.

**Deliberately not fixed in this PR.** It is an input-validation fix on a different endpoint (the right answer is a 422 from the outbox route, not coercion here — coercing would silently drop the link rather than surface the caller's bad id), and it causes none of the failures this PR is scoped to. Recorded here and in all three docs instead of silently widening the change; the PR description calls it out explicitly.

llun (commented):
Round 2 of the review loop — two reviewers, this time pointed at **the round-1 fixes themselves**, since that is where fixes tend to introduce new problems. That paid off: the two substantive findings are both about claims round 1 added, one of which was hiding a real 500.

Verified clean this round, so it does not need re-checking: the regex is ReDoS-free (1M-char adversarial inputs in <9ms) and not subject to the JS `$`-before-newline trap (that is Python; `'12\n'` is rejected); `createMedia` returning a string is a no-op for every production consumer and changes no API response body; there is only one `MediaDatabase` implementation, so no two backends can disagree; the harness switch cannot destroy another suite's database (per-worker naming, one file at a time — confirmed by running three pg suites under `--no-file-parallelism`); and the alias test is row-id independent.

The PR title and description have been rewritten — round 2 correctly flagged the body as stale in five places, since it still described the round-0 `Number()` guard and claimed route tests could not cover this.

lib/database/sql/media.ts
```diff
@@ -36,9 +36,51 @@ import { Attachment } from '@/lib/types/domain/attachment'
...
+//
+// The shape test matters as much as the range check. `Number()` alone also
+// accepts '0x10' (16), '0b101' (5), '1e3' (1000), '+12' and ' 12 ', each of
+// which resolves a DIFFERENT row than the id names — `GET /api/v1/media/0x10`
+// would answer with media 16 where every backend previously answered 404. And
+// '1e21' round-trips back to PostgreSQL as the string '1e+21' (the driver
+// stringifies with `toString()`), reproducing the exact `invalid input syntax
+// for type integer` this guard exists to prevent.
+//
+// What IS accepted, beyond a bare decimal, is any spelling that still names the
```
llun: **Round 2 finding — the rationale for accepting `'12.0'` was unsupported, and I have corrected it rather than quietly leaving it.**

The previous commit justified the allowance as compatibility with ids "existing SQLite deployments already hold". Nothing produces that form:

- Every `createAttachment` caller stringifies (via `parseMediaRow`'s `String()` or `MediaStorageSaveFileOutput.id`, which is `z.string()`).
- The one bulk writer, the #307 backfill, copies an INTEGER — SQLite's TEXT affinity renders that as `'1'`, not `'1.0'`. Verified by running it:
  ```
  UPDATE attachments SET mediaId = (SELECT id FROM medias …)  ->  '1'|text
  ```
- Only a *bound JS number* becomes `'1.0'` — which is exactly what the un-stringified `createMedia` return did, and what the previous commit fixed.

The allowance is kept, because the form names the same row and rejecting it would be a behaviour change rather than a fix — but it is now described as defence in depth, not as a shim for data that does not exist.

The same passage also claimed a trailing all-zero fraction was "the one accepted alias". It is not — leading zeros (`'0012'`) are accepted too, and both resolved before this guard existed on both backends. The tolerance test now covers each.

lib/database/sql/status.ts
```diff
@@ -722,6 +722,11 @@ export const StatusSQLDatabaseMixin = (
             existingMediaIds.has(attachment.id) &&
             (attachment.name ?? '') !== existingNameByMediaId.get(attachment.id)
         )
+        // `attachments.mediaId` is an `integer` column on PostgreSQL, so a
```
llun, line 725: **Round 2 finding (low, consistency).** This is the one remaining place in the codebase matching the exact pattern the PR's own new AGENTS.md rule tells reviewers to guard: a caller-supplied string compared against `attachments.mediaId`, which is `integer` on PostgreSQL.

It is unreachable with a bad value — every `attachment.id` in `keptAttachments` has already resolved through `getMediaByIdForAccount` — but a reader following the new rule would have to re-derive that. Added the note rather than adding a redundant guard.

.github/copilot-instructions.md
```diff
@@ -44,6 +44,7 @@ Always refer to and follow the guidelines in [AGENTS.md](../AGENTS.md) for all c
 - All database operations must work with SQLite and PostgreSQL, and should avoid assumptions that break MySQL-compatible Knex clients where possible.
 - Use Knex query builder; avoid raw SQL unless necessary.
 - Avoid database-specific features unless they include backend-specific fallback logic.
+- Coerce a client-supplied id before comparing it against a numeric column. `medias.id` is `integer` on PostgreSQL, so a malformed id raises `invalid input syntax for type integer` (or `value out of range`) instead of simply missing — a 500 where a 404 was intended. SQLite's dynamic typing hides this, and CI pins `TEST_DATABASE_TYPE=sqlite`, so only a `TEST_DATABASE_TYPE=pg` run catches it. See `toMediaRowId` in `lib/database/sql/media.ts`, and give test fixtures values the column can actually hold.
```
llun, line 47: **Round 2 finding — Documentation Maintenance sweep was incomplete.**

`AGENTS.md` requires rule changes to be mirrored into the pointer files. Round 1 updated `CLAUDE.md` but missed this one, which has its own **Database Operations** section and existing precedent for mirroring specific conventions. Added the coercion rule here.

(`GEMINI.md` is a short key-gates list that carries no per-convention detail, so it correctly stays as-is. Separately: `AGENTS.md`'s pointer list names `.cursor/rules/agents.mdc`, which does not exist in the checkout — pre-existing and left alone rather than guessed at.)

lib/database/sql/media.ts
```diff
@@ -36,9 +36,61 @@ import { Attachment } from '@/lib/types/domain/attachment'
...
+
+// `medias.id` is an integer column. Mastodon clients send ids as strings and
+// put whatever they like in them, so coerce before comparing: PostgreSQL
+// rejects an integer column compared against text with `invalid input syntax
+// for type integer`, which turns a miss into a 500 rather than a 404. SQLite's
+// dynamic typing merely matches nothing, which is why this only ever showed up
+// on PostgreSQL. Returns null for anything that is not a plain decimal row id
+// so the caller can report "not found" without touching the database.
+//
+// Accepted is exactly: optional leading zeros, one or more digits, an optional
```
llun, line 53: **Round 3 finding — my own round-1/2 descriptions of this guard were wrong in three ways, all in the same direction: they understated it as a pure bug fix.**

Measured against PostgreSQL 17 and SQLite through this app's own drivers (not psql):

| spelling | PostgreSQL 17 | SQLite | now |
| --- | --- | --- | --- |
| `0012` | row 12 | row 12 | row 12 — unchanged |
| `12.0` | `invalid input syntax` | row 12 | row 12 — 500 → hit |
| `+12`, `' 12 '` | row 12 | row 12 | **404 — tightened** |
| `0x10` | **row 16** | no match | **404 — tightened** |
| `1e3` | `invalid input syntax` | row 1000 | 404 |
| `2147483648` | `value out of range` | no match | 404 |

What was wrong:

1. *"'12.0' resolved before the guard existed — PostgreSQL casts it."* It does not; PG raises `invalid input syntax for type integer`. Of the two tolerated spellings only `'0012'` resolved there.
2. *"`GET /api/v1/media/0x10` would answer media 16 where every backend previously answered 404."* Inverted. PostgreSQL has accepted non-decimal integer literals since 16, so it **already** answered media 16 — only SQLite 404'd.
3. *"each of which resolves a DIFFERENT row than the id names"* — false of `'+12'` and `' 12 '`, which name the same row on both backends, and which the guard now **rejects**.

So this is partly a deliberate **tightening on PostgreSQL**, not purely a bug fix. The tightening is still right — a media id is a row id, and a client asking for `0x10` did not ask for media 16 — but describing it as "these never worked anyway" would mislead the next reader. Replaced with the measured table above, in the comment and all three docs, and disclosed in the PR description.

Round 3 also caught that the documented acceptance *principle* ("spellings that still name the same row stay accepted") did not describe the guard: `'+12'` satisfies it and is rejected, `'12.0'` violates it on PG and is accepted. A future agent applying that principle to a new numeric-column guard would build a looser one. The docs now state the guard's actual rule.

llun (commented):
Round 3 — convergence check. **No functional defects found**: the code and tests are correct, and the reviewer independently re-ran both suites (818 files / 9805 tests on each backend) and confirmed the PR description's claims.

One finding, and it is about my own round-1/2 doc corrections rather than the code — the same class of overstatement those rounds existed to remove. Details inline; the short version is that the guard is partly a deliberate tightening on PostgreSQL and I had described it as purely a bug fix. Fixed in `cf8048b5`, along with a sentence CLAUDE.md had ended up stating twice.

Verified correct and needing no action: the outbox claim (traced end to end, including "after the status row is committed"); `POST /api/v1/statuses` and `PUT /api/v1/statuses/[id]` being safe; the `status.ts` comment; "every method that compares" covering all six methods; every assertion in `.github/copilot-instructions.md`; and the new tolerance test being collision-free and genuinely regression-catching.

**One flake to note honestly:** a full pg run hit `fitnessRouteHeatmapTile.test.ts > refuses a claim whose build was taken over between its read and its write`. It is a compare-and-swap race test in a file this PR does not touch; it passed 4/4 in isolation, the full pg suite is green on re-run, and it also passes with this branch's changes stashed. Pre-existing pg-only flakiness, not caused here — recording it rather than quietly re-running until green.

PR #1475, opened 2026-08-21
minor: draw the route heatmap from tiles, so zooming in reveals street detail

llun:
Phase 6 of the route heatmap precision work, on top of #1467. This is the first change in the series a user can see.

Pan or zoom a heatmap and the map now fetches tiles for what it is looking at, so a world view stays coarse and a street view is drawn at street precision. Before, both were the same geometry — simplified once to fit one global vertex budget and reused at every zoom, which is why zooming in never revealed more road.

> **Status: not finished.** Component tests for the two tiled render paths and the browser verification with before/after screenshots are still outstanding — see the checklist at the bottom. Opening now for visibility; the review loop starts once those land.

## What this adds

- **`useHeatmapTiles`** — the whole of the client logic. Picks the smallest ladder rung at least as detailed as the view, fetches what it does not already hold, caches by `(version, z, x, y)`, and hands back drawable runs.
- **`heatmapTileGeometry`** — places a decoded tile's runs on the globe and builds the GeoJSON the GL layer draws. A *sibling* of `buildRouteGeoJson`, not a widening of it: the untiled blob has no visit count and never will, so giving its features one would mean inventing it.
- **GL + MapKit rendering** driven by visit count, and **`PublicRouteHeatmapMap`** so both public surfaces can bind a share token on the client.

## Three things measured rather than assumed

**A constant's comment was wrong, and I wrote it.** `MAX_TILES_PER_REQUEST` claimed a viewport "fits in a single round trip with headroom (~64 tiles)". The ladder steps by 2, so a view just past a rung rounds up nearly two levels:

| Surface | View zoom | Stored zoom | Tiles |
| --- | --- | --- | --- |
| In-app (`h-[420px]`) | 10.01 | 12 | **160** |
| Embed (`h-dvh`) | 10.01 | 12 | **300** |
| Whole world | 4 | 4 | **256** |

Batching is the normal path, not a safety net. Comment corrected.

**Two bounds interlock, and getting them backwards is silent.** The cache must hold *more* than one view. Tiles are inserted batch by batch while a view loads and eviction takes the oldest first — so a cache smaller than a view would evict the batches it had already fetched and assemble a map with holes it has no way to notice. Found by sizing a test fixture, not by reasoning.

**A rejected tile fetch escaped as an unhandled rejection** until a test surfaced it. Tiles are detail on top of geometry the surface already drew, so a failure now leaves the previous view standing.

## Design notes

- Tiles **replace** the untiled geometry rather than drawing over it — the two describe the same roads at different fidelities, so together every line renders at twice its opacity — and only once a batch resolves, so a pan never blanks the map.
- The GL paint **generates its opacity stops from `heatOpacityForCount`** rather than copying numbers out of it, so the client ramp cannot drift from the one the server documents. MapKit has no data-driven styling, so the same ramp becomes a small fixed set of styles; `heatOpacityForCount` saturates, so there are twelve, built once and shared.
- MapKit has no overlay identity and **no zoom at all**: the per-tile keying is ours, and the zoom is derived from its region and element size. Its framing now happens once per map — assigning `region` fires `region-change-end`, which now drives a fetch, so re-framing on every geometry change would have been a loop.
- Both public surfaces render from Server Components, which **cannot pass a function across the RSC boundary**, so `PublicRouteHeatmapMap` binds the token on the client for both.
- Initial framing still comes from the untiled payload — tiles cannot be requested until a viewport exists.

## Verification so far

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (**838 files / 10,414 tests**), rebased onto current `main`.

Five mutants against the hook, all killed: `storedZoomForView` rounds down; cache never consulted; per-view ceiling removed; debounce reduced to zero; and the tile-geometry conversions.

A local dataset is seeded for the browser pass — 40 activities on deliberately overlapping routes, a completed pyramid of **219 tiles** across the ladder (z4: 7 points → z16: 32,626), with visit counts up to 30, which saturates the colour ramp.

## Remaining

- [ ] Component tests for the GL tiled path and the MapKit per-tile diff
- [ ] Browser verification + before/after screenshots at world and street zoom
- [ ] Sub-agent review loop to clean


llun:
## Review round 3 — 15 raised, 6 survived; critic clears the merge

Scoped to **defects only** this round, with coverage reportable solely where a *plausible* mutant survived and a user could perceive the difference. That distinction is what ended the loop: round 2's 18 survivors were 5 defects and 13 coverage ideas, and treating both as blocking made the process look bottomless.

### The one defect, found independently by two reviewers

A raw longitude span within one column-width of a whole world — 337.5° to 360° at the coarsest rung — normalises into the **same tile column at both ends**. Reading coverage from that pair alone answered "one column", so the view fetched one sixteenth of the world; and because those tiles came back non-empty, the untiled geometry was blanked. The reader saw a single 22.5° strip and nothing else, with no error.

Column coverage is now the larger of what the normalised pair implies and what the raw span needs, stepped with modulo so the antimeridian wrap and this case are one rule. Pinned; the mutant that reverts to the normalised pair fails.

Narrower in practice than it first appears — the surviving column is always the antipode of the view centre, which after an unfocused `fitBounds` holds none of the actor's routes — but it is exactly the "assemble with holes nothing downstream can detect" failure this file's comments say must not happen.

### The critic verified round 2's two blockers by running code

> **(1)** "The hook could draw one place's streets while the map looked at another": I held a fetch open across both transitions that can cause it. Superseding — Tokyo's batch held open, an Amsterdam view loads and paints, Tokyo's released afterwards — left `runs` at the SAME array identity, no repaint. Abandoning left `runs: []` and `zoom: null`. […] The class is also structurally closed: `tileRunsToLngLat(z, x, y, …)` places every run from its own tile coordinates, so a tile cannot be drawn anywhere but where it belongs; the worst available failure is drawing the wrong SET, never the wrong PLACE.
>
> **(2)** "The zoom maths was unverified on both renderers": I re-derived both. GL's world is 512·2^z px against the pyramid's 256px grid, so `+1` is exact […] MapKit's `log2(widthPx·360/(256·span))` is the Mercator definition verbatim, pinned numerically […] and again with a 0.07° scope chosen so the coarsening loop cannot walk a wrong rung back onto the right one.

### Deliberately not fixed here

Five coverage findings remain, none indicating a known-wrong behaviour: the version-0 fallback and `evict`'s protected-key guard start from states their tests don't reach; MapKit's pan diff is asserted on object identity rather than geometry, so dropping either half of the diff survives; and `abandon()`'s abort is unpinned. They are real gaps and will be filed as a follow-up — but after three rounds and sixteen fixed defects, more test-writing is not what stands between this and shipping.

### Verification

Full gate green: `prettier --write .` → `lint` (0 errors) → `build` → `test` (**839 files / 10,448 tests**). Rebased onto current `main`, CI 20/20, `MERGEABLE CLEAN`.

| Round | Raised → survived | Worst |
| --- | --- | --- |
| 1 | 35 → 30 | **high** — wrong ladder rung on both renderers, frozen maps, blank MapKit rebuilds |
| 2 | 29 → 18 (5 defects) | **high** — unaborted repaints, 202 MB cache, no tiles at the antimeridian |
| 3 | 15 → 6 (1 defect) | **medium** — near-world view collapsed to one column |

Merging.

llun:
## Follow-up: five deferred test-coverage gaps

Issues are disabled on this repo, so recording them here where they stay findable. None indicates a known-wrong behaviour — each is a case where a plausible mutation of *correct* code survives the suite. Each was reproduced by a review agent actually running the mutant.

**1. `abandon()`'s abort is unpinned** — `lib/components/fitness/useHeatmapTiles.ts`
Removing `abortRef.current?.abort()` lets an in-flight batch land after a view is abandoned and repaint it. This was a **shipped defect fixed in this PR**, so it is silently revertible — the highest-value gap of the five.
*Test:* hold the first fetch pending, report a ≥360° span, assert `runs` empty, resolve the held promise, assert `runs` still empty and `zoom` still null.

**2. The `version: 0` fallback starts from an empty view** — same file
The existing test cannot see the difference because there is nothing to clear.
*Test:* resolve a view at the current version, then answer `version: 0` for a new viewport; assert `runs` empties.

**3. `evict`'s protected-key guard is unpinned** — same file
Dropping the skip lets a dense view evict the tiles it is assembling from — exactly the holes the tile-count floor exists to prevent.
*Test:* one view whose decoded tiles exceed `TILE_CACHE_MAX_VERTICES` (~80 tiles × ~12k vertices); assert `groups` has one entry per requested tile.

**4. MapKit's per-tile diff is asserted on object identity** — `lib/components/fitness/RouteHeatmapMapKit.tsx`
Both halves can be dropped with the test still green: removing the "already held" skip duplicates every retained overlay; removing the removal half leaks every tile ever visited.
*Test:* assert on geometry (`JSON.stringify(overlay.points)`) rather than identity — no duplicates after a pan — and add a pan to a **disjoint** region asserting every pre-pan overlay is gone.

**5. MapKit's tiled privacy-hidden style branch is untested** — same file
The hidden class keeps blue with its own opacity ramp; nothing exercises it on the tiled path.

PR #1489, opened 2026-08-21
minor: add a link preview card to the public heatmap share page

llun:
## Summary

`/u/heatmaps/<token>` published **no** OpenGraph tags at all — the whole head was `<title>Route heatmap</title>` plus a `robots` noindex. Pasting a share link anywhere (Mastodon, X, Slack, Discord, or this service's own timeline) produced a bare URL.

It now emits OpenGraph and Twitter card tags from `generateMetadata`.

## Why `format=png` is load-bearing, not a nicety

The `og:image` reuses the existing `GET /embed/heatmap/<token>/image` rather than adding a fifth public surface — so the share guards that route already enforces (completed-only, `resolveSharedHeatmapRegionBounds`, `buildHeatmapTileSource` gating) are the ones that apply, with nothing new to keep in sync.

But that route falls through to the **keyless SVG renderer** on any instance with neither a Mapbox token nor an Apple key. Checked against production before writing anything:

```
$ curl -sI 'https://llun.social/embed/heatmap/<token>/image?w=1200&h=600'
content-type: image/svg+xml; charset=utf-8
```

No card crawler renders SVG, so an `og:image` pointing at one yields a card with **no image** — the tags would have looked correct and produced nothing. A new opt-in `?format=png` rasterizes that fallback through `sharp` (the two basemap renderers already answer with raster bytes, so it costs those paths nothing).

Opt-in because the SVG scales and the embed snippet the share dialog hands out already points at the same URL. Anything but `png` (trimmed, case-insensitive) takes the default, so the parameter collapses onto two cache variants rather than widening the surface `DIMENSION_STEP` exists to bound. A rasterization failure degrades to the SVG rather than 500ing — a browser still draws it, and the crawler was going to ignore the image either way.

The declared `1200x600` is what the route will actually serve: each axis snaps to a step of 100, so a requested 630 would come back as 600. Only the Apple path differs (1280x640, same 2:1 ratio).

Both callers build that URL through one `buildHeatmapEmbedImageUrl` (`lib/fitness/heatmapEmbedImageUrl.ts`) — the share dialog in the browser and the card on the server — so the route's query shape has a single owner rather than two implementations free to drift. `HeatmapShareEmbed`'s own tests are unmodified and still pass, which is the evidence its URL is byte-identical.

## One loader behind both surfaces

`generateMetadata` and the page body share one `cache()`-memoized `loadSharedHeatmap`, so the card cannot describe a share the page refuses — a second lookup path would be free to drift — and Next runs the reads once per request instead of once per surface.

A revoked token, a share re-queued for generation, and an unresolvable region all 404 and get the bare noindex metadata with **no card**: an `og:title` for one would publish the region name, the owner's handle and a thumbnail that the refusal exists to withhold.

The page keeps `noindex, nofollow` — the token is the secret. That costs the card nothing: crawlers read OpenGraph without consulting robots, this instance's own `parseOpenGraph` included.

The page's bbox caption is deliberately left out of `og:description`; as a card subtitle `TL 52.50°N 4.80°E → BR 52.30°N 5.00°E` reads as debug output, and the map locates the share better than its corners do.

## Verification

Local SQLite instance with a seeded completed+shared heatmap, dev server on `:3010`. Rendered head:

```html
<title>Route heatmap — Amsterdam</title>
<meta name="robots" content="noindex, nofollow"/>
<meta property="og:title" content="Route heatmap — Amsterdam"/>
<meta property="og:description" content="testuser (@testuser@localhost:3010) · Generated August 21, 2026"/>
<meta property="og:url" content="https://localhost:3010/u/heatmaps/localsharetoken123"/>
<meta property="og:site_name" content="Activities.next"/>
<meta property="og:image" content=".../image?w=1200&amp;h=600&amp;format=png"/>
<meta property="og:image:width" content="1200"/>
<meta property="og:image:height" content="600"/>
<meta name="twitter:card" content="summary_large_image"/>
```

The image endpoint over that same server:

```
content-type: image/png
cache-control: public, max-age=60, s-maxage=60
access-control-allow-origin: *
41168 bytes · PNG magic ok · 1200x600
```

And the round trip through **this service's own** link-preview parser (`parseOpenGraphMetadata` run against the live page):

```json
{
  "type": "link",
  "title": "Route heatmap — Amsterdam",
  "description": "testuser (@testuser@localhost:3010) · Generated August 21, 2026",
  "siteName": "Activities.next",
  "imageUrl": ".../image?w=1200&h=600&format=png",
  "imageWidth": 1200,
  "imageHeight": 600
}
```

The share page itself was checked in a browser and renders unchanged (heading, owner line, bbox caption, pannable map), before and after the loader refactor.

## Tests

- `lib/fitness/heatmapEmbedImageUrl.test.ts` (new) — the shared builder, with and without `format`, and the trailing-slash origin.
- `app/u/heatmaps/[token]/sharedHeatmapMetadata.test.ts` (new) — title/description/image/twitter shape, noindex retained, blank `og:site_name` dropped, bbox excluded, and `UNRESOLVED_SHARE_METADATA` carrying no card.
- `app/u/heatmaps/[token]/page.test.tsx` — `generateMetadata` builds a card for a renderable share and publishes none for a revoked token, a re-queued share, or an unresolvable region; both surfaces agree on the same share.
- `app/embed/heatmap/[token]/image/route.test.ts` — `format=png` returns real PNG bytes at the requested size, the trim/case normalization is covered, unknown and blank values still serve SVG, and a rasterization failure degrades to SVG and records it.

Full gate green: prettier → lint → typecheck → build → test (844 files, 10511 tests).

## Review

Three rounds of the sub-agent review loop. Round 1 raised three findings, all fixed in `5237557e8`: a doc sentence that contradicted the code, `cardImageUrl` on the wrong view model (it added a fresh `TS2741` to an already-ratcheted test file), and the duplicated URL template. Rounds 2 and 3 were clean on code; round 3 caught this description being stale, and one untested normalization branch, fixed in `f848fbd83`.

## Docs

`AGENTS.md` (Fitness Route Heatmap Pyramid), `CLAUDE.md`, `docs/architecture.md`, and `docs/fitness-file-storage.md` (new "The link-preview card" subsection).

## Note

No migration, no schema change, no new environment variable, no new dependency (`sharp` was already used for media).


docs/fitness-file-storage.md
```diff
@@ -168,6 +168,42 @@ itself, and an instance keeps its basemap instead of trading it away for
...
+takes the default, so the parameter collapses onto two cache variants instead
+of widening the surface `DIMENSION_STEP` exists to bound, and a rasterization
+failure degrades to the SVG rather than 500ing — a browser still draws it, and
+the crawler was going to ignore the image either way. The declared `1200x600`
+is what the route will actually serve, since each axis snaps to a step of 100
+and a requested 630 would come back as 600; only the Apple path differs, at
+1280x640 for the same 2:1 ratio.
+
+Everything the card says — the region title, the owner's name and handle, the
+generated date, the bounding box — comes from the same `buildSharedHeatmapView`
```
llun: **Doc contradicts the code it describes, and this PR added both.**

This sentence lists "the bounding box" among the things the card says:

> Everything the card says — the region title, the owner's name and handle, the generated date, the bounding box — comes from the same `buildSharedHeatmapView` the page renders

But `sharedHeatmapMetadata.ts` deliberately excludes it, with a comment saying why ("as a card subtitle `TL 52.50°N 4.80°E → BR 52.30°N 5.00°E` reads as debug output"), and there is a dedicated regression test asserting the description does **not** contain `view.bboxLabel`.

Failure scenario: someone reads this durable reference doc in isolation, believes bbox coordinates are exposed in the OG card, and either "fixes" the code to match — undoing a deliberate decision that has its own test — or wrongly reasons about what a share link discloses. `docs/architecture.md` got the parallel wording right; only this line is stale.

app/u/heatmaps/[token]/sharedHeatmapView.ts
```diff
@@ -46,6 +64,12 @@ export interface SharedHeatmapView {
   owner: SharedHeatmapOwner
   generatedLabel: string
   publicUrl: string
+  /**
+   * Static image for the link-preview card. `format=png` is what makes it
+   * usable: the keyless renderer answers with SVG, which every card crawler
+   * refuses.
+   */
+  cardImageUrl: string
```
Comment (llun) [inline app/u/heatmaps/[token]/sharedHeatmapView.ts]: **`cardImageUrl` as a required field on `SharedHeatmapView` adds a NEW type error to an already-ratcheted test file, and puts a card-only concern on the page's view model.**

Two problems from one field:

1. `yarn tsc --noEmit -p tsconfig.json` now reports a fresh `TS2741: Property 'cardImageUrl' is missing in type ... but required in type 'SharedHeatmapView'` at `app/u/heatmaps/[token]/SharedHeatmapPage.test.tsx:15`, whose fixtures build `SharedHeatmapView` literals. It does not fail CI only because that file is already in the `tsconfig.typecheck.json` ratchet for an unrelated pre-existing gap (`mapProvider`/`token`). AGENTS.md describes that list as a ratchet to be emptied, never grown — this quietly grows it, and the next person trying to remove that file hits two gaps instead of one.

2. Every other field on `SharedHeatmapView` (`title`, `isWorld`, `bboxLabel`, `owner`, `generatedLabel`, `publicUrl`) is read by `SharedHeatmapPage` for the page's own render. `cardImageUrl` is read only by `sharedHeatmapMetadata.ts`, in a module whose docstring calls itself "the read-only view model for the public shared heatmap page".

Both go away if the card image URL is built in `buildSharedHeatmapMetadata` from `origin` + `token` instead, leaving the view strictly about what the page renders. (`publicUrl` genuinely belongs on the view — the page shows it in its Copy link affordance.)

app/u/heatmaps/[token]/sharedHeatmapView.ts
```diff
@@ -109,8 +136,10 @@ export const buildSharedHeatmapView = ({
       initials: computeInitials(ownerName)
     },
     generatedLabel: formatGeneratedDate(heatmap.updatedAt),
-    // Drop any trailing slash so a base like `https://host/` can't yield `//`.
-    publicUrl: `${origin.replace(/\/+$/, '')}/u/heatmaps/${token}`,
+    publicUrl: `${base}/u/heatmaps/${token}`,
+    cardImageUrl:
```
Comment (llun) [inline app/u/heatmaps/[token]/sharedHeatmapView.ts]: **Second independent implementation of the embed-image URL template.**

This builds `${base}/embed/heatmap/${token}/image?w=…&h=…&format=png` after `const base = origin.replace(/\/+$/, '')`. `lib/components/fitness/HeatmapShareEmbed.tsx:199-203` already does the same thing — same route, same query shape, same trailing-slash normalization, down to a near-identical comment.

Failure scenario: the route's query contract changes (a renamed parameter, a new required one, different snapping) and only one caller is updated. This repo has been bitten by exactly this — AGENTS.md records two tile-query parsers that drifted until `?z=4&z=16` resolved differently on each route, with the note that "a divergence between two routes that must agree is how one becomes a leak the next time either grows a scope-bearing parameter."

Worth a single dependency-free `buildHeatmapEmbedImageUrl({ origin, token, width, height, format })` that both the client share dialog and the server-side card import, so the route's URL shape has one owner.

docs/fitness-file-storage.md
llun: Fixed in 5237557e8. The sentence now stops at the generated date and states explicitly that the bounding-box caption is deliberately excluded, with the same reasoning the code comment and the regression test carry. Good catch — the doc paragraph was written before the bbox was dropped from the description and I did not revisit it.

app/u/heatmaps/[token]/sharedHeatmapView.ts
Comment (llun) [reply app/u/heatmaps/[token]/sharedHeatmapView.ts]: Fixed in 5237557e8. `cardImageUrl` and the `HEATMAP_CARD_IMAGE_*` constants are off `SharedHeatmapView` entirely. `buildSharedHeatmapMetadata` now takes `origin` + `shareToken` and builds the card's image URL itself, so the view is strictly what the page renders; `publicUrl` stays on it, since the page shows that in its Copy link affordance.

Confirmed the type error is gone: `yarn tsc --noEmit -p tsconfig.json` reports zero `cardImageUrl` errors now (it reported `TS2741` at `SharedHeatmapPage.test.tsx:15` before). The pre-existing `TS2739` (`mapProvider`, `token`) errors in that file are untouched — they are what the ratchet entry was for.

app/u/heatmaps/[token]/sharedHeatmapView.ts
Comment (llun) [reply app/u/heatmaps/[token]/sharedHeatmapView.ts]: Fixed in 5237557e8. Added `lib/fitness/heatmapEmbedImageUrl.ts` — dependency-free, so both sides can import it — and routed both callers through it: `HeatmapShareEmbed.tsx` in the browser and `buildSharedHeatmapMetadata` on the server. `format` is optional there, so the share dialog keeps asking for the route's default and the card asks for `png`.

`HeatmapShareEmbed.test.tsx` was not modified and still passes, which is the evidence that the dialog's URL is byte-identical. The new helper has its own test covering both forms and the trailing-slash case. `docs/fitness-file-storage.md` and `AGENTS.md` now record that the route's query shape has a single owner.

PR #1493, opened 2026-08-22
fix: cut the ActivityPub outbox status count query roughly in half

llun:
## Why this query does not use the `counters` table

The trace in Query insights (`query_hash=7017339302753509205`, avg **622 ms**, 61 calls / 7 days, 1 row returned) is
`getActorStatusesCount({ actorId, publicOnly: true })` — the `totalItems` of the ActivityPub outbox
`OrderedCollection` served by `GET /api/users/[username]/outbox` with no `page` param. It is a public,
unauthenticated endpoint, so every remote crawl pays for it.

`getActorStatusesCount` **does** use the counter table — but only on the other branch:

- `publicOnly: false` → `getCounterValue(CounterKey.totalStatus(actorId))`, one row by primary key.
- `publicOnly: true` → the recursive publicly-readable-ids CTE.

There is no `total-public-status:<actorId>` counter, and there cannot be an exact one. A status counts as
publicly readable if it is a public non-Announce, **or** a public Announce whose boosted original (transitively)
is a public non-Announce. That second half depends on a status belonging to a **different actor** — whose
visibility edits, deletes and boosts never touch this actor's rows. Keeping a stored counter exact would mean
fanning every actor's writes out to every actor who has ever boosted them. So the count stays computed; this PR
makes computing it cheaper.

## What changed

Both changes land in the shared `buildPubliclyReadableStatusIdsQuery`, so the other four call sites
(`getActorStatuses`, `getStatusReplies`, `getStatusRepliesCount`, `getRebloggedBy`) get them too.

**1. Count straight off the subquery instead of filtering `statuses` by it.**
The subquery is already scoped to the actor's own statuses, so the outer
`where "actorId" = $1 and "statuses"."id" in (<subquery>)` re-applied a predicate the subquery had already
decided — one `statuses_pkey` lookup per returned id, plus a sort for the `COUNT(DISTINCT)`. That outer re-join
alone was 7,884 of 37,731 shared buffers.

**2. Follow the announce chain through one pointer column.**
The recursive join was `originalStatusId = id OR (originalStatusId IS NULL AND content = id)` — the modern and
legacy spellings of "the status this Announce boosts". PostgreSQL turns that into a `BitmapOr` of two
`statuses_pkey` probes per row, and the CTE has to carry `originalStatusId` **and** the wide `content` text
through the whole recursion for every row, including the Notes that never follow a pointer. It is now a single
`CASE WHEN type = 'Announce' THEN COALESCE("originalStatusId", content) END` column joined by plain equality.

Semantics are unchanged: the pointer is only ever read on rows the recursive term has already filtered to
`type = 'Announce'`, and `COALESCE` picks exactly what the `OR` picked. A NULL pointer joins to no row, which is
what the `OR` did for a Note. Multi-level boost chains still resolve — there is now a positive test for that
alongside the existing negative one.

## Measurements

Local PostgreSQL **18** (production `mastodon-sql` is 18.2, so the same planner — including the skip scan on
`statuses_announce_actor_original_idx` that shows up in the sampled production plan), seeded to the shape that
plan implies: 500k statuses, ~995k recipients, one actor with 2,000 statuses of which 1,000 are Announces whose
originals are scattered across the heap. Warm cache, so shared-buffer counts are the portable number and the
timings are a floor — production is a 1 vCPU / 3.75 GB instance where those buffer touches are largely random IO.

| | before | after |
| --- | --- | --- |
| outbox `totalItems` count | 37,731 buffers / 71 ms | **27,018 buffers / 36 ms** |
| outbox page (`limit 30`) | 37,734 buffers / ~48 ms | **34,905 buffers / ~41 ms** |
| count end-to-end through knex | 54.1 ms/run | **35.7 ms/run** |

Same result (1,971) before and after, on SQLite and PostgreSQL.

Plan changes for the count: the outer `Nested Loop` + `Index Scan using statuses_pkey` (1,971 searches) and its
`Sort` are gone, and the recursive term's `Bitmap Heap Scan` + `BitmapOr` of two `statuses_pkey` scans (1,936
index searches) collapses to one `Index Scan using statuses_pkey` (993 searches). The CTE's working table drops
from 1,086 kB to 426 kB.

## A knex trap worth knowing

knex emits a UNION'd CTE term's `FROM` bindings **before** its select-list bindings. So a bound `?` for the
`Announce` literal inside the new `CASE` swaps places with the caller's target-set value: the `CASE` compares
the status type against the actor id, and the target set filters on `'Announce'`. Both queries compile and run
without error and quietly return nothing useful — I hit exactly this on the first attempt. The literal is
therefore escaped into the fragment (`database.raw('?', [...]).toString()`) and the raw carries identifier
placeholders only, which never reach the binding array.

Because both spellings run, this is pinned by compiled-SQL assertions rather than left to behaviour tests.

## Review-round additions

Two follow-up commits from the sub-agent review loop:

- **The `COALESCE` argument order is now pinned by a behaviour test.** It is load-bearing and nothing held it: swapping the two arguments passed all 165 tests. It matters on real rows, because `20260517000000_add_status_original_status_id.js` backfilled `originalStatusId` by JSON-parsing `content` and left the raw body behind — migrated Announces hold two different non-null strings, and only `originalStatusId` names the boosted status. `createAnnounce` writes the two identically, so no fixture built the normal way can tell the orders apart; the new test hand-writes rows where they disagree in both directions.
- **The MySQL branch of the same helper now builds its chain condition from `announceOriginalPointer` too**, instead of its own hand-spelled `OR`. MySQL has no CI coverage — no service, no docker-compose entry, no testcontainer, and `TEST_DATABASE_TYPE` only accepts `sqlite`/`pg` — so shared construction is the only thing keeping the two branches from drifting, and a compiled-SQL assertion is the honest ceiling for guarding it. One behavioural note traced during review: for a row with `type IS NULL` the old `OR` ignored `type` and could match, while the new `CASE` cannot. That brings MySQL into agreement with the recursive branch, and it is unreachable anyway — every insert into `statuses` sets `type` explicitly.

## Tests

- New: compiled-SQL guards for the binding order and for the single-equality chain join, on both `pg` and
  `better-sqlite3`.
- New: `counts nested announces when the boosted original is publicly readable` — a boost of a boost of a public
  note, which only resolves if the recursion runs past its first iteration.
- Reverting the fix fails 5 tests (2 new SQL guards, the new nested-announce case, and 2 pre-existing behaviour
  tests), so the guards are not vacuous.
- New: `resolves a boost through originalStatusId when content disagrees`, and a compiled-SQL guard for the MySQL
  branch. The swapped-`COALESCE` mutant now fails 2 tests where it previously passed all 165.
- Full gate green: `prettier:check`, `lint`, `typecheck`, `build`, `yarn test` (851 files / 10,567 tests), plus
  `lib/database/sql/status.test.ts` under `TEST_DATABASE_TYPE=pg`.

## Not in scope

Roughly 58% of what remains is the `recipients` public-audience semi-join — 4,000 index searches for a 2,000-row
actor, because `PUBLIC_ACTIVITY_RECIPIENTS` holds both the full `.../activitystreams#Public` URI and the
compacted `as:Public`, and PostgreSQL runs one index search per array element per row. Removing that needs
either a denormalised visibility column on `statuses` or dropping one of the two spellings; both are bigger
calls than a query rewrite.


lib/database/sql/status.ts
```diff
@@ -157,6 +157,31 @@ const publicRecipientStatusIds = (database: Knex) =>
...
+// a bound raw in its select list and a bound subquery in its FROM: the FROM's
+// value is emitted first, so the two swap and the CASE silently compares the
+// status type against the caller's target-set argument. The recursive branch of
+// `buildPubliclyReadableStatusIdsQuery` is exactly that shape. Identifier
+// placeholders are unaffected — they are resolved at compile time and never
+// reach the binding array. Guarded by a compiled-SQL test.
+const announceOriginalPointer = (database: Knex, table: string) => {
+  const announceType = database.raw('?', [StatusType.enum.Announce]).toString()
+  return database.raw(
+    `case when ?? = ${announceType} then coalesce(??, ??) end`,
```
llun, line 180: **COALESCE argument order is load-bearing on real rows but nothing pins it**

The old join preferred `originalStatusId` and only consulted `content` when `originalStatusId IS NULL`, so `COALESCE("originalStatusId", content)` has to keep that precedence. It does — but the order is untested: swapping the two arguments to `coalesce(content, originalStatusId)` passes all 165 tests in `lib/database/sql/status.test.ts` (verified by mutation).

That is not a hypothetical. `migrations/20260517000000_add_status_original_status_id.js` backfills `originalStatusId` by JSON-parsing `content` and writing out `.url` or `.id`, leaving `content` as the raw JSON blob. Production rows therefore exist where both columns are non-null and hold *different strings*, and only `originalStatusId` is the real id. No test — old or new — ever builds such a row: `createAnnounce` always writes the two identically, and the one legacy-row fixture sets `originalStatusId: null`.

Suggested fix: a `getActorStatusesCount({ publicOnly: true })` test with an Announce whose `originalStatusId` points at a PRIVATE note and whose `content` points at a PUBLIC one. It must not count. A swapped `COALESCE` would count it.

AGENTS.md
```diff
@@ -1891,6 +1891,13 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
 
+## Publicly Readable Status Ids
+
+- **There is no counter for the public status count, and there cannot be one.** `getActorStatusesCount({ publicOnly: true })` — the ActivityPub outbox's `totalItems` — is the one status count that is computed rather than read from `counters`. An Announce is publicly readable only while the status it boosts still is, and that status belongs to a **different** actor, whose visibility edits and deletes never touch this actor's rows. So a stored `total-public-status:<actorId>` would need fan-out from every other actor's writes to stay exact. The plain total (`publicOnly: false`) has no such dependency and does read `CounterKey.totalStatus`.
+- **Count straight off `buildPubliclyReadableStatusIdsQuery`; do not filter `statuses` by it.** The subquery is already scoped to the caller's target set, so `where actorId = ? and id in (<subquery>)` re-applies a predicate the subquery has already decided — at one `statuses_pkey` lookup per returned id, plus a sort for the `COUNT(DISTINCT)`. On a PostgreSQL 18 seed shaped like production (500k statuses, an actor with 2,000 of them, half Announces) that outer re-join alone was 7,884 of 37,731 shared buffers.
```
llun: **This reads as a blanket rule the codebase deliberately does not follow**

As written this is general guidance to every caller of `buildPubliclyReadableStatusIdsQuery`, but `getStatusRepliesCount` (`lib/database/sql/status.ts`, ~line 3418) keeps its outer `statuses` query on purpose and must keep it: its `.whereNot('type', StatusType.enum.Announce)` filters the TARGET row, whereas the subquery's `type` column is the *resolved original's* type. A target that is an Announce resolving to a public Note is returned by the subquery, and only the outer filter excludes it. Converting that call site the way this bullet implies would start counting boosts as replies.

Scope the bullet to the case it actually covers — an outer query that adds nothing beyond the subquery's own scoping — and record why `getStatusRepliesCount` is not that case, so a later reader does not "fix" it.

AGENTS.md
```diff
@@ -1891,6 +1891,13 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
+- **knex mis-orders positional bindings when a term of a `UNION`'d CTE mixes a bound raw in its select list with a bound subquery in its `FROM`** — the `FROM`'s value is emitted first, so the two swap silently. That is exactly the shape of `buildPubliclyReadableStatusIdsQuery`, so the `Announce` literal in the chain pointer is escaped into the fragment (`database.raw('?', [...]).toString()`) instead of bound; identifier placeholders (`??`) are resolved at compile time and never reach the binding array. Both queries still compile and run, they just compare the wrong things, so this is guarded by compiled-SQL assertions in `lib/database/sql/status.test.ts` rather than left to behaviour tests.
```
llun: **Restates the code comment rather than adding altitude**

This bullet is nearly the same sentences as the second paragraph of the `announceOriginalPointer` comment in `lib/database/sql/status.ts`, reordered. The only genuinely new content is the closing clause about why this is guarded by compiled-SQL assertions rather than behaviour tests. The neighbouring "Local Actors" bullets each carry a fact its code comment does not. Trim this one to the delta.

lib/database/sql/status.ts
```diff
@@ -157,6 +157,31 @@ const publicRecipientStatusIds = (database: Knex) =>
...
+    [`${table}.type`, `${table}.originalStatusId`, `${table}.content`]
+  )
+}
```
llun, line 183: **Second encoding of the announce-chain pointer, now textually divergent**

The MySQL branch still spells the chain as `original_statuses.id = target_statuses.originalStatusId OR (target_statuses.originalStatusId IS NULL AND original_statuses.id = target_statuses.content)`. That is the same rule `announceOriginalPointer` now centralises for the recursive dialects, and after this PR the two encodings no longer even look alike. Nothing catches drift if the announce-chain semantics change again, and MySQL has no CI coverage at all.

The `CASE`/`COALESCE` expression is standard SQL and dialect-neutral, so this branch can reuse `announceOriginalPointer` directly — the `CASE WHEN type = 'Announce'` guard makes a non-Announce row's pointer NULL, which cannot join, exactly as the old `OR` behaved for those rows.

lib/database/sql/status.ts
llun: Fixed in e2960f8d0. Added `resolves a boost through originalStatusId when content disagrees` in `lib/database/sql/status.test.ts` — it hand-writes two public Announce rows whose two columns disagree in opposite directions: `originalStatusId` → private / `content` → public must not count, and the mirror must. Verified by mutation: swapping the `COALESCE` arguments now fails 2 tests (this behaviour test plus the new MySQL compiled-SQL guard), where before it passed all 165.

AGENTS.md
llun: Fixed in e2960f8d0. The bullet now scopes the rule to an outer query that adds nothing beyond the subquery's own scoping, and calls out `getStatusRepliesCount` by name as deliberately not that case — its `whereNot('type', Announce)` filters the target row while the subquery's `type` is the resolved original's, so collapsing it would count boosts as replies. Recorded so nobody 'fixes' it later.

AGENTS.md
llun: Fixed in e2960f8d0. The bullet now points at `announceOriginalPointer` for the hazards themselves and keeps only the delta: why both the `COALESCE` order and the binding order are pinned by compiled-SQL assertions, and why the `COALESCE` order additionally needs a hand-written row (the backfill migration is what makes the two columns diverge, and `createAnnounce` always writes them identically).

lib/database/sql/status.ts
llun: Fixed in e2960f8d0. The MySQL branch's `.orWhereExists` now builds its chain condition from `announceOriginalPointer` instead of spelling the `OR` out, so both dialect branches share one construction. Added `resolves the MySQL announce chain through the same pointer expression` as a compiled-SQL guard, since MySQL has no CI coverage. Full gate re-run green: prettier, lint, typecheck, build, 851 files / 10,566 tests, plus `lib/database/sql/status.test.ts` under `TEST_DATABASE_TYPE=pg`.

AGENTS.md
```diff
@@ -1891,6 +1891,13 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
..., `COALESCE("originalStatusId", content)`, not an `OR` of the two forms.** Modern Announce rows set both; legacy rows only ever wrote the original's id into `content`. Spelling that as `originalStatusId = id OR (originalStatusId IS NULL AND content = id)` makes PostgreSQL build a `BitmapOr` of two `statuses_pkey` probes per row, and forces the recursion's working table to carry the wide `content` text for every row — including the Notes that never follow a pointer. The `CASE` guard on `type` is what keeps a Note's body out of that column. Together with the point above: **37,731 → 27,018 buffers and 71 → 36 ms** for the count, 37,734 → 34,905 buffers for the outbox page. **Both dialect branches build the pointer from the one `announceOriginalPointer` helper**, which is all they share: the MySQL branch resolves the chain with a correlated `EXISTS` that follows exactly ONE hop, so a boost of a boost is not publicly readable there even when the recursive branch says it is. That gap predates the helper and is not closed by it. Sharing the pointer is still what keeps the two from drifting further, since MySQL has no CI coverage at all — nothing would catch a second hand-spelled encoding.
```
llun: **"Both dialect branches build the pointer" invited a wrong inference about parity**

Round 2 scope review: the wording added in e2960f8d0 — "the MySQL branch resolves the chain with a correlated `EXISTS` rather than a recursive CTE" — reads as though sharing `announceOriginalPointer` makes the two branches equivalent. It does not. The MySQL `EXISTS` follows exactly one hop (`whereNot('original_statuses.type', Announce)` rejects a re-boosted Announce), so a boost of a boost is not publicly readable on MySQL even where the recursive branch resolves it. The limitation predates the helper and is not closed by it.

llun: Fixed in b58a5f6c7. The bullet now states the one-hop limit explicitly and that it predates the shared helper, so the shared construction can't be read as parity.

Two other round-2 observations that did not need a change, recorded here for the log:

- Correctness lens traced a genuine three-valued-logic divergence in the MySQL branch: for a row with `type IS NULL`, the old hand-spelled `OR` ignored `type` entirely and could match on the pointer columns, while the new `CASE WHEN type = 'Announce'` yields NULL and cannot. The new behaviour agrees with the recursive branch, which already excluded NULL-type rows via its final `WHERE NOT type = 'Announce'`. Traced as unreachable — every `insert()` into `statuses` sets `type` explicitly and no API writes arbitrary rows — so this is a strengthening on already-corrupt data, not a behaviour change on anything a client can produce.
- Testing lens confirmed compiled-SQL is the honest ceiling for the MySQL branch: there is no MySQL service, docker-compose entry or testcontainer anywhere in the repo, and `TEST_DATABASE_TYPE` only accepts `sqlite`/`pg`. The assertion earns its brittleness — reverting the MySQL branch to its hand-spelled form fails it.

lib/database/sql/status.ts
```diff
@@ -3409,6 +3432,13 @@ export const StatusSQLDatabaseMixin = (
...
 
+    // This one cannot be collapsed onto the subquery the way
+    // `getActorStatusesCount` was. The filter below is about the REPLY row,
+    // while the subquery's `type` column is the resolved *original's* — an
+    // Announce whose boosted original is a public Note comes back from the
+    // subquery, and only this line keeps it out of a reply count. Nothing
+    // writes `reply` on an Announce today, so dropping it changes no number
+    // anyone can observe, which is exactly why it needs saying here.
     const result = await query
       .whereNot('type', StatusType.enum.Announce)
```
llun, line 3443: **Two invariants lived only in AGENTS.md, and one of them had no test**

Round 3 quality review: the new AGENTS.md section states as a hard rule that `getStatusRepliesCount` must keep its outer `whereNot('type', Announce)`, and that the MySQL branch follows only one hop. Neither fact had any anchor in the code it protects — `getStatusRepliesCount` carried no comment at all, and neither did the MySQL branch — while every other hazard in the section is anchored at `announceOriginalPointer` or in `getActorStatusesCount`'s own comment.

Worse for the first one: nothing tested it. `describe('getStatusRepliesCount')` never passes `publicOnly`, and the route test mocks the database method, so the real SQL was never exercised for an Announce-of-a-reply-to-a-public-Note. A future refactor applying the same collapse this PR applied to `getActorStatusesCount` — plausible, since the commit message lists `getStatusRepliesCount` among the call sites that benefit from the shared subquery — would have had no code-level warning and no failing test.

llun: Fixed in 1bdc39d6e.

- Added `keeps a boosted reply out of the public replies count`: a public Note, a real reply to it, and a hand-written Announce that also points `reply` at it. The subquery returns that Announce (it resolves to a publicly readable Note), so only the outer filter keeps the count at 1. Verified by mutation — removing the `whereNot` fails this test and nothing else in the file.
- Anchored both facts at the code: a comment on that filter, and one on the MySQL branch's single-hop `EXISTS`.
- Reworded the AGENTS.md bullet, because the previous phrasing was overstated in the same way round 2's MySQL sentence was: `createAnnounce` always writes `reply: ''`, so no reachable data makes the filter matter today. Dropping it moves no observable number — which is exactly why only a hand-written row can pin it, and the bullet now says that instead of implying live breakage.

Gate re-run green: prettier, lint, typecheck, build, 851 files / 10,567 tests, plus `lib/database/sql/status.test.ts` under `TEST_DATABASE_TYPE=pg`.

AGENTS.md
```diff
@@ -1891,6 +1891,13 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
+- **The announce chain is followed through one pointer column, `COALESCE("originalStatusId", content)`, not an `OR` of the two forms.** Modern Announce rows set both; legacy rows only ever wrote the original's id into `content`. Spelling that as `originalStatusId = id OR (originalStatusId IS NULL AND content = id)` makes PostgreSQL build a `BitmapOr` of two `statuses_pkey` probes per row, and forces the recursion's working table to carry the wide `content` text for every row — including the Notes that never follow a pointer. The `CASE` guard on `type` is what keeps a Note's body out of that column. Together with the point above: **37,731 → 27,018 buffers and 71 → 36 ms** for the count, 37,734 → 34,905 buffers for the outbox page. **Both dialect branches build the pointer from the one `announceOriginalPointer` helper, and that is all they share** — the MySQL branch's own comment records what it does not do. Sharing the pointer is what keeps the two from drifting further, since MySQL has no CI coverage at all.
```
llun, line 1941: **The code comments and the AGENTS.md bullets now both carry the full reasoning**

Round 4 quality review: round 3 anchored two invariants at the code, which was right, but left the full explanation in both places.

- The MySQL one-hop comment on this line restates this bullet almost clause for clause — "gap predates the shared pointer" here vs "That gap predates the helper" in the code comment; "MySQL has no CI coverage" appears near-verbatim in both.
- Line 1897's `getStatusRepliesCount` bullet restates its code comment near-verbatim too ("filters the REPLY row, while the subquery's `type` column is the resolved original's").

That is the duplication round 1 flagged, arriving from the other direction — two copies of one piece of reasoning with nothing keeping them in step. Line 1899, the section's own last bullet, already sets the convention this should follow: "The hazards themselves are documented at `announceOriginalPointer`... What the tests add: ..." — point at the code, state only the incremental fact.

Separately, line 1897 cites a test by its literal `it()` description ("keeps a boosted reply out of the public replies count"). Every other test cross-reference in AGENTS.md names a file, which is stable and greppable; `it()` names are free-form behaviour prose that a later cleanup pass reworders, and nothing checks the quoted string still matches a real test.

llun: Fixed in 507583a1d. Both bullets now point at the code and keep only what the doc adds — the general rule, the buffer/ms numbers, and the pointer; the reasoning lives at the filter it protects. The test citation is by file (`status.test.ts`) rather than by `it()` description.

The section is 517 words across four bullets now, against 583 for the neighbouring Local Actors section, and all four follow the same shape. No fact was dropped: everything trimmed from the bullets is stated in the comment the bullet now points at.

AGENTS.md
```diff
@@ -1891,6 +1891,13 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
+- **Count straight off `buildPubliclyReadableStatusIdsQuery` when the outer query adds nothing beyond the subquery's own scoping — but check that it doesn't first.** `getActorStatusesCount` was `where actorId = ? and id in (<subquery>)`, and the subquery is already scoped to that actor's statuses, so the outer half re-applied a settled predicate at one `statuses_pkey` lookup per returned id plus a sort for the `COUNT(DISTINCT)`: 7,884 of 37,731 shared buffers on a PostgreSQL 18 seed shaped like production (500k statuses, an actor with 2,000 of them, half Announces). **`getStatusRepliesCount` is deliberately NOT that case and must keep its outer query** — the reason is recorded at that filter in `lib/database/sql/status.ts`, and `lib/database/sql/status.test.ts` pins it with a row no writer can produce, because dropping the filter moves no number anyone can observe.
```
llun, line 1940: **The pinned test is cited by a basename that is not unique**

Round 5 quality review: the bullet cited `status.test.ts`, but the repo has two files with that basename — `lib/database/sql/status.test.ts` (the intended one) and `lib/types/domain/status.test.ts`. Every other test citation in this file (`extractUrl.test.ts`, `getMastodonPreviewCard.test.ts`, `fileName.test.ts`, `localPublicQueryShape.test.ts`, `deleteActorJob.test.ts`) names a file that is unique in the repo, so a reader can open it by name alone. This one broke that convention and could land someone in the wrong file.

Round 5 scope review also flagged a dropped word in the neighbouring clause: "must keep its outer query — why is recorded at that filter" reads as though a word is missing.

llun: Fixed in 20739959b — the citation is now the full `lib/database/sql/status.test.ts`, and the neighbouring clause reads 'the reason is recorded at that filter'.

Round 5's whole-branch pass found nothing else: scope is three files, no migrations and no schema-dump regeneration needed (query-only change), `applyPublicReadableStatusFilter` still has four live call sites, no TODOs, dead code or unused imports, and the commit prefixes match their diffs. The PR body's test count is corrected to 10,567 after round 3 added one.

PR #1496, opened 2026-08-22
fix: post Strava webhook activities at import time, not ride start

llun:
## Summary

A fitness status created by an import was always backdated to the activity's start time. That is right for history replays, but wrong for the Strava webhook: it fires minutes after a ride ends, so the post — which is the news that the ride happened — arrived already buried under everything published while the ride was still in progress. A four-hour ride landed four hours down the timeline.

**What changed**

- Added `postAtImportTime` to `importFitnessFilesJob` and `importStravaActivityJob`, threaded from the webhook route exactly the way `notifyOnComplete` and `publishSendNote` already are: set by the originating publisher, defaulting to `false`. When set, the status's `createdAt` — and the v7 `publicId` minted from it — become the moment the import published the post.
- **Only the webhook opts in.** Every bulk path keeps backdating, because each replays history: the archive walker, multi-file uploads, retry-all and the `scripts/fitness` recovery tools would otherwise reorder an actor's whole timeline around whenever a sweep happened to run.
- It applies to **any** activity the webhook delivers, however old. Strava fires `create` for a late device sync, a back-dated manual entry and a bulk upload too, and none is treated differently — so restoring a season of old rides files them at the top in upload order. That is a deliberate choice (see **Review** below), and the docs say so rather than implying a recency check the code never makes.
- A webhook import that merges into a sibling's existing post (the same ride recorded on two devices) reuses that status and keeps whatever stamp it already carries.
- The Strava fallback path for a streamless activity never reaches `importFitnessFiles` and is unchanged — it has always stamped at import time on every caller.

**The recorded start time is unaffected** — it lives on `fitness_files.activityStartTime`, which is what the activity page and every fitness rollup read. One follow-on was needed for that to stay true: `FitnessStatusDetail` used `status.createdAt` as the placeholder activity date until its by-status fetch lands, and those two are no longer equivalent for webhook imports, so the status payload now carries `activityStartTime`. It is already public at the same visibility — `/api/v1/fitness-files/by-status` serves it to every viewer who can see the status — so this exposes nothing new.

Documented in `docs/fitness-file-storage.md` (Processing Pipeline), and the invariant recorded in `AGENTS.md` (**Fitness Activity Dates**) and `CLAUDE.md`. No migrations, no new env vars, no route changes.

**Tests added**, each verified to fail on revert:

- `importFitnessFilesJob` — an opt-in import stamps the status at a pinned import time, mints the URI tail from that same stamp, and leaves `activityStartTime` on the file untouched. A second test covers the **merge** path with the flag on: a sibling joining an existing post must not re-stamp it. The existing backdating test still pins the default.
- `importStravaActivityJob` — table-driven: withholds the flag by default, forwards the webhook opt-in.
- Webhook route — the published job payload carries `postAtImportTime: true` (exact-object assertions, so the field cannot be dropped with a green build).
- `getStatus` — the fitness payload surfaces the recorded start time, exercised on both backends.
- `FitnessStatusDetail` — dates the activity from the recorded start, not the post time, while the by-status fetch is still unresolved.

## Screenshots

No layout or styling change; the visual surface is the timestamp itself, so the evidence is the rendered values rather than a diffable image (the session cannot attach files to a PR).

Verified in Chromium against local SQLite with a seeded webhook-shaped activity — ride started 08:27, post published 13:27:

- Profile feed: the post sits at the top reading **"2 minutes"**, above posts seeded minutes earlier.
- Activity detail page: still reports **"8:27 AM, August 22, 2026"** — confirmed in the raw SSR HTML, not only after hydration, which is what the payload change buys.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn typecheck`, `yarn build`, and `yarn test` pass, run in that order — 853 test files, 10,584 tests. **Caveat:** the local runs used Node 22; this repo pins Node 24 and that version could not be installed in the session's container. CI runs the full gate on Node 24 and is green on `b96bddb`.
- [x] Docs updated: grepped `*.md` and `docs/` for the identifiers this PR touches; `docs/fitness-file-storage.md`, `AGENTS.md` and `CLAUDE.md` all updated
- [x] N/A — no migration is added, edited, or removed, so neither schema dump moves
- [x] `version` in `package.json` is untouched (the bump present here arrived by merging `main`)

## Review

**The sub-agent code-review loop ran to a clean round.** (An earlier version of this description said the loop had not run — true when the PR was opened, because the session was configured not to spawn sub-agents. That restriction was lifted afterwards.)

| Round | Scope | Outcome |
|---|---|---|
| 1 | 4 reviewers in parallel over the whole diff | 5 findings → `569c3f3`; 1 justified won't-fix; 2 design questions |
| 2 | Fresh eyes on whether round 1's fixes broke anything | **No code defects.** 2 doc findings → `3360609` |
| 3 | Fresh eyes, instructed to disagree with rounds 1-2 | **No code defects.** 1 doc finding → `b96bddb` |
| 4 | Doc accuracy and cross-surface consistency | **Clean — loop terminates** |

Every round after the first found the **code** clean; all three later findings were in prose, twice in the fix for the previous round's finding. Rounds 3 and 4 re-derived the key claims from code rather than from earlier rounds' conclusions, including checking the deliberate truthiness guard in `lib/database/sql/status.ts` empirically. Details in the round summaries below, findings as inline comments.

**Two design questions were raised and deliberately declined by the author:** closing the retry-path gap (a retried webhook import is backdated again) and adding a recency guard for bulk or back-dated Strava uploads. Neither is a regression against `main`; the reasoning is recorded in the thread so it is findable if the bulk-backfill case ever bites.

## Note on semantics

The stamp is taken when the status row is created — after the file is parsed, slightly before the route map finishes and the `Create` federates. That is the only point where it can be set without rewriting the post's identity (the `publicId` is minted from it), and it lands within seconds of "finished processing".


lib/database/sql/status.ts
```diff
@@ -3285,6 +3285,13 @@ export const StatusSQLDatabaseMixin = (
               ...(fitnessFile.activityType
                 ? { activityType: fitnessFile.activityType }
                 : null),
+              ...(fitnessFile.activityStartTime
```
llun, line 3311: **Finding (low) — WON'T FIX, deliberately. The truthiness guard treats epoch 0 differently on the two backends.**

Two reviewers disagreed here, so I checked the reachable values directly:

| value | `truthy` | `!= null` | `typeof === 'number'` |
|---|---|---|---|
| PG `Date(0)` | ✅ | ✅ | ❌ |
| SQLite `0` | ❌ | ✅ | ✅ |
| PG `Date(now)` | ✅ | ✅ | ❌ |
| `null` / `undefined` | ❌ | ❌ | ❌ |

So for a file whose recorded start is exactly `1970-01-01T00:00:00Z`, PostgreSQL emits `activityStartTime: 0` and SQLite omits the key. Reachable: `getStravaActivityStartTimeMs` returns `0` for that `start_date`, and a GPX `<metadata><time>` reaches the column the same way.

**Why it stays as-is anyway.** Switching to `typeof === 'number'` — to match the numeric fields just above — would drop the field on *every* PostgreSQL row, since `pg` hands back a `Date`. `!= null` is the only correct guard, but this line is deliberately character-identical to `parseSQLFitnessFile` (`lib/database/sql/fitnessFile.ts:365-367`), which feeds `/api/v1/fitness-files/by-status`. Fixing only this site makes the two surfaces disagree on SQLite for an epoch-0 row, so the activity detail page would render `1970` and then flip to the post time once the fetch lands — worse than the asymmetry it fixes.

Fixing both sites is the right change, but it also moves what `importFitnessFilesJob.ts:169` (`typeof x === 'number'`) sees, which changes overlap-merge grouping. That needs its own tests and its own PR rather than riding along here.


---

app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx
```diff
@@ -2159,7 +2159,14 @@ export const FitnessStatusDetail: FC<Props> = ({
         movingTimeSeconds: status.fitness.movingTimeSeconds ?? null,
         elevationGainMeters: status.fitness.elevationGainMeters ?? null,
         activityType: status.fitness.activityType ?? null,
-        activityStartTime: status.createdAt,
+        // Must come off the payload, not from `status.createdAt`: a Strava
+        // webhook import is stamped when it published rather than when the ride
+        // began, so a post-time stand-in dates the activity wrongly until the
+        // by-status fetch below replaces this placeholder. The `??` arm is only
+        // a shape default — the single fallback that actually renders is the one
```
Comment (llun) [inline app/(timeline)/[actor]/[status]/FitnessStatusDetail.tsx:2166]: **Finding (low) — FIXED in `569c3f3`. New read was missing from the `useMemo` dependency array.**

Flagged independently by three reviewers. `status.fitness.activityStartTime` was read in the memo body but absent from the dep list, which names its 18 siblings individually. Nothing catches it — `.oxlintrc.json` declares no `react-hooks` plugin, so there is no `exhaustive-deps` rule; lint, build and the suite were all green with it missing.

Impact was small (a re-render where only `activityStartTime` changed would return the stale placeholder, and the by-status fetch overwrites it shortly after), but it is an omission this diff introduced, so it is fixed.

A separate reviewer also noted the `?? status.createdAt` arm here is behaviourally dead — `activityDate` further down owns the fallback that actually renders — and that the old comment overstated it. The comment now says so rather than inviting someone to "simplify" the real fallback away.

`deviceGearId` / `deviceGearName` are still missing from the same array; that is pre-existing and left alone.


---

docs/fitness-file-storage.md
```diff
@@ -308,6 +308,10 @@ This endpoint is **anonymous / unauthenticated** on purpose: public embeds and s
...
 
+**When a post created by `importFitnessFiles` is stamped depends on why the import ran.** Every bulk path — the Strava archive walker, a multi-file upload, the retry-all endpoint and the `scripts/fitness` recovery tools — backdates the status to the earliest target activity's start time, falling back to the fitness file row's own `createdAt` (the upload moment) when the parse yielded no start time at all, as a GPX carrying no timestamps does. The URI tail is a v7 `publicId` minted from whichever of those two the import resolved, so it always sorts with the post. The **Strava webhook** is the single caller that opts out, through `postAtImportTime`: it fires minutes after a ride ends and its post is the news that the ride happened, so backdating buried a four-hour ride four hours down the timeline, below everything published while it was still being ridden. Only the stamp on the post moves — the recorded start time lives on `fitness_files.activityStartTime`, which is what the activity page and every fitness rollup read — and a webhook import that merges into a sibling's existing post (the same ride recorded on two devices) keeps whatever stamp that post already carries.
```
llun: **Finding (medium) — FIXED in `569c3f3`. The original paragraph made three false claims.**

Caught by three reviewers independently:

1. *"Every bulk path … backdates the status it creates"* — false for the **streamless** Strava activity. One with no positive duration and no streams never reaches `importFitnessFiles` at all; it gets `getOrCreateStravaFallbackNote` (`lib/jobs/importStravaActivityJob.ts:349-410`), which calls `createNote` with no `createdAt`, so `createNote` defaults to now. That path has always stamped at import time on *every* caller, including retry-all and the recovery scripts.
2. *"The URI tail is a v7 `publicId` minted from that same stamp"* — false on the same branch, which mints a deterministic sha256 of the Strava activity id (`getStravaFallbackPostId`) so a re-import finds the same status. The code itself calls this "a deliberate exception to the publicId-tail convention".
3. *"…which is what the activity page, **the stat strips** and every fitness rollup read"* — `FitnessStatGrid` renders distance / moving time / pace / elevation and **no date at all**.

It also omitted the `?? fitnessFile.createdAt` fallback (`importFitnessFilesJob.ts:502-504`) that applies when a parse yields no start time — a GPX carrying no timestamps — which this same document is careful about 87 lines earlier.

The paragraph is now scoped to statuses created by `importFitnessFiles`, states the fallback, drops the stat-strips claim, and a second paragraph documents the streamless exception explicitly.


---

llun (commented):

The PR body says the review loop did not run because this session was configured not to spawn sub-agents. **That restriction has since been lifted and the loop has now run**, so please disregard that section — I'll correct the body once the rounds finish.

Round 1 fanned out four reviewers in parallel over the whole diff: job-pipeline correctness, `AGENTS.md`/`REVIEW.md` invariants, the data layer and payload, and tests plus UI surfaces. Findings inline above and below. `569c3f3` addresses everything that was mine to fix; CI is green on it.

**Fixed in `569c3f3`**
- Three false claims and one omission in the new `docs/fitness-file-storage.md` paragraph (inline).
- Missing `useMemo` dependency for the new `activityStartTime` read (inline).
- The import-time test asserted its stamp fell between two real `Date.now()` reads. A window only proves the stamp is not the activity start, and it fails spuriously if the wall clock steps backwards. Now pinned with `vi.setSystemTime` — the pattern this repo already uses — and asserts exact equality.
- **New test for the merge path with the flag on.** `docs` claims a webhook import merging into a sibling's post keeps that post's stamp; the code does that, but nothing tested it, so a refactor that re-stamped on reuse would have shipped green and jumped an already-federated ride back to the top of every follower's timeline. Both new tests were verified to fail on revert.
- Recorded the invariant in `AGENTS.md` (new **Fitness Activity Dates** section) and `CLAUDE.md`: a fitness post's `createdAt` is no longer the activity's start time, so anything rendering, sorting or grouping by when the activity *happened* must read `fitness_files.activityStartTime`. That trap already bit `FitnessStatusDetail` inside the original diff, which is the argument for writing it down.

**Won't fix, with reasoning** — the epoch-0 guard asymmetry in `lib/database/sql/status.ts` (inline).

**Two open questions for you** — both are design calls I don't think I should make unilaterally, so they're in a separate comment below rather than silently patched.

**Verified clean** across the four reviewers: every other caller of both jobs keeps its previous behaviour; the overlap/merge grouping reads only fitness-file fields, never a status `createdAt`; idempotency is unchanged (dedup is `isNewImport` plus the `existingStatusId` lookup, neither timestamp-derived); no new exposure, since `/api/v1/fitness-files/by-status` already serves `activityStartTime` at the same visibility gate and `status.fitness` reaches neither the Mastodon nor the ActivityPub serializer; no added query cost; no hydration-rule violation; test conventions and the commit prefix check out.

---

llun:
## Two open questions from the review — author's call

Both fall outside what the change was asked to do, and both would alter semantics rather than fix a slip, so I've left the code as-is and am raising them instead of quietly patching.

### 1. Retrying a failed webhook import silently reverts to backdating

`lib/services/fitness-files/retryImports.ts:135-150` rebuilds the `strava-activity:<id>` retry job as `{ actorId, stravaActivityId }` only, so `postAtImportTime` defaults to `false`.

Concretely: the webhook fires for a ride that ended at 18:00, the import fails during parse or during status creation, no status exists yet, and the file goes `importStatus: 'failed'`. The owner clicks **Retry** on `/fitness/files` — and the post that finally appears is backdated to the ride start, hours down their own timeline. Exactly the symptom this PR fixes, reachable from a button. Same for a worker killed mid-import, which `isFitnessImportStuck` re-offers 15 minutes later.

The awkward part is that one function serves two callers with opposite needs: `POST /api/v1/fitness/import/[batchId]` (the user retrying **one** activity, usually fresh) and `POST /api/v1/fitness/retry-failed` (a sweep over **every** failed batch, which may be months old). `retryFitnessImportBatch` already special-cases `strava-activity:` batches — it deliberately omits `visibility` for them — so the shape is distinguishable if you want it to be.

### 2. Nothing enforces "the webhook fires minutes after a ride ends"

The route filters only on `object_type === 'activity' && aspect_type === 'create'`. Strava also emits `create` for a late device sync, a manually back-dated entry, and a bulk file upload — so restoring three weeks of rides after a new head unit would stamp ~20 posts at queue-drain order on top of the timeline, each federating a `Create`, where before they read chronologically.

A recency guard — fall back to the activity's start time when `start_date` is more than N hours old — would answer **both** questions at once, including the retry case, since a just-failed ride is still fresh on retry while a months-old sweep is not. The cost is a threshold to pick and a conditional rule instead of a flat one.

My default, absent a steer, is to leave both as they are: the PR does what it was asked to do, and neither is a regression against `main`. Happy to implement either.

---

llun:
**Decision on both open questions: leave as-is.** Confirmed by the author.

1. **Retry gap** — retries keep backdating, matching the documented rule that recovery paths replay history. Not a regression against `main`: a retried import behaves exactly as it did before this PR. The retry-all sweep over old failed batches stays safe by construction.
2. **No recency guard** — every webhook import posts at import time, flat rule, no threshold. Predictable and easy to explain; the bulk-restore case is rare and self-inflicted.

Both are recorded here rather than in the code so the reasoning is findable if the bulk-backfill case ever does bite someone.

---

llun:
## Sub-agent code review — rounds 2-4 (loop complete)

**Round 4 surfaced no findings. The loop terminates on a clean round**, well inside the 20-round cap.

| Round | Scope | Outcome |
|---|---|---|
| 1 | 4 reviewers in parallel over the whole diff | 5 findings → `569c3f3`; 1 justified won't-fix; 2 design questions to the author |
| 2 | Fresh eyes, focused on whether round 1's fixes broke anything | **No code defects.** 2 doc findings → `3360609` |
| 3 | Fresh eyes, told to disagree with rounds 1-2 | **No code defects.** 1 doc finding → `b96bddb` |
| 4 | Narrow: doc accuracy and cross-surface consistency | **Clean** |

Every round after the first found the **code** clean; all three later findings were in prose I had written, which is worth stating plainly rather than dressing up as thoroughness.

**Round 2** caught that the new `AGENTS.md`/`CLAUDE.md` entry overstated its own invariant. **Round 3** caught that round 2's fix removed the false claim from the middle of the bullet and left an equivalent one in the lead sentence — the half a reader actually absorbs — contradicting the counter-examples below it. **Round 4** confirmed `b96bddb` did not repeat that pattern.

Rounds 3 and 4 independently re-derived, from code rather than from prior rounds' conclusions:

- All three counter-cases the docs now assert (merged multi-device where the primary is not the earliest; no parsed start time; the streamless fallback note).
- The deliberate truthiness won't-fix in `lib/database/sql/status.ts` — checked **empirically**, confirming knex/better-sqlite3 returns that column as a number while `pg` returns a `Date`, so `typeof === 'number'` would drop every PostgreSQL row and `!= null` would split `getStatus` from `/api/v1/fitness-files/by-status`.
- That `postAtImportTime: true` appears only in the webhook route, with all seven other entry points defaulted to `false`.
- That a back-dated manual Strava entry has positive elapsed time and so goes through `importFitnessFiles` with the flag set, rather than the fallback-note path — which is what makes the "for ANY activity however old" sentence describe the right code path.

**One disclosure.** The local pre-commit gate ran on **Node 22**; this repo pins **Node 24** (`engines`, `CLAUDE.md`, and every `.github/workflows/ci.yml` job) and Node 24 is not installable in this container. CI is the authority here and runs the full gate on 24 — it is green on `b96bddb`, including all four test shards, Build, Type Check, Lint and Prettier and Schema Dump Sync.

---

PR #1498, opened 2026-08-22
fix: cap composer media at the configured attachment limit

llun:
## Problem

The status composer refused an 11th image on every instance, no matter what **Admin → Posts & media → Max media attachments** was set to. The setting defaults to **20**, so this was wrong out of the box.

The picker and the reducer guard behind it both read a build-time constant:

- `lib/components/post-box/upload-media-button.tsx` — `availableSlots`, the disabled state, and the `Add media (n/10)` label
- `lib/components/post-box/reducers.ts` — the `addAttachment` cap

…all off `MAX_ATTACHMENTS = 10` in `lib/services/medias/constants.ts`. The admin setting `posts.maxMediaAttachments` only ever changed what `/api/v1/instance` and `/api/v2/instance` advertised as `max_media_attachments`. Third-party Mastodon clients honoured it; this instance's own composer did not.

This is the rule in **Instance Limits in Client Components** (`AGENTS.md`) — client authoring UI sizes itself to the resolved server settings, never to a hardcoded constant — with one limit left behind when that context was introduced.

## Fix

Publish `posts.maxMediaAttachments` through `useInstanceLimits()` like every other admin-configured authoring limit, and read it where the cap is actually applied:

- `lib/components/instance-limits.tsx` — new `maxMediaAttachments` field, defaulting to `MAX_STORED_MEDIA_ATTACHMENTS` so an unwrapped surface matches `DEFAULT_SERVER_SETTINGS`
- `app/(timeline)/layout.tsx` — publishes the resolved value
- `upload-media-button.tsx` — slots, disabled state and label
- `post-box.tsx` and `posts/status-reply-box.tsx` — both composers pass it to the reducer

The reducer cap now rides on the action, `addAttachment(attachment, maxAttachments)`, mirroring how `addPollChoice(maxOptions)` already carries `polls.maxOptions`. The reducer stays pure.

## What is deliberately unchanged — and a gap the review found

**No route enforces the resolved setting, and the two families of route do not even agree on a fallback.** `lib/config/serverSettings.ts` says so in a comment on the field — *"the setting drives the value advertised to clients, not the hard create-route cap"* — so enforcing it server-side would contradict an explicit design decision, and this PR does not.

The review loop then found that the picture is worse than "bounded at 20". `POST`/`PUT /api/v1/statuses[/:id]` do fall back to `MAX_STORED_MEDIA_ATTACHMENTS`. But **`POST /api/v1/accounts/outbox` bounds the attachment count by nothing at all** — its request schema puts no `.max()` on `attachments`, the route checks neither constant, and `createNoteFromUserInput` maps every entry straight into `database.createAttachment`. That is the route both the composer and the inline reply box actually post through (`post-box.tsx` → `createNote()` in `lib/client.ts`).

So on the path the web UI uses, **the client cap this PR adds is the only limit that exists.** That is a pre-existing bug, deliberately not widened into this PR; it is tracked separately. What this PR does do is stop four documentation surfaces from claiming otherwise — they previously said, variously, that every context limit was server-enforced, or that the worst case was 20.

**`MAX_ATTACHMENTS` keeps its value of 10.** It still bounds how many photos a Strava import attaches to one activity post (`importStravaActivityJob`, `importStravaArchiveJob`, `scripts/fitness/importStravaArchive.ts`). It is now commented as that, so it is not mistaken for the composer's cap again.

**Federation is untouched.** `MAX_FEDERATION_MEDIA_ATTACHMENTS = 4` still trims the outbound note, so a 20-image post shows 20 locally and 4 on Mastodon.

The composer can never outrun the API: the setting's schema is `.max(MAX_STORED_MEDIA_ATTACHMENTS)` and it has no environment variable, so the resolved value is always ≤ the route ceiling.

## Tests

`lib/components/post-box/upload-media-button.test.tsx`, `reducers.test.ts` and `instance-limits.test.tsx` — the picker tests now render inside an `InstanceLimitsProvider` and state their own cap instead of importing the constant. New cases cover accepting 12 files at a limit of 20, stopping at a limit of 4, and the reducer honouring the limit on the action.

**Proven to guard the fix:** with the implementation reverted and the tests kept, 6 tests fail (`expected 12 times, but got 10`, `expected length 11 but got 10`, and the label/disabled assertions). Restored, all pass.

## Verification

Full gate, in order: `prettier --write` → `lint` → `typecheck` → `build` → `test` (852 files, 10,574 tests). All pass.

Browser-verified against a local SQLite instance with a mock user:

| | composer | inline reply box |
|---|---|---|
| default (20) | `Add media (0/20)`, accepted all 12 selected images | `Add media (0/20)` |
| `posts.maxMediaAttachments` = 5 | `Add media (5/5)`, button disabled, 5 of 8 selected attached | `Add media (0/5)` |

Before this change both read `n/10` and refused the 11th regardless.

## Docs

`docs/environment-variables.md` claimed lowering `posts.maxMediaAttachments` "changes what clients are told, not what is accepted" — still true of the API, now stale about the built-in composer. Updated, along with the consumer list in **Instance Limits in Client Components** (`AGENTS.md`) and its `CLAUDE.md` reminder.

## Review loop

Five rounds, five lenses per round (correctness, design-system, quality, scope/regression, testing), every finding put through an adversarial verifier before posting. **12 findings confirmed, 1 refuted, 3 dropped as prose preference or pre-existing scope.** All 12 are fixed, replied to and resolved in the threads below.

Round 1 found 8 on the original diff: the `InstanceLimits` JSDoc and the `AGENTS.md`/`CLAUDE.md` bullets all claimed every context limit was server-enforced, which stopped being true the moment `maxMediaAttachments` joined that interface; plus four test gaps, each **proven by mutation** — reverting the reply box to a hardcoded 10, deleting the layout's publishing line, and deleting the `post-box.tsx` ref guard each left the suite fully green, and one test asserted the enabled state while its name promised the disabled one.

Rounds 2-5 reviewed each round's *fixes* rather than the original diff, which is what surfaced the outbox gap above. No production logic has changed since the first commit; everything after it is comments, docs and tests.

Two things the loop is worth recording for:
- The `postExtensionRef` guard looked redundant with the picker's own gate. It is not — the ref is written synchronously before dispatch, so the reducer's cap never covers it, and when the reducer rejects an over-cap add it returns the same state object, so React bails out of re-rendering and no corrective effect fires. Two overlapping picker batches could push the ref past the cap while the UI showed the right count, and `uploadMediaAttachments()` reads that ref directly.
- The regression test for that race initially **passed with the bug present** — React's effect flush happened to heal the ref. It needed deferred `resizeImage` promises to sequence the batches deterministically, and was then re-verified over repeated shuffled runs.


lib/components/instance-limits.tsx
```diff
@@ -32,6 +33,13 @@ export interface InstanceLimits {
   maxStatusCharacters: number
   /** Resolved `media.maxFileSize` in bytes — the upload picker's size budget. */
   maxMediaFileSize: number
+  /**
+   * Resolved `posts.maxMediaAttachments` — how many attachments the media
+   * picker lets a single status carry. This is the same value the instance
+   * entity advertises as `max_media_attachments`, so the built-in composer
+   * caps itself exactly where a third-party Mastodon client would.
+   */
+  maxMediaAttachments: number
```
llun, line 52: **JSDoc's blanket "enforced server-side" claim is false for the field this PR adds**

The interface docblock (lines 25-29, just above) says these values "drive UX only, never authorization: the same resolved limits are enforced server-side on every status create/edit route (`validateStatusContentLimits`) ... so a missing or stale provider can only make the client optimistic, never let something through."

That is not true of `maxMediaAttachments`, the field added right here by this PR. Verified by reading every candidate path: `validateStatusContentLimits` (lib/services/statuses/contentLimits.ts) checks only text length and poll fields; `resolveStatusMedia` (app/api/v1/statuses/route.ts:115) and the edit route (app/api/v1/statuses/[id]/route.ts:211,225) bound against the fixed `MAX_STORED_MEDIA_ATTACHMENTS` (20), never the resolved setting; `getAttachmentsFromMediaIds` applies no count check; `exceedsMaxMediaUploadSize` is bytes-per-file only. `app/api/v1/accounts/outbox/route.ts` has no count check at all.

Failure scenario: an admin sets `posts.maxMediaAttachments` to 5. Any client that ignores the advertised `max_media_attachments` — or our own composer with a stale/missing provider — submits 6-20 media_ids and the server accepts them. That is exactly the "let something through" the docblock rules out. The risk of leaving it: a future reader trusts this comment (it is the closest, most load-bearing copy of the claim) and treats the client cap as a security boundary.

Fix: narrow the claim so this new field is called out as the exception — the character, poll and file-size limits are server-enforced; `maxMediaAttachments` is bounded only by the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling, not by the admin-configured value.

(Anchored here at the new `maxMediaAttachments` field declaration rather than lines 25-29 because those lines are unchanged context and outside this PR's diff hunks — line must be part of the diff to post an inline comment.)

AGENTS.md
```diff
@@ -172,7 +172,7 @@ ActivityPub objects are **JSON-LD**, so the same logical object can arrive in ma
...
 - Add a field to `InstanceLimits` and publish it from the layout rather than threading a prop: the composer renders inline under posts across the whole route group, so prop-threading touches ~30 files. Keep the context to values the browser genuinely needs; it is not a mirror of `ResolvedServerSettings`.
 - **Do not import `lib/config/serverSettings` into a Client Component** for its value exports — it carries `process.env`-reading closures. Defaults belong in a dependency-free module (see **Server/Client Module Boundary**).
 - The context is UX only. Every limit is enforced server-side too (`validateStatusContentLimits` on the status create/edit routes, `exceedsMaxMediaUploadSize` on every upload path); a client that is out of date can only be optimistic, never permissive. When a new surface can author or upload, put it under the provider and read the limit from it.
```
llun: **AGENTS.md and CLAUDE.md still claim every context limit is server-enforced, contradicting this PR's own docs edit**

This PR edited line 175 to add `posts.maxMediaAttachments` to the list of limits the context publishes, but left this line unqualified: "The context is UX only. Every limit is enforced server-side too (`validateStatusContentLimits` ... `exceedsMaxMediaUploadSize`); a client that is out of date can only be optimistic, never permissive."

The same commit edited `docs/environment-variables.md` to state the opposite for this specific setting: "neither create endpoint rejects a status carrying more than the configured number of attachments — they bound themselves by the stored-media ceiling instead." `CLAUDE.md` carries the same unqualified claim ("Limits are still enforced server-side") in the bullet this PR also edited.

So three docs touched by one PR now disagree with each other, and the contradiction is self-inflicted by this diff — before it, neither AGENTS.md nor CLAUDE.md mentioned attachment count under this section. This repo explicitly guards against exactly this drift (Documentation Maintenance).

Fix: add the `posts.maxMediaAttachments` carve-out to both this line and the corresponding CLAUDE.md bullet, pointing at the fixed `MAX_STORED_MEDIA_ATTACHMENTS` (20) as what the routes actually enforce.

lib/components/post-box/post-box.tsx
```diff
@@ -1094,6 +1095,12 @@ export const PostBox: FC<Props> = ({
               isMediaUploadEnabled={isMediaUploadEnabled}
               attachments={postExtension.attachments}
               onAddAttachment={(attachment) => {
+                if (
```
llun, line 1107: **Load-bearing ref guard ships with no comment saying why it is not redundant**

This new early return is NOT redundant with the picker's `availableSlots` gate or the reducer's cap, but nothing at the call site says so — it reads like a copy of a rule already checked twice, and a future "simplify the triple check" cleanup will delete it.

Verified mechanism: unlike `status-reply-box.tsx` (a bare dispatch), this handler writes `postExtensionRef.current = nextExtension` synchronously and unconditionally BEFORE dispatching, so the reducer's `case 'addAttachment'` guard — which protects only the committed `postExtension` — never sees it. The picker's disabled state and `availableSlots` both read `postExtension.attachments.length`, which does not advance until after `await resizeImage(...)` resolves and React commits. So two overlapping picker batches each compute the same stale `availableSlots` and each call `onAddAttachment`.

Concrete failure with the guard removed, at `maxMediaAttachments = 4` and 3 attachments: batch A (1 large photo) is still inside `resizeImage`; the button is still enabled (3 < 4) so the user picks a second photo as batch B, which also computes `availableSlots = 1`. A resolves — ref 4, reducer 4. B resolves — `nextExtension` built from the 4-item ref becomes 5, `postExtensionRef.current` is set to 5; its dispatch is correctly rejected so the UI still shows 4. If the user hits Post before the `[postExtension]` resync effect fires, `uploadMediaAttachments()` (post-box.tsx:493) reads the ref directly and submits 5. `isEditSubmittable`/`isEditDirty` read the same ref.

Fix: add a comment stating that this bounds the REF, that the reducer's cap bounds only `postExtension`, and that it must not be removed as redundant.

lib/components/posts/status-reply-box.tsx
```diff
@@ -383,7 +383,7 @@ export const StatusReplyBox: FC<Props> = ({
                 isMediaUploadEnabled={isMediaUploadEnabled}
                 attachments={postExtension.attachments}
                 onAddAttachment={(attachment) => {
-                  dispatch(addAttachment(attachment))
+                  dispatch(addAttachment(attachment, maxMediaAttachments))
```
llun, line 386: **Inline reply box attachment cap has no test coverage**

Proven by mutation: changing this line back to `dispatch(addAttachment(attachment, 10))` — reintroducing the exact pre-fix bug, scoped to the reply box — leaves all 4 tests in `lib/components/posts/status-reply-box.test.tsx` passing. That test file wraps in `<InstanceLimitsProvider maxStatusCharacters={...}>` only, never sets `maxMediaAttachments`, and makes no attachment-cap assertion.

The reply box is one of the two surfaces this PR fixes; half the fix is unguarded. Add a test that renders it under a provider with a low `maxMediaAttachments` and asserts the cap holds.

app/(timeline)/layout.tsx
```diff
@@ -37,6 +37,7 @@ const Layout: FC<LayoutProps> = async ({ children }) => {
   const instanceLimits = {
     maxStatusCharacters: posts.maxCharacters,
     maxMediaFileSize: media.maxFileSize,
+    maxMediaAttachments: posts.maxMediaAttachments,
```
llun, line 40: **The layout's publishing of maxMediaAttachments is untested and fails silently**

Proven by mutation: deleting this line leaves `app/(timeline)/layout.test.tsx` passing 6/6 AND `yarn typecheck` clean, because `maxMediaAttachments` is optional on `ProviderProps`. Both safety nets are silent, and the consequence is the whole fix reverting to the 20 default with no signal.

`layout.test.tsx` mocks `InstanceLimitsProvider` and asserts only on `data-max={maxStatusCharacters}`; it was not touched by this PR. Extend it to assert the new field is published.

lib/components/post-box/post-box.tsx
llun, line 1107: **The post-box ref guard has no test**

Proven by mutation: deleting this guard block entirely leaves `lib/components/post-box/post-box.test.tsx` passing 31/31. No test exercises `onAddAttachment` or the cap in this component at all.

This is the same code flagged in the sibling comment on this file (the missing-comment finding) and matters for the same reason — the guard protects `postExtensionRef.current`, which the submit path reads directly. A comment stops a human deleting it; a test stops everything else. Add one covering the over-cap add.

lib/components/post-box/upload-media-button.test.tsx
```diff
@@ -117,10 +120,57 @@ describe('UploadMediaButton', () => {
...
+
+      fireEvent.change(input, { target: { files: [file] } })
+
+      await waitFor(() => {
+        expect(mockOnUploadStart).toHaveBeenCalledTimes(1)
+      })
+      expect(mockOnAddAttachment).not.toHaveBeenCalled()
+    })
+
+    it('counts against the instance limit in its label and disabled state', () => {
```
llun, line 164: **Test named for the disabled state never asserts the disabled state**

`counts against the instance limit in its label and disabled state` asserts only `.toBeEnabled()` for a below-cap case. Proven vacuous for the half its name promises: changing `disabled={attachments.length >= maxMediaAttachments}` to `disabled={false}` in upload-media-button.tsx leaves all 14 tests in this file passing.

The disabled logic predates this PR, but this PR's test claims to cover it. Either assert `.toBeDisabled()` at the cap or rename the test to what it actually checks.

lib/components/instance-limits.test.tsx
```diff
@@ -30,6 +35,37 @@ describe('useInstanceLimits', () => {
...
+  })
+
+  it.each([
+    {
+      description: 'serves the resolved attachment cap',
+      value: 12,
+      expected: '12'
+    },
+    {
+      description: 'falls back to the default attachment cap when omitted',
```
llun: **Two it.each descriptions collide under Vitest's 40-char truncation**

`falls back to the default attachment cap when omitted` (this line) and `falls back to the default attachment cap for zero` (5 lines below) both render as `falls back to the default attachment …` in Vitest output — observed directly as two identically-named FAIL lines during a revert experiment. If one of the two regresses, the CI failure list alone will not say which. Shorten both descriptions so they differ within the first 40 characters.

lib/components/instance-limits.tsx
llun: Fixed in 776b962c5: narrowed the docblock in `lib/components/instance-limits.tsx` to carve out `maxMediaAttachments` as the one exception to "enforced server-side," pointing at the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling and `docs/environment-variables.md` for the full explanation.

AGENTS.md
llun: Fixed in 776b962c5: added the same `maxMediaAttachments` carve-out to AGENTS.md:178 and the corresponding CLAUDE.md bullet, so all three docs (AGENTS.md, CLAUDE.md, docs/environment-variables.md) now agree that the create/edit routes bound this field by the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling, not the resolved admin value.

lib/components/post-box/post-box.tsx
llun: Fixed in 776b962c5: added a comment directly above the ref guard explaining it bounds `postExtensionRef` (written synchronously before dispatch) rather than the reducer-capped `postExtension`, and that it must not be deleted as a redundant "already checked twice" simplification.

lib/components/posts/status-reply-box.tsx
llun: Fixed in 776b962c5: added `StatusReplyBox attachment cap` in `status-reply-box.test.tsx` — renders under `InstanceLimitsProvider maxMediaAttachments={1}` and fires two overlapping picker batches, asserting the button stays at `(1/1)` and never reaches `(2/1)`. Proven by mutation: reverting the fix line back to `dispatch(addAttachment(attachment, 10))` fails this test.

app/(timeline)/layout.tsx
llun: Fixed in 776b962c5: `layout.test.tsx` now mocks `maxMediaAttachments` through to `data-max-media-attachments` and adds a test that seeds `posts.maxMediaAttachments` server setting to 7 and asserts it's published. Proven by mutation: deleting the line publishing `maxMediaAttachments` in `app/(timeline)/layout.tsx` fails this new test.

lib/components/post-box/post-box.tsx
llun: Fixed in 776b962c5: added `PostBox attachment ref guard` in `post-box.test.tsx`, using deferred `resizeImage` promises to sequence two overlapping picker batches under `maxMediaAttachments={1}`, then asserting only 1 attachment is uploaded/submitted. Proven by mutation: deleting the ref guard block makes `uploadAttachmentMock` fire twice instead of once, failing the test.

lib/components/post-box/upload-media-button.test.tsx
llun: Fixed in 776b962c5: `upload-media-button.test.tsx:164` now also renders at the cap (`maxMediaAttachments: 20`, 20 existing attachments) and asserts `.toBeDisabled()`, matching what the test's name already promised. Proven by mutation: hardcoding `disabled={false}` in `upload-media-button.tsx` fails this assertion.

lib/components/instance-limits.test.tsx
llun: Fixed in 776b962c5: shortened both `it.each` descriptions in `instance-limits.test.tsx` to "falls back to the default cap when omitted" and "falls back to the default cap for a zero value" — they now diverge within Vitest's 40-char truncation (verified: their first 40 characters differ).

docs/environment-variables.md
```diff
@@ -53,10 +53,22 @@ section: its pages and API keep working, so an existing link or bookmark still
...
-more than the configured number of attachments. Lowering it changes what
-clients are told, not what is accepted.
+`posts.maxMediaAttachments` is the exception, and the exception is wider than it
+looks: it is advertised to clients and honoured by this instance's own composer
+and inline reply box (which read it through `useInstanceLimits()`), but **no**
+route enforces the resolved value, and the routes do not even agree on a
+fallback. `POST`/`PUT /api/v1/statuses[/:id]` reject an attachment list longer
+than the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling (20) rather than the
+configured number. `POST /api/v1/accounts/outbox` — the endpoint the web
+composer actually posts through — enforces no attachment ceiling at all: its
```
llun: **The attachment-limit carve-outs understate the gap — the outbox route bounds attachment count by nothing at all**

The carve-outs added in 776b962c5 said the create and edit routes fall back to the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling, so a stale or missing provider could let media through "up to that fixed ceiling".

That is true for `POST`/`PUT /api/v1/statuses[/:id]` (route.ts:115 and [id]/route.ts:211,225 both reject above 20). It is NOT true for `POST /api/v1/accounts/outbox`, which is the route the web composer actually posts through (`lib/components/post-box/post-box.tsx` -> `createNote()` in `lib/client.ts` -> `fetch('/api/v1/accounts/outbox')`). That route enforces no attachment ceiling at all: `app/api/v1/accounts/outbox/types.ts` declares `attachments: PostBoxAttachment.array().optional()` with no `.max()`, the route checks neither constant, and `lib/actions/createNote.ts` (~556-570) maps every entry straight into `database.createAttachment`.

This mattered more after 776b962c5 because that commit pointed two other surfaces at `docs/environment-variables.md` as "the full explanation", making that paragraph authoritative. Compounding it, the same doc file (~lines 22-24) establishes "every endpoint that creates or edits a status" as explicitly including `POST /api/v1/accounts/outbox`, so a reader carries that grouping into the paragraph and concludes the worst case is 20 everywhere.

Consequence: on the path the web UI uses, the client cap this PR adds is the ONLY limit that exists. A stale or missing provider lets media through unbounded, not merely up to 20.

llun: Fixed in 9f290c56d. All four descriptions of this now say the same accurate thing: no route enforces the resolved value; `POST`/`PUT /api/v1/statuses[/:id]` fall back to the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling; `POST /api/v1/accounts/outbox` bounds the count by nothing, so on that path the client is the only limit. Updated in `docs/environment-variables.md`, the `InstanceLimits` JSDoc, `AGENTS.md` and `CLAUDE.md`. Docs only, no behaviour change — the unbounded outbox attachment list is a pre-existing bug and is tracked separately rather than widened into this PR. Full gate re-run green (853 files, 10,588 tests).

docs/environment-variables.md
```diff
@@ -53,10 +53,25 @@ section: its pages and API keep working, so an existing link or bookmark still
...
+composer-only — enforces no attachment ceiling at all: its
+request schema puts no `.max()` on `attachments`, the route checks neither
+constant, and `createNoteFromUserInput` maps every entry straight through to
+`database.createAttachment`. So lowering `posts.maxMediaAttachments` changes
+what clients are told and what the built-in composer offers, never what any of
+these routes accept — and on the outbox path there is currently no accepted
+attachment count limit whatsoever. Note that this is specific to the attachment
+count: the outbox route _does_ enforce the resolved `posts.maxCharacters` and
+`polls.*` through `validateStatusContentLimits`, which is why the general claim
+at the top of this section holds for every limit but this one.
```
llun, line 74: **Positional cross-reference pointed at the wrong paragraph**

The paragraph added in 9f290c56d ended with "Note the contrast with the paragraph above: that one is about `posts.maxCharacters` and `polls.*`...". The paragraph immediately above is the `features.*` navigation-flags one (line ~45); `network.linkPreviews` sits above that; the `posts.maxCharacters`/`polls.*` enforcement text is at line ~22, three paragraphs up. A reader following the pointer top-to-bottom lands on navigation flags.

It was also wrong on the day it was written, not merely fragile — and a positional reference across a multi-paragraph gap will not survive future insertions either. The same paragraph's opening ("is the exception") relied on the same loose positional link.

lib/components/instance-limits.tsx
```diff
@@ -26,12 +27,29 @@ import { MAX_FILE_SIZE } from '@/lib/services/medias/constants'
  * (`validateStatusContentLimits`) and on every upload endpoint
  * (`exceedsMaxMediaUploadSize`), so a missing or stale provider can only make
  * the client optimistic, never let something through.
+ *
+ * `maxMediaAttachments` is the one exception, and it is not a small one. No
+ * route enforces this resolved value. `POST`/`PUT /api/v1/statuses[/:id]`
+ * fall back to the fixed `MAX_STORED_MEDIA_ATTACHMENTS` ceiling, but
+ * `POST /api/v1/accounts/outbox` — the route both the composer and the inline
```
llun, line 34: **Unbounded outbox path described as composer-only, and "this composer" in a context module**

Two precision problems in the text added by 9f290c56d, both verified against the code:

1. All four surfaces (`docs/environment-variables.md`, this JSDoc, `AGENTS.md`, `CLAUDE.md`) framed the unbounded `POST /api/v1/accounts/outbox` path as the route "the web composer" posts through. The inline reply box posts through the identical route with identical exposure — `lib/components/posts/status-reply-box.tsx` imports the same `createNote` from `lib/client.ts`, which POSTs to `/api/v1/accounts/outbox`. Naming only the composer reads as though replies-with-attachments were bounded. They are not.

2. This JSDoc said "the route this composer actually posts through" while sitting in `lib/components/instance-limits.tsx`, which is the context/provider module, not a composer.

docs/environment-variables.md
llun: Fixed in 893a057b5. Both back-references in that paragraph now name what they refer to — the opening reads "the exception to the status-limit enforcement described at the top of this section", and the closer names `posts.maxCharacters`/`polls.*` and what the outbox route does enforce, instead of pointing at "the paragraph above". Survives paragraph insertions.

lib/components/instance-limits.tsx
llun: Fixed in 893a057b5. All four surfaces now name both the composer and the inline reply box, and the docs paragraph says explicitly that the exposure is not composer-only. The JSDoc no longer calls itself a composer. Docs and comments only, no behaviour change; full gate re-run green (853 files, 10,588 tests).

lib/services/medias/constants.ts
```diff
@@ -6,6 +6,11 @@ export const MAX_FILE_SIZE = 209_715_200
 // storing media the driver would refuse to serve. This ceiling exists because
 // the read path buffers an object in memory, so an unbounded cap is an OOM.
 export const MAX_CONFIGURABLE_FILE_SIZE = 1_073_741_824
+// How many photos a Strava import may attach to one activity post. This is an
+// import-side fan-out bound, NOT the composer's cap: the authoring UI caps
+// itself at the admin-configured `posts.maxMediaAttachments` read through
+// `useInstanceLimits()` (see lib/components/instance-limits.tsx), which is also
+// the value the instance entity advertises as `max_media_attachments`.
```
llun: Is this for mastodon? If so, isn't it should limit to 4? Also the constant name is not obvious for that purpose now.

llun: Good catch on both halves, and the answer to the first is no.

**It is not a Mastodon limit.** Mastodon's cap is `MAX_FEDERATION_MEDIA_ATTACHMENTS = 4`, and it is already applied — in `getNoteFromStatus` and in `toActivityPubObject`, both of which `.slice(0, MAX_FEDERATION_MEDIA_ATTACHMENTS)` when serialising a note outbound. That covers imported statuses like any other, so an imported activity federates 4 attachments no matter how many it stores, while local surfaces keep showing all of them. Capping storage at 4 would throw away photos that are visible in the app, which is the whole point of the stored-vs-federated split.

**Your instinct was still half right, though — the webhook path already does limit to 4**, via a *different* constant: `MAX_STRAVA_PHOTOS_TO_ATTACH = 4` in `importStravaActivityJob.ts:101` (undocumented, no comment). So in that path `MAX_ATTACHMENTS` was never the binding constraint. The archive path (`importStravaArchiveJob.ts`) has no photo-specific cap at all and fills up to this one. **That asymmetry looks unintentional** — a webhook import attaches at most 4 photos, an archive import of the same activity can attach ~9 (10 minus the route map). Worth a look, but it is a behaviour question rather than a naming one, so I have not touched it here.

**On the name — agreed, and my comment on it was wrong too.** I described it as "how many photos a Strava import may attach", but both call sites do `MAX_ATTACHMENTS - existingAttachments.length`, so it bounds the *status total* — route map included — not the photo count. Renamed to `MAX_IMPORTED_ACTIVITY_ATTACHMENTS` and the comment corrected, in 2049e68b6. Value stays 10, no behaviour change.

One thing worth flagging about that 10: it is inherited, not reasoned. `git log -S` puts its origin in #145 ("Improve media upload UX") — it was the *composer's* upload cap, and the import jobs borrowed it later in #354. This PR is what removed its last non-import consumer, so nothing justifies 10 any more except history. I have said so in the comment rather than silently changing it. If you would rather it tracked the real storage ceiling, `MAX_STORED_MEDIA_ATTACHMENTS` (20) is the obvious candidate — say the word and I will switch it.

PR #1506, opened 2026-08-22
fix: bound the public actor status page to the page it returns

llun:
## Problem

`getActorStatuses({ publicOnly: true })` backs the signed-out profile page (`GET /api/v1/accounts/:id/statuses`) and the ActivityPub outbox page. It attached readability as `whereIn('statuses.id', <recursive CTE over the actor's entire history>)`.

That subquery is uncorrelated, so PostgreSQL has to finish it before `LIMIT` sees a single row. The cost is linear in the actor's whole public history on every page load, and grows forever. Production measured **116.9 ms / 25,066 buffers** for one 30-row page: the CTE built over all 1,928 of the actor's public statuses, ran 3,918 `recipients` index searches inside it, HashAggregated the ids, nested-loop probed `statuses_pkey` 1,928 times, and only then top-N heapsorted to 30.

Same class as #1484 fixed for the local public timeline.

## Fix

Add `wherePubliclyReadableStatus` — the same predicate, correlated to the row the outer query is already looking at — so the ordered index scan can early-terminate:

```
where actorId = ?
  and exists (<public recipient for statuses.id>)
  and (statuses.type <> 'Announce' or exists (<recursive chain from statuses' pointer>))
order by createdAt desc, id desc limit 30
```

`getStatusReplies` and `getStatusRepliesCount` move over too.

**`getActorStatusesCount` and `getRebloggedBy` deliberately keep the set form**, for two different measured reasons — see the table below. Both spellings share `announceOriginalPointer`, and an equivalence test pins them together.

> **Review round 1 changed this PR.** `getRebloggedBy` was originally moved onto the correlated form too, and that was a 12x wall-clock regression. It is reverted, and the shape test now pins it. Details below.

No migration: `statuses_actorId_idx` and `recipients_actorId_statusId_idx` already serve the new plan.

## Measurements

PostgreSQL 18.6 in a container, seeded to production's shape — 176,000 statuses, 14,900 actors, 435,500 recipients, target actor with 2,097 statuses (1,889 publicly readable, half Announces), rows inserted in `createdAt` order so they scatter across the heap. Median of 5 runs, measured on this branch's schema (post-#1499).

A seed reproduces production's row counts, not its statistics. What it does reproduce here is the **plan**: same parallel recursive CTE, ~4,100 `recipients` index searches against production's 3,918, and a 1,889-row HashAggregate plus pkey nested loop against production's 1,928.

### The page query

| limit | before | after |
|---|---|---|
| 20 | 33,308 buf / 23.3 ms | **325 buf / 0.46 ms** |
| 30 | 33,311 buf / 24.1 ms | **476 buf / 0.63 ms** |
| 40 | 33,311 buf / 24.2 ms | **573 buf / 0.74 ms** |
| deep page (cursor 1,500 rows back) | 33,311 buf / 22.0 ms | **1,888 buf / 2.2 ms** |

Before is flat at every page size — the limit bounded nothing. After scales with it. The plan is now `Index Scan Backward using statuses_actorId_idx` → `Incremental Sort` (for the `id DESC` tiebreak the index does not carry) → `Limit`, with `Heap Fetches: 0` on the recipients semi-join.

Deep pages cost more because the cursor's `OR` spelling is a filter rather than an index start condition, so the scan still walks from the head. That is linear in page depth instead of constant-and-huge, and still 17× better.

### The count query — why it was left alone

| | buffers | exec |
|---|---|---|
| set form (kept) | 25,752 | **18.3 ms** |
| correlated (rejected) | 15,421 | **54.1 ms** |

### The other three call sites

Measured from the **real compiled SQL**, captured off the actual code path rather than reconstructed by hand:

| call site | before (set) | after | outcome |
|---|---|---|---|
| `getStatusReplies` | 1,467 buf / 1.34 ms | 1,451 buf / 1.17 ms | moved, neutral |
| `getStatusRepliesCount` | 1,464 buf / 1.27 ms | 1,448 buf / 1.06 ms | moved, neutral |
| `getRebloggedBy` | 49,474 buf / 33 ms | 11,383 buf / **415 ms** | **reverted** |

`getRebloggedBy` embeds its filtered reblog set **twice** — as the `visible_reblogs` FROM subquery and again inside the correlated `whereNotExists` that dedupes to each actor's newest reblog — so a correlated predicate is re-evaluated per row *pair* while a materialised id set is computed once and shared. Its candidate set is not this status's reblogs either: `reblogBase`'s legacy `originalStatusId IS NULL` branch admits every pre-backfill Announce in the instance. It now matches `origin/main` exactly (49,474 buffers, identical 40 rows).

The figure first quoted for it here (3,206 → 3,040 buffers) was wrong: it came from a hand-written approximation that omitted the legacy `OR` branch and so never reproduced the real plan. Every number in this PR is now taken from SQL captured off the running code.

The rule this settles: **what costs is the number of evaluations, not the absence of a `LIMIT`.** `getStatusRepliesCount` is an unbounded `COUNT` that legitimately uses the correlated form, because one status's replies are a small set.

### A rejected optimisation worth recording

Adding a depth-1 fast-path `EXISTS` beside the recursive chain — check the common case cheaply, recurse only when needed — makes things dramatically **worse**. PostgreSQL decorrelates it into a **hashed SubPlan**: a sequential scan of every non-Announce status in the instance, hashed against every public recipient row, to answer one page. 476 buffers → 11,294 and 0.63 ms → 82 ms; the count went to 123 ms. The recursive CTE resists that rewrite, which is the only reason the predicate reads as one branch rather than two. Documented in `AGENTS.md` so nobody re-derives it.

Bounding the existing set builder to a page window was also measured (1,272 buf / 1.4 ms at a 60-row window) and rejected: it scales with the window, needs a JS top-up loop, and needs the other filters duplicated into the window query.

## Correctness

The two forms must never disagree, so `publiclyReadableStatusEquivalence.test.ts` runs both over the same rows and compares. Fixtures cover a public note, a followers-only note, a boost of each, a followers-only boost, a **depth-2 boost-of-boost**, a chain through a **non-public middle hop**, a **legacy `content`-only pointer**, **disagreeing pointer columns** (only `originalStatusId` names the real target), a **cycle**, a self-referencing boost, and a dangling pointer.

On the seeded database the same check ran across all 176,000 statuses: both forms selected exactly the same 140,333 rows, zero disagreements in either direction, with 79 announce-of-announce rows and a cycle present. The shipped query returns the identical 30 rows in the identical order as before the change.

`UNION` (not `UNION ALL`) is what terminates a cycle. Verified by mutation: switching to `UNION ALL` hangs the suite rather than failing it — same property the existing set form has.

## Why the shape test exists

`publiclyReadableStatusQueryShape.test.ts` pins the compiled SQL, following `localPublicQueryShape.test.ts`. Verified by mutation: reverting `getActorStatuses` to the set form leaves **all 169 tests in `status.test.ts` and the whole equivalence suite green** — the two forms select identical rows, which is the point — and fails only the two assertions that read the SQL. Without it the revert lands silently.

It also pins the set form for `getActorStatusesCount` and `getRebloggedBy`, so "finishing the unification" has to delete an assertion that explains why not — verified by mutation for the reblog case, which is the regression round 1 caught.

`publiclyReadableStatusCallSites.test.ts` closes a separate gap review found: **no call site had a behavioural test that `publicOnly` excludes a private row.** Deleting the filter outright from `getStatusReplies`, `getStatusRepliesCount` or `getRebloggedBy` left the whole suite green. `getStatusRepliesCount` is the one with no backstop — the ActivityPub handlers re-filter `getStatusReplies`'s array with `isStatusPubliclyReadable`, but the count flows straight into the replies collection's `totalItems`. For `getActorStatuses` the distinguishing fixture is a **public boost of a followers-only note**, which the recipients-only fallback returns and only the announce chain drops. All five are mutation-proven.

## Verification

- `prettier --check` → `lint` → `typecheck` → `build` → `test` all pass (856 files, 10,624 tests), on the current `main` (v1.144.0).
- `TEST_DATABASE_TYPE=pg` against PostgreSQL 18.6: `lib/database` 81 files / 1,438 tests pass. CI pins SQLite, and this is a plan-sensitive change.
- Mutation-tested: dropping the per-hop public check fails 3 equivalence tests; reverting the call site fails only the shape test.
- No UI change, so no screenshots.
- One unrelated flake seen once in a full-suite run (`app/(timeline)/[actor]/ActorTimelines.test.tsx`); it passes in isolation on both this branch and a clean baseline, and the full suite passed on re-run.

## Review

The sub-agent review loop ran eight rounds (correctness, quality, testing and scope/regression, plus a dedicated privacy-boundary pass on the final state). All 20 findings are posted as inline threads and resolved. Rounds 4 and 8 were clean; round 7 produced only a cosmetic simplification, which is where it converged.

Four findings changed the shipped code rather than its comments:

- **Round 1** caught the `getRebloggedBy` regression above — the reason that call site is excluded from the change at all.
- **Round 1** also found that no `publicOnly` call site had a behavioural test: deleting the filter outright from three of them left the entire suite green. `publiclyReadableStatusCallSites.test.ts` exists because of that.
- **Round 5** found that `publicOnly` combined with `excludeReblogs` / `pinned` / `tagged` — the argument shape the Mastodon accounts endpoint actually sends — had no coverage at all.
- **Round 6** showed by mutation that those new cases were too weak: asserting absences plus a non-empty result let a regression that drops the pin correlation *only* when `publicOnly` is true pass 177/177. They now assert exact sets.

Every test claim in this PR is mutation-proven: reverting each call site, and breaking the predicate three different ways, was checked to fail the specific assertion that should catch it and nothing else — on both SQLite and PostgreSQL 18.

A **pre-existing** issue found in passing and deliberately left out of this PR: `app/(timeline)/[actor]/getProfileData.ts:58` calls `getActorStatuses` with no visibility argument at all, and `getActorStatuses` applies no visibility filter unless one of `publicOnly` / `visibleToActorId` / `includeFollowersOnly` is set. Confirmed at the database layer that this call shape returns followers-only rows. Not touched here; it needs its own change and its own browser verification.

Found while investigating #1499 and deliberately left out of that PR's scope.


lib/database/sql/status.ts
```diff
@@ -2639,10 +2596,27 @@ export const StatusSQLDatabaseMixin = (
...
+        // inside the correlated `whereNotExists` that keeps only each actor's
+        // newest reblog. A correlated readability predicate is therefore
+        // re-evaluated per row PAIR, instantiating one recursive CTE each time,
+        // while a materialised id set is computed once and reused by both
+        // copies. The candidate set is also not this status's reblogs alone:
+        // the legacy `originalStatusId IS NULL` branch of `reblogBase` admits
+        // every Announce in the instance that predates the backfill, so the
+        // correlated cost scales with instance-wide legacy Announce volume
+        // rather than with this status. Measured on the production-shaped
+        // PostgreSQL 18 seed: 49,474 buffers / 33ms set-based against 11,383 /
```
llun: **Correctness/performance: `getRebloggedBy` was moved onto the correlated predicate — a 12x wall-clock regression.**

On the production-shaped PostgreSQL 18 seed: 49,474 buffers / 33ms set-based vs 11,383 buffers / 415ms correlated.

Cause: `visibleReblogsQuery` is embedded twice — once as the `visible_reblogs` FROM subquery, and again inside the correlated `whereNotExists` that dedupes to each actor's newest reblog — so a correlated predicate is re-evaluated per row PAIR, while a materialised id set is computed once and shared by both copies. The originally-quoted figure (3,206 → 3,040 buffers) came from a hand-written approximation that omitted the legacy `originalStatusId IS NULL` branch.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,203 @@
+import { getTestDatabaseWithInstance } from '@/lib/database/testUtils'
+import { ACTIVITY_STREAM_PUBLIC } from '@/lib/utils/activitystream'
+
+// Every `publicOnly` call site must actually drop a followers-only row.
```
llun, line 4: **Missing behavioural test: no `publicOnly` call site had a test proving the filter actually excludes a non-public row.**

Deleting the filter outright from `getStatusReplies`, `getStatusRepliesCount`, or `getRebloggedBy` left the entire suite green. `getStatusRepliesCount` has no app-level backstop — its number is the ActivityPub replies collection's `totalItems`.

lib/database/sql/utils/publiclyReadableStatus.ts
```diff
@@ -0,0 +1,182 @@
...
+
+// "This status is publicly readable", as a predicate correlated to the row the
+// outer query is already looking at.
+//
+// Written for knex's `modify`, which keeps it chainable mid-builder:
+//
+//     query.modify(wherePubliclyReadableStatus, database)
+//     query.modify(wherePubliclyReadableStatus, database, 'reply_statuses')
+//
+// Reach for this form when the number of rows the predicate is evaluated over
```
llun, line 138: **Docstring overstates the rule as LIMIT-based, which `getStatusRepliesCount` appears to violate.**

The module states a blanket rule ("this is the form for any query a LIMIT bounds; keep the set form only where the query has to touch every row anyway"), but `getStatusRepliesCount` is an unlimited COUNT using the correlated form — so a reader working from source alone would read that call site as violating the module's own rule.

lib/database/sql/utils/publiclyReadableStatus.ts
```diff
@@ -0,0 +1,182 @@
...
+
+// "Addressed to the public collection", correlated to one status row.
+//
+// Spelled as an EXISTS rather than the set-based
+// `id IN (SELECT statusId FROM recipients …)` the count path uses, because the
+// point of this whole module's correlated form is that nothing materialises
+// ahead of the outer row. On PostgreSQL this is served by
+// `recipients_actorId_statusId_idx` at `Heap Fetches: 0`.
+//
+// `publicRecipientStatusIds` in `status.ts` is the same test spelled as a
```
llun, line 45: **Two live spellings of "has a public recipient" with no comment explaining why they weren't converged.**

`publicRecipientExistsQuery` here and `publicRecipientStatusIds` in `status.ts` say the same thing two different ways, with nothing recorded about why — notable on a PR that did converge `announceOriginalPointer`.

lib/database/sql/status.test.ts
```diff
@@ -451,12 +451,13 @@ describe('StatusDatabase', () => {
     })
 
     it('keeps a boosted reply out of the public replies count', async () => {
-      // `getStatusRepliesCount` cannot be collapsed onto the readable-ids
-      // subquery the way `getActorStatusesCount` was: its
-      // `whereNot('type', Announce)` filters the REPLY row, while the
-      // subquery's `type` is the resolved original's, so an Announce whose
-      // original is a public Note comes back from the subquery and only the
-      // outer filter drops it. `createAnnounce` always writes `reply: ''`, so
+      // `getStatusRepliesCount`'s `whereNot('type', Announce)` filters the
```
llun, line 454: **Stale comment: describes `getStatusRepliesCount` as filtered by "the readable-ids subquery", a path it no longer takes.**

This PR moved `getStatusRepliesCount` onto the correlated `wherePubliclyReadableStatus` predicate, but this comment still describes the old subquery-based filtering. AGENTS.md "Documentation Maintenance" requires fixing docs a change makes stale in the same PR.

AGENTS.md
```diff
@@ -1967,6 +1967,10 @@ The app (`yarn migrate`) runs Knex migrations, but the test suite does **not** 
...
+- **`publiclyReadableStatusQueryShape.test.ts` pins which form each call site uses, and it is the only thing that would catch a swap in either direction.** Putting `getActorStatuses` back on `whereIn(id, buildPubliclyReadableStatusIdsQuery(…))` leaves all 169 tests in `status.test.ts` and the whole equivalence suite green — the two forms select identical rows, which is the point — and fails exactly the two assertions that read the SQL. It pins the set form for `getActorStatusesCount` and `getRebloggedBy` the same way. Separately, `publiclyReadableStatusCallSites.test.ts` asserts each `publicOnly` call site actually drops a followers-only row: deleting the filter outright from `getStatusReplies`, `getStatusRepliesCount` or `getRebloggedBy` used to leave the entire suite green. `getStatusRepliesCount` is the one with no backstop — the ActivityPub route handlers re-filter `getStatusReplies`'s array with `isStatusPubliclyReadable`, but the count flows straight into the replies collection's `totalItems`. For `getActorStatuses` the distinguishing fixture is a **public boost of a followers-only note**: the recipients-only fallback returns it and only the announce chain drops it.
```
llun, line 2023: **Two factual errors in this new prose.**

1. The function is called `getStatusReblogs` here, but the real symbol is `getRebloggedBy` — a different `getStatusReblogsCount` also exists, so the wrong name is actively misleading in a file whose job is to be greppable.
2. States "168 tests in status.test.ts" — the suite has 169.

lib/database/sql/status.ts
llun: Fixed in b03879a99: reverted `getRebloggedBy` to the set-based readability form. It is now byte-identical in plan to origin/main (49,474 buffers, same 40 rows). Pinned by a query-shape test, mutation-verified.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Addressed in b03879a99: added this new test file with five mutation-proven cases, including a public boost of a followers-only note for `getActorStatuses` — the only fixture that distinguishes the readability predicate from the recipients-only fallback.

lib/database/sql/utils/publiclyReadableStatus.ts
llun: Fixed in b03879a99: the rule is now stated in terms of the number of evaluations rather than the presence of a LIMIT, with both set-form exceptions and their measurements named.

lib/database/sql/utils/publiclyReadableStatus.ts
llun: Addressed in b03879a99: added a comment explaining the split is deliberate (converging would re-plan the query #1493 tuned, for no gain) and why these two cannot drift semantically the way `announceOriginalPointer` could.

lib/database/sql/status.test.ts
llun: Fixed in b03879a99: rewritten to describe the correlated predicate while keeping the historical reason the filter is not redundant.

AGENTS.md
llun: Fixed in b03879a99: both corrected — `getStatusReblogs` no longer appears in AGENTS.md or CLAUDE.md, and the test count now reads 169.

lib/database/sql/publiclyReadableStatusQueryShape.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+  ])(
+    'keeps the set-based readable-id form for $description',
+    async ({ run }) => {
+      const statements = await capture(run)
+
+      const matches = statements.filter((sql) =>
+        sql.includes('publicly_readable_statuses')
+      )
+      expect(matches).toHaveLength(1)
```
llun, line 155: **Test relaxed under `it.each` conversion: lost its ability to catch a regression.**

Converting this test to `it.each` silently relaxed `expect(matches).toHaveLength(1)` to `expect(matches.length).toBeGreaterThanOrEqual(1)`. The relaxation buys nothing — both call sites (`getActorStatusesCount` and `getRebloggedBy`) emit exactly one matching statement — and it loses the ability to catch a future change that splits the set-based subquery into a second round trip.

lib/database/sql/status.ts
```diff
@@ -2628,21 +2585,47 @@ export const StatusSQLDatabaseMixin = (
...
-      : applyPublicReadableStatusFilter({
-          query: visibleReblogsQuery,
-          targetStatusIds: reblogTargetStatusIds
+    if (visibleToActorId) {
+      visibleReblogsQuery = applyPotentiallyReadableStatusFilter({
+        query: visibleReblogsQuery,
+        visibleToActorId
+      })
+    } else {
+      // The set-based readability form, for the same reason
```
llun, line 2651: **Fifteen-line rationale living inside a ternary's `:` arm hides the one line of logic that matters.**

The comment explaining why the set-based form is kept here is buried in the `:` arm of a ternary, ahead of the single `whereIn(...)` call it explains. Restructure to if/else so the comment sits above the statement it justifies, and move `reblogTargetStatusIds` inline into the branch that actually uses it.

lib/database/sql/status.ts
```diff
@@ -2628,21 +2585,47 @@ export const StatusSQLDatabaseMixin = (
...
+      // `whereNotExists` that keeps only each actor's newest reblog. A
+      // correlated readability predicate is therefore re-evaluated per row
+      // PAIR, instantiating one recursive CTE each time, while a materialised
+      // id set is computed once and reused by both copies.
+      //
+      // The rows it would be evaluated over are not this status's reblogs
+      // either. `reblogBase`'s legacy `originalStatusId IS NULL` branch has no
+      // index to narrow it — `statuses_announce_original_actor_idx` covers
+      // `(type, "originalStatusId", "actorId")` and nothing indexes `content` —
+      // so every pre-backfill Announce in the instance is a candidate row the
```
llun, line 2664: **Wording implies the legacy pre-backfill Announces come back in the result — they don't.**

"admits every Announce in the instance that predates the backfill" reads as if those rows are part of the result set. They aren't: the result stays scoped to this status. The real point is that nothing indexes `content`, so they're candidate rows the planner has to filter through — reword to say "candidate row" explicitly and state the result set is unaffected, only the work to reach it.

lib/database/sql/publiclyReadableStatusQueryShape.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      // A COUNT has to touch every row anyway, and the correlated form costs
+      // one recursive-CTE instantiation per Announce: on the production-shaped
+      // seed it read 40% fewer buffers and still took three times as long
+      // (54ms against 18ms). Reasoning at `getActorStatusesCount`.
+      run: () => database.getActorStatusesCount({ actorId, publicOnly: true })
+    },
+    {
+      description: 'the anonymous reblogged-by list',
+      // Shipped correlated on this branch and caught in review, which is why
+      // it is pinned. Reasoning and measurements at `getRebloggedBy` in
```
llun, line 143: **Same `getRebloggedBy` measurement pair restated in full for a fourth time.**

This entry's comment repeats the buffer/wall-clock numbers already stated at the call site in `status.ts`. Its sibling entry in the same table (`getActorStatusesCount`) already points at the rationale instead of restating it ("Reasoning at `getActorStatusesCount`") — make this entry consistent and point at `status.ts` instead of repeating the numbers.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,209 @@
...
+  const publicOwnId = `${actorId}/statuses/public-own`
+  const privateOwnId = `${actorId}/statuses/private-own`
+  // A PUBLIC boost of the actor's own FOLLOWERS-ONLY note. This is the row that
+  // separates the readability predicate from the plain "addressed to the public
+  // collection" filter `getActorStatuses` falls back to for a signed-in viewer:
+  // the boost itself is public, so the recipients test alone lets it through,
+  // and only following the announce chain to a non-public original drops it.
+  const publicBoostOfPrivateId = `${actorId}/statuses/public-boost-of-private`
+
+  const audience = (isPublic: boolean, followersUrl: string = followers) =>
```
llun, line 47: **`audience()` hardcodes the parent actor's followers collection; two call sites reinvent it.**

`audience()` here only ever returns the outer `followers` URL for the non-public branch, so the booster fixtures further down inline an equivalent ternary by hand — the same one-liner spelled two ways across three fixture loops in this file. Parametrize `audience` with an optional `followersUrl` and use it at all three sites.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,209 @@
...
+  })
+
+  it('drops a public boost of a followers-only note from the public actor status page', async () => {
+    // The recipients-only fallback would return this row: the boost is
+    // addressed to the public collection. Only following the announce chain to
+    // a non-public original excludes it, so this is the case that distinguishes
+    // the readability predicate from that fallback — the sibling test above,
+    // with a plainly private note, passes under that fallback too.
+    //
+    // `status.test.ts`'s 'excludes nested announces when the boosted original
```
llun: **This scenario duplicates coverage that already exists elsewhere in the suite.**

This case is already covered by `status.test.ts`'s "excludes nested announces when the boosted original is not publicly readable", and at predicate level by `publiclyReadableStatusEquivalence.test.ts` — it's the one test in this file that isn't filling a gap. Worth a comment saying so explicitly so a future reader doesn't mistake it for gap-filling and doesn't try to "fix" the apparent duplication.

lib/database/sql/status.ts
```diff
@@ -1514,6 +1462,18 @@ export const StatusSQLDatabaseMixin = (
       // `statuses` by it re-applied the same `actorId` predicate through one
       // primary-key lookup per returned id, for a result the subquery had
       // already decided.
+      //
+      // One of the two readability call sites that keep the set-based form
+      // (`getRebloggedBy` is the other, for a different reason recorded there).
+      // The correlated `wherePubliclyReadableStatus` costs one recursive-CTE
+      // instantiation per Announce row — a LIMIT amortises that to a page, and
+      // a COUNT over the actor's whole history cannot amortise it at all. On
```
llun, line 1524: **Comment line runs to ~130 characters after a clause was appended, breaking this block's wrapping.**

The rest of this comment block wraps carefully; this line was left long after a clause was appended without rewrapping. Rewrap to match.

lib/database/sql/status.test.ts
```diff
@@ -451,14 +451,16 @@ describe('StatusDatabase', () => {
...
-      // outer filter drops it. `createAnnounce` always writes `reply: ''`, so
-      // no number anyone can observe moves if that filter goes — only a
-      // hand-written row pins the invariant.
+      // `getStatusRepliesCount`'s `whereNot('type', Announce)` filters the
+      // REPLY row, and is not redundant with the readability predicate: an
+      // Announce whose boosted original is a public Note is publicly readable,
+      // so only the outer filter keeps it out of a reply count. That held when
+      // the filter was the readable-ids subquery (whose `type` was the resolved
+      // original's, a different row again) and still holds now that
+      // `getStatusRepliesCount` uses the correlated
```
llun, line 460: **Comment line runs to ~110 characters after `wherePubliclyReadableStatus` was appended, breaking this block's wrapping.**

Same issue as the block in `status.ts`: this line grew past the wrapping the rest of the comment uses once the predicate name was appended inline. Rewrap.

lib/database/sql/publiclyReadableStatusQueryShape.test.ts
llun: Fixed in c4ac80a7e: restored strict `expect(matches).toHaveLength(1)`. Verified both `it.each` rows (the public status count and the anonymous reblogged-by list) still pass under the strict assertion.

lib/database/sql/status.ts
llun: Fixed in c4ac80a7e: restructured to if/else so the comment sits above the statement it explains, and moved `reblogTargetStatusIds` inline into the only branch that uses it. Behaviour-neutral — captured the compiled SQL off the real code path before and after; byte-identical to origin/main.

lib/database/sql/status.ts
llun: Fixed in c4ac80a7e: reworded to say "candidate row" explicitly and to state the result set is unaffected — only the work to reach it.

lib/database/sql/publiclyReadableStatusQueryShape.test.ts
llun: Fixed in c4ac80a7e: this entry now points at the rationale in `status.ts` instead of restating it, matching its sibling entry's style.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Fixed in c4ac80a7e: parametrized as `audience(isPublic, followersUrl = followers)` and used at all three fixture sites.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Kept, per the finding: the file's job is to state the same invariant once per call site in one place, so this stays even though it isn't filling a gap. Fixed in c4ac80a7e by adding a comment that says so explicitly, pointing at the two other places this scenario is already covered.

lib/database/sql/status.ts
llun: Fixed in c4ac80a7e: rewrapped to match the rest of the block.

lib/database/sql/status.test.ts
llun: Fixed in c4ac80a7e: rewrapped to match the rest of the block.

lib/database/sql/utils/publiclyReadableStatus.ts
```diff
@@ -0,0 +1,178 @@
...
+// size, against 325/476/573 here at limits of 20/30/40.
+//
+// Keep the SET form in two cases, both measured on that seed:
+//
+//   - The predicate has to be evaluated over a large unbounded set anyway. A
+//     COUNT cannot amortise the per-Announce recursive-CTE instantiation the
+//     way a LIMIT does — see the note at `getActorStatusesCount`.
+//   - The query embeds the predicate MORE THAN ONCE, so a correlated form is
+//     re-evaluated per row PAIR while a materialised id set is computed once
+//     and shared — see the note at `getRebloggedBy` in `status.ts`.
```
llun, line 153: **Round 3 finding (already fixed):** the two "keep the SET form" bullets in this docblock did not follow the same discipline — the first defers to `getActorStatusesCount` in one line, but the second restated the entire `getRebloggedBy` rationale plus the exact measurement pair (49,474/33ms vs 11,383/415ms) that already lives in fuller form at that call site in `status.ts`. Round 2 already established the pointer pattern for this exact text in `publiclyReadableStatusQueryShape.test.ts`, so this module docstring was the last place still restating it.

Fixed in 59860468f: the second bullet now points at `getRebloggedBy` in `status.ts` instead of restating, matching its sibling. The same commit also corrected an undercount in `publiclyReadableStatusCallSites.test.ts` — both `getActorStatuses` cases there are already covered in `status.test.ts` (by 'excludes nested announces when the boosted original is not publicly readable' and 'counts only publicly readable actor statuses when requested'), not just one.

llun: Addressed in 59860468f. Round 3's other three lenses (correctness, scope/regression, testing) returned clean with no findings — this docstring-discipline finding from the quality lens was the only one this round. Resolving.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,273 @@
...
+  // `AND`s today, so none of these failed when first written — this is coverage
+  // for a real argument shape, not a fix. `onlyMedia` is left out because an
+  // attachment fixture needs a `medias` row whose id is a valid integer on
+  // PostgreSQL, which buys no additional composition signal.
+  it.each([
+    { description: 'excludeReblogs', filters: { excludeReblogs: true } },
+    { description: 'pinned', filters: { pinned: true } },
+    { description: 'tagged', filters: { tagged: '#combo' } }
+  ])(
+    'still drops followers-only rows with publicOnly and $description',
```
llun, line 248: **Testing lens (round 5): coverage gap for the real production argument shape**

`app/api/v1/accounts/[id]/statuses/route.ts` passes `publicOnly` together with `excludeReblogs`, `pinned`, `tagged` and `onlyMedia` in a single call — that is the actual production argument shape — but nothing in this suite exercised `publicOnly: true` in combination with any of the other filters. This PR rewrites exactly the predicate `publicOnly` relies on, so a composition regression in the rewrite would have shipped unnoticed.

Please add coverage for `publicOnly` combined with at least `excludeReblogs`, `pinned`, and `tagged`.

lib/database/sql/utils/publiclyReadableStatus.ts
```diff
@@ -0,0 +1,184 @@
...
+    .select(database.raw('1'))
+    .from('statuses as original_statuses')
+    .where('original_statuses.id', announceOriginalPointer(database, table))
+    .whereNot('original_statuses.type', StatusType.enum.Announce)
+    .whereExists(publicRecipientExistsQuery(database, 'original_statuses.id'))
+
+// "This status is publicly readable", as a predicate correlated to the row the
+// outer query is already looking at.
+//
+// Written for knex's `modify`, which keeps it chainable mid-builder:
```
llun, line 127: **Docs (raised independently across rounds 1, 4 and 5): misleading `modify` usage example**

The docblock advertised `query.modify(wherePubliclyReadableStatus, database, 'reply_statuses')`. No call site in the codebase actually passes a third argument, and `reply_statuses` is not an alias used anywhere — so the example showed a call that does not exist. Worse, passing an alias the outer query does not expose is a runtime SQL error (correlated references resolve to nothing), not a wider result set, which is a sharp edge for the next reader. Three separate review passes read this as a real call site.

Please trim the example to the real two-argument form and note what the third argument requires when it is used.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Fixed in 6a9331238. Added an `it.each` covering `publicOnly` combined with `excludeReblogs`, `pinned`, and `tagged` (the actual argument shape `app/api/v1/accounts/[id]/statuses/route.ts` passes). They compose as independent top-level ANDs and all three passed first time, so this is coverage rather than a behavioral fix. The pinned case pins the public boost of a followers-only note, which is what lets it distinguish the readability predicate from the plain "addressed to the public collection" fallback — mutation-verified: forcing that fallback fails the pinned case, and making the filter match nothing fails all three via their non-empty-list guard. `onlyMedia` was left out on purpose: an attachment fixture needs a `medias` row whose id is a valid integer on PostgreSQL, and it would add no composition signal beyond the other three.

Also from round 5: a dedicated privacy-boundary review ran a 200-row adversarial corpus against both SQLite and PostgreSQL 18.6, plus an independent reference implementation, and found no leak and no divergence between the two predicate forms. Round 4 was clean across all four review lenses.

lib/database/sql/utils/publiclyReadableStatus.ts
llun: Fixed in 6a9331238. The docblock now shows only the real two-argument form (`query.modify(wherePubliclyReadableStatus, database)`); the prose explains the optional third argument must name an alias the outer query actually exposes, that passing one that is not in scope is a runtime SQL error rather than a silently wider result, and that no caller needs it today.

Also from round 5: a dedicated privacy-boundary review ran a 200-row adversarial corpus against both SQLite and PostgreSQL 18.6, plus an independent reference implementation, and found no leak and no divergence between the two predicate forms. Round 4 was clean across all four review lenses.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,290 @@
...
+      expected: () => [publicOwnId]
+    },
+    {
+      description: 'tagged',
+      filters: { tagged: '#combo' },
+      // Two rows carry the tag; only this one is publicly readable.
+      expected: () => [taggedPublicId]
+    }
+  ])(
+    'still drops followers-only rows with publicOnly and $description',
```
llun, line 248: **Testing: assertions too weak to catch the regression they exist for.**

The three `publicOnly` combination cases (`excludeReblogs`, `pinned`, `tagged`) only assert that specific ids are absent, plus `ids.length > 0`. That doesn't prove the filter combination is wired correctly: dropping `status_pins.statusId = statuses.id` from the join *only* when `publicOnly` is also true still satisfies every absence assertion here while leaving the list non-empty — and it also passes `status.test.ts`'s dedicated pinned test, because that test never sets `publicOnly`. So this single combination-specific regression slips through both files at once, 177/177 green.

Please assert the exact (sorted) set of surviving ids for each case instead of absence + non-empty.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,290 @@
+import { getTestDatabaseWithInstance } from '@/lib/database/testUtils'
+import { ACTIVITY_STREAM_PUBLIC } from '@/lib/utils/activitystream'
+
+// Every `publicOnly` call site must actually drop a followers-only row.
```
llun, line 4: **Scope/regression: PR description has drifted from the actual review history.**

The Review section of the PR body still claims the loop ran five rounds, lists all 15 findings as resolved, and describes round 4 as though the loop had concluded there — while the round-5 commit's own message already describes two further findings, and this round (6) adds two more on top of that. The verification counts (files/tests, PostgreSQL files/tests) are also stale by several tests against a fresh full run.

(Anchoring here at the top-of-file docblock since this finding is about the PR description, not a source line in this file.)

Please rewrite the Review section to state the actual round count, note that round 4 is the only clean round so far, list the findings that changed shipped code, and call out the pre-existing `getProfileData` visibility issue as explicitly out of scope. Refresh the verification counts too.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Addressed. Each combination case (`excludeReblogs`, `pinned`, `tagged`) now asserts the exact sorted set of surviving ids instead of absence + non-empty — `excludeReblogs` expects every publicly readable non-Announce the actor owns, `pinned` expects the one publicly readable row of the three pinned, `tagged` expects the one publicly readable row of the two tagged. Re-ran the mutation described above (dropping the pin-correlation only under `publicOnly`): it now fails the pinned case and nothing else.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
llun: Addressed. The Review section now states the actual round count (six so far), notes round 4 as the only clean round so far and that the loop continues until two consecutive clean rounds, lists the four findings that changed shipped code, and calls out the pre-existing `getProfileData` visibility issue at `app/(timeline)/[actor]/getProfileData.ts:58` as explicitly out of scope. Verification counts refreshed to 856 files / 10,624 tests and 81 files / 1,438 PostgreSQL tests.

lib/database/sql/publiclyReadableStatusCallSites.test.ts
```diff
@@ -0,0 +1,290 @@
...
+      description: 'excludeReblogs',
+      filters: { excludeReblogs: true },
+      // Every publicly readable non-Announce the actor owns.
+      expected: [parentId, publicReplyId, publicOwnId, taggedPublicId]
+    },
+    {
+      description: 'pinned',
+      filters: { pinned: true },
+      // Three rows are pinned; only this one is publicly readable.
+      expected: [publicOwnId]
```
llun, line 239: **[Testing][Non-blocking]** Round 6 introduced these expected sets as thunks (`expected: () => [...]`), unwrapped at the assertion as `[...expected()]`. The laziness bought nothing: an `it.each` table is evaluated synchronously at collection time, and every id here is a plain template-literal string constant declared at the top of the describe block, not an async DB-generated value. A plain array behaves identically, so the thunk was unnecessary indirection.

llun: Fixed in 3574448ca: the thunks are gone — `expected: [...]` with `[...expected].sort()` at the assertion. Re-ran all three mutations the cases guard against afterwards and confirmed they are still caught: dropping the pin correlation only when `publicOnly` is true fails the pinned case alone; forcing the recipients-only fallback fails the pinned case plus the standalone boost-of-a-private-note test; making the readability filter match nothing fails all three combination cases plus the followers-only-status test.

Separately, round 7's scope/regression lens came back clean with no findings — it independently re-verified the PR body's claims and verification counts against HEAD and found them accurate.

PR #1507, opened 2026-08-22
minor: raise the stored media attachment ceiling to 50

llun:
## Summary

Raises `MAX_STORED_MEDIA_ATTACHMENTS` (`lib/services/mastodon/constants.ts`) from **20** to **50**. That constant is the hard safety ceiling on how many media a single status may store and accept through the create/edit routes — `POST`/`PUT /api/v1/statuses[/:id]` and `POST /api/v1/accounts/outbox` all `422` above it.

## The admin form had to move with it

`posts.maxMediaAttachments` is the admin-configurable value clients are told about, and its registry schema is already bounded by the constant (`z.number().int().min(1).max(MAX_STORED_MEDIA_ATTACHMENTS)`), so it followed automatically. The **form** did not: `PostsMediaSettingsForm` hardcoded `max={20}` on the "Media per post" field, and `NumberField` settles its buffer against `max` **on blur** — not just as an advisory attribute. So without this half, an admin typing `50` would have watched it snap back to `20` on blur, and the raised ceiling would have been unreachable from the UI. The field and its help line now read `MAX_STORED_MEDIA_ATTACHMENTS` and `MAX_FEDERATION_MEDIA_ATTACHMENTS` instead of spelling both numbers out.

## Blast radius

Because `DEFAULT_SERVER_SETTINGS.posts.maxMediaAttachments` is *defined as* the constant, any instance that never explicitly saved that setting picks up the raise with no admin action: both `/api/v1/instance` and `/api/v2/instance` start advertising `max_media_attachments: 50`, and the create/edit routes start accepting it. An instance whose admin explicitly saved `20` keeps `20` — the stored value wins over the default. No migration is needed: defaults are never auto-persisted, and no stored row can exceed the new max because the old ceiling was lower.

## Not changed

The federation cap is untouched — an outbound ActivityPub note still carries only the first `MAX_FEDERATION_MEDIA_ATTACHMENTS` (4), so remote servers keep receiving a Mastodon-compatible payload however many a status stores locally. `MAX_IMPORTED_ACTIVITY_ATTACHMENTS` (10, the Strava-import bound) is a different constant and is untouched.

Nothing about *where* the ceiling is enforced changes: the resolved setting still is not enforced by any route, which is the standing exception documented in **Instance Limits in Client Components**.

## Tests

Two cases in `PostsMediaSettingsForm.test.tsx`, each pinned to the constants so the form and the ceiling cannot drift apart again:

- `caps media per post at the stored-media ceiling` — the `max` attribute and the blur clamp.
- `states both media ceilings in the help copy` — the derived help string.

Each was proven by partial revert, and each fails for precisely its own half: reverting `max={20}` trips only the first, reverting the `help` literal only the second.

Everything else that asserts on this limit already referenced the constant symbolically (the instance-entity routes, the outbox/statuses ceiling tests) and adapted on its own.

## Verification

Full pre-commit gate, green on every commit: `prettier`, `yarn lint` (exit 0, pre-existing warnings only), `yarn typecheck`, `yarn build`, and `yarn test` — **853 files / 10,599 tests**.

## Notes

- No migration, no config change. Docs updated in the same PR: `docs/environment-variables.md` and the `CLAUDE.md` reminder both quoted the ceiling as a literal `20`.
- No UI screenshot: the only visual change is the admin field's help text and its numeric cap.

## Review

The sub-agent review loop ran over three rounds and is complete — two consecutive clean rounds ended it.

- **Round 1** (correctness, design-system, quality, scope-regression, testing): four lenses clean. Testing raised two findings, both sent to an adversarial verifier. One — "no test pins the literal 50, so reverting the constant leaves the suite green" — was **refuted**: all 15 comparable tuning constants in this repo are tested symbolically and none pins its literal, so a hardcoded `50` would be inconsistent, redundant with the existing symbolic boundary tests, and one more literal to hand-bump next time. The other, the unguarded help copy, was **confirmed** and fixed in 73cbc0cbb.
- **Round 2** (scope-regression, testing) on that fix: clean, with a non-blocking note that the test name no longer described everything it asserted — split in 706b85b4e.
- **Round 3** (testing, quality) on the split: clean. Both proofs reproduced independently.


lib/components/admin/settings/PostsMediaSettingsForm.test.tsx
```diff
@@ -78,6 +79,28 @@ describe('PostsMediaSettingsForm', () => {
     expect(screen.getByLabelText('Upload size limit')).toHaveValue(200)
   })
 
+  // The field's `max` is not decoration: NumberField settles the buffer against
+  // it on blur, so a hardcoded cap here silently clamps away anything the
+  // stored-media ceiling allows. Assert both halves against the constant so the
+  // two cannot drift the next time the ceiling moves.
+  it('caps media per post at the stored-media ceiling', () => {
```
llun, line 89: **testing — the derived help text is unguarded (minor)**

Verified by partial revert: changing only the `help` prop on `PostsMediaSettingsForm.tsx:191` back to the old literal `"Up to 20. The fediverse still only ever sees the first 4."` — leaving `max={MAX_STORED_MEDIA_ATTACHMENTS}` and the constant itself untouched — leaves all 18 tests in this file **green**.

So half of what this PR changed on that field is regression-tested (the `max` attribute and the blur clamp) and half is not. If the copy drifts back, an admin reads "Up to 20" while the field accepts and clamps at 50 — misleading admin-facing copy rather than a functional defect, which is why this is minor rather than blocking.

This file already has precedent for asserting rendered help text derived from dynamic values (the "Post size" help assertion further down), so the fix is one assertion inside the existing test:

```ts
expect(
  screen.getByText(
    `Up to ${MAX_STORED_MEDIA_ATTACHMENTS}. The fediverse still only ever sees the first ${MAX_FEDERATION_MEDIA_ATTACHMENTS}.`
  )
).toBeInTheDocument()
```

_Note on a related claim that did **not** survive verification: a reviewer read the test's comment "Assert both halves against the constant" as meaning "max and help text", and concluded the test does not do what it claims. That is a misreading — "both halves" is the `max` attribute and the blur clamp, both of which the test does assert. That part of the finding is withdrawn; the coverage gap above stands on its own._

llun: Fixed in 73cbc0cbb.

Added the assertion to the existing `caps media per post at the stored-media ceiling` test and imported `MAX_FEDERATION_MEDIA_ATTACHMENTS` alongside the stored ceiling:

```ts
expect(
  screen.getByText(
    `Up to ${MAX_STORED_MEDIA_ATTACHMENTS}. The fediverse still only ever sees the first ${MAX_FEDERATION_MEDIA_ATTACHMENTS}.`
  )
).toBeInTheDocument()
```

Proven the same way the gap was found — reverting only the `help` prop to the old `"Up to 20 … first 4."` literal now fails:

```
× caps media per post at the stored-media ceiling
TestingLibraryElementError: Unable to find an element with the text:
  Up to 50. The fediverse still only ever sees the first 4.
1 failed | 17 passed (18)
```

Also reworded the test's comment, which said "both halves" — it meant the `max` attribute and the blur clamp, but it read ambiguously enough that a reviewer took it as covering the copy too. It now names all three things the test pins.

Full gate re-run green on the fix: prettier, lint, typecheck, build, and 853 files / 10,598 tests.

PR #1509, opened 2026-08-22
fix: store fitness activity types as canonical sport keys

llun:
## Problem

`fitness_files.activityType` stored whatever string the source file carried, and four vocabularies reach it:

| Source | Same ride stored as |
| --- | --- |
| FIT `sport`/`sub_sport` | `cycling`, `gravel_cycling` |
| Garmin TCX `Sport` | `Biking` |
| Strava `sport_type` | `Ride`, `GravelRide` |
| GPX `<trk><type>` | free-form text |

Gear was never affected — it matches through `normalizeActivityTypeToSportKey` — but everything that **groups or filters** on the raw string was:

- the fitness overview breakdown counted "Cycling", "Biking" and "Ride" as three separate activities
- the calendar filter offered three entries for one sport
- the per-type route-heatmap cache keyed three separate rows

## Fix: normalize on the write path

`toActivityData` in `parseFitnessFile.ts` is the single funnel all three parsers return through, so one call there covers uploads **and** the Strava webhook — the webhook writes its `sport_type` into a generated TCX and parses it straight back. The generated file stays a faithful archive; only the column is canonical.

Verified the funnel is complete:

- only `processFitnessFileJob` and `importFitnessFilesJob` write the column, both from `activityData.activityType`
- no API route writes it — it is not owner-editable
- the streamless Strava fallback note creates no `fitness_files` row at all

A sport the keys deliberately do not model (swimming, gym work, Garmin's `Other`) is stored **verbatim** rather than dropped: there is no gear kind to attribute it to, but it is still an activity the breakdown must show.

## Backfill script

`scripts/fitness/normalizeFitnessActivityTypes.ts` — dry run by default, `--apply` to write, idempotent (every sport key normalizes to itself).

**Gear attribution cannot shift as a result.** Auto-assign reads the column through `normalizeActivityTypeToSportKey`, and every value written is a fixed point of that function — asserted in a test rather than assumed.

Per-type heatmap caches are **reported stale, never rewritten**: the unique index is `(actorId, activityTypeKey, periodType, periodKey, region)`, so renaming a `cycling` row to `ride` collides with the row already built from `Ride` instead of merging into it. `recreateFitnessRouteHeatmaps.ts` is what actually reconciles them, and the script prints that command.

## Caption vocabulary — required, not incidental

`getActivityPresentation` moves out of `processFitnessFileJob` into `lib/services/fitness-files/activityPresentation.ts` and keys on `SportKey`. This is forced by the change above: the old table matched raw strings and had `cycling` but no `ride`, so it would have captioned every new import `Gravel_ride 🏋️`.

It keeps its own **gerund** labels rather than reusing `SPORT_LABELS`. Post text is federated and stored forever, so the wording stays exactly what the old table produced for the vocabularies it covered — five existing job tests caught the reword when the nouns were tried first. Only the gaps the old table left are new.

That also fixes a live bug: **no Strava spelling was in the old table**, so Strava rides were already being captioned `Ride 🏋️` while the same ride from a FIT file got `Cycling 🚴`.

## Deliberately left alone

- `getActivityLabel` (`FitnessStatusDetail`) and `formatActivityType` (`ActorFitnessDashboard`) — pure display, already render sport keys correctly, and changing them risks a chip-layout regression for no gain here.
- `getActivityLabelAndEmoji` (`lib/services/strava/activity.ts`) — a third vocabulary, but it labels a Strava **API** activity for the streamless fallback note (preferring Strava's own `rawType`) and never writes the column.

## Trade-off

Normalizing loses the raw string, so an ambiguous GPX type such as `trail` reads as `trail_run`. The dry run lists every transition before anything is written.

## Testing

Full gate green: prettier, lint, typecheck, build, **855 test files / 10,670 tests**.

Because `scripts/` is lint-ignored and untested, the script was **run for real** against a local SQLite database seeded with all four vocabularies:

- 11 of 16 rows normalized; three spellings of one ride collapsed onto `ride`
- swims and `Other` left untouched and reported
- a second run reported nothing to do (idempotency)

New unit tests cover `normalizeStoredActivityType`, `getActivityPresentation` (including that pre- and post-normalization spellings caption identically), and the script's planner across all four vocabularies.

## Review

The sub-agent code-review loop ran to completion: **11 rounds, 10 commits, two consecutive clean rounds**. Eleven findings confirmed and fixed, one refuted by an adversarial verifier, one declined with reasoning. Per-round detail is in the comments below.

Five findings were defects in a previous round's *own* fix — most notably one that was a no-op at one of its two callers, and a commit that contradicted a line of `AGENTS.md` added in this same PR. Two of the confirmed findings overturned judgement calls stated in an earlier version of this description:

- `getActivityLabelAndEmoji` was described as serving only the streamless fallback note. It is in fact the caption writer for the ordinary successful Strava import, so every mountain-bike, gravel and e-bike ride from the webhook carried the wrong glyph. Now delegated to the shared presenter.
- The backfill script's untested `--apply` write loop was first judged acceptable by convention, then re-argued and covered — along with a pagination path where two realistic typos passed the whole suite.

Gate at every round: prettier, lint, typecheck, build, full suite. Final: **855 files / 10,695 tests** green.


llun:
## Sub-agent review — round 1

Four lenses in parallel (correctness, quality, testing, scope/regression), then an adversarial verifier on each finding whose validity depended on reachability. `review-design-system` was skipped: this diff touches no `app/` or `lib/components/` code.

| # | Lens | Finding | Verdict | Action |
|---|---|---|---|---|
| 1 | quality | `getActivityLabelAndEmoji` is a third sport vocabulary; `MountainBikeRide` → 🚴 not 🚵 | **CONFIRMED** | fixed in 352207859 |
| 2 | quality | `parseArgs` duplicated byte-for-byte across three `scripts/fitness/` scripts | real, **declined** | see below |
| 3 | testing | gear-invariance test is vacuous | **CONFIRMED** (by mutation) | fixed in 352207859 |
| 4 | scope | `AGENTS.md` still describes the column as raw/free-form | **CONFIRMED** | fixed in 352207859 |
| 5 | correctness | `getActivityLabel` folds `hike` → "Walk" | pre-existing, untouched | out of scope |
| 6 | scope | email `getActivityTitle` fallback would render "Gravel_ride" | **REFUTED** | none |

### #1 — the PR's own rationale was wrong

The original description claimed `getActivityLabelAndEmoji` only served the streamless fallback note. It does not. `importStravaActivityJob` creates a file-backed import's status with `text: ''` and fills it via `buildStravaActivitySummary` before the deferred process job runs, so it is the caption writer for the **ordinary successful Strava import**. The emoji prefixes the caption unconditionally, so every mountain-bike / gravel / e-bike / virtual ride through the live webhook carried the wrong glyph. Now delegates to the shared `getActivityPresentation`.

### #3 — the test proved nothing

It asserted `normalize(to) === normalize(from)`, but `to` is *defined* as `normalize(from)`, so it reduced to `f(f(x)) === f(x)` — true for any deterministic `f`. A reviewer forced the classifier to return `walk` for every input (which would misattribute every activity to shoes) and the test still passed. Rewritten against a hardcoded bike/shoes oracle via `getGearKindForActivityType`; it now fails under that same mutation.

### #6 — refuted empirically

The fallback is unreachable once a fitness file exists: `processFitnessFileJob` backfills the status text through the shared presenter before `notifyActivityImported` re-reads it. Reproduced: the email renders `🚴 Gravel cycling`.

### #2 — declined, with reasoning

The duplication is real (`parseArgs` is byte-identical to `backfillFitnessDevices.ts`, and `backfillFitnessMovingTime.ts` carries a third near-copy). Extracting it means editing two scripts this PR otherwise doesn't touch, in a data-fix PR — which the scope lens would rightly flag next round. Worth a follow-up that does only that, across all three call sites at once.

Gate after fixes: prettier, lint, typecheck, build, **855 files / 10,678 tests** green.

llun:
## Sub-agent review — rounds 2 and 3

Each round reviews the **previous round's fixes**, not the original diff.

### Round 2 — on `c9896d253..352207859`

Correctness: clean. Independently reproduced both mutation tests, and confirmed the dropped `.trim()` at the new call site is harmless (`toComparableToken` strips non-alphanumerics; the fallback trims internally).

Scope/regression: **one real regression from round 1's own fix.** Brute-forced all 49 Strava `sport_type` values through the old and new implementations side by side. No emoji got worse anywhere, and `MountainBikeRide` got the correct 🚵 — but three labels lost specificity, because `normalizeActivityTypeToSportKey` folds them into a coarser key for *gear* purposes:

| | before round 1 | after round 1 | now |
|---|---|---|---|
| `Handcycle` | Handcycle 🚴 | Cycling 🚴 | **Handcycling 🚴** |
| `Velomobile` | Velomobile 🏋️ | Cycling 🚴 | **Velomobile 🚴** |
| `VirtualRun` | VirtualRun 🏃 | Running 🏃 | **Indoor running 🏃** |

A sport key answers "which bike or which shoes", so folding a handcycle into `ride` is right for attribution. A caption is not answering that question, and "Cycling" erases a distinction the athlete made. Fixed in 9b32c7149 with a `SPECIFIC_ACTIVITY_LABELS` table consulted before the sport-key lookup; the key itself is untouched, so gear attribution does not move, and a test asserts both halves.

### Round 3 — on `352207859..9b32c7149`

**Round 2's fix was a no-op at one of its two callers.** `getActivityPresentation` is called with a raw string by the Strava summary builder, but with the *already-normalized* key by `processFitnessFileJob`. So an uploaded handcycle GPX arrived as `ride`, never matched the new table, and was still captioned "Cycling" — while the same ride from Strava said "Handcycling". Round 2's tests called the presenter directly and structurally could not catch this.

Fixed in a447ffd48: the source file's own spelling is carried through parsing as a transient `FitnessActivityData.rawActivityType` (never persisted — the column's job is the canonical key), and `buildActivitySummary` prefers it.

Guarded **end to end** in `processFitnessFileJob.test.ts` this time, not on the presenter: a unit test on a shared function cannot catch a caller handing it the wrong value, which is the entire shape of this bug. Reverting the caller fails it with `expected '🚴 Cycling — 5.20 km in 28:15 min' to contain 'Handcycling'`.

Round 3 also caught that round 2 **contradicted AGENTS.md** — the same PR wrote "do not add a fourth table" and then added one. The block now documents the table, its ordering, and the raw-versus-key rule.

Gate after each round: prettier, lint, typecheck, build, full suite. Currently **855 files / 10,685 tests** green.

llun:
## Sub-agent review — rounds 4, 5 and 6

### Round 4 — on `9b32c7149..a447ffd48`

Scope/regression: clean. Confirmed `rawActivityType` cannot reach the database — the DB write type (`UpdateFitnessFileActivityData`) does not contain the field, and there is no `...activityData` spread anywhere in `lib/` or `app/`. Also confirmed the Strava and upload paths now produce byte-identical captions for `Handcycle` and `MountainBikeRide`.

Quality (fresh-eyes read of the whole change): one finding. `activityPresentation.ts` carried its own copy of `toLowerCase().replace(/[^a-z0-9]/g, '')` — byte-for-byte the private `toComparableToken` in `sportTypes.ts`, the module that owns that vocabulary and which `activityPresentation` already imports from. Two copies key the caption tables and `EXACT_SPORT_KEYS` off the same rule today, and would diverge the moment a fifth vocabulary needed the separator set widened. Fixed in c7be86114 by exporting and importing it.

The same review explicitly considered and **rejected** two larger changes: merging `SPECIFIC_ACTIVITY_LABELS` into `UNMODELLED_ACTIVITY_LABELS` (they encode a real conceptual distinction — has-a-key-but-needs-specificity versus has-no-key), and moving normalization to read time instead of write time (the heatmap cache's unique index is keyed on the type string, and the breakdown and calendar filter `GROUP BY` it, so every consumer would have to re-derive the canonical value independently — exactly the drift this PR exists to remove).

### Round 5 — on `a447ffd48..c7be86114` plus a full audit

Correctness: clean. Verified the exported collapser is character-for-character equivalent, including for unicode, emoji and accented input.

Testing: a full **revert-and-prove** pass over all eight production behaviours in the cumulative diff. Every one is pinned by a test that fails meaningfully when reverted — including each of the three `SPECIFIC_ACTIVITY_LABELS` entries individually, and the *ordering* of that table before the sport-key lookup (swapping the two blocks breaks three tests). One finding: the new `describe('activity caption')` had been spliced inside `describe('gear auto-assignment')`. Fixed in e3b4c3c52.

### Round 6 — clean

Scope: verified the reparenting was structurally pure and — the real risk — that the caption tests' loss of the inherited `beforeEach(clearGear)` changes nothing, because captioning never reads gear. Also confirmed all six commits trace back to the original two asks: no migration, no schema-dump change, no version bump, no unrelated files.

Quality: no material findings. Division of responsibility between `sportTypes.ts` and `activityPresentation.ts` is clean, no rule expressed twice, no stale identifiers from earlier rounds, and every factual claim in all three touched docs re-verified against the final code.

---

Rounds 1–5 each found something; round 6 was the first clean one. A confirming round 7 is running on correctness and testing, whose last look predated the final commit.

llun:
## Sub-agent review — rounds 7 to 11, and outcome

**The loop is complete: 11 rounds, two consecutive clean (10 and 11), cap 20 not reached.**

### Round 7 — one finding

Correctness clean (112 assertions over 36 adversarial inputs — prototype keys, unicode, a 300-char string — nothing throws). Testing re-opened a judgement round 5 had made: the `--apply` write loop was untested. Round 5 called that acceptable because `backfillFitnessDevices.ts` skips it too; round 7 argued `repairFailedFitnessImports.test.ts` is the closer precedent, being the one sibling that mocks the DB and asserts the failure branch of this exact pattern — and this script mutates `fitness_files` in bulk against production with nothing but stdout in between. Agreed and fixed in e8921cb2d.

### Round 8 — one finding, proven

The new script-level tests all supplied a handful of rows, so the scan always ended on its first iteration and **the pagination boundary was never exercised**. Proven: `offset += 0` and an off-by-one in the break condition *both* passed the entire file. Either silently truncates the scan at 200 rows — and an actor with more history than that is the normal case for what this script exists to fix.

Fixed in aedf8c62c with a full first page plus a short second one. The stub caps its own call count deliberately: the scan only ends on a short page, so a non-advancing offset would be handed the same full page forever and *hang* rather than fail. With the cap both typos fail as plain assertions.

### Round 9 — one finding, present since the first commit

The heatmap-rebuild advice was derived from the **plan**, not from what was written, and printed unconditionally after `--apply`. A run where every write threw still produced:

```
Done: 0 updated, 512 error(s).

Per-activity-type route heatmaps cached under "cycling", "GravelRide" are now stale. Rebuild them with:
  NODE_ENV=production scripts/fitness/recreateFitnessRouteHeatmaps.ts --actor-id …
```

Nothing had moved, so nothing was invalidated — but the guidance was paste-ready and worded identically whether one write failed or all of them did. Eight rounds missed it. Fixed in 862cb7f38 by naming only the values a row actually moved away from.

### Rounds 10 and 11 — clean

Round 10 confirmed the fix across every failure combination (including a `from` value spanning several rows where only some succeed — still correctly named, since `transitions` cannot contain duplicate `from` values), and flagged one advisory: nothing pinned the *success* path of the block just edited. Closed in 364d8fd21.

Round 11 confirmed that test is non-vacuous against three mutations, judged the literal key-order assertion the right trade-off (all three assertions are substring matches, so rewording the sentence would not false-fail), and returned a merge-readiness verdict on the whole change.

---

### Outcome

| | |
|---|---|
| Rounds | 11 (cap 20) |
| Commits | 10 |
| Findings confirmed and fixed | 11 |
| Findings refuted by the verifier | 1 |
| Findings declined with reasoning | 1 |
| Consecutive clean rounds | 2 |

Five of the eleven findings were defects in a **previous round's own fix** — including a fix that was a no-op at one of its two callers, and a commit that contradicted a line of `AGENTS.md` written in the same PR. That is the loop doing the job it exists for.

Gate at every round: prettier, lint, typecheck, build, full suite. Final: **855 files / 10,695 tests** green.

PR #1512, opened 2026-08-23
refactor: resolve the viewer's follow row once per profile render

llun:
## Summary

Follow-up to #1510, which deliberately left this alone to keep a security fix narrow. (#1510 merged mid-session; this is rebased onto `main` and targets it directly.)

An authenticated non-owner viewing `/@user@domain` issued the identical
`getAcceptedOrRequestedFollow({ actorId: viewer, targetActorId: profileActor })`
query **twice** in one request, from two call sites that cannot see each other:

1. `getProfileData` → `resolveActorStatusesAudience`, which needs it to decide
   whether followers-only posts are in scope (`includeFollowersOnly`).
2. `page.tsx` → `getRelationship`, which needs the same row for the follow
   button's state plus its `showing_reblogs` / `notifying` preferences and its
   language filter.

Neither was memoized, so it was one genuinely redundant database round trip per
signed-in non-owner profile view.

## What changed

- **`lib/services/getViewerFollow.ts`** (new) wraps the lookup in React `cache()`,
  the same way `getActorFromSession`, `getServerAuthSession` and `loadSharedHeatmap`
  already deduplicate their per-request reads. Its parameters are **positional** on
  purpose: React keys primitive arguments through a `Map` and object arguments through
  a `WeakMap`, so the options-object spelling the database method itself uses would
  allocate a fresh, never-equal key on every call and memoize nothing.
- The two read call sites above now go through it. `person.id === persistedActor.id`
  on the local branch, and on the remote branch `person.id` is the same object the page
  later hands `getRelationship` — the tests take the relationship's target from the
  returned profile's own `person.id` on both branches so that identity is asserted
  rather than assumed.
- `statusAccess`'s `isAcceptedFollowerOf` becomes `isAcceptedFollow(follow)`. The
  *rule* #1510 unified — a Requested follow is not a follower — stays shared; only the
  lookup feeding it differs between the two halves.

## What deliberately did not change

- **`database.getAcceptedOrRequestedFollow` is untouched.** It stays the entry point
  for every write path. Follow, unfollow, block and the follow-request authorize/reject
  routes all read the row, mutate it, and then report the result through
  `getRelationship` — they reach the cached helper only *after* their own write, which
  is a cold read, so their responses still describe the state they just produced. That
  ordering is now the helper's documented contract; routing one of those pre-mutation
  reads through it would silently break those five routes.
- **`canActorReadSingleStatus` stays on the direct call** — a scope boundary, not a
  claim that it never repeats. It *does*: it runs once per status in `getProfileData`'s
  own per-status pass with no `followerStateByActorId` prefetch, so a profile carrying
  two boosts of the same followers-only author issues that query twice. Routing it
  through the helper would collapse that too and is safe, but it reaches a dozen further
  call sites (inbox handling, status create/edit, search, polls), most on paths with no
  request scope at all. **Worth a follow-up.**
- `getRelationship`'s **incoming** follow lookup (target → viewer) is a different row
  that nothing else on a request asks for, so it stays a plain read.

## Testing

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` →
`yarn test` all green (860 files / 10,736 tests).

The request-level assertion lives in `app/(timeline)/[actor]/profileFollowLookup.test.ts`:
it drives `getProfileData` and then `getRelationship` exactly as the page does, inside one
request scope, on the local **and** remote branches, and asserts the viewer→profile follow
row is queried **once**. Each half of the fix was proved load-bearing by reverting it
alone — either revert turns that 1 into a 2.

A call count alone cannot tell "the two lookups collapsed" from "the audience half never
ran", so each branch also pins what the audience *produced* — `includeFollowersOnly: true`
reaching `getActorStatuses` locally, and the full resolved scope reaching
`getAttachmentsForActor` remotely. Without the latter, removing the remote branch's audience
resolution outright — an unscoped remote profile, the followers-only leak #1510 exists to
prevent — left every test in the file passing.

One thing worth flagging, because it is easy to get wrong and easy to "fix" into
vacuousness: **React `cache` does not memoize under Vitest at all.** React ships two builds,
and the client build Vitest resolves is a hard passthrough; only the `react-server` build —
what Next.js loads for Server Components and route handlers — keys on arguments, and only
while an async cache dispatcher is installed. That dispatcher *is* the request scope.
`lib/testing/reactCacheScope.ts` loads the server build and stands a real scope up around
the code under test. Scopes are sequential and a concurrent or nested one throws: the
dispatcher is a single mutable global, so two live scopes cannot be isolated from each other
at all, and saving/restoring it on a stack would fix only the value left behind while
leaving the mid-flight bleed open.

## Docs

`AGENTS.md` (Who May See an Actor's Statuses + Testing Guidelines), `REVIEW.md` and
`CLAUDE.md` record the read-path-only contract, the positional-argument requirement, and
the Vitest `cache` trap.

## Review

Three sub-agent review rounds (correctness, testing, quality, scope/regression, security),
with every finding adversarially verified before posting and each round pointed at the
previous round's fixes. Eight findings across rounds 1–2, all fixed, replied to and
resolved; round 3 clean. The two that mattered:

- `runInReactCacheScope` leaked a live React dispatcher when two scopes were started
  concurrently, so later out-of-scope code got memoized where production hands it a
  passthrough. Fixed with a reentrancy guard.
- A justification comment of mine was simply false — it claimed the boosted-original
  lookup never repeats within a request. Corrected in the code and all three doc files.


lib/testing/reactCacheScope.ts
```diff
@@ -0,0 +1,78 @@
...
+export const serverCache = loadServerReact().cache
+
+/** Runs `fn` as if it were one request, with an empty `cache()` store. */
+export const runInReactCacheScope = async <T>(
+  fn: () => Promise<T>
+): Promise<T> => {
+  const internals =
+    loadServerReact()
+      .__SERVER_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE
+  const previousDispatcher = internals.A
```
llun, line 104: **[CONFIRMED] `runInReactCacheScope` leaks a stale dispatcher when two scopes are started concurrently.**

The save/restore of `internals.A` is LIFO-only. `Promise.all([runInReactCacheScope(a), runInReactCacheScope(b)])` runs A's whole synchronous prefix (installing dispatcher SA) before B is invoked, so B captures `previousDispatcher = SA` rather than the true original. Whichever settles first restores a value captured before the other scope existed; after both settle, `internals.A` holds a leftover non-null dispatcher. Reproduced against the real `react-server` build: out-of-scope calls afterwards get memoized (both returned 3 where 3 and 4 were expected), because `cache()` only takes its passthrough branch when `internals.A` is falsy.

A stack-based push/pop is NOT the right fix: `internals.A` is a bare mutable global, not `AsyncLocalStorage`-scoped, so two genuinely concurrent scopes cannot be isolated at all — A's continuation after an `await` would read whichever dispatcher is currently installed, possibly B's. A stack would make that look fixed while leaving the mid-flight bleed open. Fix with a reentrancy guard that throws on concurrent/nested entry.

The two test files shipped in this PR are unaffected — every scope in them is awaited before the next begins, and the one `Promise.all` is two calls to the cached function inside a single scope.

lib/testing/reactCacheScope.ts
```diff
@@ -0,0 +1,78 @@
...
+ * So a test that wants to observe request-level deduplication has to do both
+ * halves itself: load the server build (via `serverCache`, which the test hands
+ * to `vi.mock('react', …)` so the module under test gets it) and run the code
+ * inside a scope (via `runInReactCacheScope`). Without the first, nothing
+ * memoizes; without the second, the server build falls back to the passthrough
+ * and nothing memoizes either — the same result a caller outside a request gets
+ * in production.
+ *
+ * Each `runInReactCacheScope` call is one request: a fresh store, restored on
+ * the way out, so scopes neither leak into each other nor into unrelated tests.
```
llun: **[CONFIRMED] JSDoc overclaims isolation.**

"a fresh store, restored on the way out, so scopes neither leak into each other nor into unrelated tests" is stated unconditionally, but holds only for sequential/nested use. A future author testing "do two concurrent requests stay isolated" would reasonably reach for `Promise.all([runInReactCacheScope(a), runInReactCacheScope(b)])` — exactly the case that silently breaks. State the sequential-only constraint explicitly.

lib/services/statusAccess.ts
```diff
@@ -33,21 +34,20 @@ export const isStatusPubliclyReadable = (status: Status): boolean => {
...
+// of this module need, and which each had its own copy of. A Requested follow is
+// not a follower: it is the pending state, and treating it as one would hand a
+// stranger the followers-only audience by asking.
+//
+// The rule is shared; the lookup that feeds it deliberately is not.
+// `resolveActorStatusesAudience` asks about the profile's own actor, which
+// `getRelationship` asks about again on the same render, so it reads through the
+// request-memoized `getViewerFollow`. `canActorReadSingleStatus` asks about
+// whoever wrote a boosted original — a question nothing else on the request
+// repeats, and one that also runs from inbox and status-create paths where there
```
llun: **[CONFIRMED] The comment's justification is factually wrong: the boosted-original lookup CAN repeat within one request.**

The comment says `canActorReadSingleStatus` asks "a question nothing else on the request repeats". But that call runs once per status inside `getProfileData`'s own `Promise.all(scopedStatuses.map(...))`, with no `followerStateByActorId` prefetch on that path. A profile containing two or more boosts whose ORIGINALS were written by the same followers-only author issues the identical query twice — the very duplication this PR exists to remove, inside this PR's own code path.

Note the scope constraint: routing this third call site through `getViewerFollow` was explicitly excluded from this change's brief ("only the two read call sites"). So the code stays as it is, but the comment must stop asserting something untrue — it should state the real reason (deliberately out of scope) and record the known repeat case.

lib/services/statusAccess.ts
```diff
@@ -1,6 +1,7 @@
 import { Database } from '@/lib/database/types'
+import { getViewerFollow } from '@/lib/services/accounts/getViewerFollow'
```
llun: **[CONFIRMED] Placement inverts the repo's only precedent for this module boundary.**

This is the sole import anywhere from a top-level `lib/services/*.ts` module into a `lib/services/<subdir>/` module. The established direction runs the other way — `lib/services/accounts/*` imports `@/lib/services/serverSettings`, `@/lib/services/medias`, `@/lib/services/email` and so on, never the reverse. Verified by grepping every top-level `lib/services/*.ts`: `statusAccess.ts` is the only one.

Since `getViewerFollow` is consumed by both a top-level module and an `accounts/` module, it should live at `lib/services/getViewerFollow.ts`, a sibling of `statusAccess.ts`. `accounts/relationship.ts` can then import it in the direction every other `accounts/*` file already does.

app/(timeline)/[actor]/profileFollowLookup.test.ts
```diff
@@ -0,0 +1,159 @@
+import { Database } from '@/lib/database/types'
```
Comment (llun) [inline app/(timeline)/[actor]/profileFollowLookup.test.ts:5]: Coverage gap: the remote-actor branch of `getProfileData` is not covered by the dedup assertion.

The test only exercises the local-account branch. The remote branch also calls `resolveActorStatusesAudience` (now via `getViewerFollow`) with `targetActor.id = person.id`, where `person` comes from a FETCHED ActivityPub document rather than a database row — so the id match that makes the two cache keys collapse is less self-evidently true there than on the local path. No bug today (verified), but nothing would catch a future edit that threaded a different id through the remote branch.

lib/services/accounts/getViewerFollow.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      viewerId: OTHER_VIEWER,
+      targetActorId: TARGET
+    },
+    {
+      description: 'a different target',
+      viewerId: VIEWER,
+      targetActorId: OTHER_TARGET
+    }
+  ])(
+    'reads its own row for $description',
```
llun: `it.each` title should be the bare `$description` form.

AGENTS.md's Testing Guidelines specify a table-driven `it.each([...])('$description', …)` with a `description` column. This uses the wrapper form `'reads its own row for $description'`. That is also the riskier pattern given Vitest truncates the interpolated value at 40 characters while a literal prefix does not — harmless for these two short descriptions, but a trap for anyone extending the table.

llun (commented):
Round 1 automated review findings.

lib/testing/reactCacheScope.ts
llun: Fixed in 73e455167. Confirmed and reproduced before fixing, including against the real `react-server` build.

Took the reentrancy guard rather than a stack, for the reason given: `internals.A` is a bare mutable global, so a stack would correct only the value left behind and leave the mid-flight bleed open while looking handled. `runInReactCacheScope` now throws on concurrent or nested entry, and the guard runs before it touches the dispatcher, so a refused scope cannot disturb the live one.

`lib/testing/reactCacheScope.test.ts` covers it, including the exact reported shape — `Promise.all` of two scopes now rejects, and the surviving scope still cleans up so a later out-of-scope call is not memoized.

lib/testing/reactCacheScope.ts
llun: Fixed in 73e455167. The unconditional isolation claim is gone; the doc now states the sequential-only constraint and points at the guard for why concurrent scopes cannot work rather than merely being unsupported. AGENTS.md and REVIEW.md carry the same constraint.

lib/services/statusAccess.ts
llun: Fixed in 73e455167 — the comment was wrong and is rewritten.

You are right that it repeats: once per status in `getProfileData`'s own per-status pass, with no `followerStateByActorId` prefetch on that path, so two boosts of the same followers-only author issue the identical query twice. The comment now says that explicitly and frames leaving the call site direct as a scope boundary, not an absence of duplication. AGENTS.md and CLAUDE.md are corrected the same way.

The code stays as it is: routing that third call site through `getViewerFollow` is safe and would collapse the repeat, but it reaches a dozen further call sites (inbox handling, status create/edit, search, polls), so it is its own change. Flagged in the PR body as follow-up work.

lib/services/statusAccess.ts
llun: Fixed in 73e455167. Verified the claim independently before acting — grepping every top-level `lib/services/*.ts`, this import was indeed the only one reaching into a `lib/services/<subdir>/` module, while `accounts/*` already imports upward (`@/lib/services/serverSettings`, `@/lib/services/medias`, `@/lib/services/email`).

The helper now lives at `lib/services/getViewerFollow.ts`, so `statusAccess.ts` imports a sibling with `./getViewerFollow` and `accounts/relationship.ts` imports upward like the rest of that directory. Its test moved with it, and the three doc files were updated to the new path.

app/(timeline)/[actor]/profileFollowLookup.test.ts
Comment (llun) [reply app/(timeline)/[actor]/profileFollowLookup.test.ts:5]: Fixed in 73e455167. The request-level test now has a `for a remote actor` case alongside the local one, driving the branch through mocked WebFinger, `getActorPerson`, `getActorPosts` and `getActorCollectionCounts`.

It also takes the relationship's target from the returned profile's own `person.id` rather than from a constant — on both branches — so the id identity the memoization depends on is asserted rather than assumed. Proved it is not vacuous by reverting the `statusAccess.ts` half of the fix: the remote case fails alongside the local ones.

lib/services/accounts/getViewerFollow.test.ts
llun: Fixed in 73e455167. Now the bare `'$description'` form, with each description rewritten as a full behavior statement ("gives a different viewer its own row" / "gives a different target its own row") — both well under the 40-character truncation point and distinct well before it.

Note the file moved to `lib/services/getViewerFollow.test.ts` in the same commit (thread 4).

app/(timeline)/[actor]/profileFollowLookup.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      })
+    })
+
+    it('reads the viewer follow once for a signed-in non-owner profile view', async () => {
+      const profile = await runInReactCacheScope(() =>
+        renderProfileForViewer(REMOTE_HANDLE)
+      )
+
+      expect(profile.person.id).toBe(REMOTE_ACTOR_ID)
+      expect(viewerFollowCalls(REMOTE_ACTOR_ID)).toHaveLength(1)
```
Comment (llun) [inline app/(timeline)/[actor]/profileFollowLookup.test.ts:248]: **[CONFIRMED] The remote-actor assertion cannot tell "memoized" from "never ran".**

`expect(viewerFollowCalls(REMOTE_ACTOR_ID)).toHaveLength(1)` counts calls, so it is satisfied both when the two lookups collapse into one AND when the audience half never happens at all. Proved by revert-proof: stubbing out the `resolveActorStatusesAudience` call in `getProfileData`'s remote branch entirely — simulating "the remote branch stopped scoping the audience" — leaves all four tests in this file passing. The single remaining call comes from `getRelationship` alone.

The local branch does not have this hole: `still reports the viewer as a follower to both call sites` asserts `getActorStatuses` was called with `includeFollowersOnly: true`, and that test does fail under the same mutation. There is no remote analogue.

To be clear about what the test DOES still guard — reverting either memoization call site (`resolveActorStatusesAudience` in `lib/services/statusAccess.ts`, or `getViewerFollow` in `lib/services/accounts/relationship.ts`) fails it 1 vs 2, on the local and remote cases alike. So it is not vacuous about memoization. It is blind to the more severe regression: an unscoped remote profile, which is the followers-only leak this PR's own comments warn about.

The remote branch passes `remoteVisibilityScope` to `database.getAttachmentsForActor`, so asserting that call carries `includeFollowersOnly: true` would close it and make the two branches symmetric.

app/(timeline)/[actor]/profileFollowLookup.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+      expect(viewerFollowCalls(PROFILE_ACTOR_ID)).toHaveLength(2)
+    })
+  })
+
+  // The remote branch resolves the actor from a FETCHED ActivityPub document
+  // rather than a database row, so the id the audience lookup is keyed on and
+  // the id the page hands `getRelationship` arrive by different routes. They
+  // still have to be the same string for the two reads to collapse.
+  describe('for a remote actor', () => {
```
Comment (llun) [inline app/(timeline)/[actor]/profileFollowLookup.test.ts:217]: Adjacent pre-existing gap: no test anywhere covers the remote branch admitting followers-only content to an actual follower.

Not introduced by this PR. `app/(timeline)/[actor]/getProfileData.test.ts`'s remote `describe` only ever mocks `getAcceptedOrRequestedFollow` returning falsy — its two audience-propagation tests never mock an `Accepted` follow, so `includeFollowersOnly: true` is never exercised on the remote path. The local branch has that case (`admits followers-only statuses once the viewer follows the actor`); the remote branch does not.

Noting it here because this PR's new remote `describe` already mocks an Accepted follow, so closing the sibling finding on this file lands this coverage as a side effect rather than needing its own change.

app/(timeline)/[actor]/profileFollowLookup.test.ts
Comment (llun) [reply app/(timeline)/[actor]/profileFollowLookup.test.ts:248]: Fixed in ffea76cf4. You were right, and I reproduced your mutation before and after.

Added `scopes the remote gallery with the memoized follow row`, which asserts the remote branch's `getAttachmentsForActor` call actually carries the resolved audience — `publicOnly: false`, `visibleToActorId`, `includeFollowersOnly: true`, and the followers collection. That is the remote analogue of what the local case pins through `getActorStatuses`; on this branch it has to be the attachment query, since `getActorPosts` reads the remote outbox and that is public by construction.

Verified against exactly the mutation you used — replacing `remoteVisibilityScope` with `{}` so the remote branch stops scoping. Before: all four tests passed. After: the new test fails and the other four still pass, so it isolates the regression rather than just adding noise.

The count assertion stays as well; the two now pin different properties.

app/(timeline)/[actor]/profileFollowLookup.test.ts
Comment (llun) [reply app/(timeline)/[actor]/profileFollowLookup.test.ts:217]: Closed as a side effect of thread 1, as you predicted. The new `scopes the remote gallery with the memoized follow row` runs with an `Accepted` follow mocked and asserts `includeFollowersOnly: true` reaches the query, so the remote branch admitting followers-only content to an actual follower is now covered.

Left `app/(timeline)/[actor]/getProfileData.test.ts` alone — the coverage now exists, and widening that file's remote block was not part of this change.

app/(timeline)/[actor]/profileFollowLookup.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    mockDatabase.isDomainBlockedByActor.mockResolvedValue(false)
+    mockDatabase.getMute.mockResolvedValue(null)
+    mockDatabase.getAccountNote.mockResolvedValue('')
+    mockDatabase.getEndorsement.mockResolvedValue(null)
+    mockDatabase.getActorPublicIds.mockResolvedValue(new Map())
+    mockDatabase.getAcceptedOrRequestedFollow.mockResolvedValue({
+      id: 'follow-id',
+      actorId: VIEWER_ACTOR_ID,
+      targetActorId: PROFILE_ACTOR_ID,
+      status: FollowStatus.enum.Accepted
```
Comment (llun) [inline app/(timeline)/[actor]/profileFollowLookup.test.ts:147]: **Every case in this file seeds an `Accepted` follow, so together they only ever prove the permissive branch.**

An audience that returned `includeFollowersOnly: true` unconditionally for any non-owner viewer would satisfy every assertion here. The deny outcome is covered in `getProfileData.test.ts`, but that runs through React's client-build `cache`, which is a passthrough — never through the real memoized path this PR adds. The two halves ("the predicate denies correctly" and "the cache does not leak across keys") are each covered; they are never combined into one assertion that a real non-follower is denied through the production request-cache path.

No live vulnerability — the backstops hold today. But this is the surface that served followers-only posts and direct messages to anyone before #1510, so it is worth closing rather than leaving to the backstops.

lib/services/getViewerFollow.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Rendering `/@user@domain` asks this exact question twice from two places that
+ * cannot see each other: `resolveActorStatusesAudience` needs it to decide
+ * whether the viewer may be shown followers-only posts, and `getRelationship`
+ * needs the same row for the follow button's state, its reblog/notify
+ * preferences and its language filter. Both are reads on one render, so the
+ * second was a wasted round trip for every signed-in visitor to someone else's
+ * profile.
+ *
+ * **Read paths only.** `database.getAcceptedOrRequestedFollow` stays the entry
+ * point everywhere else. The follow, unfollow and block routes read *this* row,
```
llun, line 19: **The read-path-only warning overstates the hazard for two of the routes it names.**

`follow`, `unfollow` and `block` do read this row — the outgoing direction, `viewer → target` — mutate it, then report through `getRelationship`, so the warning is exactly right for them. But the follow-request `authorize`/`reject` routes read and mutate the *incoming* direction (`accountId → currentActor`), which is a different cache key and cannot collide with this helper either way.

Not a functional defect — over-inclusive prose. Worth tightening, since the whole value of this comment is that a future contributor trusts it precisely.

app/(timeline)/[actor]/profileFollowLookup.test.ts
Comment (llun) [reply app/(timeline)/[actor]/profileFollowLookup.test.ts:147]: Fixed in 575ad0fb2. Added a Requested-follow case to each branch, asserting the audience reaching `getActorStatuses` (local) and `getAttachmentsForActor` (remote) withholds the followers-only scope — and that the lookup is still made exactly once, so the memoization is unaffected by which way the answer goes.

Verified against exactly the mutation described: patching `resolveActorStatusesAudience`'s `isFollower` to be unconditionally true for a non-owner leaves the other five tests green and fails both new ones, on the local and remote branches alike.

lib/services/getViewerFollow.ts
llun: Fixed in 575ad0fb2. The paragraph now states the hazard for follow/unfollow/block, which genuinely mutate this row, and notes parenthetically that authorize/reject mutate the opposite direction — a different key that cannot collide — while remaining the same shape of hazard if the direction ever changes.

PR #1530, opened 2026-08-24
refactor: retire gear components without creating blank replacement

llun:
## Summary
- Replaces the gear component **Replace** action with a **Retire** action that closes the fitted component's installation window at `now` without creating a blank duplicate row.
- Replaces `POST /api/v1/fitness/gear/:id/components/:componentId/replace` with `POST .../retire`, returning the updated `{ component }` entity.
- Renames `replaceFitnessGearComponent` to `retireFitnessGearComponent` across database layer, client SDK, and callers.
- Removes unused `ReplaceGearComponentRequest` schema.
- Aligns UI terminology and state in `GearComponentsCard` and `GearDetailView` from "Replace / Replaced" to "Retire / Retired".
- Updates documentation in `docs/fitness-file-storage.md` and `AGENTS.md`.
- Removes deleted route test from `tsconfig.typecheck.json` ratchet list and verifies new test is fully type-clean.

## Verification
- `yarn run prettier --write .` passed
- `yarn lint` passed
- `yarn typecheck` passed
- `yarn build` passed
- `yarn test` passed (867 test files, 10,915 tests passing)


lib/database/sql/fitnessGear.ts
```diff
@@ -880,49 +875,27 @@ export const FitnessGearSQLDatabaseMixin = (
...
-      await trx('fitness_gear_components').insert(replacement)
+  async retireFitnessGearComponent({ id, gearId, actorId }) {
+    const existing = await getOwnedComponentRow(database, {
+      id,
+      gearId,
+      actorId
+    })
+    // An already-removed part cannot be retired again — it is history, and the
+    // part that succeeded it is the one on the bike.
+    if (!existing || existing.removedAt) return null
```
llun: This decides the state transition from a read in front of the write: `getOwnedComponentRow` checks `existing.removedAt`, then the `UPDATE` below writes unconditionally. AGENTS.md's Fitness Gear section requires the opposite — *"a state change is a predicate on the UPDATE, never a decision taken from a read in front of it — that goes for the reminder claim and for the retire/unretire toggle alike"* — and `setFitnessGearRetired` further down this same file implements exactly that pattern, with the rationale spelled out in its comment.

As written, two concurrent retire requests (two tabs, a client retry) both read `removedAt: null`, both update, and both return 200 — where the route's own contract answers the second one 404 (`answers 404 when the component was already retired`). The old `replaceFitnessGearComponent` at least kept the check and the writes inside one transaction; this drops both.

The ownership read can stay (`gearId`/`actorId` are immutable, so that check is race-free), but the transition guard belongs in the statement itself, e.g. `.where('id', id).whereNull('removedAt').whereNull('deletedAt').update(...)`, with the affected-row count deciding `null`. Note this method is the easy case for that rule: unlike the gear toggle, "already retired" here is an error rather than a no-op that must still return the row, so 0 rows affected maps straight to the route's 404 and no distinguishing re-read is needed.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -506,9 +506,9 @@ export const GearComponentsCard: FC<Props> = ({
                           variant="ghost"
                           className="text-primary-text"
                           disabled={isPending}
-                          onClick={() => handleReplace(component.id)}
+                          onClick={() => handleRetire(component.id)}
                         >
-                          Replace
+                          Retire
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:512]: This label now collides with the gear-level Retire button in `GearDetailView` (`{isRetired ? 'Unretire' : 'Retire'}`): a bike page renders one "Retire" for the gear itself plus one per installed component row, and they do very different things — one freezes the gear's totals and excludes it from auto-assign, the other closes a single part's install window. Before this PR the row action was "Replace", so there was no collision; the refactor introduces it.

For a screen-reader user the page now offers a list of identical "Retire" buttons with no way to tell which is which, and the test suite already pays for the ambiguity (see the order-dependent `getAllByRole('button', { name: 'Retire' })[0]` in `GearDetailView.test.tsx`). Give the row action a component-scoped accessible name, e.g. `aria-label={\`Retire ${component.componentType}\`}` — the visible text can stay "Retire" since the row context disambiguates it visually.

app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx
```diff
@@ -451,13 +451,14 @@ describe('GearDetailView', () => {
...
-    // Expand the replaced rows: the child's own state is what a remount loses.
+    // Expand the retired rows: the child's own state is what a remount loses.
     fireEvent.click(
-      await screen.findByRole('button', { name: 'Show 1 replaced component' })
+      await screen.findByRole('button', { name: 'Show 1 retired component' })
     )
     expect(screen.getByText('Cassette')).toBeInTheDocument()
 
-    fireEvent.click(screen.getByRole('button', { name: 'Retire' }))
+    const [gearRetireButton] = screen.getAllByRole('button', { name: 'Retire' })
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx]: This works only because the gear header's Retire button happens to render before the components card's Retire buttons in DOM order. Any layout change silently re-points this click at a component row's button instead — the test would then retire a chain while believing it retired the bike, and fail (or worse, pass) for the wrong reason. It's a symptom of the duplicate accessible name flagged on `GearComponentsCard.tsx`; once the row action carries a component-scoped `aria-label`, this can go back to an unambiguous `getByRole('button', { name: 'Retire' })`.

app/api/v1/fitness/gear/[id]/components/[componentId]/retire/route.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const rejection = await rejectComponentsForDevice({
+      req,
+      database,
+      actorId: currentActor.id,
+      gearId: id
+    })
+    if (rejection) return rejection
+
+    const result = await database.retireFitnessGearComponent({
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/components/[componentId]/retire/route.ts:52]: The replaced endpoint deliberately distinguished "sent no body" from "sent something broken": a malformed/truncated payload got a 400 rather than retiring the fitted part while silently discarding whatever the client meant to say (the old route's comment spelled this out). This route now ignores the body entirely, so a client POSTing garbage — or an outdated client still sending the replace payload shape — gets a 200 and a retired component.

Low stakes, since retirement is reversible via PATCH on `removedAt`, and dropping the schema is the right call now that nothing is consumed from the body. But the 400-on-garbage behavior was a considered decision, not boilerplate — worth either keeping it (read the body, 400 if non-empty and unparsable) or a short comment here recording that the body is intentionally ignored so the distinction isn't re-litigated later.

llun (commented):
Review pass over the replace→retire refactor. The rename is carried through consistently — no stale references to `replaceFitnessGearComponent` / `ReplaceGearComponentRequest` / the `/replace` endpoint remain in code or docs, the ratchet entry was correctly removed from `tsconfig.typecheck.json` rather than added to, and the route/database/client layers all line up on the new `{ component }` response shape. Four things need addressing before this lands: one correctness regression against the repo's own state-change rule in the database layer, one accessible-name collision the refactor introduces on the gear detail page, the brittle test selector that collision forces, and one deliberate-behavior question on the route's dropped body validation. Details inline.

lib/database/sql/fitnessGear.ts
llun: Updated `retireFitnessGearComponent` to express the state transition as an atomic predicate directly on the statement (`.where('id', id).whereNull('removedAt').whereNull('deletedAt')`) and return `null` when 0 rows are affected.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:512]: Added `aria-label={`Retire ${component.componentType}`}` to the row's Retire button in `GearComponentsCard.tsx`, giving it a component-scoped accessible name while keeping the visible button text as 'Retire'.

app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearDetailView.test.tsx]: Reverted the test selector in `GearDetailView.test.tsx` back to an unambiguous `getByRole('button', { name: 'Retire' })` now that component rows carry their own component-scoped accessible names.

app/api/v1/fitness/gear/[id]/components/[componentId]/retire/route.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/components/[componentId]/retire/route.ts:52]: Updated `POST /api/v1/fitness/gear/:id/components/:componentId/retire` to read `req.text()` and answer 400 `Bad Request` if a non-empty body contains invalid/unparsable JSON, with a doc comment explaining the contract.

PR #1531, opened 2026-08-24
fix: federate Strava webhook imports at the configured visibility regardless of Strava activity privacy

llun:
## Summary

Fixes Strava webhook fitness imports failing to federate their `Create(Note)` activity upon creation.

### Root Cause
Previously, Strava webhook imports were gated by `isStravaSharedActivity` (`activity.visibility === 'everyone' || activity.visibility === 'followers_only'`). When Strava's API returned `only_me`, omitted `visibility` in its JSON payload, or returned an unexpected string, federation was suppressed, setting `publishSendNote: false` during import and process jobs. The status was created locally at the configured default visibility, but was never delivered to follower inboxes until the user manually updated its visibility on the UI (which sent an `Update(Note)` activity).

### Changes
- In `lib/jobs/importStravaActivityJob.ts`, removed the `isStravaSharedActivity` gate, `resolvedVisibility` mapping, and the unrecognised visibility warning. Status visibility is determined directly from the Strava fitness settings (`fitnessSettings.defaultVisibility`), and `publishSendNote` is used directly across all federation gates.
- Removed unused `mapStravaVisibilityToMastodon` from `lib/services/strava/activity.ts`.
- Updated `docs/fitness-file-storage.md` to document that webhook imports are created and federated at the configured visibility without consulting Strava's privacy flags.
- Updated unit and integration tests in `lib/jobs/importStravaActivityJob.test.ts` and `lib/jobs/importStravaActivityJob.federation.test.ts`.

> **Note**: Posts that previously failed to federate cannot be re-delivered automatically by this change. To federate existing posts, continue using the per-post visibility edit.

## Checklist


lib/jobs/importStravaActivityJob.ts
```diff
@@ -469,44 +469,7 @@ export const importStravaActivityJob = createJobHandle(
     const resolvedVisibility =
```
llun: The should not have any resolvedVisibility any more. This should take from the setting page in Strava fitness setting

lib/jobs/importStravaActivityJob.ts
```diff
@@ -469,44 +469,7 @@ export const importStravaActivityJob = createJobHandle(
...
-      activity.visibility !== 'only_me'
-    ) {
-      logger.warn({
-        message: 'Not federating an import with an unrecognised visibility',
-        actorId,
-        stravaActivityId,
-        stravaVisibility: activity.visibility ?? null
-      })
-    }
+    const shouldFederateImport = publishSendNote
```
llun: If this just assign the variable directly, the use it directly.

lib/jobs/importStravaActivityJob.ts
llun: This should be visibility from the setting, don't use resolvedVisibility one

lib/jobs/importStravaActivityJob.ts
llun: This removes the fail-closed `only_me` handling entirely: an activity the athlete explicitly marked **Only Me** on Strava is now federated at the configured Webhook Activity Visibility — including `public`. The deleted comment block spelled out why that case was gated: "Now that imports federate it would be a leak, and the account default is the one setting a user would never think to check for it."

The bug being fixed is real — Strava omitting or misreporting `visibility` silently suppressed federation of genuinely shared rides. But this fixes the absent/unrecognised-value case by also dropping the guard for the one value that is an explicit per-activity privacy decision. A user who set the webhook visibility to `public` and marks a single ride "Only Me" on Strava (e.g. one starting at their front door) now has it broadcast to every follower inbox, and to relays.

Could we keep the narrow gate for the explicit value only — `publishSendNote && activity.visibility !== 'only_me'` — and fail open just when Strava sends nothing or something new? That fixes the reported bug without overriding a deliberate per-activity choice. If the product decision is to ignore Strava-side privacy entirely, the Strava settings UI should say so: `StravaSettingsForm`'s helper text currently reads "Automatically imported Strava activities will use this visibility" with no hint that the per-activity Strava setting is overridden — which is exactly the "setting a user would never think to check" trap the removed comment warned about.

docs/fitness-file-storage.md
```diff
@@ -309,7 +309,7 @@ This endpoint is **anonymous / unauthenticated** on purpose: public embeds and s
...
...is a v7 `publicId` minted from whichever of those two the import resolved, so it always sorts with the post. The **Strava webhook** is the single caller that opts out, through `postAtImportTime`: it fires minutes after a ride ends and its post is the news that the ride happened, so backdating buried a four-hour ride four hours down the timeline, below everything published while it was still being ridden. That is deliberately unconditional: Strava fires `create` for a late device sync, a back-dated manual entry and a bulk file upload as well, and none of them is treated differently, so restoring a season of old rides files them all at the top in upload order rather than chronologically. Only the stamp on the post moves — the recorded start time lives on `fitness_files.activityStartTime`, which is what the activity page and every fitness rollup read — and a webhook import that merges into a sibling's existing post (the same ride recorded on two devices) keeps whatever stamp that post already carries. Webhook imports are created and federated at the instance's configured "Webhook Activity Visibility" (from fitness settings); the activity's Strava-side privacy setting is not consulted.
```
llun: Two accuracy issues in the added sentence:

1. "the instance's configured 'Webhook Activity Visibility'" — this is a per-**account** fitness setting (`fitness_settings.defaultVisibility` resolved from the webhook token's actor), not an instance-level config. "Instance's" reads as though the admin controls it.
2. "the activity's Strava-side privacy setting is not consulted" is only true for the live webhook path. `retryFitnessImportBatch` deliberately omits `visibility` for a `strava-activity:<id>` batch, so a failed-then-retried import re-derives it via `mapStravaVisibilityToMastodon` — a recovered `only_me` activity is still created at `direct` rather than the configured visibility. If the follow-up your review comments describe (dropping `resolvedVisibility` in favor of the settings value, which the job already loads via `getFitnessSettings`) lands, that inconsistency disappears and this sentence becomes fully true — otherwise it's worth a half-sentence here so the doc doesn't overpromise.

llun (commented):
Review of the federation-gate removal. CI is green and the test changes cover the new behavior well (the `only_me` end-to-end federation case and the `it.each` matrix over all Strava visibility values). Two inline findings below, plus one that sits outside the diff hunks:

**Stale comment outside the diff — `lib/jobs/importStravaActivityJob.ts:944-946`.** The photo-Update gate's rationale still says: "Ungated, a recovery sweep — or an only_me activity, which reaches here having deliberately skipped its Create — would ship a status precisely because it had a photo." After this PR an `only_me` activity no longer skips its Create on the webhook path (`shouldFederateImport` is just `publishSendNote`), so that half of the example describes the behavior this PR removed and will mislead the next reader about what the gate is for. The recovery-sweep half still holds. Per the repo's documentation-maintenance rule (stale guidance is a bug, fixed in the same PR), please update it here.

Note on verification: I reviewed the diff and the full caller graph (`app/api/v1/webhooks/strava/[webhookToken]/route.ts`, `retryImports.ts`, `scripts/fitness/*`) at the head commit but did not run the suite locally; test verification relies on the green CI shards.

llun:
Follow-up to my review's verification note: I have since run the two touched test files locally against the head commit (`vitest run lib/jobs/importStravaActivityJob.test.ts lib/jobs/importStravaActivityJob.federation.test.ts`) — 60/60 pass. The findings in the review stand as posted.

lib/jobs/importStravaActivityJob.ts
```diff
@@ -466,48 +465,10 @@ export const importStravaActivityJob = createJobHandle(
       activityId: stravaActivityId,
       accessToken
     })
-    const resolvedVisibility =
-      visibility ?? mapStravaVisibilityToMastodon(activity.visibility)
+    const statusVisibility =
+      visibility ?? fitnessSettings.defaultVisibility ?? 'private'
```
llun: Is there a visibility constant or enum somewhere in the code? This should use that instead of it's available also in the test.

lib/jobs/importStravaActivityJob.ts
llun: Resolved in da4cebe9f and 43945546f: Removed `resolvedVisibility` and `mapStravaVisibilityToMastodon`. Status visibility is now derived directly from `visibility ?? fitnessSettings.defaultVisibility ?? Visibility.enum.private`.

lib/jobs/importStravaActivityJob.ts
llun: Resolved in da4cebe9f: Removed the `shouldFederateImport` variable and used `publishSendNote` directly across all job publish calls.

lib/jobs/importStravaActivityJob.ts
llun: Resolved in da4cebe9f: `getOrCreateStravaFallbackNote` and `importFitnessFiles` now use `statusVisibility` from the setting directly.

lib/jobs/importStravaActivityJob.ts
llun: Resolved in da4cebe9f: Per requirements, Strava-side activity visibility is no longer consulted; imports and retries consistently use the account's configured default visibility from Strava fitness settings.

docs/fitness-file-storage.md
llun: Resolved in da4cebe9f and 43945546f: Updated docs/fitness-file-storage.md to refer to the per-account fitness setting, and with `mapStravaVisibilityToMastodon` removed, both live webhook and retries consistently use the settings value.

lib/jobs/importStravaActivityJob.ts
llun: Resolved in 43945546f: Replaced raw visibility strings with `Visibility.enum` constants from `@/lib/types/mastodon/visibility` across the job, webhook route, and test files.

lib/jobs/importStravaActivityJob.ts
llun: Acknowledged — dropping the Strava-side gate in favour of the instance setting is your call to make, and the follow-up commits make the behaviour coherent end to end: the job now reads `fitnessSettings.defaultVisibility` itself, so the retry path agrees with the webhook path and the deleted `mapStravaVisibilityToMastodon` has no remaining consumers. Resolving. One leftover from this change: the photo-Update gate comment above (now lines 942-944) still cites "an only_me activity, which reaches here having deliberately skipped its Create" as rationale — that described the removed behaviour, worth a touch-up here or in a fast follow.

docs/fitness-file-storage.md
llun: Both points addressed in da4cebe9f: the sentence now says "the account's configured", and since the job reads `defaultVisibility` itself, the retry path no longer re-derives from Strava — "the activity's Strava-side privacy setting is not consulted" is now unconditionally true. Resolving.

llun:
One small leftover from da4cebe9f (can't anchor inline — the file isn't in the diff): `lib/services/fitness-files/retryImports.ts` lines 137-141 still say the Strava-activity retry omits `visibility` because "the job re-derives the activity's actual Strava visibility. Do not add it back." The job no longer re-derives anything from Strava — omitting `visibility` now means "fall back to the account's configured default" — and `mapStravaVisibilityToMastodon` is gone. The docstring at line ~75 ("re-fetching caption/photos/visibility") is stale for the same reason. Comment-only touch-up; fine here or as a follow-up.

llun:
Addressed in 88b1fe829: Updated the docstring and comments in `lib/services/fitness-files/retryImports.ts` to reflect that omitting `visibility` causes the job to fall back to the account's configured default visibility from fitness settings rather than re-deriving it from Strava.

llun:
Stale-comment re-check at `88b1fe8` (I grepped the head for every phrase tied to the removed behaviour — `only_me`, `mapStravaVisibilityToMastodon`, "re-derive", "visibility from Strava", "restored"):

**Fixed** ✓ — `retryImports.ts`: both the docstring ("re-fetching caption and photos") and the inline comment at 137-141 (now correctly says the omission makes the job fall back to the account's configured default) are accurate.

**Still stale** ✗ — three spots in two files:

1. `lib/jobs/importStravaActivityJob.ts:942-944` — unchanged: "Ungated, a recovery sweep — or an only_me activity, which reaches here having deliberately skipped its Create — would ship a status precisely because it had a photo." The `only_me` half describes the removed behaviour.
2. `scripts/fitness/repairFailedFitnessImports.ts:10-11` — "re-fetches the Strava activity, so caption/photos/visibility are restored". Visibility is no longer restored from Strava; the Strava path here publishes the job with `data: { actorId, stravaActivityId }` (no `visibility`), so the repaired post takes the account's configured `defaultVisibility`.
3. `scripts/fitness/repairFailedFitnessImports.ts:31-34` — the `--visibility` help text: "Only applies to manual-upload batches; Strava-activity retries re-derive the activity's own visibility from Strava." That re-derivation is gone with `mapStravaVisibilityToMastodon`; Strava-activity retries use the configured default instead.

All comment/help-text only; nothing behavioural. CI is green at `88b1fe8`.

llun:
Addressed in 3d75650c0: Cleaned up all remaining stale comments and CLI help text across `lib/jobs/importStravaActivityJob.ts`, `lib/jobs/importStravaActivityJob.test.ts`, `scripts/fitness/repairFailedFitnessImports.ts`, `app/api/v1/fitness/retry-failed/route.ts`, `app/api/v1/fitness/import/[batchId]/route.test.ts`, `lib/components/settings/FitnessFileManagement.tsx`, and `lib/services/fitness-files/retryImports.test.ts`.

llun (commented):
**Review round 3 — clean.** Verified at `3d75650c`:

- All three stale-comment flags from the last round are fixed, and the commit also caught four more I hadn't flagged (the test comments in `import/[batchId]/route.test.ts`, `importStravaActivityJob.test.ts:1412`, `retryImports.test.ts:170`, and the component comment in `FitnessFileManagement.tsx`). The reworded photo-Update gate rationale now cites only the recovery sweep, which is accurate.
- Repo-wide sweep at head: no remaining references to `mapStravaVisibilityToMastodon`, `isStravaSharedActivity`, `shouldFederateImport`, or any "re-derive visibility from Strava" phrasing. The only `only_me` left in non-test code is the `StravaActivityVisibility` API-response type in `lib/services/strava/activity.ts` — correctly retained, since Strava still sends the field.
- Local run of every affected test file at this commit: `importStravaActivityJob.test.ts`, `importStravaActivityJob.federation.test.ts`, `retryImports.test.ts`, `import/[batchId]/route.test.ts`, `retry-failed/route.test.ts` — 93/93 pass. CI is fully green.

One observation, not a finding: `scripts/fitness/runImportStravaActivity.ts` (the CLI `stravaAuth` path) builds synthetic settings with no `defaultVisibility`, so those imports now land at `private` where they previously mirrored the Strava-side value (`everyone` → `public`). Fail-safe direction, they never federate (`publishSendNote` unset), and the script's header never documented a visibility — mentioning only so the change is a conscious one.

The branch shows `mergeable_state: behind` — no conflicts, so squash-merge works as-is, but you may want to update it first. LGTM.

llun:
Updated `scripts/fitness/runImportStravaActivity.ts` in e35b426b1 to accept an optional `--visibility <public|unlisted|private|direct>` CLI flag, forwarding it to `importStravaActivityJob`, and added unit tests in `scripts/fitness/runImportStravaActivity.test.ts`.

llun:
Updated `scripts/fitness/runImportStravaActivity.ts` in 725d8846a to make `--visibility <public|unlisted|private|direct>` a required CLI parameter, and updated unit tests in `scripts/fitness/runImportStravaActivity.test.ts` accordingly.

PR #1536, opened 2026-08-25
fix: delete status locally before queueing federation fan-out

llun:
## Summary

Deleting a status ran the entire ActivityPub fan-out **inline, ahead of the local delete**. `deleteStatusFromUserInput` resolved every delivery inbox (followers + relays + explicit recipients, including live remote actor fetches for anyone not already in the local database), `Promise.all`'d a signed `Delete`/`Tombstone` POST to each one at a 10s timeout with a retry, and only then called `database.deleteStatus`.

Two consequences:

- **The post stayed on screen until every remote inbox had been contacted.** The delete dialog awaits the `DELETE` response before the timeline drops the row, so the author's own post remained visible for as long as the slowest server took to answer or time out.
- **The delete could be abandoned before it happened.** Inbox resolution ran ahead of `database.deleteStatus`, and that path throws — a database error while collecting followers, relays or recipients skipped the delete entirely, after Deletes had already gone out to remotes.

This makes the local delete happen first and moves delivery to a new `SendDeleteNoteJob`, matching what create (`SendNoteJob`), edit (`SendUpdateNoteJob`) and boost already do. Status deletion was the last status-lifecycle action still federating inline. The status now leaves the author's timelines as soon as the transaction commits.

Both delete entry points — `DELETE /api/v1/statuses/:id` and the legacy `DELETE /api/v1/accounts/outbox` — are fixed through the shared action, with no route logic changes. Media and fitness cleanup stay in the route, where the ids are already collected before the delete.

### The job cannot load the status it federates

`database.deleteStatus` is a cascading hard delete, so by the time the job runs the row is gone. The payload therefore carries the audience captured **before** the delete, and `getFederatedStatusDeliveryInboxes` was narrowed to accept just the `to` and `cc` fields — already all it read of the status — so no call site changed.

This is exactly what `sendUndoAnnounceJob` gets wrong today: `undoAnnounce` hard-deletes then enqueues, and the job bails on a missing status, so unboosts never federate. Noted here as a **pre-existing bug left for a follow-up PR** — it is not touched by this change.

### Two deliberate behaviour choices

- **Dedup id is `hash(statusId + '#delete')`, not `hash(statusId)`.** The queue deduplicates globally on message id across job names with a window that outlives consumption, and the create/update jobs already publish under the bare status id. Without the suffix, deleting a status that was posted or edited inside that window would be silently dropped and never federate.
- **The enqueue is wrapped in try/catch.** The local delete has already committed and cannot be undone, so surfacing a queue failure as a failed request would tell the author their post is still there when it is gone — and would skip the route's media cleanup. The failure is logged with a stack; remote copies reconcile on their next fetch, which 404s.

Under the default in-process queue (`NoQueue`), `publish` still runs the job inline, so a single-instance deployment continues to wait for fan-out — the same as create/edit/boost. What changes there regardless of backend: the delete has already committed first, inbox-resolution failures can no longer abandon it, and an inline failure can no longer fail the request. QStash deployments get the fully async path.

## Test plan

- **`lib/jobs/sendDeleteNoteJob.test.ts`** (new) — every delivery case deliberately uses a `statusId` that is **not** in the database, the regression guard for the whole design. Covers fan-out for a deleted status, direct statuses preserving their original `to`/`cc` (and not reaching the follower audience), non-direct statuses being addressed to Public instead of their raw audience, an unknown actor as a no-op, per-inbox failure isolation, and that the job is registered in the `JOBS` map.
- **`lib/actions/deleteStatus.test.ts`** (rewritten) — asserts the local delete is ordered *before* the publish via `invocationCallOrder`, the exact job payload, non-owner and missing-status no-ops, and that a rejecting `publish` still leaves the status deleted.
- **`app/api/v1/statuses/[id]/route.test.ts`** — added a route-level assertion that the job is queued with the pre-delete audience. The `deleteStatus` entry was dropped from the `@/lib/activities` mock, which now fails loudly if delivery ever creeps back into the request path.
- **`app/api/v1/accounts/outbox/route.test.ts`** — success-path coverage for the action's other caller, which previously only tested invalid JSON.

Full suite green: 869 files, 10,936 tests.

## Review

Two sub-agent review rounds (four reviewers, then three), each finding attacked by two independent skeptics. Round 1 refuted five design challenges as pre-existing or intentional, and found three of my own tests that **could not fail** — the per-inbox isolation test in particular, because `postActivityToInbox` swallows socket errors so the sender never rejects. All three now fail against a mutant of the code they guard. Round 2 produced no surviving findings, but did prompt a factual correction: a failing *remote server* was never able to abort the old local delete (the sender and `getActorPerson` both swallow) — only inbox resolution could, which is what the summary above and the docs now say. Details in the [round 1 comment](https://github.com/llun/activities.next/pull/1536#issuecomment-5415857719).

`AGENTS.md` and `REVIEW.md` gained a **Status Delete Federation** section recording the invariants, chiefly that this job must never load the status it federates.

## Checklist

- [x] Docs updated: grepped `*.md` and `docs/` — nothing described the delete flow or enumerated jobs, so nothing went stale; added the new invariants to `AGENTS.md` / `REVIEW.md`
- [x] If migrations changed: n/a — no schema changes, payload-only job

https://claude.ai/code/session_01REnirYGfFfThi1SqsFX9zw

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1557,6 +1557,41 @@ consistency is enforced by keeping the wiring in one place rather than per page.
   own. There is deliberately no prop for hiding one of the menu's own items;
   that is the per-page drift this whole section exists to prevent.
 
+## Status Delete Federation
+
+- **The local delete commits FIRST and the `Delete` fans out from
+  `SendDeleteNoteJob` afterwards — and that job must never load the status it is
+  announcing the death of.** `database.deleteStatus` is a cascading hard delete,
+  so by the time the job runs the row (and its whole same-actor reply subtree) is
+  gone. `deleteStatusFromUserInput` (`lib/actions/deleteStatus.ts`) therefore
+  captures the audience before deleting and publishes it in the payload
+  (`{ actorId, statusId, to, cc }`), and `getFederatedStatusDeliveryInboxes`
+  takes `Pick<Status, 'to' | 'cc'>` rather than a whole `Status` so a caller
+  holding only that payload can still resolve inboxes. **Do not "unify" this job
+  onto `loadStatusAndActor`**: that is exactly the bug `sendUndoAnnounceJob` has
+  — `undoAnnounce` hard-deletes then enqueues, the job bails on `!status`, and
+  the `Undo` never federates. Every delivery test in
+  `lib/jobs/sendDeleteNoteJob.test.ts` uses a statusId that is deliberately NOT
+  in the database for that reason.
+- **The dedup id is `getHashFromString(`${statusId}#delete`)`, and the suffix is
+  correctness.** The queue deduplicates globally on `message.id` across job
+  names, with a window that outlives consumption, and `SendNoteJob` /
+  `SendUpdateNoteJob` already publish under the bare `getHashFromString(statusId)`
+  — without the suffix, deleting a status posted or edited inside that window is
+  silently dropped and never federates.
+- **The `publish` is wrapped in try/catch on purpose.** The hard delete has
+  already committed and cannot be undone, so a failed enqueue must not surface as
+  a failed request — that would tell the author their post is still there when it
+  is gone, and would skip the route's media cleanup. It is logged with a stack;
+  remote copies reconcile on their next fetch, which 404s.
+- Delivery errors never reach that catch: `postActivityToInbox` swallows every
+  network failure and returns `undefined`, so the activities `deleteStatus`
+  sender does not reject. A test that fails the _socket_ therefore proves nothing
+  about the job's per-inbox guard — mock the **sender** to reject, which is the
+  only seam that can.
+- Known gap, unchanged by the move to a job: a cascaded reply subtree is deleted
+  locally but only the top status's `Delete` federates.
+
 ## Better-auth Plugin Guidelines
 
 - **Do not register a better-auth plugin unless its required database tables exist** in the Knex migrations. The custom `knexAdapter` does not auto-create tables; missing tables will cause runtime errors.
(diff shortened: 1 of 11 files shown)


llun:
## Sub-agent review round 1

Four reviewers over independent dimensions (federation correctness, queue/job semantics, project conventions, test quality), with each design-level finding then attacked by two independent skeptics.

### Refuted — design challenges that did not hold

Five findings went through adversarial verification and **all five were refuted** as pre-existing properties or the PR's stated design. Recording them so they don't get re-litigated:

| Finding | Why refuted |
| --- | --- |
| Delete is now at-most-once (`retries: 0`) | Pre-existing and global to all 15+ jobs; untouched here. The QStash SDK also retries transport-level failures internally (~6 attempts) before `publish` can throw. |
| A failed enqueue loses the Delete forever | Same commit-then-best-effort-enqueue shape already ships in `undoAnnounce`, `createNote`, `announce`, `updateNote` and others; no durable outbox exists anywhere. Deliberate, and directly tested. |
| Cascaded reply deletes never federate | Confirmed against `main`: the old inline code sent exactly one Delete too. Real gap, zero delta from this PR. |
| The `catch` mislabels an inline fan-out failure under NoQueue | Matches ~14 existing `"Failed to queue …"` call sites, and this one is better instrumented than those (it passes `err`, so the formatter emits `stack_trace`). |
| Actor-not-found drops the Delete with no record | Matches seven sibling jobs; with no actor row there is no key to sign with, so nothing deliverable is being discarded. |

One verifier also noted the old path was strictly worse than described: a throw during inbox resolution left the post **undeleted locally after Deletes had already been POSTed to remotes** — split-brain, not just slowness.

### Fixed — the review found three tests that could not fail

The most valuable findings were about my own tests, and three reviewers independently converged on the first one:

1. **The per-inbox isolation test asserted nothing.** `postActivityToInbox` catches every network error and returns `undefined`, so the activities `deleteStatus` sender never rejects — failing the *socket* never reached the job's guard. Now mocks the **sender**, the only seam that can reject, and lets the job resolving be the assertion (without the guard, `Promise.all` propagates and `withSpan` rethrows).
2. **Nothing covered the JOBS registry** — the one new wiring point this PR adds. Dropping that line would have sent every delete to a job nothing handles, with a single "Unknown job name" log as the only symptom.
3. **The `isDirect` branch was uncovered.** The non-direct test asserted a `to` that is identical whether or not the branch runs. A followers-only case now pins it, asserting on the raw request body so the absent `cc` is actually checked.

All three now **fail against a mutant** of the code they guard, verified individually.

Also added: a success-path test for the legacy `DELETE /api/v1/accounts/outbox` (the action's other caller, previously only tested for invalid JSON), a log on the actor-not-found drop, and the invariants written into `AGENTS.md` / `REVIEW.md` — chiefly that this job must never load the status it federates, which is exactly the trap `sendUndoAnnounceJob` is in today.

Local suite: 869 files, 10,936 tests, green.

---

PR #1537, opened 2026-08-25
minor: show activity emoji and duration on the fitness overview, and filter recent activities by type

llun:
## Summary

Reworks the fitness overview's activity breakdown to match the design system's Activities card (`ui_kits/web/FitnessPage.jsx`), and makes each activity name a filter for the feed below.

- **"Activity Mix" → "Activities"**, with a bar-chart icon in front of the heading.
- **Emoji per row**, resolved through the existing `getActivityPresentation` table rather than a new lookup — so a qualified bike sport keeps its own glyph (🚵 for a mountain bike ride, not the generic 🏋️ a raw-string-only lookup gave it).
- **New Duration column** between Count and Distance, from the `totalDurationSeconds` the summary call already returned. No new query.
- **Each activity name is a link** that narrows "Recent activities" to that stored `activityType`. Clicking the selected row clears the filter, as does a chip beside the "Recent activities" heading; a filter that matches no posts renders "No recent Walk activities have been posted." instead of collapsing the section, because the chip is the only way back out of it.

### Notes for review

- **The filter lives in a `?activity=` search param, not dashboard state.** The list is then narrowed by the same server query that builds it. A client-side filter over the five posts already fetched would show whichever of them happened to match, not the five most recent of that type. Verified in a browser that the dashboard's client state (a custom date range) and the scroll position both survive the navigation.
- **A supplied `?activity=` value is resolved against the actor's own stored types** (`getDistinctActivityTypesForActor`) and only then becomes a filter. Unvalidated, it reached three surfaces that cannot defend themselves — the empty-state sentence and the chip's text and `aria-label`, where a crafted link put attacker-chosen words on the victim's own signed-in page in the instance's voice, plus `.where('activityType', …)`, where PostgreSQL rejects a NUL byte and turned a junk URL into a 500. Matching against the actor's own list needs no length or character rules that could drift from the column. The lookup runs only when a param was supplied. That method's predicate is a strict superset of the summary's, so no link the table offers can be rejected.
- **The row name keeps the column's own word, not the post caption's.** `getActivityPresentation` deliberately folds several stored spellings onto one caption (`ride` and `cycling` both read "Cycling"), which is right for a post and wrong for a table whose two rows carry different numbers and different filters. `formatActivityType` therefore moves out of the client dashboard into `activityPresentation.ts` as `formatActivityTypeLabel` — the chip is rendered by a Server Component, which must not read a runtime value out of a `'use client'` module — and the heatmap view's copy of the same expression now calls it too.
- **Capitalising is itself case-insensitive**, so `run` and `Run` still land on one label, and both are real on an instance whose pre-canonicalisation rows have not been swept. `buildActivityTypeLabels` qualifies a colliding label with the value that produced it ("Run (run)" / "Run (Run)") and only a colliding one, so an ordinary instance renders exactly as before.
- **An `sr-only` `role="status"` region reports the filter's outcome.** Nothing else can: `scroll={false}` suppresses the post-navigation focus move, the page title is static so Next's route announcer has nothing to read, and `aria-current` flipping on the already-focused link is not reliably re-announced. The section is always rendered and only its contents are gated, so the region keeps one stable position — a live region that mounts, or that React re-creates because it moved, announces nothing.
- **`prefetch={false}` on the activity links**: one link per activity type, all pointing at this same `force-dynamic` page, so prefetching them would re-run the whole overview render once per row on screen.
- **`break-words`, not `wrap-anywhere`, on the activity name.** This table auto-sizes rather than snapping, and breaking anywhere drops the name column's min-content contribution to one character — which is what let table layout squeeze "Walk" into "Wal / k" at 390px. Caught in the browser pass, not by a test.
- `tsconfig.typecheck.json` loses one ratchet entry: `RecentFitnessActivities.test.tsx` is edited by this PR and is now type-clean.
- No migrations, no new env vars, no new API routes, and no docs went stale (grepped `*.md` and `docs/` for the renamed card and the touched helpers).

### Review

Four rounds of the sub-agent loop ran on this PR — 74 agents, every finding put to two adversarial refuters and kept only on a unanimous confirm. **Ten issues found and fixed** (one major, nine minor); round 4 came back clean. Round 2 was cut short by a rate limit and its findings were self-verified instead — [that is disclosed on its own review](https://github.com/llun/activities.next/pull/1537#pullrequestreview) rather than papered over. Full record in the round summaries on this PR.

Every fix was mutation-tested: the defect each one describes now fails a test. `main` was merged in afterwards (`next` 16.3.1 → 16.3.2) and the two Next-behavioural guarantees — live-region node identity across an RSC navigation, and `scroll={false}` — were re-verified in a browser against the new version.

## Screenshots

Verified in Chromium against a local SQLite seed (`scripts/mock/createMockUser.ts` + `createMockFitnessData.ts`) at 1280px, 768px and 390px, plus dark mode:

| State | What it shows |
| --- | --- |
| Overview, unfiltered | "Activities" heading with the bar-chart icon; 🚴 Ride / 🏃 Run / 🚶 Walk rows over Count · Duration · Distance |
| `?activity=run` | "Run" underlined with `aria-current="true"`, a `Run ×` chip beside "Recent activities", and only the run posts below |
| `?activity=walk` | The dashed "No recent Walk activities have been posted." panel, with the chip still offering the way out |
| Crafted / NUL / 2000-char param | Renders the plain unfiltered overview: 200, no chip, no horizontal overflow |
| Legacy mixed-case row | "Run (run)" and "Run (Run)" as separate rows with their own totals |
| 390px | Each name on one line beside its emoji, all four columns visible |

Screenshots are attached in the session handoff rather than inline — this environment can't upload binaries to the PR. Happy to re-post them here if you'd like them in the thread.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn typecheck`, `yarn build`, and `yarn test` all pass locally, run in that order (871 test files, 10,985 tests, on the merged tree)
- [x] Docs updated: grepped `*.md` and `docs/` — nothing referenced "Activity Mix" or the moved helper, so no doc went stale
- [x] If migrations changed: n/a, this PR adds none

---
_Generated by [Claude Code](https://claude.ai/code/session_01DmZkaf8Y5j42gJ9SBhF3W1)_


app/(timeline)/fitness/activityFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const ACTIVITY_FILTER_PARAM = 'activity'
+
+const FITNESS_OVERVIEW_PATH = '/fitness'
+
+/**
+ * The stored `activityType` the recent-activities list is narrowed to, or
+ * `undefined` for no filter.
+ *
+ * A repeated or blank param is no filter: the value is matched verbatim against
```
llun: **minor — the doc contradicts the code, in the module that exists to be the contract**

This says "A repeated or blank param is no filter", but line 29 is `Array.isArray(value) ? value[0] : value`, so a repeated param filters on its **first** value — and `activityFilter.test.ts` pins exactly that (`{ activity: ['run', 'ride'] }` → `'run'`). Only the blank half is true, and the justification that follows ("`''` is not a value any activity is stored under") explains only the blank case.

No runtime bug today, because `getActivityFilterHref` only ever emits a single param. It matters because this module was created precisely so three callers agree on one spelling of the `?activity=` contract, so its JSDoc is the spec they get read against: a maintainer trusting it either believes repeated params are already neutralised, or "fixes" the code toward the comment and silently changes behaviour a test asserts.

Raised independently by three of the six reviewers (correctness, invariants, quality).

---

app/(timeline)/fitness/activityFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+ * `fitness_files.activityType`, and `''` is not a value any activity is stored
+ * under.
+ */
+export const readActivityTypeParam = (
+  params: Record<string, string | string[] | undefined>
+): string | undefined => {
+  const value = params[ACTIVITY_FILTER_PARAM]
+  const activityType = Array.isArray(value) ? value[0] : value
+  return activityType ? activityType : undefined
```
llun, line 30: **minor (two reviewers, converging) — the param is returned unchecked, and three surfaces downstream cannot defend themselves**

`readActivityTypeParam` returns whatever the query string carried behind a bare truthiness check. Nothing verifies the value names an activity type that exists, and it then reaches:

1. **Prose on the victim's own signed-in page.** `RecentFitnessActivities.tsx:60` splices the label into `No recent {label} activities have been posted.`, and the empty state renders *precisely when the value matches nothing* — so a crafted value always takes the prose path. `/fitness?activity=rides.%20Security%20notice:%20re-verify%20at%20evil.example` reads as ordinary muted body copy in the instance's own voice. React escapes it, so no XSS, but this is the same reflected-copy shape `REVIEW.md` documents for the auth error page — and unlike that page there is neither an allow-list nor a length cap.
2. **The chip's text and `aria-label`**, so a screen reader announces the attacker's sentence as "Clear … filter".
3. **Layout.** The chip is `inline-flex` with no `min-w-0`/wrapping, so a single long unbroken token sets its min-content width and overflows the `max-w-content` column.

A third reviewer raised, and the verifiers refuted on the narrower grounds that `AGENTS.md`'s coercion rule is numeric-column-only, that **a `%00` in the param reaches `.where('activityType', …)` and PostgreSQL rejects a NUL in a text comparison (22021)** — a junk URL becoming a 500. The scope objection is fair, but the failure is real and the same root cause covers it.

The bound that needs no upkeep is the actor's own data: resolve the param against `getDistinctActivityTypesForActor` (same completed/primary/not-deleted predicate as the recent-activities query, so the accepted set is exactly what that query could return and exactly what the Activities table links to) and treat a non-match as no filter. That closes all three surfaces plus the NUL path at once, with no length or character rules that could drift from the column.

---

app/(timeline)/fitness/RecentFitnessActivities.tsx
```diff
@@ -1,40 +1,79 @@
...
+            className="inline-flex items-center gap-1 rounded-full border px-2.5 py-0.5 text-xs font-medium text-primary-text transition-colors hover:bg-muted"
+          >
+            {label}
+            <X className="size-3" aria-hidden="true" />
+          </Link>
+        )}
+      </div>
+      {statuses.length === 0 ? (
+        <p className="rounded-lg border border-dashed p-6 text-center text-sm text-muted-foreground">
+          No recent {label} activities have been posted.
```
llun: **minor — applying or clearing the filter is silent to assistive tech**

The navigation swaps the whole list below and moves nothing else, and three things line up against a screen-reader user noticing:

- `scroll={false}` on both filter controls suppresses Next's post-navigation focus move, so focus stays on the link and the changed region is never entered.
- `metadata.title` on this page is the static `'Activities.next: Fitness'`, so Next's `AppRouterAnnouncer` — which reads `document.title` after a navigation — has nothing new to say.
- The only other DOM signal is `aria-current` flipping on the link the user is already focused on, which screen readers do not reliably re-announce for an attribute mutation on the focused element.

On a phone the list is well below the fold, so there is no incidental visual feedback either. An NVDA user pressing Enter on "Gravel Ride" hears nothing at all and cannot tell whether the link did anything — same for the reverse toggle and for the filtered empty state appearing.

A `role="status"` `sr-only` region naming the current scope is server-rendered, changes text on every filtered render, and costs nothing visually.

---

app/(timeline)/fitness/page.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+    // learns that from the same param the query above was built from.
+    expect(
+      findElementByType(element, ActorFitnessDashboard)?.props
+    ).toMatchObject({ selectedActivityType: 'gravel_ride' })
+  })
+
+  // How a param is read is `activityFilter.test.ts`; what this asserts is that
+  // "no filter" leaves the key off the query entirely rather than passing
+  // `activityType: undefined`, which the database method reads differently.
+  it('leaves the filter off the query for a blank param', async () => {
```
llun, line 224: **major — this assertion cannot fail for the thing its comment says it guards**

The comment claims the database method "reads differently" an explicit `activityType: undefined` versus an absent key. It does not: `getFitnessFilesByActor` gates on `activityType !== undefined` (`lib/database/sql/fitnessFile.ts:493`), so the two are the same code path and only `null` means anything else. The `page.tsx:87` comment beside the spread has the same slip in reverse — it says "only an ABSENT key" means every type, then immediately concedes `undefined` behaves identically.

Worse, the test could not catch the distinction even if it existed. `toHaveBeenCalledWith` compares with `toEqual` semantics, which treat an absent key and a key set to `undefined` as equal — a reviewer reproduced it in this repo (Vitest 4.1.10): `fn({a: 1, b: undefined}); expect(fn).toHaveBeenCalledWith({a: 1})` **passes**. So both this test and the "lists every activity type" one hold whether or not the spread is there; deleting the `...(activityType ? … : {})` guard breaks nothing.

Two reviewers (tests, quality) landed on this from different directions. The code is fine — it is the comments that are wrong and the guard that is unpinned, and in this repo a comment asserting something false about a neighbouring module is a real defect. Assert on the recorded call's own keys so the test can fail, and fix both comments to say what is actually true.

---

app/(timeline)/fitness/ActorFitnessDashboard.test.tsx
```diff
@@ -210,4 +246,95 @@ describe('ActorFitnessDashboard', () => {
...
+  it('reports a duration column beside the count and distance', async () => {
+    render(
+      <ActorFitnessDashboard
+        actorId={ACTOR_ID}
+        currentTime={FIXED_CURRENT_TIME}
+      />
+    )
+
+    expect(
+      await screen.findByRole('columnheader', { name: 'Duration' })
```
llun: **minor — the new column is verified in two halves that are never tied together**

`findByRole('columnheader', { name: 'Duration' })` proves only that a header with that text exists *somewhere*; the assertions below check cell ordering by index. Nothing correlates a header position with the cells under it, so the header row can be reordered independently of the body and the suite stays green.

That is not hypothetical here: this PR inserts Duration between two columns that are both right-aligned and both `tabular-nums`, i.e. visually interchangeable. The reviewer applied the mutation — swap only the `Duration` and `Distance` `<th>` labels, leave the `<td>`s alone — and all 12 tests still passed, with the Run row rendering `1h 30m` under "Distance" and `15.0 km` under "Duration".

Asserting the header row as a sequence pins it.

---

app/(timeline)/fitness/activityFilter.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      params: { sport: 'run' },
+      expected: undefined
+    }
+  ])('reads $description', ({ params, expected }) => {
+    expect(readActivityTypeParam(params)).toBe(expected)
+  })
+})
+
+describe('getActivityFilterHref', () => {
+  it('links an unselected activity to its own filter', () => {
```
llun: **minor — three `it` blocks that differ only by data**

Each body is a single `expect(getActivityFilterHref(<type>, <isSelected>)).toBe(<expected>)` with an identical shape and no setup. `AGENTS.md` → **Testing Guidelines** asks for a table-driven `it.each([...])('$description', …)` for cases that differ only by input and expected value, and the `readActivityTypeParam` block directly above already does exactly that — so the file is inconsistent with itself as well as with the convention.

---

llun (commented):

Six reviewers fanned out over the diff (correctness, project invariants, UI/a11y, security, tests, reuse/quality), each finding then put to two adversarial refuters with distinct lenses and kept only on a unanimous confirm.

**19 raised → 10 confirmed → 6 distinct issues** after deduplication (the `activityFilter.ts` JSDoc was found independently by three reviewers). Nothing was a blocker; one major, five minor.

Notable **refutations**, recorded so they are not re-litigated later:

- *"`formatActivityTypeLabel` prints 'Ebike Ride' where the app says 'E-bike ride'"* — refuted on scope: the helper is a character-for-character move of the dashboard's pre-existing `formatActivityType`, so `/fitness` printed that before this PR and prints it after. Routing it through `getSportLabel` instead would also break the thing the split exists for (one row per stored value).
- *"The filtered empty state lies when postless files fill the 5-row window"* — mechanically real, but the `!hasMore` invariant it invokes is explicitly scoped to a paginated feed, and this list does not page.
- *"The heatmap view's new import feeds an unreachable branch"* — true, but the branch was already dead on the parent commit; this PR only replaces a local copy of the expression with the shared one.
- *"A NUL byte 500s the page on PostgreSQL"* — refuted on the narrow ground that the `AGENTS.md` coercion rule is numeric-column-only. The underlying failure is real, and the fix for the unbounded-param thread closes it anyway.

Fixes for all six are being pushed now; I'll reply on each thread with what changed and re-run the loop against the new head.

---

app/(timeline)/fitness/activityFilter.ts
llun: Fixed in 58ded3a. Kept the first-value behaviour (it is what the test pins and what a hand-typed duplicate should sensibly do) and rewrote the doc to describe it: "The `?activity=` value as it arrived, or `undefined` when the param is absent or blank. A repeated param takes the FIRST value."

The doc block also now says explicitly that this returns the raw request rather than a filter, and points at the new `resolveActivityTypeFilter` as the thing every caller must go through — which is the other half of this module's contract after the unbounded-param thread.

---

app/(timeline)/fitness/activityFilter.ts
llun: Fixed in 58ded3a, taking the suggested approach.

New `resolveActivityTypeFilter(requested, availableActivityTypes)` in the same module, and `page.tsx` now honours `?activity=` only when it names one of the actor's own stored types, read from `getDistinctActivityTypesForActor`. A non-match resolves to `undefined`, so the page renders as the plain unfiltered overview — no chip, no filtered empty state, nothing reaching `.where()`. The lookup runs only when a param was actually supplied, so the unfiltered page still issues the same two queries it did before.

That closes all four surfaces at once: prose, `aria-label`, chip layout, and the NUL byte. Belt-and-braces on the two rendering surfaces anyway, since a genuine 255-character stored type would still be wide: the chip gains `max-w-full` + `break-words` (and the `X` icon `shrink-0`), and the empty-state paragraph gains `break-words`.

Verified in a browser against the local seed — for each of an unknown type, the crafted `rides. Security notice: re-verify at evil.example`, `run%00`, and a 2000-character value: HTTP 200, no chip, `document.scrollWidth === clientWidth` (no horizontal overflow), and the status region reading "Showing all recent activities". A valid `?activity=run` still filters, and `?activity=run&activity=ride` still resolves to `run`.

Covered by `resolveActivityTypeFilter`'s own table-driven tests (unknown type, wrong case, crafted prose, 2000 chars, NUL byte, empty stored list) and by a new `it.each` in `page.test.tsx` asserting each of those reaches neither the query nor either child component. Mutation-tested: dropping the `includes` check fails 10 tests.

---

app/(timeline)/fitness/RecentFitnessActivities.tsx
llun: Fixed in 58ded3a with the suggested `role="status"` region, rendered inside the `<section>` between the heading row and the list:

```tsx
<p role="status" className="sr-only">
  {label ? `Showing recent ${label} activities` : 'Showing all recent activities'}
</p>
```

It is server-rendered and its text differs on every filtered render, so the RSC update that swaps the list also swaps this text and the announcement fires. A comment above it records the three reasons nothing else announces (the `scroll={false}` focus suppression, the static page title, and `aria-current` on the already-focused link), so a later change that removes one of those knows what it is standing on.

Two table-driven tests assert the region's text in both states; mutation-tested by dropping `role="status"`, which fails both.

Confirmed in the browser that it does not disturb the visual layout — the page renders pixel-identically to before.

---

app/(timeline)/fitness/page.test.tsx
llun: Confirmed and fixed in 58ded3a. You're right on both halves — I'd flagged the same slip myself while re-reading, and the reproduction settles the `toEqual` semantics.

**The comments.** `page.tsx` now says what is actually true: the spread exists so "no filter" passes no `activityType` at all, the method *happens* to read an explicit `undefined` the same way, but `null` means something else there entirely ("activities with no recorded type") — so the narrowing is stated only when meant. The `page.test.tsx` comment no longer claims a distinction the method does not make.

**The assertion.** Added a small helper that reads the recorded argument's own keys, so presence and value are separable:

```ts
const activityTypeArgumentOf = (call: Record<string, unknown>) =>
  Object.hasOwn(call, 'activityType')
    ? { present: true, value: call.activityType }
    : { present: false }
```

Used in the unfiltered case, the blank-param case and the new invalid-param cases (`{ present: false }`) and in the valid-filter case (`{ present: true, value: 'gravel_ride' }`), with a doc block recording *why* the object-literal form cannot express this.

Mutation-tested exactly as you described: replacing the spread with a plain `activityType` key now fails 6 tests in `page.test.tsx`, where before it failed none.

---

app/(timeline)/fitness/ActorFitnessDashboard.test.tsx
llun: Fixed in 58ded3a — the header row is now asserted as a sequence, replacing the bare existence check:

```ts
expect(
  (await screen.findAllByRole('columnheader')).map((header) => header.textContent)
).toEqual(['Activity', 'Count', 'Duration', 'Distance'])
```

with a comment recording why presence alone was not enough here (Duration sits between two right-aligned `tabular-nums` columns, so a swapped label is invisible to the eye as well as to the old assertion).

Reproduced your mutation before and after: renaming just the `Duration` `<th>` left all 12 tests green previously, and now fails this one.

---

app/(timeline)/fitness/activityFilter.test.ts
llun: Fixed in 58ded3a — the three blocks are now one `it.each([...])('$description', …)`, matching the `readActivityTypeParam` block above it. The two case-specific comments (URL encoding, the clear-filter toggle) moved onto their table rows rather than being lost.

The new `resolveActivityTypeFilter` suite added in the same commit follows the same shape.

---

app/(timeline)/fitness/RecentFitnessActivities.tsx
```diff
@@ -1,40 +1,102 @@
...
-      />
+      <p role="status" className="sr-only">
+        {label
+          ? `Showing recent ${label} activities`
+          : 'Showing all recent activities'}
+      </p>
+      {hasNothingToShow ? null : (
+        <>
+          <div className="flex flex-wrap items-center gap-2">
+            <h2 className="text-base font-medium">Recent activities</h2>
```
llun, line 75: **minor — the live region added in round 1 is below the early return, so it cannot announce the one transition it was added for**

Raised independently by both reviewers that completed this round (regression, a11y/HTML). Their verifiers were killed by the rate limit, so I verified this one myself.

A live region only announces a change to content it **already held**. One that mounts carrying its own text is silent. This region sits under `if (statuses.length === 0 && !activityType) return null`, so it is created rather than updated whenever that branch flips.

That branch is reachable, and it is the worst possible case to miss: an actor whose fitness activities have no surviving posts (every recent `fitness_files` row's `statusId` nulled by a deleted status) renders no "Recent activities" section at all while unfiltered — but `hasFitnessData` is still true, the Activities table is fed by the summary rather than by these posts, and its rows are still links. So they can apply a filter, and *that* navigation moves nothing else on the page either. It is precisely the case the region exists for, and the only one it cannot serve.

Worth recording that the obvious fix is **not enough**, which I only found because the new identity assertion caught it: hoisting the region above the early return still leaves it moving in the tree, because with a filter set the section renders and the region goes from a bare top-level `<p>` to a child of `<section>`. React reconciles those as different nodes and re-creates it — silent again.

---

llun: Fixed in eed40f2.

The section is now **always** rendered and only its contents are gated, so the region keeps one stable position in both states:

```tsx
return (
  <section className="space-y-3">
    <p role="status" className="sr-only">
      {label ? `Showing recent ${label} activities` : 'Showing all recent activities'}
    </p>
    {hasNothingToShow ? null : (<>…heading, chip, list…</>)}
  </section>
)
```

As noted above, hoisting it above the early return was my first attempt and it was wrong — the new `expect(after).toBe(before)` assertion failed, because with a filter set the section renders either way and the region moved from a bare `<p>` into it. Good argument for writing the identity assertion rather than another text assertion.

Two tests cover it now: the empty-and-unfiltered branch asserts the heading, list and chip are all absent while the region is present, and a rerender test asserts the *same DOM node* carries the updated text across the postless actor's first filter. Mutation-tested — restoring `return null` fails both.

Also confirmed in a real browser rather than only in jsdom, since this is ultimately about how React reconciles an RSC navigation: I stamped `data-probe` on the live region, clicked a filter link, and after the navigation the attribute is still on the same node with the text updated in place. `sr-only` keeps it out of flow, so the always-rendered section costs no layout — document height is byte-identical to before (1271px unfiltered, both before and after this change).

One thing I did not do: `aria-live` is left implicit via `role="status"` rather than spelled out. That is the same pattern the rest of this codebase uses, and adding both is redundant.

---

app/(timeline)/fitness/ActorFitnessDashboard.tsx
```diff
@@ -381,31 +406,107 @@ export const ActorFitnessDashboard: FC<Props> = ({ actorId, currentTime }) => {
...
+                      >
+                        Count
+                      </th>
+                      <th
+                        scope="col"
+                        className="px-2 py-2 text-right font-medium"
+                      >
+                        Duration
+                      </th>
+                      <th
```
llun, line 444: **minor — two stored spellings that differ only in case become two identically-named, indistinguishable filter links**

`formatActivityTypeLabel` capitalises word-initial letters, which is case-insensitive, so `run` and `Run` both produce "Run" — as do `ride`/`Ride`, `walk`/`Walk`, `hike`/`Hike`, i.e. exactly the single-word Strava `sport_type` vocabulary. `getActivityPresentation` returns the same emoji for both too.

Both spellings are real and can coexist: the canonical form is applied on write, and `AGENTS.md` states plainly that "rows imported before this rule keep whichever one they came in with until `scripts/fitness/normalizeFitnessActivityTypes.ts` sweeps them" — a script that is per-actor and dry-run unless `--apply`. `getFitnessActivitySummary` groups on the raw column with no case-insensitive collation, so both survive as separate rows with separate totals.

Before this PR that was a cosmetic duplicate row of inert text. This PR is what makes it functional: the label is now a `<Link>` to `getActivityFilterHref(item.activityType, …)`, with `title`, `aria-current` and a matching chip. So the user gets two adjacent rows reading "🏃 Run" with different Count/Duration/Distance, linking to `?activity=run` and `?activity=Run`, and whichever they pick shows a chip reading "Run" that silently omits the other half of their runs. `aria-current` and the underline mark one of two visually identical rows.

**And the doc comment I added for this helper is wrong on its own example.** It justifies not using `getActivityPresentation(type).label` by saying "`ride` and `Ride` both read 'Cycling'" and that the fold "would print two identically named rows carrying different numbers and different filters" — which is precisely what its replacement does for that pair. The fold it genuinely prevents is `ride` vs `cycling`. That is a durable maintainer-facing defect independent of any instance's data.

---

app/(timeline)/fitness/RecentFitnessActivities.tsx
```diff
@@ -1,40 +1,112 @@
...
+  //
+  // A live region only announces a change to content it ALREADY held: one that
+  // mounts carrying its own text is silent, and so is one React re-creates
+  // because it moved in the tree. That is why the section below is always
+  // rendered and only its CONTENTS are gated — an actor whose activities have
+  // no surviving posts shows nothing here while unfiltered, yet still sees the
+  // Activities table (fed by the summary, not by these posts) and can still
+  // click a row. Returning early past this region made that first filter — the
+  // one transition where nothing else on the page moves either — the one it
+  // could not announce. The region is `sr-only`, so it is out of flow and an
```
llun, line 57: **minor — the live region announces "Showing recent &lt;label&gt; activities" even when the filter matched nothing**

Rounds 1 and 2 established this region as the only thing that tells assistive tech the scope changed, and round 2 went to some length to keep the node mounted so its text is *updated* rather than re-created. Neither round checked what that text actually says in the branch immediately below it.

It is derived from `activityType` alone and never reads `statuses.length`, so with a filter set and an empty list it announces that activities are being shown while the page renders "No recent &lt;label&gt; activities have been posted." That paragraph is a sibling of the region, not inside it, and is not itself a live region — and with focus unmoved and the page unscrolled, nothing carries the reader to it. The one mechanism this PR added to report the filter change reports the wrong outcome, in a branch the same PR added.

The state is reachable and not exotic: `getDistinctActivityTypesForActor` has no `statusId` predicate, so a type whose posts were all deleted is still an accepted `?activity=` value and still linked from the Activities table. It is also broader than the deleted-post case — `RECENT_LIMIT` is 5, so an actor with many posted runs lands here whenever the five most recent rows of that type happen to be postless, and `getStatus`'s `.catch(() => null)` can drop rows too.

---

app/(timeline)/fitness/ActorFitnessDashboard.test.tsx
```diff
@@ -18,6 +25,33 @@ vi.mock('@/lib/client', () => ({
...
 
+// next/link swallows `prefetch` and `scroll` instead of reflecting them in the
+// DOM, so the only way to assert on them is to render them ourselves. Neither
+// may be spread onto the `<a>`: they are not valid DOM attributes.
+vi.mock('next/link', () => ({
+  default: ({
+    children,
+    href,
+    prefetch,
+    scroll,
```
llun, line 36: **minor — both `next/link` mocks swallow `scroll`, so dropping `scroll={false}` fails no test**

Both mocks added by this PR destructure `scroll` into an unused `_scroll` and never reflect it (here, and `RecentFitnessActivities.test.tsx:18`) — even though this same mock deliberately reflects `prefetch` as `data-prefetch` on the next line, precisely so line 323 can assert it. The author's own convention is applied to one prop and not the other in the same mock.

`scroll={false}` is not decorative here; it is the premise the accessibility design rests on. `RecentFitnessActivities.tsx` argues that the filter navigation "swaps the list below and moves nothing else: `scroll={false}` keeps the page still … This is the one thing that tells them the scope changed", and the test file repeats that rationale verbatim while asserting only the live region. So the prop that makes the argument true is the one prop no test can see.

A verifier reproduced it: deleting `scroll={false}` from both links leaves the whole `app/(timeline)/fitness/` suite green — 16 files, 325 tests — and a repo-wide grep finds no other assertion on it. `AGENTS.md` documents this exact mock-and-reflect pattern for `prefetch`, so the fix is the established one.

(The `RecentFitnessActivities` mock also swallows `prefetch`, leaving the clear chip's `prefetch={false}` unasserted as well.)

---

app/(timeline)/fitness/ActorFitnessDashboard.tsx
llun: Confirmed and fixed in 74e7aa0 — this is the best catch of the three rounds, and the wrong comment was mine.

**The comment.** Corrected to use `ride` vs `cycling`, which is a fold this helper genuinely prevents, and it now states the limit rather than overclaiming: capitalising is itself case-insensitive, so `ride`/`Ride` still land on one label, and it points at the new helper for what makes the survivors distinguishable.

**The defect.** New `buildActivityTypeLabels(activityTypes)` in the same module returns one label per stored value, qualifying a colliding label with the value that produced it — and *only* a colliding one, so an instance whose labels are already unique renders exactly as before:

```
['run', 'Run', 'ride']  →  { run: 'Run (run)', Run: 'Run (Run)', ride: 'Ride' }
```

The dashboard builds it over the whole `topActivities` set rather than per row, which is what makes the collision visible at all.

Verified against a real mixed-case row rather than only in tests — I set one seeded `run` row to `Run` and loaded `/fitness`:

```
🏃  Run (run)    15   8h 17m   98.7 km
🏃  Run (Run)     1      37m    7.7 km
```

Two rows, their own totals, `?activity=run` and `?activity=Run`, and neither able to be mistaken for the other.

Test coverage: a dashboard test with a colliding summary asserting both links and their hrefs and that no bare "Run" link survives, a companion asserting unambiguous labels stay unqualified, and four cases on the helper itself. One of those started as a bad test of mine — I had asserted `RIDE` collides with `ride`, and it does not, because only word-initial letters are touched. That case now pins the opposite, so the qualifier cannot start appearing on labels nothing is ambiguous with. Mutation-tested: reverting to the per-row label fails the collision test.

Scope note, since it came up in verification: the fuller fix would also qualify the chip, which cannot detect a collision because it holds one activity type. I left it naming the sport plainly. The ambiguity this finding is about is between two *controls*, and that is resolved in the table where both are visible; the URL and the underlined row already identify which one is active.

---

app/(timeline)/fitness/RecentFitnessActivities.tsx
llun: Fixed in 74e7aa0. The region now reports the outcome rather than the requested scope:

```tsx
{statuses.length === 0
  ? label
    ? `No recent ${label} activities have been posted.`
    : 'No recent activities have been posted.'
  : label
    ? `Showing recent ${label} activities`
    : 'Showing all recent activities'}
```

Text content does not affect node identity, so round 2's keep-it-mounted property is untouched — the rerender identity test still passes unchanged.

Confirmed in the browser: `/fitness?activity=walk`, a type whose activities have no surviving posts, now announces "No recent Walk activities have been posted." where it previously said "Showing recent Walk activities".

New test asserts the filtered-empty branch announces the outcome and specifically does *not* contain "Showing"; the two existing empty-state assertions were updated to the new wording. Mutation-tested — restoring the scope-only text fails three tests.

One consequence worth naming: the announcement and the visible empty state now say the same sentence, so a screen-reader user browsing the page encounters it twice. That is the normal cost of the `role="status"` pattern, and the alternative — putting the role on the visible paragraph — would reintroduce exactly the mount/unmount problem round 2 fixed, since that paragraph does not exist when the list is non-empty. The existing "keeps the section rendered" test now filters the two matches by role, so it still asserts what a sighted reader sees.

---

app/(timeline)/fitness/ActorFitnessDashboard.test.tsx
llun: Fixed in 74e7aa0. Both mocks now reflect `scroll` the way one already reflected `prefetch`, and the `RecentFitnessActivities` mock reflects `prefetch` too:

```tsx
<a href={href} data-prefetch={String(prefetch)} data-scroll={String(scroll)} {...rest}>
```

Asserted in two places: the dashboard's "links each activity name to that type filter without prefetching it" test now pins `data-scroll="false"` beside the existing `data-prefetch` assertion, and the chip test pins both. Mutation-tested — deleting `scroll={false}` from the two links now fails those two tests, where before it failed none.

Since this was ultimately a claim about runtime behaviour, I also stopped taking `scroll={false}` on trust and measured it. The naive check looks like a failure: scrolled to y=398 on the unfiltered page, clicking a filter leaves y=32. But that is the browser clamping to a document that got shorter — 32 is exactly the new maximum scroll. Driving it the other way settles it, because a page that grows cannot clamp:

```
unfiltered → filtered:  before y=398 | after y=32, max=32   (clamped, not reset)
filtered → unfiltered:  before y=30  | after y=30, max=398  (preserved exactly)
```

A router scroll-to-top would have produced 0 in both. So the premise the comments rest on holds, and it is now guarded.

On the parenthetical: you're right that the chip's `prefetch={false}` isn't strictly required by the feed-row rule in `AGENTS.md` — a single chip is not a per-row link. I've pinned it anyway, since it points at the same `force-dynamic` page and the intent is deliberate either way.

---

llun:
## Sub-agent code review — round 4: clean

Fourth round at `74e7aa0`, with the bar raised explicitly ("after three rounds an empty findings array is a likely and entirely acceptable answer; do not invent something marginal"). Ten agents, no errors.

**Three findings raised, none confirmed.** Each split its two adversarial verifiers 1–1, so none met the unanimity bar. That ends the loop: a full round with no new actionable comments.

The two substantive ones were the same observation from different reviewers — that the clear chip, empty state and live region name the filter with the *un-disambiguated* label while the table uses `buildActivityTypeLabels`. That is a real asymmetry, and it is deliberate; I called it out in the round-3 thread before this round found it. The refutations landed on the same reasoning:

- The helper's contract is "disambiguated against the others it is rendered **beside**". The table renders a set; there is exactly one chip, and it only clears, so nothing is ambiguous with it.
- The selected row already carries `aria-current="true"` and the qualified label, so the active filter is never unidentifiable.
- Both suggested fixes make things worse: the page-side set is the **all-time** distinct list while the table's is the **date-ranged** summary, so qualifying the chip trades one mismatch for its mirror image — a chip reading "Run (Run)" above a table showing a single unqualified "Run" — and pushes a raw stored value into prose and an accessible name ("Clear Run (Run) filter").

The third asked for the labeling convention to be lifted from JSDoc into `AGENTS.md`. Refuted on the ground that the helper is a *relocation* of a previously private, entirely undocumented function into the canonical module that `AGENTS.md` already points at, with new JSDoc explaining the choice — the PR narrows that gap rather than widening it. Recording it here rather than acting on it; if you'd prefer an `AGENTS.md` bullet, it's a one-line follow-up.

### Loop summary

| Round | Agents | Raised | Confirmed | Outcome |
| 1 | 44 | 19 | 10 → **6 distinct** | fixed in `58ded3a` |
| 2 | 9 (rate-limited) | 2 | verifiers died | self-verified, fixed in `eed40f2` |
| 3 | 11 | 3 | **3** | fixed in `74e7aa0` |
| 4 | 10 | 3 | **0** | clean |

Ten issues found and fixed across the loop; one major (an assertion that could not fail), the rest minor. Every fix was mutation-tested — the defect it describes now fails a test — and the behavioural ones were verified in a browser, not only in jsdom.

`main` has since been merged in (`f10d587`), which brought `next` 16.3.1 → 16.3.2. Since two of this PR's guarantees are Next-behavioural, I re-verified both against the new version rather than assuming: the live region's DOM node still survives an RSC filter navigation with its text updated in place, and `scroll={false}` still preserves scroll position exactly (30 → 30) on the navigation where a growing page makes clamping impossible.

---

PR #1539, opened 2026-08-25
minor: show a click-to-play YouTube player on link preview cards

llun:
## Summary

A post linking a YouTube video now renders the video in place of the generic link card: a 16:9 poster with a play button that swaps in the embedded player when the reader presses it.

**Nothing is loaded from YouTube's player until that press.** A live iframe per feed row would pull a megabyte of player code and announce the reader to Google for every video that merely scrolled past, so the facade is what keeps a timeline cheap and quiet; pressing play is the consent that makes the request reasonable.

**The branch reads the URL, never the metadata.** `getYouTubeVideoFromUrl` (`lib/utils/youtube.ts`) parses `linkPreview.url` — the redirect-resolved final URL this server itself fetched — so a page can only be treated as YouTube by actually being served from a YouTube host. `og:type` and the stored `type` column are deliberately ignored even though `mapCardType` already writes `'video'` for these pages: that column is page-claimed, and gating on it would make one URL render two ways depending on what a cache entry happened to capture. Hosts are matched **exactly** and the id must be 11 base64url characters, so a look-alike host, a playlist, a channel or a percent-encoded segment all fall back to the ordinary card. `/embed/videoseries` is refused by name — it is eleven lowercase letters that embed a whole playlist.

**Two fixed hosts are the entire third-party surface.** The player is framed from `https://www.youtube-nocookie.com`, now the only origin in `frame-src` — there was no `frame-src` at all before this, and `default-src 'none'` blocks every iframe. The poster comes from `https://i.ytimg.com`, added to `img-src` **unconditionally** rather than through `remoteMediaSources`, so narrowing (or emptying) `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` cannot blank the thumbnail of every video card; `next.config.test.ts` pins that against the emptied allowlist.

**Two bugs found while verifying in a real browser, both fixed here.** Neither was visible to the test suite:

- **A thumbnail that failed before hydration was never dropped.** These cards are server-rendered, so the browser starts fetching while it parses the HTML and usually finishes first. `error` does not bubble, so React attaches that listener during hydration and a failure that landed first is never delivered — `onError` simply never fires. Every jsdom test passed because they fire the event by hand, which real life does not. Both cards now also ask the element what already happened when the ref attaches. The **standard** card's documented text-only fallback had been dead for the same reason, and its boolean reset effect ran *after* the ref and overwrote what it had just found — that state is now derived from the failing URL instead of reset in an effect.
- **Focus was dropped to `document.body`** when the play button unmounted and the iframe took its place, so a keyboard user lost their position with no signal anything had happened. Focus now moves onto the player, which also hands it their keys (space to pause) — the same move `StatusReplyBox` makes when its inline UI replaces its trigger.

The Mastodon API is untouched: `card.html` and `card.embed_url` stay empty because the embed URL is built in the browser, so this is not oEmbed consumption. Backend, database and migrations are untouched, and existing cached cards get the player with no refetch (the parser reads only `url`).

Not in scope, deliberately: other video providers, a settings toggle, and the trends page's own link card.

## Screenshots

Verified against a production build (`yarn build` + `yarn start`) on a local SQLite instance, in light and dark themes at 900px and 390px widths:

- **Before the click** — the post shows a full-width 16:9 poster with a centred play badge, and a caption below it carrying `youtube.com` and the video title, linking out to YouTube. A non-YouTube link in the same timeline still renders the ordinary compact card with its 88px thumbnail.
- **After the click** — the poster is replaced in place by the `youtube-nocookie.com` player at the same 16:9 size; the caption stays as the escape hatch to YouTube.
- **Keyboard** — the facade takes the standard `focus-visible` ring, Enter activates it, and `document.activeElement` is then the iframe (confirmed by assertion, not just by eye).

Observed in that run: `t=1m30s` became `start=90` in the player URL, a 404ing stored thumbnail fell back to `hqdefault.jpg` (`naturalWidth=1280`), and there were **zero CSP violations** in the console.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn typecheck`, `yarn build`, and `yarn test` all pass locally, run in that order — 869 test files, 11,010 tests
- [x] Docs updated: `AGENTS.md` (Link Preview Cards), `docs/features.md`, `docs/mastodon-api-compatibility.md`, and the `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` row in `docs/environment-variables.md`

One caveat for review: this environment only has Node 20/21/22, not the Node 24 `AGENTS.md` mandates, so the whole gate ran on **Node 22**. Worth a re-run on 24 before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VoTU8vVAKMscydCqxoFuMy)_


lib/components/posts/link-preview-card.tsx
```diff
@@ -121,3 +358,43 @@ export const LinkPreviewCard: FC<LinkPreviewCardProps> = ({
...
+    return (
+      // Keyed on the video so React CLEARS the card's state when the linked
+      // video changes, rather than the card comparing against state it kept.
+      // Deriving `isPlaying` from a remembered id defends A→B but not A→B→A:
+      // the remembered id was never cleared, so returning to a video the reader
+      // had played re-entered the playing branch with no click — mounting an
+      // autoplaying frame and taking focus. A remount starts from `null`, in
+      // the same commit, so there is still no painted frame of autoplay.
+      <YouTubeLinkPreviewCard
+        key={video.videoId}
```
llun, line 421: **MAJOR — play consent survived a video change in one direction** (fixed in c369bbca)

Two reviewers reproduced this independently, with a test rather than by reading.

Deriving `isPlaying` from a remembered `playingVideoId` defended A→B — the existing "returns to the facade when the post is edited to a different video" test covers exactly that hop — but the remembered id was **never cleared**, only compared against. So `render(A) → click → rerender(B) → rerender(A)` flipped `isPlaying` false→true with no gesture: a fresh `<iframe src=".../embed/A?…&autoplay=1">` mounted, sending an un-consented request to `youtube-nocookie.com`, and the focus effect fired and yanked focus into the player.

Reachable in production, not just in a harness: `post.tsx` renders `<LinkPreviewCard>` with no key, so an edit updates props on the *same mounted instance*. Two edits — one away, one back — is all it takes.

The fix keys the card on the video so React clears the state via remount, rather than the card comparing against state it kept. That preserves the property the id-keying was defending (a remount happens in the same commit, so there is still no painted frame of autoplay) and now holds in both directions.

Regression test added: `still asks for a click when the post is edited back to a video already played`. Mutation-verified — removing the `key` fails that test and only that test.

---

lib/components/posts/link-preview-card.tsx
```diff
@@ -24,6 +52,196 @@ const getDomain = (linkPreview: StatusLinkPreview) => {
...
+                ? `Play video: ${linkPreview.title}`
+                : 'Play video'
+            }
+            // `ring-inset` is load-bearing, not decoration: a ring is an
+            // OUTWARD box-shadow, and this button is flush with the edges of a
+            // wrapper that must clip (`overflow-hidden`, to round the video's
+            // square corners) — so an outward ring is clipped away on every
+            // flush side and the focus indicator all but disappears. Painting
+            // it inward keeps it inside the button's own box.
+            className="group absolute inset-0 flex cursor-pointer items-center justify-center focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/50 focus-visible:ring-inset"
```
llun: **MAJOR — focus indicators were almost entirely invisible** (fixed in c369bbca)

Found by compiling the project's real `tailwindcss@4.3.3` config against `app/globals.css` and rendering this markup in headless Chromium, rather than by inspection.

A Tailwind `ring` is a **non-inset** box-shadow — it paints outward from the element's box. Both focusable elements here sit flush against the edges of a wrapper that must clip (`overflow-hidden`, needed to round the video's square corners), so the outward ring was clipped away on every flush side. The play button kept only a ~2px sliver at the one edge that isn't flush; the caption `<a>` had no focus classes at all and fell back to a native outline that doesn't render in this nested layout.

Net effect: a sighted keyboard-only or low-vision user got no reliable indication of what was focused, on the primary interactive surface of the feature, in both themes.

Both rings now paint inward (`focus-visible:ring-inset`), so they stay inside each element's own box and cannot be clipped. Verified after the fix in a real browser — computed style on both elements is `… 0px 0px 0px 2px inset`, and the ring is visible on all four edges.

Note for future readers: the button's half of this mirrors an existing pattern in `attachments.tsx:36`, which has the same clipped-ring shape. Out of scope here, but worth a follow-up.

---

lib/components/posts/link-preview-card.tsx
```diff
@@ -24,6 +52,196 @@ const getDomain = (linkPreview: StatusLinkPreview) => {
...
+            ) : null}
+            {/* Decoration: the button's own label already says what this does,
+                and with no poster left to show this is the whole facade. */}
+            <span
+              aria-hidden="true"
+              // The hairline is what keeps the badge legible when NO poster
+              // loaded: the area behind it is then plain `bg-black`, and black
+              // at any opacity composites to black, so the circle and its hover
+              // state would both vanish and leave a bare floating triangle.
+              className="relative rounded-full bg-black/60 p-4 ring-1 ring-white/25 transition-colors group-hover:bg-black/80"
```
llun, line 224: **MINOR — the play badge disappeared against its own fallback background** (fixed in c369bbca)

When both poster candidates fail — the scenario the `keeps the play button when no poster loads at all` test exercises — the only backdrop is the media area's plain `bg-black`. Black composited over black is black at any opacity, so `bg-black/60` and its `group-hover:bg-black/80` were both indistinguishable from what sat behind them: a bare white triangle floating with no circle and no hover feedback.

Fixed with a `ring-1 ring-white/25` hairline, which reads in every case rather than only when a poster happens to load.

---

lib/components/posts/link-preview-card.tsx
```diff
@@ -24,6 +52,196 @@ const getDomain = (linkPreview: StatusLinkPreview) => {
...
+            // embedding to allowlisted domains. The poster below keeps
+            // `no-referrer` because it loads before the reader has consented to
+            // anything; this frame only exists because they pressed play.
+            // No sandbox either — but not because one would be useless. A
+            // sandbox is only neutered by `allow-scripts allow-same-origin`
+            // when the framed document is SAME-origin with us, which this is
+            // not, so one would still deny top-level navigation and popups.
+            // It is left off because the player's own affordances need popups
+            // ("Watch on YouTube", end-screen links), and the origin is already
+            // pinned twice over: by the builder and by the CSP frame-src.
```
llun, line 162: **Comments that stated wrong or imprecise reasoning** (fixed in c369bbca)

Three of my own comments justified correct decisions with incorrect arguments. That's worse than no comment, because the next reader inherits the argument and may act on it.

1. **The sandbox rationale was wrong** (this line). I wrote that the player "needs `allow-scripts` and `allow-same-origin`, which together neutralize it". That neutering only applies when the framed document is **same-origin with the embedder** — `youtube-nocookie.com` is not, so a sandbox *would* still deny top-level navigation and popups. The conclusion (no sandbox) stands, but for a different reason: the player's own affordances need popups, and the origin is already pinned twice, by the builder and by CSP `frame-src`.

2. **The poster-chain termination argument was wrong** (~line 99). I credited the identical-array guard in `markPosterFailed`, but `posterUrl` is by construction a URL the failed list does not yet contain, so that branch never runs on the advance path. What actually terminates the chain is that there are at most two candidates and each failure advances past one. The guard still earns its place — for a late `error` from an element already replaced.

3. **`hasImageAlreadyFailed` has a known false positive** (~line 33). `complete && naturalWidth === 0` also describes a viewBox-only SVG that loaded and painted fine. There is no cleaner DOM question to ask and the degradation is graceful, so it's accepted — but now documented rather than silently assumed away.

Also noted at the standard card's ``: it needs no `key` only because a recorded failure *unmounts* it, so no broken element survives to be handed a new `src`. Anything that keeps it mounted through a failure has to bring the key along.

---

docs/features.md
```diff
@@ -49,7 +49,7 @@ This document tracks the implemented and planned features for Activity.next.
...
...h the page's title, description, publisher and thumbnail, the way Mastodon, Bluesky and other social apps do. The server fetches each page once through the SSRF-guarded outbound fetcher, caches the result per URL (shared by every post linking the same page, refreshed weekly, failures suppressed for an hour), and serves it as the Mastodon `Status.card` so third-party clients render it too. Cards are fetched for remote posts as well as local ones, with a random delay so a widely-shared link is not hit by every receiving server at once. The card yields to media, a quoted post or a fitness activity when a post has one. A link to a **YouTube video** shows the video itself instead of the usual card: a thumbnail with a play button that swaps in the embedded player when you press it, so none of the player — around a megabyte of code, and the tracking that comes with it — is loaded until you choose to watch. The thumbnail itself is still fetched from YouTube's image host up front, without a referrer, the same way any other card's thumbnail is. The player is always framed from the cookie-light `youtube-nocookie.com` host. Admins can turn fetching off under **Admin → Network → Link previews**
```
llun, line 52: **MINOR — the privacy claim was overstated** (fixed in c369bbca)

The sentence originally read "…so nothing is loaded from YouTube's player — **and no request identifies you to it** — until you choose to watch."

That second clause promises more than the code delivers. The poster thumbnail is fetched from `i.ytimg.com` on render, unconditionally — the component's own comment says as much ("the poster keeps `no-referrer` because it loads before the reader has consented to anything"). `no-referrer` withholds *which page* you were reading, but the request still reaches a Google-owned host carrying your IP.

The PR description and the AGENTS.md bullet both scope this correctly to *the player*; this one sentence didn't. Now rewritten to say what actually happens: the player and its tracking are gated behind the click, the thumbnail is fetched up front without a referrer like any other card's.

Worth stating plainly for reviewers weighing the `img-src` decision below: for a **default-configured** instance this is not a new request — a YouTube card's `og:image` already pointed at `i.ytimg.com` before this PR. It is only new for an operator who had set `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS='[]'`.

---

lib/components/posts/link-preview-card.test.tsx
```diff
@@ -172,4 +197,306 @@ describe('LinkPreviewCard', () => {
...
+      ).toBe('1')
+      expect(iframe?.getAttribute('allow')).toContain('autoplay')
+      expect(iframe).toHaveAttribute('allowfullscreen')
+    })
+
+    // The player deliberately does NOT carry the posters' `no-referrer`: some
+    // videos are embeddable only from allowlisted domains, and withholding the
+    // origin breaks them. It sits two lines from two images that DO carry it,
+    // so pin the absence — otherwise harmonising the three reads as tidying up.
+    it('sends the instance origin to the player rather than no referrer', () => {
```
llun, line 280: **MINOR — nothing pinned the player's deliberate lack of `no-referrer`** (fixed in c369bbca)

The reviewer added `referrerPolicy="no-referrer"` to the iframe and reran the whole suite: **106/106 still passed**. Nothing asserted the attribute's presence *or* absence.

That matters because the omission is deliberate and load-bearing — videos whose owners restrict embedding to allowlisted domains break without an origin — and the iframe sits three lines from two `` elements that both explicitly carry `no-referrer`. That adjacency is exactly what a future contributor "harmonizes" in good faith, silently breaking restricted embeds with no test to catch it.

This test pins the absence, with the reason in a comment so the next person reads it before changing it.

---

AGENTS.md
```diff
@@ -1359,6 +1359,41 @@ it; there is no legacy shape left to copy.
...
+  — never the metadata.** `getYouTubeVideoFromUrl` (`lib/utils/youtube.ts`)
+  parses `linkPreview.url`, which is the **redirect-resolved final URL this
+  server fetched** (`fetchLinkPreview` stores `normalizePreviewUrl(response.url)`),
+  so a page can only be treated as YouTube by actually being served from a
+  YouTube host. Neither `og:type` nor the stored `type` column is consulted,
+  even though `mapCardType` already writes `'video'` for these pages: that
+  column is page-claimed, and gating on it would make one URL render two ways
+  depending on what a cache entry happened to capture. Hosts are matched
+  **exactly** (`youtube.com.evil.example` ends with nothing that matters) and
+  the id must be 11 base64url characters — which rejects a percent-encoded PATH
```
llun, line 1374: **NIT — two imprecise claims in the documentation bullet** (fixed in c369bbca)

The security reviewer checked each documented claim against the code and found two that were slightly broader than the truth:

1. "the id must be 11 base64url characters, which also rejects a percent-encoded segment" — true for a **path** segment, since `pathname` is never decoded. But a `?v=` value *is* decoded by `URLSearchParams` before validation, so `?v=%41AAAAAAAAAA` resolves to the id `AAAAAAAAAAA`. Harmless (the decoded value still has to satisfy the same 11-character shape), but the sentence read as if decoding never happens anywhere. Now split explicitly.

2. The `videoseries` carve-out read as broader than it is — it is path-scoped, so `youtu.be/videoseries` isn't covered by it. Also harmless, and for a better reason than the carve-out itself: `list` is **never carried into the embed URL**, so no playlist can be framed however it is spelled. That is now stated, since it's the property that actually holds the line.

Everything else in the bullet was verified accurate against the code: the redirect-resolved `url`, `mapCardType` writing `'video'` while the branch ignores it, exact `Set` host matching, `videoseries` being exactly eleven lowercase letters, `frame-src` genuinely not existing on `main`, and the poster host being unconditional rather than part of `remoteMediaSources`.

---

llun (commented):

Four reviewers over the full diff: **React lifecycle** and **security** on Opus, **tests/conventions** and **UI/a11y** on Sonnet. All findings below are already fixed and pushed in c369bbca; the inline comments record what was found and why, per AGENTS.md → Code Review Loop.

What makes this round worth reading is that three of the four verified empirically instead of reasoning from the source: the React reviewer scripted `render → click → rerender(B) → rerender(A)` against the real component, the UI reviewer compiled the project's actual `tailwindcss@4.3.3` config and rendered the markup in headless Chromium, and the tests reviewer mutation-tested the suite by neutering the implementation and checking that exactly the right tests failed.

**2 MAJOR** — both real, both would have shipped:
- Play consent survived a video change in one direction (A→B→A re-autoplayed with no click). Found independently by two reviewers.
- Focus rings were clipped to near-invisibility on both interactive elements.

**3 MINOR** — the play badge vanishing against its own fallback background; nothing pinning the player's deliberate lack of `no-referrer`; and an overstated privacy sentence in `docs/features.md`.

**Nits** — three code comments whose stated reasoning was wrong or imprecise (fixed, since a wrong rationale outlives the code it explains), two doc imprecisions, an untested host, and three copies of a CSP test helper reduced to one.

**Security: clean.** The reviewer ran a live battery against the real parser — host confusion, userinfo, backslash authority termination, IDN/punycode/soft-hyphen, percent- and double-encoding, Unicode digits, ReDoS, and hand-built `YouTubeVideo` objects bypassing the parser entirely. Nothing reaches an embed or escapes the two fixed origins. It also independently confirmed the premise the design rests on: no path lets a remote server supply `linkPreview.url`, because `parseOpenGraphMetadata` returns no `url` key and only `completed` rows hydrate.

One judgement call left standing deliberately, flagged rather than silently reversed: the unconditional `i.ytimg.com` in `img-src` overrides an explicit `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS='[]'`. For a default instance it adds no request that wasn't already there; for that one configuration it does. It is documented and test-pinned, and reverting it would blank the thumbnail on every video card.

Two follow-ups noted and deliberately **not** done here, to avoid widening the PR: `attachments.tsx:36` has the same clipped-ring shape, and `fetchLinkPreview.ts` spreads `...metadata` after `url: resolvedUrl` (inert today, since `LinkPreviewMetadata` has no `url` key, but one `og:url` feature away from mattering).

Gate after the fixes: prettier, lint, typecheck, build, and **11,013 tests** all green; focus rings re-verified in a browser by computed style.

---

lib/components/posts/link-preview-card.tsx
```diff
@@ -24,6 +52,214 @@ const getDomain = (linkPreview: StatusLinkPreview) => {
...
+            >
+              <Play className="size-8 fill-white text-white" />
+            </span>
+            {/* The focus indicator, as an overlay rather than a style on the
+                button, because this is the only place it survives: painted
+                LAST, after the poster that covers the button's box, and inside
+                its own bounds so the wrapper's clip cannot take it. */}
+            <span
+              aria-hidden="true"
+              className="pointer-events-none absolute inset-0 ring-inset ring-ring/50 group-focus-visible:ring-2"
```
llun: **HIGH — round 1's focus fix did not work, and I reported that it did** (fixed in b71d24f9)

This is the finding that justifies running a second round over the first round's own fixes.

Round 1 replaced a clipped outward ring with `focus-visible:ring-inset`, and I verified it by reading `getComputedStyle` (which returned `… 2px inset`) and glancing at a screenshot where I couldn't see the ring — and talked myself out of it as "subtle against a dark poster". **A box-shadow that is computed is not a box-shadow that is painted.** An inset box-shadow paints above the element's own background but *below its content and descendants*, and the poster here is an `absolute inset-0 size-full` child covering the button's box exactly. So with a thumbnail loaded — the ordinary case — the indicator was behind it.

Worse, it was a net regression: the pre-round-1 outward ring at least left a 2px seam along the bottom edge, painted over the caption below. After round 1 there was nothing.

Measured this time instead of asserted, by counting indicator pixels on the focused element with a poster loaded:

| indicator on the button | indicator pixels |
|---|---|
| outward ring (original) | clipped on 3 sides |
| `ring-inset` (round 1) | **0** |
| `outline` + negative offset | **0** |
| overlay span, painted last (this fix) | **4368**, spanning the full height |

An outline was the obvious next try and also came out at zero. What works is putting the indicator on its own `pointer-events-none absolute inset-0` overlay rendered *after* the poster, so it paints last and sits inside its own bounds where the wrapper's clip can't reach it.

The caption link is genuinely different and keeps its outline — its children are in-flow text that never covers the padding edge, and it measures 3945 pixels on all four edges. It also gained `rounded-b-[11px]` so its corners aren't nipped by the wrapper's `rounded-xl` clip.

I also caught a measurement bug of my own along the way: Playwright's `page.screenshot({clip})` is page-relative while `getBoundingClientRect()` is viewport-relative, so my first pixel scans were sampling a different card entirely. The numbers above come from element screenshots.

---

AGENTS.md
```diff
@@ -1359,6 +1359,58 @@ it; there is no legacy shape left to copy.
...
+  `csp.ts`: the player is framed from `https://www.youtube-nocookie.com`, the
+  only origin in `frame-src` (there was no `frame-src` at all before this —
+  `default-src 'none'` blocks every iframe), and the poster comes from
+  `https://i.ytimg.com`, which is an **unconditional** `img-src` source rather
+  than one of `remoteMediaSources`, so narrowing (or emptying)
+  `ACTIVITIES_ALLOW_REMOTE_MEDIA_DOMAINS` cannot blank the thumbnail of every
+  video card. `next.config.test.ts` pins that against the emptied allowlist.
+  **Nothing loads from the player until the reader presses play** — a live
+  iframe per row is a megabyte of player code and a Google request for every
+  video that merely scrolled past — and **the card is mounted under a React
```
llun, line 1391: **MEDIUM — the docs still credited the guard that round 1 disproved, and never mentioned the `key` doing the work** (fixed in b71d24f9)

After round 1, `AGENTS.md` still read: *"the play state is keyed on the video id, not a boolean, so a card replaced in place cannot inherit the previous video's consent."* That is precisely the claim round 1 falsified — it holds for A→B and not for A→B→A — and the round-1 commit edited this file for other reasons without touching it. The component's own comment said the same.

The reviewer proved the `key` is now the sole load-bearing mechanism: with `key={video.videoId}` present, replacing the id-keyed state with a plain `useState(false)` boolean passes **all 36 tests**, including both regression tests. Once the component only ever sees one `videoId` for its lifetime, `playingVideoId === video.videoId` is just `playingVideoId !== null`.

The concrete hazard is a future reader — or this repo's own `/simplify` pass — deleting the `key` as redundant *on the documentation's authority*, and restoring the un-consented-autoplay bug. The A→B→A test would catch it, so this is a wasted round rather than a shipped bug, but the docs shouldn't be the thing that causes it.

Both now name the remount as the mechanism and the id-keying as defence in depth (kept so that removing the key degrades to the old one-directional guard rather than to none). `AGENTS.md` also now carries the paint-order trap, including the instruction to verify a focus change by screenshotting the focused element with a poster loaded rather than by reading `getComputedStyle` — the exact mistake that cost this PR a round.

---

llun (commented):
## Sub-agent code review — round 2 (reviewing round 1's fixes)

One reviewer on Opus, pointed specifically at the round-1 fixes, since that is where a review loop most often introduces a new defect. It did, and this round paid for itself:

- **HIGH** — round 1's focus fix didn't work at all, and I'd reported it as working off `getComputedStyle` plus a screenshot I misread. Zero indicator pixels with a poster loaded. Now fixed and **measured**: 4368 pixels focused, 0 unfocused.
- **MEDIUM** — `AGENTS.md` and the component still credited the guard round 1 disproved, and never mentioned the `key` that replaced it — an invitation for a later pass to delete the fix as redundant.
- **LOW** — a comment claiming the native outline "does not render at all" (it renders, and is clipped).

Everything else the reviewer interrogated came back sound, with the reasoning checked rather than assumed: the `key` causes no unwanted remount (a title-only or `?t=`-only edit keeps the same key), removing `w-full` changes nothing under `box-sizing: border-box`, `getSourcesFrom` is exactly equivalent to the three helpers it replaced, and both new tests were mutation-verified — removing the `key` fails *only* the A→B→A test while the A→B test still passes, which is exactly the asymmetry the fix targets.

Worth recording what made this catchable, since it's the transferable lesson: round 1 verified a paint change by reading a computed style. `ring-inset` compiles correctly and computes correctly — it just paints below a full-bleed child. The same is true of the `outline` I reached for next, which also measured zero. Only counting pixels on the focused element found it, and only with a poster loaded; the no-poster state (which is what round 1's other fix, the play badge, was diagnosed in) shows the ring fine. `AGENTS.md` now carries that trap so it isn't re-derived.

Gate after these fixes: prettier, lint, typecheck, build, **11,013 tests** green. Round 3 is running now over the round-2 fixes; the loop closes when a round comes back with nothing.

---

lib/components/posts/link-preview-card.tsx
```diff
@@ -24,6 +52,224 @@ const getDomain = (linkPreview: StatusLinkPreview) => {
...
+                span is the button's LAST child and has no children of its own,
+                so an outline on it paints over the poster — and an outline,
+                not a ring, because forced-colors mode (Windows High Contrast)
+                drops box-shadows entirely and would leave the card's only
+                control with no focus indicator at all. The radius matches the
+                wrapper's inner corner so the top two corners are not nipped by
+                its clip. */}
+            <span
+              aria-hidden="true"
+              className="pointer-events-none absolute inset-0 rounded-t-[11px] group-focus-visible:outline-2 group-focus-visible:-outline-offset-2 group-focus-visible:outline-ring/50"
```
llun, line 243: **MEDIUM — no focus indicator at all in forced-colors / Windows High Contrast (WCAG 2.4.7)** (fixed in 2bf3f8bb)

Round 2 put the indicator on an overlay as a `ring` — i.e. a `box-shadow`. Forced-colors mode drops box-shadows entirely, and Tailwind v4's `outline-none` (on the button) is a plain `outline-style: none` with no forced-colors fallback — that's `outline-hidden`'s job in v4. So in High Contrast the card's only real control had **zero** indicator pixels while the caption link beside it kept 3516. The working technique was sitting one element away from the broken one.

Fixed by making the overlay's indicator an **outline** rather than a ring. It can be one *there* and not on the button, and that asymmetry is the crux: the overlay has no descendants of its own, whereas the button's full-bleed poster covers its box and paints over anything the button itself carries.

Re-measured on the focused element with a poster loaded, counting pixels that change on focus:

| | button | caption |
|---|---|---|
| `forced-colors: none` | **4409** | 3065 |
| `forced-colors: active` | **4412** | 3066 |

The overlay also gained `rounded-t-[11px]` (the wrapper's 12px radius less its 1px border), so the indicator turns the top two corners instead of stopping ~7px short of each — the same nipping round 2 fixed on the caption and left here. Corner indicator pixels went 29 → 45.

The reviewer confirmed the rest of the overlay approach independently: `pointer-events-none` holds (`elementFromPoint` inside the media corner returns the poster; mouse, Enter and Space all still mount the player), `group-focus-visible:ring-2` could only ever fire from the button since it's the only `.group` ancestor, and `aria-hidden` on the decorative span has no screen-reader or find-in-page impact.

One latent fragility noted and deliberately left: `group` is unnamed and shared with the badge's `group-hover:`. A future focusable ancestor carrying a bare `group` would light this overlay spuriously; `group/facade` would pin it. Not a defect today, and not worth the churn mid-review.

---

AGENTS.md
```diff
@@ -1359,6 +1359,72 @@ it; there is no legacy shape left to copy.
...
+  inherit the previous video's consent and autoplay something nobody clicked.
+  Do not delete that key as redundant: the component also remembers which video
+  was played, but that alone was the original guard and it was insufficient —
+  the remembered id was only ever compared against, never cleared, so it
+  defended a change A→B and not a change back A→B→A, which re-entered the
+  playing branch with no gesture. A remount clears it in both directions, and
+  happens in the same commit, so there is still no painted frame of autoplay.
+  `autoplay` is set only on a player the reader mounted, and must be in the
+  iframe's `allow` list to reach the frame at all.
+  **The facade's focus indicator lives on a child OVERLAY, not on the button**,
```
llun, line 1402: **HIGH — this paragraph documented the opposite of what shipped, and would have instructed the next agent to reintroduce the bug** (fixed in 2bf3f8bb)

The worst finding of the three rounds, and entirely self-inflicted.

I wrote that paragraph while pursuing an outline *on the button*, then measured that at zero, moved to the overlay — and never went back to the doc. So `AGENTS.md` said "**The focus indicator on the facade is an `outline`, never a `ring`**", blamed `ring-inset` for painting below descendants, and never mentioned the overlay at all. Meanwhile the code had `focus-visible:outline-none` on the button and a `ring-inset … group-focus-visible:ring-2` overlay — a ring, using the exact class the paragraph condemned. It also contradicted my own commit message.

The failure scenario is specific and likely: an agent trims the "redundant" overlay and moves to `focus-visible:outline-2 focus-visible:-outline-offset-2` on the button, *because the documentation says that is what the facade uses*. That is round 1's zero-pixel indicator, restored on the authority of the file whose whole purpose is to prevent it. The reviewer measured that configuration at **34 px** (invisible) with the overlay removed.

Rewritten to describe the real mechanism — a last-child overlay with no descendants of its own, why nothing on the button itself survives (clipped when painted outward, hidden by the poster when painted inward), why it carries an outline rather than a ring (forced-colors), and why the caption may keep its outline on the element itself.

I also folded in the two harness traps that cost this feature time, so they aren't re-derived: Playwright's `page.screenshot({clip})` is page-relative while `getBoundingClientRect()` is viewport-relative, and `:focus-visible` won't match a programmatic `.focus()` without keyboard modality first. Plus the standing instruction — verify a focus change by **counting indicator pixels** on an element screenshot, with a poster loaded and with `forced-colors: active`, never by reading `getComputedStyle`.

---

llun (commented):
## Sub-agent code review — round 3 (reviewing round 2's fixes)

One Opus reviewer over round 2's focus-indicator rework, re-measuring everything from a fresh build rather than trusting the numbers in my commit message. Five findings; the two that matter are fixed in 2bf3f8bb.

- **HIGH** — `AGENTS.md` documented the opposite of the shipped code and would have walked the next agent straight back into round 1's zero-pixel indicator.
- **MEDIUM** — no focus indicator at all under forced-colors / Windows High Contrast: the overlay's ring is a box-shadow, which that mode drops. Zero pixels on the button, 3516 on the caption beside it.
- **LOW ×3** — the overlay's top corners nipped by the wrapper's radius (the same defect round 2 fixed on the caption and left here), plus two stale comments: the caption claiming it used the "same offset outline as the button" (the button has none, deliberately), and a test still crediting the id-keyed play state for a case the `key` now handles.

All five collapse into essentially one change — the overlay carries an **outline** instead of a ring, plus the matching inner radius — which fixes High Contrast, makes the documentation true as written, unifies the two mechanisms, and turns the corners.

Verified, not asserted:

| | button | caption |
|---|---|---|
| `forced-colors: none` | 4409 px | 3065 px |
| `forced-colors: active` | 4412 px | 3066 px |
| corner pixels (TL/TR) | 29 → 45 | — |

The reviewer independently confirmed the parts of round 2 that were sound: `pointer-events-none` genuinely doesn't intercept the click (mouse, Enter and Space all mount the player), `group-focus-visible` can only fire from the button, `aria-hidden` on the decorative overlay is right, and `rounded-b-[11px]` is arithmetically correct against the compiled `--radius-xl` of 12px less the 1px border.

Three rounds, three real defects, each one found in the *previous* round's fix — the loop is doing exactly what it's for. Round 4 is running over these changes; the loop closes when a round returns nothing.

Gate: prettier, lint, typecheck, build, **11,013 tests** green.

---

AGENTS.md
```diff
@@ -1359,6 +1359,81 @@ it; there is no legacy shape left to copy.
...
+  and that is not a style preference — it is the only place it survives. The
+  button is flush with the edges of a wrapper that must clip
+  (`overflow-hidden`, to round the video's corners), so anything painted
+  outward (a plain `ring`, a zero-offset outline) is clipped on the three sides
```
llun, line 1406: **LOW — an outward indicator is not "clipped away", and the paragraph's own prescribed measurement is what produced that wrong number** (fixed in e4b0900d)

The only finding of round 4, and the more interesting half is the methodology.

The paragraph claimed an outward indicator on the button is "clipped away" and that both outward and inward variants "measured at essentially zero visible pixels". The inward half is true — 14 px, hidden beneath the poster. The outward half isn't: only **three** of the button's four edges are flush with the wrapper. Its bottom edge is interior, so an outward ring survives there as a full-width 2px line between the video and the caption. Measured on the wrapper: **1416 px**, against the shipped overlay's 4412.

That still argues for the overlay — a line on one edge is not a usable indicator — but "unusable" and "invisible" are different claims, and only one of them is true.

**Why it got recorded wrong is the part worth keeping.** The same paragraph told the next agent to count pixels on *an element screenshot of the focused element*. That method structurally cannot see anything painted outside the element's own box, so it reports zero for every outward candidate — including ones painting perfectly well. Round 1's commit message actually had the precise version ("clipped away on every flush side"); the qualifier was lost when it moved into AGENTS.md, and the measurement recipe then made the loss self-confirming. Anyone re-testing an outward variant on those instructions would reproduce the same false zero and draw the same wrong conclusion.

Now corrected on both counts: the real numbers, and an instruction to screenshot the **wrapper** whenever the candidate paints outward. Same fix applied to the code comment carrying the claim in the same words.

No behaviour change — the shipped indicator is untouched.

---

llun (commented):
## Sub-agent code review — round 4: **no code defects, loop closed**

Round 4 reviewed round 3's fixes and then re-read the whole feature with fresh eyes, explicitly instructed that an empty finding list was the expected answer and not to manufacture findings. It found **no code defects**. One low-severity documentation inaccuracy, fixed in e4b0900d with no behaviour change.

The shipped indicator was re-measured independently in all four states — the earlier rounds only ever checked two:

| state | play button | caption |
|---|---|---|
| `forced-colors: none`, poster loaded | 4412 | 3065 |
| `forced-colors: none`, no poster | 4411 | — |
| `forced-colors: active`, poster loaded | 4415 | 3066 |
| `forced-colors: active`, no poster | 4416 | — |

4412 is *exactly* the full 2px band of a 708×399 box (708·399 − 704·395), so the indicator is complete on all four sides in every state. The counterfactual confirms round 3's reasoning: the round-2 ring measures 4412 normal / **0** forced-colors.

One risk the reviewer checked that nobody had thought to: `outline-2` compiles to `outline-style: var(--tw-outline-style)`, and the button sets that property to `none`. Had Tailwind registered it as inheriting, the shipped indicator would have been dead in exactly the focused state. The compiled CSS registers it `inherits: false`, so it doesn't leak — but that was a live way for this fix to have been silently broken.

Also verified end to end, none of it previously covered: the playing state leaves no dangling overlay and Tab from the iframe reaches the caption; the radius geometry is genuinely concentric (outer arc radius 11, inner 9, meeting the straight edges exactly at 11px); `fetchLinkPreview` is the **only** writer of `link_previews.url`, so the "redirect-resolved URL this server fetched" premise the whole design rests on actually holds; the served CSP header is correct and the iframe mounts with no violation; and the `hqdefault` fallback is pixel-exact under `object-cover` — 480×360 letterboxed 16:9, cropped 9.375% per side, precisely the bar height, so no black bands.

**Four rounds: three real defects, each found in the previous round's fix, plus this documentation correction.** The loop is closed — a round finally returned nothing actionable about the code.

Two follow-ups remain deliberately out of scope, both recorded earlier: `attachments.tsx:36` carries the same clipped-ring shape this PR diagnosed, and `fetchLinkPreview.ts` spreads `...metadata` after `url: resolvedUrl` (inert today, one `og:url` feature away from mattering).

Final gate: prettier, lint, typecheck, build, **11,013 tests** green, CI green on every pushed commit.

---

PR #1540, opened 2026-08-25
fix: lay post media out as a scrollable row and left-align a lone attachment

llun:
## Summary

Post attachments rendered as a 2-column grid capped at four items with a `+N` overlay, and a single attachment was cropped to a full-width 16:9 box. This ports the design system's media strip (`ui_kits/web/Post.jsx` → `Attachments`) into `lib/components/posts/attachments.tsx`, which owns media for every surface that renders a post.

**A lone picture** keeps its own aspect ratio and hugs the post's left edge. It is scaled by **width** — `min(100%, round(SINGLE_MAX_HEIGHT * ratio)px)`, capped at the file's own pixels so a thumbnail is never upscaled — never by capping the height of an `aspect-ratio` box, which leaves the ratio to be re-derived from a clamped axis and resolves inconsistently across browsers.

**Two or more** become one fixed-height horizontally scrolling row, each item as wide as its own ratio makes it. Four details are load-bearing:

- `STRIP_ITEM_MAX_WIDTH` (78%) means no item can fill the strip, so the next one always peeks past the edge. That peek is what says "this scrolls" on a touch screen, where the back chevron never appears at all.
- `flex-none`, so the items overflow instead of shrinking to fit. It is the single declaration whose removal silently turns the whole feature off: measured in Chromium, three 800×600 photos in a 600px column lay out at 320px each with it (`scrollWidth` 972 vs `clientWidth` 600) and at 196px each without, making `scrollWidth === clientWidth` forever — no chevrons, no fade, no peek, no scrolling, and every photo squeezed and cropped.
- `scroll-snap-type: x proximity`, never `mandatory`: mandatory snapping pulls the peeking item flush as soon as the scroll settles and destroys the affordance the 78% cap creates.
- The forward chevron stays visible while there is more to the right; the back chevron appears only on hover, and is `pointer-events-none` while `opacity-0` — `opacity-0` alone still hit-tests, and `group-hover` never latches on touch, leaving a dead column over the leftmost photo.

**The four-item cap and the `+N` overlay are gone**, because scrolling now reaches everything. Anything the strip renders unboundedly therefore needs a deferral: strip images pass `loading="lazy"` (a photo dump was 4 requests, now one per photo), and `Media` turns that into `preload="none"` for a video — but **only one carrying a `poster`**. A posterless video's sole pre-playback frame comes from the `#t=0.01` source fragment, which needs metadata to decode, and the strip hides its controls, so deferring one would leave a bare empty box; federated video always lands there, since `thumbnailUrl` is written on the local-upload path alone. A lone picture or video stays eager, being the post's largest element.

### Decisions that deviate from the design kit, deliberately

- **The edge fade is a `mask-image`, not a background gradient.** Posts render on four surfaces — `bg-card` when framed, `bg-background` on the status detail, `bg-muted/30` for an ancestor row, and the page itself when unframed — so a fade painted in any one token is visibly wrong on the other three, and the kit's literal white is wrong in dark mode everywhere. A mask fades the strip's own pixels and lets whatever is behind show through. It lives in `buildEdgeFadeMask` so the string itself is unit-testable; jsdom's CSS parser drops the two variants carrying `calc()`, but the left-edge-only form round-trips, so a rendered node can be asserted against it.
- **Boxes are container-sized from the attachment's dimensions rather than letting the image size itself.** The kit has no blurhash; this app does, and self-sizing images would leave the placeholder nothing to paint into and the strip measuring zero before load.

### Consequences of filtering, handled

`isVisualAttachment` selects image and video; `isAudibleAttachment` renders audio as left-aligned players below the strip; anything else — a fitness `.fit` file, a PDF — is skipped rather than rendering the empty box the grid gave it. That filtering has three knock-on effects, all addressed:

- The **lightbox is handed exactly the pictures on screen**, indexed into that list. Passing the raw array gave `MediasModal` a blank slide, an empty thumbnail and a wrong "n of m".
- **`post.tsx`'s link-preview suppression** asks `isRenderableAttachment` instead of reading `attachments.length`, so a post carrying only a PDF still gets its link card.
- **A stored dimension of `0` means "unknown", not "zero pixels."** Several media-storage paths persist `metaData.width ?? 0`, so all reads go through one `getMediaGeometry` guard; reading `attachment.width` raw collapsed the box to `0×0` and the photo vanished. It also clamps pathological shapes (a 10×10000 sliver otherwise rounds an axis to zero) and falls back to a 4:3 box.

### Also in this PR

- Defines the `no-scrollbar` utility in `app/globals.css`. It had been applied in the emoji and reaction pickers while defined **nowhere** — Tailwind v4 has no such built-in — so it silently did nothing. Defining it for the strip would have taken away those rows' only overflow cue (measured: 8 tabs fit at exactly 286/286; a 9th, present on instances with custom emoji, overflows by 34px with zero peek), so both were switched back to a plain scrollbar rather than quietly inheriting it.
- Every picture button now carries an explicit `aria-label`; federation writes `attachment.name || ''`, so an undescribed photo announced as a bare "button".
- Keeping the chevrons out of the tab order takes **both** `tabIndex={-1}` and a mousedown guard. `tabindex="-1"` removes an element from the *sequential* order only, and Chrome and Firefox focus a button on click, so without `preventFocusOnPress` a chevron unmounted by its own scroll dropped focus to `<body>`.
- A strip item's focus indicator is an **`outline` with a negative offset**: an outset ring is clipped by the strip's `overflow-x-auto`, and an inset ring is invisible because an inset `box-shadow` paints beneath the opaque image.
- `MessageBubble`'s local `isVisualMedia` is replaced by the shared predicate.
- `scripts/mock/createMockStatuses.ts` can now seed single-image statuses and mixed aspect ratios — the layouts this PR changes were not reachable from the sanctioned local seeder at all.

Scope was confirmed with the author as **status attachments only**: DM bubbles and the fitness Photos tab are deliberately untouched.

## Screenshots

Verified in Chromium against a local SQLite instance, seeded with images labelled with their own dimensions so the per-item widths are readable. Rendered screenshots (desktop, mobile, strip scrolled mid-way, dark mode) were captured and shared with the author; the measured geometry behind them:

| Case | Before | After |
| --- | --- | --- |
| Single landscape 800×600 | full-width, cropped to 16:9 | `560×420`, flush left (`leftOffset: 0`) |
| Single portrait 600×900 | full-width, cropped to 16:9 | `280×420`, flush left |
| 3 images (800×600, 600×900, 1200×500) | 2×2 grid of squares | one 240px row, items `320 / 160 / 576`, `scrollWidth 1068` in an `822` box |
| 7 images | 4 shown + `+N` overlay | all 7 in the strip |

Also confirmed in the browser: strip images carry `loading="lazy"` and a lone picture does not; both chevrons report `tabindex="-1"`; the strip's `aria-label` reads `7 media attachments, scroll for more` only when it actually overflows; clicking the second strip item opens the lightbox on the second picture. In dark mode the mask fades to the dark surface rather than to white.

## Review loop

Five rounds of the AGENTS.md sub-agent review loop ran on this diff, each with two lanes (a code/regression read and adversarial mutation testing, the latter in a disposable worktree) and an independent Opus verifier per lane tasked with *refuting* its lane's report.

**26 findings across rounds 1–4, all fixed and mutation-verified; round 5 was clean on both lanes and both verifiers, which ended the loop.** Roughly 14 further candidates were refuted by the verify passes. All 23 review threads are resolved.

Three findings are worth calling out because a fix of mine caused two of them:

- **`ring-inset` was worse than the bug it fixed.** Round 1's clipped-focus-ring fix used `focus-visible:ring-inset`; an inset `box-shadow` paints *beneath* content, and the button's only child is an opaque image, so no indicator showed at all. The shipped fix is an outline with a negative offset, verified visually in Chromium.
- **Round 2's video deferral broke posterless video.** `preload="none"` was applied unconditionally, leaving every federated video an empty box — and the tests round 2 added *pinned the broken case*. Round 3 gated it on the poster; round 4 then found that only the `poster` half of that gate was defended and closed the other half.
- **`flex-none` had no test at all**, and it is the one class whose removal silently turns the entire feature off.

Two items were deliberately **not** fixed, for reasons recorded on the threads: `lib/components/posts/post.test.tsx` remains on the `tsconfig.typecheck.json` exclude ratchet (taking it off needs nine unrelated pre-existing fixes — note `lib/types/domain/attachment.test.ts` came off the ratchet on `main` and the blocks this PR adds to it typecheck clean), and `MessageBubble`'s pre-existing invisible `focus-visible:ring-inset`, which this PR only swaps a helper import in.

## Rebased onto current `main`

This branch was rebased (not merged) onto `main` at `e663078c`, replacing the earlier merge commit with a linear five-commit history. Two doc sections conflicted — `AGENTS.md` and `REVIEW.md`, where `main`'s new "Status delete federation" section had been added at the same place as this branch's "Post media layout" section. Both are kept; in `AGENTS.md` this branch's section is deliberately ordered *first*, because it is an `###` belonging to `## Status Posts & Actions` and would otherwise read as nested under the new `##`.

Worth noting for review: [#1539](https://github.com/llun/activities.next/pull/1539) landed a click-to-play YouTube player on link preview cards, which touches the same area as this PR's link-preview change. There is no semantic conflict — #1539 changed `link-preview-card.tsx` (what the card renders), while this PR changes only the condition in `post.tsx` deciding *whether* the card renders at all. Both sets of `post.test.tsx` cases survive and pass.

## Checklist

- [x] `yarn run prettier --write .`, `yarn lint`, `yarn typecheck`, `yarn build`, and `yarn test` all pass locally, run in that order — 878 test files, 11,263 tests, on the rebased branch
- [x] Docs updated: new **Post media layout** sections in `AGENTS.md` and `REVIEW.md` plus the matching `CLAUDE.md` reminder; grepped for the conventions this PR reshapes

---
_Generated by [Claude Code](https://claude.ai/code/session_0155eLvoCNCxk9foh7t6SPKU)_


lib/components/posts/useMediaStripScroll.test.tsx
```diff
@@ -0,0 +1,287 @@
...
+    value: scrollLeft
+  })
+}
+
+// A ResizeObserver whose only job here is to prove it gets disconnected — the
+// hook's `measure` reads the element directly rather than the entry, so the
+// stub never needs to deliver anything for the assertions in this file.
+let disconnected = 0
+
+class ResizeObserverStub {
```
llun, line 49: **HIGH — the ResizeObserver → `measure` wiring, the hook's entire reason to exist, is asserted by nothing.**

This stub never invokes its constructor callback (its own comment says so), and the only assertion about it is `disconnected === 1`. Two mutations to `useMediaStripScroll.ts` survive with the whole suite green: deleting `observer.observe(element)`, and constructing `new ResizeObserver(() => {})` instead of `new ResizeObserver(measure)`.

**Failure scenario:** a three-photo post sits in a 500px column; the viewer collapses the sidebar and the column becomes 900px, so nothing overflows any more. With either mutation the right edge is never re-read, so the forward chevron and the right-edge fade stay painted over a strip that has nothing further to reach, and `aria-label` keeps saying "3 media attachments, scroll for more". CI reports success.

Capture the constructor callback in the stub, assert `observe` was called with the node, then re-stamp `scrollWidth` and invoke the captured callback inside `act()`, asserting `right` flips to `false`.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,343 @@
...
+  const items = useMemo(
+    () =>
+      pictures.map((attachment) => ({
+        attachment,
+        width: Math.round(STRIP_ROW_HEIGHT * getMediaGeometry(attachment).ratio)
+      })),
+    [pictures]
+  )
+
+  const strip = useMediaStripScroll(items.map((item) => item.width).join(','))
```
llun, line 212: **HIGH — the "content key is widths, not count" rule is unpinned at the component level.**

`useMediaStripScroll.test.tsx` proves the *hook* honours `contentKey`, but replacing this line with `useMediaStripScroll(String(items.length))` leaves both test files green. That is precisely the defect an earlier review round fixed, and nothing now stops it coming back.

**Failure scenario:** a post with a 1200×500 panorama and a 600×900 portrait (576px + 160px) is edited to swap the panorama for another portrait (160px + 160px). Item count, container box and `scrollLeft` are all unchanged, so the observer never fires and a count-keyed effect never re-runs — the forward chevron and right-edge fade remain over a strip that now fits.

Assert the key the component actually passes, or rerender with a swapped attachment set and assert the affordances clear.

---

lib/components/posts/post.tsx
```diff
@@ -377,11 +378,14 @@ export const Post: FC<PostProps> = (props) => {
...
-          presence of a fitness row. */}
+          presence of a fitness row. The media clause asks the same question of
+          the attachments for the same reason: `Attachments` skips anything
+          `Media` renders as nothing, so a bare `.length` would yield the card
+          to a PDF that puts no block on screen to yield to. */}
       {actualStatus.linkPreview &&
       !actualStatus.quote &&
       !fitnessFile &&
-      actualStatus.attachments.length === 0 ? (
+      !actualStatus.attachments.some(isRenderableAttachment) ? (
```
llun, line 388: **HIGH — this new gate has no test anywhere, and neither does `isRenderableAttachment` itself.**

Reverting this to `actualStatus.attachments.length === 0` leaves `post.test.tsx` and `posts.test.tsx` fully green. The existing `yields to media attachments` case uses an `image/png`, which both spellings suppress, so it cannot see the difference. Separately, `isRenderableAttachment` is the one new export with zero direct coverage — neutering it to `isVisualAttachment(attachment)` alone survives `attachment.test.ts`, `post.test.tsx` and `attachments.test.tsx` together.

**Failure scenario:** a status with a link card and one `.fit` attachment. `Attachments` renders nothing for it, so under the reverted gate the card silently disappears and the post shows bare link text with no block in its place — and CI reports success.

Needs one `post.test.tsx` case with an `application/vnd.ant.fit` attachment asserting the card still renders, plus an `isRenderableAttachment` table beside the `isVisualAttachment` one.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,343 @@
...
+                : `${items.length} media attachments`
+            }
+            className="no-scrollbar flex gap-1.5 overflow-x-auto"
+            style={stripStyle}
+          >
+            {items.map(({ attachment, width }, index) => (
+              <button
+                key={attachment.id}
+                type="button"
+                onClick={openMedia(index)}
```
llun, line 317: **MEDIUM — the strip's per-item lightbox index is not pinned.** Changing this to `onClick={openMedia(0)}` survives the whole suite.

The test named "hands the lightbox exactly the pictures on screen, indexed into that list" only ever clicks `getAllByRole('button')[0]`, where a correct `index` and a hardcoded `0` are indistinguishable.

**Failure scenario:** a four-photo post; the reader taps the third photo and `MediasModal` opens on the first, reading "1 of 4".

Fix in that same test: click the *second* picture and assert the index is `1`.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,343 @@
...
-            className="h-full w-full object-cover"
-            attachment={attachments[1]}
-          />
-        </button>
-      </div>
-    )
+  const stripStyle: CSSProperties = {
+    height: STRIP_ROW_HEIGHT,
+    scrollSnapType: STRIP_SNAP_TYPE,
+    maskImage: edgeFadeMask,
```
llun, line 290: **MEDIUM — the mask is never asserted to be *attached* to the strip, and the comment claiming a test cannot read it back is overstated.**

`buildEdgeFadeMask` is unit-tested as a pure function, but replacing `maskImage`/`WebkitMaskImage` here with `undefined` leaves every test green. The comment at the top of `buildEdgeFadeMask` says a test cannot read the value off the node — that is true only of the two variants carrying `calc()`. The left-edge-only value (`linear-gradient(to right, transparent 0, #000 48px, #000 100%)`) round-trips through jsdom fine, so `renderScrolledStrip({ scrollLeft: 500 })` does yield a non-empty `strip.style.maskImage`.

**Failure scenario:** someone "cleans up" `stripStyle`, the fade stops rendering, every photo butts hard against both edges of the post with no hint of clipping, and no test objects.

Add the assertion in the scrolled-to-end case, and narrow the comment to the `calc()` variants.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,343 @@
...
           >
             <Media
               className="h-full w-full object-cover"
               attachment={attachment}
             />
           </button>
-        ))}
-      </div>
+        </div>
+        {audioPlayers}
```
llun, line 282: **MEDIUM — `{audioPlayers}` in the single-picture branch is untested; deleting this line leaves the suite green.**

The audio section only renders audio *alone*, and the single-image cases only use images — no test renders one picture plus one audio.

**Failure scenario:** a post with one photo and one `audio/mpeg` file loses the `<audio controls>` element entirely. The attachment is silently dropped from the post while `Attachments` still returns the picture, so nothing looks broken.

Add one case with `[image, audio]` asserting both the button and the `<audio>` render.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,343 @@
...
+const FALLBACK_ASPECT_RATIO = 4 / 3
+const FALLBACK_ASPECT_RATIO_CSS = '4 / 3'
+
+/**
+ * The shapes a media box is allowed to take. A 10x10000 sliver would round one
+ * axis of its box to zero and render an invisible, unclickable item, and remote
+ * dimensions are whatever the origin server said they were. `object-cover`
+ * crops whatever the clamp trims.
+ */
+const MIN_ASPECT_RATIO = 1 / 3
```
llun, line 62: **LOW — `MIN_ASPECT_RATIO` is not pinned to a value.** Changing `1 / 3` to `1 / 10` survives the suite.

The sliver case asserts only that the emitted `aspect-ratio` is `not.toBe('10 / 10000')`, and its expected `44px` width comes from the 10-pixel natural-width cap and the target-size floor, not from the clamp — so the clamp's value is invisible to it. The `MAX_ASPECT_RATIO` side *is* pinned, by the panorama's `1260px`.

Assert the positive value, or use a sliver whose natural width doesn't mask the clamp — e.g. 400×10000, which should give `min(100%, 140px)`.

---

lib/components/posts/attachments.test.tsx
```diff
@@ -0,0 +1,629 @@
...
+      const filler = buildAttachment({ width: 800, height: 600 })
+      render(
+        <Attachments
+          status={buildNoteStatus([target, filler])}
+          onMediaSelected={vi.fn()}
+        />
+      )
+
+      const buttons = screen.getAllByRole('button')
+      const button = buttons.find(
```
llun, line 368: **LOW — this lookup is ambiguous for one row of the table.**

The button is located by matching its width, and the `800×600` row's filler attachment is also `800×600` — so that row would pass against the filler even if the target rendered wrong.

Give the filler a distinct shape (e.g. 400×400 → 240px) so the match identifies exactly one element.

---

llun (commented):
**Sub-agent review round 1 — test quality & coverage lane.**

Every finding below was verified empirically: the source was mutated, the suite run, and the mutation reverted. A finding is listed only where the mutation survived with CI green.

What already holds up, for the record: `renderScrolledStrip` genuinely drives the component (deleting `onScroll={strip.measure}` fails 6 tests); the hook test's rerender case proves what it claims (removing `contentKey` from the effect deps fails exactly that test, so ref-identity churn is not what drives it); and `x proximity`, the 78% cap, `scroll-snap-align`, `stopPropagation` on the picture buttons, `MIN_MEDIA_WIDTH`, `SINGLE_MAX_HEIGHT`, `STRIP_ROW_HEIGHT`, `SCROLL_PAGE_RATIO`, the tolerance boundary (`>` vs `>=`), the lazy/eager split, the unfiltered-lightbox regression and the live `aria-label` are all pinned. House rules are clean — jsdom docblocks, `vi.*` only, no name sigils, `$description` tables, same-directory imports, no `console.*`, no tsconfig ratchet change, and no `.ts` twin to trip `lib/testFileTypeCoverage.test.ts`.

Two further gaps this lane's mutation probes surfaced are **already fixed** in the working tree and are not filed below: the chevrons' `stopPropagation` and the back chevron's scroll direction were both unpinned, and now have tests.

Reviews from the invariants and docs/surfaces lanes follow separately.

---

lib/components/posts/useMediaStripScroll.test.tsx
llun: Fixed in d2dc3b8. The stub now captures both the constructor callback and the observed node, and there are two new cases: `observes the scroll container itself, not the viewport` asserts `observe` was called with the scroller node, and `re-measures when the observer fires` widens `clientWidth` to 1000, invokes the captured callback inside `act()`, and asserts `right` flips to `false`.

Both mutations you named now fail: deleting `observer.observe(element)` → 1 failed / 14 passed; `new ResizeObserver(() => {})` → 1 failed / 14 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in d2dc3b8, using your rerender suggestion: `re-measures when an edit changes item widths but not their count` renders a 1200×500 panorama plus a 600×900 portrait (576 + 160), stamps `scrollWidth` behind a getter so it can change without re-stamping the node React reuses, then rerenders with the panorama swapped for another portrait (160 + 160) and asserts the forward chevron is gone.

`useMediaStripScroll(String(items.length))` now fails that test — 1 failed / 41 passed.

---

lib/components/posts/post.tsx
llun: Fixed in d2dc3b8, both halves.

`post.test.tsx` gains two cases either side of the line: `does not yield to an attachment nothing can render` (a link card plus one `application/vnd.ant.fit`, asserting the card still renders) and `yields to an audio attachment, which does render` — the second matters because it is what stops the gate being narrowed to `isVisualAttachment`. `attachment.test.ts` gains `isRenderableAttachment` and `isAudibleAttachment` tables.

Both mutations now fail: reverting the gate to `attachments.length === 0` → 1 failed / 72 passed; neutering `isRenderableAttachment` to `isVisualAttachment(attachment)` → 3 failed / 107 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in d2dc3b8 — the test now clicks the second rendered picture and asserts index `1`, with a comment saying why it must not be the first.

`onClick={openMedia(0)}` now fails — 1 failed / 41 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in d2dc3b8, and you were right about the comment. `attaches the edge fade to the strip itself` renders at `scrollLeft: 500` — the one variant with no `calc()` stop — and asserts `strip.style.maskImage` contains `linear-gradient` and `48px`. The `buildEdgeFadeMask` doc comment now says the parser rejects *the two variants carrying `calc()`* rather than claiming the value can never be read back.

`maskImage: undefined` now fails — 1 failed / 41 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in d2dc3b8 — new `a picture alongside audio` case renders `[image, audio]` and asserts exactly one button plus an `<audio>` element.

Deleting `{audioPlayers}` from the single-picture branch now fails — 1 failed / 41 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in d2dc3b8 with your 400×10000 suggestion — a new row, `lays a clamped sliver out at the 1:3 floor, not narrower`, expects `min(100%, 140px)`, wide enough that neither the natural-width cap nor the 44px target floor masks the clamp. The original 10×10000 row stays, with a comment explaining that its 44px comes from those two and not from the clamp.

`MIN_ASPECT_RATIO = 1 / 10` now fails — 1 failed / 41 passed.

---

lib/components/posts/attachments.test.tsx
llun: Fixed in d2dc3b8 — the filler is now 400×400 (240px), distinct from all three rows, with a comment saying that is why. The width lookup can only match the target.

---

lib/components/posts/media.tsx
```diff
@@ -109,6 +120,7 @@ export const Media: FC<Props> = ({
       <img
         onClick={onClick}
         key={id}
+        loading={loading}
```
llun, line 132: **MEDIUM — removing the 4-item cap has no mitigation for `<video>`; `loading="lazy"` only reaches ``.**

`isVisualAttachment` routes `video/*` into the strip, and `ACCEPTED_FILE_TYPES` accepts `video/mp4`, `video/webm` and `video/quicktime`. `loading` is forwarded only to the two `` branches; the `<video>` branch below sets no `preload`, so the UA default (`metadata`) applies, and its `poster` is fetched eagerly too. `posts.maxMediaAttachments` defaults to `MAX_STORED_MEDIA_ATTACHMENTS` = 50.

**Failure scenario:** a Note with 20 video attachments in the home timeline. Before this PR: 4 `<video>` mounted → 4 metadata range requests. After: 20 mount → 20 metadata requests plus 20 poster fetches on render, ~16 of them for items scrolled off the right of the strip that the viewer never looks at — multiplied by every such post in the feed.

`preload="none"` on the strip's videos is the exact analogue of the image fix. Related: `attachments.test.tsx` uses only `image/jpeg`, `audio/mpeg` and `application/vnd.ant.fit`, so the video path through the new layout has no coverage at all.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+ * A strip item cannot use that ring. Its border box is exactly the strip's
+ * height and `overflow-x-auto` forces `overflow-y` to compute to `auto`, so an
+ * OUTSET ring's top and bottom bars fall outside the scrollport and are clipped
+ * away. An INSET one is worse: an inset `box-shadow` paints with the element's
+ * background, underneath its content, and the button's only child is an opaque
+ * image filling the whole box — so it is occluded on all four sides and there
+ * is no indicator at all. An outline with a negative offset draws inside the
+ * border box AND paints above content, which is the combination this needs.
+ */
+const STRIP_FOCUS_CLASS =
```
llun, line 157: **LOW — the one behavioural fix round 1 made is pinned by no test at any level.**

Verified empirically in a scratch worktree: replacing `cn(MEDIA_BOX_CLASS, STRIP_FOCUS_CLASS, 'h-full flex-none')` with `cn(MEDIA_BOX_CLASS, 'h-full flex-none')` leaves **29 files / 329 tests passing**. Reverting instead to the already-shipped-and-rejected `focus-visible:ring-2 focus-visible:ring-inset` spelling keeps the constant referenced, so it survives the unused-binding check too.

This indicator has now been got wrong twice — an outset ring clipped by the strip's overflow, then an inset ring occluded by the image — and it is the only rule in the new AGENTS.md section with no executable guard. The repo has direct precedent for a class-string guard of exactly this kind (`lib/components/tailwindCssVariableSyntax.test.ts`).

Two assertions close it: `toHaveClass('focus-visible:outline-2', 'focus-visible:-outline-offset-2')` on a strip item, and `toHaveClass('focus-visible:ring-2')` on the lone picture.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+  'focus-visible:outline-2 focus-visible:-outline-offset-2 focus-visible:outline-ring/50'
+
+/**
+ * The chevrons are POINTER affordances and deliberately sit outside the tab
```
llun: **LOW — this comment overclaims: `tabIndex={-1}` does not prevent the focus loss it says it prevents.**

`tabindex="-1"` removes an element from the *sequential* tab order only. It stays click-focusable, and Chrome and Firefox focus a `<button>` on click — so the scenario the comment describes still happens via the mouse.

**Failure scenario:** strip at `scrollLeft: 200`, `clientWidth: 500`. Mouse-click "Previous media" → the button takes focus on mousedown → `scrollByPage(-1)` scrolls by −350 → `scrollLeft` reaches 0 → `canScrollLeft` goes false → the button holding DOM focus unmounts → `document.activeElement` falls back to `<body>` → the next Tab restarts from the top of the page. Same at the right end for "More media".

`tabIndex={-1}` is still right for the *keyboard* half of the claim, but the mouse half needs the button not to take focus in the first place — `onMouseDown={(event) => event.preventDefault()}` is the usual way — and the comment should stop asserting more than the attribute delivers.

---

AGENTS.md
```diff
@@ -1557,6 +1557,116 @@ consistency is enforced by keeping the wiring in one place rather than per page.
...
+  which is the wrong trade for the post's largest element.
+- **The edge fade is a `mask-image`, not a background gradient.** Posts render on
+  four different surfaces — `bg-card` when framed, `bg-background` on the status
+  detail, `bg-muted/30` for an ancestor row, and the page itself when unframed —
+  so a fade painted in any one token is visibly wrong on the other three, and a
+  literal white one is wrong in dark mode everywhere. A mask fades the strip's
+  own pixels and lets whatever is behind show through. It lives in
+  `buildEdgeFadeMask` rather than inline because it is the one part of the strip
+  a test cannot read back off the node: jsdom's CSS parser rejects a gradient
+  carrying `calc()` and stores nothing. Known cosmetic cost: the mask also fades
```
llun: **LOW — this section now contradicts itself and describes a ring that strip items no longer have.**

Two stale claims survived round 1's edit:

1. "It lives in `buildEdgeFadeMask` rather than inline because it is **the one part of the strip a test cannot read back off the node**" — round 1 narrowed that in the source JSDoc (now "can only ever be asserted against the left-edge-only form") *and* disproved it by adding a test that reads `strip.style.maskImage` off the node.
2. "the mask also fades the outer edge of a focused item's **ring**" — three lines below, the same section states that a strip item's indicator is an outline with a negative offset, precisely because an outset ring is clipped away. A reader following the first sentence looks for a ring that cannot exist. What the mask actually fades is the leading edge of the *outline*, which is where a Tab-focused item gets scrolled to.

(The reviewer also claimed `CLAUDE.md` repeats the "cannot read back" rationale — it does not; it gives only the accurate jsdom/`calc()` reason.)

---

CLAUDE.md
```diff
@@ -128,6 +128,7 @@
...
...ndexed into that list, or `MediasModal` shows a blank slide and a wrong "n of m". Anything asking "do I have media" asks `isRenderableAttachment`, not `attachments.length` (`post.tsx`'s link-preview suppression does). Every picture button carries an explicit `aria-label`, since federation writes `attachment.name || ''` and an undescribed photo announced as a bare "button". **A stored `0` means "unknown", not "zero pixels"** — media-storage paths persist `metaData.width ?? 0` — so all dimensions go through `getMediaGeometry`, which also clamps the shape (a 10x10000 sliver otherwise rounds an axis to zero) and falls back to `FALLBACK_ASPECT_RATIO` so blurhash has a box. Affordances come from `useMediaStripScroll` (container-measured, **callback** ref because the strip is conditional), whose `contentKey` must describe item **widths**, not their count. The `no-scrollbar` utility it needs is now defined in `app/globals.css` and belongs **only** on a row with its own overflow affordance — it had been applied in two pickers while defined nowhere, and defining it would have taken away their only cue, so both were switched back to a plain scrollbar. See **Post media layout** in `AGENTS.md`.
```
llun: **LOW — `REVIEW.md`, the authoritative review checklist, was never updated.**

Filing this here because `REVIEW.md` is not in the diff, so there is no line to anchor to.

This PR adds a ~110-line convention block to `AGENTS.md` and a bullet to this file, but `REVIEW.md` is untouched — against its own docs-hygiene rule at `REVIEW.md:527`: *"conventions → `AGENTS.md` + this checklist."* Grepping it for `Post media`, `attachments.tsx`, `no-scrollbar`, `isRenderableAttachment` or `mask-image` returns nothing.

It is not a category mismatch either: `REVIEW.md` already carries feature-specific sections of exactly this kind (`## Link preview cards`, `## Apple Maps basemap`, `## Stored media`). So a reviewer working from the authoritative checklist has no entry for the no-grid rule, the 78%/`proximity` pair, the `isRenderableAttachment`-not-`.length` rule, or the strip focus-indicator rule.

---

llun (commented):
**Sub-agent review round 2** — two lanes (regression: what round 1 broke or missed; mutation: what the current suite still fails to catch), each adversarially verified. The mutation lane ran in a disposable git worktree this time, so it could mutate freely without racing the working tree.

Eight findings were raised; the verify pass **refuted two** and they are not filed:

- *"the strip's focus outline paints on the photograph at 50% alpha with no contrast floor"* — refuted on three counts. The geometry is misstated (with `-outline-offset-2` the band runs 2px inward from the border box, covering the 1px border plus 1px of image, not a line laid across the photo); the SC is miscited (the 3:1 focus-indicator requirement is 2.4.13 *Focus Appearance*, **AAA** — 2.4.11 is *Focus Not Obscured*); and `ring-ring/50` is the app-wide focus idiom, 24 occurrences, including the pre-PR grid this replaced. Neither the token nor its alpha is introduced by this PR.
- *"the invisible `ring-inset` in `MessageBubble` was documented rather than fixed"* — technically accurate but a pre-existing bug in code this PR only touches to swap a local helper for a shared import. AGENTS.md records it explicitly as "a pre-existing bug, not a precedent to copy", i.e. a deliberate scope boundary.

One further finding is **acknowledged and not being fixed**: `post.test.tsx` and `attachment.test.ts` both sit on the `tsconfig.typecheck.json` exclude ratchet, so `yarn typecheck` does not see the blocks round 1 added to them. I confirmed the added blocks are type-clean today — running `tsc -p tsconfig.json`, every error in those two files is at a pre-existing line — so nothing is broken; taking them off the ratchet means fixing 9 unrelated pre-existing type errors, which does not belong in a media-layout PR. Flagging rather than silently widening.

Also verified clean, so nobody re-derives it: `twMerge` preserves every focus utility in both `cn(...)` call sites; `MEDIA_BOX_CLASS` has exactly two consumers and both carry a focus class; `SINGLE_FOCUS_CLASS` keeps `outline-none` so no UA outline is re-exposed; an element's own `overflow: hidden` does not clip its own outline; hooks all run before both early returns; empty/audio-only/`Announce` statuses all return correctly; and the `.no-scrollbar` removal from the two pickers is a genuine no-op against `main`.

---

lib/components/posts/media.tsx
```diff
@@ -87,6 +97,7 @@ export const Media: FC<Props> = ({
             }}
             onClick={onClick}
             key={id}
+            loading={loading}
```
llun, line 109: **MEDIUM — the two tests named for the lazy-loading rule exercise only the branch production does *not* take.**

`buildAttachment` in `attachments.test.tsx` never sets `blurhash`, so both `lazy-loads strip images, which are now uncapped` and `loads a lone picture eagerly…` assert against the plain `` further down. `media.test.tsx` matches "loading" nowhere. Deleting `loading={loading}` from *this* line — the blurhash `` — survives 126 tests.

**Failure scenario:** locally-uploaded images all carry a blurhash (`S3StorageFile` computes it on save; `lib/database/sql/media.ts` selects and returns it), so this is the branch that actually renders in production. A 12-photo post fires 12 full-size image requests on render — exactly the regression `loading="lazy"` was added to prevent — with the test named for it still green.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+            {items.map(({ attachment, width }, index) => (
+              <button
+                key={attachment.id}
+                type="button"
+                onClick={openMedia(index)}
+                aria-label={mediaLabel(attachment, index)}
+                className={cn(
+                  MEDIA_BOX_CLASS,
+                  STRIP_FOCUS_CLASS,
+                  'h-full flex-none'
```
llun, line 322: **MEDIUM — `flex-none` is the single declaration that makes the strip overflow at all, and nothing asserts it.**

Mutating `'h-full flex-none'` → `'h-full'` leaves the suite green, because every scroll test fakes `scrollWidth`/`clientWidth` with `Object.defineProperty` — no test can observe real overflow.

**Failure scenario**, measured in Chromium rather than reasoned: three 800×600 photos in a 600px post column lay out at 320px each with `flex-none` (`scrollWidth` 972 vs `clientWidth` 600). Without it, the default `flex-shrink: 1` squeezes them to 196px each and `scrollWidth` 600 === `clientWidth` 600. `canScrollRight` is then permanently false — no chevrons, no edge fade, no 78% peek, no scrolling at all — and every photo is squeezed and cropped by `object-cover`. The entire feature is off and CI is green.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+
+/**
+ * A stored `0` means "dimensions unknown", NOT "zero pixels wide" — several
+ * media-storage paths persist `metaData.width ?? 0` (`lib/services/medias/`),
+ * and a federated `Document` carries whatever the remote server sent. So every
+ * consumer of a dimension has to apply this same guard; reading
+ * `attachment.width` raw collapses the box to nothing.
+ */
+const getMediaGeometry = ({ width, height }: Attachment) => {
+  if (!width || !height || width <= 0 || height <= 0) {
```
llun, line 80: **MEDIUM — this guard's four clauses are only covered in pairs, so the one-sided-dimension case is untested and each clause can be dropped alone.**

The table covers *both* dimensions missing and *both* negative. Removing `!width ||` survives 122 tests: the both-undefined row is still caught by `!height`, the `0,0` row by `width <= 0`, the `-800,-600` row by `width <= 0`.

**Failure scenario:** `width` and `height` are independently `.optional()` on `Attachment`, so a federated `Document` can arrive with `height: 600` and no `width`. Under that mutant `width / height` is `NaN`, and the rendered output was measured as `aspectRatio: ''` and `width: min(100%, nanpx)` — both invalid declarations the browser drops — so the photo collapses to a zero-size box. That is the same class of bug round 1 fixed for the both-missing case.

Needs rows for `{800, undefined}`, `{undefined, 600}`, `{-800, 600}`, `{800, -600}`.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+        <div className="mt-3 flex justify-start">
           <button
-            key={attachment.id}
-            onClick={(e) => {
-              e.stopPropagation()
-              handleClick(i)
-            }}
-            className="relative aspect-square cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/50"
+            type="button"
+            onClick={openMedia(0)}
```
llun, line 271: **LOW — the lone picture's lightbox index is unpinned, and getting it wrong throws rather than degrading.**

Round 1 pinned the *strip's* per-item index by clicking the second picture, but nothing asserts what `onMediaSelected` receives on the single-picture branch — the only single-picture click test passes `vi.fn()` and asserts nothing about it. Mutating to `openMedia(1)` survives 131 tests.

**Failure scenario:** `MediasModal` sets `currentIndex` straight from `initialSelection` with no wrapping (`getWrappedIndex` is applied only on prev/next), so `medias=[a], initialSelection=1` gives `visibleIndices = [0, 1, 0]` and `medias[1].id` throws `TypeError: Cannot read properties of undefined (reading 'id')`. The timeline render blows up — not a blank slide.

---

lib/components/posts/attachments.tsx
```diff
@@ -1,144 +1,368 @@
...
+/**
+ * The row height two or more attachments lay out at, and the height a single
+ * one is scaled down to. Both come from the design system's media strip.
+ */
+const STRIP_ROW_HEIGHT = 240
+const SINGLE_MAX_HEIGHT = 420
+
+/**
+ * No single item may fill the strip, so the next one always peeks past the
+ * edge. That peek — not the chevrons, which need a pointer to appear — is what
```
llun: **LOW — this comment contradicts another comment in the same file, and AGENTS.md carries the same contradiction.**

Here: the peek matters "not the chevrons, **which need a pointer to appear**". 320 lines later the back-chevron comment says the opposite: "The forward chevron opposite **stays visible** so 'there is more' is discoverable without a pointer" — and the code agrees with the second one, since only the back chevron carries `opacity-0`/`group-hover` gating. AGENTS.md repeats both claims two bullets apart. One of them has to go.

Separately (same area): the forward chevron's always-visible rule is unasserted — adding `opacity-0 pointer-events-none group-hover/media:opacity-100` to it survives 122 tests, and on a touch screen `group-hover` never latches, so the cue the comment calls pointer-free would be permanently invisible.

---

lib/components/posts/useMediaStripScroll.ts
```diff
@@ -0,0 +1,94 @@
+'use client'
+
+import { useCallback, useEffect, useState } from 'react'
+
+/**
+ * Sub-pixel scroll offsets — and a scroller whose content is a hair wider than
+ * its box — must not flicker the edge affordances on and off, so each end has a
+ * dead zone before it counts as scrollable.
+ */
+export const SCROLL_EDGE_TOLERANCE = 8
```
llun, line 10: **LOW — `SCROLL_EDGE_TOLERANCE` is tested against itself, so its value is unpinned.**

The boundary table imports the constant and derives every expectation from it (`scrollLeft: SCROLL_EDGE_TOLERANCE`, `+ 1`, `MAX - TOLERANCE`, `MAX - TOLERANCE - 1`), so setting it to `0` keeps all four rows green: `0 > 0` false, `1 > 0` true, `500 < 500` false, `499 < 500` true. This is the same gap round 1 closed for `MIN_ASPECT_RATIO` by asserting the literal `140px`.

**Failure scenario:** the dead zone the doc comment exists for is gone, so a scroller whose content is one rounded pixel wider than its box reports `canScrollRight` permanently — a forward chevron and right-edge fade over a strip with nothing to reach.

Also unasserted on the same theme: `object-cover` on both `Media` call sites. Switching it to `object-contain` survives, and since `Media` branches on `className?.includes('object-contain')` the mutation flips the blurhash branch too — the clamped 10×10000 sliver that `MIN_ASPECT_RATIO`'s comment says `object-cover` crops gets letterboxed instead.

---

llun (commented):
**Sub-agent review round 2 — mutation lane.**

119 mutants applied, run and reverted in a disposable git worktree: **100 killed, 19 survived**. The verify pass then refuted 5 and confirmed 7. Filed above.

Refuted, for the record:

- *"the focus indicator has zero coverage"* — the reviewer ran against committed `d2dc3b8`, where that was true; the `describe('focus indicators')` block had already landed in the working tree. Reverting `STRIP_FOCUS_CLASS` to the outset-ring form now fails 2 tests.
- *"the identity-preserving `setEdges` updater is unpinned"* — the code is correct and the only guard proposed is a render-counting probe on a two-boolean state object. Brittle enough to cost more than it protects.
- *"cosmetic classes in `MEDIA_BOX_CLASS` / `CHEVRON_CLASS` are unasserted"* — pinning `rounded-xl`, `border-border/60` and a centring transform is asserting Tailwind spelling, not behaviour.
- *"no `no-scrollbar` regression guard on the two pickers"* — a suggestion for a new repo-wide guard, not a defect in this diff. The existing repo-wide guards exist because their rules are invisible at review time; a literal class name in a two-line diff is not.
- *"the two early returns each survive deletion"* — the reporter itself concluded these are equivalent mutants (the `useMemo` returning `[]` and the `items.length ? … : null` guard already yield an empty fragment). Worth knowing that `renders nothing for a non-Note status` passes for a reason other than the guard it is named after, but nothing to fix.

The two that matter most are the `flex-none` one — dropping one class silently turns the entire feature off, and I measured that in Chromium rather than trusting the reasoning — and the lazy-loading one, where the tests named for the rule turn out to assert the branch production doesn't take.

---

lib/components/posts/media.tsx
llun: Fixed in b2f804b. `Media`'s `loading` prop now reaches a video as `preload={loading === 'lazy' ? 'none' : undefined}`, so strip videos defer their metadata and a lone video stays eager like a lone picture. Three new tests cover the video path, which had none: it lays out in the strip at its ratio-derived width, strip videos carry `preload="none"`, and a lone video carries no `preload`.

One correction to the finding, now in the prop's doc comment: `preload="none"` does **not** defer the `poster` — nothing declarative does. That is one image request per clip, the same as a photo, and the strip's videos hide their controls anyway, so losing the preloaded duration costs nothing here.

Dropping the `preload` now fails — 1 failed / 58 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed — a `describe('focus indicators')` block landed with the round-2 changes and is in b2f804b: it asserts `focus-visible:outline-2`, `focus-visible:-outline-offset-2` and `focus-visible:outline-ring/50` on a strip item, plus `expect(item.className).not.toContain('focus-visible:ring-')` so the rejected spelling cannot come back silently, and `focus-visible:ring-2`/`focus-visible:ring-ring/50` on the lone picture.

Both mutations you named now fail: dropping `STRIP_FOCUS_CLASS` from the `cn()` call → 1 failed / 48 passed; reverting it to `focus-visible:ring-2 focus-visible:ring-inset` → 1 failed / 48 passed.

---

lib/components/posts/attachments.tsx
llun: You were right, and this was a real hole rather than only an overclaim. Fixed in b2f804b: both chevrons now carry `onMouseDown={preventFocusOnPress}`, which cancels the default so the browser never moves focus to them, leaving the click and every keyboard path untouched.

The comment no longer asserts more than the attribute delivers — it now says keeping them out of the tab order takes **both** `tabIndex={-1}` and the mousedown guard, and spells out that `tabindex="-1"` covers the sequential order only.

Two new table-driven tests assert `defaultPrevented` on a mousedown against each chevron; deleting the guard fails 2 of them.

---

AGENTS.md
llun: Both fixed in b2f804b.

The "one part of the strip a test cannot read back off the node" sentence now matches the source JSDoc: it lives in `buildEdgeFadeMask` so the string is unit-testable, and jsdom rejects only the two variants carrying `calc()`, so a *rendered* node can be asserted against the left-edge-only form.

The "Known cosmetic cost" sentence no longer says "ring" — it now says the mask fades the leading edge of a focused item's **outline**, which is where the browser scrolls a Tab-focused item to.

And thanks for the correction about `CLAUDE.md` — you were right that it never carried the "cannot read back" claim, so I left it alone.

---

CLAUDE.md
llun: Fixed in b2f804b — `REVIEW.md` gains a `## Post media layout` section, placed beside the other feature sections you named. It covers the no-grid rule, the three load-bearing strip details (78% peek, `flex-none`, `x proximity`), the no-cap/deferral pair, the mask, the layout-only nature of the filtering and the `isRenderableAttachment`-not-`.length` rule, the `0`-means-unknown dimension guard, the focus-indicator rule, `contentKey`, and where `no-scrollbar` may be applied.

---

lib/components/posts/media.tsx
llun: This was the best catch of the round — the tests named for the rule were pinning the branch production doesn't take. Fixed in b2f804b: both are now table-driven across a plain image and one carrying a blurhash, using a `BLURHASH` constant with a comment saying why (`Media` takes the blurhash branch for every locally uploaded image).

Deleting `loading={loading}` from the blurhash `` now fails — 1 failed / 58 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in b2f804b — a new test asserts `flex-none` on every strip item, with a comment recording exactly what you measured: without it the items shrink to fit, `scrollWidth === clientWidth` forever, and the whole feature is off.

Pinning the class rather than the layout is a compromise I've noted in the test: jsdom lays nothing out, so the geometry your Chromium measurement showed can't be reproduced in the suite. The class is the only thing at this level that can carry the rule.

Dropping `flex-none` now fails — 1 failed / 58 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in b2f804b — the unknown-dimensions table gains all four rows you named: `{undefined, 600}`, `{800, undefined}`, `{-800, 600}` and `{800, -600}`, each asserting the 4:3 fallback box rather than a `NaN`-derived one.

Dropping `!width ||` now fails — 1 failed / 58 passed.

---

lib/components/posts/attachments.tsx
llun: Fixed in b2f804b — a new `opens the lightbox on the lone picture itself` case asserts `onMediaSelected` is called with `([attachment], 0)`, with a comment recording that `MediasModal` reads `initialSelection` unwrapped so an index past the end throws rather than degrading.

`openMedia(1)` now fails — 1 failed / 58 passed.

---

lib/components/posts/attachments.tsx
llun: Both fixed in b2f804b.

The contradiction is gone from all four places it appeared — the `STRIP_ITEM_MAX_WIDTH` comment, `AGENTS.md`, `CLAUDE.md` and the new `REVIEW.md` section. The peek now earns its keep by a claim that is actually true: it is what says "this scrolls" on a touch screen, *where the back chevron never appears at all* and the forward one is easy to miss. The back-chevron comment 320 lines down was the accurate one and is unchanged.

The forward chevron's always-visible rule is now asserted too: a new test checks it carries neither `opacity-0` nor `pointer-events-none`. Hover-gating it fails — 1 failed / 58 passed.

---

lib/components/posts/useMediaStripScroll.ts
llun: Both fixed in b2f804b.

`SCROLL_EDGE_TOLERANCE` now has a standalone `expect(SCROLL_EDGE_TOLERANCE).toBe(8)`, with a comment saying why the derived boundary rows below cannot carry the value themselves. The rows stay — they pin the comparison, the literal pins the constant.

`object-cover` is asserted on every strip image by a new `crops rather than letterboxes, which the ratio clamp relies on` case. Switching it to `object-contain` fails — 1 failed / 58 passed.

---

lib/components/posts/media.tsx
```diff
@@ -128,6 +149,13 @@ export const Media: FC<Props> = ({
         width={width}
         height={height}
         poster={poster}
```
llun, line 151: **MEDIUM — round 2's own fix is a regression for a posterless video: `preload="none"` leaves it a completely empty box.**

Both round-3 lanes found this independently. The video branch has exactly two ways to paint anything before playback: `poster`, and the `#t=0.01` fragment on `<source>`, which needs at least `metadata` to decode a frame. `preload="none"` fetches nothing, so the fragment is inert; there is no blurhash fallback outside the image branch; and the strip leaves `showVideoControl` at `false`.

**Failure scenario:** a boosted remote post with two attachments, one `video/mp4`. `thumbnailUrl` is written on exactly one path — `lib/actions/createNote.ts:576`, the local-upload one — and `lib/jobs/createNoteJob.ts` never passes it, so federated video is always posterless. `isVisualAttachment` admits it, `pictures.length === 2` takes the strip branch, `loading="lazy"` is passed unconditionally → `preload="none"`. The item renders as a bare `bg-muted/20` rounded rectangle with nothing identifying the clip. Before this PR, `origin/main` had no `preload` at all and the same attachment showed its first frame.

Worse, the tests round 2 added **pin the broken case**: both build their videos without `thumbnailUrl` and assert `preload="none"`.

Gate `'none'` on the attachment actually having a poster.

---

AGENTS.md
```diff
@@ -1557,6 +1557,134 @@ consistency is enforced by keeping the wiring in one place rather than per page.
...
+  literal white one is wrong in dark mode everywhere. A mask fades the strip's
+  own pixels and lets whatever is behind show through. It lives in
+  `buildEdgeFadeMask` rather than inline so the string itself is unit-testable:
+  jsdom's CSS parser rejects the two variants carrying `calc()` and stores
+  nothing for them, so a RENDERED node can only be asserted against the
+  left-edge-only form. Known cosmetic cost: the mask also fades the leading edge
+  of a focused item's outline, which is exactly where the browser scrolls a
+  Tab-focused item to.
+- **A strip item's focus indicator is an `outline` with a NEGATIVE offset, not
+  a ring.** Its border box is exactly the strip's height and `overflow-x-auto`
```
llun, line 1699: **LOW — round 2 updated `CLAUDE.md` and the code comments but not the canonical section `CLAUDE.md` points readers at, so the source of truth is now the less complete of the two.**

`CLAUDE.md:131` reads "Keeping the chevrons out of the tab order takes **both** `tabIndex={-1}` and a mousedown guard", names `flex-none` as the declaration that silently turns the whole feature off, and ends "See **Post media layout** in `AGENTS.md`". That section still reads "**The chevrons sit outside the tab order (`tabIndex={-1}`).**" with no mention of `preventFocusOnPress` — i.e. exactly the claim round 2 established is insufficient — and still enumerates "Three details are load-bearing" without `flex-none`. `grep -n flex-none AGENTS.md REVIEW.md` returns nothing.

Same class the loop already accepted and fixed in round 2, when round 1's focus fix left two sentences stale here.

---

llun (commented):
**Sub-agent review round 3** — a regression lane (what round 2 broke) and a second mutation lane, both Opus, each adversarially verified. Round 3 was scoped at the round-2 code specifically, since it was the newest and had never been mutation-tested.

**Two findings, both filed above**, and the first is the one that justifies running a third round: round 2's own fix was a regression for posterless video. Both lanes reached it independently.

The verify pass refuted the rest:

- *"`h-full w-full` is unpinned"* and *"`group/media relative` is unpinned"* — the classes are present and correct; the failure needs a future edit that deletes them. That is an unbounded coverage increment, since every class in those strings is equally unpinned. Rounds 1–2 already took the ones whose removal silently disables the feature (`flex-none`, `object-cover`, `pointer-events-none`, the focus spelling); this is the next tier down with no distinguishing rationale. Dropping `relative` would also throw both chevrons to the page edges — not a silent failure.
- *"the `object-cover` test pins the wrong branch"* — refuted on the merits, and the proposed change would have made the test worse: in the blurhash branch the caller's `className` lands on the wrapper `<div>` where `object-fit` is inert, and the `` gets its own `object-cover` from `Media`'s ternary.

Also re-checked clean: `preventFocusOnPress` suppresses focus without cancelling the click and shadows no other pointer handler in `lib/components/posts/`; no other `Media` caller passes `loading`, so `preload` stays unset everywhere else including the lightbox; and `BLURHASH` is a genuinely valid hash (`isBlurhashValid` → true, `decode` → 4096 bytes), so round 2's new table-driven rows really do exercise the blurhash branch rather than passing through the catch.

---

lib/components/posts/media.tsx
llun: Fixed in 34031442, and you were right that this was my own regression rather than a pre-existing gap. `preload` is now gated on the poster existing:

```
preload={loading === 'lazy' && poster ? 'none' : undefined}
```

so a video with a poster still defers its metadata and a posterless one keeps the default, letting the `#t=0.01` fragment decode a frame. The prop's doc comment no longer reasons only about the poster-present case, and the `preload` line carries its own comment naming federated video as the case that lands there.

The test that pinned the broken case is now two: `defers a strip video that has a poster to paint instead` (both built with `thumbnailUrl`) and `does not defer a posterless strip video, which would show nothing`. Both directions are pinned — ungating the condition back to round 2's form fails, and deleting `preload` outright also fails.

---

AGENTS.md
llun: Fixed in 34031442. The AGENTS.md bullet now reads "**Keeping the chevrons out of the tab order takes BOTH `tabIndex={-1}` and a mousedown guard**", explains that `tabindex="-1"` covers the sequential order only and that Chrome and Firefox focus a button on click, and names `preventFocusOnPress` — so it no longer states the claim round 2 established is insufficient.

`flex-none` is now the **fourth** load-bearing strip detail (the bullet said "Three" before), with the silent-turn-off consequence spelled out and a note that a test pins the class because jsdom lays nothing out. REVIEW.md gained the same two, plus the poster condition from the other thread.

---

lib/components/posts/media.tsx
```diff
@@ -128,6 +149,13 @@ export const Media: FC<Props> = ({
...
+        // Only skip the fetch when there is a poster to paint instead. With no
+        // poster this element's ONLY way to show anything before playback is
+        // the `#t=0.01` fragment below, which needs metadata to decode a frame
+        // — and the strip hides its controls, so `preload="none"` would leave a
+        // bare empty box. Federated video always lands here: `thumbnailUrl` is
+        // written on the local-upload path alone.
+        preload={loading === 'lazy' && poster ? 'none' : undefined}
```
llun, line 158: **Round 4 — only the `poster` half of this gate is pinned; the `loading === 'lazy'` half is free.**

Mutating this line to `preload={poster ? 'none' : undefined}` leaves the whole suite green (28 files, 311 tests under `lib/components/posts` + `lib/components/medias-modal`, and `thumbnailUrl` appears in no other suite that would reach it).

All three existing `preload` assertions live in `attachments.test.tsx`: the postered strip case still yields `'none'` under the mutation, and the other two fixtures are posterless so they still yield `undefined`. `media.test.tsx`'s `renders video with poster and focal point object position` is the one test that renders a **postered video with `loading` unset** — exactly the case that separates the two spellings — but it asserts `poster` and `objectPosition` and never `preload`.

*Failure scenario:* with the mutation shipped, `MediasModal` (`medias-modal.tsx:287` — `<Media showVideoControl …>`, no `loading`) opens a locally-uploaded video with visible controls showing no duration and an empty scrub bar until play, and the lone-video branch (`attachments.tsx:276`) stops preloading the post's largest element. Round 3 established that this gate is exactly where a plausible-looking simplification does damage; half of it is currently undefended.

Fixed in the follow-up commit: one line added to that `media.test.tsx` case — `expect(video).not.toHaveAttribute('preload')`. Mutation-verified — with `loading === 'lazy' &&` removed the assertion fails with `Received: preload="none"`, and passes once restored.

---

CLAUDE.md
```diff
@@ -128,6 +128,7 @@
...
```
llun: **Round 4 — this bullet still states the unconditional `preload` rule that round 3 identified as the bug.**

`34031442` changed the code to `preload={loading === 'lazy' && poster ? 'none' : undefined}` and added the poster qualifier to `AGENTS.md` and `REVIEW.md`, but not here — `git show 34031442 --stat` lists only `AGENTS.md`, `REVIEW.md`, `attachments.test.tsx` and `media.tsx`. The string `poster` appears zero times in the committed `CLAUDE.md`, so this is now the only doc asserting that strip videos get `preload="none"` with no condition.

It went the opposite direction from the round-3 edits around it: round 2 added this clause plus `flex-none` and the mousedown guard to `CLAUDE.md` first, so round 3's `AGENTS.md`/`REVIEW.md` edits for those two were catching `AGENTS.md` *up*. Only the poster condition was left behind here.

*Failure scenario:* `CLAUDE.md` is the file this repo's instructions load unconditionally and tell an agent to read first. An agent reads this line, compares it to `media.tsx:158`, treats `&& poster` as a stray condition and drops it — reintroducing exactly the round-3 regression, where every federated video renders in a strip as a bare empty rounded rectangle (`thumbnailUrl` is written on one path only, `lib/actions/createNote.ts:576`, local upload; the inbound `createNoteJob.ts` never sets it). The two tests at `attachments.test.tsx:482`/`:513` would fail, but this doc would be the stated reason for the change.

This is the third time this loop has filed a "stale doc sentence, three files disagree" finding.

Fixed in the follow-up commit: the poster qualifier appended here, matching `AGENTS.md:1601-1607`.

---

llun (commented):
**Sub-agent review round 4** — two lanes (regression re-read of the round-3 code, and mutation testing against `34031442`), each with an adversarial Opus verifier.

**2 findings, both confirmed by their verifier and both now fixed:**
- `lib/components/posts/media.tsx:158` — only the `poster` half of the round-3 gate is pinned; `preload={poster ? 'none' : undefined}` survives the whole suite.
- `CLAUDE.md:131` — the one doc of the three that `34031442` did not update, still stating the unconditional rule round 3 identified as the bug.

**4 candidates refuted by the verify pass**, all for the same reason — they targeted lines this PR did not introduce, or asked for assertion completeness rather than naming a defect:
- `#t=0.01` unasserted — pre-existing and untouched; it appears in the diff only as context.
- The blurhash-branch `object-cover` ternary — also context; the PR inserted only `loading={loading}` above it, and the crop test already kills a mutation of the new caller string at `attachments.tsx:331`.
- `group/media` / `left-2` vs `right-2` — the existing chevron test already pins the group name via its `group-hover/media:` prefix; side is unverifiable in jsdom and immediately visible if wrong.
- `flex` on the strip container and `h-full w-full` on the strip `<Media>` — present and correct; every remaining class in a new JSX literal is an unbounded regress.

The code lane found nothing else. It independently re-derived and cleared `preventFocusOnPress` (the `MouseEvent` import from `react` is load-bearing — without it the DOM global is picked up and the handler stops type-checking; cancelling `mousedown` does not cancel `click`; no ancestor `mousedown` listener exists near a post, and Radix dismissal uses `pointerdown`), confirmed the `preload` gate is correct for **every** caller in the codebase, compiled `app/globals.css` through `@tailwindcss/postcss` to confirm `.no-scrollbar` genuinely survives into `@layer utilities`, and decoded the test blurhash against the installed package (28 chars, 4x3 components, `decode` returns 4096 bytes).

Cumulative: **26 findings across four rounds**, all fixed and mutation-verified; ~14 further candidates refuted.

Round 5 will follow. Full local gate green on the fix: prettier, lint, typecheck, build, and 870 test files / 11,025 tests.

---

lib/components/posts/media.tsx
llun: Fixed in [03df7bf3](https://github.com/llun/activities.next/commit/03df7bf3). One assertion added to `media.test.tsx`'s `renders video with poster and focal point object position` — the one test that renders a postered video with `loading` unset:

```
expect(video).not.toHaveAttribute('preload')
```

with a comment saying why that case is the one that separates the two spellings: a poster alone must not defer the fetch, because only a strip item asks for that by passing `loading="lazy"`, and every other caller — the lightbox, which shows controls, and the lone-video branch — omits it and keeps the element's `metadata` default.

Mutation-verified: with `loading === 'lazy' &&` removed the test fails with `Received: preload="none"`, and passes once restored.

---

CLAUDE.md
llun: Fixed in [03df7bf3](https://github.com/llun/activities.next/commit/03df7bf3). The bullet now carries the qualifier, matching `AGENTS.md:1601-1607` and `REVIEW.md:558-563`:

> …so strip images pass `loading="lazy"` and strip videos `preload="none"` (`loading` is image-only) — but **only a video carrying a `poster`**, because a posterless one's sole pre-playback frame comes from the `#t=0.01` source fragment, which needs metadata to decode, so deferring one leaves a bare empty box, and federated video never has a poster (`thumbnailUrl` is written on the local-upload path alone) — while a lone picture or video stays eager…

Your reading of the history is right and worth recording: round 2 added this clause to `CLAUDE.md` first, so round 3's `AGENTS.md`/`REVIEW.md` edits were catching those two files *up* — the poster condition is the one thing that moved the other way and left `CLAUDE.md` behind.

All three docs now agree, and the code half of the same round-4 batch closed the mutation gap that would have let `&& poster` be dropped without a test objecting.

---

llun:
## Review loop round 5 — clean. The loop is closed.

Both lanes reported no defect, and both adversarial verifiers returned **NO SURVIVING FINDINGS**. That is the terminating condition in AGENTS.md, so the loop ends here at 5 of a maximum 20 rounds.

Round 5 deliberately did not chase the round-4 delta (which changed no production code — one test assertion and one doc line). The code lane was pointed back at what `dc92aa30` introduced and had been reviewed least recently, on the theory that a surviving defect was likelier there.

**What it cleared, so the ground covered is on the record:** hook order against both early returns and the memo dependency chain; `getMediaGeometry` at the clamp boundaries and for `NaN`/`Infinity`/negatives; the single-picture width expression's ordering; 2 vs 50 items and first paint before any measurement; every input that can change overflow in `useMediaStripScroll` (observer, `contentKey`, `onScroll` — there is no fourth); keyboard reachability, which needs no `role`/`tabIndex` on the scroller because focusing a picture scrolls it into view; that `mask-image` fades neither the chevrons nor a focused outline; and that `tailwind-merge@3.6` does not collapse either focus-indicator class pair (verified empirically, and the tests read the *rendered* className, so it is pinned rather than merely true today).

**Two candidates were raised and both were refuted**, one by the reviewer itself:

- *The 44px `MIN_MEDIA_WIDTH` floor is width-only*, so a 100×20 image lands in a 100×20 target. Refuted: it only breaches the WCAG 2.5.8 24px floor for a spacer-gif shape (`naturalHeight < 24` **and** `naturalWidth > 44`), and rendering a 100×20 image at 100×20 is arguably correct anyway.
- *The single-picture branch is not pinned against a companion audio attachment* — `pictures.length === 1 && !players.length` does survive the suite. Refuted on three independent grounds. The stated harm was wrong: the reviewer claimed the photo would be "cropped to a band, precisely the pre-PR bug", but a strip item's width is derived from the image's own clamped ratio, so an 800×600 photo gets a 320×240 box at exactly 4:3 and `object-cover` has nothing to trim — the real effect is a smaller image and spurious strip semantics. Every *standard* mutation operator on that line is already killed by `renders no scroll strip`; the survivor requires inserting a novel conjunct, which by that standard makes every branch in the codebase "uncovered". And the remedy is an assertion on an already-green test where the shipped code is correct — the class round 4's verifier refused, applied consistently here.

**Cumulative: 26 findings across rounds 1–4, all fixed and mutation-verified; ~14 candidates refuted.** All 23 threads resolved.

---

Separately, this branch had gone `mergeable_state: dirty` while round 5 ran. `main` has been merged in ([af95cfdc](https://github.com/llun/activities.next/commit/af95cfdc)). The only true conflicts were `AGENTS.md` and `REVIEW.md`, where main's new "Status delete federation" section landed at the same place as this branch's "Post media layout" section — both kept, since they document unrelated subsystems. `lib/types/domain/attachment.ts` and its test auto-merged cleanly: [#1541](https://github.com/llun/activities.next/pull/1541)'s `AttachmentMediaMetadata` and this branch's three predicates are additive to each other.

Worth flagging from that merge: **`main` took `lib/types/domain/attachment.test.ts` off the `tsconfig.typecheck.json` exclude ratchet**, so the tables this PR adds to that file are now typechecked — they pass. `post.test.tsx` remains excluded, so that half of the recorded "deliberately not fixed" item still stands.

Full gate green after the merge: prettier, lint, typecheck, build, and **873 test files / 11,057 tests**.

The PR description has been brought in line with what the loop changed — the load-bearing strip details are now four rather than three, the `preload` rule carries its poster condition, the mask rationale no longer claims a test cannot read it back, and the chevron and focus-indicator rules are stated as shipped.

---

PR #1553, opened 2026-08-26
fix: federate undo announce after the boost row is deleted

llun:
## Summary

**Unboosting never federated.** `userUndoAnnounce` hard-deletes the Announce row and *then* enqueues `SendUndoAnnounceJob`, which called `loadStatusAndActor` and bailed because the row it needed was already gone. Every remote instance kept showing a boost the author had removed, while locally it disappeared — so it looked like it had worked.

Under the default in-process queue this is not even a race: `publish` runs the job inline immediately after the delete, so it failed on **every** unboost. With QStash it failed just as reliably, only later.

This is the follow-up flagged in [#1536](https://github.com/llun/activities.next/pull/1536), which fixed the same pattern for status delete and documented the trap.

### The payload now carries what the activity needs

`undoAnnounce` builds its object from six fields of the Announce — `id`, `actorId`, `createdAt` (→ `published`), `to`, `cc`, and `originalStatus.id` (→ the boosted post). All six are captured before the delete, and the sender's `announce` parameter is narrowed to exactly that shape rather than a whole `StatusAnnounce`, so nothing has to fake a `Status` it can no longer load. That mirrors the `Pick<Status, 'to' | 'cc'>` narrowing #1536 applied to `getFederatedStatusDeliveryInboxes`, and is safe here because the job is the sender's only consumer.

### A second defect in the same function

The publish used `getHashFromString(status.id)` — the id `SendAnnounceJob` already publishes under. The queue deduplicates on that id **across job names**, in a window that outlives consumption, so a boost followed by an unboost inside the window dropped the `Undo` regardless of the payload. Fixing only the reload would have left this live. The id is now suffixed to match the emitted `<id>#undo`, exactly as delete uses `#delete`.

Inbox resolution is deliberately unchanged: `sendAnnounceJob` resolves inboxes with the same `getFollowersInbox` + `filterFederatedUrls` pair, so boost and unboost stay symmetric. Switching to the shared delivery helper would have widened the audience beyond this fix.

## Test plan

**The suite could not see any of this.** The old job test created the Announce with `database.createAnnounce` and *left it in the database*, so the job found what production had already deleted and passed against permanently broken code. A second test asserted the bail-out as intended behaviour, encoding the bug. Both are gone.

- **`lib/jobs/sendUndoAnnounceJob.test.ts`** — the delivery test now drives an Announce id that is deliberately **absent** from the database (the production condition) and asserts the real activity body: `<id>#undo`, `type: 'Undo'`, and the nested `Announce` with its `published`, `to`/`cc` and boosted `object`. The old test asserted only that *some* POST happened. Plus actor-not-found, a rejected legacy payload, and the `JOBS` registry entry.
- **`lib/actions/undoAnnounce.test.ts`** — the exact new payload, the `#undo` dedup id, an explicit assertion that it differs from `getHashFromString(status.id)` (the collision guard), and that the delete is ordered *before* the publish.

Each new guard was verified to **fail against a mutant** of the code it covers — restoring the reload-and-bail, restoring the bare dedup id, unregistering the job, and swapping the delete/publish order each failed exactly one test and no others.

Full suite green: 873 files, 11,007 tests.

## Checklist

- [x] Docs updated: `AGENTS.md` / `REVIEW.md` — the existing **Status Delete Federation** sections described this bug as live, so they are extended to cover unboost rather than duplicated
- [x] If migrations changed: n/a — no schema changes, payload-only

https://claude.ai/code/session_01REnirYGfFfThi1SqsFX9zw

---
_Generated by [Claude Code](https://claude.ai/code/session_01REnirYGfFfThi1SqsFX9zw)_


llun:
## Sub-agent review round 1

Three reviewers (federation correctness, types/payload, tests/docs), each finding then attacked by two independent skeptics. Three findings survived verification — all three are addressed in `c58a957f`, and the verifiers reproduced each one with their own mutation experiments rather than taking the reviewer's word for it.

### Two tests that could not fail

This is the same class the previous PR's review caught, and it recurred in tests **this PR rewrote**. My own mutation pass covered the guards I *added* and missed the pre-existing guards these tests claimed to cover — I confirmed both by removing the guard and watching the suite stay green:

| Test | Why it was hollow | Fix |
| --- | --- | --- |
| `does nothing if actor is not found` | An unknown actor has no followers either, so `getFollowersInbox` returns `[]` and the fan-out is empty **with or without** the `!actor` guard. The assertion was satisfied by the empty follower list, not by the guard. | Now spies on `getFollowersInbox` — the call the guard exists to prevent. |
| `returns null when status is not owned by the current actor` | It targeted `announce-1`, which the *first* test in the file hard-deletes. The action short-circuits on the missing row, so the ownership branch was never reached. | Creates an Announce that still exists, and additionally asserts a non-owner cannot delete it. |

The second one matters beyond hygiene: the job's comment now explicitly leans on that ownership guard (it dropped its own `status.type` check), so it was the one thing standing between a signed-in actor and unboosting someone else's post.

### A regression this fix introduced

The reviewers split on whether the missing `try/catch` around `publish` was pre-existing or new, and the distinction is real. The bare `await publish` is unchanged from `main` — but **what it can throw is not**. Pre-fix the job returned at `!status` before touching `getFollowersInbox` or `filterFederatedUrls`; post-fix both run synchronously inside `publish` under the default in-process queue, after the Announce row is already hard-deleted. A pool timeout there now 500s an unboost that has already happened locally.

So it is wrapped now, matching `deleteStatusFromUserInput`, with a test that fails if the wrapper is removed. The `AGENTS.md` bullet asserting that invariant is true for both paths as a result — it was flagged as inaccurate when it described only delete.

### Verified, not assumed

All six guards were mutation-tested — reload-the-deleted-status (the original bug), the `!actor` guard, the ownership check, the bare dedup id, the post-commit `try/catch`, and the `JOBS` registry entry. Each fails exactly one test and no others.

Local suite: 873 files, 11,008 tests, green.

---

PR #1556, opened 2026-08-26
fix: cache outbox totalItems and let shared caches hold the collection

llun:
## What went wrong

Cloud SQL Query Insights flagged query hash `1282900130835943839` averaging **899 ms over 560 calls** in a 75-minute window on 2026-08-25. That query is the ActivityPub outbox collection's `totalItems` — `getActorStatusesCount({ publicOnly: true })`, a recursive CTE over the actor's entire public history.

Its real cost is ~100–160 ms of CPU for an actor with ~1,900 public statuses. The 899 ms was **CPU queueing, not more work**:

- **530 outbox root fetches landed in 11 minutes** (09:35–09:46 CEST), in sub-second clusters. Each recomputed the count from scratch; nothing deduplicated it.
- Query Insights showed this one query demanding **~4 CPU-sec/sec against a CPU capacity of 1**. All "CPU and CPU wait" — zero IO wait, zero lock wait. Mid-flood, one request took **6.84 s** end-to-end at Cloud Run.
- The sampled plans inside the slow window are identical to the fast ones, which rules out a plan regression and points at contention.

The response also sent no `Cache-Control`, so the CDN forwarded every one of those requests to origin.

## The fix

**A 60 s per-actor cache that holds the query's promise** (`lib/services/statuses/actorPublicStatusesCount.ts`). Caching the promise rather than the resolved count is what makes it bound *both* halves of a burst: the TTL bounds requests arriving after a count is computed, and a request arriving while one still runs joins it instead of starting its own. A stored counter remains impossible for this value — an Announce is publicly readable only while the status it boosts still is, and that status belongs to another actor — but a count that lags by up to a minute is within what the collection already promises, since that same argument is why it can never be exact.

This is the shape `getMySQLFullTextMinTokenSize` (`lib/database/sql/search/documents.ts`) already used, down to deleting the entry on failure. Bounded at 512 actors, keyed `WeakMap<Database, Map<actorId, …>>`, errors never cached.

**`Cache-Control: public, max-age=60, s-maxage=60` on the root**, as a third layer that is deliberately not trusted as the only one — see the correction below. The root also **`Vary`s on `x-activity-next-host`, `x-forwarded-host`, `Host` and `Origin`**: `headerHost` trusts the two custom headers ahead of `Host`, so on a multi-domain instance the same path serves a *different actor* per header, and a shared cache that forwarded them without keying on them would serve one domain's collection for another's.

**The `?page=true` branch sends an explicit `Cache-Control: no-store`.** Its `orderedItems` reflects live per-status visibility, so a status hidden mid-window must stop being served at once — and omitting a header asserts nothing, it just defers to whatever the cache in front does with an unlabelled 200.

`activityPubResponse` grew an `additionalHeaders` pass-through for this. It appends, which cuts both ways: `Vary` is a list header so a caller's entry correctly joins the `Accept` it already sends, while `Content-Type` is single-valued and a second one would corrupt it.

## Two claims from my first draft that review disproved

Worth stating plainly, since both were load-bearing:

1. **"No single-flight dedup — the CDN absorbs the burst."** It does not. llun.dev's CloudFront policy keys on `Signature` and `Date`, which differ per request, so every *signed* server-to-server fetch — what federation actually sends — takes its own cache entry and collapses into nothing. An unauthenticated caller can vary an ignored query parameter to the same effect. The origin has to survive a burst alone. (The general form is what's durable: whether a shared cache collapses a burst is a property of its configuration, not of this response.)
2. **"No precedent for promise-caching in this codebase."** `documents.ts` had been doing exactly that for hundreds of commits. Adopting it removed a second map, a `.finally`, and an argument about microtask ordering.

## Review

Five rounds of sub-agent review (correctness, scope/regression, testing, quality, plus a security pass on the shared-cache decision). It found, and this PR fixes:

- A **cold-start race** that silently dropped a concurrently-computed entry for another actor — the map was read before the query and written after, so two callers on a cold database each built their own and the later write replaced the earlier. Independently reproduced with realistic query latency.
- The `Vary` gap and the page branch's reliance on an *absent* header (security lens).
- **Eviction order** and the **touch branch**, both unproven: mutations that evicted the wrong key passed the entire suite.
- An **untested identity guard** — deleting the failed entry unconditionally, which would let a late rejection evict the good entry that replaced it, passed all 14 tests.

Every fix is pinned by a test proved non-vacuous by reverting the thing it guards. One correction worth recording: my first attempt at the touch-branch tests refreshed the *oldest* entry, which is degenerate — an implementation that evicts the oldest and re-inserts the refreshed actor lands on the same answer. They now refresh a middle entry.

**Gate**: prettier, lint, typecheck, build, and **879 test files / 11,293 tests** passing, over two consecutive clean full-suite runs.

## Deployment note (manual, AWS side — not in this repo)

The `s-maxage` only takes effect if the CloudFront behavior covering `/users/*` permits caching, and its cache key must include the headers now named in `Vary` — this route negotiates content type *and* resolves its actor from request headers. Note the current policy's `Signature`/`Date` keying means the CDN will not collapse signed federation fetches regardless; that is why the origin-side dedup is the load-bearing half.

Separately: the flood's true client can only be identified from CloudFront access logs — the origin sees only `userAgent: "Amazon CloudFront"` and edge IPs.


llun (commented):
Automated sub-agent review: 5 confirmed findings posted inline.

lib/services/statuses/actorPublicStatusCount.ts
```diff
@@ -0,0 +1,103 @@
...
+  const entries = cacheByDatabase.get(database)
+  const entry = entries?.get(actorId)
+  if (entry && entry.expiresAt > Date.now()) return entry.count
+
+  const count = await database.getActorStatusesCount({
+    actorId,
+    publicOnly: true
+  })
+
+  const actorEntries = entries ?? new Map<string, CacheEntry>()
```
llun: **Cold-start race drops a concurrently-computed entry for another actor** (real bug, low blast radius)

`const entries = cacheByDatabase.get(database)` is read synchronously BEFORE the `await`, and reused after it. When the WeakMap has no value yet for this database — the cold-start window this cache exists to survive — two concurrent calls for DIFFERENT actorIds both see `undefined`, each builds its own `Map`, and the later `cacheByDatabase.set` replaces the earlier one wholesale. The dropped actor's entry is gone entirely, so its next request recomputes even seconds later, well inside the TTL.

Independently reproduced with realistic (setTimeout-based, ~120ms) query latency: `Promise.all` of actors A and B on a cold cache leaves the cache with 1 entry instead of 2, and a call for B staggered 30ms after A still drops one — the window is the full query duration, not a same-tick coincidence. Once any entry exists for that database the race cannot reoccur, and `getDatabase()` is `memoize()`-wrapped so the object is a true process singleton, making this reachable by two real concurrent requests on a fresh container.

Nothing wrong is ever SERVED — both callers get their own correct count — the cost is one lost cache entry. This is a different mechanism from the documented "no single-flight dedup" tradeoff, which covers same-actor redundant work, not cross-actor entry loss.

Fix: get-or-create and `cacheByDatabase.set` the Map synchronously before the `await`, so every racing caller mutates one shared Map.

lib/services/statuses/actorPublicStatusCount.ts
```diff
@@ -0,0 +1,103 @@
...
+// status it boosts does, and that status belongs to another actor whose edits
+// and deletes never touch this actor's rows. So it is a recursive CTE over the
+// actor's whole public history, ~100–160ms of CPU for an actor with ~1,900
+// public statuses.
+//
+// Nothing above it deduplicated the work. On 2026-08-25 a burst of 530 outbox
+// root fetches in 11 minutes each recomputed the count and queued behind one
+// another on a 1-vCPU instance, taking the query's average wall time to 899ms
+// against ~150ms of real work — the executions were waiting for CPU, not doing
+// more of it. This bounds the recomputation to once per actor per TTL, and the
```
llun: **Comment claims a guarantee the code does not provide** (inaccurate comment)

There is no in-flight/single-flight dedup, so two concurrent misses for the SAME actor both run the query — the cache is written only after the query resolves. "Bounds the recomputation to once per actor per TTL" is therefore not literally true under exactly the burst this cache exists for. The AGENTS.md paragraph hedges this correctly ("no single-flight dedup … a sub-TTL stampede is a different, unmeasured failure mode"); the in-file comment should not overstate it.

lib/services/statuses/actorPublicStatusCount.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    const { database } = mockDatabase(1)
+
+    for (let index = 0; index < MAX_CACHED_ACTORS + 88; index++) {
+      await getCachedActorPublicStatusCount(
+        database,
+        `https://example.com/users/actor-${index}`
+      )
+    }
+
+    expect(getActorPublicStatusCountCacheSizeForTests(database)).toBe(
```
llun: **Eviction ORDER is unproven — only final size is asserted** (test gap)

Mutating `setBoundedEntry` to evict the second-to-last-inserted key instead of the true oldest (`entries.keys().next().value`) leaves all 8 tests green, because the test only checks the map didn't grow past the bound, never WHICH actor was evicted. A wrong-order eviction regression would ship silently. Assert that the oldest actor is the one gone and a recent one survives.

lib/services/statuses/actorPublicStatusesCount.ts
```diff
@@ -0,0 +1,103 @@
...
+
+// Insertion-order eviction, mirroring `setBoundedCacheValue` in
+// `lib/utils/host.ts` — that one is module-private to an unrelated domain, so
+// this is a deliberate second copy rather than an export widened for one caller.
+const setBoundedEntry = (
+  entries: Map<string, CacheEntry>,
+  actorId: string,
+  entry: CacheEntry
+) => {
+  if (entries.has(actorId)) {
```
llun, line 75: **The touch branch and its interaction with the size bound are untested** (test gap)

The `entries.has(actorId)` delete-then-reinsert branch exists so refreshing an already-cached actor moves it to the newest position rather than being evicted for age. Two mutations prove nothing guards it: (a) removing the branch entirely so a touch never re-orders — all 8 tests pass; (b) making the size-based eviction fire unconditionally, so refreshing an existing entry at full capacity spuriously evicts one EXTRA unrelated actor — all 8 tests pass. No test combines "cache is at MAX_CACHED_ACTORS" with "refresh an existing entry"; the bound test only inserts new ids and the TTL tests only use 1-2 actors.

lib/services/statuses/actorPublicStatusCount.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * `database.getActorStatusesCount({ actorId, publicOnly: true })` behind a
+ * per-actor TTL cache.
+ *
+ * Errors are not cached: a rejection propagates before anything is written, so
+ * the next caller retries rather than being served a failure for a minute.
+ */
+export const getCachedActorPublicStatusCount = async (
```
llun: **Name drops the 'es' its wrapped method and modeled sibling both keep** (minor naming)

This wraps `database.getActorStatusesCount` (plural) and is modeled on `getCachedLocalPublicStatusesCount`, which preserves its wrapped method's name verbatim under the `Cached` prefix. `getCachedActorPublicStatusCount` is singular, breaking that convention and reading like a typo beside `getActorStatusesCount` in the same route file. Suggest `getCachedActorPublicStatusesCount`.

lib/services/statuses/actorPublicStatusCount.ts
llun: Fixed in 63952c877. Extracted `getActorEntries(database)`, which does the get-or-create AND the `cacheByDatabase.set` synchronously, before the `await` — so every caller racing on a cold database key mutates the one published map instead of each building its own. Publishing an empty map early is free: only `setBoundedEntry` puts anything in it, and a reader that finds it empty simply misses. Pinned by a new test, `does not drop a concurrently cached actor on a cold database`, which uses `Promise.all` deliberately — serialized calls cannot reproduce the bug, since the second would read the map the first already published. Reverting the fix fails it with `expected length 2, got 1`.

lib/services/statuses/actorPublicStatusCount.ts
llun: Fixed in 63952c877. The comment now says the cache holds each actor's count for a TTL, and states plainly what it does not do: there is no in-flight dedup, so concurrent misses for one actor all run the query, because the entry is only written once the query resolves. It also names the CDN as what actually absorbs a burst, matching the AGENTS.md wording.

lib/services/statuses/actorPublicStatusCount.test.ts
llun: Fixed in 63952c877. The test-only size accessor is now `getActorPublicStatusesCountCachedActorIdsForTests`, returning the cached ids in insertion order, so a test can assert WHICH actor went rather than only that the map did not grow. New test `evicts the actor cached longest ago, not an arbitrary one` asserts the oldest is gone, a newer one survives, and the just-inserted actor is last. Verified against your exact mutation (`[...entries.keys()][1]`), which now fails two tests, and also against evicting the newest and against an off-by-one bound.

lib/services/statuses/actorPublicStatusesCount.ts
llun: Fixed in 63952c877 — and you were right that it needed care. My first attempt refreshed actor-0, the OLDEST entry, and it was vacuous for exactly the reason you would expect: an implementation that evicts the oldest and then re-inserts the refreshed actor lands on the same answer, so both your mutations still passed. Both tests now refresh an actor from the MIDDLE of the map (actor-3), with a comment recording why. `refreshing an expired actor at capacity evicts nobody` catches both mutations — removing the touch branch, and evicting the oldest before the touch delete — and `refreshing an expired actor moves it clear of the next eviction` covers the reordering.

lib/services/statuses/actorPublicStatusCount.ts
llun: Fixed in 63952c877. Renamed to `getCachedActorPublicStatusesCount`, with the module, the TTL constant and both test-only exports following, so the plural now matches both `database.getActorStatusesCount` and `getCachedLocalPublicStatusesCount`, which this is modelled on. The file moved via `git mv` so history follows; AGENTS.md and CLAUDE.md were updated in the same commit and a grep for the old singular names returns nothing.

PR #1517, opened 2026-08-23
minor: support blurhash and smart focus for images and video thumbnails

llun:
## Summary

Adds comprehensive support for **blurhash** placeholders and **smart focal points** across all image formats and video preview frames.

### Key Changes
1. **Coordinate Math & Image Analysis (`lib/utils/focalPoint.ts`, `lib/services/medias/imageAnalysis.ts`)**:
   - Computes 4x3 blurhash strings using `blurhash`.
   - Auto-detects focal point using Sharp's attention saliency strategy (`position: sharp.strategy.attention`).
   - Normalizes coordinates to [-1.0, 1.0] and converts to CSS `object-position` percentage values.
   - Wraps analysis in error-isolated decoration logic (analysis failures never block media uploads).
2. **Storage Drivers (`localFile.ts`, `S3StorageFile.ts`)**:
   - Hooks image analysis and video preview frame extraction into `saveFile`, `saveThumbnail`, and presigned upload completion.
3. **Database & Migrations (`migrations/`)**:
   - Added migration `20260823123348_add_media_blurhash_and_attachment_focus.js` adding `blurhash` to `medias`, and `blurhash`, `focusX`, `focusY`, `thumbnailUrl` to `attachments`.
   - Regenerated both `migrations/schema.sqlite.sql` and `migrations/schema.sql` in lockstep.
4. **Federation & Serialization (`createNote.ts`, `createNoteJob.ts`, `attachment.ts`, `getMediaAttachment.ts`)**:
   - Emits `toot:blurhash` and `toot:focalPoint` in ActivityPub JSON-LD Note attachments.
   - Validates inbound Document `blurhash` and `focalPoint` before storing in database.
   - Emits `blurhash` and `meta.focus` in Mastodon API attachments (`getMastodonAttachment`, `getMediaAttachment`).
   - Updates `PUT/PATCH /api/v1/media/:id` to handle focal point updates and thumbnail replacement with blurhash recomputation.
5. **Frontend Rendering (`BlurhashCanvas.tsx`, `media.tsx`)**:
   - Decodes blurhash placeholders onto `<canvas>` element behind loading images with smooth fade-in.
   - Applies `objectPosition` matching the focal point.
   - Sets video `poster` to `thumbnailUrl`.
6. **Maintenance & Documentation**:
   - Added `scripts/maintenance/backfillMediaBlurhash.ts` backfill script.
   - Updated `docs/features.md` and `docs/maintenance.md`.

## Checklist

[...]

PR #1559, opened 2026-08-26
fix: harden the blurhash backfill script's URLs, downloads and --force

llun:
## Summary

Hardens `scripts/maintenance/backfillMediaBlurhash.ts` against four verified issues found reviewing the blurhash work (#1517 / #1526 / #1528). Three of them are things the script got wrong; the fourth is a promise the docs made that the code never kept.

### 1. It wrote a host-relative `thumbnailUrl` where every live path writes an absolute one

`thumbnailUrl` is served to clients verbatim as Mastodon's `preview_url` (`getMastodonAttachment`, three branches) and as a `<video>` poster, with no absolutisation anywhere in between. A native client that is not talking to this origin cannot load `/api/v1/files/…` — the web UI only appeared to work because it is same-origin.

The script now builds it with `getMediaFileUrl`, the same helper `getAttachmentMediaMetadata` uses on the live path, and on the same host that path uses: **the owning actor's domain**, not the configured primary host, so a multi-domain instance mints the right origin. `getAttachmentMediaHost` falls back to the configured host rather than reusing `getActorDomain`, which echoes a non-URL `actorId` back verbatim and would mint `https://<garbage>/api/v1/files/…`.

### 2. Unbounded, unguarded fetches over remote-actor-supplied URLs

The #1528 addition was a bare `fetch(row.url)` + `arrayBuffer()`. Those URLs come off notes federated to us (`createNoteJob` inserts `url: attachment.url` straight off an inbound Note), so this was stored-data SSRF reachable from inside the deployment, with no size cap and no timeout.

Now: an SSRF URL guard before the request — **re-run on every redirect hop** — a 10s `AbortSignal.timeout`, an `image/` content-type check, and a 10 MB body cap via `readResponseArrayBufferWithLimit`. A new `--local-only` flag skips remote downloads entirely.

The redirect part is not incidental. The first version of this PR guarded only the URL it was handed, and `fetch` defaults to `redirect: 'follow'` — so a public host answering `302` with `Location: http://169.254.169.254/…` sent the request somewhere the guard never saw, and Node applies no mixed-content rule, so https could redirect to plain http on loopback. Review round 1 caught it; reproduced against two loopback servers, where the guard approved the first URL and the bytes came back from the internal service. `safeImageFetch` now takes each hop by hand and re-guards the target, capped at three hops like `safeRemoteFetch`. The storage-redirect read in `getFileBuffer` is capped too, at `PRESIGNED_ANALYSIS_MAX_BYTES` — the same ceiling the S3 driver analyses uploads under. It needs no SSRF guard (it is our own presigned URL) but it was still an uncapped read into memory.

`safeRemoteFetch` is deliberately **not** used here: its result body is `Buffer.concat(chunks).toString('utf8')`, which mangles image bytes. That is why the Strava photo path hand-rolls a URL guard and reaches for a plain `fetch`. Rather than write that guard a second time, this PR **extracts it** to `lib/utils/safeImageDownload.ts` and points both callers at it. The move is verbatim — same blocklist, same order of checks — and it gains the test coverage it never had (that guard had zero tests before this PR).

Sharing it also closes **the same redirect hole in the Strava photo path**, which was pre-existing and which this PR would otherwise have left for the next caller to inherit. The consequences are worse there: those bytes reach `saveMedia` and become a real stored attachment, not a 30-character placeholder.

One accepted residual, documented in the module rather than closed: the guard resolves the hostname and `fetch` resolves it again, so a DNS record flipping between the two wins the race. `safeRemoteFetch` pins the resolved address into its transport; closing this the same way would mean a binary-mode entry point on `createSafeRemoteFetch`, which is more than these two callers justify. A rebind needs a race; a redirect needed only a header.

### 3. `--force` did not do what `docs/maintenance.md` said

For attachments, `--force` only dropped `whereNull('blurhash')` from the query while every per-row step still filled a *missing* value (`if (media.blurhash && !blurhash)`, `if (!blurhash && …)`). "Recompute wrong hashes" updated **zero** rows.

Fixed in the code for the blurhash, and in the docs for the focal point:

- `--force` now genuinely recomputes the blurhash. A blurhash is purely derived, so overwriting one is safe.
- It still only ever **fills in** a focal point. `PUT /api/v1/media/:id` lets an owner set one by hand and no column records whether a stored point was set that way, so recomputing would silently discard the owner's choice. The docs now say so instead of promising otherwise.

While restructuring `backfillMedias` for this I also fixed a bug I had introduced one commit earlier: gating the whole update on the blurhash having changed meant a `--force` pass over a row with a correct blurhash and a *missing* focal point stopped filling the focal point. Updates are now collected field by field, and a row where nothing changed is no longer counted as updated.

### 4. `url.includes('/api/v1/files/')` matched other activities.next instances

`/api/v1/files/` is this project's own route, so **every other activities.next instance** serves attachment URLs under exactly that path. A substring match called those local storage paths, looked them up in our storage, missed, and skipped the row — and it let a remote actor choose the path string we handed to storage (`…/api/v1/files/../../secret.jpg`).

Now the URL is parsed and its host matched against this instance's own — `ACTIVITIES_HOST` plus `ACTIVITIES_TRUSTED_HOSTS` — through the app's own `normalizeHost`/`hostMatchesRule` pair, so a wildcard entry like `*.example.com` matches exactly as it does for a request `Host` header. (The first version used a `Set` of literals and `.has()`, which silently disowned every wildcard-covered subdomain; review round 1 caught that too.) `isHostTrustedByRules` itself is not reused because `normalizeHost` deliberately answers null for loopback names, which would stop a dev instance on `localhost:3000` recognising its own URLs — so a loopback authority falls back to literal equality.

The path is taken from `pathname` rather than by substring, so `https://remote.example/redirect?to=/api/v1/files/x.jpg` no longer matches at all. A decoded path that walks upwards is refused rather than handed to storage: `new URL().pathname` normalises a literal `../` away, but a percent-encoded `%2e%2e%2f` survives it and decoding puts it back.

## Also

- The silent `catch {}` around per-attachment analysis now warns with the row id; every hardening rejection was previously invisible.
- `main()` fails fast when `ACTIVITIES_HOST` is unset, since absolute URLs cannot be built without it.

## Docs

`docs/maintenance.md` — corrected the `--force` description, documented `--local-only`, and added sections on what `--force` does and does not recompute and on how remote attachment URLs are treated.

## Testing

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test`, all green (880 files / 11,462 tests).

New coverage, all **proved by reverting the subject** — and two of the first round's tests turned out not to be, which is why the table is longer now:

| Revert | Tests that fail |
| --- | --- |
| `redirect: 'manual'` → `'follow'` | 1 |
| the hop loop (guard hop 0, then follow) | 9 |
| the redirect target not re-guarded | 4 |
| the 3-hop budget | 1 |
| wildcard host matching → literal equality | 1 |
| decoded-traversal rejection | 3 |
| IPv6 bracket stripping | 3 |
| absolute `thumbnailUrl` → relative | 1 |
| `--force` → fill-only | 1 |
| focus fill gated behind a blurhash change | 1 |
| host-checked path → substring match | 4 |
| the `getFileBuffer` redirect branch | 3 |
| the direct-analysis write block | 2 |
| the local-storage-miss warning | 1 |

Two tests from the first round were vacuous and are rewritten. One had `../../` in its fixture URL, which `new URL().pathname` normalises away *before* the host check it claimed to prove — so it returned null for an unrelated reason and passed with that check deleted. The other drove `fetch` directly rather than through `safeImageFetch`, so it passed with `redirect: 'manual'` reverted — the single most important line in the PR. Both were caught by revert-proofing, not by reading.

The `getFileBuffer` redirect branch and the entire direct-analysis write path had **no** coverage: each could be deleted whole with all 39 tests still green. The S3 driver answers with a redirect whenever a CDN hostname is configured, so that branch is the only way an object-storage deployment reads a file at all.

`scripts/maintenance/backfillMediaBlurhash.test.ts` grows from 1 execution test to 70, and `lib/utils/safeImageDownload.test.ts` adds 106 for the extracted guard.

## Review

Four rounds so far, each pointed at the previous round's fixes rather than the original diff. **Every round found a real defect in the round before it**, and all but one were in code this PR added.

- **Round 1** (5 lenses) — 11 findings, posted as threads on this PR. The headline: `fetch` defaults to `redirect: 'follow'` and the guard only inspected the URL it was handed, so a 302 to a private address defeated it. Reproduced against two loopback servers. Two of the round-0 tests were also vacuous.
- **Round 2** (4 lenses) — 5 findings, all in round 1's fix. Stripping IPv6 brackets handed the decision to a `BlockList` that had no NAT64/6to4/Teredo/ORCHID entries; and `new URL` parses `*` in an authority, so the loopback fallback compared against a wildcard rule's own spelling and read attacker-chosen paths out of local storage.
- **Round 3** (2 lenses) — 3 findings, all in round 2's fix. The `BlockList` copy blanket-blocked NAT64 where the audited policy decodes the embedded IPv4 (an availability break on DNS64 deployments), missed two ranges, and the protocol-relative check was defeated by `/\host` and `/<TAB>/host`, which the WHATWG parser turns into authorities.
- **Round 4** (4 lenses + an adversarial verifier on every finding) — 13 raw, **10 refuted**, and the only real one was a test of mine that was a tautology. No defect in shipped behaviour. It independently verified the extraction: byte-identical modulo three `export` keywords, zero policy disagreements across 777 addresses, and the protocol-relative check survived 184,324 attack forms.
- **Round 5** (4 lenses, fresh angles) — 10 raw, 8 refuted, 2 real. The important one: **the relative-URL fix wasn't reaching the data it exists for.** The pre-PR script set the blurhash and a relative `thumbnailUrl` in the same pass, so `whereNull('blurhash')` never revisited those rows — an operator upgrading and re-running would have repaired none of them. A normal run now selects and rewrites them. The other: a cap test that OOM-killed the Vitest worker instead of failing an assertion.
- **Round 6** (3 lenses, strict actionability bar) — **clean**. Two findings judged real but pre-existing and out of scope, filed as a follow-up.

From round 4 on, every finding was routed through a verifier whose job was to *refute* it. That killed 18 of 23 — including several that read as serious until someone tried to reproduce the consequence rather than the evidence.

### What round 3 changed structurally

Three rounds in a row found the same underlying cause: this branch had grown a **second, hand-rolled implementation** of "is this IP address safe", expressed as a `BlockList`, which diverged from the audited one in `safeRemoteFetch` in five places. Rather than add the two missing ranges and wait for round 4 to find a sixth, the policy moved to `lib/utils/unsafeAddress.ts` and **both guards import it**. `safeRemoteFetch`'s own 25 tests pass unchanged, proving the move is behaviour-preserving.

## Deferred, filed separately

Deleting a `medias` row leaves `attachments.mediaId` dangling (`deleteMediaByConditions` never nulls it). This script then selects such a row but can never repair it — the rewrite lives inside `if (media)` — counting it in `processed`, never in `updated`, with no warning. Left out because the pattern pre-exists (video and PDF attachments already behave this way under the old predicate), nothing wrong is written, and the delete path destroys the stored bytes anyway. It needs a decision about the root cause, not a patch here.

## Follow-up filed separately

Two more copies of this URL-parsing logic exist and both still lack the host check.

`lib/utils/getAttachmentMediaPath.ts` is the one that matters: it is in `lib/`, used by four production files (`status.ts:397`, `processFitnessFileJob.ts:634`, `regenerateFitnessMapsJob.ts:244`, `mapAttachments.ts:31`), and does a bare `url.indexOf(marker)` with no host check at all — weaker even than the script versions, since it matches the marker anywhere in the string including inside a query string. Its callers may not have the instance host to hand, so fixing it is its own change rather than a widening of this one.

`getMediaStoragePathFromUrl` in `scripts/backup/actorArchive.ts:167` has the same host-check bug as issue 4 — its doc comment claims it returns null for foreign URLs but it only checks the pathname, so a remote instance's attachment is mapped into the archive as a local path that is never fetched. Its existing "foreign host URL" test uses a foreign host with a *different* path, so it does not catch this. Left out of this PR because it is a different script with a different symptom and needs its own tests.


llun (commented):
Round-1 review findings against a47f5d58a. One finding could not be anchored inline and is included here instead of being dropped:

**lib/jobs/importStravaActivityJob.ts — `const photoResponse = await fetch(photoUrl)` (line 193 at a47f5d58a)**

The Strava photo path has the identical redirect hole as the backfill script's `downloadRemoteImage`: a bare `fetch` with default `redirect: 'follow'`, guarding only the initial URL via `getSafeImageDownloadUrl`. Pre-existing rather than introduced by this PR, but this PR promoted its companion guard into a shared module, so the next caller would inherit the hole. Consequences are worse on this side than in the backfill script: these bytes reach `saveMedia` and become a real stored attachment, not a 30-character blurhash.

Fixed in the follow-up commit on this branch.

(This line sits outside the diff hunk touched by this PR, so it can't be anchored as an inline comment — recorded here instead.)

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -47,7 +81,154 @@ export const parseArgs = (args: string[]): CliOptions => {
...
+export const downloadRemoteImage = async (
+  rawUrl: string
+): Promise<Buffer | null> => {
+  const safeUrl = await getSafeImageDownloadUrl(rawUrl)
+  if (!safeUrl) {
+    console.warn(`Skipping unsafe or non-HTTPS attachment URL: ${rawUrl}`)
+    return null
+  }
+
+  const response = await fetch(safeUrl, {
```
llun: **Redirect-based SSRF: the guard validates hop 0, fetch follows the rest unchecked**

`fetch` defaults to `redirect: 'follow'`, and `getSafeImageDownloadUrl` inspects only the URL it is handed — so a public host answering `302` with `Location: http://169.254.169.254/…` sends the request somewhere the guard never saw. Node applies no mixed-content rule, so https can redirect to plain http on loopback.

Reproduced with two loopback servers: the guard approved the first URL and the bytes came back from the internal service.

Fully reachable: `lib/jobs/createNoteJob.ts:240` stores `attachment.url` verbatim off a federated Note, and omitting `blurhash` puts the row squarely in this script's own `whereNull('blurhash')` selection. Capability gained is an arbitrary attacker-chosen GET against any internal host, plus port scanning by timing and low-bandwidth exfiltration where the response decodes as an image.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -47,7 +81,154 @@ export const parseArgs = (args: string[]): CliOptions => {
...
+ */
+export const buildInstanceHosts = ({
+  host,
+  trustedHosts
+}: {
+  host: string
+  trustedHosts?: readonly string[] | null
+}): InstanceHosts => {
+  const fallbackHost = canonicalAuthority(host)
+  const ownHosts = new Set(
```
llun: **Wildcard `ACTIVITIES_TRUSTED_HOSTS` entries never match**

`ownHosts` is a `Set` of literal canonicalised strings queried with `.has()`, so a `*.example.com` entry is stored literally and never matches `tenant.example.com`. `ACTIVITIES_TRUSTED_HOSTS` supports wildcards everywhere else in the app (`lib/utils/host.ts` `hostMatchesRule`, tested at `lib/utils/host.test.ts`).

Effect on a wildcard-scoped multi-domain deployment: a genuinely local attachment is misclassified as remote and re-downloaded over the public network — or, under `--local-only`, skipped forever with no error. It also made the new `docs/maintenance.md` sentence about `ACTIVITIES_TRUSTED_HOSTS` false.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -47,7 +81,154 @@ export const parseArgs = (args: string[]): CliOptions => {
...
+  ownHosts: ReadonlySet<string>
+): string | null => {
+  const pathname = getOwnPathname(rawUrl, ownHosts)
+  if (!pathname?.startsWith(MEDIA_FILE_URL_PATH)) return null
+
+  const encodedPath = pathname.slice(MEDIA_FILE_URL_PATH.length)
+  if (!encodedPath) return null
+
+  try {
+    return decodeURIComponent(encodedPath)
```
llun: **Percent-encoded traversal survives URL normalisation and is handed to storage**

`new URL().pathname` normalises a literal `../` away, but a percent-encoded `%2e%2e%2f` survives it and `decodeURIComponent` turns it back into `../../`. The host-relative branch did no normalisation at all, so `/api/v1/files/../../../../etc/passwd` reached `storage.getFile` verbatim.

Contained by `LocalFileStorage.getFile`, which resolves against the storage root and rejects an escaping path — but the pre-change code passed the still-encoded form, so this was a regression introduced by adding the decode.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -47,7 +81,154 @@ export const parseArgs = (args: string[]): CliOptions => {
...
+  }
+}
+
+/**
+ * Downloads a remote attachment image for analysis.
+ *
+ * The URL comes off a note a remote actor federated to us, so it is untrusted
+ * input reached from INSIDE the deployment: the guard rejects a URL naming the
+ * local network before the request is made, the content type must claim an
+ * image, and the body is capped. Returns null (never throws) so one hostile row
```
llun: **Doc comment claims "never throws"; the function throws, and the PR's own test asserts it**

`readResponseArrayBufferWithLimit` raises `StreamByteLimitError` past the cap, and the new test asserted `.rejects.toThrow(/exceeds byte limit/)` — directly contradicting the comment. Behaviour was safe because the caller wrapped it, but a reader taking the contract at face value, or a refactor that unnests the call, would be wrong.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -215,39 +413,52 @@ export const backfillAttachments = async (
...
-          if (filesIndex !== -1) {
-            const targetPath = row.url.slice(
-              filesIndex + '/api/v1/files/'.length
-            )
+          const targetPath = getLocalStoragePath(
+            row.url,
+            instanceHosts.ownHosts
+          )
+          if (targetPath) {
             buffer = await getFileBuffer(storage, targetPath)
```
llun, line 563: **A local-storage miss is silent, unlike the identical case in `backfillMedias`**

When the URL resolves as ours but `getFileBuffer` returns `null`, nothing is logged and the row moves on. `backfillMedias` warns `Could not read file buffer at …` for exactly this. The PR's own stated goal — that every hardening rejection is visible — was only half delivered: the remote path warns on every rejection, the local-miss path stayed silent, so an operator cannot tell "nothing needed doing" from "the file is gone".

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -58,10 +239,19 @@ export const getFileBuffer = async (
   if (!fileOutput) return null
   if (fileOutput.type === 'buffer') return fileOutput.buffer
   if (fileOutput.type === 'redirect') {
+    // Our own presigned URL rather than untrusted input, so no SSRF guard — but
+    // still capped, at the same ceiling the S3 driver analyses uploads under.
     const res = await fetch(fileOutput.redirectUrl)
```
llun: **The storage-redirect fetch has no timeout and no redirect policy**

Unlike the remote branch's 10s budget, this fetch had none — a stalled CDN hangs the sweep indefinitely. It also left `redirect` at its default, so a CDN that can be made to redirect gives a second unguarded hop. This URL is meant to serve bytes, not redirect.

Fixed in the follow-up commit on this branch.

lib/utils/safeImageDownloadUrl.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (url.protocol !== 'https:' || url.username || url.password) {
+    return null
+  }
+
+  const hostname = url.hostname.trim().toLowerCase()
+  if (!hostname || isRestrictedDownloadHostname(hostname)) {
+    return null
+  }
+
+  if (isIP(hostname)) {
```
llun: **Every IPv6 literal URL is rejected, for the wrong reason**

`new URL(...).hostname` keeps the brackets on an IPv6 literal and `isIP('[::1]')` is 0, so every IPv6 URL fell through to the DNS branch, where resolving an address rather than a name fails and the empty-result check rejected it. Fail-closed, but it meant the blocklist never actually evaluated an IPv6 literal — `[::1]` was refused by accident and a legitimate public literal was refused too. `safeRemoteFetch` strips the brackets (`stripIpv6Brackets`).

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -87,28 +360,338 @@ describe('backfillMediaBlurhash execution', () => {
...
+    await backfillAttachments(db, mockStorage, options(), HOSTS)
+
+    const updated = await db('attachments').where('id', 'att-1').first()
+    expect(updated.blurhash).toBe('STALEHASHSTALEHASH')
+  })
+
+  // A remote instance runs this same code, so its attachment URLs carry exactly
+  // the `/api/v1/files/` path. Reading one out of OUR storage is both wrong and
+  // a way to make the sweep look at a path the remote actor chose.
+  it('does not read a remote instance URL out of local storage', async () => {
```
llun, line 676: **This test passes with the host check removed — it proves nothing**

The fixture URL is `https://remote.example/api/v1/files/../../secret.jpg`, and `new URL().pathname` normalises the `../../` before the host check ever runs, collapsing the path so it no longer starts with `/api/v1/files/`. The test therefore returns null for an unrelated reason. Verified by deleting the host check: this test still passed, and only its unit-level sibling caught the regression.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -58,10 +239,19 @@ export const getFileBuffer = async (
   if (!fileOutput) return null
   if (fileOutput.type === 'buffer') return fileOutput.buffer
   if (fileOutput.type === 'redirect') {
```
llun, line 326: **The storage-redirect branch and the direct-analysis write path had zero coverage**

Verified by deletion: removing the entire `redirect` block left all 39 tests passing, and turning the `if (buffer) { analyzeImageBuffer(...) }` block in `backfillAttachments` into a no-op also left all 39 passing — every test that reaches it ends with `buffer === null`, so `analyzeImageBuffer` was never once invoked for an attachment. The S3 driver answers with a redirect whenever a CDN hostname is configured, so that branch is the only way an object-storage deployment reads a file at all.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -4,41 +4,75 @@
...
  *
  * Usage:
- *   NODE_ENV=production scripts/maintenance/backfillMediaBlurhash.ts [--dry-run] [--batch-size 50] [--force]
+ *   NODE_ENV=production scripts/maintenance/backfillMediaBlurhash.ts [--dry-run] [--batch-size 50] [--force] [--local-only]
  *
  * Options:
  *   --batch-size   Number of rows to process per batch (default: 50)
  *   --dry-run      Analyze and print planned updates without modifying the database
- *   --force        Recompute blurhash/focus even on rows that already have them
+ *   --force        Recompute the blurhash even on rows that already have one.
```
llun, line 12: **The `--force` rationale is duplicated verbatim between the script header and `docs/maintenance.md`**

Both spell out, in near-identical prose, why `--force` recomputes the blurhash but never overwrites a focal point. This PR exists because a previous version of exactly this doc/code pair drifted apart — the docs promised a recompute the code did not do. Two independently maintained copies invite the same drift; one surface should own the rationale.

Fixed in the follow-up commit on this branch.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. `safeImageFetch` in `lib/utils/safeImageDownload.ts` now takes each hop by hand with `redirect: 'manual'` and re-runs the guard on the resolved target, capped at 3 hops like `safeRemoteFetch`. Both this script and the Strava photo path use it, so the same hole is closed there too. Proved by reverting four ways — `manual`→`follow`, the hop loop deleted, the target not re-guarded, and the budget removed — which fail 1, 9, 4 and 1 tests respectively.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea, then hardened in a687cdf05. The own-host list is no longer a `Set` queried with `.has()`; it goes through the app's own `normalizeHost`/`hostMatchesRule` pair, so a `*.example.com` entry matches a real subdomain exactly as it does for a request Host header. `isHostTrustedByRules` itself is not reused because `normalizeHost` deliberately answers null for loopback, which would stop a dev instance on `localhost:3000` recognising its own URLs — so a loopback authority falls back to literal equality. Round 2 then found that fallback compared an authority against a wildcard rule's own spelling (`new URL` parses `*` in a host), letting `https://*.<trusted-domain>/api/v1/files/<path>` read an attacker-chosen path out of local storage; a `*.`-prefixed candidate or rule is now excluded from it.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. A decoded path that starts with `/` or contains a `..` segment is refused before it reaches storage. `new URL().pathname` normalises a literal `../` away but a percent-encoded one survives it, so the check has to happen after decoding.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. `downloadRemoteImage` now genuinely returns null: the byte-cap error, a fetch throw and a non-ok response are all caught and warned about, so the contract matches the comment and every rejection is visible.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. It now warns with the row id and the storage path, matching what `backfillMedias` already did for the identical case.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. The presigned/CDN read now carries the same 10s timeout as the remote branch and `redirect: 'error'` — that URL is meant to serve bytes, not redirect. Also covered by tests now; the whole branch could previously be deleted with every test still passing.

lib/utils/safeImageDownloadUrl.ts
llun: Fixed in a687cdf05, then superseded by 2d03050f6. Stripping the brackets made the blocklist authoritative — but round 2's blocklist had no NAT64/6to4/Teredo/ORCHID entries, and round 3 found it also blanket-blocked NAT64 where the audited policy decodes the embedded IPv4, which would have refused every public IPv4-only origin on a DNS64 deployment. The address policy now lives in `lib/utils/unsafeAddress.ts` and both this guard and `safeRemoteFetch` import it, so there is nothing left to diverge. Verified byte-identical to the original modulo three `export` keywords, with zero policy disagreements across 777 addresses.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in f6ec053ea. The fixture URL contained `../../`, which `new URL().pathname` normalised away before the host check ran, so it returned null for an unrelated reason and passed with that check deleted. The `../../` is gone and the test now fails when the host check is removed.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. Both had no coverage — each could be deleted whole with all 39 tests still green. There are now tests for the storage-redirect branch (including its size cap and `redirect: 'error'`) and for the direct-analysis write path, on both the local and remote sides. Reverting each fails 3 and 2 tests.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in f6ec053ea. The script header now points at `docs/maintenance.md` for the reasoning instead of restating it, so the two cannot drift apart again — which is the failure this PR exists to fix.

PR #1560, opened 2026-08-26
fix: check the host before reading an actor archive attachment URL as a stored path

llun:
## Problem

`getMediaStoragePathFromUrl` in `scripts/backup/actorArchive.ts` documented itself as returning null "for fitness-file URLs (served from a different route) and for foreign/remote URLs", but the implementation only checked `parsed.pathname.startsWith('/api/v1/files/')`. It never compared the URL's host against this instance's own hosts.

`/api/v1/files/` is **this project's own route**, so every other activities.next instance serves its attachment URLs under exactly that path. At the three call sites (`registerAttachmentUrl`, and the actor's `iconUrl` / `headerImageUrl`), a remote instance's attachment URL was therefore:

- added to `mediaPaths` and mapped to an archive path via `getArchiveMediaPath`,
- missed by the local storage lookup, so the archive referenced a file it never included (and, without `--allow-missing-storage`, aborted the export),
- never handed to the `fetchRemoteAttachments` branch that would have downloaded it properly.

`scripts/maintenance/backfillMediaBlurhash.ts` shipped the same bug independently and is being fixed in #1559.

## Fix

The parser moves next to its inverse `getMediaFileUrl` as **`getMediaPathFromFileUrl`** (`lib/services/medias/mediaFileUrl.ts`) and now takes the instance's host config, so the two scripts that need it stop re-deriving it. #1559 keeps a local copy (`buildInstanceHosts` / `getLocalStoragePath`) while it is in flight; whichever of the two lands second can drop that copy for this helper — the behaviour is the same, including the host-relative branch, which is new here (the old script returned null for a relative URL because `new URL` threw). Happy to do that swap here if you'd rather it merge as one.

The host question itself is **`isOwnInstanceHost`** (`lib/utils/host.ts`), beside the existing host matchers rather than as a fourth private spelling of "is this us":

- it includes `ACTIVITIES_TRUSTED_HOSTS`, because a multi-domain instance mints media URLs on the **owning actor's** domain (`getAttachmentMediaMetadata` is handed `currentActor.domain`), which is not necessarily the configured primary host;
- it cannot be `isHostTrustedByRules` alone. That answers the narrower "may an inbound `X-Forwarded-Host` be believed", and `normalizeHost` deliberately rejects loopback names for it — so on a `localhost:3000` development instance it says no to the instance's own host, and a local export would have stopped archiving its own media. Canonical authorities are compared first (covering that case and every exact `host:port` match), then the rules matcher adds the wildcard entries (`*.example.com`) only it understands. The exact pass is deliberately more permissive than the rules matcher on `:80`; `canonicalAuthority`'s comment says why, and a test row pins it.

`getAttachmentMediaPath` (`lib/utils/getAttachmentMediaPath.ts`) is deliberately left alone: it never returns null and matches on `indexOf`, because its callers hold a URL this instance just wrote to its own storage.

Also removes an unused `getMediaFileUrl` import from `actorArchive.ts` (the module imported it but never called it).

## Tests

- `lib/services/medias/mediaFileUrl.test.ts` — 15 cases for `getMediaPathFromFileUrl`, including **a foreign host carrying the `/api/v1/files/` path**, a foreign host on some other path, a protocol-relative foreign URL, a suffix-only near-match (`evil-llun.test`), an unrelated subdomain, our own host on a different port, a loopback development host, a trusted host, a wildcard trusted host, and a `getMediaFileUrl` round trip.
- `lib/utils/host.test.ts` — `isOwnInstanceHost`, including the loopback pair that asserts `isHostTrustedByRules` answers `false` where `isOwnInstanceHost` answers `true`, so the reason the second pass exists is pinned.
- `scripts/backup/actorArchive.test.ts` — the parser block is replaced with a call-site regression test on the now-exported `registerAttachmentUrl`: a foreign-host `/api/v1/files/` URL leaves `mediaPaths`/`mediaIds`/`urlToArchivePath` untouched and warns; with `--fetch-remote-attachments` it reaches the download branch and lands in `media_attachments/remote/`; the memoization early-return and the fetch-failure warning branch are covered too.

**Revert proof:** replacing the host check with a bare `return parsed.pathname` fails 8 tests — 4 in `mediaFileUrl.test.ts` and 4 in `actorArchive.test.ts`. Every other new test was proven the same way, by reverting the one production line it names.

## Docs

`AGENTS.md` (new "A stored media URL is only ours if the host says so" section), `REVIEW.md` (Stored media), and `CLAUDE.md` get the invariant, since two scripts hit it independently.

## Review loop

Eight rounds of the sub-agent review loop ran on this PR (correctness, quality, scope/regression and testing lenses; the design-system lens was skipped — no `app/` or `lib/components/` changes). Eight findings were raised, all posted as inline threads, addressed, replied to and resolved. Rounds 7 and 8 were both clean, which is where it stopped.

Production code changed exactly once, in the first commit; every follow-up commit is tests and docs. The findings were: a doc bullet that claimed `backfillMediaBlurhash.ts` was covered by the shared parser when this PR does not touch it; two untested `registerAttachmentUrl` branches; a test case lost in the swap; duplicated fetch-mock teardown; a repeated call shape; a comment that wrongly claimed a `finally` does not run when an assertion throws; and a `canonicalAuthority` comment that described scheme-aware port stripping the helper does not do.

Two pre-existing issues were found and deliberately left alone: `copyProfileImage` joins the recovered storage path without the `..` guard `buildStoragePlan`'s `normalizeStoragePath` applies (not reachable — a local actor's `iconUrl` is always server-generated), and `exportActorArchive` itself has no test coverage, so the two icon/header call sites are only covered indirectly.

## Gate

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test` all green (878 test files, 11308 tests). No UI change, so no screenshots. No migration, so no schema-dump regeneration.


AGENTS.md
```diff
@@ -2238,6 +2238,12 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
 - Covered by `lib/services/medias/fileName.test.ts` plus entry-point regression tests in the `S3StorageFile.test.ts` / `localFile.test.ts` of both `medias/` and `fitness-files/`.
 
+### A stored media URL is only ours if the host says so
+
+- **Recover a stored media path from a URL with `getMediaPathFromFileUrl` (`lib/services/medias/mediaFileUrl.ts`) — never by checking the `/api/v1/files/` prefix on its own.** That route is this project's own, so **every other activities.next instance serves its attachment URLs under exactly that path**. A path-only check therefore reads a remote instance's URL as one of our storage paths: the lookup misses in storage, and at a caller that treats "not local" as "fetch it over HTTP instead" the download branch is skipped entirely, so the archive or backfill references a file it never obtained. Both `scripts/backup/actorArchive.ts` (attachment URLs plus the actor's `iconUrl`/`headerImageUrl`) and `scripts/maintenance/backfillMediaBlurhash.ts` shipped that bug independently, which is why the parser is shared rather than re-derived per caller. It is the inverse of `getMediaFileUrl` and lives beside it.
```
llun: **The doc claims a fix this PR did not make.** This bullet says "Both `scripts/backup/actorArchive.ts` … and `scripts/maintenance/backfillMediaBlurhash.ts` shipped that bug independently, which is why the parser is shared rather than re-derived per caller." That reads as "both callers now go through the shared, host-aware parser", but this PR does not touch `scripts/maintenance/backfillMediaBlurhash.ts` (`git diff origin/main...HEAD -- scripts/maintenance/backfillMediaBlurhash.ts` is empty). On `main` that script still has a looser version of the same bug at `scripts/maintenance/backfillMediaBlurhash.ts:237-241` — `row.url.indexOf('/api/v1/files/')`, an unanchored substring search with no host check — and because the download fallback is the `else if` of that same branch, a federated attachment from another activities.next instance is read as a local storage path, misses, and never reaches the HTTP fetch. Since `AGENTS.md` is what every session reads first, this line would tell a future contributor that script is already safe. Reword to state plainly that `backfillMediaBlurhash.ts` still carries it on `main`, that it is fixed on the open PR #1559 with its own local copy of the logic, and that whichever lands second should adopt this shared helper.

Found independently by the correctness, quality and scope/regression lenses.

CLAUDE.md
```diff
@@ -50,6 +50,7 @@
...
...is this project's own, so every OTHER activities.next instance serves its attachment URLs under exactly that path; matching on the path alone reads a remote URL as one of our storage paths, misses in storage, and — where the caller treats "not local" as "fetch it over HTTP instead" — skips the download branch entirely, leaving the archive referencing a file it never obtained. `scripts/backup/actorArchive.ts` and `scripts/maintenance/backfillMediaBlurhash.ts` shipped that bug independently, which is why the parser is shared and lives beside its inverse `getMediaFileUrl`. The host question is `isOwnInstanceHost` (`lib/utils/host.ts`): it must include `ACTIVITIES_TRUSTED_HOSTS`, because a multi-domain instance mints media URLs on the OWNING actor's domain, and it cannot be `isHostTrustedByRules` alone, which answers the narrower "may an inbound `X-Forwarded-Host` be believed" and whose `normalizeHost` rejects loopback names — so on a `localhost:3000` dev instance it says no to the instance's own host. `getAttachmentMediaPath` is NOT this check: it never returns null and is for URLs this instance just produced. See **A stored media URL is only ours if the host says so** in `AGENTS.md`.
```
llun: Same overclaim as Finding 1, in the condensed reminder list: "`scripts/backup/actorArchive.ts` and `scripts/maintenance/backfillMediaBlurhash.ts` shipped that bug independently, which is why the parser is shared and lives beside its inverse `getMediaFileUrl`." The second script is not migrated by this PR. Reword to match whatever `AGENTS.md` ends up saying.

scripts/backup/actorArchive.test.ts
```diff
@@ -121,27 +122,105 @@ describe('parseExportActorArgs', () => {
...
+    expect([...mediaIds]).toEqual(['media-1'])
+    expect(urlToArchivePath.get(url)).toBe('media_attachments/files/ab/cd.webp')
+    expect(warnings).toEqual([])
+  })
+
+  // `/api/v1/files/` is this project's own route, so every OTHER
+  // activities.next instance serves attachment URLs under exactly that path.
+  // Reading one as a local storage path put a file the archive never contains
+  // into the manifest and skipped the download branch below.
+  it('does not treat another instance media URL as a stored path', async () => {
```
llun, line 190: **Two branches of `registerAttachmentUrl` gain their first unit tests in this PR but are left uncovered.** (a) The memoization early return — `if (urlToArchivePath.has(attachment.url)) return` (`scripts/backup/actorArchive.ts:676`) — is never exercised, because every `runRegister` call starts with a fresh empty `Map`; a second call for the same URL should not duplicate a `mediaIds` entry or a warning. (b) The fetch-failure branch — the `catch` that pushes `Failed to fetch remote attachment ${url}: ${message}` (`scripts/backup/actorArchive.ts:709-713`) — has no test at all, for either a non-OK response or a thrown fetch. Both branches predate the PR, but this PR is what exported `registerAttachmentUrl` and wrote its first direct tests, so this is the natural point to close them.

lib/services/medias/mediaFileUrl.test.ts
```diff
@@ -37,3 +37,105 @@ describe('getMediaFileUrl', () => {
...
+      url: 'https://llun.test/api/v1/fitness-files/id-1',
+      expected: null
+    },
+    {
+      description: 'our own media route with nothing after it',
+      url: 'https://llun.test/api/v1/files/',
+      expected: null
+    },
+    {
+      description: 'a value that is not a URL at all',
```
llun, line 122: **A case lost in the swap.** The deleted `getMediaStoragePathFromUrl` block in `scripts/backup/actorArchive.test.ts` had a foreign host carrying a *different* path (`https://other.example/some/path.jpg`). Every foreign-host case in the new table reuses the same `/api/v1/files/ab/cd.webp` path. The behaviour is provably the same (`getOwnPathname` returns null on the host check before the pathname is read), so this is a lost literal case rather than a lost guarantee — but it is one row and it keeps the table describing both rejection reasons.

AGENTS.md
llun: Reworded in 31f94c43f. The bullet now says what is actually true: `scripts/backup/actorArchive.ts` is migrated to the shared host-aware parser, while `scripts/maintenance/backfillMediaBlurhash.ts` hit the same bug independently and still carries it on `main` — fixed on the open PR #1559 with its own local copy of the logic, and whichever of the two lands second should drop that copy for this helper. Deliberately not migrating that script here: #1559 rewrites the same function (adding `buildInstanceHosts`/`getLocalStoragePath`, `--local-only`, an SSRF guard and a stream cap), so touching those lines would conflict and duplicate its work. A follow-up commit (5be3a9425) also dropped the `:237` line reference, since #1559 moves that exact line.

CLAUDE.md
llun: Same reword in 31f94c43f, matching the AGENTS.md wording, with the line number dropped in 5be3a9425.

scripts/backup/actorArchive.test.ts
llun: Both branches now covered in 31f94c43f. `does not re-register a URL already resolved to an archive path` calls `registerAttachmentUrl` twice with the same URL but different `mediaId`s — a Set hides a duplicate value, so the differing id is what makes a missing memoization guard visible. An `it.each` table covers the fetch-failure branch for both a non-OK response and a rejected fetch, asserting exactly one `Failed to fetch remote attachment …` warning and no `urlToArchivePath` entry. Proven non-vacuous: neutralising the `urlToArchivePath.has` guard and removing the catch block's warning push fails all three new cases.

lib/services/medias/mediaFileUrl.test.ts
llun: Row restored in 31f94c43f — `https://other.example/some/path.jpg` → null, so the table again describes both rejection reasons (foreign host, and own host with a non-media route).

scripts/backup/actorArchive.test.ts
```diff
@@ -121,28 +122,175 @@ describe('parseExportActorArgs', () => {
...
+
+    return { mediaPaths, mediaIds, urlToArchivePath, warnings }
+  }
+
+  // `vitest.setup.ts` enables the fetch mock but leaves it passing through, so
+  // the tests that reach the download branch opt in with `doMock()`. One
+  // teardown covers all of them, including a later one that opts in without
+  // cleaning up after itself.
+  afterEach(() => {
+    fetchMock.resetMocks()
```
llun, line 161: **Duplicated fetch-mock setup/teardown.** Round 3, quality lens: the fetch-mock setup/teardown was duplicated across the two tests that opt into mocking (`fetchMock.doMock()` … `finally { fetchMock.resetMocks(); fetchMock.dontMock() }`), and `actorArchive.test.ts` was the only file in the repo using that per-test pattern — every other file that opts into mocking resets in a hook.

scripts/backup/actorArchive.test.ts
```diff
@@ -121,28 +122,175 @@ describe('parseExportActorArgs', () => {
...
+    // invisible either way — use a different `mediaId` per call so a second,
+    // non-memoized registration would show up as a second Set entry.
+    const url = 'https://example.test/api/v1/files/ab/cd.webp'
+    const mediaPaths = new Set<string>()
+    const mediaIds = new Set<string>()
+    const urlToArchivePath = new Map<string, string>()
+    const warnings: string[] = []
+    // Not `runRegister`: this needs the same collections across both calls,
+    // and that helper allocates a fresh set of them per call.
+    const register = (mediaId: string) =>
```
llun, line 243: **Repeated inline `registerAttachmentUrl` call.** Round 3, quality lens: the test repeated the full seven-field `registerAttachmentUrl({...})` call twice inline, differing only in `attachment`. It legitimately cannot use `runRegister` — that helper allocates fresh collections per call, and this test needs them shared — but the repetition should be a local closure.

scripts/backup/actorArchive.test.ts
```diff
@@ -121,28 +122,175 @@ describe('parseExportActorArgs', () => {
...
+      urlToArchivePath,
+      stagingDir,
+      warnings
+    })
+
+    return { mediaPaths, mediaIds, urlToArchivePath, warnings }
+  }
+
+  // `vitest.setup.ts` enables the fetch mock but leaves it passing through, so
+  // the tests that reach the download branch opt in with `doMock()`. One
```
llun, line 157: **Comment states a false fact about `try`/`finally`.** Round 4, quality lens: the comment justifying the hoisted teardown claimed the hook runs when an assertion throws "unlike a `finally`". That is wrong — a `finally` block does run when its `try` body throws, which is exactly what the replaced code relied on. The comment taught a false fact about `try`/`finally` semantics.

llun (commented):
Round 3/4 review-loop findings for `scripts/backup/actorArchive.test.ts`, posted for the loop's record. All three were already addressed by `eaceb75dc` and `f116c6c58`; reply + resolve follow immediately below each.

scripts/backup/actorArchive.test.ts
llun: Fixed in `eaceb75dc`. One `afterEach` on the `registerAttachmentUrl` describe now restores the passthrough mock, so the failure-case `it.each` needs no `try`/`finally` at all and the download test's `finally` is left with just its temp-directory cleanup. Round 4's testing lens re-proved all five behaviours in the block by reverting each corresponding piece of production code one at a time.

scripts/backup/actorArchive.test.ts
llun: Fixed in `eaceb75dc` — one `register(mediaId)` closure over the shared collections, with a comment stating why `runRegister` does not fit. Verified still non-vacuous: removing the `urlToArchivePath.has` guard fails it with `['media-1','media-2']` vs `['media-1']`.

scripts/backup/actorArchive.test.ts
llun: Fixed in `f116c6c58`. The comment now gives the real reason — the teardown was duplicated, and one hook also covers a later test that opts into `doMock()` without cleaning up after itself. `eaceb75dc`'s commit message carries the same wrong claim; it is corrected in `f116c6c58`'s message rather than by rewriting a pushed commit.

lib/utils/host.ts
```diff
@@ -201,6 +201,55 @@ export const isHostTrustedByRules = (
   )
 }
 
+/**
```
llun, line 204: **`canonicalAuthority`'s `:80|:443` stripping is scheme-blind, and the comment did not say so.** It strips a literal `:80` or `:443` off any authority regardless of the original URL's scheme, so `isOwnInstanceHost`'s exact-authority pass accepts `https://llun.test:80/...` as ours, while `hostMatchesRule`/`getPortForRuleMatching` — which treats only `:443` as an implied default — would say no. The old JSDoc claimed the helper drops "the port the scheme implies anyway", which it cannot do: it is handed a bare authority with no scheme to consult. The test table pinned the `:443` case but had no `:80` case, so the asymmetry was accidental rather than chosen. Not attacker-reachable — `isOwnInstanceHost` is reached only through `getMediaPathFromFileUrl`, and every URL the archive script feeds it is minted by this codebase's own `getMediaFileUrl`, which never emits an explicit `:80`/`:443`.

llun: Addressed in `ea4bbd3d6`. The behaviour is kept: because `canonicalAuthority` is applied to BOTH the incoming authority and every configured rule, and never alters the hostname text, the only equivalence a blanket strip can newly create is between an incoming host and a rule sharing the identical hostname — our own host on the other default port, which this instance answers to either way. Round 8 tried to construct a counter-example to that and could not. What changed is the comment, which now states what is actually stripped, why the strip is deliberately not scheme-aware, and that the exact pass is more permissive than `hostMatchesRule` on `:80` — plus a test row pinning `llun.test:80` → true, so the behaviour is chosen rather than accidental. Proven load-bearing: tightening `canonicalAuthority` to strip only `:443` fails that row and nothing else.

PR #1566, opened 2026-08-26
fix: report attachments the blurhash backfill cannot repair

llun:
## Summary

Deleting a `medias` row leaves every `attachments.mediaId` that pointed at it naming a row that is gone, and `scripts/maintenance/backfillMediaBlurhash.ts` cannot repair such an attachment — its blurhash, focal point and `thumbnailUrl` all come from the linked media row, and `thumbnailUrl` has no other source. Because a host-relative `thumbnailUrl` keeps matching the selection predicate, the row is selected on every run, counted in `processed`, and nothing is logged. An operator reading `processed N, updated 0` could not tell that from a run with nothing to do. Found during review of #1559 and deferred from it.

This PR reports the gap and leaves the data alone. It warns per row and adds **two counts to the summary line, deliberately never summed**, because they call for opposite responses:

```text
[attachments 0f3c…] media 412 no longer exists; cannot restore blurhash, focus or thumbnailUrl from it
[attachments 91ab…] mediaId "wat" is not a media row id; cannot restore blurhash, focus or thumbnailUrl from it
Attachments complete: processed 1204, updated 6, 37 whose media row is gone, 2 with an invalid mediaId
```

- **`whose media row is gone`** — a real row id whose media row was deleted. The expected residue of owners deleting their own media; nothing to act on.
- **`with an invalid mediaId`** — a value that was never a row id, so nothing was deleted. That is a bad *write*, from the unvalidated `createAttachment` path AGENTS.md documents under **Database Compatibility Guidelines**, and is worth investigating. Not SQLite-only: `-5` and `0` are valid `integer` values PostgreSQL stores and the id guard refuses either way.

Neither count partitions `processed` — a warned row can still be repaired from its own image bytes.

## Why not clear the pointer, and why not delete the attachment

Three fixes were on the table. Only the reporting one is right; the other two are regressions, and the review loop independently verified both claims rather than taking them on trust.

**Nulling `attachments.mediaId` would make a broken attachment permanent.** A NULL `mediaId` is how a *federated* attachment is stored, and `isReplaceableMediaAttachment` (`lib/database/sql/status.ts`) reads exactly that to decide an attachment is not the owner's to replace. Today a dangling pointer is still non-null, so editing the post removes the attachment; null it and `updateNote` never touches it again. The existing test `clears only editable media while preserving legacy and fitness attachments` pins that behaviour.

**Refusing the delete would revoke a shipped capability.** The two DELETE routes differ on purpose: `DELETE /api/v1/media/:id` is Mastodon-compatible and answers 422 for in-use media, while `DELETE /api/v1/accounts/media/:mediaId` — the Settings → Media Storage button — deliberately allows it, promising "Posts containing this media will show a placeholder image". That promise is kept **only when the attachment already carries a BlurHash**: `lib/components/posts/media.tsx` gates the `` at `opacity-0` until `onLoad` (which a 404 never fires) solely inside its `if (blurhash)` branch, and a falsy one renders a bare ``. Review caught that I had stated this unconditionally; the docs now carry the caveat.

**Deleting the attachment row** would remove that placeholder and silently rewrite a published status whose federated copies keep the attachment.

## Test plan

- Table-driven cases covering a `mediaId` whose media row is gone and one that is not a row id, asserting the warning, the exact summary line, and that the row really is left untouched — so the warning is not redundant with an update.
- A second table-driven pair pinning that **both** branches keep warning on a row the analysis step still repairs, i.e. that the counts do not partition `processed`. Self-healing is reachable for a deleted media row too: the delete route removes stored bytes best-effort and drops the row regardless, so the file behind `url` can outlive the row that named it.
- A negative case proving a federated attachment (NULL `mediaId`) is counted under neither cause, or every remote attachment on an instance would read as a gap.
- **Proved by mutation, not just by passing.** Reverting the whole change fails all the new tests; conflating the two counters fails two; forcing the counts to partition `processed` fails exactly the overlap case; and suppressing the warning on either branch fails that branch's own case.
- Full gate green on Node 24: `prettier --write .` → `lint` (0 errors) → `typecheck` → `build` → `test` (881 files, 11545 tests).
- No migration, no schema change, no runtime code touched — the change is confined to a maintenance script.

## Docs

- `AGENTS.md` — new **Deleting Media a Post Uses** section recording why the two routes differ and why the pointer must not be cleared. The trap is live: clearing it was the first fix proposed.
- `CLAUDE.md` — matching reminder bullet.
- `docs/maintenance.md` — operator-facing subsection explaining what each count means and which one warrants investigation.

## Review

Seven rounds of the sub-agent loop (correctness, testing, quality, scope/regression), each pointed at the previous round's fixes rather than the original diff. Seventeen findings, all addressed; round 7 came back clean on code and tests. The design-system lens was skipped throughout — this diff touches no `app/` or `lib/components/` code.

Rounds 1–2 were posted as inline threads and resolved. Rounds 3–6 were fixed in commits but not recorded here at the time — a process failure caught by round 7 and reconciled in a follow-up review comment on this PR.

Worth stating plainly: five of the findings were errors in my own reasoning rather than style — a counter conflating a deliberate deletion with a bad write, an unconditional placeholder claim, a "never counted in `updated`" claim that was false, a "SQLite-only" claim that was false, and a comment asserting a permanence the next round's own test disproved. Four separate mutations passed the full suite before being caught, two of them in tests I had just written to close an earlier mutation.

### Known gaps, deliberately excluded

All four predate this PR and concern code it does not touch; each was confirmed by mutation:

1. `backfillMedias` has no `--dry-run` coverage — stripping its gate passes the suite.
2. `options.batchSize` is unpinned for `backfillAttachments` — hardcoding `.limit(50)` passes.
3. The summary log's position after the loop is unguarded.
4. `vi.restoreAllMocks()` does not reset `vi.mock` factory mocks (traced into `@vitest/spy`); no test currently depends on the leak, and adding a `mockReset` could break a sibling that does.

Tracked for a follow-up rather than widened into this change.


llun (commented):
Round 1 of the sub-agent review loop: four lenses (correctness, testing, quality, scope/regression). Quality returned clean. Testing independently re-ran the revert-proof and passed it. Four findings below survived — three of them are inaccuracies in prose this PR itself added, and two lenses reached the counter-conflation independently.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -542,6 +543,23 @@ export const backfillAttachments = async (
...
+          // block would have copied is unreachable — and `thumbnailUrl` has no
+          // other source at all, because rebuilding it needs the media row's
+          // stored thumbnail path. Deleting a `medias` row does not clear the
+          // `attachments.mediaId` pointing at it (nor should it: a null
+          // `mediaId` marks a federated attachment, and `updateNote` uses that
+          // to decide an attachment is not the owner's to replace), so these
+          // rows are permanent. Without this they were invisible: counted in
+          // `processed`, absent from `updated`, re-selected by every later run
+          // — a sweep reporting "processed N, updated 0" and no reason why.
+          totalUnresolvedMedia += 1
```
llun: **The counter folds two causes with opposite operator responses into one number.** (found independently by the correctness and scope/regression lenses)

`rowId === undefined` means the `mediaId` was **never a row id** — which on a SQLite instance is reachable through the unguarded `createAttachment` write from `POST /api/v1/accounts/outbox`, the separately-tracked open bug this repo already documents at `lib/database/sql/media.ts:350`. `rowId` defined but absent from `mediaMap` means the media row was **deleted on purpose**.

The first wants investigation; the second wants nothing. Folding them means the summary line — the thing an operator reads first, which is the entire argument for adding it — cannot tell them apart.

Fix: split into two counters and report both.

docs/maintenance.md
```diff
@@ -631,6 +631,19 @@ NODE_ENV=production ./scripts/maintenance/backfillMediaBlurhash.ts --local-only
...
+
+Such a row cannot be repaired by this script: the BlurHash, focal point and `thumbnailUrl` it fills in all come from the linked `medias` row, and `thumbnailUrl` has no other source. Because a host-relative `thumbnailUrl` keeps matching the selection predicate, the row is re-read on every run and never updated. The script warns for each one and reports the total:
+
+```text
+[attachments 0f3c…] media 412 no longer exists; cannot restore blurhash, focus or thumbnailUrl from it
+Attachments complete: processed 1204, updated 0, 37 with an unresolvable mediaId
+```
+
+A non-zero count is not an error to act on — it is the number of posts whose media the owner deleted on purpose. It is there so a run that updated nothing is distinguishable from a run that had nothing to do. The author can drop the leftover attachment by editing the post.
```
llun: **Factually wrong, and wrong in the direction that tells an operator to ignore a real bug.**

"it is the number of posts whose media the owner deleted on purpose" is only true for one of the two branches the counter feeds from. A `mediaId` that was never a row id lands in the same number, and that is evidence of the unvalidated-`mediaId` write path, not of a deliberate deletion.

This is exactly the "operator-facing all-clear that isn't" failure this PR exists to prevent, reintroduced one paragraph below the fix.

AGENTS.md
```diff
@@ -1815,6 +1815,42 @@ system's `Attachments` component.
...
+  pins the surviving-null behaviour. Deleting the attachment row alongside the
+  media is wrong for a different reason: it removes the placeholder the dialog
+  promises and silently rewrites a published status whose federated copies keep
+  the attachment.
+- **What a dangling pointer does cost is a maintenance sweep that cannot report
+  it.** `scripts/maintenance/backfillMediaBlurhash.ts` rebuilds an attachment's
+  BlurHash, focal point and `thumbnailUrl` from the linked `medias` row, and
+  `thumbnailUrl` has no other source — so a row whose media is gone is selected
+  (a host-relative `thumbnailUrl` matches the predicate), counted in
+  `processed`, never counted in `updated`, and re-selected by every later run.
```
llun: **"counted in `processed`, never counted in `updated`" is false for the never-a-row-id branch.**

When a `mediaId` is malformed, nothing was actually deleted — `row.url` is usually still a live file, so the direct-analysis fallback further down still runs and succeeds. The testing lens built the case and ran it:

```text
console.warn: [attachments att-1] media 404 no longer exists; ...
console.log:  [attachments att-1] updated
console.log:  Attachments complete: processed 1, updated 1, 1 with an unresolvable mediaId
```

The same row is in both counters. The two do **not** partition `processed`, and the prose here (mirrored in `CLAUDE.md` and the commit message) claims they do.

AGENTS.md
```diff
@@ -1815,6 +1815,42 @@ system's `Attachments` component.
...
+## Deleting Media a Post Uses
+
+- **The two media DELETE routes disagree on purpose, and neither is a bug.**
+  `DELETE /api/v1/media/:id` is the Mastodon-compatible one and refuses media
+  that a posted status already uses (`deleteMediaForAccount` answers `in-use`,
+  which the route turns into a 422), matching Mastodon's `in_usage_error`.
+  `DELETE /api/v1/accounts/media/:mediaId` is this instance's own, behind the
+  Settings → Media Storage delete button, and deliberately allows it: its
+  confirmation dialog promises "Posts containing this media will show a
+  placeholder image", and that promise is kept by `lib/components/posts/media.tsx`,
```
llun: **The placeholder claim is stated unconditionally but only holds when the attachment already carries a blurhash.**

`lib/components/posts/media.tsx` only applies the `opacity-0`-until-`onLoad` treatment inside its `if (blurhash)` branch. With a falsy blurhash it falls through to a bare `` with no canvas underneath — a 404 there is a broken-image icon, not a placeholder.

That matters for precisely the rows this PR is about: once the media row and its bytes are deleted, the script's own direct-analysis fallback cannot recover a blurhash either, so such a row is permanently placeholder-less. Qualify the sentence here and in `CLAUDE.md` so a future reader does not dismiss a genuine broken-image report as handled by design.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -807,6 +807,79 @@ describe('backfillMediaBlurhash execution', () => {
     )
   })
 
+  // Deleting a `medias` row leaves every `attachments.mediaId` that pointed at
+  // it dangling, and the sweep cannot repair such a row: the thumbnailUrl
+  // rebuild is the media block's alone, so a host-relative value stays
+  // host-relative and the row is re-selected by every later run. Before this
+  // warning the operator saw only "processed 1, updated 0".
+  it.each([
```
llun, line 821: **Both fixtures pre-set `blurhash`, which hides the overlap.** With `blurhash` already present, `shouldAnalyze` is always false and `updated` always stays 0 — so nothing in the suite pins that a warned row can also be updated.

Add a case with a dangling `mediaId`, `blurhash: null` and a successful direct analysis, asserting both counters and the resulting row, so the non-partitioning is pinned as intentional rather than left to be rediscovered.

Separately: the `'abc'` case is only reachable on SQLite — `attachments.mediaId` is `integer` on PostgreSQL, so that INSERT fails before this branch is reached. Worth a comment saying so, since this file hardcodes `better-sqlite3` and no CI configuration can ever exercise the pg side.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in 735deace2. Split into `totalDeletedMedia` and `totalInvalidMediaId`, reported as two clauses that are never summed, with the comment on the declaration saying why. Agreed on the reasoning: the summary line is the thing an operator reads first, which was the whole argument for adding it, so conflating the two partly defeated the fix.

docs/maintenance.md
llun: Fixed in 735deace2. The section is rewritten and retitled (`Attachments whose mediaId resolves to nothing`), and now splits the guidance: a gone media row is not an error to act on, an invalid `mediaId` is a bad write and is worth investigating, with the `createAttachment` / outbox path named and the SQLite-vs-PostgreSQL asymmetry called out. The false sentence is gone.

AGENTS.md
llun: Fixed in 735deace2. You are right and I was wrong — nothing is deleted in the invalid-id case, so the file behind `url` is still there and the analysis step repairs the row anyway. Dropped the claim from AGENTS.md, CLAUDE.md and the commit message, and stated the opposite explicitly: neither count partitions `processed`. Added `counts a warned attachment that the analysis step still repairs`, which asserts `updated 1` alongside `1 with an invalid mediaId`.

AGENTS.md
llun: Fixed in 735deace2. Confirmed against `media.tsx` directly — the opacity guard and canvas live only inside the `if (blurhash)` branch, and the fallthrough is a bare ``. AGENTS.md and CLAUDE.md now say the promise holds only for an attachment that already carries a BlurHash, that a falsy one renders a broken-image icon, and that the sweep cannot compute one once the bytes are gone — with an explicit "do not read a broken-image report as handled by design". Worth noting this does not change the decision to leave the delete route alone: the deliberateness of allowing an in-use delete rests on the dialog copy and the route split, not on the placeholder always rendering. My supporting evidence was overstated, not the conclusion.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Both addressed in 735deace2. Added `counts a warned attachment that the analysis step still repairs` (dangling `mediaId`, `blurhash: null`, successful analysis) asserting `processed 1, updated 1, 0 whose media row is gone, 1 with an invalid mediaId` plus the written row, so the non-partitioning is pinned as intentional. Proved it by mutation: gating `shouldAnalyze` on having warned makes the counts partition, and that test alone fails. The `abc` fixture now carries a comment saying it is reachable on SQLite only, that PostgreSQL rejects the insert at the `integer` column, and that this file builds its own better-sqlite3 database so no `TEST_DATABASE_TYPE` exercises the pg side.

llun (commented):
Round 2 (four lenses, pointed at round 1's fixes rather than the original diff). Correctness and scope/regression both cleared the counter-split attribution — including the Map-key-identity question across the number-vs-string backend difference. Four findings, all fixed in e2fae2059:

1. **A mutation that survived.** Suppressing the warning for exactly the rows the analysis step self-heals passed 71/71 — the overlap test asserted the summary and the repaired blurhash but never that the warning fired, and the sibling case that does assert it has a blurhash pre-set so it never self-heals. Asserted now; reproduced the mutation and it fails.
2. **Another wrong claim of mine.** "expected only on SQLite-backed instances" is false — `-5` and `0` are valid `integer` values PostgreSQL stores and `toMediaRowId` refuses either way, so the count is reachable on both backends.
3. **Over-documentation.** The `createAttachment`/outbox/backend-typing chain was re-derived in five places while AGENTS.md's Database Compatibility Guidelines already documents it — six copies to update when the outbox bug is fixed. They point at that section now, which also deletes most of finding 2 rather than restating it correctly five times.
4. **Verbosity.** AGENTS.md section 578 → 395 words, CLAUDE.md bullet 312 → 222, and four comment blocks in `backfillAttachments` that each justified "not summed" folded into one at the declarations.

Net -18 lines. Quality explicitly endorsed keeping the two-counter split itself.

Two things deliberately not actioned: `vi.restoreAllMocks()` does not reset `vi.mock` factory mocks (traced into `@vitest/spy`), but it is pre-existing in this file, no test currently depends on the leak, and 20 shuffled runs plus 8 clean consecutive runs all pass — adding a `mockReset` could break a sibling that relies on it, so it belongs in its own change. And a transient flake sighting during review was an artifact of two agents mutating this worktree concurrently (stray `TRACE` writes were still in the tree afterwards); on a clean tree the file passes 8/8.

llun (commented):
## Review loop rounds 3–7 (backfilled record)

Rounds 1 and 2 were posted as inline threads at the time. Rounds 3–6 were addressed in commits but **never recorded here**, which is a process failure on my part: this project's policy is that the PR threads are the source of truth, and the visible trail showed two rounds where seven happened. Caught by round 7's scope lens. Reconstructing it below, each finding with the commit that fixed it.

### Round 3 — correctness clean, scope clean, testing found 1
- **The twin branch was left unguarded.** Round 2 pinned that a row with an invalid `mediaId` keeps its warning when the analysis step repairs it, but suppressing the same warning on the `totalDeletedMedia` branch still passed 71/71. Fixed in `f0a7c7197`, as a table over both branches. That branch is live: the delete route removes stored bytes best-effort and drops the row regardless, so the file can outlive the row that named it.

### Round 4 — testing found 2, scope found 1
- **Warn-per-row collapsed to once-per-run survived.** Every fixture had one affected row per batch, so an operator would have seen only the first id. Fixed in `5d1cd946c`.
- **`--dry-run` never exercised `backfillAttachments` at all**, though `docs/maintenance.md` tells operators to run it first. Fixed in `5d1cd946c`.
- **A stale comment round 2 wrote and round 3 disproved** — it claimed a gone media row can never be repaired. Corrected in `5d1cd946c`, in the code comment and `docs/maintenance.md`.

### Round 5 — testing found 1, scope found 1 (same root cause)
- **The dry-run test was vacuous.** Both fixture rows had `blurhash` pre-set, so the update block was never entered and "without writing" asserted nothing — stripping the gate to an unconditional write passed 73/73. Fixed in `ad3672d72` with a row that genuinely diverges.
- **Self-caught while proving that fix:** the multi-row test gave its two rows *different* causes, so each counter only reached 1 and the warn-once mutation suppressed nothing. Now a table over both causes, two same-cause rows each, across two batches.

### Round 6 — testing clean for this PR's surface, quality found 1
- **A comment clause narrating the review history** rather than the code. Trimmed in `1c422db47`.
- Testing found two survivors in code this PR does not touch (`options.batchSize` unpinned; summary log position unguarded) — see the exclusions below.

### Round 7 — clean
No code or test finding. Verified the four accepted exclusions against source, confirmed no scope creep, and flagged the two process items being fixed by this comment and the description update.

### Known gaps, deliberately excluded
All four predate this PR and concern code it does not touch, each confirmed by mutation:
1. `backfillMedias` has no `--dry-run` coverage — stripping its gate passes.
2. `options.batchSize` is unpinned for `backfillAttachments` — hardcoding `.limit(50)` passes.
3. The summary log's position after the loop is unguarded.
4. `vi.restoreAllMocks()` does not reset `vi.mock` factory mocks (traced into `@vitest/spy`); no test currently depends on the leak.

They are tracked for a follow-up rather than widened into this change.

PR #1568, opened 2026-08-26
fix: confirm a gear component retire, and let it be undone

llun:
## Summary

Two problems with retiring a fitted component, from the review of #1530.

### Retire fired on one unconfirmed click

The adjacent Delete deliberately arms first and acts on the second click, with a rationale comment explaining why. Retire did not — and it is the *harder* of the two to notice going wrong. Delete removes a row you can see is gone; retire closes the install window silently, so the part stops accruing distance and nothing says so until someone reads the Retired column.

Retire now arms the same way, with both disarm rules Delete already had: leaving the button (`onBlur`), and hiding the retired rows.

### A misclick was unrecoverable from the UI

`PATCH /api/v1/fitness/gear/:id/components/:componentId` has always accepted `removedAt: null`, and `lib/client.ts` has always exported `updateFitnessGearComponent` — with **zero** callers outside its own definition and route tests. So the only way back from a mis-retire was the database.

Retired rows now carry an **Unretire** beside Delete, deliberately *not* armed.

Review round 1 corrected the reasoning I first gave for that, so it is worth stating precisely. Clearing `removedAt` reopens the **original** `[addedAt, now)` window, and a component has one window with no per-period history — so Unretire is *not* the mirror of Retire. Undoing a retire seconds later credits nothing extra. Undoing one from last season credits every activity ridden while the part sat retired, and **re-retiring does not take that back**, because it only closes the window at the new `now`. Verified against the rollup query with a throwaway database: a component retired at `2026-02-01` reading 10 km went to 30 km on unretire and stayed at 30 km after re-retiring.

It stays unarmed anyway. Arming it would add friction to the misclick this exists to recover from without preventing the misattribution — and the UI cannot tell a mistaken unretire from a legitimate refit of the same part, which is a real thing people do.

This is a pure UI change. `CreateFitnessGearComponentInput.removedAt` is already `number | null`, the request schema's `optionalEpochMilliseconds` is `.nullish()` (no `.transform()`, so the ordering trap does not apply), and the route's `removedAt <= addedAt` cross-check cannot trip when `removedAt` is null. `route.test.ts` already pins the server side.

### One armed id, not one per action

The obvious implementation adds `confirmingRetireId` beside `confirmingDeleteId`. That is wrong: a row armed for Delete and then unretired keeps that arm, and retiring it again brings Delete back **already armed** — a one-click delete, which is the exact hazard the disarm rules exist to prevent.

A row offers exactly one confirmable action and `removedAt` decides which, so there is one `confirmingActionId`. That makes the bad state unrepresentable rather than something to remember to clear, and preserves the existing "only one row armed at a time" invariant for free.

## Layout

The retired row's actions cell keeps its declared **84px** minimum rather than growing for the pair. Three reasons:

- A `<td>` minimum is per-row, but the column width is the max across rows — so a `isRetired ? 168 : 84` ternary is misleading: the installed rows' 84 is dead the moment one retired row is visible.
- The pair occupies ~185px anyway, so any smaller declared minimum never binds.
- 84 is quoted in **both** `AGENTS.md` and `CLAUDE.md` as one of the seven per-cell minimums summing to ~740px. Changing it silently makes both wrong; changing it honestly drags a two-file docs edit into a one-component PR.

`flex flex-wrap justify-end gap-1` is what handles the width. Below `GEAR_TABLE_SNAP_WIDTH` (480px) `dataColumnStyle` returns a *fixed* panel, and an overhang there eats the value from the right and cannot be scrolled to under `x mandatory` — so wrapping, not overflowing, is the behaviour that matters. Measured at 375px: the panel is 221px and Unretire (78px) + Delete (67px) sit on one line, right edges at 1376/1447 inside a cell ending at 1463. No overhang, no stacking needed; the wrap is insurance for narrower containers.

## Browser verification

Against a worktree-local SQLite `dev.sqlite3` on `localhost:3200` — never a remote database — with a seeded bike carrying one installed and one retired component.

| Check | Result |
| --- | --- |
| First click on Retire | arms to "Confirm retire", no request sent |
| Showing retired rows | disarms a pending confirm |
| Unretire | count 1→2 installed, row un-dims, action flips back to Retire |
| Full retire → unretire → retire | round-trips |
| Retired row actions | Unretire + Delete on one line, no overflow |
| Dark mode | reads correctly |
| 375px | both buttons on one line inside the cell |

## Testing

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test`, all green (884 files / 11,572 tests).

`GearComponentsCard.test.tsx` goes from 33 tests to 41. Two existing tests changed, because they pinned the single-click retire this PR removes.

Every new guard is **proved by reverting its subject**:

| Revert | Tests that fail |
| --- | --- |
| retire back to a single unconfirmed click | 5 |
| Unretire button removed | 4 |
| unretire sends something other than `removedAt: null` | 1 |
| retire `onBlur` disarm removed | 1 |
| `flex-wrap` dropped from the actions cell | 1 |
| unretire no longer clears the arm | 1 |
| `aria-label` loses the visible label | 5 |

That last one is WCAG 2.5.3 (Label in Name): the accessible name must *contain* the visible label, so the armed button is `Confirm retire <type>`, not `Confirm retiring <type>`.

## From review round 1

- **The new unretire tests were order-dependent.** `vi.clearAllMocks()` resets call history and leaves implementations in place, so with no `beforeEach` default for `updateFitnessGearComponent` they passed only on whichever neighbour had last set one — including the rejection from "surfaces an unretire failure". Removing the default and running under `--sequence.shuffle` fails; with it, every seed tried passes. (Round 2 noted that specific seed numbers are environment-dependent, so the claim is stated as the property rather than as particular seeds.) CI does not shuffle, so this would have sat latent.
- **The layout claim was asserted in this description and tested nowhere.** jsdom does no layout, so it is now pinned as the classes that produce it. Round 2 caught that asserting `flex-wrap` alone was not enough: `flex-wrap` does nothing without `display: flex`, and the enclosing `<td>` carries `whitespace-nowrap`, so dropping only `flex` gave a block container that put both buttons on one line and overhung the panel by 18px — measured in a browser — while the test still passed. It now asserts both tokens, and each drops independently to a failure.
- **A false claim survived in a second place.** Round 1 corrected the rationale comment in the component but left the identical retracted wording in the test file, 200 lines from its correction. Both now say the same true thing.
- A duplicated `setConfirmingActionId(null)`, left by collapsing two ids into one.

## Deliberately not in scope

- **Backdating a retire.** The API supports it — the PATCH route takes an arbitrary `removedAt` and already validates it against the stored `addedAt`, returning a real `422` with `removedAt must be after addedAt`. But a backdated retire silently rewrites derived distance history everywhere it is read (the gear page, the stat grid, the fitness overview), and it turns a one-click action into a form. Different blast radius, and it deserves its own change.
- **Guarding a future `addedAt`.** The inverted-window case the review notes named is **already handled**: `gearRequests.ts`'s `.refine` plus the route's cross-check against the stored other end. A future `addedAt` alone is not blocked, but that is arguably legitimate — planning a part swap — and it is not the same defect.
- **Per-period component history.** The gap-credit above is inherent to one `[addedAt, removedAt)` window per component; fixing it means recording install periods, which changes the schema and every rollup. Filed separately.
- **The gear-level Retire** in `GearDetailView` is also a single-click toggle. It is genuinely reversible in a way the component one was not — the same button unretires — so it does not share this defect, and changing it here would widen a one-component PR into the gear surface.
- **A better error than `404` when retiring an already-retired row.** Real, but the `404` is a deliberate outcome of a compare-and-swap: `retireFitnessGearComponent` predicates its `UPDATE` on `whereNull('removedAt')` so two concurrent requests produce one write, and 0 rows affected maps to null. Distinguishing "not found" from "already retired" means a second read to undo a design that is correct. With the two-click confirm and the row re-rendering as retired, the window is now narrow.

### DIFF
--- a/app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
+++ b/app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
@@ -101,6 +101,13 @@ describe('GearComponentsCard', () => {
     mockRetireFitnessGearComponent.mockResolvedValue(
       createComponent({ removedAt: Date.UTC(2025, 5, 1) })
     )
+    // Without this default the unretire tests pass only on a neighbour's
+    // leaked implementation: `vi.clearAllMocks()` resets call history and
+    // leaves implementations in place, so whichever test last set one on this
+    // mock decides what the next test sees. Shuffled seeds 3, 8 and 10 failed.
+    mockUpdateFitnessGearComponent.mockResolvedValue(
+      createComponent({ removedAt: null })
+    )
   })
 
   it('renders the header with the installed count', () => {
@@ -416,6 +423,24 @@ describe('GearComponentsCard', () => {
 
     expect(await screen.findByText('Component not found')).toBeInTheDocument()
     expect(onChanged).not.toHaveBeenCalled()
+  })
+
+  // jsdom does no layout, so the wrap can only be asserted as the class that
+  // produces it. Below GEAR_TABLE_SNAP_WIDTH `dataColumnStyle` returns a FIXED
+  // panel, and an overhang there eats the value from the right and cannot be
+  // scrolled to under `x mandatory` — so a retired row's two actions must be
+  // able to wrap rather than spill.
+  it('lets the retired row actions wrap instead of overflowing', () => {
+    renderCard([createComponent({ removedAt: Date.UTC(2025, 5, 1) })])
+
+    fireEvent.click(
+      screen.getByRole('button', { name: 'Show 1 retired component' })
+    )
+
+    const actions = screen
+      .getByRole('button', { name: 'Unretire Chain' })
+      .closest('div')
+    expect(actions).toHaveClass('flex-wrap')
   })
 
   it('offers both unretire and delete on a retired row', () => {
--- a/app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
+++ b/app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
@@ -7,7 +7,8 @@ import { act, fireEvent, render, screen, waitFor } from '@testing-library/react'
 import {
   createFitnessGearComponent,
   deleteFitnessGearComponent,
-  retireFitnessGearComponent
+  retireFitnessGearComponent,
+  updateFitnessGearComponent
 } from '@/lib/client'
 import type { GearComponentEntity } from '@/lib/services/fitness-gears/gearEntities'
 
@@ -16,7 +17,8 @@ import { GearComponentsCard } from './GearComponentsCard'
 vi.mock('@/lib/client', () => ({
   createFitnessGearComponent: vi.fn(),
   deleteFitnessGearComponent: vi.fn(),
-  retireFitnessGearComponent: vi.fn()
+  retireFitnessGearComponent: vi.fn(),
+  updateFitnessGearComponent: vi.fn()
 }))
 
 const mockCreateFitnessGearComponent =
@@ -30,6 +32,10 @@ const mockDeleteFitnessGearComponent =
 const mockRetireFitnessGearComponent =
   retireFitnessGearComponent as jest.MockedFunction<
     typeof retireFitnessGearComponent
+  >
+const mockUpdateFitnessGearComponent =
+  updateFitnessGearComponent as jest.MockedFunction<
+    typeof updateFitnessGearComponent
   >
 
 const createComponent = (
@@ -277,6 +283,9 @@ describe('GearComponentsCard', () => {
     const onChanged = renderCard([createComponent()])
 
     fireEvent.click(screen.getByRole('button', { name: 'Retire Chain' }))
+    fireEvent.click(
+      screen.getByRole('button', { name: 'Confirm retire Chain' })
+    )
 
     await waitFor(() =>
       expect(mockRetireFitnessGearComponent).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
```diff
@@ -95,6 +101,14 @@ describe('GearComponentsCard', () => {
     mockRetireFitnessGearComponent.mockResolvedValue(
       createComponent({ removedAt: Date.UTC(2025, 5, 1) })
     )
+    // Without this default the unretire tests pass only on a neighbour's
+    // leaked implementation: `vi.clearAllMocks()` resets call history and
+    // leaves implementations in place, so whichever test last set one on this
+    // mock decides what the next test sees — including the rejection from
+    // "surfaces an unretire failure". Remove it and `--sequence.shuffle` fails.
+    mockUpdateFitnessGearComponent.mockResolvedValue(
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:109]: **Finding (round 1, high): unretire tests passed only on a leaked mock implementation**

`vi.clearAllMocks()` resets call history but leaves implementations in place. `mockUpdateFitnessGearComponent` had no `beforeEach` default, so the two unretire tests passed only because a neighbouring test had last set a resolved implementation that leaked forward. Under `--sequence.shuffle` they fail: when "surfaces an unretire failure" runs first, its rejection leaks into the refetch test, the component's catch branch fires, and `onChanged()` is never called.

CI does not shuffle, so this would have sat latent until someone reordered a test.

Fixed in 59d32556d — a `beforeEach` default matching the pattern the other three client mocks already use. Removing it again reproduces the failures.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
```diff
@@ -95,6 +101,14 @@ describe('GearComponentsCard', () => {
     mockRetireFitnessGearComponent.mockResolvedValue(
       createComponent({ removedAt: Date.UTC(2025, 5, 1) })
     )
+    // Without this default the unretire tests pass only on a neighbour's
+    // leaked implementation: `vi.clearAllMocks()` resets call history and
+    // leaves implementations in place, so whichever test last set one on this
+    // mock decides what the next test sees — including the rejection from
+    // "surfaces an unretire failure". Remove it and `--sequence.shuffle` fails.
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:108]: **Finding (round 2, info): cited shuffle seeds were environment-dependent**

The comment and PR description originally named the specific seeds that failed (3, 8, 10). Seed-to-order mapping varies by environment and vitest version — one reviewer could not reproduce those numbers in isolation.

Fixed in 522e4bcc6: the comment now states the property instead of the seeds — remove the default and a shuffled run fails.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
```diff
@@ -287,6 +304,165 @@ describe('GearComponentsCard', () => {
...
+      )
+    )
+    expect(onChanged).toHaveBeenCalled()
+  })
+
+  // Unretire is not armed. It is NOT the mirror of Retire — see the comment on
+  // `handleUnretire`: a delayed unretire credits every activity ridden while
+  // the part sat retired, and re-retiring does not take that back. It stays
+  // unarmed because arming adds friction to the misclick without preventing
+  // the misattribution.
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:401]: **Finding (round 2, low): the retracted claim survived in a second place**

Round 1 corrected the Unretire rationale in `GearComponentsCard.tsx`, but this test file still carried the identical retracted wording — "it restores state rather than destroying it, and a stray click is undone by one more click" — about 200 lines from its own correction, and directly beside the tests that exercise unretire, where a reader is most likely to trust it. The two comments contradicted each other within one PR.

Fixed in 522e4bcc6.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
```diff
@@ -287,6 +304,165 @@ describe('GearComponentsCard', () => {
...
+      screen.getByRole('button', { name: 'Show 1 retired component' })
+    )
+
+    const actions = screen
+      .getByRole('button', { name: 'Unretire Chain' })
+      .closest('div')
+    // Both tokens: `flex-wrap` does nothing without `display: flex`, and the
+    // enclosing `<td>` carries `whitespace-nowrap`, so a block container puts
+    // the two buttons on one line and overhangs the panel instead of wrapping.
+    expect(actions).toHaveClass('flex', 'flex-wrap')
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:450]: **Finding (round 2, low): the wrap guard passed with the defect present**

Round 1 added `expect(actions).toHaveClass('flex-wrap')`. That does not guard the behaviour: `flex-wrap` has no effect without `display: flex`, and the enclosing `<td>` carries `whitespace-nowrap`, so dropping only `flex` gives a block container whose inline-flex buttons cannot wrap.

Measured in a real browser against the fixed panel `dataColumnStyle` produces below the snap width: with `flex`, `display:flex`, buttons on two lines, 0 overflow. Without it, `display:block`, `whiteSpace:nowrap`, both buttons on one line, **18px overflow to the right** — exactly the overhang the test's own comment says must not happen — and the test still passed.

Fixed in 522e4bcc6: `toHaveClass('flex', 'flex-wrap')`. Each token now drops independently to a failure.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -199,17 +216,49 @@ export const GearComponentsCard: FC<Props> = ({
     }
   }
 
+  // Clearing `removedAt` reopens the ORIGINAL `[addedAt, now)` window. A
+  // component has one window and no per-period history, so this is not the
+  // mirror of Retire: undoing a retire seconds later credits nothing extra,
+  // but undoing one from last season credits every activity ridden while the
+  // part sat retired — and re-retiring does NOT take that back, because it
+  // only closes the window at the new now. Verified against the rollup query.
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:224]: **Finding (round 1, medium): the comment claimed a symmetry that does not exist**

The comment said Unretire "restores state rather than destroying it, and a stray click is undone by one more click". Verified false against the real rollup query with a throwaway database:

- component `addedAt` 2026-01-01, rides 10 km on Jan 15 and 20 km on Feb 15
- `removedAt` = 2026-02-01 → rollup `{10 km, 1}` — correct
- unretire (`removedAt: null`) → `{30 km, 2}`
- re-retire → still `{30 km, 2}`

A component has one `[addedAt, removedAt)` window and no per-period history, so a delayed unretire retroactively credits every activity ridden while the part sat retired, and re-retiring does not take that back — it only closes the window at the new `now`.

Fixed in 59d32556d: the comment now states this. The behaviour is unchanged (still unarmed — arming adds friction to the misclick this exists to recover from without preventing the misattribution), and the underlying schema gap is filed separately.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -531,7 +609,7 @@ export const GearComponentsCard: FC<Props> = ({
             // delete on the first click after the next "Show ...".
             onClick={() => {
               setShowRetired((current) => !current)
-              setConfirmingDeleteId(null)
+              setConfirmingActionId(null)
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:612]: **Finding (round 1, info): duplicated `setConfirmingActionId(null)`**

Two identical calls in a row, left behind by collapsing `confirmingDeleteId` and `confirmingRetireId` into one id. Behaviourally inert, but dead code that reads against the comment this PR adds explaining why there is only one id.

Fixed in 59d32556d.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -478,40 +527,69 @@ export const GearComponentsCard: FC<Props> = ({
...
-                          className="text-destructive"
-                          disabled={isPending}
-                          onClick={() => handleDelete(component.id)}
-                          // Leaving the button disarms it: an armed row that
-                          // stays armed is a destructive single click waiting
-                          // for whoever comes back to this table.
-                          onBlur={() => {
-                            if (confirmingDeleteId === component.id) {
-                              setConfirmingDeleteId(null)
+                      <div className="flex flex-wrap justify-end gap-1">
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:530]: **Finding (round 1, low): the wrap that keeps the actions from overhanging was asserted nowhere**

The PR description claimed the two buttons wrap rather than overflow, and nothing tested it — removing `flex-wrap` passed all 41 tests. jsdom does no layout, so the only feasible guard is asserting the classes that produce the behaviour.

Fixed in 59d32556d, then corrected again in 522e4bcc6 (the guard initially asserted only `flex-wrap`, which passes even with the overhang present — see the follow-up comment on `GearComponentsCard.test.tsx`).

llun (commented):
Posting verified findings from two prior sub-agent review rounds that were fixed on the branch but never posted at review time. All seven were confirmed against the diff and are already addressed as of 522e4bcc6.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:109]: Fixed in 59d32556d by giving `mockUpdateFitnessGearComponent` a `beforeEach` default.

Round 3 re-verified this independently rather than taking the fix on trust: with the default removed and `--sequence.shuffle` run over seeds 1–8, seeds 2 and 7 fail (`unretires a retired component and refetches`, `expected "vi.fn()" to be called at least once`); with the default restored all eight pass. That control is what attributes the failure to the missing default rather than to shuffle flakiness, so the fix is load-bearing.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:108]: Fixed in 522e4bcc6 by replacing the seed numbers with the property they were standing in for.

One correction from round 3 worth recording: the property is probabilistic, not universal — only orderings that place `surfaces an unretire failure` first fail. The commit body reads slightly stronger than that ("Remove it and `--sequence.shuffle` fails"). The behaviour is right and the guard holds; the prose overstates its reach.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:401]: Fixed in 522e4bcc6.

This one is the more useful finding of the two low-severity ones: round 1 retracted the claim in the component and I did not grep for the same sentence elsewhere, so the retracted wording sat in the test file 200 lines away, where it would have outlived the correction.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:450]: Fixed in 522e4bcc6 by widening the assertion to `toHaveClass('flex', 'flex-wrap')`.

Round 3 checked the fix in the direction it was made for: on the round-1 spelling, dropping `flex` while keeping `flex-wrap` left the test green (a block container overhanging by 18px). On the current spelling both tokens drop independently to a failure — mutation 8 fails with `Received: flex-wrap justify-end gap-1`, mutation 9 with the converse.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:224]: Fixed in 59d32556d.

The comment asserted that Unretire "restores rather than destroys". I checked that against the database instead of reasoning about it and it is false: a component retired at 10km, ridden to 30km, then unretired reads 30km, not 10km — the install window is reopened, not resumed. The comment now says what the code does. Behaviour is unchanged and the schema gap (no record of the interrupted window) is filed separately.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:612]: Fixed in 59d32556d — the duplicate call is gone.

Round 3 confirms the surviving one is guarded rather than merely present: removing `setConfirmingActionId(null)` from the show/hide-retired toggle fails `disarms a pending delete when the retired rows are hidden and shown again`.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:530]: Fixed in 522e4bcc6 by adding `lets the retired row actions wrap instead of overflowing`.

See the thread on the test itself for the mutation evidence — the assertion as first written would not have caught the defect it was added for.

PR #1569, opened 2026-08-26
fix: refuse a traversing media path when recovering it from a file URL

llun:
> **Severity:** this started as "hardening for an unreachable path". Round 1 of the review loop disproved that — see **Reachability**. It is an arbitrary local-file read a signed-in user can set up.

## What

`getMediaPathFromFileUrl` (`lib/services/medias/mediaFileUrl.ts`) `decodeURIComponent`s the segment it recovers from `/api/v1/files/<path>`, and a percent-encoded slash survives the WHATWG URL parser's dot-segment normalization. So `https://<our-host>/api/v1/files/..%2f..%2fsecrets/env` arrives at the decoder still spelled `..%2f` and comes back as `../../secrets/env`. The host-relative branch parses no URL at all, so there even a plain `/api/v1/files/../../secrets/env` was never normalized.

`copyProfileImage` in `scripts/backup/actorArchive.ts` joined that straight onto the staging directory:

```
path.join(stagingDir, MEDIA_ARCHIVE_DIR, storagePath)
```

and copied whatever it found into the archive as `avatar.<ext>` / `header.<ext>`.

## Reachability

`POST /api/v1/accounts/profile` declares `iconUrl: z.string().optional()` and `headerImageUrl: z.string().optional()` — bare strings, no URL parse, no host check, no media-id lookup — and spreads them straight into `database.updateActor`, which stores them verbatim. The route sits behind `AuthenticatedGuard`, so **any account can set its own avatar URL** to such a path. The next archive export of that account then ships an arbitrary file off the operator's machine inside it, as `avatar.bin`.

One nuance stated precisely: the export is **CLI-only** (`scripts/backup/exportActorArchive.ts`), so the read fires when an operator runs the export for that account — there is no self-service endpoint. That is a timing constraint, not a barrier: the payload is planted in advance and sits in the row until someone runs the tool for its intended purpose, and the resulting archive goes to the account holder.

The attachment route into the same traversal was already defanged — `normalizeStoragePath` drops any `..` path before `buildStoragePlan` reaches the filesystem — so `copyProfileImage` genuinely was the only unguarded sink.

*(The unvalidated `iconUrl` field is its own problem, independent of the archive: that value is also rendered as an `` for every profile viewer and federated in the actor's `icon.url`. Left as a follow-up.)*

## What changed

**The parser refuses the path** (`isTraversingStoragePath`), covering all three of its call sites rather than leaving each to remember — I checked them first, and every one already has a "not one of ours" branch. The check runs **after** decoding, since that is where the traversal reappears. Only a segment that *resolves* to `..` is refused, so `ab/..cd.webp` stays an ordinary stored file name.

Windows traversal is covered deliberately, because these scripts run wherever the operator runs them: `\` is a separator, `C:` is absolute, and a segment of bare dots can reach the parent — Node's own `path.win32` models none of that, so the containment check downstream wouldn't catch one either. The shape refused is deliberately wider than what Win32 actually collapses; nothing this instance stores is named out of dots, so over-refusing costs nothing. That is *traversal* coverage, not a claim that every Win32 quirk is handled — a reserved device name like `CON` still resolves to a device, as it already does for `LocalFileStorage.getFile`.

A NUL byte is refused too: it used to survive and abort the whole export from `archiveStorage`.

**`copyProfileImage` confirms containment for itself as well.** It is the step that turns a stored path into a file read off the operator's machine, and its signature says nothing about where that path came from — the same reason `createMediaTempFilePath` asserts its own result is still under `tmpdir()`. It `resolve`s rather than `join`s, which is what lets an absolute path be *seen* as outside instead of silently reinterpreted.

This cannot be left to the storage driver: `LocalFileStorage.getFile` does make the identical check, but **`S3FileStorage.getFile` makes none** — a traversing key is merely inert there, and with a CDN `hostname` configured it is string-concatenated into a redirect URL.

**The rule is exported and shared.** `scripts/maintenance/backfillMediaBlurhash.ts` now imports it instead of carrying its own weaker copy, which refused neither backslashes nor drive letters — the dedup #1560 owed. Only the traversal half is shared; the host halves stay separate.

## Tests

Both guards were written test-first and proven by reverting them:

- `lib/services/medias/mediaFileUrl.test.ts` — 14 refusal cases (encoded slash, upper-case `%2E%2E%2F`, traversal after a leading directory, plain traversal on the host-relative branch, backslash, three absolute spellings, drive letter on both branches, trailing-space and three-dot segments, NUL) plus cases pinning what must keep working: `ab/..cd..webp`, a `.` segment, `cd:ef.webp` (a colon that is not a drive), and a malformed escape whose undecoded `%2f` is inert to `path.join`.
- `scripts/backup/actorArchive.test.ts` — `copyProfileImage` gains its own suite. The traversal case plants a secret outside the staging directory and asserts it is *not* read into the archive; against the unguarded sink that test copies the secret to `avatar.bin` and returns the filename.

Mutation-tested rather than assumed: the `relativePath === '..'` disjunct had no test reaching it, and now does. `path.isAbsolute(relativePath)` is win32 cross-drive only — on POSIX `path.relative` between two absolute paths never returns an absolute string — so it carries a comment saying so instead of an unreachable test.

`copyProfileImage` is exported for this, following the precedent #1560 set for `registerAttachmentUrl` in the same file.

## Review

Five rounds of the sub-agent review loop (correctness, quality, testing, scope/regression, security). Round 1 found six issues, all confirmed by running the real code — including the reachability assessment above, which the original PR body got backwards. The traversal fix has had no findings since.

Rounds 2–5 kept finding gaps in an adjacent trusted-host wildcard control that this PR had picked up along the way. On two reviewers' recommendation that work is now **#1578**, based on this branch, so the traversal fix isn't held behind a control that was still sprouting gaps.

## Gate

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test` all green (884 files, 11598 tests). No UI change, so no browser verification. No migration, so no schema-dump regeneration.


lib/services/medias/mediaFileUrl.ts
```diff
@@ -38,6 +38,31 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
...
+ * segment is refused: `ab/..cd.webp` is an ordinary file name, and rejecting
+ * every path that merely contains two dots would turn one into a lost file.
+ *
+ * `scripts/maintenance/backfillMediaBlurhash.ts` carries its own copy of this
+ * rule (`isTraversingStoragePath` there) because it landed first with the same
+ * bug; the two agree so that dropping the copy for this helper stays a
+ * deletion rather than a behaviour change.
+ */
+const isTraversingStoragePath = (mediaPath: string) =>
+  /^[/\\]/.test(mediaPath) || mediaPath.split(/[/\\]/).includes('..')
```
llun: **Windows drive-letter absolute paths defeat the guard**

The guard only recognises a POSIX-style leading `/` or `\` as absolute, so a drive-letter path is returned as a "safe" storage path — contradicting this helper's own doc comment ("or an absolute decoded path, is never one of our storage paths").

Verified against the real code:
```
getMediaPathFromFileUrl('/api/v1/files/C:\Windows\System32\config\SAM', config)
  -> "C:\\Windows\\System32\\config\\SAM"
getMediaPathFromFileUrl('https://llun.test/api/v1/files/C:\Windows\...', config)
  -> "C:/Windows/System32/config/SAM"
```
`copyProfileImage` is saved by its own containment check, but `registerAttachmentUrl` is not: the path flows to `buildStoragePlan`/`archiveStorage`, whose `normalizeStoragePath` only rejects a literal `..` substring and never calls `path.isAbsolute`, and `path.win32.resolve(base, 'C:/...')` discards the base entirely.

Also unhandled by the same helper, same class:
- a NUL byte (`/api/v1/files/ab%00.webp` -> `"ab\u0000.webp"`) survives and aborts the whole export from `archiveStorage` (a DoS on the operator's tool);
- Win32 strips trailing dots/spaces per component, so `..%20%2f..%20%2f` -> `".. /.. /"` is refused by neither this helper nor `copyProfileImage`'s lexical check, and Win32 would open it as the parent directory.

lib/services/medias/mediaFileUrl.ts
```diff
@@ -38,6 +38,31 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
...
+ * happen AFTER decoding.
+ *
+ * A backslash counts as a separator because Windows resolves `..\..` the same
+ * way, and nothing this instance stores has one in a path. Only a whole `..`
+ * segment is refused: `ab/..cd.webp` is an ordinary file name, and rejecting
+ * every path that merely contains two dots would turn one into a lost file.
+ *
+ * `scripts/maintenance/backfillMediaBlurhash.ts` carries its own copy of this
+ * rule (`isTraversingStoragePath` there) because it landed first with the same
+ * bug; the two agree so that dropping the copy for this helper stays a
```
llun: **The "the two agree" claim is false, and three docs repeat it**

`scripts/maintenance/backfillMediaBlurhash.ts`'s `isTraversingStoragePath` splits on `/` only and has no backslash awareness. Verified against both real functions:
```
url = 'https://llun.test/api/v1/files/..%5c..%5csecrets'
getMediaPathFromFileUrl(url, config)   -> null              (refused)
getLocalStoragePath(url, ['llun.test']) -> "..\\..\\secrets" (NOT refused)
```
The same false claim is repeated in `AGENTS.md` and `CLAUDE.md` ("kept in agreement deliberately so that collapsing them stays a deletion rather than a behaviour change"). A future reader could use it to justify a merge that is not a no-op.

lib/services/medias/mediaFileUrl.ts
```diff
@@ -38,6 +38,31 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
   }
 }
 
+/**
+ * A decoded segment that walks upwards, or an absolute decoded path, is never
+ * one of our storage paths — both storage drivers refuse such a path anyway.
```
llun: **Overstates S3 — that driver has no containment check**

`LocalFileStorage.getFile` does make this check (`lib/services/medias/localFile.ts`). `S3FileStorage.getFile` does not: it passes the key straight to `GetObjectCommand`, and when a CDN `hostname` is configured it string-concatenates into `https://${hostname}/${filePath}` with no validation. Such a key is merely inert, not refused. The same sentence was added to `AGENTS.md`.

scripts/backup/actorArchive.ts
```diff
@@ -633,7 +638,29 @@ const copyProfileImage = async ({
...
+  // for itself rather than trusting a caller three hundred lines away, the
+  // same reason `createMediaTempFilePath` asserts its own result is still
+  // under `tmpdir()`. Resolving rather than joining is what lets an absolute
+  // path be seen as outside instead of being silently reinterpreted as a
+  // relative one.
+  const relativePath = path.relative(mediaDir, source)
+  if (
+    relativePath === '..' ||
+    relativePath.startsWith(`..${path.sep}`) ||
+    path.isAbsolute(relativePath)
```
llun, line 662: **Two of three disjuncts are unexercised by the new tests**

Mutation-confirmed: deleting `relativePath === '..'`, or `path.isAbsolute(relativePath)`, or both, leaves all 34 tests in `scripts/backup/actorArchive.test.ts` green.

On POSIX `path.relative()` between two absolute paths never returns an absolute string, so the test row named "is an absolute path outside the archive" is actually caught by the `startsWith('..' + path.sep)` branch and adds no branch coverage. `path.isAbsolute(relativePath)` is only reachable on Windows cross-drive (`path.win32.relative('C:\\staging\\media','D:\\secrets\\env')` returns `'D:\\secrets\\env'`) — it is not dead code, but it needs a comment saying so rather than a test. `relativePath === '..'` IS reachable on POSIX (a `storagePath` of exactly `'..'`) and should get its own test row.

scripts/backup/actorArchive.ts
```diff
@@ -618,7 +618,12 @@ const relocateStorageDir = async (
   await fs.rename(source, destination)
 }
 
-const copyProfileImage = async ({
+/**
+ * Copies one profile image out of the archive's media directory to the
+ * archive root, where the Mastodon format expects `avatar.<ext>` /
+ * `header.<ext>`.
+ */
+export const copyProfileImage = async ({
```
llun, line 626: **The PR's "not reachable today" assessment is wrong — this is a live vulnerability**

`POST /api/v1/accounts/profile` (`app/api/v1/accounts/profile/route.ts`) declares `iconUrl: z.string().optional()` and `headerImageUrl: z.string().optional()` — bare strings, no URL parse, no host check, no media-id lookup — and spreads them straight into `database.updateActor`, which stores them verbatim (`lib/database/sql/actor.ts`, `...(iconUrl ? { iconUrl } : null)`). The route is behind `AuthenticatedGuard`, so ANY signed-in local user can set their own actor's `iconUrl`.

End-to-end on `origin/main`: a user sets `iconUrl` to `https://<our-host>/api/v1/files/..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2fetc/hosts`, then requests an account export; `copyProfileImage` resolves the source to `/etc/hosts` and ships it inside their own archive as `avatar.bin`.

So the fix in this PR is a security fix, not hardening, and the PR body's Reachability section needs correcting. (Separately worth a follow-up issue: an unvalidated `iconUrl` is also rendered as `` for every viewer and federated in the actor's `icon.url`, which is its own third-party-URL problem independent of the archive.)

llun:
**Round 1 finding — NOT postable as an inline review comment (see note below)**

`lib/utils/host.ts:248` (`isOwnInstanceHost`, the exact-authority pass: `rules.some((rule) => canonicalAuthority(rule) === authority)`)

**A literal `*.`-prefixed authority passes the own-host check**

`canonicalAuthority` compares rules literally, and `new URL()` accepts `*` in a host, so a rule spelled `*.cdn.llun.test` matches an authority spelled `*.cdn.llun.test`. Verified with this PR's own test config:
```
'https://evil.example/api/v1/files/ab/cd.webp'      -> null
'https://*.cdn.llun.test/api/v1/files/ab/cd.webp'   -> "ab/cd.webp"
'https://*.CDN.llun.test/api/v1/files/secret.jpg'   -> "secret.jpg"
```
This is pre-existing (introduced by #1560) and re-opens exactly the hole `scripts/maintenance/backfillMediaBlurhash.ts` documents fixing in #1559. `isOwnInstanceHost`'s only production consumer is `getMediaPathFromFileUrl`, i.e. this PR's own module, so it is inside this change's blast radius. A wildcard rule should never be compared literally — the `isHostTrustedByRules` pass immediately after already handles it correctly.

---
Note: `lib/utils/host.ts` has no diff in this PR (confirmed via `git diff <merge-base>..fff121d5e -- lib/utils/host.ts`, and the GitHub review-comments API rejects it with `"Path could not be resolved"`), so this cannot be attached as a resolvable inline thread on this PR. Posting here as a top-level comment so it isn't lost; recommend a follow-up PR/issue to fix `isOwnInstanceHost` and add a regression test.

lib/services/medias/mediaFileUrl.ts
llun: Fixed in ca3dd26f4. `isTraversingStoragePath` now refuses a drive-letter prefix as absolute, alongside `/` and `\`. All three of the extra cases you raised are covered: the NUL byte (which was aborting the whole export from `archiveStorage`, not just failing one copy) and the Win32 trailing-dot/space stripping, via an `isUpwardSegment` helper that refuses any segment made only of dots and spaces carrying two or more dots — so `.. `, `...` and `. .` go with `..`, while a lone `.` stays legal. New cases in `mediaFileUrl.test.ts` cover the drive letter on both the absolute and host-relative branches, the trailing-space and three-dot spellings, and the NUL; `cd:ef.webp` pins that an ordinary colon is still not a drive letter.

lib/services/medias/mediaFileUrl.ts
llun: You were right, and the fix is to stop making the claim rather than to restate it. `isTraversingStoragePath` is now exported from `lib/services/medias/mediaFileUrl.ts` and `scripts/maintenance/backfillMediaBlurhash.ts` imports it, so the traversal rule is single-sourced — there is no parity to assert. That also tightens the backfill, which refused neither backslashes nor drive letters; new cases in its test file pin the newly-refused spellings. Only the traversal half is shared: as the security lens noted, its `isOwnAuthority` is deliberately stricter than `isOwnInstanceHost` and sweeps remote-controlled `attachments.url` rows, so AGENTS.md/CLAUDE.md/REVIEW.md now say the owed host dedup must tighten `isOwnInstanceHost` rather than replace `isOwnAuthority` with it.

lib/services/medias/mediaFileUrl.ts
llun: Fixed in ca3dd26f4. I had propagated that sentence from the backfill's own comment without checking it. The JSDoc, AGENTS.md and REVIEW.md now say what is actually true: `LocalFileStorage.getFile` makes the containment check, `S3FileStorage.getFile` makes none — a traversing key is merely inert there, and with a CDN `hostname` configured it is string-concatenated into a redirect URL — which is precisely why the decision has to be made in the parser rather than deferred to 'the driver will reject it'. REVIEW.md carries that as its own checklist line now.

scripts/backup/actorArchive.ts
llun: Fixed in ca3dd26f4, and I reproduced your mutation result before and after. Added a `buildStoragePath: () => '..'` row; deleting the `relativePath === '..'` disjunct now fails exactly that test and nothing else. For `path.isAbsolute(relativePath)` I took your recommendation and did not spend a row on it — I confirmed with `path.win32.relative('C:\\staging\\media','D:\\secrets\\env')` returning `'D:\\secrets\\env'` that it is genuinely win32 cross-drive only, and it now carries a comment saying it cannot be covered here rather than looking like untested code. Also took the `$description` truncation note: the row descriptions are now under the 40-char cap, so `vitest -t` matches the printed title.

scripts/backup/actorArchive.ts
llun: Confirmed and corrected — this was the most important finding in the round. I verified the route myself: `iconUrl`/`headerImageUrl` are bare `z.string()`, spread into `updateActor`, stored verbatim, behind `AuthenticatedGuard`. The PR body's Reachability section has been rewritten to say this is a live arbitrary-file read exploitable by any signed-in local user, and the commit message for ca3dd26f4 records that the previous commit's assessment was wrong. AGENTS.md, CLAUDE.md and REVIEW.md now all say the same, so the invariant is not remembered as a hardening exercise. Filing the unvalidated-`iconUrl`-as-``/federated-`icon.url` problem as a separate follow-up, since it is independent of the archive.

llun:
Fixed in ca3dd26f4 rather than deferred. `isOwnInstanceHost`'s only production consumer is `getMediaPathFromFileUrl` — this PR's own module — so the hole was inside this change's blast radius, and leaving it meant the hardened parser still accepted a spoofed authority. The exact-authority pass now skips any rule containing `*`; `isHostTrustedByRules` immediately after already expands wildcards correctly. Verified non-vacuous by mutation: reverting the one-line guard fails three tests (`lib/utils/host.test.ts` for the rule spelled literally and in another case, plus the `mediaFileUrl.test.ts` case for a URL whose authority is spelled as the rule). Your warning about the dedup direction is now recorded in AGENTS.md, CLAUDE.md and REVIEW.md: the owed consolidation must tighten `isOwnInstanceHost`, never replace `isOwnAuthority` with it, precisely because the backfill sweeps remote-controlled `attachments.url` rows.

lib/utils/host.ts
```diff
@@ -245,7 +258,21 @@ export const isOwnInstanceHost = (
...
+  // a host, so comparing `*.cdn.example` against itself would let a URL
+  // spelling that authority pass as one of ours — the hole
+  // `scripts/maintenance/backfillMediaBlurhash.ts` documents closing for its
+  // own matcher. Wildcards belong to the rules pass below, which expands them.
+  //
+  // This pass reads the RAW rules, so it needs its own skip: `normalizeHost`
+  // refusing a misplaced `*` covers the rules pass, not this one.
+  if (
+    rules.some(
+      (rule) => !rule.includes('*') && canonicalAuthority(rule) === authority
```
llun: Round 1's wildcard fix is incomplete. Skipping `*`-bearing rules in the exact pass rests on "wildcards belong to the rules pass, which expands them" — true only of the documented `*.example.com` form. `normalizeHost` derives `hasWildcard` from `startsWith('*.')` alone, so every other spelling survives normalization as a LITERAL hostname carrying a `*`, which `hostMatchesRule` then compares literally. `new URL()` percent-decodes `%2a` in an authority, so a caller can spell one exactly. Verified against the real `getMediaPathFromFileUrl` with a `https://<authority>/api/v1/files/ab/cd.webp` URL per rule:

```
*.cdn.llun.test   -> null   (the one round 1 covered)
*llun.test        -> "ab/cd.webp"
foo.*.llun.test   -> "ab/cd.webp"
cdn.llun.test.*   -> "ab/cd.webp"
cdn.*             -> "ab/cd.webp"
*-cdn.llun.test   -> "ab/cd.webp"
```

`*llun.test` is a plausible typo of the documented `*.example.com` form and fails silently today, so an operator would not notice it. The backfill's `isOwnAuthority` has the same gap, which makes the "stricter matcher" claim true only for the `*.`-prefixed case.

lib/services/medias/mediaFileUrl.ts
```diff
@@ -38,6 +38,58 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
   }
 }
 
+// A segment of nothing but dots and spaces is refused whenever it carries two
+// of those dots. Windows normalises trailing dots away, so `.. ` and `...` can
+// reach the parent directory a bare `..` names — and Node's own `path.win32`
+// does NOT model that, so neither `path.resolve` nor the containment check
+// downstream would see it. Refusing the whole shape is deliberately wider than
```
llun, line 45: The comment states as a blanket fact that "Win32 strips trailing dots and spaces from a path component," making `.. `, `...` and `. .` act as `..` "there". Neither lens could verify that on a real Windows host, and the public research is narrower: trailing-dot stripping is described as applying per element, but trailing-space stripping only to the last element of the path, and an all-dots name strips toward an empty component rather than literally becoming `..`. Separately confirmed by running Node directly: `path.win32.join`/`resolve` do NOT special-case `...` or `.. ` at all, so this check is stricter than Node's own path arithmetic — which is the safe direction, but not what the comment claims. Same text is restated in AGENTS.md, CLAUDE.md and REVIEW.md.

llun:
Two round-2 observations accepted without a code change, recorded so they are not silently dropped.

**`rule.includes('*')` is a raw substring test that also kills the `https://example.com/*` spelling.** Previously `canonicalAuthority` reduced that to `example.com` and matched; now the exact pass skips it. It fails closed and `normalizeHost` already returned null for it (`hasOnlyAuthority` rejects the `/`), so such a rule matched nothing through the other pass either. `/*` is not a documented spelling in `docs/environment-variables.md`, so this is left as-is.

**Reserved Win32 device names (`CON`, `NUL`, `COM1`) pass both the guard and `copyProfileImage`'s containment check.** On a Windows operator machine an `iconUrl` ending `/api/v1/files/CON` would make `fs.copyFile` read the console and hang the export. It is not an escape, it is Windows-only, and it is pre-existing in `LocalFileStorage.getFile` too — so rather than chase it here, the PR's Win32 claim has been narrowed to say explicitly that it covers traversal and not every Win32 quirk.

Also corrected in the PR body from the round-2 security review: the export is CLI-only (`scripts/backup/exportActorArchive.ts`), so the arbitrary-file read fires when an operator runs an export for the account, not on a self-service request. That is a timing constraint rather than a barrier — the payload is planted in advance and the resulting archive goes to the account holder — but the earlier wording implied an endpoint that does not exist.

lib/utils/host.ts
llun: Fixed in 3d8f6b6f6, at normalization rather than in the pass — `normalizeHost` now returns null for any parsed hostname still containing `*`. That covers the rules pass and both matchers; nothing legitimate reaches that line with a `*`, since the documented form has already had its prefix stripped off `hostToParse`. Both guards are kept and neither subsumes the other: the exact pass reads RAW rules, the rules pass normalizes first — the comments now say so. All six spellings above now return null, and the fix is mutation-proven: removing the one line fails four new cases in `lib/utils/host.test.ts`. The typo now surfaces as a logged invalid `ACTIVITIES_TRUSTED_HOSTS` entry instead of a silent rule matching one attacker-chosen authority.

lib/services/medias/mediaFileUrl.ts
llun: Fixed in 3d8f6b6f6 in all four places. The prose no longer asserts per-component Win32 stripping semantics as fact. It now says what is actually verifiable: Windows normalises trailing dots away, Node's `path.win32` models none of it (so the containment check downstream would not catch one either), and the shape refused is DELIBERATELY wider than what Win32 collapses — nothing this instance stores is named out of dots, so over-refusing costs nothing and does not depend on getting the platform's rules exactly right. Also added the scope limit your security counterpart raised: this is traversal coverage, not a claim that every Win32 quirk is handled — a reserved device name like `CON` still resolves to a device, as it already does for `LocalFileStorage.getFile`.

lib/utils/host.ts
```diff
@@ -90,9 +90,18 @@ export const normalizeHost = (
...
+  // The wildcard marker sits on the AUTHORITY, which an operator may write
+  // behind a scheme. Reading it off the raw value missed
+  // `https://*.edge.example.com`, which used to work only by accident: it fell
+  // through as a literal hostname that `getHostParts` then re-read as a
+  // wildcard. The `*` refusal below closes that accident, so the marker has to
+  // be recognised properly here or a scheme-prefixed wildcard rule stops
+  // matching its own subdomains.
+  const scheme = /^[a-z][a-z0-9+.-]*:\/\//i.exec(firstHost)?.[0] ?? ''
+  const authority = firstHost.slice(scheme.length)
+  const hasWildcard = authority.startsWith('*.')
```
llun: **Round 2 regression** — `hasWildcard` in `normalizeHost` was read off the RAW value, so `https://*.edge.example.com` never took the wildcard branch — it parsed to the literal `*.edge.example.com`, which `getHostParts` then re-read as a wildcard, so it matched subdomains by accident. Round 2's `hostname.includes('*')` kills that accident and the rule with it:

```
rules=['https://*.edge.example.com'] host='a.edge.example.com'   OLD=true NEW=false
resolveAuthBaseURL:  OLD https://a.edge.example.com -> NEW https://example.com
```

Blast radius on that config: the wrong WebAuthn rpID/origin for that domain (passkeys stop verifying), `hasSameOriginProof` rejects legitimate mutations from it, and `isOwnInstanceHost` stops recognising its own media URLs — while `buildTrustedOrigins` still lists it, so the two disagree.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -130,10 +133,15 @@ const isOwnAuthority = (
   // `https://*.<trusted-domain>/api/v1/files/<path>` match the rule's own
   // spelling and read an attacker-chosen path straight out of local storage.
   const candidate = authority.trim().toLowerCase()
-  if (!candidate || candidate.startsWith('*.')) return false
+  // `includes`, not `startsWith('*.')`: a rule is only a wildcard in the
+  // documented `*.example.com` spelling, so `*example.com`, `cdn.*` and
+  // `foo.*.example.com` reach here as LITERAL rules carrying a `*` — and
+  // `new URL()` percent-decodes `%2a` in an authority, so a federated
+  // attachment URL can spell one exactly and be read out of local storage.
+  if (!candidate || candidate.includes('*')) return false
```
llun: The fallback guards only a LEADING `*.`, so every other wildcard spelling reaches it as a literal rule. Proven with the script's own exported `getLocalStoragePath`, rule `*cdn.llun.test`:

```
authority %2acdn.llun.test  -> app getMediaPathFromFileUrl = null
                            -> script getLocalStoragePath = 'media/other-users-private.webp'
```

Same for `cdn.*`, `foo.*.cdn.example`, `cdn.example.*`, `*-cdn.example`. `attachments.url` is federated attacker input; the sweep reads the named path out of local storage and writes its blurhash onto the attacker's own row. Not a regression — but AGENTS.md, CLAUDE.md and REVIEW.md now all assert `isOwnAuthority` is the stricter matcher, and make that the reason the host halves stay duplicated. That claim is false for exactly this class.

lib/services/auth/trustedOrigins.ts
```diff
@@ -16,6 +16,21 @@ export const buildTrustedOrigins = (
...
+    // better-auth routes any pattern containing `*` through `wildcardMatch`,
+    // so a misplaced wildcard becomes a GLOB rather than an inert literal:
+    // `*example.com` trusts `evilexample.com`, and `isTrustedOrigin` gates the
+    // Origin check on state-changing auth requests as well as `callbackURL` /
+    // `redirectTo`, which makes it an open redirect carrying auth callbacks.
+    // Only the documented `*.example.com` spelling is a wildcard; anything
+    // else is an operator typo, and `new URL()` will not throw on it, so it
+    // has to be refused here rather than by the catch below.
+    const authority = host.replace(/^[a-z][a-z0-9+.-]*:\/\//i, '')
+    if (authority.replace(/^\*\./, '').includes('*')) {
```
llun: `buildTrustedOrigins` bypasses `normalizeHost` entirely and builds origins with its own `new URL(...).origin`, so a typo'd entry `*example.com` becomes the origin `https://*example.com` — and better-auth's `matchesOriginPattern` routes any pattern containing `*` through `wildcardMatch`. Proven against the installed better-auth:

```
true   https://*example.com        https://evilexample.com
true   https://*example.com        https://evilexample.com/steal?token=x
true   https://cdn.*               https://cdn.evil.test
true   https://foo.*.example.com   https://foo.evil.example.com
```

`isTrustedOrigin` gates the Origin check on state-changing auth requests AND `callbackURL`/`redirectTo`/`errorCallbackURL`/`newUserCallbackURL`. So an attacker registers `evilexample.com` and gets CSRF-origin acceptance plus an open redirect carrying auth callbacks — strictly worse than the media-path hole this PR closes, on the same operator typo.

llun:
Round 3's other item, the false "logs it as an invalid entry" claim added in round 2, is fixed in b43a411d9 across `lib/utils/host.ts` and AGENTS.md. It logged nothing: `normalizeHostRules` flat-maps the null away, and `buildTrustedOrigins`' catch only fires when `new URL()` throws, which `*example.com` does not. The prose now says the drop is silent and that a per-request path behind a cache is the wrong layer to log from — with the exception of `buildTrustedOrigins`, which after this round genuinely does warn.

That is the third prose claim in this PR that review falsified by running the code (after "the two copies agree" and "not reachable today"). Recorded here because the pattern is the finding: a comparative or reachability claim written into AGENTS.md is exactly the kind a future reader relies on to skip work, so it needs executing before it is committed.

lib/utils/host.ts
llun: Fixed in b43a411d9. The marker is read off the AUTHORITY now — an optional scheme is split off first and re-spliced onto `hostToParse` — so the spelling is a wildcard on purpose rather than by accident. Three cases in `lib/utils/host.test.ts` pin the bare, scheme-prefixed and ported forms; reverting the detection fails the scheme-prefixed one.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in b43a411d9 — the fallback tests `includes('*')` on both the candidate and the rule. Worth noting for the record that this is confined to the fallback: `isOwnAuthority`'s primary path already expands wildcards correctly through `normalizeHost`/`hostMatchesRule`, and the fallback only runs when `normalizeHost` refuses the authority (loopback), so legitimate wildcard matching is untouched. Three cases in the script's test file cover the spellings; reverting fails all three. I also dropped the "stricter matcher" claim from all three docs rather than restore it — it was only ever true of this wildcard gap, which is now closed on both sides, so the owed dedup needs measuring rather than an assumed ordering.

lib/services/auth/trustedOrigins.ts
llun: Fixed in b43a411d9. `buildTrustedOrigins` now refuses anything but a leading `*.` on the authority (checked after stripping an optional scheme) and skips it through the `logger.warn` path already beside it, so the typo is now the one place in this codebase where it IS surfaced to an operator. Four spellings plus two keep-cases are covered in `lib/services/auth/trustedOrigins.test.ts`; reverting fails all four drops. Fixing this here rather than deferring was deliberate: it is the same input and the same typo as the other two, and it is the most dangerous of the three consumers — hardening two and leaving this one would have been worse than fixing none.

llun:
**Restructured after review.** This branch was rewritten to hold only the path-traversal fix; the trusted-host wildcard work it had accumulated is now #1578, based on this branch.

Why: rounds 2–5 of the review loop each found a new gap in the wildcard control (twice in a fix from the round before), while the traversal fix had no findings after round 1. Two independent scope reviewers recommended the split, so the live traversal hole isn't held behind a control that was still moving. Reverting one no longer takes the other with it.

The 11 review threads above are all resolved but now point at rewritten history, so GitHub renders them as outdated. Their content still applies — every finding was addressed, and the fixes are in this branch or in #1578. For the record, the six round-1 findings and their outcomes:

1. Windows drive-letter paths defeated the guard → fixed here, with NUL and Win32 bare-dot segments.
2. My claim that the two `isTraversingStoragePath` copies agreed was false → removed the need for the claim by exporting the rule and having the backfill import it.
3. "Both storage drivers refuse such a path anyway" overstated S3 → corrected; `S3FileStorage.getFile` has no containment check at all.
4. Two of three disjuncts in `copyProfileImage`'s guard were unexercised → `'..'` now has a test; `isAbsolute` is win32 cross-drive only and carries a comment saying so.
5. The wildcard-authority hole in `isOwnInstanceHost` → moved to #1578.
6. "Not reachable today" was wrong → `iconUrl` is a bare `z.string()` any signed-in user can set. The PR body now says so.

Worth recording: four of those six were false claims I had written into the PR body or AGENTS.md, each caught by an agent executing the claim rather than reading it.

PR #1570, opened 2026-08-26
fix: consolidate the media-URL host check onto the shared helper

llun:
## Summary

`scripts/maintenance/backfillMediaBlurhash.ts` carried its own copy of "what stored media path does this URL name, and is it ours" — added in #1559, before #1560 landed the shared `getMediaPathFromFileUrl` for the same question.

#1569 shared the **traversal half** and named the host half as the remaining dedup, on the condition that the two matchers be *measured* rather than assumed to agree. They were. They differ in exactly one way, described below. So the script now calls the shared parser, and its `getLocalStoragePath`, `getOwnPathname`, `isOwnAuthority`, fallback-host regex chain and private `MEDIA_FILE_URL_PATH` are all deleted.

> **This PR was rebased after #1569 and #1578 merged, and shrank a lot as a result.** (Rebased again since, onto #1573, #1575, #1576 and #1577 — those needed only doc reconciliation, described at the bottom.) It originally also fixed a path-traversal gap and a wildcard-as-authority hole in `isOwnInstanceHost`. Both landed on main first, in stronger forms than mine — #1569's traversal check covers Windows separators, drive letters, `...`/`.. ` trailing-dot collapse and NUL bytes; #1578 fixed the wildcard hole at every consumer, including a scheme-prefixed spelling and two auth consumers I never looked at. My versions of both are dropped rather than reconciled. What remains is the consolidation this PR was actually asked for.

## The one behaviour delta

`isOwnInstanceHost` strips **both** default ports, where the deleted copy's non-loopback branch went through `hostMatchesRule` and treated only `:443` as implied. So `https://<our-host>:80/api/v1/files/…` is now accepted by the sweep where it was previously refused.

Same hostname either way, only ever our own host, documented as deliberate in `host.ts`, and now pinned by a test. That is the entire difference — every other input class the deleted suite exercised (traversal, wildcard-trusted subdomains, the loopback dev host, protocol-relative and query-string smuggling, foreign hosts) answers identically.

## What else moved

- **`canonicalAuthority` → exported `getCanonicalAuthority`.** Normalising the configured host into the authority `getMediaFileUrl` mints a URL on was the *fourth* copy of that same regex chain; `buildInstanceHosts` calls the helper now instead of respelling it.
- **`MEDIA_FILE_URL_PATH` exported** for the narrower question the script still asks: whether a stored `thumbnailUrl` is a host-relative media URL an earlier run wrote.
- **`getConfiguredHost` gains the doc comment it never had** — it falls back to the RAW value whenever `normalizeHost` refuses one, so it must not decide trust. It sits one line from `getCanonicalAuthority` and is the easy wrong reach.
- **`getAttachmentMediaHost` stays.** Which host an attachment's media URL is minted on is a different question, and it reads `actorId`, not the configured host list.

## Coverage

Moved, not dropped. The deleted `getLocalStoragePath` suite's traversal cases were already a subset of the shared parser's own; its wildcard-trusted-subdomain, loopback and host-matching cases are covered there too. The three misplaced-wildcard spellings #1578 added to it are **re-pinned against `getMediaPathFromFileUrl`**, which refuses them one layer further in — that check would otherwise have died with `isOwnAuthority`.

New alongside them:

- the six smuggled-authority prefixes a leading-characters test misses (`/\`, `/\\`, `/<TAB>/`, `/<LF>/`, `/<CR>/`, `/<TAB>\`) — the WHATWG parser grows an authority out of each, and the raw-string branch must keep refusing them if it is ever rewritten to parse;
- a foreign host carrying our media route inside a query string;
- both default-port spellings of our own host, pinning the delta above;
- an unconfigured instance, since `getTrustedHostRules` always contributes `config.host`;
- a `selectHeaderHost` guard, so the **inbound** half of #1578's wildcard refusal cannot be tightened past the documented `*.` form unnoticed — `normalizeHost` governs every `Host`/`X-Forwarded-Host` decision, not just this helper, and nothing exercised it through that entry point.

The script keeps two wiring tests proving the trusted-host list actually reaches the shared parser and that its answer is what storage is asked for.

Full gate green after the rebase: prettier → lint (0 errors) → typecheck (0) → build → 887 files, 11697 tests passing. No UI change, so no screenshots.

## Docs

`AGENTS.md`, `CLAUDE.md` and `REVIEW.md` said the host halves stay separate and enumerated `isOwnAuthority` as one of four consumers refusing a misplaced wildcard. Both are now false: the halves are consolidated, and that consumer is gone — the sweep inherits `isOwnInstanceHost`'s guards instead. Corrected in all three, with the measurement result recorded so the next reader does not have to redo it. The later rebase onto #1573 and #1575 needed no further correction here — this branch never claimed anything about `LocalFileStorage.deleteFile`'s containment or about `deleteSavedMedia`, both of which those PRs changed, so main's new **A stored path is confined to the storage root** section carries over untouched.

## Review

Seven sub-agent review rounds ran against the pre-rebase branch; 15 threads, all resolved. Those rounds reviewed the traversal and wildcard work that has since been dropped in favour of main's, so **the review loop is owed another pass over what actually remains** — the consolidation, the port delta, and the moved tests.

Three adjacent issues surfaced by those rounds are already fixed or in flight elsewhere: #1572, #1573, #1575.


lib/utils/host.ts
```diff
@@ -215,12 +235,46 @@ export const isHostTrustedByRules = (
...
+  // a rule stops being matched literally. `isHostTrustedByRules` still applies
+  // it as a pattern, because it re-derives the rules from the config and
+  // `normalizeHost` accepts the documented leading `*.` there.
+  //
+  // The test is `*` ANYWHERE, not the `*.` prefix: an operator who wrote an
+  // undocumented `*llun.test` or a bare `*` had it compare equal to a URL
+  // spelled the same way, and a prefix test does not see those. `normalizeHost`
+  // refuses the same class, which is what stops the rules pass from matching
+  // them; this stops the exact pass.
+  return authority.includes('*') ? '' : authority
```
llun: **Wildcard guard only checked a leading `*.`.** `new URL` parses `*` anywhere in a host, so an operator's undocumented `ACTIVITIES_TRUSTED_HOSTS` entry like `*llun.test` (or a bare `*`) still compares equal to a URL spelled the same way through this exact-authority pass. Separately, `**.edge.llun.test` matches a `*.edge.llun.test` rule through `hostMatchesRule`'s suffix test, which no change to this function alone can close. Suggest blanking here on `authority.includes('*')`, and refusing `*` anywhere in `normalizeHost` so the rules pass is closed too.

lib/utils/host.ts
```diff
@@ -215,12 +235,46 @@ export const isHostTrustedByRules = (
...
     .toLowerCase()
     .replace(/:(?:80|443)$/, '')
+
+  // A `*.` value is a wildcard RULE — a pattern, never an authority — and only
+  // `hostMatchesRule` knows how to read one. `new URL` parses `*` in a host
+  // quite happily, so without this `https://*.<trusted-domain>/…` compared
+  // equal to the rule's own spelling and was answered for as our own.
+  //
+  // That bought an attacker no NEW path, because `getTrustedHostRules` always
+  // contributes the configured host, so `https://<configured-host>/…` was
```
llun: **Impact claim looks inflated.** This doc comment (and the same claim in `AGENTS.md`/`CLAUDE.md`/`REVIEW.md`/the PR body) says the wildcard hole "handed an attacker-chosen storage path to whichever caller asked." But `getTrustedHostRules` always contributes `config.host`, so `https://<configured-host>/api/v1/files/<path>` was already reachable by anyone who could write a URL — the wildcard spelling grants no new path. Please reword this (and the other docs) as a latent trap for a caller that *uses* the host, not an exploit.

lib/services/medias/mediaFileUrl.ts
```diff
@@ -38,6 +63,25 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
...
+// media path. `new URL` normalises a LITERAL `../` out of a pathname, but a
+// percent-encoded one survives it and `decodeURIComponent` puts it back — so
+// the check has to happen after decoding. Backslashes are folded first because
+// that same parser treats one as a separator for a special scheme, and
+// `%5C` reaches the decoder intact.
+//
+// This is the layer that stops a caller ASKING for such a path, and it cannot
+// be left to the consumers: only `LocalFileStorage.getFile` re-checks
+// containment (its `deleteFile` does not, and an S3 key is taken literally),
+// while `copyProfileImage` in `scripts/backup/actorArchive.ts` `path.join`s the
```
llun: **"The storage drivers refuse such a path too" is inaccurate.** Only `LocalFileStorage.getFile` re-checks containment — `LocalFileStorage.deleteFile` has none, and an S3 key is taken literally. More importantly, `copyProfileImage` in `scripts/backup/actorArchive.ts` does a `path.join` under the staging directory with no containment check at all. Please narrow the comment to name exactly which consumer re-checks and which don't — the drivers are not what makes this safe.

lib/services/medias/mediaFileUrl.ts
```diff
@@ -48,6 +92,15 @@ const getOwnPathname = (url: string, config: HostRuleConfig): string | null => {
  * local storage path, which then misses in storage — and, at a caller that
  * treats "not local" as "fetch it over HTTP instead", skips the branch that
  * would have retrieved the file correctly.
+ *
+ * A path that walks out of storage — read on the DECODED value; see the note
+ * at the decode below for the one shape that stays encoded — is answered for as
+ * not ours either, so no caller has to re-derive that check, and at one of them
+ * nothing else does:
+ * `scripts/backup/actorArchive.ts` reads an actor's `iconUrl`, which
+ * `POST /api/v1/accounts/profile` stores as a free string, and `path.join`s
```
llun: **Traversal impact is understated here.** `POST /api/v1/accounts/profile` declares `iconUrl: z.string().optional()` behind `AuthenticatedGuard` and spreads it into `updateActor`; `scripts/backup/actorArchive.ts` reads that value through this helper, and `copyProfileImage` copies whatever the path resolves to into the tarball the account owner receives. Before this refusal, an authenticated account could read an arbitrary server file into its own export — worth documenting explicitly here (and in `AGENTS.md`/`CLAUDE.md`/`REVIEW.md`/the PR body).

lib/services/medias/mediaFileUrl.ts
```diff
@@ -59,9 +112,18 @@ export const getMediaPathFromFileUrl = (
...
+  // An undecodable escape anywhere in the path leaves the WHOLE value encoded,
+  // so what comes back is then not canonical — `%2e%2e/x%zz` reads as segments
+  // that are not `..`. That is safe only because no consumer decodes again:
+  // `path.join`/`path.resolve` treat `%2e` literally, an S3 key is literal,
+  // and `scripts/backup/productionArchive.ts` re-checks for `..` on the raw
+  // string. A consumer that DOES decode needs its own check.
+  let storagePath: string
   try {
-    return decodeURIComponent(encodedPath)
+    storagePath = decodeURIComponent(encodedPath)
```
llun: **Decode failure returns the still-encoded string, which isn't canonical.** When `decodeURIComponent` throws, `..%2f..%2fetc%2fpasswd%zz` falls through as one harmless-looking segment instead of being recognized as a traversal attempt — it's inert only because neither current consumer (`path.join`/S3 key) decodes again. Please comment on why that's safe today and what a future consumer must do, and pin it with a test so it's visible rather than assumed.

lib/utils/host.ts
```diff
@@ -165,6 +175,15 @@ export const getTrustedHostRules = (config: HostRuleConfig): string[] => [
...
+/**
+ * The configured host as a request should see it. NOT an authority for
+ * comparing against — that is `getCanonicalAuthority`. This one falls back to
+ * the RAW value whenever `normalizeHost` refuses it, which includes every
+ * loopback name and every malformed one, so it can hand back something
+ * unlowercased, untrimmed, or carrying a path. `selectHeaderHost` wants that
+ * (it is echoing a configured value back, not deciding trust); a security
+ * comparison must not use it.
+ */
 export const getConfiguredHost = (host: string | undefined | null) =>
```
llun, line 203: **No doc comment, and it sits right beside the new `getCanonicalAuthority`.** `getConfiguredHost` falls back to the raw, unnormalized value whenever `normalizeHost` refuses one. A future caller could reach for this by name for an authority comparison and silently defeat the trust check this PR just tightened. Please add a doc comment making clear it's not for deciding trust.

lib/services/medias/mediaFileUrl.test.ts
```diff
@@ -103,11 +103,31 @@ describe('getMediaPathFromFileUrl', () => {
       url: 'https://sub.llun.test/api/v1/files/ab/cd.webp',
       expected: null
     },
+    {
+      description: 'a media URL on our own host carrying a query string',
+      url: 'https://llun.test/api/v1/files/ab/cd.webp?v=2',
```
llun, line 113: **Coverage lost in the move from the deleted script suite.** The deleted suite had an absolute URL carrying a query string; only the host-relative equivalent (`/api/v1/files/ab/cd.webp?v=2`) survived here. Also currently unpinned: the `:80` behavior delta this consolidation newly exposes (`isOwnInstanceHost` strips both default ports where the deleted copy's non-loopback branch treated only `:443` as implied), the decode-throw fallback, and a double-encoded path. Please add cases for all four.

llun (commented):
Round 1 sub-agent review (five lenses, verified findings) against commit 252ff5132. Posting each finding inline.

lib/utils/host.ts
llun: Fixed in ccf490db3: `getCanonicalAuthority` now blanks on `authority.includes('*')` (host.ts:277), and `normalizeHost` refuses a `*` anywhere in an authority (host.ts:105) so the rules pass is closed too. Each guard has its own failing test, proved by isolated revert.

lib/utils/host.ts
llun: Fixed in ccf490db3: reworded here (lines 260-265) and in AGENTS.md/CLAUDE.md/REVIEW.md/the PR body as a latent trap for a caller that USES the host, not an exploit.

lib/services/medias/mediaFileUrl.ts
llun: Fixed in ccf490db3: comment narrowed (lines 73-79) to name exactly which consumer re-checks containment (`LocalFileStorage.getFile`) and which don't (`deleteFile`, an S3 key, and `copyProfileImage` in `scripts/backup/actorArchive.ts`).

lib/services/medias/mediaFileUrl.ts
llun: Fixed in ccf490db3: documented in this doc comment (lines 96-103), and in AGENTS.md/CLAUDE.md/REVIEW.md and the PR body.

lib/services/medias/mediaFileUrl.ts
llun: Fixed in ccf490db3: commented (lines 115-120) with the reason it's safe today and what a future consumer must do, and pinned by tests in mediaFileUrl.test.ts (the double-encoded and undecodable-escape cases) so it's visible rather than assumed.

lib/utils/host.ts
llun: Fixed in ccf490db3: added a doc comment (lines 178-186) stating it's not for deciding trust and documenting its raw-value fallback.

lib/services/medias/mediaFileUrl.test.ts
llun: Fixed in ccf490db3: all four cases added to mediaFileUrl.test.ts — the absolute-URL-with-query-string case (line 108), the `:80`-spelled-out case, the decode-throw fallback case, and the double-encoded-path case.

llun:
### Adjacent issues surfaced by the review, deliberately not fixed here

Recording these so they are on the record. None was introduced by this PR.

1. **The production-archive download path does not use the guarded fetch helper the rest of the codebase does.** Specifics and a proposed fix have been handed over separately rather than written up here — it is live and unfixed, and this is a public repository. The shape to copy is `downloadRemoteImage` in `scripts/maintenance/backfillMediaBlurhash.ts`, which pairs `safeImageFetch` with `readResponseArrayBufferWithLimit`.

2. **`LocalFileStorage.deleteFile` has no containment check while `getFile` does** (`lib/services/medias/localFile.ts`). Latent rather than live: every current caller passes a server-generated `medias.*.path`. It is the reason the comment above `isTraversingStoragePath` in this PR names which consumer re-checks rather than claiming "the storage drivers" do.

3. **`deleteSavedMedia` in `app/api/v1/admin/custom_emojis/route.ts` is a third "URL → storage path" spelling** — `indexOf` rather than a prefix check, no host check, `decodeURIComponent` with no traversal check — feeding a delete. Not exploitable today: its only argument is a URL the server just minted, which is the `getAttachmentMediaPath` case. It is the last remaining copy of the question this PR consolidates.

lib/utils/host.ts
```diff
@@ -93,6 +93,16 @@ export const normalizeHost = (
...
+  // `*` is legal in neither a DNS name nor an IP literal, and `new URL` parses
+  // one into a host regardless — so an authority still carrying one after a
+  // documented leading `*.` has been stripped is a malformed RULE, never a
+  // host. Without this an operator who wrote `*llun.test` (or a bare `*`) in
+  // `ACTIVITIES_TRUSTED_HOSTS` had it match a URL spelled exactly the same
+  // way, and `**.edge.llun.test` matched a `*.edge.llun.test` rule through
+  // `hostMatchesRule`'s suffix test. `getCanonicalAuthority` blanks the
+  // documented `*.` form for the exact-authority pass; this closes the rest of
+  // the class in the one place the rules pass goes through.
+  if (hostToParse.includes('*')) return null
```
llun: **Blast radius: this `*` guard also governs inbound header selection, not just the media helper.**

`normalizeHost` backs both `normalizeHostRules` (the trusted-host list) and `selectHeaderHost` (inbound `Host`/`X-Forwarded-Host`), so this `*` refusal decides every inbound header selection too — and nothing exercised it through that entry point. A future tightening of this check could regress inbound header selection for a wildcard-trusted deployment without any test failing where it matters.

lib/services/medias/mediaFileUrl.test.ts
```diff
@@ -127,6 +147,172 @@ describe('getMediaPathFromFileUrl', () => {
...
+    {
+      description:
+        'a host-relative URL whose literal dot segments resolve inside',
+      url: '/api/v1/files/ab/../cd.webp'
+    }
+  ])('normalises $description', ({ url }) => {
+    expect(getMediaPathFromFileUrl(url, config)).toBe('cd.webp')
+  })
+
+  // NOT A TRAVERSAL GUARANTEE — these two pin what this helper RETURNS for a
```
llun, line 306: **Both `it.each` cases here assert a traversal-shaped string IS returned — easy to misread as "traversal is handled here".**

A reader skimming only the green assertions could take this as proof the helper resolves/blocks traversal for these inputs, when it's actually pinning what falls out of an unresolvable decode. The explanatory prose is there, but only in a comment a skim would miss.

lib/utils/host.test.ts
```diff
@@ -174,6 +175,53 @@ describe('isHostTrustedByRules', () => {
   })
 })
 
+// `*` is legal in neither a DNS name nor an IP literal, but `new URL` parses
+// one into a host — so an operator's malformed `ACTIVITIES_TRUSTED_HOSTS`
+// entry could be matched by a URL spelled exactly the same way.
+describe('normalizeHost wildcard handling', () => {
+  afterEach(() => {
```
llun, line 183: **Missing `afterEach` resetting the module-level host caches, unlike both neighbouring `describe` blocks.**

Not a correctness bug — the caches key on their full input — but inconsistent with the surrounding tests in this file.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -63,233 +62,40 @@ describe('backfillMediaBlurhash parseArgs', () => {
...
-        'https://remote.example/redirect?to=/api/v1/files/medias/a.jpg',
-        HOSTS.ownHostRules
-      )
-    ).toBeNull()
+  // `getCanonicalAuthority` blanks a wildcard-shaped value, and this pins that
+  // the fallback host inherits that rather than minting
+  // `https://*.llun.test/api/v1/files/...` — which is what leaves `main`'s
+  // `if (!instanceHosts.fallbackHost) throw` reachable. It does NOT cover that
+  // throw: `main` is not exercised by this suite at all.
+  it('answers an empty fallback host for a wildcard-shaped configured host', () => {
```
llun: **Nothing pins a `*.`-shaped `ACTIVITIES_HOST` producing an empty `fallbackHost`.**

That's the case that leaves the script's startup throw (`if (!instanceHosts.fallbackHost) throw`) reachable, rather than a nonsense wildcard authority being minted into a media URL.

lib/utils/host.test.ts
```diff
@@ -201,18 +249,21 @@ describe('isOwnInstanceHost', () => {
       host: 'cdn.edge.llun.test',
       expected: true
     },
+    // These three lead with the port: the reporter shows only the first 37
+    // characters of `$description`, and spelled the other way round the first
+    // two rendered as the same line.
     {
-      description: 'the configured host with the default HTTPS port',
+      description: 'port 443 on the configured host',
```
llun, line 228: **These two default-port test names print identically in the reporter.**

Vitest's reporter shows only the first 37 characters of an `it.each` `$description` (chai's `truncateThreshold` of 40, less `objDisplay`'s reserves and the ellipsis), so `'the configured host with the default HTTPS port'` and `'…default HTTP port'` both printed as the same truncated line — a regression in either case could not be told apart in the output. Predates this PR (came in with #1560) but sits in a file this PR edits.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: **This case's comment overstates what the assertion covers.**

It says this makes `main`'s `if (!instanceHosts.fallbackHost) throw` reachable, which reads as though the test covers that throw. It doesn't — `main` isn't exercised by this suite at all — and the assertion duplicates one already in `lib/utils/host.test.ts`.

llun:
**PR body described a live, unfixed SSRF-class issue in exploitation-ready detail on a public repository.**

It named the vulnerable function, the attacker-controlled field, and the CLI flag that triggers it — while this PR ships the fix for a different issue. That detail shouldn't be public before a fix lands.

lib/utils/host.ts
llun: Addressed in 6eae5f30a: `lib/utils/host.test.ts` gained `still selects a wildcard-trusted subdomain from a forwarded host`, which exercises this guard through `selectHeaderHost` directly (a real `*.edge.llun.test`-trusted subdomain arriving via `X-Forwarded-Host` must still be selected). It fails if this `*` check is ever moved ahead of the wildcard strip — proved by that mutation: 8 tests fail, this one among them. The PR body also gained a "blast radius" paragraph stating the same.

lib/services/medias/mediaFileUrl.test.ts
llun: Addressed in 6eae5f30a: added an explicit `NOT A TRAVERSAL GUARANTEE` marker on this comment block (and the matching marker at the decode site in `lib/services/medias/mediaFileUrl.ts`, line 115), so the returned traversal-shaped strings can't be misread as "handled" rather than "safe only because no consumer decodes again".

lib/utils/host.test.ts
llun: Addressed in 6eae5f30a: added `afterEach(() => { resetHostCachesForTests() })` to this describe, matching the neighbouring `describe` blocks in this file.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Addressed in 6eae5f30a: added `answers an empty fallback host for a wildcard-shaped configured host`, asserting `buildInstanceHosts({ host: '*.llun.test' }).fallbackHost` is `''`.

lib/utils/host.test.ts
llun: Addressed in 1146a91db: all three port cases (this line, and the `port 80`/`port 8443` cases just below) renamed to lead with the port — `port 443 on the configured host`, `port 80 on the configured host`, `port 8443 on the configured host` — the way `mediaFileUrl.test.ts`'s default-port cases already did.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Addressed in 1146a91db: comment reworded to state exactly what this assertion proves (`getCanonicalAuthority` blanks a wildcard-shaped value and the fallback host inherits that) and to note explicitly that `main`'s throw is uncovered by this suite.

llun:
Addressed in 6eae5f30a: the PR body and the summary issue comment were trimmed to a neutral pointer at that unfixed issue, and the exploitation detail was handed over privately instead of staying in this public thread.

lib/services/medias/mediaFileUrl.test.ts
```diff
@@ -127,6 +147,176 @@ describe('getMediaPathFromFileUrl', () => {
...
+
+  // An operator's malformed `ACTIVITIES_TRUSTED_HOSTS` entry must not be
+  // matched by a URL spelled the same way. Which guard catches which, proved
+  // by isolated revert: the first two are `getCanonicalAuthority`'s
+  // `includes('*')` alone — they never reach `normalizeHost`, because
+  // `isOwnInstanceHost` returns on the blanked authority first — and the third
+  // is caught by BOTH, since `**.` also fails `normalizeHost` on the rules
+  // pass. `getCanonicalAuthority`'s own table in `lib/utils/host.test.ts`
+  // covers the same extension directly.
+  it.each([
```
llun, line 345: This comment attributes the refusal of these three wildcard spellings to "`getCanonicalAuthority` only blanks the documented leading `*.`; the rest of the class is refused by `normalizeHost`" — which describes the code as it stood *before* this same commit broadened `getCanonicalAuthority` to refuse `*` anywhere, not just a leading `*.`.

Isolated reverts show the opposite: the first two cases (`a wildcard rule written without its dot`, `a bare wildcard rule`) are caught by `getCanonicalAuthority` alone — `isOwnInstanceHost` returns on the blanked authority before the rules pass is ever reached — and only the third (`a doubled wildcard under a real wildcard rule`) is caught by both guards.

Separately, `getCanonicalAuthority`'s own case table in `lib/utils/host.test.ts` covers nothing but leading-`*.` values, so the broadening this commit makes is proven only indirectly, through this mislabelled block.

lib/utils/host.ts
```diff
@@ -93,6 +93,13 @@ export const normalizeHost = (
   const hasWildcard = firstHost.startsWith('*.')
   if (hasWildcard && !allowWildcard) return null
   const hostToParse = hasWildcard ? firstHost.slice(2) : firstHost
+  // An authority still carrying a `*` once a documented leading `*.` has been
+  // stripped is a malformed RULE, never a host — see `getCanonicalAuthority`
+  // below for why one gets this far. This is the RULES pass's half of that
+  // guard, and it catches one spelling the other cannot: `**.edge.llun.test`
+  // matched a `*.edge.llun.test` rule through `hostMatchesRule`, which only
+  // ever compares the tail.
+  if (hostToParse.includes('*')) return null
```
llun: This comment and `getCanonicalAuthority`'s comment (further down in this file) had each grown their own copy of the same `*llun.test` / bare-`*` example — the one place across the diff where two comments restate the same fact with the same illustration instead of dividing the explanation between the two guards.

lib/services/medias/mediaFileUrl.test.ts
llun: Verified and fixed in 5fd138d42 (`test: attribute the wildcard guards to the ones that actually catch`): the comment above now carries the isolated-revert attribution (first two caught by `getCanonicalAuthority` alone, the doubled-wildcard case by both), and three direct cases — a dotless wildcard rule, a bare wildcard rule, and a mid-label wildcard — were added to `getCanonicalAuthority`'s own table in `lib/utils/host.test.ts`, each of which fails if the check reverts to a leading-`*.`-only prefix test.

lib/utils/host.ts
llun: Verified and fixed in 6ccb0876e (`none: keep the wildcard example in one comment, not two`): the `*llun.test` / bare-`*` example now lives only with `getCanonicalAuthority`, the guard those two spellings actually meet. This comment, above `hostToParse.includes('*')`, keeps only what's unique to the rules pass — that `**.edge.llun.test` matched a `*.edge.llun.test` rule through `hostMatchesRule`, which compares the tail alone.

llun:
## Sub-agent PR review loop — closed out

Six rounds ran: five review rounds plus this close-out round. Lenses used across the loop: correctness, security, testing, quality, and scope-and-regression. Round 5's testing and quality lenses both signed off with no new findings.

15 review threads were opened across the loop (13 in rounds 1-3, 2 in rounds 4-5) and all 15 are now resolved.

The three adjacent issues named in the earlier summary comment remain deliberately out of scope for this PR, with follow-ups filed:
1. The production-archive download path does not use the guarded fetch helper (`safeImageFetch` + `readResponseArrayBufferWithLimit`, the shape used by `downloadRemoteImage` in `scripts/maintenance/backfillMediaBlurhash.ts`). Live and unfixed; specifics handed over separately rather than written up here.
2. `LocalFileStorage.deleteFile` has no containment check while `getFile` does (`lib/services/medias/localFile.ts`). Latent, not live — every current caller passes a server-generated path.
3. `deleteSavedMedia` in `app/api/v1/admin/custom_emojis/route.ts` is a third "URL → storage path" spelling (`indexOf`, no host check, no traversal check) feeding a delete. Not exploitable today — its only argument is a server-minted URL, the `getAttachmentMediaPath` case — but it is the last remaining copy of the question this PR consolidates.

No further action expected on this PR from the review loop.

llun:
### Review loop closed

Seven rounds. Lenses: correctness, security, testing, quality, scope-and-regression — fanned out in parallel each round and pointed at the previous round's fixes rather than the original diff.

Rounds 6 and 7 were both clean, which is the stopping condition. 15 threads were opened and all 15 are resolved.

What the loop actually changed, beyond the original consolidation:

- The wildcard guard I first wrote tested a leading `*.` only. Isolated reverts showed `*llun.test`, a bare `*`, and `**.edge.llun.test` still matched, so the fix became two guards — `getCanonicalAuthority` for the exact pass, `normalizeHost` for the rules pass — each with its own failing test.
- Two of my own claims were wrong. The wildcard hole's impact was inflated: it granted no path the configured host did not already grant. The traversal fix's impact was understated: it stops a reachable authenticated arbitrary-file read, not just defence in depth. Both are corrected in the code comments and in `AGENTS.md`/`CLAUDE.md`/`REVIEW.md`.
- The PR body originally spelled out the trigger for a live, unfixed issue on a public repo. That was trimmed to a neutral pointer.
- Several tests were re-proved by isolated rather than combined revert, which promoted one apparently-vacuous table to real and demoted one real-looking case to vacuous (removed).

The three adjacent issues in the earlier summary comment stay deliberately out of scope, with follow-ups filed.

PR #1575, opened 2026-08-26
fix: confine the local storage drivers' delete path to the storage root

llun:
## Summary

`LocalFileStorage.getFile` (`lib/services/medias/localFile.ts`) resolved the requested path against the media root and refused it when it escaped. `deleteFile` directly below it resolved and `fs.unlink`ed with no such check.

Every caller reaches `deleteFile` with a server-generated `medias.*.path`, so this was latent rather than live. What makes it worth fixing is the shape: containment was an invariant of the **callers**, not of the driver, so it held only until someone added a caller — and an unlink is the more damaging half. It also made "the storage drivers refuse such a path too" false as a claim about the pair, which is why a review of #1570 had to narrow a comment in `mediaFileUrl.ts` to `getFile` alone.

## What changed

- **New `lib/services/medias/storagePath.ts`** — `resolveStorageFilePath` (null when the path escapes, and logs the refusal) and `assertStorageFilePath` (throws, for a write whose only alternative is creating a file the driver does not own). It lives under `medias/` and is shared, the same way `medias/fileName` and `medias/quota` already are.
- **`medias/localFile.ts`** — `getFile` and `deleteFile` resolve through it; both write paths assert. That also unified the two different spellings of "the storage root" the file carried, so `import process from 'process'` is gone.
- **`lib/services/fitness-files/path.ts` deleted.** It held a near-identical copy that the fitness driver already applied correctly on all three of `getFile`/`deleteFile`/`saveFile` — so the answer to "does the fitness driver need this fix?" was no, it needed to stop being a second implementation. That driver and `lib/jobs/importStravaArchiveJob.ts` now import the shared module.
- **`scripts/maintenance/cleanupMediaStorage.ts`** — a review found a **third** copy of this check guarding an `fs.unlink`, and its boundary was wrong: `startsWith` against a bare `path.resolve(basePath)` accepts a sibling directory the root's name prefixes, so `/srv/uploads` accepted `../uploads-backup/x`. Now uses the shared helper, keeping its own throw so the operator still learns which path was refused.
- **`lib/services/medias/storagePathCallSites.test.ts`** — a deliberately narrow guard against a *new* unguarded call site. See below.
- Docs: **A stored path is confined to the storage root** in `AGENTS.md`, a `REVIEW.md` bullet, a `CLAUDE.md` reminder.

`origin/main` is merged in — #1569 landed mid-review, six commits after this branch forked.

## Not in scope

The object-storage drivers are untouched: an S3 key has no filesystem root to escape, and neither `S3StorageFile.ts` imports `fs` or `path`.

## Tests

- `storagePath.test.ts` covers the helper directly, including the refusal log.
- A `deleteFile` block in `localFile.test.ts`: removes a file inside the root, refuses a relative escape and an absolute path (asserting the outside file survives), reports a missing file inside the root as already deleted, and confines under a **relative** storage root — which is what `./uploads` ships.

**Proved by reverting.** With `deleteFile` restored to the bare `path.resolve` + `unlink`, both refusal tests fail — and `resolves.toBe(true)` there means the unlink genuinely removed the file outside the root. Stubbing the containment branch additionally fails the pre-existing `LocalFileFitnessStorage path containment` test, confirming the fitness driver really is on the shared module.

Full suite: 886 files, 11,660 tests.

## The guard test is deliberately narrow, and that is a finding

It catches the obvious spelling of a missing call site and **makes no larger claim**. It does not catch a destructured `const { resolve } = path`, `path?.resolve`, a renamed import, or a helper resolving on a driver's behalf — a reviewer is the enforcement for those until an AST rule lands in `lint/agentsRules.mjs`, which is filed as a follow-up.

Five review rounds each broke a cleverer version of that file, in both directions, and **every failure came from a mechanism added to widen coverage**: alias tracking fell to prettier wrapping a call across lines (step 1 of the commit gate); a substring ban fell to `path['resolve']`; an extracted-variable heuristic failed on an unrelated `const resolvedArchivePath = 'database.'`; an import-shape regex matched *nothing* on a wrapped import and passed vacuously; and a comment-stripping pass read the `/*` inside the MIME string `'*/*'` as a comment opener, blanking 2,126 characters of `lib/services/link-previews/` so the scan silently stopped covering it. The current version reads raw text with no interpolation and no stripping, so it cannot pass vacuously, and a comment naming a banned pattern fails loudly rather than quietly disabling the scan. The history lives in the test's own header, where the next person to make it cleverer will read it.

## Disclosed, not changed

- `assertStorageFilePath` throws `'Storage path escapes storage root'` where `assertFitnessStoragePath` threw `'Fitness storage path escapes storage root'`. If it ever fires inside `resolveArchivePath`, `toImportErrorMessage` persists that string verbatim to `fitness_files.importError`, which the owner sees. Unreachable today; kept generic because the row it lands on is already a `fitness_files` row.
- The check is **lexical** — a symlink planted under a storage root defeats it. Inherited from both predecessors; nothing creates symlinks in either root (the Strava archive reader decompresses into memory, so there is no zip-slip route to plant one).
- `scripts/backup/productionArchive.ts` keeps its own stricter `assertRelativeFilePath`/`normalizeStoragePath` for archive-entry names. Different question, different input — documented as deliberate, not drift.
- `copyProfileImage` in `scripts/backup/actorArchive.ts` still hand-rolls an equivalent check in the `path.relative` form. Not broken, and filed as a follow-up.

## Review

Seven rounds of the sub-agent loop, each pointed at the **previous round's fixes** rather than the original diff. Round 1's nine findings are on their own resolved threads; rounds 2–5 and 6–7 are recorded in comments below. An adversarial verifier over the complete diff could not refute the core claim.

The production code has not changed since round 1 — `storagePath.ts`, both drivers and the cleanup script were settled early and every round since confirmed them.

## Checklist

- [x] Docs updated: grepped `*.md` and `docs/` for every name this PR reshapes
- [x] If migrations changed: n/a, no migration


lib/services/medias/storagePath.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * The absolute path `filePath` names inside `storageRootPath`, or null when it
+ * escapes that root.
+ *
+ * Every filesystem path a local storage driver builds goes through here — the
+ * media driver and the fitness-file driver both. A stored path is data: it
+ * reaches a driver from a `medias` / `fitness_files` row, from a URL path
+ * segment, or from an archive entry, and `path.resolve` walks straight out of
+ * the root given `../` or an absolute path. The escape is silent, because the
+ * read or the unlink simply lands somewhere else on disk.
```
llun, line 14: The containment claim in this doc comment is defeated by a symlink.

The check is purely lexical — `path.resolve` plus a prefix compare, never `fs.realpath`. A symlink planted inside the storage root pointing outside it resolves to a path that passes the prefix compare, so `getFile` would read through it (an `unlink` would only remove the link itself). This is inherited unchanged from both predecessors — the old inline `getFile` check and `fitness-files/path.ts` — and nothing in the codebase creates symlinks in either root (the Strava archive reader reads entries into memory rather than extracting to disk, so there is no zip-slip route to plant one). So it is a residual, not a regression. But this doc comment and the new AGENTS.md section both state confinement without qualifying it, and a future reader will take that at face value. Add a sentence naming the limitation rather than reaching for `realpath`.

AGENTS.md
```diff
@@ -2282,6 +2282,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The rule covers delete and write, not just read.** `LocalFileStorage.getFile` carried the check inline while `deleteFile` beside it resolved and `unlink`ed with none at all. Every caller passed a server-generated `medias.*.path`, so it was latent rather than live — but the asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun: "every caller passed a server-generated `medias.*.path`" is stronger than the code supports.

True for every caller that READS a path from a row, but `deleteSavedMedia` in `app/api/v1/admin/custom_emojis/route.ts:186-193` DERIVES one: it takes a URL, `indexOf`s the `/api/v1/files/` marker, slices the tail, and `decodeURIComponent`s it before calling `deleteMediaFile`. Its input is `saved.url` from `saveMedia`, so it is server-built and carries no percent-encoding today — the latent-vs-live claim still holds — but it is the one caller whose argument is parsed rather than read, and that `decodeURIComponent` turns `%2e%2e%2f` into traversal the moment the input shape changes. Worth one clause of accuracy in the prose, since this PR's whole thesis is that containment must not rest on a claim about callers.

lib/services/medias/storagePath.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * The root itself resolves to itself rather than to null; only a path that
+ * leaves the root is refused.
+ */
+export const resolveStorageFilePath = (
```
llun, line 27: The new invariant is documentary, not enforceable — no call-site guard.

Nothing stops a new driver method from writing `path.resolve(this._config.path, x)` directly and bypassing the helper. This repo guards comparable prose invariants with dedicated grep-style tests — `lib/database/statusVisibilityCallSites.test.ts`, `lib/clientModuleBoundary.test.ts`, `lib/components/tailwindCssVariableSyntax.test.ts`. The tests added here prove the helper is correct and that today's call sites use it; none can catch a call site added tomorrow. A guard test asserting the two local drivers contain no bare `path.resolve`/`path.join` against their own `_config.path` would make the rule enforceable at CI rather than at review time.

lib/services/medias/storagePath.ts
```diff
@@ -0,0 +1,54 @@
...
+ * replaced was exactly that claim, and it held only until someone added a
+ * caller.
+ */
+export const assertStorageFilePath = (
+  storageRootPath: string,
+  filePath: string
+): string => {
+  const fullPath = resolveStorageFilePath(storageRootPath, filePath)
+  if (!fullPath) {
+    throw new Error('Storage path escapes storage root')
```
llun, line 70: The thrown message changed, and it is persisted to an operator-visible column.

`assertFitnessStoragePath` threw `'Fitness storage path escapes storage root'`; the shared `assertStorageFilePath` throws `'Storage path escapes storage root'`. If it ever fires inside `resolveArchivePath` (`lib/jobs/importStravaArchiveJob.ts:202`), `toImportErrorMessage` writes that exact string to `fitness_files.importError`, which the owner sees. Nothing pattern-matches the old string and the branch is unreachable today (the archive path is a trusted `fitness_files.path` row), so nothing breaks — but it is an unannounced change to persisted, operator-facing text riding inside what the PR body frames as a mechanical rename. It should at least be disclosed in the PR description.

lib/services/medias/storagePath.ts
```diff
@@ -0,0 +1,54 @@
...
+  storageRootPath: string,
+  filePath: string
+): string | null => {
+  const storageRoot = path.resolve(storageRootPath)
+  const fullPath = path.resolve(storageRoot, filePath)
+  const storageRootPrefix = storageRoot.endsWith(path.sep)
+    ? storageRoot
+    : `${storageRoot}${path.sep}`
+
+  if (fullPath !== storageRoot && !fullPath.startsWith(storageRootPrefix)) {
```
llun, line 37: A refused path is silent — no log on either driver.

Both drivers return `null`/`false` on refusal with nothing logged, matching `getFile`'s pre-existing behaviour. Defensible, since every current caller is server-generated or pre-sanitized upstream and the branch should never fire. But the stated point of pushing this into the driver is to catch a FUTURE caller's mistake or a corrupted row — and if that ever happens the operator gets no log line, no metric, just a quietly failed delete or a 404-shaped read. One `logger.warn` at the refusal point would cost nothing today and would be the only signal that the invariant was ever exercised.

lib/services/medias/localFile.test.ts
```diff
@@ -96,6 +96,56 @@ describe('LocalFileStorage.getFile', () => {
...
+
+  it('removes a file inside the media root', async () => {
+    const storedPath = path.join(mediaRoot, 'avatar.png')
+    await fs.writeFile(storedPath, 'image-data')
+
+    await expect(createStorage().deleteFile('avatar.png')).resolves.toBe(true)
+    await expect(fs.access(storedPath)).rejects.toThrow()
+  })
+
+  it('refuses a relative path that escapes the media root', async () => {
```
llun, line 134: Driver-level tests never exercise a relative storage root, which is the shipped default.

`ACTIVITIES_MEDIA_STORAGE_PATH` defaults to `./uploads` (`lib/config/environmentTemplates.ts:69`), so a relative root is the out-of-the-box production config. Every driver-level test here and in `fitness-files/localFile.test.ts` uses an absolute `mkdtemp` root; only `storagePath.test.ts:52` covers the relative case, and only at the pure-function level. Low risk because the unit test does cover the arithmetic, but it is the one config shape actually shipped that the class-level regression tests skip.

lib/services/medias/localFile.test.ts
```diff
@@ -96,6 +96,56 @@ describe('LocalFileStorage.getFile', () => {
...
+    new LocalFileStorage(
+      {
+        type: MediaStorageType.LocalFile,
+        path: mediaRoot
+      },
+      'llun.test',
+      {} as Database
+    )
+
+  it('removes a file inside the media root', async () => {
```
llun, line 126: `deleteFile`'s ENOENT-returns-true branch is still untested.

`deleteFile` deliberately reports a missing file as success (`error.code === 'ENOENT'` → `true`). Pre-existing and untouched by this diff, but the new `deleteFile` describe block is its natural home, and the branch now sits directly beside the refusal path that returns `false` — the two are easy to conflate, and nothing currently pins which is which.

lib/services/medias/storagePath.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    },
+    {
+      description: 'a nested file name',
+      filePath: '2026-08-26/ride.fit',
+      expected: `${STORAGE_ROOT}/2026-08-26/ride.fit`
+    },
+    {
+      description: 'an inner traversal that stays inside the root',
+      filePath: 'nested/../avatar.webp',
+      expected: `${STORAGE_ROOT}/avatar.webp`
```
llun, line 37: This `it.each` description is truncated in the reporter.

Vitest cuts an interpolated `$description` at 40 characters, so `'an inner traversal that stays inside the root'` renders as `resolves 'an inner traversal that stays inside …'`. Cosmetic only — no collision with another case's truncated prefix — but shortening it gives a clean report.

llun (commented):
Round-1 automated review: 8 findings anchored inline below. A 9th finding (scripts/maintenance/cleanupMediaStorage.ts:267) targets a file this PR does not touch, so the GitHub review API refuses to anchor a line comment there ("Path could not be resolved"); it is posted as a separate top-level PR comment instead.

llun:
Round-1 finding — `scripts/maintenance/cleanupMediaStorage.ts:267` (not inline: this file is untouched by the PR diff, so the GitHub review-comments API refuses to anchor a line comment there — "Path could not be resolved" — hence this is a top-level comment instead)

**A third copy of this containment check survives the consolidation, and its boundary is wrong**

`deleteLocalFile` does `path.resolve(basePath, filePath)` then `if (!fullPath.startsWith(path.resolve(basePath)))` before `fs.unlink`. That is the same predicate this PR extracts, minus the separator boundary: `resolveStorageFilePath` appends `path.sep` to the root precisely because a bare `startsWith` accepts a sibling directory the root's name prefixes — root `/srv/uploads` accepts `../uploads-backup/x`, which resolves to `/srv/uploads-backup/x` and gets unlinked. `storagePath.test.ts` has a dedicated case for exactly this ("a sibling directory of the root"). Not exploitable today: `filePath` comes from `listLocalFiles(basePath)`, a readdir walk of the root, so it is relative by construction. But the PR deletes `fitness-files/path.ts` on the grounds that the copies must not drift, and the new AGENTS.md rule reads as covering any `unlink` under a storage root — so either this call site joins the consolidation or the prose must be narrowed to the drivers. Suggest `resolveStorageFilePath(basePath, filePath)` here, keeping the script's own `Invalid file path: ${filePath}` throw so the operator still learns which path was refused.

lib/services/medias/storagePath.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding2.md

AGENTS.md
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding3.md

lib/services/medias/storagePath.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding4.md

lib/services/medias/storagePath.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding5.md

lib/services/medias/storagePath.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding6.md

lib/services/medias/localFile.test.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding7.md

lib/services/medias/localFile.test.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding8.md

lib/services/medias/storagePath.test.ts
llun: @/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding9.md

llun:
@/private/tmp/claude-501/-Users-llun-Documents-llun-activities-next--claude-worktrees-zen-haibt-47256b/003143ac-6d78-4908-9393-0a7ba5ca3617/scratchpad/replies/finding1_full.md

lib/services/medias/storagePath.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents. Apologies for the noise.)

**Fixed.** The doc comment now states the limitation instead of claiming confinement flatly: the check is lexical, never calls `fs.realpath`, so a symlink planted inside the root and pointing outside resolves to a path that passes — a read follows it, an unlink removes only the link.

It stays a documented residual rather than a `realpath` call for two reasons: both predecessors were lexical too, so this is not a regression; and nothing creates symlinks in either storage root — `lib/services/strava/archiveReader.ts` reads zip entries into memory via `yauzl` rather than extracting to disk, so there is no zip-slip route to plant one. The comment says explicitly that anything gaining the ability to create a link under a storage root changes that answer.

Same wording added to the AGENTS.md section and the REVIEW.md bullet.

AGENTS.md
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed.** The AGENTS.md bullet and the CLAUDE.md reminder now distinguish callers that **read** a server-generated path from a row from the one that **derives** one, naming `deleteSavedMedia` and its `decodeURIComponent` of the `/api/v1/files/` tail, and stating the condition under which it would matter.

Agreed this was worth the clause. The PR's whole thesis is that containment must not rest on a claim about callers, so prose that stated exactly such a claim — and stated it more strongly than the code supports — was undermining its own argument.

lib/services/medias/storagePath.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed, then fixed again** — this finding was the most productive one in the round.

Round 1 added `lib/services/medias/storagePathCallSites.test.ts`, following `statusVisibilityCallSites.test.ts` / `clientModuleBoundary.test.ts`, tracking `this._config.path` and its aliases.

Round 2 reviewers then broke that version four ways: a destructured rename (`const { path: root } = this._config`), a nested call in the argument list (the `[^)]*` cannot cross a `)`), an extracted `startsWith` variable, and — decisively — **prettier wrapping the call across lines**, since the check scanned line by line and `prettier --write` is step 1 of the commit gate. A guard a formatting pass disarms is worse than no guard, because it reads as covered.

So the rule changed rather than acquiring four patches. Neither driver needs `path.resolve` or `path.join` at all — their only `path` calls are `extname`/`dirname`, which read a path rather than build one — so both constructors are now **banned outright** in those two files, which no amount of aliasing, nesting or wrapping evades. The containment-idiom check scans each file as one string and also catches the extracted-variable spelling. All four bypasses have a revert proof.

lib/services/medias/storagePath.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Addressed, no code change.** Confirmed the trace: `toImportErrorMessage` would write that exact string to `fitness_files.importError` for the owner to read.

Keeping the generic message. The row it lands on is a `fitness_files` row, so "Fitness" was redundant context there; nothing pattern-matches either spelling; and the throw carries a stack that pinpoints which of the four call sites raised it.

What the finding was right about is that it needed disclosing rather than riding along inside a rename, so it is now called out in the PR description. The AGENTS.md bullet also documents the pattern for a caller that *does* want its own context — take `resolveStorageFilePath` and throw your own — which is what `scripts/maintenance/cleanupMediaStorage.ts` now does to keep its `Invalid file path: ${filePath}`.

lib/services/medias/storagePath.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed.** `resolveStorageFilePath` now emits `logger.warn({ message: 'Refused a storage path outside the storage root', storageRootPath, filePath })` before returning null.

The argument in the finding is what settled it: the branch being unreachable from today's callers is precisely why it needs a line. It exists for a caller nobody has written yet and for a row that should not hold what it holds — and without it the only symptom is a delete that quietly did nothing, or a read shaped like a 404.

Two tests pin it: one asserts the warn fires with those fields, one asserts a contained path stays quiet. The `beforeEach(vi.clearAllMocks)` is load-bearing for the second — a reviewer confirmed that removing it leaks five prior calls in and the quiet assertion fails. The logger is mocked in `storagePath.test.ts` and in `fitness-files/localFile.test.ts`, whose containment test deliberately makes it refuse twice.

lib/services/medias/localFile.test.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed.** Added a driver-level test that builds the storage config from `path.relative(process.cwd(), mediaRoot)`, asserts the root really is relative, then checks both halves: a contained delete succeeds, and a traversal is refused with the outside file still readable.

Good catch that `./uploads` is the shipped default rather than a hypothetical — that is the difference between a config shape worth covering and one worth skipping.

lib/services/medias/localFile.test.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed.** Added `reports a missing file inside the media root as already deleted`.

Its comment names why it belongs here rather than anywhere else: it returns the *opposite* boolean to the refusal test directly above it, for a path that is also not present. The two are easy to conflate and nothing pinned which was which. A reviewer separately confirmed it exercises the intended branch by asserting `logger.warn` is not called — i.e. the `true` comes from the ENOENT catch, not from a refusal.

lib/services/medias/storagePath.test.ts
llun: (Reposting — my earlier reply here posted a file path instead of its contents.)

**Fixed.** Shortened to `'an inner traversal'`, which renders untruncated in the reporter.

llun:
## Sub-agent review loop — rounds 2, 3 and 4

Round 1's nine findings are on their own threads above, each replied to and resolved. Rounds 2–4 each reviewed **the previous round's fixes** (`git diff <prev>..HEAD`) rather than the original diff, which is where all of the value came from. Recording them here rather than as a dozen threads I would open and close myself.

The headline: **the production code has not changed since round 1.** `storagePath.ts`, both drivers, and `cleanupMediaStorage.ts` were settled early and every subsequent round confirmed them. Rounds 2–4 are entirely the story of one *test* — the call-site guard — being wrong four times in a row, which is itself the most useful thing the loop produced.

### Round 2 — reviewing round 1

- **The guard test was disarmed by `prettier --write`.** It scanned line by line, so wrapping `path.resolve(this._config.path, filePath)` across lines hid the call — and prettier is step 1 of the commit gate. Three more bypasses alongside it: a destructured rename, a nested call in the argument list, and an extracted `startsWith` variable.
- **A false claim in `AGENTS.md`/`CLAUDE.md`.** They said PR #1569 should fix the archive traversal with `assertStorageFilePath`. It does not, and should not — it uses `isTraversingStoragePath` in `getMediaPathFromFileUrl`, a shape check that takes no root and runs after decoding. Corrected to state the two are complementary boundaries.
- CodeQL flagged partial regex escaping in the round-1 guard; the rewrite removed the interpolation rather than patching it.

### Round 3 — reviewing round 2

- **False negative:** the "outright ban" banned a *spelling*, not a function. `path['resolve'](…)` and `import { resolve as pathResolve } from 'path'` both walked past it.
- **False positive:** the extracted-variable check matched a bare identifier with no scope awareness, so an unrelated `const resolvedArchivePath = 'database.'` in `scripts/backup/productionArchive.ts` failed the suite with a confident, wrong diagnosis.
- That check was **dropped rather than patched a third time**, and its residual documented. A false positive in a guard is worse than a gap, because the fix is to weaken the guard.
- The over-corrected #1569 wording was corrected again — `isTraversingStoragePath` does not live in `mediaFileUrl.ts` in this repo yet; #1569 only proposes it. The docs now state the durable invariant and drop the unmerged PR's internals.

### Round 4 — reviewing round 3

- **False negative:** the import-shape regex had no `s` flag, so a **wrapped** `path` import matched *nothing* — the assertion passed vacuously while the file called a bare `resolve(…)`. Prettier keeps multi-specifier imports wrapped, so this is an ordinary shape. Round 3's "leaves no spelling open" was therefore false, in the test comment and in `AGENTS.md`.
- **False positive, again:** the substring scan counted matches inside **comments**, so documenting this very rule in a driver failed the suite.

The import rule is now **pinned rather than pattern-matched**: the drivers must reach the path module through exactly one `import path from 'path'`, counted by every reference to the module, so a second import, a replacement, a wrapped named import, a `require` or a dynamic `import` all fail. With the import pinned, banning member and computed access *is* exhaustive for those two files — the claim the previous version could not support. Scanning strips comments.

Seven bypass shapes and two false-positive shapes now have revert proofs: plain `path.resolve`, destructured alias, prettier-wrapped call, `path['resolve']`, renamed named import, multi-line named import, dynamic `import('path')`, and the inline `startsWith` idiom all fail the guard; the `productionArchive.ts` name reuse and a driver comment naming `path.resolve` both pass.

### What this says about the guard

Four rounds, four leaks, each from reaching for more than a text scan can decide. The residual it does not cover — following an identifier back to its declaration — is now named in the test header, in `AGENTS.md` and in `REVIEW.md`, rather than implied away, and its proper home is an AST rule in `lint/agentsRules.mjs` (`no-component-fetch` is the precedent), where formatting and aliasing are invisible by construction. That is filed as a follow-up, with the note that `.oxlintrc.json` currently ignores `scripts/**` — exactly where the real bug was found — so the migration must not narrow coverage.

### Disclosed, not changed

- `assertStorageFilePath` throws `'Storage path escapes storage root'` where `assertFitnessStoragePath` threw `'Fitness storage path escapes storage root'`. If it ever fires inside `resolveArchivePath`, `toImportErrorMessage` persists that string verbatim to `fitness_files.importError`, which the owner sees. Unreachable today; kept generic because the row it lands on is a `fitness_files` row.
- The containment check is **lexical** — a symlink planted under a storage root defeats it. Inherited from both predecessors; nothing creates symlinks in either root.
- `scripts/backup/actorArchive.ts` has a live instance of this class that is **not** fixed here; PR #1569 is open against it.

Gate on each commit: `prettier` → `lint` → `typecheck` → `build` → `test`, all green (886 files, 11,593 tests).

llun:
## Round 5 — and the guard test's last round

Round 5 reviewed round 4's changes and found five more issues. One was severe and **live**, not hypothetical:

**`stripComments` read the `/*` inside the MIME string `'*/*'` as opening a block comment.** On `lib/services/link-previews/fetchLinkPreview.ts` that fake comment ran 2,126 characters to the next real `*/`, blanking `LinkPreviewFetchError`, `getErrorCode`, `parseContentType` and `isFresh` from the scan. The reviewer proved it by injecting the exact `startsWith(path.resolve(...))` idiom the guard exists to catch into that span — all assertions passed. `csp.ts` and `activityPubContentNegotiation.ts` carry the same wildcard strings.

That is the same vacuous-pass class the file's own header comment boasted about having fixed, reintroduced with a far larger blast radius. Alongside it: three more bypasses (destructuring off the default import, `path?.resolve`, a helper module re-export) and two false positives (`node:path`, any quoted `'path'` string).

### The decision

Five rounds, five sets of holes, in both directions — and **every one of them came from a mechanism added to widen coverage.** Alias tracking, an import-shape regex, comment stripping: each was the thing that broke.

So the guard is cut back rather than patched a sixth time. Two assertions over **raw** source — no interpolation to escape, no stripping pass to corrupt — which means it cannot pass vacuously. A comment naming a banned pattern now fails loudly instead of quietly disabling the scan, and loud beats silent.

**What it no longer claims is the point.** It catches the obvious spelling of a missing call site. A destructured `const { resolve } = path`, `path?.resolve`, a renamed import and a helper resolving on a driver's behalf all pass it — and a **reviewer** is the enforcement for those until the AST rule lands in `lint/agentsRules.mjs`. `AGENTS.md`, `CLAUDE.md` and `REVIEW.md` now say exactly that instead of implying completeness.

Proofs on the current version: the round-5 injection hidden inside the `*/*` span is caught, a plain `path.resolve` in a driver is caught, the inline `startsWith` idiom in the cleanup script is caught, and `node:path` no longer false-positives.

### Merged `main`

`origin/main` is merged in. #1569 landed at 15:30 today, six commits after this branch forked — so this branch was describing it as open, and `copyProfileImage` looked like a live arbitrary-file read here purely because the fix wasn't in its ancestry. The `REVIEW.md` conflict is resolved keeping both sides.

### A scope claim the verifier disproved

`AGENTS.md` said "every filesystem path built under a storage root" goes through the helper. `scripts/backup/productionArchive.ts` does not — it has its own `assertRelativeFilePath` / `normalizeStoragePath`, stricter than this module and applied to archive-entry names rather than root-relative stored paths. Different question, different input, not drift. The claim is now scoped to the storage drivers, and that checker is documented as deliberate.

### The adversarial verifier's verdict

A verifier pointed at the complete diff **could not refute** the PR's core claim within its scope: every call site it traced goes through the shared helper, both `S3StorageFile.ts` drivers import neither `fs` nor `path`, and `DELETE /api/v1/media/:id` and `DELETE /api/v1/statuses/:id` are clean from request to `fs.unlink`.

**The production code has not changed since round 1.** `storagePath.ts`, both drivers and `cleanupMediaStorage.ts` were settled in the first round and every round since has confirmed them. Rounds 2–5 were entirely about one test.

Gate on this commit: prettier → lint → typecheck → build → test, all green (886 files, 11,660 tests).

llun:
## Rounds 6 and 7 — loop closed

**Round 6** produced no correctness findings and a mergeable scope verdict, with one documentation trim: the `AGENTS.md` and `CLAUDE.md` entries had grown a blow-by-blow of all five failed revisions of the guard test, near-verbatim with what already lives in that test's header. This project's idiom is to cite recurrence in a clause as evidence and then state the rule, never to replay it. Trimmed in `b5af1e8c5` — the forensics stay beside the code they explain, where the next person to make that file cleverer will read them first. (Round 3 cut this same narrative and round 5 grew it back, which is itself the argument for it having one home.)

Round 6 also found `copyProfileImage` in `scripts/backup/actorArchive.ts` still hand-rolls an equivalent containment check in the `path.relative` form — not broken, since that form lacks the separator-boundary bug, but a third living implementation. Filed as a follow-up rather than widening this PR again.

**Round 7** independently re-verified everything and came back clean:
- `deleteFile` routes through `resolveStorageFilePath` before `fs.unlink`; both drivers route every read/write/delete through the helper, with zero stray `path.resolve`/`path.join` remaining in either.
- `cleanupMediaStorage.ts`'s missing-separator-boundary bug is fixed; `fitness-files/path.ts` is gone with no dangling references.
- Happy paths pass: upload, media read, fitness save and read back, the Strava archive job's path resolution, and both delete routes.
- Full suite: 886 files, 11,660 tests.
- **Revert proof, one last time:** restoring `deleteFile` to the unguarded `path.resolve` + `unlink` fails exactly four assertions — the three traversal cases plus the call-site guard catching the reintroduced raw call. Restored byte-for-byte, 40/40 green again.

That is two consecutive clean rounds, so the loop stops here at seven rounds of a twenty-round cap.

### What the loop actually found

Round 1's nine findings included one real bug the original task did not cover — a **third** copy of this check in `scripts/maintenance/cleanupMediaStorage.ts`, guarding an `fs.unlink`, whose `startsWith` against a bare resolved root accepted a sibling directory the root's name prefixes.

Everything after round 1 was about one test. **The production code has not changed since round 1** — `storagePath.ts`, both drivers and the cleanup script were settled early, and rounds 2 through 7, including an adversarial verifier over the complete diff, only ever confirmed them.

### Standing caveats for the reviewer

- The containment check is **lexical**; a symlink under a storage root defeats it. Inherited, documented, and not reachable today.
- The guard test catches the obvious spelling only. A reviewer is the enforcement for a destructured `resolve`, `path?.resolve`, a renamed import, or a helper resolving on a driver's behalf, until the AST rule lands.
- Two follow-ups are filed: move the guard to an AST rule in `lint/agentsRules.mjs` (noting `.oxlintrc.json` ignores `scripts/**`, where the real bug was found, so it must not narrow coverage), and consolidate `copyProfileImage`'s check.
- GitHub Actions was unreliable across this repo during the review — `startup_failure` and no-job `failure` runs on several unrelated branches, and some pushes here got no run at all. The local gate ran green on every commit.

PR #1576, opened 2026-08-26
none: pin the blurhash backfill's dry run, paging, summary and mock reset

llun:
## Summary

Four coverage gaps in `scripts/maintenance/backfillMediaBlurhash.test.ts` that mutation testing showed would let a real regression ship silently. They were found while reviewing #1566 and left out of it because they concern code that PR did not touch.

Every mutation below was applied to `backfillMediaBlurhash.ts` and the whole file run against it. **All four passed before this change**; each new test fails against its own mutation now, and every one was reverted afterwards.

| # | Mutation | Before | After |
| - | -------- | ------ | ----- |
| 1 | `backfillMedias`: collapse the `if (options.dryRun)` gate to an unconditional write | 75 passed | `counts a row it would write under --dry-run without writing it` fails |
| 2 | `backfillAttachments`: `.limit(options.batchSize)` → `.limit(50)` | 75 passed | `pages the attachment scan at the requested batch size` fails (`expected [ …(2) ] to have a length of 4 but got 2`) |
| 3 | `backfillAttachments`: move the `Attachments complete:` summary into the per-row loop — and, separately, into the batch loop | 75 passed | `logs the attachments summary once, as the last line of the sweep` fails against **both** placements |
| 4 | Remove the new `beforeEach` mock reset | n/a | the guard test **and** `skips the remote download entirely under --local-only` both fail |

### What each gap was

**1 — `backfillMedias` had no `--dry-run` coverage at all.** `--dry-run` is the first command `docs/maintenance.md` tells an operator to run, and a "simplify this" edit that inlined the write would have shipped undetected. The fixture has to diverge from the stored row, or the update block is never reached and "without writing" asserts nothing. The `backfillAttachments` sibling case added in #1566 is the model.

**2 — `options.batchSize` was unexercised for `backfillAttachments`.** At fixture scale one 50-row batch and several 1-row batches write the same rows, emit the same warnings and log the same counters. The `accumulates … across batches` case already runs at `batchSize: 1`, but on two rows — so it genuinely pins counter accumulation (resetting the counters per batch does fail it) and not that batching happens. Paging is only observable by counting the SELECTs off knex's `query` event.

**3 — the summary's position was unguarded.** Every assertion on it is `toHaveBeenCalledWith`, which asks whether the line was ever logged, not whether it was the only one — and the final row happens to carry correct cumulative totals. Two rows at `batchSize: 1` separate all three placements: per-row and per-batch each log twice.

**4 — `vi.restoreAllMocks()` never reset `analyzeImageBuffer` / `safeImageFetch`.** They are `vi.fn()`s created inside `vi.mock(...)` factories, and `restoreAllMocks` only iterates the set `vi.spyOn` populates. A probe confirmed the leak concretely: entering a test at the end of that block, `analyzeImageBuffer` still returned a neighbour's `{blurhash: 'L6PZ…', focus: null}` and carried five prior calls, and `safeImageFetch` still resolved `null`.

No existing test depended on the leak (verified: adding the reset alone left all 75 passing), so the fix is an explicit `beforeEach` reset. Two tests carried a local `vi.mocked(safeImageFetch).mockReset()`; both are removed. Only one of the two was actually load-bearing — `skips the remote download entirely under --local-only`, whose local reset was clearing call-history leaked from the preceding `downloadRemoteImage` describe, and which now genuinely depends on the shared hook (deleting the `beforeEach` fails it as well as the guard test). The other, in `runs a remote attachment URL through the download guard by default`, was already redundant with the `mockResolvedValue(null)` on the very next line, which overwrites any prior implementation regardless.

The guard test runs last in the block on purpose — but for its FAILING case, not its passing one: `beforeEach` reruns before every test either way, so what position buys is that a guard running first would find only `safeImageFetch` dirty (leaked from the preceding `downloadRemoteImage` describe) while `analyzeImageBuffer` is still untouched. `sequence.shuffle` is not configured, so file order is deterministic. The review loop measured the shuffled case: with the reset deleted and `--sequence.shuffle.tests --sequence.shuffle.suite`, 1 of 8 orders passed vacuously, and 5+ shuffled runs of the unmutated code never failed spuriously — a weakness, never a flake.

**Round 4** — testing and scope-regression, pointed at round 3's fixes. One real finding, plus two declined.

The finding was in this description, not the code: it credited the guard-comment rewrite to "round 2", which produced no commit — that text was written by the round-1 fix. Corrected here and in the commit message, along with an ambiguous "under one" in the trimmed comment.

Round 4 also confirmed the narrowed paging filter costs no coverage. It re-ran all three settled batch-size mutations (still caught), tried `.limit(options.batchSize).offset(0)` — which passes, correctly, because `.offset(0)` is a no-op against a `where('id', '>', lastId)` cursor and does not actually break paging — and could construct no mutation that breaks paging while keeping `limit` in the SQL. It verified by capturing live bindings that knex fully parameterises, so `includes('limit')` cannot match a fixture id. It also reproduced the hang empirically: freezing `lastId` and running under `timeout 8` exits 124 rather than failing an assertion.

Two declined, with reasons:

- **Re-expand the guard comment's mock-by-mock counterfactual.** Round 3's quality lens called that detail transcript and asked for it out; round 4's scope lens called it the evidence that made the claim falsifiable and asked for it back. Round 4 itself grades this "not a fact lost". Restoring it would just oscillate between two defensible readings, so the trimmed version stands.
- **Also flag the hang hazard beside `lastId` in `backfillMediaBlurhash.ts`.** A fair consistency argument — that script does keep its hazard notes beside the code they concern — but acting on it means touching production code this PR deliberately leaves alone, and round 4 grades it "could be better, not a broken fix". It is listed as a remaining gap instead; the follow-up that closes the cursor gaps touches that loop anyway and is its natural home.

## Known remaining gaps (NOT addressed here)

These are pre-existing holes in code this PR does not touch — the same category the four above were carved out of. Each was found by mutation during the review loop and each is flagged rather than silently folded in, because widening the PR past its stated scope would make it unreviewable against its own purpose. All are cheap to close in a follow-up.

| Surviving mutation | Why nothing catches it |
| ------------------ | ---------------------- |
| Delete the `focusX !== row.focusX` clause of `backfillAttachments`' change-detection OR — and, separately, the `focusY` one | Every fixture exercising the focal-point path also changes `blurhash` or `thumbnailUrl`, so those clauses mask the missing guard. Deleting either passes all 79. Deleting the `blurhash` clause fails 5 tests and the `thumbnailUrl` clause fails 1, so only these two are unguarded. A regression here would silently stop persisting focal-point-only repairs. |
| Delete the whole non-force `whereNull('blurhash').orWhere('thumbnailUrl', 'like', …)` filter | Every per-row write is *also* gated on `options.force` / existing-value checks, so removing the SQL filter never produces a wrong write — only a needless scan. Proven with a probe: one clean + one dirty attachment gives `processed 1` unmutated and `processed 2` with the filter gone. The filter's own comment frames it as scale-motivated, so this is load-bearing behaviour with zero coverage. |
| `lastId = String(rows[rows.length - 1].id)` → `rows[0].id` in `backfillAttachments` | No test uses `batchSize > 1` with more rows than one batch holds, so the two are indistinguishable everywhere they run. Lower severity — self-correcting, since `rows[0].id` still exceeds the previous `lastId`, so the loop terminates but reprocesses rows and inflates `processed`. |
| `.limit(options.batchSize)` in **`backfillMedias`** | The direct sibling of gap 2. Every test in that block inserts exactly one row, so hardcoding its limit passes too. |

One non-test item belongs with these: the cursor-loop hang hazard is documented beside the paging test, but not beside the `lastId` assignments in `backfillMediaBlurhash.ts` that actually cause it. Whoever closes the cursor gaps above will be editing those lines and should carry a one-line pointer there.

Say the word and I will close any or all of these.

## Docs

`AGENTS.md` → Testing Guidelines, `REVIEW.md` → Style, imports & tests, and the `CLAUDE.md` key reminders pick up the two reusable conventions: `vi.restoreAllMocks()` not reaching a `vi.mock` factory's `vi.fn()`, and `toHaveBeenCalledWith` being "was ever called" rather than "was the only call" (with the hardcoded-page-size blind spot that hides the same way).

## Review

The sub-agent review loop ran. **Round 1** — four lenses in parallel (design-system skipped: no `app/` or `lib/components/` changes):

- **correctness** — clean. Traced every asserted string, count and JSON key order against the untouched script; fact-checked the doc claims against Vitest's documentation. Raised one comment-precision nit.
- **testing** — all four guards independently proven by applying each mutation and reverting. Went further: confirmed the batch-size test is sensitive to the *value* (`.limit(2)` and `.limit(options.batchSize + 1)` are both caught, not just the hardcoded `.limit(50)`), and confirmed `TEST_DATABASE_TYPE=pg` is moot here by running against an unreachable PG host — 79 pass identically, because this file builds its own better-sqlite3 database. Raised the duplicate-name finding.
- **quality** — clean, no required changes.
- **scope-regression** — clean. Enumerated every test in the `execution` describe against the script's real branching and confirmed none previously depended on a leaked implementation to reach a path it now misses.

Both findings were posted as inline threads, fixed in `5cf4a0fa6`, and both subjects re-proven after the edits. Round 2 then verified that commit and found nothing, so it produced no commit of its own — the branch has three commits, not one per round.

**Round 2** — correctness and scope-regression, pointed at the round-1 fixes: **clean, no findings.** Both independently verified the rewritten comment's mechanical claim by reading the file — nothing before the execution block touches `analyzeImageBuffer`, and `downloadRemoteImage`'s last test does leave `safeImageFetch` with a rejected-value implementation and a non-empty call list — so the comment teaches a true mechanism.

**Round 3** — testing and quality, hunting for what rounds 1-2 missed rather than re-confirming. Three findings against this PR's own code, all fixed in `50493b090`:

- The paging test matched *every* SELECT against `attachments`, so an unrelated query added to the loop later would break it spuriously (proven: a per-batch diagnostic `count()` took it to 8 selects against an expected 4). Requiring a `limit` clause narrows it to the paging query; re-proven that `.limit(50)` still fails at 2, and that dropping `.limit()` entirely now fails at 0.
- The cursor-loop hang hazard lived only in this description. It is now a comment beside the paging test, where whoever extends that coverage will meet it.
- The guard test's comment, as rewritten by the round-1 fix (`5cf4a0fa6`), had over-corrected into arguing with round 1's finding in the source. Trimmed to the durable facts.

Round 3 also **disproved** two suspected weaknesses rather than confirming them: the summary test is not over-constrained (an unconditional per-row log does not break it), and appending a dirtying test *after* the guard does not silently reopen the hole it closes — re-running the reset-deleted mutation with such a test appended still failed the guard. It also found the four remaining gaps now listed below.

Two soft suggestions from the quality lens were declined with reasons: the fixture duplication between the two new attachment tests matches this file's own house style (~12 other inline `insert` call sites, no fixture helper exists to reuse), and merging them into an `it.each` would violate the project's rule that `it.each` is for cases differing only by input and expected value — they assert different subjects.

## Test results

Full gate run in order, all green:

- `yarn lint` — exit 0 (pre-existing warnings only)
- `yarn typecheck` — exit 0; this file is **not** in the `tsconfig.typecheck.json` ratchet, so it is genuinely covered
- `yarn build` — exit 0
- `yarn test` — **884 files, 11576 tests, all passing**

Re-run in full after each review round's fixes, not just at the start.

## Notes

- Tests and docs only; no runtime behaviour changes, hence the `none:` prefix.
- No migrations, so no schema-dump regeneration.
- No UI surface, so no screenshots.
- While pinning gap 2 I confirmed the hazard noted during #1566's review: forcing `lastId` never to advance does not fail cleanly, it hangs past Vitest's 30s `testTimeout` indefinitely, because better-sqlite3's synchronous driver wrapped in immediately-resolving promises inside `while (true)` never yields to Node's timer phase. No test here relies on that loop terminating incorrectly.


scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1251,6 +1339,36 @@ describe('backfillMedias', () => {
     expect(updated.focusY).toBe(-0.25)
   })
 
+  // `--dry-run` is the first command `docs/maintenance.md` tells an operator to
+  // run, and nothing exercised this half of the gate — collapsing it to an
+  // unconditional write passed the whole file. The row has to be one that
+  // genuinely diverges, or the update block is never reached and "without
+  // writing" asserts nothing; the `backfillAttachments` case is the model.
+  it('counts a row it would write under --dry-run without writing it', async () => {
```
llun: **Duplicate test name.** This new test is named `counts a row it would write under --dry-run without writing it`, which is character-for-character the name of the pre-existing test at line 945 — that one covers `backfillAttachments`' dry-run gate inside `describe('backfillMediaBlurhash execution')`, this one covers `backfillMedias`' gate.

Vitest disambiguates them by their parent `describe`, so nothing fails. But the two names are indistinguishable in `-t` filters and in CI reporter output, and this repo has already paid for that once: #1566 shipped `1146a91db test: separate two test names the reporter rendered identically` for exactly this.

Rename the new one so it names its own subject — e.g. `counts a media row it would write under --dry-run without writing it`.

Found by the testing lens (mutation-testing round 1).

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1077,6 +1149,22 @@ describe('backfillMediaBlurhash execution', () => {
       'https://remote.example/media/photo.jpg'
     )
   })
+
+  // A guard on the fixture rather than on the script, and it runs LAST on
```
llun: **The comment's ordering rationale is imprecise.** It says the test "runs LAST on purpose: it asserts the `beforeEach` reset cleaned up after every test above that dirtied these mocks."

Position buys nothing for the *passing* case — `beforeEach` reruns before every test, so with the reset present these assertions pass wherever the test sits. What position actually buys is the *failing* case: with the reset deleted, a guard running first would see only `safeImageFetch` dirty (leaked from the preceding `downloadRemoteImage` describe) and `analyzeImageBuffer` still clean, so it would miss half the leak.

That distinction matters because it also bounds the risk: the testing lens ran this under `--sequence.shuffle.tests --sequence.shuffle.suite` with the reset deleted, and 1 of 8 shuffled orders passed vacuously — a weakness, never a false failure (5+ shuffled runs of the unmutated code never spuriously failed). `sequence.shuffle` is not configured anywhere in this repo, so it is not a live problem, but the comment should say what ordering is for and note the shuffle caveat rather than implying the test would break without it.

Found by the correctness lens, quantified by the testing lens (round 1).

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in `5cf4a0fa6`. The `backfillMedias` case is now `counts a media row it would write under --dry-run without writing it`; the pre-existing `backfillAttachments` one at line 945 keeps its name, since it was there first and its own `describe` already says which function it covers.

Re-proven after the rename: collapsing `backfillMedias`' `if (options.dryRun)` gate to an unconditional write fails exactly the renamed test (`× counts a media row it would write under --dry-run without writing it`, 78/79 pass), then reverted.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in `5cf4a0fa6`. The comment now separates the two cases explicitly: position buys nothing for the passing case, because `beforeEach` reruns before every test regardless; what it buys is the failing case, where a guard running first would find only `safeImageFetch` dirty — leaked from the `downloadRemoteImage` block — while `analyzeImageBuffer` is still untouched.

It also now records the shuffle caveat you quantified: `sequence.shuffle` is not configured, so the order is deterministic, and a shuffled run degrades the guard to a vacuous pass rather than a false failure.

Re-proven after the rewrite: deleting the two `beforeEach` reset lines still fails the guard test, alongside `skips the remote download entirely under --local-only`, then reverted.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1031,6 +1035,90 @@ describe('backfillMediaBlurhash execution', () => {
...
+      }))
+    )
+
+    // `limit` is what narrows this to the PAGING query. Matching every SELECT
+    // against the table counts any other one the loop might later issue — a
+    // per-batch diagnostic count inflated this to 8 — while a limit clause is
+    // exactly what the assertion is about, and a mutation that drops `.limit()`
+    // altogether still fails, on zero matches rather than four.
+    const selects: string[] = []
+    db.on('query', ({ sql }: { sql: string }) => {
```
llun, line 1079: **The SELECT filter was too loose, making this test brittle.** It matched every `select` mentioning `attachments`, not just the paging query. Any other query the batch loop might later issue counted as a page — injecting a plausible per-batch diagnostic `count()` broke the test spuriously at 8 selects against an expected 4.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1077,6 +1164,25 @@ describe('backfillMediaBlurhash execution', () => {
...
+  // A guard on the fixture rather than on the script, placed last so it runs
+  // after the tests that dirty these mocks — `beforeEach` reruns regardless, so
+  // placement is what gives the guard a failing case rather than a passing one.
+  // The leak it guards: `vi.restoreAllMocks()` never reaches a `vi.fn()` a
+  // `vi.mock` factory created, so without the reset above these two carry their
+  // implementation and their call history across the whole block, and two tests
+  // here used to work around that with a local `mockReset()`. Nothing depended
+  // on the leak, but a test written against either mock's DEFAULT behaviour
+  // would silently inherit a neighbour's. `sequence.shuffle` is not configured;
+  // under one this degrades to a vacuous pass, never a false failure.
```
llun: **This comment over-corrected.** Round 1 flagged the original as imprecise about what the test's placement buys; the rewrite answered that objection *in the source*. The clauses "Position is not what makes it PASS…" / "Position is what gives it a FAILING case" are structured as a rebuttal, and the counterfactual is walked through mock by mock — forensic detail that reads as transcript rather than as a fact a maintainer needs.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1031,6 +1035,90 @@ describe('backfillMediaBlurhash execution', () => {
     }
   )
 
+  // `options.batchSize` bounds the SELECT, and nothing observed the paging. At
+  // fixture scale one 50-row batch and several 1-row batches produce identical
+  // writes, identical warnings and identical counters — the accumulation case
+  // above runs at a batch size of one on two rows and passes either way — so
+  // hardcoding the limit was invisible to the whole file. Counting the SELECTs
+  // is what tells the two apart.
+  it('pages the attachment scan at the requested batch size', async () => {
```
llun, line 1058: **A real hazard around these cursor loops is documented nowhere durable.** Forcing `lastId` never to advance does not fail cleanly — it hangs well past the 30s `testTimeout`, because better-sqlite3 is a synchronous driver and the awaits inside the sweep's `while (true)` settle on the microtask queue without ever yielding to the timer phase the watchdog lives in.

That fact currently lives only in this PR's description. Once this merges it survives as PR archaeology, and the next contributor to write the natural follow-up test — stall the cursor, assert the loop does not spin — hangs CI with no clue why.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in `972483fed` — the filter now also requires a `limit` clause, which is exactly what the assertion is about.

Verified all three ways: with the diagnostic `count()` injected into the loop the test now passes (79/79); the real `.limit(50)` mutation still fails it at 2 against 4; and dropping `.limit()` altogether now fails it at 0 against 4, so a removed limit is caught rather than silently matching nothing.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in `972483fed`. The PASS/FAIL framing and the mock-by-mock counterfactual are gone; the same fact now reads as one clause — placement is what gives the guard a failing case rather than a passing one — followed by the mechanism (`vi.restoreAllMocks()` not reaching a `vi.mock` factory's `vi.fn()`) and the shuffle caveat.

The shuffle sentence is deliberately kept in the source rather than moved to the PR body. It is a forward-looking caveat about this guard's own strength — it matters to whoever later enables `sequence.shuffle` globally — and the companion finding on this same round argues the opposite direction for PR-body prose, that load-bearing facts should not survive only as PR archaeology. Keeping it inline is the consistent reading.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in `972483fed` — it is now a comment beside this test, which is where whoever extends the paging coverage will be reading.

Deliberately placed in the test file rather than beside the loops in `backfillMediaBlurhash.ts`: the person who trips over it is writing a test, and it keeps this PR's "no production code touched" property, which several review lenses relied on when reasoning about the diff.

PR #1577, opened 2026-08-26
fix: reject a blurhash that decode cannot read, and pin the guard

llun:
Pins the blurhash decode guard, and makes the validator agree with what `decode` will actually accept.

## The guard

`lib/components/posts/BlurhashCanvas.tsx` wraps `decode` in a `try/catch`. That `catch` is the app's only containment for a hostile or malformed blurhash: there is no error boundary between a post and the root, so the nearest one is `app/error.tsx` — an escaping throw replaces the whole page with the error screen, and one bad attachment would take the timeline with it. It had zero coverage, and `media.test.tsx` mocked the component out entirely, so deleting the `catch` broke nothing.

`BlurhashCanvas.test.tsx` now covers it. Deleting the `try/catch` fails three of its tests.

## The validator

`isValidBlurhash` was charset-only, so `'aaaaaa'` — legal base83, legal length — passed and then threw inside `decode`, which derives the required length (`4 + 2 * componentX * componentY`) from the size flag in the first character. It now also runs the blurhash package's own `isBlurhashValid`. Both halves are load-bearing and neither subsumes the other: `isBlurhashValid` does not look at the alphabet at all, and `decode` happily returns pixels for characters outside base83.

Two further problems came out of the review loop, both of which this PR now fixes.

**The validator checked a different string than the caller stored.** It compared `hash.trim()`, but `createNoteJob` stored `attachment.blurhash` — the value it was handed. A federated hash padded with whitespace was therefore approved on the trimmed copy and persisted untrimmed, and `decode` threw on every render:

```
trailing space   valid=true   decode(as stored) THROWS: length is 29 but it should be 28
```

The blur placeholder painted as an empty canvas, and the padded value was re-served verbatim to third-party Mastodon clients. The `.trim()` predates this PR; tightening the check while leaving it aimed at a different string is what left the hole open. A predicate cannot say "and store this instead", so `normalizeBlurhash` returns the form to store or `null`, and the caller stores what it returns. `isValidBlurhash` is removed rather than kept alongside — once `createNoteJob` moved over it had no production caller, and what remained was an exported predicate whose own documentation had to warn that saying yes to a value did not make it safe to store.

**The length cap rejected hashes `decode` reads fine.** A well-formed blurhash runs to `4 + 2 * 9 * 9` = 166 characters, so the old 100-character cap sat two below a 7x7 hash and silently dropped the blurhash of any remote encoding above 6x8 components — the same bug pointing the other way. Raised to 166; the column is `varchar(255)`, so the widest legal hash fits.

## Scope

The fix is **inbound-only**. `createNoteJob` is the sole path that stores a remote-supplied blurhash (verified against every `createAttachment` caller: `updateNoteJob` handles no attachments, and the local, Strava and fitness writers all use a locally computed hash). Rows written before this change can still hold a padded or structurally invalid blurhash, and nothing re-validates on read — the `try/catch` this PR put under test is what keeps those harmless. `backfillMediaBlurhash.ts` does not reach them by default, since its selection matches on a null blurhash or a host-relative `thumbnailUrl`. `--force` does repair them, but only by re-downloading and recomputing every remote image in the instance — and not at all for a non-image attachment, whose `mediaType` fails the `shouldAnalyze` guard. Repairing the residue properly needs no image downloads, only a re-run of `normalizeBlurhash` over stored rows; filed as a follow-up rather than widened into this PR.

## Tests

Every guard was proved by breaking its subject, not by being read:

| Mutation | Result |
| --- | --- |
| `normalizeBlurhash` returns the original, not the trimmed form | 5 fail |
| structural check dropped | the 3 size-flag rows fail |
| regex check dropped | the 3 charset rows fail |
| `createNoteJob` stores the raw federated value | 1 fails, on the stored value |
| length cap back to 100 | 7x7 and 9x9 fail; 6x8 correctly still passes |
| space admitted to the base83 charset | the interior-space row fails |
| `` ref callback loses its `complete` check | the remount test fails |
| effect's `else` reset deleted | the url-change test fails |
| whole mount effect deleted | the url-change test fails |
| `BlurhashCanvas` `try/catch` deleted | 3 fail |

Three test defects the review loop found, all introduced by this PR:

- the charset block's space case used a **trailing** space, which is trimmed away before the regex sees it; the 27-character body left behind is pure base83, so the regex accepted it and the structural half did the rejecting — it pinned the opposite of what its comment claimed. Now an interior space, which survives the trim.
- `vi.restoreAllMocks()` sat as the last statement of a test body, so a failing assertion above it skipped restoration and left `HTMLImageElement.prototype.complete` mocked `true` for the rest of the file. Forcing that test to fail costs 1 failure with an `afterEach` and 2 without, the second being the test directly below it.
- the cached-image test claimed to cover the `ref` callback; it covered neither guard on its own. The inline `ref` arrow is a new function every render, so React reattaches it constantly and it can only set `isLoaded` true, while the effect re-runs only on a url change and is the only thing that can set it false. On mount both fire, so deleting either alone left the test green. Two tests now isolate them — one holds the url steady, the other changes it — and the second closes a branch that previously had no coverage at all.

Review: two rounds of the sub-agent loop, plus a third on the fixes. Round 1 found 5 real issues (this PR did not close the class it was opened for); round 2 found 0 actionable.


lib/services/medias/imageAnalysis.ts
```diff
@@ -18,12 +18,47 @@ const BLURHASH_MAX_DIMENSION = 32
...
+ * Returning the string rather than a boolean is load-bearing. The check has
+ * always compared `hash.trim()`, but `createNoteJob` stored the value it was
+ * handed, so a federated hash padded with whitespace validated on the trimmed
+ * copy and was persisted untrimmed. `decode` then threw on every render
+ * (`length is 29 but it should be 28`), the blur placeholder rendered as an
+ * empty canvas, and the padded value was served on to third-party clients
+ * verbatim. A predicate cannot say "and store this instead", so callers store
+ * what this returns.
+ *
+ * BOTH checks are load-bearing and neither subsumes the other.
```
llun, line 42: **Finding (round 1, medium): the validator checked a different string than the caller stored**

`isValidBlurhash` compared `hash.trim()`, but `createNoteJob` stored `attachment.blurhash` — the value it was handed. A federated hash padded with whitespace was therefore approved on the trimmed copy and persisted untrimmed, and `decode` threw on every render (`length is 29 but it should be 28`). Measured, not inferred. The padded value was also re-served verbatim to third-party Mastodon clients via `lib/types/domain/attachment.ts`.

Two independent lenses found this; the verifier reproduced it end to end against a real SQLite database. Two corrections it made, both adopted: the `.trim()` PREDATES this PR (main's `BLURHASH_REGEX.test(hash.trim())` reproduces it identically), so this PR narrowed the hole rather than opening it; and the harm is the blur placeholder painting as an empty canvas, not the `opacity-0` pre-load state, which `onLoad` clears.

lib/services/medias/imageAnalysis.test.ts
```diff
@@ -1,26 +1,157 @@
...
+    // accepts and the structural half rejects for its length - which would
+    // pin the opposite of what this block claims.
+    it.each([
+      {
+        description: 'an exclamation mark',
+        blurhash: 'L6PZfSi_.AyE_3t7t7R**0o#DgR!'
+      },
+      { description: 'a backslash', blurhash: 'L6PZfSi_.AyE_3t7t7R**0o#DgR\\' },
+      {
+        description: 'an interior space',
```
llun, line 100: **Finding (round 1, low): the charset test's space case was rejected by the structural half, not the regex**

*Anchor note: the finding was originally filed at `imageAnalysis.test.ts:68`. At the current head commit, line 68 falls inside the "rejects a hash $description" (structural) `it.each` block rather than the charset block — the file grew a "cap" test block between round 1 and now, shifting line numbers. Anchoring instead to line 99, the actual `'an interior space'` row this finding is about.*

The block is introduced by "The structural check does NOT look at the alphabet, so the regex is not redundant", but the space row used a TRAILING space. `isValidBlurhash` trims first, so the regex saw the 27-character body `L6PZfSi_.AyE_3t7t7R**0o#DgR`, which is pure base83 and ACCEPTED; the false verdict came from `isBlurhashValid` rejecting the length. The row would keep passing if the whitespace exclusion were removed from `BLURHASH_REGEX`, so it pinned the opposite of what its comment claimed.

lib/components/posts/media.test.tsx
```diff
@@ -99,6 +107,104 @@ describe('Media', () => {
...
+  // This covers the mount `useEffect`, NOT the `ref` callback: refs attach
+  // before effects fire, but the effect reads the same two getters through
+  // `imgRef.current` and reaches the same result, so deleting the callback's
+  // copy leaves this test green. The test below pins the callback itself.
+  it('reveals a cached image that was already complete on mount', () => {
+    const attachmentWithBlurhash: Attachment = {
+      ...baseAttachment,
+      blurhash: 'L6PZfSi_.AyE_3t7t7R**0o#DgR4'
+    }
+
```
llun, line 126: **Finding (round 1, low): mock restoration sat in the test body, leaking prototype spies on failure**

`vi.restoreAllMocks()` was the last statement of the cached-image test, so a failing assertion above it skipped restoration and left `HTMLImageElement.prototype.complete` mocked `true` for every later test in the file — turning one failure into a cascade that masks it.

lib/services/medias/imageAnalysis.ts
```diff
@@ -18,12 +18,47 @@ const BLURHASH_MAX_DIMENSION = 32
 const BLURHASH_COMPONENTS_X = 4
 const BLURHASH_COMPONENTS_Y = 4
 
-// Standard blurhash format regex: base83 characters, length 6 to 100
-const BLURHASH_REGEX = /^[0-9a-zA-Z#$%*+,\-.:;=?@[\]^_{|}~]{6,100}$/
+// The base83 alphabet, and the length a well-formed blurhash can occupy: the
+// shortest is 1x1 (`4 + 2 * 1 * 1` = 6) and the longest 9x9 (`4 + 2 * 9 * 9` =
```
llun, line 22: **Finding (round 1, low): the 100-character cap rejected blurhashes decode reads fine**

A well-formed blurhash runs to `4 + 2 * 9 * 9` = 166 characters, so the 100-character cap sat two characters below a 7x7 hash and silently dropped the blurhash of any remote encoding above 6x8 components. Same bug as the sibling finding on this validator, pointing the other way: rejecting what `decode` accepts. Round 1 rated this informational and pre-existing.

lib/components/posts/BlurhashCanvas.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+ * @vitest-environment jsdom
+ */
+import '@testing-library/jest-dom'
+import { render } from '@testing-library/react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { BlurhashCanvas } from './BlurhashCanvas'
+
+// A real blurhash from a stored image, and two that are well-formed base83 of a
+// legal length but still throw inside `decode`, which derives the required
```
llun, line 11: **Finding (round 1, low): the canvas test's comments described a validator that no longer exists**

Not from the review lenses — found while grepping for stale references to the removed export. The comments asserted these fixtures "pass `isValidBlurhash`, which only checks the base83 charset and a 6..100 length". After this PR the inbox rejects both, and `isValidBlurhash` is gone. The tests still pass because they drive the canvas directly, but the comments would have been actively misleading about why the guard exists.

lib/components/posts/media.test.tsx
```diff
@@ -99,6 +107,104 @@ describe('Media', () => {
...
+    expect(screen.getByRole('img')).toHaveClass('opacity-100')
+    expect(screen.getByTestId('blurhash-canvas')).toHaveClass('opacity-0')
+  })
+
+  // The `ref` callback's own `node.complete && node.naturalWidth > 0` check.
+  // The mount effect covers the same ground, but its dependency array is
+  // `[attachment?.url]`, so it does not re-run when the `<img>` remounts on a
+  // changed `key={id}` with the url unchanged. In that case the callback is
+  // the only thing that reveals an already-painted picture.
+  it('reveals a cached image when the img remounts without the url changing', () => {
```
llun, line 145: **Finding (round 1, low): the ref callback's complete/naturalWidth check was covered by no test**

*Anchor note: this finding was originally filed at `lib/components/posts/media.tsx:103` (the `` ref callback). That file has NO changes in this PR — it isn't part of the diff at all, so the GitHub API refuses a line comment there ("could not be resolved"). Redirecting to `lib/components/posts/media.test.tsx`, at the new regression test this finding led to, which pins the exact line in `media.tsx` the finding is about.*

Deleting `if (node?.complete && node.naturalWidth > 0 && !isLoaded)` from the ref callback left the whole media suite green, including the test whose own comment claimed to cover that branch. The mount `useEffect` reads the same two getters through `imgRef.current` and independently reaches the same result. The callback matters only for an `` remounted by a changed `key={id}` with the url unchanged, which the effect's `[attachment?.url]` dependency does not re-run for — and no test exercised that.

lib/services/medias/imageAnalysis.ts
llun: Fixed in 9f516bf7d. `normalizeBlurhash` returns the form to store or null, and `createNoteJob` stores what it returns — a predicate cannot say "and store this instead". Revert-proof: making the normalizer return the original instead of the trimmed form fails 4 tests; making `createNoteJob` store the raw value fails the new federation test with `expected '  LEHV6nWB2yk8pyo0adR*.7kCMdnj\n' to be 'LEHV6nWB2yk8pyo0adR*.7kCMdnj'`.

lib/services/medias/imageAnalysis.test.ts
llun: Fixed in 9f516bf7d — the row now uses an INTERIOR space (`L6PZfSi_.AyE_3t7 7R**0o#DgR4`), which survives the trim, keeps the length structurally valid, and is rejected by the regex. Verified: `isBlurhashValid` returns true for it and `decode` returns pixels, so the charset check is the only thing rejecting it. Revert-proof: admitting a space into the character class fails this row with `expected 'L6PZfSi_.AyE_3t7 7R**0o#DgR4' to be null`. A comment now records why the space must be interior.

lib/components/posts/media.test.tsx
llun: Fixed in 9f516bf7d by moving restoration to an `afterEach` at the describe level. Proved by measurement rather than assertion: forcing the cached-image test to fail costs exactly **1** failure with the `afterEach`, and **2** without it — the second being `leaves the placeholder painted when the image fails to load`, the test directly below, exactly the masking cascade predicted.

lib/services/medias/imageAnalysis.ts
llun: Fixed in 9f516bf7d, and this is a judgement call worth flagging rather than burying — round 1 explicitly declined to call it a defect. Included because this PR's whole thesis is aligning the validator with `decode`'s actual contract, and leaving a known false-reject in the same function would contradict the doc comment being shipped. Verified empirically: 7x7 = 102 chars and 9x9 = 166 chars are both structurally valid and decode fine, and the column is `varchar(255)` so the widest legal hash fits. Revert-proof: restoring the 100 cap fails the 7x7 and 9x9 rows while 6x8 (exactly 100) correctly still passes. Happy to split this back out if it is unwanted scope.

lib/components/posts/BlurhashCanvas.test.tsx
llun: Fixed in 9f516bf7d. The comments now say the inbox rejects these today but did not always, and is not consulted again on read — so rows written before the check became structural still reach this component, which is what keeps its `try/catch` load-bearing.

lib/components/posts/media.test.tsx
llun: Fixed in 9f516bf7d. The overclaiming comment is corrected to say it covers the mount effect, and a new test `reveals a cached image when the img remounts without the url changing` pins the callback: it renders with the image incomplete, then re-renders with a new `id` and the same url while the stubbed getters flip to complete. Revert-proof: stripping the ref callback's check fails that test and only that test.

lib/services/medias/imageAnalysis.ts
llun: Confirmed by round 2 (0 actionable) and round 3 (0 actionable) — two consecutive clean rounds, so the loop stops here. Round 2 independently re-ran the mutation matrix: returning the untrimmed form fails 5 tests (the four padding rows plus the createNoteJob federation test), dropping the structural check fails exactly the three size-flag rows, and dropping the regex fails exactly the three charset rows. It also confirmed `createNoteJob` is the sole federated writer by tracing every `createAttachment` caller.

lib/services/medias/imageAnalysis.test.ts
llun: Confirmed by round 2 (0 actionable) and round 3 (0 actionable) — two consecutive clean rounds, so the loop stops here. Round 2 also flagged that the test was restating BLURHASH_REGEX as a hand-synced literal whose cap portion asserted nothing — mutating the test copy's cap to 100, to 1000, or to `+` left all 26 tests green. Fixed in a01b4199b by exporting the production constant and importing it, so the two cannot drift.

lib/components/posts/media.test.tsx
llun: Confirmed by round 2 (0 actionable) and round 3 (0 actionable) — two consecutive clean rounds, so the loop stops here. Round 3 re-ran media.test.tsx under `--sequence.shuffle` across 12 seeds and confirmed the `afterEach` makes the prototype spies order-independent, including between the two tests added since.

lib/components/posts/media.test.tsx
llun: Round 2 found this fix was itself still wrong, and it is now properly closed in a01b4199b.

My round-1 comment claimed the cached-image test covered the mount `useEffect` rather than the `ref` callback. Deleting the effect outright left all 693 tests green, so it covered neither. The real structure: the inline `ref` arrow is a new function every render, so React reattaches it constantly and it can only ever set `isLoaded` true; the effect re-runs only on a url change and is the only thing that can set it false. On mount both fire, which is why either could be deleted invisibly.

That exposed a genuinely uncovered branch, not just a wrong comment — the effect's `else` reset, which repaints the placeholder when a new url swaps in. Without it the incoming picture shows at full opacity wearing the previous image's loaded state. Two tests now isolate the two guards and the mutations are disjoint: stripping the ref callback fails only the remount test; deleting the reset, or the whole effect, fails only the url-change test.

Round 3 tried to refute the remount test as not actually remounting and could not — a node-identity probe gives `sameNode=false` on the id change against `sameNode=true` without it.

lib/services/medias/imageAnalysis.ts
llun: Confirmed by round 2 (0 actionable) and round 3 (0 actionable) — two consecutive clean rounds, so the loop stops here. Round 2 specifically hunted for a consumer assuming the old 100-character bound — zod schemas, every blurhash column in both schema dumps, the Mastodon serializers and the AP emitters — and found none. Flagging again that this half was my judgement call rather than a review finding, and it is still easy to drop if you would rather keep the PR to the reject-side fix.

lib/components/posts/BlurhashCanvas.test.tsx
llun: Confirmed by round 2 (0 actionable) and round 3 (0 actionable) — two consecutive clean rounds, so the loop stops here. Round 3 checked every doc and comment this PR touches for staleness and found none outstanding, bar one prose hyphen fixed in b1ab45c6a.

PR #1583, opened 2026-08-26
refactor: consolidate the actor archive's containment check onto the shared helper

llun:
## What

`copyProfileImage` in [scripts/backup/actorArchive.ts](scripts/backup/actorArchive.ts) hand-rolled the containment check that `resolveStorageFilePath` ([lib/services/medias/storagePath.ts](lib/services/medias/storagePath.ts)) already owns — "does this path resolve inside that root?" — with `path.relative` and a three-way disjunct instead of the shared separator-bounded compare.

```diff
-  const relativePath = path.relative(mediaDir, source)
-  if (
-    relativePath === '..' ||
-    relativePath.startsWith(`..${path.sep}`) ||
-    path.isAbsolute(relativePath)
-  ) {
+  const source = resolveStorageFilePath(mediaDir, storagePath)
+  if (!source) {
```

## This fixes nothing, and that is worth stating

The `path.relative` form carries the separator boundary for free, so it never had the bug that made `scripts/maintenance/cleanupMediaStorage.ts` accept `../uploads-backup/x` for a root of `/srv/uploads` before #1575. It moves because #1575 extracted the helper and converted three other copies (`fitness-files/path.ts`, `importStravaArchiveJob`, `cleanupMediaStorage`) and stopped one file short of this one. A fourth spelling of one question is a fourth place to forget when the shared answer changes.

The two forms agree on every input that can reach this call site. They are **not** identical everywhere, and review caught me overstating that: `path.win32.relative` compares case-insensitively where the prefix compare does not, so on Windows an *absolute* path differing from the root only by letter case was accepted by the old spelling and is refused by this one. Windows-only, fail-safe in direction, unreachable (`isTraversingStoragePath` refuses every absolute path upstream), and inherited from the helper as shipped in #1575 rather than introduced here. A 200,000-case fuzz over both implementations found **0** inputs the new form accepts that the old refused, against 1,241 in the documented direction.

## What is deliberately unchanged

- **The local failure mode.** The `warnings.push` and `return null` stay — byte-for-byte the same warning strings as `main` — so the archive still tells its operator which profile image it dropped and why. This is the pattern `cleanupMediaStorage.ts` uses: take `resolveStorageFilePath`, supply your own message on null. The helper's `logger.warn` is additive.
- **Both boundaries.** `getMediaPathFromFileUrl` already refuses a traversing path at URL recovery via `isTraversingStoragePath` (#1569). That is a shape check with no root; this is a root check. Neither subsumes the other, and `copyProfileImage` keeps both.
- **`scripts/backup/productionArchive.ts`.** Untouched. Its `assertRelativeFilePath` / `normalizeStoragePath` reject any `..` outright and apply to archive-entry names rather than root-relative stored paths, which AGENTS.md documents as intentional.

## Siblings checked

- `headerPath` needs no separate change — it flows through this same function; `copyProfileImage` is one implementation for both `avatar` and `header`.
- The only other path builder in the file, `registerAttachmentUrl`'s remote-attachment write, names its file from a sha256 under the staging directory, so it has no stored path to confine.
- The remaining `path.relative` sites in `lib/config/fitnessStorage.ts` and `cleanupMediaStorage.ts`'s `getContainedRelativePath` ask a different question — is one configured root nested inside another, and what is the prefix — and need the relative string as output. `resolveStorageFilePath` returns an absolute path or null and cannot answer it.

## Tests

The three traversal cases already on `main` covered the refusal. Review added a fourth, `'../files-backup/x'` — a sibling directory whose name the media dir prefixes — because the other three are all refused by a separator-less `startsWith(mediaDir)` too, so none of them would catch such a check being inlined here again; `storagePath.test.ts` pins that boundary for the helper, nothing pinned it for this call site.

Proven by reverting, three ways:

| Revert | Result |
| --- | --- |
| Guard removed entirely | 3 failed \| 32 passed |
| Back to the hand-rolled `path.relative` check | 3 failed \| 32 passed |
| Separator-less `startsWith` that still logs | 1 failed \| 35 passed — exactly the new case |

The second matters most. Both forms return the same null and push the same warning, so a revert to an inline check is invisible in results — the tests assert the shared helper's `logger.warn`, which is the only observable difference. The file gains the logger mock the two driver tests already use, with an explicit `mockClear` in `beforeEach` because a `vi.mock` factory's `vi.fn()` accumulates calls across a whole file.

`MEDIA_ARCHIVE_DIR` is exported so the test computes the expected root exactly as `copyProfileImage` does, rather than reverse-engineering it through a throwaway filename.

## Docs and one route comment

`AGENTS.md` / `CLAUDE.md`: added this call site to the helper's importer list, recorded that it is the one migration that fixed nothing, narrowed "the backup scripts keep their own check" to `productionArchive.ts`, and noted that the helper is case-sensitive where `cleanupMediaStorage.ts` deliberately is not.

Looking for the right home for one of those facts turned up a claim that had been wrong since #1575: *"Every caller **reads** a server-generated path from a row."* `GET /api/v1/files/[...pathname]` reads neither — it hands the client's own URL segments to `getMedia` and so to the driver. That is corrected, and the route now carries a comment saying why its own absolute-path refusal is load-bearing, beside the comment that already explains its `isReservedFitnessMediaPath` guard.

## Review

13 rounds of the sub-agent loop, 25 findings, all resolved. Two consecutive clean rounds ended it. Four findings were errors in my own corrections — twice a fix was itself wrong — and the last one cut a clause rather than correcting it a fifth time: a passage needing four correction rounds was carrying a tangent, not unlucky wording.

## Gate

`prettier --write .`, `yarn lint` (0 errors), `yarn typecheck` (clean), `yarn build`, `yarn test` — 887 files, 11701 tests, all passing. Re-run in full after every round's fixes.


AGENTS.md
```diff
@@ -2313,13 +2313,13 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
...dias/quota` already are.** `lib/services/fitness-files/localFile.ts`, `lib/jobs/importStravaArchiveJob.ts`, `scripts/maintenance/cleanupMediaStorage.ts` and `scripts/backup/actorArchive.ts`'s `copyProfileImage` import it too — it replaced the near-identical `fitness-files/path.ts` **and** the cleanup script's own hand-rolled check, which compared against a bare `path.resolve(basePath)` and so accepted a sibling directory the root's name prefixes (`/srv/uploads` accepted `../uploads-backup/x`, and then `unlink`ed it). That is why the helper appends `path.sep` to the root. Note the root itself resolves to itself; only a path that leaves the root is refused. **`copyProfileImage` is the one migration that fixed nothing**, and it is worth knowing which kind a call site is: it asked the same question with `path.relative` — `'..'`, a `..` + separator prefix, or an absolute answer — which carries the separator boundary for free and so never had the cleanup script's bug. It moved because a fourth spelling of one question is a fourth place to forget, not because it was wrong; its test pins the shared helper's log for exactly that reason, since every result the two forms produce is identical.
```
llun: **The "identical results" claim is falsified on win32 by a case-variant absolute path.**

`path.win32.relative` compares case-insensitively; `resolveStorageFilePath`'s `fullPath.startsWith(storageRootPrefix)` does not. Reproduced:

```
root = C:\staging\media_attachments\files
p    = c:\STAGING\media_attachments\FILES\ab\cd.webp
path.win32.relative(root, resolve(root, p)) -> 'ab\cd.webp'   old form: ACCEPT
prefix compare                              -> refuse          new form: REFUSE
```

So the two forms are not identical on every input. The divergence is Windows-only, needs an *absolute* path differing from the root only by letter case, and fails in the safe direction (spurious refusal, never a missed traversal). It is also unreachable through both real call sites: `isTraversingStoragePath`'s `/^([/\\]|[A-Za-z]:)/` (`lib/services/medias/mediaFileUrl.ts:89`) refuses any absolute path before `getMediaPathFromFileUrl` returns. And it is inherited from `resolveStorageFilePath` itself — shipped in #1575 for the local storage drivers — not introduced here.

No code change wanted. The prose claim needs narrowing to "identical on every input reachable here", with the win32 case-sensitivity divergence named as the exception, since that is a property of the shared helper the next adopter should know. `lib/services/medias/storagePath.test.ts` has no case-sensitivity coverage either.

CLAUDE.md
```diff
@@ -51,9 +51,9 @@
...
...toragePathCallSites.test.ts` catches only the OBVIOUS spelling — raw-text assertions that neither driver names `path.resolve`/`path.join` and that nothing in `app`/`lib`/`scripts` carries the inline `startsWith(path.resolve(…))` idiom. It does NOT catch a destructured `const { resolve } = path`, `path?.resolve`, a renamed import or a helper resolving on the driver's behalf; a **reviewer** is the enforcement there until the AST rule lands in `lint/agentsRules.mjs`. That narrowness is deliberate — five review rounds each broke a cleverer version, and the test's own header records how, so read it before making that file smarter. Object storage is a different question — an S3 key has no filesystem root to escape — and so is **"is this string SHAPED like traversal"**, — `isTraversingStoragePath` in `lib/services/medias/mediaFileUrl.ts`, from #1569 — which takes no root, must run after decoding, and is what a URL-recovered path needs; don't fold the two together. `scripts/backup/productionArchive.ts` keeps its own stricter `assertRelativeFilePath`/`normalizeStoragePath` for archive-entry names, which is also not drift. See **A stored path is confined to the storage root** in `AGENTS.md`.
```
llun: **The "identical results" claim is falsified on win32 by a case-variant absolute path.**

`path.win32.relative` compares case-insensitively; `resolveStorageFilePath`'s `fullPath.startsWith(storageRootPrefix)` does not. Reproduced:

```
root = C:\staging\media_attachments\files
p    = c:\STAGING\media_attachments\FILES\ab\cd.webp
path.win32.relative(root, resolve(root, p)) -> 'ab\cd.webp'   old form: ACCEPT
prefix compare                              -> refuse          new form: REFUSE
```

So the two forms are not identical on every input. The divergence is Windows-only, needs an *absolute* path differing from the root only by letter case, and fails in the safe direction (spurious refusal, never a missed traversal). It is also unreachable through both real call sites: `isTraversingStoragePath`'s `/^([/\\]|[A-Za-z]:)/` (`lib/services/medias/mediaFileUrl.ts:89`) refuses any absolute path before `getMediaPathFromFileUrl` returns. And it is inherited from `resolveStorageFilePath` itself — shipped in #1575 for the local storage drivers — not introduced here.

No code change wanted. The prose claim needs narrowing to "identical on every input reachable here", with the win32 case-sensitivity divergence named as the exception, since that is a property of the shared helper the next adopter should know. `lib/services/medias/storagePath.test.ts` has no case-sensitivity coverage either.

scripts/backup/actorArchive.ts
```diff
@@ -639,28 +640,25 @@ export const copyProfileImage = async ({
...
-  // between two absolute paths POSIX always answers with a relative string,
-  // and only a win32 CROSS-DRIVE pair (`C:\staging` to `D:\secrets`) makes
-  // `path.relative` hand back an absolute path of its own.
-  const relativePath = path.relative(mediaDir, source)
-  if (
-    relativePath === '..' ||
-    relativePath.startsWith(`..${path.sep}`) ||
-    path.isAbsolute(relativePath)
-  ) {
+  // It asks the shared helper rather than spelling the comparison out again.
```
llun: **The consolidation rationale is now told three times.**

That middle paragraph restates, at comparable length, what this PR just added to `AGENTS.md` ("`copyProfileImage` is the one migration that fixed nothing … a fourth spelling of one question is a fourth place to forget") and what the test comment in `scripts/backup/actorArchive.test.ts` already says ("the hand-rolled `path.relative` check this replaced returns the same null and pushes the same warning … so its log is the only observable difference"). Three copies of one narrative for a one-line implementation swap, and three places to keep in sync.

Compare the sibling call sites this same helper family touches: `lib/services/medias/localFile.ts:95-96` uses two lines, and `scripts/maintenance/cleanupMediaStorage.ts:266-272` uses five that state only the locally-distinguishing fact and point outward rather than retelling the design record.

Keep the first paragraph (load-bearing: why this step re-checks at all) and the closing sentence about the local `warnings.push` staying for operator messaging (not duplicated anywhere). Drop the middle retelling in favour of a pointer.

scripts/backup/actorArchive.test.ts
```diff
@@ -295,6 +305,19 @@ describe('registerAttachmentUrl', () => {
 })
 
 describe('copyProfileImage', () => {
+  // The media directory a stored path is confined to. `MEDIA_ARCHIVE_DIR` is
+  // not exported, and `getArchiveMediaPath` prefixes it, so a one-segment name
+  // put through it and then dropped names the directory itself.
+  const archiveMediaDir = (stagingDir: string) =>
```
llun, line 670: **`archiveMediaDir` reverse-engineers a constant through a throwaway filename.**

`path.resolve(stagingDir, path.dirname(getArchiveMediaPath('name.webp')))` passes a dummy filename in purely to strip it back off. `MEDIA_ARCHIVE_DIR` is not exported today, but this module already exports plenty of internals for its tests (`buildActorJson`, `csvEscape`, `forEachLike`, `getArchiveMediaPath`, …), so exporting it would let the test read `path.resolve(stagingDir, MEDIA_ARCHIVE_DIR)` — character-for-character what the production code computes at `scripts/backup/actorArchive.ts:642` — with no dummy argument to explain. The literal `'media_attachments/files'` stays pinned independently by the existing `getArchiveMediaPath` assertion, so nothing is weakened.

scripts/backup/actorArchive.test.ts
```diff
@@ -342,7 +365,9 @@ describe('copyProfileImage', () => {
 
   // `getMediaPathFromFileUrl` refuses a `..` segment before one can reach
   // here, but this is the step that turns a stored path into a file read off
-  // the operator's machine, so it answers the question for itself.
+  // the operator's machine, so it answers the question for itself — through
+  // the shared `resolveStorageFilePath`, the same helper the local storage
+  // drivers ask.
   it.each([
```
llun, line 729: **No case exercises the separator boundary the shared helper exists for.**

The three cases (`../../../secrets/env`, `..`, an absolute path outside) are all refused by a naive `fullPath.startsWith(mediaDir)` check with no separator too. The bug class `resolveStorageFilePath` was extracted for — a sibling directory whose name the root string-prefixes — is pinned at the unit level in `lib/services/medias/storagePath.test.ts` (`'../uploads-backup/secret.webp'`) but not at this call site. A future change that re-inlines a naive prefix check here, touching `storagePath.ts` not at all, would pass all three existing cases.

Add a fourth case, `'../files-backup/x'` (a sibling of `media_attachments/files`), and prove it by temporarily replacing the helper call with a separator-less `startsWith` — the new case must fail while the other three pass.

AGENTS.md
llun: Confirmed and fixed in 62bb04fbb. I reproduced it independently before acting: `path.win32.relative('C:\staging\media_attachments\files', 'c:\STAGING\media_attachments\FILES\ab\cd.webp')` returns `'ab\cd.webp'` (old form accepts) while the prefix compare refuses. The claim now reads "on every input that can reach it the two forms return the same value and push the same warning", and the divergence is recorded in AGENTS.md as its own bullet beside the symlink residual, since it is a property of `resolveStorageFilePath` rather than of this call site — the next adopter meets it first. The bullet also notes the contrast with `cleanupMediaStorage.ts`'s `getContainedRelativePath`, which deliberately goes the other way (`realpath` plus a segment-wise case-insensitive compare) because for a tool that deletes files, concluding "unrelated" is the expensive direction to be wrong in — the opposite trade to this one. No code change: the direction is a spurious refusal rather than a missed traversal, `isTraversingStoragePath`'s `/^([/\\]|[A-Za-z]:)/` refuses every absolute path upstream so nothing can reach it, and the behaviour is inherited from #1575 rather than introduced here. That `storagePath.test.ts` has no case-sensitivity coverage is filed as a follow-up rather than fixed in this PR, since changing the helper would change the shipped drivers.

CLAUDE.md
llun: Confirmed and fixed in 62bb04fbb. I reproduced it independently before acting: `path.win32.relative('C:\staging\media_attachments\files', 'c:\STAGING\media_attachments\FILES\ab\cd.webp')` returns `'ab\cd.webp'` (old form accepts) while the prefix compare refuses. The claim now reads "on every input that can reach it the two forms return the same value and push the same warning", and the divergence is recorded in AGENTS.md as its own bullet beside the symlink residual, since it is a property of `resolveStorageFilePath` rather than of this call site — the next adopter meets it first. The bullet also notes the contrast with `cleanupMediaStorage.ts`'s `getContainedRelativePath`, which deliberately goes the other way (`realpath` plus a segment-wise case-insensitive compare) because for a tool that deletes files, concluding "unrelated" is the expensive direction to be wrong in — the opposite trade to this one. No code change: the direction is a spurious refusal rather than a missed traversal, `isTraversingStoragePath`'s `/^([/\\]|[A-Za-z]:)/` refuses every absolute path upstream so nothing can reach it, and the behaviour is inherited from #1575 rather than introduced here. That `storagePath.test.ts` has no case-sensitivity coverage is filed as a follow-up rather than fixed in this PR, since changing the helper would change the shipped drivers.

scripts/backup/actorArchive.ts
llun: Agreed, fixed in 62bb04fbb. The middle paragraph is gone; the comment now keeps the first paragraph (why this step re-checks at all, which is local to this call site) plus the sentence about the local `warnings.push` staying for operator messaging, and points at AGENTS.md for the rest.

scripts/backup/actorArchive.test.ts
llun: Agreed, fixed in 62bb04fbb. `MEDIA_ARCHIVE_DIR` is now exported and the helper reads `path.resolve(stagingDir, MEDIA_ARCHIVE_DIR)` — character-for-character what `copyProfileImage` computes. Exporting is consistent with this module's own convention (it already exports `buildActorJson`, `csvEscape`, `forEachLike`, `getArchiveMediaPath` and others for its tests), and the literal stays pinned independently by the `getArchiveMediaPath` assertions.

scripts/backup/actorArchive.test.ts
llun: Agreed, added in 62bb04fbb: a fourth case, `'../files-backup/x'`, a sibling whose name the media dir prefixes. Proved it the way the instruction asks — against a separator-less `startsWith(mediaDir)` that still calls `logger.warn` (so the boundary is the only variable), exactly this case fails and the other three pass: `Tests 1 failed | 35 passed (36)`. Against the shipped code all 36 pass.

scripts/backup/actorArchive.ts
```diff
@@ -639,28 +640,21 @@ export const copyProfileImage = async ({
...
-  // and only a win32 CROSS-DRIVE pair (`C:\staging` to `D:\secrets`) makes
-  // `path.relative` hand back an absolute path of its own.
-  const relativePath = path.relative(mediaDir, source)
-  if (
-    relativePath === '..' ||
-    relativePath.startsWith(`..${path.sep}`) ||
-    path.isAbsolute(relativePath)
-  ) {
+  // Through the shared helper: "does this resolve inside that root?" is the
+  // storage drivers' question too, and AGENTS.md records why this call site
```
llun: **The AGENTS.md pointer doesn't name the section it points to.**

Every other in-repo pointer to AGENTS.md names the bolded section so a reader can jump to it — `lib/database/sql/fitnessFile.ts` ("see AGENTS.md → Fitness Gear"), `lib/services/fitness-gears/deviceIdentity.ts` ("AGENTS.md → Server/Client Module Boundary"), `app/api/v1/admin/custom_emojis/route.ts` ("AGENTS.md, 'A stored media URL is only ours if the host says so'"), and CLAUDE.md one file over spells out "See **A stored path is confined to the storage root** in `AGENTS.md`" for this exact section. This is the only pointer in the codebase that says "AGENTS.md records why" with no section name, leaving a reader to skim ~2300 lines. The trim went the right direction on length but dropped the thing that made the sibling pointers findable. Name the section.

scripts/backup/actorArchive.test.ts
```diff
@@ -342,7 +366,9 @@ describe('copyProfileImage', () => {
 
   // `getMediaPathFromFileUrl` refuses a `..` segment before one can reach
```
llun: **The defense-in-depth rationale is stated near-verbatim in both the production comment and the test comment.**

`scripts/backup/actorArchive.ts`'s first comment paragraph ("`getMediaPathFromFileUrl` already refuses a `..` segment … this is the step that turns a stored path into a file read off the operator's machine") and this test comment are near-verbatim restatements of one sentence. Round 1's fix trimmed the code comment's SECOND paragraph but left this pair, so the "told twice" problem is smaller but still live. Suggestion was to shrink the test's header to something test-specific.

> Author's position: declining the trim. This duplication predates the PR — both comments were written in #1569 and neither was added here (this PR only appended a clause naming the shared helper). More importantly the two readers are different: someone reading the test needs to know why these cases exist at all, given that `getMediaPathFromFileUrl` already refuses a `..` upstream, and cutting the rationale to a test-specific sentence removes exactly that. Duplication between a production file and its test is not automatically waste when both readers need the same fact. Flagging rather than silently ignoring.

AGENTS.md
```diff
@@ -2313,13 +2313,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The check is also CASE-SENSITIVE where the filesystem is not, and that is the one place the `path.relative` form it replaced disagreed with it.** `path.win32.relative` compares case-insensitively, so on Windows an ABSOLUTE path differing from the root only in letter case (`c:\staging\...\FILES\ab\cd.webp` against a root of `C:\staging\...\files`) was accepted by the old spelling and is refused by this one. Nothing can reach it — `isTraversingStoragePath` refuses every absolute path upstream, and each driver's root comes from config rather than from a row — and the direction is the safe one, a spurious refusal rather than a missed traversal. It is worth knowing anyway, because it is the shape a future adopter would hit first, `storagePath.test.ts` does not cover it, and `scripts/maintenance/cleanupMediaStorage.ts` deliberately went the other way in `getContainedRelativePath` (`realpath` plus a segment-wise case-insensitive compare) — for a tool that deletes files, concluding "unrelated" is the expensive direction to be wrong in, which is the opposite trade to this one.
```
llun: **The new win32 bullet is heavier than what it documents justifies.**

Six sentences with literal Windows path examples, documenting a divergence the bullet itself admits three times over is not operationally live: unreachable (`isTraversingStoragePath` refuses every absolute path upstream), safe in direction (spurious refusal, not a missed traversal), and untested. AGENTS.md is already very dense and adding to it has a cost. The durable value is one fact — that `cleanupMediaStorage.ts`'s `getContainedRelativePath` deliberately does the OPPOSITE (realpath plus a segment-wise case-insensitive compare) for a different reason, so a future reader does not try to "reconcile" the two. The reachability analysis and the literal path spellings are investigation notes, not something the next adopter needs pre-chewed. Cut to roughly a third.

scripts/backup/actorArchive.test.ts
```diff
@@ -355,6 +381,16 @@ describe('copyProfileImage', () => {
     {
       description: 'is absolute and outside',
       buildStoragePath: (secretPath: string) => secretPath
+    },
+    {
+      // The bug the shared helper was extracted for: a root of `/srv/uploads`
```
llun: **Six comment lines for one fixture, and the fifth telling of the same anecdote.**

Its three siblings carry none. Some asymmetry is earned — this case is the only one of the four that would catch a regression to a separator-less prefix compare, and that is worth stating. But it re-tells the "`/srv/uploads` accepted `../uploads-backup/x`" anecdote already told in `AGENTS.md`, `CLAUDE.md`, `scripts/maintenance/cleanupMediaStorage.ts` and `lib/services/medias/storagePathCallSites.test.ts` — a fifth telling. Cut the anecdote, keep the test-specific point, land around three lines.

scripts/backup/actorArchive.ts
llun: Agreed, fixed in d1705c752. The pointer now reads `AGENTS.md ("A stored path is confined to the storage root")`, matching how the rest of the repo cites it.

scripts/backup/actorArchive.test.ts
llun: Declining this one, and leaving the thread open so it stays visible rather than being closed by the author who disagreed with it. Reasoning, for the record: the duplication predates this branch — both comments were written in #1569, and this PR only appended a clause to the test one naming the shared helper. And the two readers are not the same reader. Someone opening the test needs to know why these cases exist at all, given that `getMediaPathFromFileUrl` already refuses a `..` before one can reach `copyProfileImage`; the suggested test-specific replacement removes exactly that. Duplication between a production file and its test is not automatically waste when both readers need the same fact. Happy to be overruled by a human reviewer.

AGENTS.md
llun: Agreed, cut in d1705c752 — from six sentences to three. It now leads with the durable fact (`cleanupMediaStorage.ts`'s `getContainedRelativePath` goes the other way on purpose; do not reconcile the two), keeps the inverted trade and the unreachability in one clause each, and drops the literal Windows path spellings, which were investigation notes. CLAUDE.md's condensed clause was updated to match so the two files still mirror each other.

scripts/backup/actorArchive.test.ts
llun: Agreed, trimmed to four lines in d1705c752. The `/srv/uploads` / `../uploads-backup/x` anecdote is gone — AGENTS.md, CLAUDE.md, `cleanupMediaStorage.ts` and `storagePathCallSites.test.ts` already carry it — leaving only the test-specific point: this is the one case of the four a separator-less prefix compare would let through.

AGENTS.md
```diff
@@ -2313,13 +2313,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The check is CASE-SENSITIVE, and `cleanupMediaStorage.ts` inverts that on macOS and Windows ONLY — do not reconcile the two.** Its `getContainedRelativePath` resolves symlinks and compares segments case-insensitively where `process.platform` says the filesystem folds case, falling through to the same case-sensitive answer this helper gives on Linux; for a tool that deletes files, concluding "unrelated" is the expensive direction to be wrong in, where here over-refusing costs nothing. Nothing reaches the difference today, for two reasons worth keeping apart: the archive's call sits downstream of `isTraversingStoragePath`, which refuses every absolute path, while the DRIVER call sites have no such guard and are safe only because nothing writes an absolute path into a `medias` / `fitness_files` row. It is also the one input where the `path.relative` spelling `copyProfileImage` used disagreed — `path.win32.relative` being case-insensitive — and `storagePath.test.ts` does not cover it.
```
llun: "`cleanupMediaStorage.ts` goes the other way" was true on macOS and Windows and false on Linux.

`getContainedRelativePath` (`scripts/maintenance/cleanupMediaStorage.ts`) does not compare case-insensitively unconditionally. It gates that on `isCaseInsensitiveFilesystem()` — `process.platform === 'darwin' || process.platform === 'win32'` — and otherwise returns the plain `path.relative` result:

```js
const direct = path.relative(realFrom, realTo)
if (!direct.startsWith('..') && !path.isAbsolute(direct)) return direct
if (!isCaseInsensitiveFilesystem()) return direct   // <- Linux stops here
```

So on Linux — the deployment target — it behaves exactly like `resolveStorageFilePath`, and the divergence the bullet framed as permanent ("do not reconcile the two") exists only on macOS and Windows. A future editor could have relied on the stronger claim.

llun: The reachability argument conflated the archive's call site with the driver ones.

"Nothing can reach it — `isTraversingStoragePath` refuses every absolute path upstream, and a driver's root comes from config rather than from a row" put those two clauses in the same breath, but they cover different call sites and only one of them is a check.

`isTraversingStoragePath` sits upstream of the ARCHIVE's call (`scripts/backup/actorArchive.ts`, via `getMediaPathFromFileUrl`) and nothing else. The other four call sites of `resolveStorageFilePath`/`assertStorageFilePath` — `lib/services/medias/localFile.ts`, `lib/services/fitness-files/localFile.ts`, `lib/jobs/importStravaArchiveJob.ts`, `scripts/maintenance/cleanupMediaStorage.ts` — take their `filePath` from a database row, a generated name or a directory walk with no such guard in between. They are safe for a different and uncodified reason: nothing in the app writes an absolute path into a `medias` / `fitness_files` row. The root coming from config says nothing about the path.

llun: Fixed in 536c1ea9d, in the same round it was found. The bullet now says `cleanupMediaStorage.ts` inverts the behaviour "on macOS and Windows ONLY", names the `process.platform` gate, and states that it falls through to this helper's own case-sensitive answer on Linux. CLAUDE.md's mirrored clause carries the same qualifier.

llun: Fixed in 536c1ea9d. The bullet now separates the two reasons explicitly — the archive's call sits downstream of `isTraversingStoragePath`, while the driver call sites have no such guard and are safe only because nothing writes an absolute path into a row. That is worth stating precisely because it is an assumption rather than a check, and it is now written down as one.

scripts/backup/actorArchive.test.ts
llun: Reversing the decline — the counter-argument is right and mine was not. The two comments shared "turns a stored path into a file read off the operator's machine" near-verbatim, which is the tell that this was a copy rather than a reframing for a different reader; and declining on "it predates the branch" is inconsistent with this same branch having cut a comparable pre-existing duplication (the `/srv/uploads` anecdote) one block down. Fixed in 536c1ea9d: the test comment now reads "As `copyProfileImage`'s own comment explains, `getMediaPathFromFileUrl` refuses a `..` segment before one can reach it. These cases exercise the local re-check it makes anyway, through the shared `resolveStorageFilePath`." That answers the test reader's "why do these exist" question without reproducing the production sentence.

AGENTS.md
```diff
@@ -2313,13 +2313,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
...e` IS `path.win32.relative` and already answers case-insensitively before that loop is reached, while on Linux it returns the same case-sensitive answer this helper gives. It goes that way because for a tool that DELETES files, concluding "unrelated" is the expensive direction to be wrong in; here the trade is inverted and over-refusing costs nothing. Nothing reaches the difference today, but through THREE separate mechanisms and it is worth knowing which one covers a given caller: `isTraversingStoragePath` guards the archive's call; `GET /api/v1/files/[...pathname]` — **the one driver caller fed unvalidated client input**, straight from the URL segments into `getMedia` — refuses an absolute or drive-lettered path in the ROUTE, not anywhere near the driver; and the remaining driver calls are handed a path from a row or a directory walk that is never absolute by construction. Only the first is a check this module can see, which is why weakening the route's own guard on the belief that "a row can never hold an absolute path" would reopen the gap. It is also the one input where the `path.relative` spelling `copyProfileImage` used disagreed, and `storagePath.test.ts` does not cover it.
```
llun: **A driver call site IS fed unvalidated client input, so "safe only because nothing writes an absolute path into a row" is false for it.**

The bullet enumerated two reasons nothing can reach the case-sensitivity difference, and assigned every driver call site to "nothing writes an absolute path into a `medias` / `fitness_files` row". That does not hold for the most reachable caller of all.

`GET /api/v1/files/[...pathname]` (`app/api/v1/files/[...pathname]/route.ts`) joins the client's own URL segments, normalises them, and passes the result to `getMedia` → `LocalFileStorage.getFile` → `resolveStorageFilePath`. That is unvalidated client input reaching a storage driver. What keeps an absolute path out of it is the route's OWN guard:

```js
if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
  return apiErrorResponse(404)
}
```

which has nothing to do with row content. So there are three mechanisms, not two, and the stated one does not cover this caller. The practical risk of the wrong version: an editor who trusts "a row can never hold an absolute path" could weaken or drop that route-level check believing it redundant.

llun: **The segment-wise case-insensitive compare is the macOS branch; Windows never reaches it.**

The bullet credited `getContainedRelativePath`'s segment-by-segment case-insensitive compare with covering "macOS and Windows". On win32 the generic `path` module IS `path.win32`, and `path.win32.relative` is itself case-insensitive:

```
path.win32.relative('C:\data\media', 'C:\DATA\MEDIA\file.png') -> 'file.png'
path.posix.relative('/data/media', '/data/MEDIA/file.png')     -> '../MEDIA/file.png'
```

So on Windows `direct` comes back clean and the function returns at its early exit (`if (!direct.startsWith('..') && !path.isAbsolute(direct)) return direct`) before the `isCaseInsensitiveFilesystem()` gate or the segment loop is reached. Only macOS, where the POSIX `path.relative` is case-sensitive, actually falls into that loop. The conclusion (Windows ends up case-insensitive) is right; the mechanism credited is not.

llun: Confirmed and fixed in d4286c00d. Verified the chain by reading the route: client segments → `path.normalize` → the absolute/drive-letter refusal → `getMedia` → `LocalFileStorage.getFile` → `resolveStorageFilePath`. The bullet now names three mechanisms and which caller each covers, calls out `GET /api/v1/files/[...pathname]` as **the one driver caller fed unvalidated client input**, says its refusal lives in the ROUTE rather than near the driver, and states the consequence — that weakening that guard on the belief a row cannot hold an absolute path is what would reopen the gap. Only the first mechanism is a check this module can see, and the text says that too. CLAUDE.md's mirror carries the same three.

llun: Confirmed and fixed in d4286c00d — I re-ran both `path.relative` flavours to check rather than taking it on trust, and they answer exactly as quoted. The bullet now says the loop is "in practice the macOS branch, because on win32 `path.relative` IS `path.win32.relative` and already answers case-insensitively before that loop is reached", and the heading clause reads "inverts that off Linux" rather than naming the two platforms as if they shared a mechanism.

AGENTS.md
```diff
@@ -2313,13 +2313,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The check is CASE-SENSITIVE, and `cleanupMediaStorage.ts` inverts that — do not reconcile the two.** Its `getContainedRelativePath` resolves symlinks and, where `process.platform` says the filesystem folds case, compares segments case-insensitively, because for a tool that DELETES files, concluding "unrelated" is the expensive direction to be wrong in; here the trade is inverted and over-refusing costs nothing. No live caller reaches the difference — but resist writing down WHY as a closed list, and resist saying which platform takes which branch. Both are claims about a moving set, and this bullet had each falsified in review: the caller enumeration twice, and "the segment compare is the macOS branch" once, since win32 skips that loop only for a NESTED pair, and reaches it for two unrelated roots exactly as macOS does. It is the one input where the `path.relative` spelling `copyProfileImage` used disagreed, and `storagePath.test.ts` does not cover it.
```
llun: The caller enumeration was falsified a second time — `deleteSavedMedia` is covered by mechanism 1, not by "a row or a directory walk".

"The remaining driver calls are handed a path from a row or a directory walk that is never absolute by construction" does not cover `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`). It recovers its path with `getMediaPathFromFileUrl`, which applies `isTraversingStoragePath` — the mechanism the text assigned to the archive's call alone — and then calls `deleteMediaFile` → `LocalFileStorage.deleteFile` → `resolveStorageFilePath`. That is a driver call protected by mechanism 1, so the three-way partition is wrong. Not a live bug: the call site is protected. The stated reason for it is what is wrong.

This is the second enumeration in this bullet to be falsified — the previous one claimed two mechanisms and missed the file-serving route.

llun: "In practice the macOS branch" is false for unrelated roots — win32 reaches the segment loop too.

`path.win32.relative` returns a contained answer, and so takes `getContainedRelativePath`'s early exit, only when the target is NESTED inside the root. For two unrelated roots — a fitness root sitting beside the media root rather than under it, an ordinary deployment shape — it returns a `..`-prefixed path, which falls past the early exit AND past the `isCaseInsensitiveFilesystem()` gate (true on win32), reaching the segment-comparison loop exactly as macOS does. Verified against Node's `path.win32`:

```
relative('C:\data\media', 'C:\data\fitness')       -> '..\fitness'   loop entered
relative('C:\data\media', 'C:\data\media\fitness') -> 'fitness'      early exit
relative('C:\data\media', 'C:\DATA\MEDIA\fitness') -> 'fitness'      early exit
```

So the loop is macOS-only for the narrow nested case-variant scenario the bullet illustrates, not in general.

CLAUDE.md
```diff
@@ -51,9 +51,9 @@
...
```
llun: The CLAUDE.md mirror re-pasted the detail instead of compressing it.

`CLAUDE.md` is the condensed per-machine index of the `AGENTS.md` rules, and round 4 grew this clause by pasting the corrected reachability text in nearly unchanged rather than compressing it — a maintenance script's platform-specific branch behaviour is below the altitude this file works at.

Author's note appended for the record: the suggestion to rewrite the bullet as a one-line index entry ("see X in AGENTS.md") was NOT taken, because every one of this file's ~40 bullets is a dense multi-sentence paragraph ending in such a pointer — that is the file's established convention, and restyling one bullet alone would make it the odd one out while restyling all of them is far outside this PR. The substance of the finding was taken: the clause shrank in step with the AGENTS.md cut.

AGENTS.md
llun: Fixed in 569b970fe, and the repeat is what drove the fix rather than the individual error. The enumeration is gone entirely. The bullet now states only the durable rule — the check is case-sensitive, `cleanupMediaStorage.ts` inverts it on a case-folding filesystem, do not reconcile the two — and instructs the next editor not to re-derive the list, giving the falsification count as the reason. That matches how the `storagePathCallSites.test.ts` bullet records what broke its own cleverer versions. A closed claim about a set that unrelated PRs keep moving is the wrong shape for this document, which is the actual lesson here.

llun: Confirmed against `path.win32` before acting — the three cases answer exactly as quoted. Fixed in 569b970fe by removing the platform-branch claim rather than re-qualifying it, for the same reason as the enumeration: it was an over-specified mechanism detail that had to be right about a moving target, and it was the third such claim in this bullet to need correcting. The bullet now says only that the cleanup script compares case-insensitively "where `process.platform` says the filesystem folds case", and explicitly tells the next editor not to say which platform takes which branch, recording that this one was falsified.

CLAUDE.md
llun: Addressed in 569b970fe. The clause lost the three-mechanism enumeration and the platform-branch claim along with AGENTS.md, and now carries the same durable rule plus the same warning not to re-derive either list. Its length is back in line with its neighbours in that file. The form of the suggestion — a bare one-line pointer — is declined for the consistency reason recorded above.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- **The check is CASE-SENSITIVE, and `cleanupMediaStorage.ts` inverts that — do not reconcile the two.** Its `getContainedRelativePath` resolves symlinks and, where `process.platform` says the filesystem folds case, compares segments case-insensitively, because for a tool that DELETES files, concluding "unrelated" is the expensive direction to be wrong in; here the trade is inverted and over-refusing costs nothing. No live caller reaches the difference — but resist writing down WHY as a closed list, and resist saying which platform takes which branch. Both are claims about a moving set, and review falsified each of them here repeatedly — including, twice, a correction that was itself wrong. It is the one input where the `path.relative` spelling `copyProfileImage` used disagreed, and `storagePath.test.ts` does not cover it.
```
llun: The sentence forbidding platform-branch claims then made one.

> The bullet read: "resist saying which platform takes which branch. Both are claims about a moving set, and this bullet had each falsified in review: the caller enumeration twice, and 'the segment compare is the macOS branch' once, **since win32 skips that loop only for a NESTED pair, and reaches it for two unrelated roots exactly as macOS does.**"
>
> The bolded clause is a which-platform-takes-which-branch claim — the exact category the sentence just told the next editor to avoid, relabelled as "what review found" rather than "the design". It is currently accurate, but that is the property that has already been wrong twice here; restating it keeps the same fragility alive under a new name. `CLAUDE.md`'s version of the same sentence was more disciplined: it stopped at the falsification and did not re-derive the mechanism.

app/api/v1/files/[...pathname]/route.ts
```diff
@@ -20,7 +20,16 @@ export const GET = traceApiRoute(
...
+    // driver: the segments below go to `getMedia`, which hands them to
+    // `LocalFileStorage.getFile`. Every other caller is given a path from a
+    // row, a directory walk, a server-generated name, or the traversal-checked
+    // `getMediaPathFromFileUrl`. So this refusal is the check for that input,
+    // and the driver's own `resolveStorageFilePath` does not make it
+    // redundant — an absolute path would land outside the storage root, which
+    // that helper answers by refusing a read this route should never have
+    // asked for. POSIX hosts do not treat Windows drive paths as absolute,
+    // hence the second test.
     if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
```
Comment (llun) [inline app/api/v1/files/[...pathname]/route.ts:34]: The one route that hands client input to a storage driver did not say so.

> This guard is the only thing standing between a client-supplied path and `LocalFileStorage.getFile` → `resolveStorageFilePath`: the segments below go to `getMedia`, which passes them straight to the driver. Every other driver caller is handed a path from a row, a directory walk, a server-generated name, or the traversal-checked `getMediaPathFromFileUrl`.
>
> The inline comment explained only the drive-letter half. Nothing said the refusal is load-bearing, or that an editor should not drop it believing `resolveStorageFilePath`'s own containment check makes it redundant. The same file carries exactly that style of "why this exists and what breaks without it" comment for the `isReservedFitnessMediaPath` guard four lines down, so there was a precedent sitting right there.

AGENTS.md
llun: The falsification tally was itself miscounted.

> The bullet said the platform claim was falsified "once". It was falsified at least twice in this PR's own review — round 4 ("the segment-wise compare is the macOS branch; Windows never reaches it", correcting "on macOS and Windows ONLY") and round 5 ("'in practice the macOS branch' is false for unrelated roots"), with a third earlier correction to the unconditional-inversion claim in round 3. The fixer's own reply to the second called it "the third such claim in this bullet to need correcting", so the tally contradicted the PR's own record as well as the sibling count beside it.

llun: Agreed and fixed in 627d4a1e2 — a clean catch. The clause is gone; the sentence now stops at "review falsified each of them here repeatedly". The `storagePathCallSites.test.ts` bullet is the model, as you noted: it records that five review rounds broke cleverer versions and defers the how to the test's own header rather than inlining it.

app/api/v1/files/[...pathname]/route.ts
Comment (llun) [reply app/api/v1/files/[...pathname]/route.ts:34]: Fixed in 627d4a1e2, and thank you for pointing at the neighbouring comment — that settled where this belonged. An earlier round had removed this fact from AGENTS.md's storage-root section as the wrong home, which was right, but removing it left it documented nowhere. It now sits beside the guard it describes.

Looking for the right home also turned up a better find: AGENTS.md's own storage-path bullet has claimed since #1575 that "Every caller READS a server-generated path from a row … with one caller that DERIVES rather than reads". This route reads neither. That claim is corrected in the same commit, in both AGENTS.md and CLAUDE.md — so the fact is not new documentation bolted onto the wrong section, it is a false claim that section was already making.

AGENTS.md
llun: Confirmed and fixed in 627d4a1e2 — by removing the numbers rather than correcting them. A tally of how often this bullet was wrong is the same brittle shape as the claims it warns about, and needing a review round to audit it made that concrete. It now reads "review falsified each of them here repeatedly — including, twice, a correction that was itself wrong", which is the durable part and cannot be miscounted.

app/api/v1/files/[...pathname]/route.ts
```diff
@@ -20,7 +20,17 @@ export const GET = traceApiRoute(
...
+    // callers pass the driver a path they did not take from a request — see
+    // "A stored path is confined to the storage root" in AGENTS.md, which
+    // keeps that inventory — so this refusal is the check for THIS input. The
+    // driver's own `resolveStorageFilePath` does not make it redundant: an
+    // absolute path resolves outside the storage root, which that helper
+    // answers with a null, and the route below turns a null into the
+    // placeholder image — a miss served for a read it should never have made.
+    // POSIX hosts do not treat Windows drive paths as absolute, hence the
+    // second test.
     if (path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/.test(normalizedPath)) {
```
Comment (llun) [inline app/api/v1/files/[...pathname]/route.ts:34]: **An enumeration is worse in a code comment than in AGENTS.md, not better.**

The comment added last round claimed "Every other caller is given a path from a row, a directory walk, a server-generated name, or the traversal-checked `getMediaPathFromFileUrl`". Three enumerations in this PR's docs were falsified by review and eventually deleted for being closed claims about a moving set, and this is the same shape — it describes four other files' behaviour from inside a route file.

The difference is not that a comment sits next to the code it describes: the first and last sentences do, but the enumeration does not. It is that a doc bullet is what the review loop keeps re-reading, which is how the three were caught, while nobody adding a fifth storage-driver caller has any reason to open this route.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
.../api/v1/files/[...pathname]` is the one that matters: it hands the CLIENT's own URL segments to `getMedia` and so to the driver. Keep its own `path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/` refusal: it is what makes that a 404 at the edge, where the check behind it — per the bullet above — answers a null `LocalFileStorage` turns into a served placeholder, and under S3 is not there at all. The other **derives** rather than reads: `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`) slices the tail off a `/api/v1/files/` URL and `decodeURIComponent`s it, which would turn `%2e%2e%2f` into traversal the moment its input stopped being `saveMedia`'s own output. The asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun: **"In the route rather than anywhere near the driver" is false for LocalFileStorage, and contradicts a bullet two above it.**

The corrected caller sentence said the route's own refusal is "what keeps an absolute path out … in the route rather than anywhere near the driver. Do not drop that guard believing the driver's containment check covers it."

For LocalFileStorage the driver's check does cover it. `resolveStorageFilePath` is `getFile`'s first statement; `path.resolve('/data/media', '/etc/passwd')` is `/etc/passwd`, which fails the storage-root prefix test, so `getFile` returns null, `getMedia` returns null, and the route serves `media-removed.svg`. Dropping the route guard changes a 404 into a placeholder — not a containment bypass.

The claim holds only for S3, whose `getFile` makes no such check. AGENTS.md already says exactly that two bullets earlier: "`LocalFileStorage.getFile` already makes the identical check against its storage root, while `S3FileStorage.getFile` makes **none**". So the new sentence contradicted the file's own single-sourced answer. Note the route's own comment did NOT make this mistake — it says "does not make it redundant" and names `LocalFileStorage`.

app/api/v1/files/[...pathname]/route.ts
Comment (llun) [reply app/api/v1/files/[...pathname]/route.ts:34]: Agreed, and this was the question I most wanted answered — fixed in 3f2da9c17. The list is gone. The comment keeps what is local to this route (a client's path reaches a driver here; `resolveStorageFilePath` does not make the refusal redundant) and points at the AGENTS.md section that holds the inventory and gets re-checked. It also states the consequence more precisely: without the guard, the helper's null becomes the route's placeholder image.

AGENTS.md
llun: Confirmed and fixed in d24c62e96 — I traced it before acting and it lands exactly as described. Both files now give the reason to keep the guard without the false part: it is the 404 at the edge, behind it `LocalFileStorage` answers a null the route serves as a placeholder, and `S3FileStorage` checks nothing. That defers to the bullet above rather than restating it, which is what the file wanted. The route comment was already correct and is untouched.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
...he one that matters: it hands the CLIENT's own URL segments to `getMedia` and so to the driver. Keep its own `path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/` refusal: it is what makes that a 404 at the edge, where the check behind it — per the bullet above — answers a null that `LocalFileStorage` returns as-is and the ROUTE turns into a served placeholder, and that S3 does not make at all. The other **derives** rather than reads: `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`) slices the tail off a `/api/v1/files/` URL and `decodeURIComponent`s it, which would turn `%2e%2e%2f` into traversal the moment its input stopped being `saveMedia`'s own output. The asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun: The sentence read "...answers a null `LocalFileStorage` turns into a served placeholder". Parsed normally that makes `LocalFileStorage` the thing serving the placeholder. It is not: `LocalFileStorage.getFile` is `if (!fullPath) return null` and knows nothing about placeholders — the `media-removed.svg` branch lives in the route's own `if (!media)`. The sentence was also missing the connector ("a null **that** `LocalFileStorage`…") that would let it parse at all.

`CLAUDE.md`, edited in the same commit for the same fact, says it correctly: "`LocalFileStorage` answers a null the route serves as a placeholder". So the two files disagreed, and the commit message matched the CLAUDE.md wording rather than the one that shipped in AGENTS.md.

Found independently by two lenses this round, which was the whole of what they found.

llun: Fixed in 1f29373bf. AGENTS.md now names the ROUTE as what turns the null into a served placeholder and carries the missing connector, so the two files agree and both match the code.

The same commit also drops the clause narrating how often this bullet's claims had been falsified in review. It read as a war story where a rule belongs, and it had already needed correcting once for miscounting — which is its own argument for not keeping a tally. Both files now say the durable thing instead: these are claims about a moving set, state the rule, not the inventory.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
...eep its own `path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/` refusal: it is what makes that a 404 at the edge. Behind it, per the bullet above, `resolveStorageFilePath` answers null on an out-of-root path and `LocalFileStorage.getFile` returns that null unchanged, leaving the ROUTE to turn it into the served placeholder — a null `S3FileStorage` never produces, since it makes no such check. The other **derives** rather than reads: `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`) slices the tail off a `/api/v1/files/` URL and `decodeURIComponent`s it, which would turn `%2e%2e%2f` into traversal the moment its input stopped being `saveMedia`'s own output. The asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun: **Title: The corrected sentence was accurate but did not parse on first read.**

> "...it is what makes that a 404 at the edge, where the check behind it — per the bullet above — answers a null that `LocalFileStorage` returns as-is and the ROUTE turns into a served placeholder, and that S3 does not make at all."

Three parse costs stack in one clause: (a) "the ROUTE turns into a served placeholder" elides its object right after a different subject was named, so the reader backtracks to supply the pronoun; (b) the trailing "and that S3 does not make at all" reaches back past two intervening clauses for its antecedent, "a null"; (c) the em-dash aside sits between "the check behind it" and its verb, adding a second parse cost where the sentence is already overloaded.

The route's own comment says the same thing far more legibly, and is worth borrowing from.

llun: Agreed, fixed in ab8b0cc32 — this is an editorial finding and it was right. Two sentences now, each actor named where it acts: "Keep its own `path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/` refusal: it is what makes that a 404 at the edge. Behind it, per the bullet above, `resolveStorageFilePath` answers null on an out-of-root path and `LocalFileStorage.getFile` returns that null unchanged, leaving the ROUTE to turn it into the served placeholder — a null `S3FileStorage` never produces, since it makes no such check." The wording comes from the route comment, as suggested.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
... it is what makes that a 404 at the edge. Behind it, per the bullet above, `resolveStorageFilePath` answers null on an out-of-root path and `LocalFileStorage.getFile` returns that null unchanged, leaving the ROUTE to turn it into the served placeholder. `S3FileStorage` makes no such check at all, so under object storage nothing refuses the path before it becomes a key or a CDN redirect URL. The other **derives** rather than reads: `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`) slices the tail off a `/api/v1/files/` URL and `decodeURIComponent`s it, which would turn `%2e%2e%2f` into traversal the moment its input stopped being `saveMedia`'s own output. The asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun: **The readability split turned a true narrow claim into a false broad one — `S3FileStorage` does return null.**

The previous wording said S3 "does not make [the check] at all" — scoped to the containment check, and true. The split rewrote it as "a null `S3FileStorage` never produces, since it makes no such check", which is a different and stronger claim about nulls in general.

It is false. `S3FileStorage.getFile` has `if (!object.Body) return null` (`lib/services/medias/S3StorageFile.ts`) for a missing object body — a real branch, unrelated to path containment — and that null reaches the route's placeholder branch exactly like the local driver's containment null. A reader taking the sentence at face value would conclude a null at the route implies the local driver's refusal.

The absence of a containment check does not imply the absence of nulls. Rewriting for style is how a claim's scope slips.

Found independently by two lenses this round.

llun: Confirmed and fixed in c66e93314. I caught this one myself while the round was still running — I had flagged the risk when handing the rewrite to review, checked `S3StorageFile.ts`, found the `if (!object.Body) return null` branch, and fixed it before the reports came back; both lenses then found the same thing and corroborated the fix sitting in the worktree. Recording it here because it was a real error in a pushed commit, not because review missed it.

Both files now state the absence of the check without the claim about nulls: `S3FileStorage` makes no containment check at all, so nothing refuses the path before it becomes a key or a CDN redirect URL.

Worth noting this is the second time in this loop that a correction was itself wrong — which is exactly the argument for the rule already in that bullet: state the rule, not the inventory.

AGENTS.md
```diff
@@ -2304,22 +2304,23 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
...ds the CLIENT's own URL segments to `getMedia` and so to the driver. Keep its own `path.isAbsolute(normalizedPath) || /^[a-zA-Z]:/` refusal: it is what makes that a 404 at the edge. Behind it, per the bullet above, `resolveStorageFilePath` answers null on an out-of-root path and `LocalFileStorage.getFile` returns that null unchanged, leaving the ROUTE to turn it into the served placeholder. The other **derives** rather than reads: `deleteSavedMedia` (`app/api/v1/admin/custom_emojis/route.ts`) slices the tail off a `/api/v1/files/` URL and `decodeURIComponent`s it, which would turn `%2e%2e%2f` into traversal the moment its input stopped being `saveMedia`'s own output. The asymmetry is the problem: containment was an invariant of the callers, not of the driver, so it held only until someone added a caller, and it made "the storage drivers refuse such a path" false as a claim about the pair (a review of PR #1570 had to narrow a comment in `mediaFileUrl.ts` for exactly that reason). An unlink is the more damaging half. The write paths generate their own names today, so the assert there is the belt rather than the braces, and is applied anyway so confinement is a property of the driver.
```
llun, line 2330: **The S3 clause was a tangent, and the churn was the evidence.**

The corrected sentence is accurate this time. The passage is still the problem: four consecutive rounds touched this one clause and two of those corrections were themselves wrong.

The cause is not unlucky wording. This bullet's job is that the media route's own absolute-path refusal is what matters, with `LocalFileStorage`'s null-to-placeholder chain as the motivating detail. What S3 does instead does not serve that point — and AGENTS.md already states it twice: in the media-URL section ("`S3FileStorage.getFile` makes **none** — an S3 key that traverses is merely inert, not refused, and with a CDN `hostname` configured it is string-concatenated into a redirect URL"), and in this very section's dedicated bullet ("**The object-storage drivers are a different question.**"). Counting the CLAUDE.md digests, the fact was stated four times.

So the clause cost four rounds to get right for zero information. Cut it and let the bullet end at "...the served placeholder"; a fact already carried by a dedicated bullet a few lines below does not need restating in a paragraph that is already threading eight claims about four components.

llun: Cut in c8c037eb4, from both files, as recommended — this was the finding worth having, and it is a better answer than a fifth correction. The bullet now ends where its point does.

Recording the pattern for whoever reads this thread later: the individual errors were not the signal. A passage needing four correction rounds for one sub-clause was carrying more chained causality than a paragraph holds reliably, and the sub-clause turned out to be the part that was not doing any work. Looking for what to delete beat looking for better wording.

PR #1584, opened 2026-08-26
fix: keep storage-driver error text out of the archive manifest

llun:
## Problem

`redactStorageSource` in `scripts/backup/productionArchive.ts` deliberately strips the bucket, region, endpoint, CDN hostname and local root out of the archive's `manifest.json` — and `storage[].failedFiles[].error`, written a few lines above it into the same manifest, carried the storage driver's message **verbatim**.

An export run with `--allow-missing-storage` against a private-network MinIO therefore shipped strings like `getaddrinfo ENOTFOUND minio.internal` inside the tarball, defeating the redaction sitting immediately next to it. The local driver shipped `ENOENT: no such file or directory, copyfile '/srv/uploads/...'`; an S3 service error can carry the bucket name, and a refused connection carries a bare internal IP and port.

`archiveStorage` is **shared with `scripts/backup/actorArchive.ts`**, whose `manifest.json` also embeds `storage: storageManifest` — so the same leak reached the per-actor export handed to the user whose data it is, not just the operator's own backup.

Pre-existing on `main`. Surfaced by the security review on [#1572](https://github.com/llun/activities.next/pull/1572) and deliberately left out of that PR's scope.

## Fix

`redactStorageError(error, source)` **extends the redaction already sitting beside it** rather than adding a second policy — both halves answer to one definition of what is sensitive about a `StorageSource`, derived from the fields `redactStorageSource` keeps (`kind`, `prefix`), so a field added to that type later is withheld from both without anyone remembering to.

It is an **allowlist, not a scrubber**: the syscall, code, name and HTTP status survive, the message never does. A message is free text a driver interpolates whatever it likes into, so a denylist over one is a guess about the forms we thought of; an allowlist is a guarantee. Two filters back it:

- a **shape** check, `/^[A-Za-z][A-Za-z0-9_]{0,63}$/` — error codes, error names and libuv syscalls are single bare words (`ENOTFOUND`, `NoSuchKey`, `getaddrinfo`), so anything carrying a dot, dash, colon, slash or space is refused. That excludes `minio.internal`, `10.4.2.11`, `https://minio.internal:9000` and `activitynext-prod-media` without naming any of them.
- a **value** check against everything `redactStorageSource` drops, for the single-label case the shape check cannot tell from an error code: a Docker-network `hostname: 'minio'`, a bucket named `activitynext`, a `region: 'auto'`.

The chain is walked through `cause`, because `fetch` reports a DNS failure as `TypeError: fetch failed` and hangs the real error underneath. Results: `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503`, and `unknown` for a failure that classifies itself only in its message.

Three further properties came out of the review rounds:

- **On the S3 path all three of `code`, `name` and `syscall` are server-controlled**, not just `name`. The SDK's error handler flattens every member of the XML `<Error>` element onto the data object and `decorateServiceException` copies each onto the exception — a server-chosen `code`+`syscall` produced `pool_ssd_openstack storage_node_7_rack_b12 HTTP 404`. Nothing leaks (all three stay inside the shape check), but none of them may be treated as a libuv constant.
- **The redactor cannot throw.** It runs inside the `catch` that makes `--allow-missing-storage` tolerant, so a hostile property getter would otherwise turn one tolerated missing file into a failed export. Its whole body sits in one `try` falling back to `unknown`.
- **The HTTP-status candidates were narrowed** to `$metadata.httpStatusCode` (the AWS SDK) and `httpStatusCode` (attached by `downloadPublicStorageFile`). A bare `status` was dropped: it is not an HTTP field by convention, so a `child_process`-style error would have been reported as `HTTP 127`.

### What is deliberately not redacted

The `console.error` beside it still prints the driver's message **in full**. It goes to the operator running the export, on their own machine, and never enters the archive — which is what makes `unknown` in the manifest cost no diagnosis. Don't "finish the job" by redacting it too.

The claim covers that one field. `actorArchive.ts`'s `warnings` array still carries a raw message for a failed **remote** attachment fetch — a third-party host whose URL the warning already names on purpose — and that is not the same disclosure.

The output is a bound on **form, not content**: a storage server writing its own error responses can put 64 arbitrary `[A-Za-z][A-Za-z0-9_]*` characters into each token, and a vendor code such as `XMinioStorageFull` names the storage software. Both are the price of a code an operator can act on.

## Tests

`scripts/backup/productionArchive.test.ts`:

- a table over `redactStorageError` — DNS failure, S3 service error, HTTP status alone, a fetch failure behind its `cause`, a wrapper with no coded cause, an unclassified error, a thrown non-error, a code shaped like a host, a code shaped like an address, a name echoing the bucket, a syscall echoing the CDN host, an out-of-range status — plus a cause cycle, a chain resolving at the depth cap, one truncating past it, and a hostile getter that throws.
- five `archiveStorage` tests that assert the whole manifest **serialized** contains no internal hostname, no bucket name, no bare IP and no local storage root. One uses the real local driver and a real `ENOENT` with no mock; another drives the real S3 → public-storage fallback.

**Every guard is proven by mutation** — each of these fails exactly the test meant to pin it, and failed nothing before it was added:

| Mutation | Fails |
|---|---|
| call site back to `error: message` | all five leak tests + the partial-file test |
| drop the `httpStatusCode` attachment | the public-storage fallback test |
| `MAX_ERROR_CAUSE_DEPTH` 5 → 4 | `reads a code at the deepest level it walks` |
| `MAX_ERROR_CAUSE_DEPTH` 5 → 6 | `stops one level past the cause cap` |
| relax the `100..599` range check | `redacts 'an out-of-range status'` |
| remove the `try/catch` | `falls back when reading the error throws` |

## Review

Six rounds of the sub-agent loop (correctness, security, testing, quality, scope/regression). 28 findings posted as inline threads — 9, 9, 6 and 4 across rounds 1-4 — all addressed and resolved; rounds 5 and 6 found none. The last three rounds changed no non-comment line: rounds 2-4 were almost entirely about comments of mine that claimed more than the code delivers. Security could not break the redaction across ~110 hostile and accidental failure shapes; correctness verified the real error shapes empirically by driving `fs`, `fetch`, AWS SDK v3 and `DOMException` failures through the exported function rather than assuming them.

## Gate

`prettier` → `yarn lint` (0 errors) → `yarn typecheck` (0) → `yarn build` (0) → `yarn test` (**11721 passed, 887 files**). All green.


scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+      )
+      .map(([, value]) => String(value).toLowerCase())
+  )
+
+/**
+ * Error codes, error names and libuv syscalls are single bare words —
+ * `ENOTFOUND`, `NoSuchKey`, `getaddrinfo`. A dot, dash, colon, slash or space
+ * means it is not one, which is what keeps `minio.internal`, `10.4.2.11`,
+ * `https://minio.internal:9000` and `activitynext-prod-media` out by SHAPE
+ * rather than by pattern-matching for the forms we happened to think of. An S3
```
llun: **The comment says only `name` is server-controlled; `code` and `syscall` are too.** The AWS SDK v3's `handleError` flattens EVERY member of the XML `<Error>` element onto the data object, and `decorateServiceException` copies each onto the exception where the exception has no value — so `<Error><Code>NoSuchKey</Code><code>…</code><syscall>…</syscall></Error>` sets `error.code` and `error.syscall` as well as `error.name`. Demonstrated: a server-controlled `code`+`syscall` produced `pool_ssd_openstack storage_node_7_rack_b12 HTTP 404`. The allowlist is NOT bypassed — those tokens still pass the shape and value checks — but the comment invites a future editor to relax the filter on `code` believing it is a libuv constant.

**Suggested fix:** say code, name and syscall are all whatever the storage server sent on the S3 path.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+
+const getErrorHttpStatusCode = (level: Record<string, unknown>) => {
+  const metadata = level.$metadata
+  const candidates = [
+    typeof metadata === 'object' && metadata !== null
+      ? (metadata as Record<string, unknown>).httpStatusCode
+      : undefined,
+    level.httpStatusCode,
+    level.statusCode,
+    level.status
```
llun: **`getErrorHttpStatusCode` reads a bare `.status` and will label a non-HTTP integer `HTTP n`.** A `child_process`-style error (`status` = exit code) renders as `HTTP 127`. Unreachable from `archiveStorage` today (its failure paths are `fs.copyFile`, the S3 client and `fetch`), but it makes the manifest say something false, and `.status` is the one candidate in the list that is not an HTTP field by convention. The testing lens separately found `.statusCode` is dead: deleting that array entry leaves all 66 tests passing.

**Suggested fix:** keep only `$metadata.httpStatusCode` and `httpStatusCode`, both of which are real shapes and both exercised.

scripts/backup/productionArchive.ts
```diff
@@ -1470,17 +1476,21 @@ export const archiveStorage = async (
...
 
+          // The console line keeps the driver's own message: it goes to the
+          // operator running the export, on their own machine, and never into
+          // the archive. Only the redacted form is written to the manifest.
+          const message = error instanceof Error ? error.message : String(error)
+
           await fs.rm(archiveFilePath, { force: true })
           failedFiles.push({
-            error: nodeError.message,
+            error: redactStorageError(error, entry.source),
```
llun, line 1483: **A throwing property turns a tolerated per-file failure into a whole-export abort.** `redactStorageError` is called inside the `catch` that exists precisely so `--allow-missing-storage` can continue. If any property read it performs throws — proven with a throwing `code` getter — the throw escapes that catch and aborts the entire export. No storage driver defines such a getter, but a function whose whole job is to run inside an error handler should not be able to throw.

**Suggested fix:** make `redactStorageError` fall back to `'unknown'` on any internal throw, and pin it with a throwing-getter fixture.

scripts/backup/productionArchive.ts
```diff
@@ -1385,8 +1385,14 @@ const downloadPublicStorageFile = async ({
   )
 
   if (!response.ok) {
-    throw new Error(
-      `Failed to download ${key} from public storage: HTTP ${response.status}`
+    // The status is attached as well as interpolated. The message is for the
+    // operator's console; only a structured field survives `redactStorageError`
+    // into the manifest.
+    throw Object.assign(
```
llun, line 1391: **The `httpStatusCode` attached to the public-storage throw has no end-to-end coverage.** Removing the `Object.assign(..., { httpStatusCode: response.status })` wrapper and rerunning the full file gives 0 failures. The `redacts 'a public storage response'` case only exercises a hand-built error object; nothing drives `archiveStorage` through the `source.hostname` fallback branch (S3 `send()` throws → public HTTP fetch) at all. A wiring break in that fallback would ship silently.

**Suggested fix:** add an integration test that drives `archiveStorage` through the fallback with a failing fetch and asserts `error: 'HTTP 503'`.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+ * `https://minio.internal:9000` and `activitynext-prod-media` out by SHAPE
+ * rather than by pattern-matching for the forms we happened to think of. An S3
+ * error `name` is whatever `Code` the storage server sent, so it is untrusted
+ * input like any other.
+ */
+const SAFE_ERROR_TOKEN_PATTERN = /^[A-Za-z][A-Za-z0-9_]{0,63}$/
+
+// `fetch` reports a DNS failure as `TypeError: fetch failed` and hangs the real
+// error off `cause`, so the code worth reporting is rarely on the top level.
+const MAX_ERROR_CAUSE_DEPTH = 5
```
llun, line 1597: **`MAX_ERROR_CAUSE_DEPTH = 5` is unconstrained by the suite.** Changing it to `2` leaves all 66 tests green — the only cause-chasing test needs depth 2.

**Suggested fix:** add fixtures with a chain that resolves at depth 5 and one that truncates past the cap, so the constant is pinned in both directions.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+
+const getErrorChain = (error: unknown) => {
+  const chain: Record<string, unknown>[] = []
+  let current = error
+
+  while (
+    chain.length < MAX_ERROR_CAUSE_DEPTH &&
+    typeof current === 'object' &&
+    current !== null
+  ) {
```
llun, line 1607: **The `chain.includes(level)` cycle guard is redundant and unproven.** `MAX_ERROR_CAUSE_DEPTH` alone guarantees termination for a cycle of any length, since the loop can never push more than 5 entries. Deleting the `includes` check leaves `stops walking a cause cycle` and the whole suite passing — that test only proves termination, not that this guard provides it.

**Suggested fix:** drop the redundant check and note that the depth cap is what bounds a cycle.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+
+// The fields `redactStorageSource` keeps.
+const STORAGE_SOURCE_MANIFEST_FIELDS: readonly string[] = ['kind', 'prefix']
+
 const redactStorageSource = (source: StorageSource) => ({
   kind: source.kind,
   ...(source.prefix ? { prefix: source.prefix } : null)
 })
 
+/**
```
llun: **The three new `/** */` JSDoc blocks depart from this file's own idiom.** Every other comment in this ~2500-line file uses plain `//` line comments, including ones on exported non-trivial functions (`exportDatabase`, `createTableRowsStream`, `sortTablesForRestore`). `grep -n '^/\*\*'` returns only the three new blocks.

**Suggested fix:** convert to `//` for consistency with the file.

AGENTS.md
```diff
@@ -2321,6 +2321,13 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
 
+### The archive manifest withholds where storage lives
+
+- **`manifest.json` never carries a storage-driver message.** `redactStorageSource` (`scripts/backup/productionArchive.ts`) keeps only `kind` and `prefix`, dropping the bucket, region, endpoint, CDN hostname and local root — and `failedFiles[].error` sits in the same manifest, so it used to put all of that straight back. An export run with `--allow-missing-storage` against a private-network MinIO shipped `getaddrinfo ENOTFOUND minio.internal` inside the tarball; the local driver shipped `ENOENT: … copyfile '/srv/uploads/…'`. **This binds the actor archive too** — `scripts/backup/actorArchive.ts` imports the same `archiveStorage`, and that manifest is handed to the user whose data it is, not to the operator.
```
llun: **Documentation overclaim.** The new section opens "**`manifest.json` never carries a storage-driver message**" and then says "**This binds the actor archive too**". But `scripts/backup/actorArchive.ts:741-745` pushes a raw `error.message` into `warnings`, which is serialized into the actor archive's `manifest.json` at line 1018 — the same file, handed to the same end user. The actual disclosure there is negligible (undici's top-level message is `fetch failed`, and the remote host is already in the warning as `attachment.url`), so this is a doc-accuracy problem rather than a second leak.

**Suggested fix:** scope the claim to `storage[].failedFiles[].error` and say what `actorArchive.ts`'s own `warnings` still carries and why it is not the same disclosure.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1525,148 @@ export const archiveStorage = async (
...
+      value >= 100 &&
+      value <= 599
+  )
+}
+
+/**
+ * Reduces a storage-driver failure to what an operator can act on —
+ * `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503`
+ * — and nothing that says where storage lives. `unknown` is the honest answer
+ * for a failure that classifies itself only in its message; the export console
```
llun: **"nothing that says where storage lives" is slightly strong.** `copyfile ENOENT` reveals the backend is a local filesystem (which `source.kind` already publishes), and a vendor-specific code such as `XMinioStorageFull` names the storage *software* where `kind: 's3'` only said S3-compatible. That is a deliberate and cheap trade for actionability, but the current wording denies the fingerprint exists.

**Suggested fix:** state the trade instead of denying it.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. The comment above `SAFE_ERROR_TOKEN_PATTERN` now says all three of `code`, `name` and `syscall` are whatever the storage server sent on the S3 path, names the mechanism (the SDK's error handler flattens every member of the XML `<Error>` element onto the data object, and `decorateServiceException` copies each onto the exception wherever it has no value of its own), and explicitly warns not to relax the filter for `code` or `syscall` believing they are libuv constants. `AGENTS.md` and `CLAUDE.md` carry the same correction.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. `getErrorHttpStatusCode` now reads only `$metadata.httpStatusCode` and `httpStatusCode` — the two shapes this path actually produces — with `level.statusCode` and `level.status` dropped, and a comment recording why the bare `status` was dropped (a `child_process`-style exit code would have rendered as `HTTP 127`). The `an out-of-range status` test case was retargeted from `status: 9000` to `httpStatusCode: 9000` so it pins a field that is still read.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. `redactStorageError` is now a thin wrapper: `try { return describeStorageError(...) } catch { return UNKNOWN_STORAGE_ERROR }`, with a comment explaining it runs inside the catch that lets `--allow-missing-storage` continue and so must not be able to throw. Pinned by the new `falls back when reading the error throws` test, which uses a `code` getter that throws.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. Added `reports the status when the public storage fallback fails`, which drives `archiveStorage` through the real S3 → public-storage fallback: the S3 client throws, `source.hostname` is set, and `fetch` returns 503. It asserts the fetch URL, `error: 'HTTP 503'` in the manifest, and that the CDN hostname never appears in the serialized manifest.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. Added `reads a code at the deepest level it walks` (a chain resolving exactly at the cap) and `stops one level past the cause cap` (one level past it), built from a `buildCauseChain` helper whose wrapper errors are plain `Error`s with no coded fields, so `unknown` means the walk never reached the coded level rather than that it read a name on the way there.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. The `chain.includes(level)` check is removed from `getErrorChain`. The comment above `MAX_ERROR_CAUSE_DEPTH` now records that the depth cap is also what bounds a `cause` cycle, since the walk can never push more entries than the cap regardless of repetition. `stops walking a cause cycle` is kept, since it still proves termination.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. All three `/** */` blocks (over `getRedactedStorageSourceValues`, `SAFE_ERROR_TOKEN_PATTERN`, and `redactStorageError`) are now `//` line comments, matching the rest of the file.

AGENTS.md
llun: Fixed in `50f8f2090`. The section now scopes its opening claim to `storage[].failedFiles[].error` rather than the whole manifest, and adds a bullet stating what `actorArchive.ts`'s `warnings` still carries — a raw message for a failed remote attachment fetch — and why that isn't the same disclosure: the fetch targets a third-party host the warning already names, and the reachable messages (`fetch failed`, `HTTP <status>`, `This operation was aborted`) say nothing about this instance's own storage. It also notes not to "fix" this by pointing `redactStorageError` at a remote URL it has no source to redact against.

scripts/backup/productionArchive.ts
llun: Fixed in `50f8f2090`. The wording now states the trade instead of denying it: the output withholds bucket, region, endpoint, hostname, address and local root, but notes `copyfile ENOENT` still implies a local filesystem (already published via `source.kind`) and a vendor code like `XMinioStorageFull` names the storage software where `kind: 's3'` said only S3-compatible. `AGENTS.md` carries the same correction as its own bullet.

AGENTS.md
```diff
@@ -2321,6 +2321,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
... `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503` — and never the message. A message is free text a driver interpolates whatever it likes into, so a denylist over one is a guess about the forms someone thought of; an allowlist is a guarantee. Two filters back it: a **shape** check (`/^[A-Za-z][A-Za-z0-9_]{0,63}$/`) refuses anything carrying a dot, dash, colon, slash or space, which is what keeps `minio.internal`, `10.4.2.11` and `activitynext-prod-media` out without naming them; on the S3 path **all three of `code`, `name` and `syscall` are whatever the storage server sent**, because the SDK flattens every member of the XML `<Error>` element onto the data object and `decorateServiceException` copies each onto the exception, so none of them is a libuv constant to be trusted and the shape check is the only thing bounding them — and a **value** check refuses any token equal to something `redactStorageSource` dropped, for the single-label case the shape check cannot see (a Docker-network `hostname: 'minio'`). Both derive from one definition of "sensitive about a storage source", so a field added to `StorageSource` later is withheld from both halves for free.
```
llun: **A dropped newline merged two bullets.** The round-1 doc edit ended one bullet `…no source to redact against.` and ran straight into `- **\`redactStorageError\` is an ALLOWLIST…**` with no line break, so the allowlist rule — which carries the just-corrected `code`/`syscall` content — renders as inline prose inside a bullet about a different file's `warnings` array, with a literal `- **` mid-sentence. Prettier does not catch it (`proseWrap: preserve` does not validate markdown prose) and no other gate reads doc content, so it would have shipped silently. Found independently by three lenses.

scripts/backup/productionArchive.ts
```diff
@@ -1470,17 +1476,26 @@ export const archiveStorage = async (
...
         } catch (error) {
-          const nodeError = error as Error
           if (!options.allowMissingStorage) throw error
 
+          // The console line keeps the driver's own message: it goes to the
+          // operator running the export, on their own machine, and never into
+          // the archive. Only the redacted form is written to the manifest.
+          // `inspect` rather than `String` for a non-Error throw — it renders a
+          // plain object as its contents instead of `[object Object]`, and it
+          // does not throw on a Symbol, which is the one value `String` would
```
llun: **The comment justifying `inspect` over `String` was factually wrong.** It claimed `String` "does not throw on a Symbol, which is the one value `String` would turn into a failed export." `String(Symbol('x'))` returns `'Symbol(x)'` and does NOT throw — only interpolation (`` `${sym}` ``) does, and the old code called `String(error)`, never interpolated it. So the named hazard never existed. Worse, `inspect` is not the throw-safe choice it was presented as: it throws on a hostile `[util.inspect.custom]`, where `String` succeeds. The swap is still correct — a plain object prints as its contents rather than `[object Object]`, and `String` genuinely throws on a null-prototype object — but the durable record of why was wrong, and nothing tests this line.

scripts/backup/productionArchive.ts
```diff
@@ -1470,17 +1476,26 @@ export const archiveStorage = async (
...
+          // turn into a failed export from inside this catch.
+          const message =
```
llun: **The message render sat first in the catch, outside any guard.** Round 1 hardened `redactStorageError` so it cannot throw, and AGENTS.md stated that guarantee — but the line computing `message` ran BEFORE `fs.rm` and `failedFiles.push`, unguarded. A throw there does exactly what the hardening was for: it escapes the `catch` implementing `--allow-missing-storage`, fails the whole export on one tolerated missing file, and additionally leaves the partial file on disk. Not reachable today (every value thrown on this path is an `Error`), so it is a documentation/ordering finding rather than a live bug — but the stated property belonged to the catch block, not to the redactor alone.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1530,173 @@ export const archiveStorage = async (
...
+// — with no bucket, region, endpoint, hostname, address or local root in it.
+// Not "nothing that says where storage lives": `copyfile ENOENT` still implies
+// a local filesystem, which `source.kind` publishes anyway, and a vendor code
+// such as `XMinioStorageFull` names the storage SOFTWARE where `kind: 's3'`
+// said only S3-compatible. That fingerprint is the price of a code an operator
+// can act on, and it is worth paying — it is just not nothing.
+//
+// `unknown` is the honest answer for a failure that classifies itself only in
+// its message; the export console still printed that message in full.
+export const redactStorageError = (error: unknown, source: StorageSource) => {
```
llun: **The wrapper/body split was not earning its keep, and it inverted the block's definition order.** `describeStorageError` had exactly one caller and no direct test — every test still went through `redactStorageError` — so it was a second private name existing only to host a body. It also placed the exported wrapper above both `UNKNOWN_STORAGE_ERROR` and the logic, reversing the bottom-up order every other helper in this block follows (`getRedactedStorageSourceValues` → `SAFE_ERROR_TOKEN_PATTERN` → `getErrorChain` → `getErrorHttpStatusCode`).

CLAUDE.md
```diff
@@ -33,6 +33,7 @@
...
...tim (`getaddrinfo ENOTFOUND minio.internal` from an `--allow-missing-storage` export against a private-network MinIO). It binds `scripts/backup/actorArchive.ts` too, which imports the same `archiveStorage` and hands its manifest to the user. The helper is an **allowlist** — syscall, code, name, HTTP status, never the message — filtered by SHAPE (no dots, dashes, colons, slashes or spaces) and by VALUE (nothing `redactStorageSource` dropped). On the S3 path **`code` and `syscall` are as server-controlled as `name`** (the SDK flattens every `<Error>` member onto the exception), so the shape check is the only bound on them — don't relax it believing they are libuv constants. It cannot throw either: it runs inside the `catch` that lets `--allow-missing-storage` continue. Don't swap it for a regex scrubber over the message, don't redact the **console** line beside it (that goes to the operator's own machine and is where the full detail still lives), and note the claim covers that one field — `actorArchive.ts`'s `warnings` deliberately still carries a remote-fetch message about a third-party host it already names. See **The archive manifest withholds where storage lives** in `AGENTS.md`.
```
llun: **The new prose restated the code comments near-verbatim.** Three additions duplicated their counterparts almost word for word: the `decorateServiceException` mechanism, the `copyfile ENOENT` / `XMinioStorageFull` fingerprint examples, and the "redactor cannot throw" paragraph. The repo's convention is AGENTS.md = durable rule, code = local why; both were carrying the same mechanism-level detail. This CLAUDE.md bullet had grown by absorbing that content rather than compressing it, weakening its role as a pointer.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1530,173 @@ export const archiveStorage = async (
...
+      typeof value === 'number' &&
+      Number.isInteger(value) &&
+      value >= 100 &&
+      value <= 599
+  )
+}
+
+// Reduces a storage-driver failure to what an operator can act on —
+// `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503`
+// — with no bucket, region, endpoint, hostname, address or local root in it.
```
llun: **"with no bucket, region, endpoint, hostname, address or local root in it" claimed an absolute the shape check does not provide.** The check bounds FORM, not content: a storage server that writes its own error responses can put up to 64 arbitrary `[A-Za-z][A-Za-z0-9_]*` characters into each of `code`, `name` and `syscall` — verified, `code: 'activitynextprodmedia'` (the bucket with its dashes removed) and `syscall: 'srvUploadsMedia'` both survive, because the value check only catches exact case-insensitive matches. Severity informational: the only actor who can do this is the storage backend, which already serves the file bytes going into the same tarball. But the comment should say "no value in the SHAPE those take".

AGENTS.md
llun: **The `actorArchive.ts` warnings enumeration reads exhaustive but is not.** It listed the reachable messages as `fetch failed`, `HTTP <status>` and `This operation was aborted`. The same `try` also spans `new URL(...)` and the `fs.mkdir`/`fs.writeFile` into the staging directory, so an `EACCES: permission denied, mkdir '<stagingDir>/remote-media'` or an `ENOSPC` can land in `warnings` with the operator's staging path interpolated. The security claim survives — a staging path is not the storage root and is not what `redactStorageSource` withholds — but the list should not read as complete.

AGENTS.md
```diff
@@ -2321,6 +2321,14 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- Covered by `scripts/backup/productionArchive.test.ts` — a table over `redactStorageError` plus five `archiveStorage` tests that assert the whole manifest, serialized, contains no internal hostname, bucket name, bare IP or local storage root. Four of them fail if the call site goes back to `error.message`; the fifth drives the S3 → public-storage fallback and is the only thing that notices if the `httpStatusCode` attachment is dropped. `MAX_ERROR_CAUSE_DEPTH` is pinned in both directions by a chain that resolves at the cap and one that truncates past it.
```
llun: **"Four of them fail if the call site goes back to `error.message`" understated it.** The fifth test (`reports the status when the public storage fallback fails`) asserts `failedFiles: [{ error: 'HTTP 503', … }]` with `toEqual`, so it fails on that revert too; only its `not.toContain` half would still pass.

scripts/backup/productionArchive.test.ts
```diff
@@ -1159,6 +1158,316 @@ describe('production archive scripts', () => {
...
+    })
+
+    it('stops walking a cause cycle', () => {
+      const error = Object.assign(new Error('failed'), {
+        cause: undefined as unknown
+      })
+      error.cause = error
+
+      expect(redactStorageError(error, source)).toBe('unknown')
+    })
```
llun, line 1434: **`it('stops walking a cause cycle')` named a mechanism that no longer exists.** The `chain.includes` guard it was written against was removed in this same commit; the test now proves the depth cap terminates the walk. Still a valid termination test — an unbounded walk would hang the suite — but the name misdescribes what it guards.

AGENTS.md
llun: Fixed in 114d799ea. AGENTS.md restores the dropped newline — the "claim is about that one field" bullet and the "`redactStorageError` is an ALLOWLIST" bullet are now two separate bullets again, so the allowlist rule no longer renders as inline prose inside the wrong bullet.

scripts/backup/productionArchive.ts
llun: Fixed in 114d799ea. The comment on the message-render line now says: "The swap is still right (a plain object prints as its contents, not `[object Object]`)" and drops the false `String`-throws-on-a-Symbol claim, while also noting `inspect` throws on a hostile `[util.inspect.custom]`.

scripts/backup/productionArchive.ts
llun: Fixed in 114d799ea. The message computation and `console.error` call now run last in the `catch`, after `fs.rm(archiveFilePath, ...)` and `failedFiles.push(...)`, with a comment explaining it is the one step not guaranteed not to throw and why it is ordered last.

scripts/backup/productionArchive.ts
llun: Fixed in 114d799ea. `describeStorageError` is gone — `redactStorageError` is a single function with one `try`/`catch`, and the block now reads bottom-up: `getErrorHttpStatusCode` → the doc comment → `UNKNOWN_STORAGE_ERROR` → `redactStorageError`, consistent with the rest of the file.

CLAUDE.md
llun: Fixed in 114d799ea. The `decorateServiceException` mechanism detail is removed from AGENTS.md/CLAUDE.md and pointed at the code comment instead ("the mechanism is recorded at the filter"); the `XMinioStorageFull` fingerprint bullet is condensed to state the trade rather than re-deriving it; and the "redactor cannot throw" bullet is shortened to the outcome, leaving the property-read rationale to the code comment.

scripts/backup/productionArchive.ts
llun: Fixed in 114d799ea. The doc comment now reads "carrying no value in the SHAPE a bucket, region, endpoint, hostname, address or local root takes" and explicitly calls out that a storage server can put up to 64 arbitrary characters into a token (e.g. `activitynextprodmedia` survives), so it no longer claims an absolute.

AGENTS.md
llun: Fixed in 114d799ea. The AGENTS.md bullet now adds: "That `try` also spans `new URL` and the staging-directory write, so a local `EACCES`/`ENOSPC` can land there naming the STAGING path — still not the storage root... but do not read the list as exhaustive."

AGENTS.md
llun: Fixed in 114d799ea. Corrected to "All five fail if the call site goes back to `error.message`, and the fifth — which drives the S3 → public-storage fallback — is additionally the only thing that notices if the `httpStatusCode` attachment is dropped."

scripts/backup/productionArchive.test.ts
llun: Fixed in 114d799ea. Renamed to `it('terminates on a cause cycle', ...)`.

scripts/backup/productionArchive.ts
```diff
@@ -1470,17 +1476,28 @@ export const archiveStorage = async (
...
           if (!options.allowMissingStorage) throw error
 
           await fs.rm(archiveFilePath, { force: true })
           failedFiles.push({
-            error: nodeError.message,
+            error: redactStorageError(error, entry.source),
             path: relativeFilePath
           })
+
+          // Rendering the message is the LAST thing this catch does, because it
```
llun: **The catch-ordering rationale was false, and both halves of it were unobservable.** The comment said rendering the message last means a throw there "cannot cost the partial-file cleanup or the manifest entry above." Neither survives: `archiveStorage` wraps its loop in `try { … } finally { s3Client?.destroy() }` with no `catch`, so the throw unwinds past `storageManifests.push` and `writeManifest` is never reached — no manifest is written at all. And both callers delete the whole staging directory recursively in their own `finally` (`productionArchive.ts` and `actorArchive.ts`), so the `fs.rm` of the one partial file buys nothing once the export unwinds. Found independently by the correctness and testing lenses; testing proved it by running the scenario (`didArchiveStorageThrow: true`, `manifest: undefined`). Secondary: "the one step not guaranteed not to throw" also overlooked `await fs.rm`, which can throw on EPERM/EACCES.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1532,176 @@ export const archiveStorage = async (
...
+      value <= 599
+  )
+}
+
+// Reduces a storage-driver failure to what an operator can act on —
+// `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503`
+// — carrying no value in the SHAPE a bucket, region, endpoint, hostname,
+// address or local root takes. Shape is the bound, not content: a storage
+// server that writes its own error responses can put up to 64 arbitrary
+// `[A-Za-z][A-Za-z0-9_]*` characters into each token, so `activitynextprodmedia`
```
llun: **The residual example overclaimed the risk and contradicted the value check.** It said `activitynextprodmedia` "survives where `activitynext-prod-media` does not." Measured: if that IS the configured bucket, `getRedactedStorageSourceValues` holds it and the token is refused — the result is `unknown`. The example only holds for a single-label name the source does not carry. As written it reads as "a dashless bucket name leaks," which argues the value check does nothing and invites its removal — directly contradicting AGENTS.md's own justification for that check.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1532,176 @@ export const archiveStorage = async (
...
+const STORAGE_SOURCE_MANIFEST_FIELDS: readonly string[] = ['kind', 'prefix']
+
 const redactStorageSource = (source: StorageSource) => ({
   kind: source.kind,
   ...(source.prefix ? { prefix: source.prefix } : null)
 })
 
+// Every string `redactStorageSource` drops, lowercased. Deriving it from the
+// source rather than listing the fields again keeps one definition of "what is
+// sensitive about a storage source", so a field added to `StorageSource` later
```
llun: **"A field added to `StorageSource` later is withheld from both halves for free" is only true for STRING fields.** `redactStorageSource` is a two-name allowlist, so it genuinely covers anything new. The value set does not: it filters `typeof value === 'string'`, so a nested `credentials: { accessKeyId, secretAccessKey }` would never enter it — and `AKIAIOSFODNN7EXAMPLE` passes the shape check cleanly (verified). The storage server receives that key id in every SigV4 `Authorization` header, so it is a value the one actor who can set these tokens already holds. `StorageSource` is all-string today, so this is latent, not exploitable — but the promise is exactly what would mislead the person adding the field. (Also flagged in AGENTS.md's matching bullet.)

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1532,176 @@ export const archiveStorage = async (
   return storageManifests
 }
 
+// Manifest redaction. The manifest travels inside the tarball, so it
```
llun: **Round 2 answered a comment-duplication finding by adding a comment.** The new shape-bounds-form paragraph was a second, independent explanation of the point already made 60 lines above over `SAFE_ERROR_TOKEN_PATTERN`, using the same example string. Of the file's 90 comment-only lines, 73 sat inside this one ~170-line block, against ~4% density elsewhere in a 2562-line file — and the block GREW in the commit that claimed to be de-duplicating.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1532,176 @@ export const archiveStorage = async (
...
+// deliberately says nothing about WHERE production storage lives:
+// `redactStorageSource` keeps `kind` and `prefix` and drops the bucket, region,
+// endpoint, CDN hostname and local root.
+//
+// `failedFiles[].error` sits directly beside it in the same manifest and used
+// to carry the driver's message verbatim, which put back exactly what the line
+// above had removed. An export run with `--allow-missing-storage` against a
+// private-network MinIO landed `getaddrinfo ENOTFOUND minio.internal` in the
```
llun: **The AGENTS.md trim addressed the shape of the duplication finding, not its substance.** Three narratives still ran in both places nearly word for word: the MinIO incident with its quoted strings (this comment block), the `decorateServiceException` mechanism (line 1589 below), and the "Docker-network `hostname: 'minio'`" example (line 1567 below). The repo convention is AGENTS.md = durable rule, code = local why.

AGENTS.md
```diff
@@ -2321,6 +2321,15 @@ A full sub-agent review round yields no new actionable comments, or you have run
...
+- Covered by `scripts/backup/productionArchive.test.ts` — a table over `redactStorageError` plus five `archiveStorage` tests that assert the whole manifest, serialized, contains no internal hostname, bucket name, bare IP or local storage root. All five fail if the call site goes back to `error.message`, and the fifth — which drives the S3 → public-storage fallback — is additionally the only thing that notices if the `httpStatusCode` attachment is dropped. `MAX_ERROR_CAUSE_DEPTH` is pinned in both directions by a chain that resolves at the cap and one that truncates past it.
```
llun: **The revert-proof count was still imprecise.** Reverting the call site to `error.message` fails six tests, not five: the five that assert the serialized manifest excludes a secret, plus the pre-existing `removes partial files for allowed missing storage downloads`, which asserts `error: 'unknown'`.

scripts/backup/productionArchive.ts
llun: Fixed in `336516923`. The catch comment no longer claims the ordering protects partial-file cleanup or the manifest entry — it now says the message computation is defensive only, and if it ever threw, the export's own `finally` deletes the whole staging directory regardless of line order, so no ordering here "contains" anything. (`await fs.rm`'s own throw risk is covered by the same correction: nothing on this path is claimed to guard against it anymore.)

scripts/backup/productionArchive.ts
llun: Fixed in `336516923`. The `activitynextprodmedia` residual example is removed entirely; the doc comment above the reducing function now states the two deliberate limits (shape bounds form not content; a vendor code still names the software) without the misleading example, and points to AGENTS.md for the full trade-off argument.

scripts/backup/productionArchive.ts
llun: Fixed in `336516923` (and the matching AGENTS.md bullet). The comment above `getRedactedStorageSourceValues` now says a STRING field added to `StorageSource` is withheld for free, but a non-string one is not — the filter skips it, so a nested `credentials: { accessKeyId }` would sail through, matching your verified access-key-id example. `StorageSource` is noted as all-string today.

scripts/backup/productionArchive.ts
llun: Fixed in `336516923`. Comment-only lines in the file dropped from 90 to 73 (2562 → 2545 total lines) — the block header, the shape-bound paragraph, and the catch comment were all cut rather than grown, per the round-3 commit message: "Cut the comment duplication rather than adding to it."

scripts/backup/productionArchive.ts
llun: Fixed in `336516923`. All three narratives are now single-sourced: the MinIO incident with its quoted strings and the Docker-network `hostname: 'minio'` example are gone from the code and live only in AGENTS.md (the code comment now points to it by name — "See AGENTS.md, 'The archive manifest withholds where storage lives'"); the `decorateServiceException` mechanism detail stays in code only, and AGENTS.md/CLAUDE.md do not repeat it (verified by grep).

AGENTS.md
llun: Fixed in `336516923`. The AGENTS.md coverage bullet now reads "Reverting the call site to `error.message` fails all five plus the pre-existing partial-file test," correcting the count to six.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1527,164 @@ export const archiveStorage = async (
...
 const redactStorageSource = (source: StorageSource) => ({
   kind: source.kind,
   ...(source.prefix ? { prefix: source.prefix } : null)
 })
 
+// Every string `redactStorageSource` drops, lowercased. Deriving it from the
+// source rather than listing the fields again keeps one definition of "what is
+// sensitive about a storage source", so a STRING field added to `StorageSource`
+// later is withheld from both halves without anyone remembering this one. A
+// non-string one is not: the filter below skips it, so a nested
```
llun, line 1554: **The commit that removed code/AGENTS.md duplication introduced a fresh instance of it.** Round 3's new non-string-field paragraph went into BOTH files at once — same claim, same `credentials: { accessKeyId }` example, same closing note — added by the very commit meant to be resolving that problem. The same diff applied the correct split elsewhere (the MinIO incident cut from code and left AGENTS.md-only with a pointer; the `decorateServiceException` mechanism kept in code with AGENTS.md reduced to "the mechanism is recorded at the filter"), so the pattern was understood and simply not applied here. It is mechanism detail about `getRedactedStorageSourceValues`, so the code should own it and AGENTS.md should hold the rule.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1527,164 @@ export const archiveStorage = async (
...
+  )
+}
+
+// Reduces a storage-driver failure to what an operator can act on —
+// `getaddrinfo ENOTFOUND`, `copyfile ENOENT`, `NoSuchKey HTTP 404`, `HTTP 503`
+// — carrying no value in the SHAPE a bucket, region, endpoint, hostname,
+// address or local root takes. Two deliberate limits come with that: the token
+// pattern above bounds form and not content, and a vendor code such as
+// `XMinioStorageFull` still names the storage software. Both are the price of
+// a code an operator can act on; AGENTS.md argues the trade.
```
llun: **A "why" was deleted and replaced with a pointer to a document that does not carry it.** Round 3 removed "That costs nothing — the only actor who can do it already serves the file bytes going into the same tarball", which is the entire argument for bounding FORM rather than content, and replaced it with "AGENTS.md argues the trade". AGENTS.md argues the vendor-fingerprint limit; it says nothing about form-vs-content. The deleted sentence survived nowhere — not AGENTS.md, not CLAUDE.md, not a test, not another comment. Failure scenario: a reviewer reads "bounds form and not content", asks whether the filter should become a fixed allowlist of known error codes (which would break `XMinioStorageFull`, `NoSuchBucket` and every future vendor code), follows the pointer, and finds no answer to the question the comment raised.

scripts/backup/productionArchive.ts
```diff
@@ -1470,17 +1476,23 @@ export const archiveStorage = async (
...
           await fs.rm(archiveFilePath, { force: true })
           failedFiles.push({
-            error: nodeError.message,
+            error: redactStorageError(error, entry.source),
             path: relativeFilePath
           })
+
+          // The message goes to the operator's own console, never into the
+          // archive — only the redacted form reaches the manifest. Nothing on
+          // this path throws a non-Error, so `inspect` is defensive only; were
```
llun: **The reason `inspect` is used rather than `String` was dropped, and `inspect` is imported for that single call.** `inspect` appears exactly twice in the file — the import and this one call. With the rationale gone, `String(error)` is the obvious simplification and it also removes the import, so the change is inviting. The cost is silent: a thrown non-`Error` would print as `[object Object]` on the operator's console, which by this block's own design is the only place the driver's full message survives at all. No test exercises that branch — the `redactStorageError` table's "a thrown non-error" case never reaches `console.error`.

scripts/backup/productionArchive.ts
```diff
@@ -1515,11 +1527,164 @@ export const archiveStorage = async (
   return storageManifests
 }
 
+// Manifest redaction. The manifest travels inside the tarball, so it withholds
```
llun, line 1533: **Comment density measurement.** After round 3 the block held 61 of the file's 73 comment-only lines — 83.6% of the file's commentary in a ~157-line span, against roughly 0.5% density across the other ~2388 lines. Round 3 cut real lines but one sub-paragraph grew by four with brand-new content, and the block's SHARE of the file's comments went up rather than down.

scripts/backup/productionArchive.ts
llun: Fixed in `0563b488b`. The AGENTS.md bullet ("`redactStorageError` is an ALLOWLIST...") now ends "so a **string** field added to `StorageSource` later is withheld from both halves for free. A non-string one is not, and has to be taught to the value set by hand — the filter records why," dropping the duplicated `credentials: { accessKeyId }` example and closing note. The mechanism-level paragraph (the example, the shape-check detail) stays only in the code comment above `getRedactedStorageSourceValues`, unchanged from before this fix — matching the split already applied elsewhere in this same file (MinIO incident, `decorateServiceException`).

scripts/backup/productionArchive.ts
llun: Fixed in `0563b488b`. The justification is restored where the form-vs-content bound is actually argued, at the `SAFE_ERROR_TOKEN_PATTERN` comment: "Bounding their FORM rather than their content is enough, and deliberately so: the only actor who can choose these tokens is the storage server itself, and it already serves the file bytes going into the same tarball." The `redactStorageError` doc comment's pointer is corrected to say which half is argued where: "the token pattern above bounds form and not content — argued there — and a vendor code such as `XMinioStorageFull` still names the storage software, which AGENTS.md argues." So a reviewer asking "should this become a fixed allowlist of known codes" now finds the answer at the token pattern rather than a dead pointer.

scripts/backup/productionArchive.ts
llun: Fixed in `0563b488b`. The comment above the `console.error` call now restores the reason: "`inspect` rather than `String` so a thrown non-Error prints as its contents instead of `[object Object]`; nothing on this path throws one, so that branch is defensive, and were `inspect` itself to throw the export unwinds and its `finally` deletes the whole staging directory — no ordering of these lines contains that." `inspect` and its import are no longer an unexplained no-op-looking choice.

scripts/backup/productionArchive.ts
llun: Checked rather than asserted. As of the current head (`0563b488b`), the redaction block (`// Manifest redaction...` through the closing brace of `redactStorageError`) is 161 lines and holds 65 of the file's 80 comment-only lines: 81.25% of the file's commentary (down from 83.6% at `336516923`'s 157-line/61-of-73 baseline), against 15 of 2391 lines outside it (0.63%, up from 0.50%).

So the block's SHARE of the file's total comments moved down this round, not up — but only because the fix for finding C added 3 comment lines *outside* the block (restoring the `inspect` rationale) while the fixes for findings A/B netted +4 comment lines *inside* it (restoring the deleted form-vs-content justification, net of the AGENTS.md-duplication trim). The block itself did not shrink — it grew again in absolute terms (157→161 lines, 61→65 comment lines), for the same reason round 3's growth did: correctness/documentation fixes landing here add lines, they don't remove them. This finding was not itself the target of a fix in this round; leaving it open rather than resolving it to tidy it away.

llun: Adjudicated rather than changed, in round 5.

Round 4 did not target density — the block went 157→161 lines because two other findings (the deleted `already serves the file bytes` justification, and the dropped `inspect`-over-`String` reason) each required restoring content that a correctness lens and this lens had both called load-bearing. Leaving this open at the time was right.

In round 5 the quality lens re-measured the final state against round 3's peak of 66 in-block comment lines and closed the concern on the merits: the +4 is "a small, targeted re-addition, not a regression to the earlier bloat", the non-string mechanism now has exactly one owner (code, with AGENTS.md holding only the rule), both pointers resolve to the comment that actually carries the argument, and it "could not find a new gap worth a fifth round".

So the density is what it is because two justifications were restored deliberately. Resolving on that basis — no change made, and none warranted.

PR #1587, opened 2026-08-26
fix: accept only our own stored media as a profile image URL

llun:
## The problem

`POST /api/v1/accounts/profile` declared `iconUrl` and `headerImageUrl` as bare `z.string()`s and spread them straight into `database.updateActor`, which stores them verbatim. The route is behind `AuthenticatedGuard`, so any signed-in local user could set their own actor's avatar or header to an arbitrary string.

That matters in three ways:

- `actor.iconUrl` is rendered as an `` for every viewer of the profile, so an arbitrary host becomes a tracking pixel leaking each viewer's IP and user agent to a third party the viewer never chose.
- It is federated as the actor's `icon.url`, so the same URL is republished to every instance that fetches the actor.
- A `data:` URL, or one naming an internal address, was equally acceptable to the schema.

This was surfaced while fixing the path-traversal bug in the actor archive script (#1569). That PR closed the arbitrary-file-read sink; the unvalidated field itself was left alone deliberately, as a separate problem with a different blast radius. This is that follow-up.

## The rule

Both fields now go through `parseProfileImageUrl` (`lib/services/accounts/profileImageUrl.ts`), which accepts **only a URL naming media this instance already stores** — a `/api/v1/files/…` path on one of our own hosts, resolved through the existing `getMediaPathFromFileUrl`.

That is not a new contract for any real client. It is exactly what the settings upload button produces, and exactly what `update_credentials` produces from an uploaded file — the Mastodon-compatible path never accepted a URL at all, only a `File`. The free-URL contract had one consumer: the web settings form.

**`z.url()` is not this check and does not stand in for it.** Measured on this repo's Zod 4.4.3:

| input | `z.string().url()` | `z.url()` |
|---|---|---|
| `javascript:alert(1)` | passes | passes |
| `data:image/svg+xml;base64,…` | passes | passes |
| `file:///etc/passwd` | passes | passes |

So shape validation alone would have left exactly the schemes worth refusing. The protocol allowlist runs **before** the host check, for the same reason `getProductUrlHostname` needs one: `new URL` parses an authority for non-special schemes too, so `javascript://<our-host>/%0aalert(1)` presents a perfectly good hostname.

A host-relative `/api/v1/files/…` is refused as well, even though `getMediaPathFromFileUrl` happily resolves one — the value is federated as `icon.url`, where a relative reference names the **reader's** origin rather than ours.

## Also in scope

**`POST /api/v1/accounts/image`** (the account avatar) shares the one helper. It already validated with `z.string().url()`, which per the table above left it the same three schemes. Both surfaces use the same `ImageUploadField`, so neither loses anything.

**Four dead fields removed from `ProfileRequest`.** It also declared `publicKey`, `followersUrl`, `inboxUrl` and `sharedInboxUrl` as free strings, and `updateActor` persists all four — so any signed-in user could rewrite the public key and inbox endpoints their own actor publishes in its ActivityPub document. No form has ever sent them. (I traced the delivery path before scoping this: `getRemoteActorInboxes` filters local actors, so it was not an SSRF into the delivery job.)

## Two consequences of the validation, not separate changes

**An empty submitted value now clears the image.** The settings form submits both fields on every save, so once validation exists an empty one has to mean something. It now clears — matching what the account image route has always done, and fixing an affordance that was silently a no-op: there was previously no way to clear an avatar from `/settings` at all.

**`ImageUploadField`'s text box is `readOnly`, with an explicit Remove button.** Its placeholder read `https://example.com/avatar.jpg`, actively inviting the remote URL the server now refuses. Removing the editable box would have taken clearing with it, hence the Remove button. Both icon-only buttons gained `aria-label`s.

## Scope deliberately not taken

**Input-time only — no backfill, no render-time gate.** A remote actor's `iconUrl` is written from its actor document by `getPersistableProfile` and is legitimately remote, so a render-time host gate would break every federated avatar. Existing rows are left alone: a local actor who pasted a remote URL keeps their avatar until they next save.

## Verification

Full gate green: `prettier --check`, `lint`, `typecheck`, `build`, and the full suite (891 files / 11,784 tests).

Every behavioural change is revert-proofed — reverting it fails the test that names it. A round-5 mutation sweep over the whole validator applied 19 mutations and killed 17; both survivors were real gaps and are now closed (see the review rounds below).

Manually verified end to end against a local SQLite instance, through the real routes:

| submitted | `/accounts/profile` | `/accounts/image` |
|---|---|---|
| own-instance media URL | 307, stored | stored |
| host `evil.example` | 422 | refused, stored value unchanged |
| `javascript://…/%0aalert(1)` | 422 | refused |
| `data:image/svg+xml;…` | 422 | refused, stored value unchanged |
| `/api/v1/files/..%2f..%2fsecrets/env` | 422 | — |
| empty | cleared, persists across reload | cleared |

The settings field renders read-only with the new placeholder, and Remove appears only for a field that has a value.

## Review rounds

Five rounds of the sub-agent review loop ran, each pointed at the previous round's fixes. Every finding was put through an adversarial verifier before being acted on. Round-by-round detail is in the comments; the short version:

| Round | Found |
|---|---|
| 1 | The validation bricked the entire `/settings` form for any actor with a pre-existing remote avatar (high severity). Plus two accessibility defects in the new Remove button. |
| 2 | The round-1 fix compared a trimmed submission against a raw stored value, so it missed the whitespace-carrying rows it existed to protect. The `bg-muted` fix worked only in light mode. |
| 3 | The round-2 fix made a whitespace-only stored value swallow an explicit clear, leaving Remove a permanent silent no-op for those rows. |
| 4 | The round-3 reorder over-corrected: a save that never touched the image wrote `null` over such a row. A test deleted in round 3 turned out to be the only guard on the comparison's left-hand side. |
| 5 | A mutation sweep found the protocol allowlist was **not pinned by any test** — every scheme case also failed the host or path check, so deleting the allowlist passed everything. Also that the `currentValue &&` guard was unpinned, where a `!== undefined` variant 500s the first upload on any new account. |

Four consecutive rounds found a regression in `parseProfileImageUrl`, each introduced by the previous round's fix. The three early-return branches now have to be in exactly the order they are in — each of the three orderings has shipped and each broke one of the other two — so that ordering, and the reason for it, is written into AGENTS.md rather than left to be rediscovered.

One suggestion was declined: converting `parseProfileImageUrl` to an options object to prevent a transposed call. The footgun fails loudly (a transposed call refuses a legitimate upload with a visible 422) and the change would churn 35 test call sites.


llun:
## Review loop — round 1

Five lenses (correctness, design-system, quality, scope-regression, testing) ran against the full diff. Every finding was then put through an adversarial verifier that tried to refute it. Three survived; all three are fixed in b6f46211a.

### 1. The validation could brick the whole `/settings` form — high severity, CONFIRMED

The regression I shipped in the first commit. `/settings` is a **single** form around post line limit, name, summary, both images and the privacy switch, with one Update button — and `ImageUploadField` seeds its hidden input from the **stored** URL and resubmits it untouched.

So an actor carrying a URL stored before this rule existed would 422 the whole form while editing only their display name. `apiErrorResponse` answers a bare JSON body with no `Location`, so a native form post renders that JSON in place of the page: the name edit is lost, with no error UI, and nothing on `/settings` could be saved again until the user worked out the image had to be removed first.

The verifier added two facts I had missed. `/settings` applies no `isRealAvatar` filter on read (unlike `/account`), so a legacy value reaches the field verbatim. And `git log -p --follow` shows the field was a bare `z.string()` with an `https://example.com/avatar.jpg` placeholder for the route's **entire history** — so the triggering precondition is not hypothetical.

**Fix:** `parseProfileImageUrl` now takes the currently stored value and reports a submission matching it as "no change" instead of re-validating it. That is the same "existing rows are left alone" rule the PR already stated, enforced at the one place a stored value comes back in. It gives nothing away — a new value still has to pass, and clearing still works, so a stale value stays removable rather than sticky.

`POST /api/v1/accounts/image` then had to **skip** its write on "no change": `updateAccountImage` always writes, so passing the result through as `iconUrl.value ?? null` would have cleared the very image the form was resubmitting unchanged.

### 2. Remove dropped keyboard focus to `<body>` — CONFIRMED

The Remove button renders only while a value is set, so activating it unmounts the button that was just pressed, and a focused element removed from the document drops focus to `<body>` — sending the next Tab back to the top of the page (WCAG 2.4.3). The verifier reproduced it with a passing jsdom assertion.

AGENTS.md documents this exact mechanism for the media strip's chevrons, but their fix (leave the control out of the tab order) does not transfer: Remove is the only way to clear the image. **Fix:** focus Upload before clearing.

### 3. The read-only field looked editable — CONFIRMED, low severity

`Input` styles `disabled` but not `readOnly`, so the field was pixel-identical to the genuinely editable Name and Summary fields above it. Four of five comparable display-only fields in this repo use a muted background. **Fix:** `bg-muted`.

It stays `readOnly` rather than becoming `disabled` — the verifier's point, and a good one: the preview above is a background-image `div` that assistive tech cannot see, so this field is the only announced representation of which image is set, and `disabled` would drop it from the tab order.

### Also closed

Two coverage gaps the testing lens named: a trusted-alias host is now exercised end to end through **both** routes (previously only in the unit test, so the route wiring of that branch was untested), and one image is cleared while the other is set.

### Cleared

Quality and scope-regression found nothing. Between them they independently confirmed that `update_credentials` and `PATCH /api/v1/profile` only ever take `File` uploads so the validator is never in their path; that `DELETE /api/v1/profile/avatar|header` bypasses it by design; that no script, mock, doc or pre-existing test wrote through either route; that the discriminated-union return shape matches the existing `OAuthGuard` convention rather than inventing one; and that every factual claim in the new comments checks out against the code.

All three fixes are revert-proofed — reverting each one fails the tests that name it. Full gate green: `prettier --check`, `lint`, `typecheck`, `build`, 891 files / 11,775 tests.

Round 2 is running against the fixes.

llun:
## Review loop — rounds 2 through 7

Round 1 is in the comment above. Since then:

**Round 2** — two gaps in round 1's own fixes, each found independently by two lenses. The "no change" bypass compared a *trimmed* submission against a *raw* stored value, so it missed exactly the legacy rows it existed to protect. And `bg-muted` only worked in light mode: `Input`'s base carries `dark:bg-input/30`, and this project compiles the dark variant as `&:is(.dark *)`, where `:is()` takes its most specific argument's specificity — so a bare `bg-muted` loses the cascade. The design-system lens proved that by compiling the project's real Tailwind and diffing computed `background-color` in headless Chrome, not by reading. A jsdom `toHaveClass('bg-muted')` passes against the broken form, so the test now asserts `dark:bg-input` is *absent* from the rendered class list.

**Round 3** — round 2's fix made a pure-whitespace stored value collide with an explicit clear. Such a row is truthy but trims to `''`, the same thing Remove's empty submission trims to, and the match was tested first — so the write was dropped. The actor sees a Remove button over a preview that cannot load, clicks it, gets a successful redirect, and nothing happens, permanently, since the field is read-only.

**Round 4** — round 3's reorder over-corrected: deciding the empty submission first made it fire for a field the user never touched, writing `null` over a legacy row while saving an unrelated field. The two signals both trim to `''` and mean opposite things, so they are now told apart *before* trimming. Round 4 also caught that a test deleted in round 3 was the only guard on the comparison's left-hand side — mutating `trimmed` to `value` there survives every other test in the file and 422s the whole form. Restored with a comment saying what it actually guards.

**Round 5** — a mutation sweep over the whole validator: 19 mutations, 17 killed, and both survivors were real.

Deleting the **protocol allowlist entirely left every test passing.** Every scheme case — `javascript:`, `data:`, `file:`, `ftp:` — also fails the host or path check, so none pinned the scheme test on its own. `new URL` gives `ftp:` and `ws:` a real authority where `javascript:` and `data:` have none, so `ftp://<our-host>/api/v1/files/abc.jpg` resolves cleanly through `getMediaPathFromFileUrl` and would have been accepted, stored and federated. This PR had been calling that allowlist "the load-bearing half" for five rounds with nothing testing it. Two cases now pin it independently.

The second survivor: swapping the trimmed match's `currentValue &&` guard for `!== undefined`. `accounts.iconUrl` is nullable and the domain type carries the DB's `null` through, so that is the state of every account that has never set an avatar — the most common path through the image route — and the mutation dereferences that null and 500s the first upload.

**Round 6** — five rounds of fixes had left the ordering rule splayed across four multi-paragraph comments interleaved with three one-line `if`s. Unreadable ordering logic is the root cause of the four consecutive regressions, so it is now a named `classifySubmission` returning `unchanged`/`clear`/`candidate`, with the rationale stated once. A literal extract-method: same conditions, same order. The mutation checks were re-run against the new structure rather than trusting the claim.

**Round 7 — clean.** Correctness executed the old and new logic across all 77 value/stored-value combinations and found zero mismatches. A dedicated security review traced every writer of `actors.settings.iconUrl`/`headerImageUrl` and `accounts.iconUrl` and confirmed the boundary: the `unchanged` path returns `value: undefined` and both callers skip the write, so it can never persist the submitted string; `getPersistableProfile`, the one attacker-influenced writer, is guarded by `if (existingActor?.privateKey) return existingActor` and can never touch a local actor. Field removal complete, no SSRF (the archive script never fetches avatars), no lockout, CSRF closed by `hasSameOriginProof` running before session resolution.

### One operational decision for the maintainer

This PR is input-time only by design — no backfill. The security review's one note is worth repeating plainly: **an instance that ran the vulnerable code in production still serves any malicious `iconUrl` already in its database**, until that user clears or replaces it. Nothing here retroactively sanitizes stored rows. Whether to audit existing local-actor rows is an operational call, not a code one.

PR #1589, opened 2026-08-26
minor: give a gear component a history of install periods

llun:
## Summary

A `fitness_gear_components` row carried **one** `[addedAt, removedAt)` install window, so a part that comes off and goes back on could not be represented. Clearing `removedAt` — which is what the components card's Unretire did — reopened the **original** window, so a part unretired months later was retroactively credited every activity ridden while it sat off the bike, and re-retiring did **not** take that back: it only closed the window again at the new today.

Verified against the real database layer before the change, with a throwaway test:

- component `addedAt = 2026-01-01`, rides of 10 km on 2026-01-15 and 20 km on 2026-02-15
- `removedAt = 2026-02-01` → rollup reads `{10_000, 1}` — correct
- `updateFitnessGearComponent({ removedAt: null })` (what Unretire sent) → `{30_000, 2}`
- `retireFitnessGearComponent` → still `{30_000, 2}`

The same shape blocked any genuine refit: a wheelset taken off for winter and put back on could not be expressed without crediting the winter.

## Model

`addedAt` and `removedAt` move **off** `fitness_gear_components` into a row per period in the new `fitness_gear_component_periods`, and survive as **derived** values — the first period's start and the **last** period's end. They therefore still mean exactly what they always did ("when it first went on", "when it last came off", null while fitted), so every existing reader is unaffected: `evaluateGearServiceReminders`' "is it fitted now", `diffGearComponents`' `(type, brand, model, addedAt)` identity in the import script, and the card's Added/Retired columns.

Keeping the columns as a denormalised copy of the latest period was the alternative. It was rejected for the reason the distance totals are derived rather than stored: a second source of truth for the same fact drifts, and nothing downstream can tell that it has.

| | before | after |
| --- | --- | --- |
| Retire | sets `removedAt = now` | closes the open period at `now` (same guarded UPDATE) |
| Put a part back on | `PATCH { removedAt: null }`, reopening the original window | `POST …/components/:id/refit` opens period *n+1* at `now` |
| `PATCH { removedAt: null }` | the only way back, and lossy | kept — it now means precisely "this retirement never happened", reopening the last period |
| Entity | `addedAt`, `removedAt` | same two, derived, plus `periods: [{ addedAt, removedAt }]` |

The reported scenario resolves: `[2026-01-01, 2026-02-01)` + `[refit, ∞)` reads 10 km, and the 20 km ride of 2026-02-15 lands in the gap and is credited to nothing.

## Refit, not Unretire

The retired row's action is renamed and repointed. It stays **one click and unarmed**, exactly as #1568 argued — what changed is that a stray click is now cheap in both directions. A misclick fixed seconds later leaves a seconds-long gap no activity can fall in, so it still reads as an undo; a misclick fixed a week later under-credits that week, which is visible in the row and strictly safer than silently over-crediting with no way back.

Install history is surfaced by stacking the dates in the columns that already exist, one line per period, only when there is more than one. The two columns read as a pair — line *N* of Added and line *N* of Retired are the two ends of the same period. A component with one period, which is every row that exists today, renders identically to before. No new component, no new column, no change to the 480px snapping widths.

## Concurrency

`UNIQUE (componentId, installSequence)` is the guard, not decoration: two open periods would make the rollup count every later activity twice, which is a permanently wrong number nothing downstream can detect. Two concurrent refits read the same highest sequence and both try to claim the next one; one loses at the database, and `isUniqueConstraintError` maps that into the same "already fitted" answer the winner's second click would get.

**What the tests do and do not prove here** (corrected after review round 1): the two-refit test pins the *invariant* — one open period, one caller told "already fitted" — but it does **not** exercise the unique-constraint path. On SQLite knex serializes the two transactions, so the second one's own read already sees the first's committed insert and returns through the ordinary branch; deleting the whole `try/catch` leaves every test passing. The index itself is proved directly at raw-SQL level by the migration test, and the mapping from a violation to "already fitted" is `isUniqueConstraintError`, which has its own tests. A test forcing a genuine interleave would need a real concurrent backend and is not attempted.

It is deliberately a **plain** unique index rather than the partial `(componentId) WHERE removedAt IS NULL` that expresses "at most one open period" directly. `20260820000000_add_local_public_timeline_indexes.js` documents that knex emits a partial predicate on PostgreSQL and SQLite but **drops it** on MySQL-compatible clients — leaving a bare `UNIQUE(componentId)` that would forbid a second period outright and break refitting there forever.

Overlap is otherwise impossible by construction: periods are created only by component creation, by the backfill, and by a refit that opens at `now` when nothing is open; and PATCH reaches only the outermost bounds (first period's start, last period's end).

That prevents *overlap* but not *inversion*, which the first draft of this PR got wrong — see the review threads below. The route now checks each bound against the other end of the period it lands on; validating against the component's own derived pair compares bounds belonging to two different periods as soon as a part has been refitted.

## Rollup

The window moves from the component to the period; it is still one grouped query per chunk. The file join gains `p.id IS NOT NULL`, which is **not** decoration either: a component with no period row is a shape nothing creates, but without that clause both window tests read as TRUE against the missing row's NULLs and the component claims every activity on the gear. The countable predicate (`deletedAt IS NULL` + `completed` + `isPrimary`) is untouched, so the numbers still agree with `getFitnessActivitySummary`.

## Service reminders

`serviceDistanceMeters` and `lastAlertedDistanceMeters` stay **per component**, summed across its periods, because a chain's wear is cumulative across refits. `evaluateGearServiceReminders` needed no change.

It does fix a live bug for free. Today an unretire makes the total *jump* by the retroactive credit, and since reminders are evaluated on write, that jump fires a spurious crossing on the next processed activity. With periods the total does not move, so there is no crossing — pinned by a new test.

## The gear-level retire has no equivalent gap

Checked, because it was the obvious next question. `getFitnessGearDistanceRollups` filters on `actorId` + `gearId` + the countable predicate and never looks at `retiredAt`. A gear's attribution is an explicit per-activity `fitness_files.gearId`, not an inferred date window, so retiring and unretiring a bike cannot retroactively credit it anything. The lossiness was specific to components.

## Migration

`20260826120000_add_fitness_gear_component_periods.js` creates the table, backfills one period per existing component carrying the window it already had (soft-deleted rows included — a periodless component is the shape above), then drops the two columns. Both reference schema dumps are regenerated in this PR; the diffs are exactly this change and nothing else.

`down` restores the columns from the outermost bounds it can still represent. That is lossy for a refitted component, which is the information the old shape could not carry — deliberately so rather than silently.

## Browser verification

Against a worktree-local SQLite `dev.sqlite3` on `localhost:3210` — never a remote database — with a bike carrying one component added 2025-03-14.

| Check | Result |
| --- | --- |
| Retire (arm → confirm) | row dims, moves to "Show 1 retired component" |
| Retired row actions | **Refit** + Delete |
| Refit | row returns to installed, offers Retire again |
| Added / Retired columns after refit | two aligned lines: `Mar 14, 2025 / Aug 26, 2026` and `Aug 26, 2026 / —` |
| API after refit | `addedAt` = first period's start, `removedAt` = `null`, `periods` = both, 17s gap |
| Re-retire | closes the **second** period at the new now; the first period's bound is untouched |
| Single-period component | one line per column, unchanged from before |
| Dark mode | reads correctly |
| 375px | stacked dates stay inside the snapped column, pinned Type column intact |

## Review

Seven rounds of the sub-agent loop ran over this change — correctness, design-system, quality, scope-regression and testing lenses. Findings per round: **5, 4, 2, 1, 1, 2, 1**. Every finding is recorded as a resolved thread on this PR, including the two that were declined with reasons.

**Two consecutive fully clean rounds were never achieved**, so AGENTS.md's stopping condition was not met — the loop was stopped by an explicit decision after round 7, not by satisfying the rule.

Three bugs and one regression were found, and all four lived in the same place: the `addedAt`/`removedAt` handling of `PATCH /components/:componentId`, which no UI calls (the components card offers create, retire, refit and delete only). They shared one shape — a check that looked right because the case it was tested against was the one where it could not be wrong:

- the route validated against the component's DERIVED pair, correct only for a single-period component, and every fixture was single-period;
- the fix's per-bound SQL predicates were correct only for a refitted component, and that fix's test used a refitted one;
- the joint-write guard's null handling was covered by nothing, found by mutating it and watching the whole suite still pass.

The feature itself — periods table, derived pair, `/refit` — went seven rounds without a lens questioning its shape, and round 5's quality pass found nothing in it that could be deleted without losing a guarantee.

Several claims in this description and in code comments were corrected during the loop rather than left standing, including one about a test that did not test what it said.

## Testing

`yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test`, all green (890 files / 11,749 tests, re-counted after the review rounds).

Every new guard is **proved by reverting its subject**:

| Revert | Failures |
| --- | --- |
| refit reopens the closed period (the old behaviour) | 8, incl. the reported scenario, the reminder crossing, and the retire→refit→retire round trip |
| drop `p.id IS NOT NULL` from the rollup join | the periodless component claims the whole 15 km instead of 0 |
| derive `removedAt` from `periods[0]` instead of the last | 3 |
| drop `UNIQUE (componentId, installSequence)` | the second open period inserts cleanly |
| drop the migration's backfill | 2 migration tests |
| collapse `PeriodDates` to the derived pair | the multi-period row test |

New files: `lib/database/sql/fitnessGearComponentPeriods.test.ts` (the two properties needing a state the public `Database` interface cannot construct, on `getTestDatabaseWithInstance` so it honours `TEST_DATABASE_TYPE`) and `lib/database/sql/fitnessGearComponentPeriodsMigration.test.ts`.

## Docs

`AGENTS.md` and `CLAUDE.md` (Fitness Gear), `docs/fitness-file-storage.md` (Gear Tracking) and `docs/maintenance.md` (the gear import's `addedAt`/`removedAt` seed the first period and never a second).


app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts
```diff
@@ -61,16 +72,32 @@ export const PATCH = traceApiRoute(
...
+
+      const inverted = [0, lastIndex]
+        .filter((index) => index >= 0)
+        .some((index) => {
+          const { addedAt, removedAt } = boundsAfterWrite(index)
+          return addedAt !== null && removedAt !== null && removedAt <= addedAt
+        })
 
-      if (addedAt !== null && removedAt !== null && removedAt <= addedAt) {
+      if (inverted) {
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts]: **Finding (high, confirmed by correctness + scope-regression lenses):** the PATCH date-validation block checked `addedAt`/`removedAt` against the component's *derived* pair (`existing.addedAt`/`existing.removedAt`), which is the FIRST period's start and the LAST period's end. Once a part has been refitted those come from two different periods, so an inverting edit could slip through: on a component whose last period is `[2026-05-01, null)`, `PATCH { removedAt: 2026-02-15 }` was accepted against the derived `addedAt` of `2026-01-01` and wrote `[2026-05-01, 2026-02-15)` — a period no activity can ever satisfy, silently zeroing that stretch's distance behind a 200. `PATCH { addedAt }` alone was worse: checked against a derived `removedAt` of `null`, the guard short-circuited and accepted anything. Not reachable from the UI today (no date-edit affordance), but live on the public route, and every fixture in `route.test.ts` was single-period so nothing caught it.

**Fixed in d4d8e03c9:** each bound is now checked against the other end of the *period it lands on*, expressed per period (`boundsAfterWrite`) so the single-period case still sees both new values. Four tests added in `route.test.ts` — the two refusal tests fail against the old derived-pair check, the two acceptance tests pass under both, so they're not mirroring the implementation.

lib/database/sql/fitnessGear.ts
```diff
@@ -887,21 +1060,107 @@ export const FitnessGearSQLDatabaseMixin = (
...
+      // click would.
+      //
+      // Classified from the ERROR, through the same shared helper every other
+      // racing insert in this directory uses — not by re-reading the component
+      // and suppressing when it happens to look fitted. That re-read cannot
+      // tell this transaction's unique violation from any other failure inside
+      // it (a dropped connection, a lock timeout, a bug in either write), so a
+      // real fault landing while some other writer had just refitted the part
+      // would be reported as a tidy 404 and leave no trace.
+      if (!isUniqueConstraintError(error)) throw error
```
llun, line 1229: **Finding (medium, quality lens):** `refitFitnessGearComponent`'s catch block classified a failed transaction by re-reading the component and suppressing the error when it happened to look fitted. That can't distinguish its own unique violation from a dropped connection, a lock timeout, or a bug in either write — a real fault landing while another writer had just refitted the part would be reported as a tidy 404 with no trace. The repo already has `isUniqueConstraintError` (`lib/database/sql/utils/isUniqueConstraintError.ts`, its own tests, five existing callers in this directory using exactly this idiom).

**Fixed in d4d8e03c9:** replaced with `if (!isUniqueConstraintError(error)) throw error; return null`.

lib/database/sql/fitnessGear.test.ts
```diff
@@ -1010,6 +1010,382 @@ describe('FitnessGearDatabase', () => {
...
+      // open period, one caller told "already fitted" — not the unique index
+      // that protects it under a genuine interleave: on SQLite knex serializes
+      // the two transactions, so the second one's own read already sees the
+      // first's committed insert and returns through the ordinary branch. The
+      // index itself is proved directly, at the raw-SQL level, by
+      // `fitnessGearComponentPeriodsMigration.test.ts`'s "refuses a second open
+      // period per component"; the mapping from that violation to "already
+      // fitted" is `isUniqueConstraintError`, which has its own tests. Do not
+      // read a pass here as proof that the catch path ran.
+      it('opens exactly one period when two refits are issued together', async () => {
```
llun, line 1318: **Finding (medium, testing lens — proved it):** the two-refit race test's try/catch coverage claim was false. Deleting the entire try/catch in `refitFitnessGearComponent` left every test passing, and instrumenting the catch showed it never runs: on SQLite knex serializes the two transactions, so the second one's own read already sees the first's committed insert and returns through the ordinary branch. The test proves the invariant (one open period, one caller told "already fitted"), not the catch path.

**Fixed in d4d8e03c9:** renamed to "opens exactly one period when two refits are issued together" and the comment above it now states plainly what it does and does not prove, pointing at `fitnessGearComponentPeriodsMigration.test.ts`'s "refuses a second open period per component" for the unique index itself, and at `isUniqueConstraintError`'s own tests for the error-mapping. The PR body's claim was corrected too.

lib/database/sql/fitnessGear.ts
```diff
@@ -887,21 +1060,107 @@ export const FitnessGearSQLDatabaseMixin = (
...
+      // carried in as a subquery rather than left to the ownership read above.
+      // That read is a decision taken in front of the write, and a soft-delete
+      // landing between the two would otherwise let this close a period on a
+      // component nothing can reach again.
+      .whereIn(
+        'componentId',
+        database('fitness_gear_components')
+          .select('id')
+          .where('id', id)
+          .whereNull('deletedAt')
```
llun, line 1151: **Finding (low, quality + scope-regression lenses):** `retireFitnessGearComponent`'s state-changing UPDATE moved to the periods table, which has no `deletedAt` of its own — so the component's soft-delete check was left to the ownership read in front of the write (check-then-act), while the comment directly above the UPDATE still claimed a predicate on the write itself. AGENTS.md has an explicit rule against exactly this pattern.

**Fixed in d4d8e03c9:** the component's `deletedAt IS NULL` is now carried into the UPDATE as a `whereIn` subquery against `fitness_gear_components`, so the write itself is the guard again. No result-based test can observe the race (the trailing re-read answers null either way and the interleave can't be forced in-process), so it's pinned by a query-shape test in `fitnessGearComponentPeriods.test.ts`, following the precedent of `localPublicQueryShape.test.ts`.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
```diff
@@ -70,6 +70,51 @@ const CELL_WRAP = 'wrap-anywhere'
...
+          {/* Which line belongs to which period is carried by POSITION, and
+              position is exactly what a screen reader drops: it reads out every
+              Added date, then every Retired date, with nothing saying that the
+              second of one goes with the second of the other. Numbering the
+              lines restores the pairing, and only where there is a pairing to
+              restore — a component with one period, which is every part that
+              has never been refitted, announces exactly the bare date it
+              always did. */}
+          {periods.length > 1 && (
+            <span className="sr-only">{`Install ${index + 1}: `}</span>
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:107]: **Finding (low, design-system lens):** stacking one line per period made the Added/Retired correspondence positional only, which a screen reader drops — it reads every Added date, then every Retired date, with nothing saying which belongs to which. Newly introduced by this PR.

**Fixed in d4d8e03c9:** each line now carries an `sr-only` "Install N:" prefix, only when there's more than one period, so a single-period row still announces exactly the bare date it always did. Two tests added.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts]: Addressed in d4d8e03c9 (already on this PR before this comment was posted): the check is now expressed per period via `boundsAfterWrite`, so it validates each bound against the other end of the period it actually lands on rather than the derived first/last pair. Four tests added in `route.test.ts` covering both the refusal and acceptance cases. Resolving.

lib/database/sql/fitnessGear.ts
llun: Addressed in d4d8e03c9: the catch block now classifies the failure with `isUniqueConstraintError(error)` and rethrows anything else, matching the five other callers in this directory. Resolving.

lib/database/sql/fitnessGear.test.ts
llun: Addressed in d4d8e03c9: renamed to "opens exactly one period when two refits are issued together" with a comment stating plainly what it proves (the invariant) versus what it does not (the catch path), pointing at the migration test for the unique index and at `isUniqueConstraintError`'s own tests for the mapping. The PR body's claim was corrected too. Resolving.

lib/database/sql/fitnessGear.ts
llun: Addressed in d4d8e03c9: the component's `deletedAt IS NULL` is now carried into the UPDATE itself as a `whereIn` subquery against `fitness_gear_components`, closing the check-then-act gap. Pinned by a query-shape test in `fitnessGearComponentPeriods.test.ts` since no result-based test can observe the race. Resolving.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.tsx:107]: Addressed in d4d8e03c9: each period line now carries an `sr-only` "Install N:" prefix, only rendered when there is more than one period, so the single-period case still announces the bare date it always did. Two tests added. Resolving.

llun:
Two suggestions from round 1's quality lens were considered and deliberately **not** actioned:

1. **Fold `updateFitnessGearComponent`'s two SELECT+UPDATE pairs into subquery UPDATEs.** Declined — `UPDATE t ... WHERE id = (SELECT ... FROM t ...)` is MySQL error 1093 (cannot select from the table being updated in a subquery), and AGENTS.md says MySQL-compatible knex paths should not be broken casually. The two-step form is portable, and everything already runs inside one transaction so no race is introduced.

2. **Extract the `component()`/`createComponent()` fixture builder, now duplicated across six test files, into a shared helper.** Declined as scope creep for this PR — the duplication predates it (each file already had its own builder), and a shared helper spanning the `app/` and `app/api/` test trees is its own change.

This is a non-blocking note, not a thread requiring resolution.

lib/services/fitness-gears/gearRequests.ts
```diff
@@ -180,6 +180,72 @@ export const UpdateFitnessFileGearRequest = z.object({
...
+ * opens at `now`, later than the period it just closed, so a new `addedAt` that
+ * clears its own period's end clears every later period too, and symmetrically
+ * for the last period's `removedAt`.
+ *
+ * `changes` uses the same presence semantics as the update params — an absent
+ * key leaves that bound alone.
+ *
+ * Returns an error message for the 422 body, or null when the edit is ordered.
+ */
+export const getComponentPeriodBoundsError = (
```
llun, line 211: **Finding (medium, quality lens; the same latent crash was reached independently by the scope-regression lens):** the inversion guard round 1 added to this route carried `.filter((index) => index >= 0)`, which read as protection against an empty `periods` array but was not: it removes only the `-1` entry, so index `0` stays and `boundsAfterWrite(0)` still dereferences `undefined` on a component with no periods — a 500 rather than a graceful answer. Separately, the route's own comment claimed parity with the gear PATCH one directory up, which delegates to the named, independently testable `getGearKindFieldError`, while this stayed an anonymous closure tested only through route integration tests.

lib/database/sql/fitnessGearComponentPeriods.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      sql.startsWith('update fitness_gear_component_periods')
+    )
+    expect(update).toBeDefined()
+    expect(update).toContain('removedat is null')
+    // The guard, as a predicate on the write itself — asserted structurally
+    // rather than by its spelling. An equivalent `whereExists` with
+    // `select(1)`, or a different order of the subquery's own clauses, is the
+    // same guard, and a shape test that fails for those is a test of the
+    // author's phrasing rather than of the behaviour.
+    expect(update).toContain('from fitness_gear_components')
```
llun, line 160: **Finding (medium, testing lens — demonstrated it):** this asserted `toContain('select id from fitness_gear_components')`, which pins the projection and the `whereIn` spelling rather than the guard. Substituting a semantically identical `whereExists` with `select(1)` — the guard fully intact — fails the test. That is a test of the author's phrasing.

AGENTS.md
```diff
@@ -773,11 +773,64 @@ it; there is no legacy shape left to copy.
...
+  activity can fall inside, which drops its distance silently and for good. The
+  bug reads as correct because for a single-period component the first and last
+  period are the same row and the two formulations coincide.
+- **Periods must never overlap, and `UNIQUE (componentId, installSequence)` is
+  what enforces it against a race.** The rollup joins the periods and sums over
+  them, so an activity inside two of them is counted twice — a permanently wrong
+  number nothing downstream can detect. Two concurrent refits both read the
+  highest sequence and both try to claim the next one; the unique index lets
+  exactly one win, and the loser reports "already fitted". The loser identifies
+  itself from the ERROR, through the shared `isUniqueConstraintError` — never by
```
llun, line 835: **Finding (low, scope-regression lens):** this bullet still read "the loser re-reads and reports 'already fitted'" — describing the mechanism round 1 REMOVED, fifteen lines above the new code comment explaining why re-reading was wrong. Doc drift introduced by round 1 itself, which edited the adjacent bullet and missed this one.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
```diff
@@ -226,6 +237,71 @@ describe('GearComponentsCard', () => {
...
+        ]
+      })
+    ])
+
+    // Once per column, so each date is announced with the install it belongs to.
+    // Anchored, so `Install 1:` does not also match `Install 10:` the day a
+    // fixture grows that far — and anchored WITHOUT the trailing space the
+    // element actually renders, because the default matcher normalizes
+    // whitespace before comparing.
+    expect(screen.getAllByText(/^Install 1:$/)).toHaveLength(2)
```
Comment (llun) [inline app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:293]: **Finding (low, testing lens):** `getAllByText(/Install 1:/)` is unanchored, so it would also match "Install 10:" in a fixture with ten or more periods.

lib/services/fitness-gears/gearRequests.ts
llun: Fixed in 576885ff6: extracted as `getComponentPeriodBoundsError` in `lib/services/fitness-gears/gearRequests.ts`, beside `getGearKindFieldError`, which makes the comment's claim true. It answers null for a component with no periods — the way the write itself no-ops on one — instead of indexing into an empty array. Ten unit cases added in `gearRequests.test.ts`, including the middle-period case the route can never reach and the single-period collapse where both bounds land on one row; the two refusal cases fail against the old derived-pair formulation. Resolving.

lib/database/sql/fitnessGearComponentPeriods.test.ts
llun: Fixed in 576885ff6: loosened to the structural `toContain('from fitness_gear_components')` plus the existing `toContain('deletedat is null')`. Still fails against a genuine removal of the guard (re-proved), but now survives an equivalent rewrite. Resolving.

AGENTS.md
llun: Fixed in 576885ff6: rewritten to say the loser identifies itself from the ERROR through `isUniqueConstraintError`, and why re-reading is not an acceptable substitute. Resolving.

app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx
Comment (llun) [reply app/(timeline)/fitness/gear/[id]/GearComponentsCard.test.tsx:293]: Fixed in 576885ff6: anchored. Worth recording that the first attempt anchored it as `/^Install 1: $/`, including the trailing space the element actually renders, and that failed — the default matcher normalizes whitespace before comparing, so the correct anchor is `/^Install 1:$/`. Noted in the test's comment. Resolving.

llun:
**Round 2 — items considered and not actioned (non-blocking):**

- The quality lens preferred `whereExists` over `.whereIn(col, subquery)` in `retireFitnessGearComponent`, since the same file uses `whereExists` in `assignFitnessFileGearIfUnset`. **Declined:** both are in-repo idioms (the correctness lens cited `collection.ts` for the `whereIn` subquery form), the `whereIn` here needs no correlation clause, and with the query-shape test now loosened it no longer pins either spelling — so this is a preference between two working idioms, not a defect.
- The quality lens flagged the `normalize`/`capture` helper in `fitnessGearComponentPeriods.test.ts` as a third near-copy. **Declined:** two copies already exist and `localPublicQueryShape.test.ts`'s own comment explains it keeps the helper local deliberately. Following that precedent is the smaller change.
- The testing lens confirmed `refitFitnessGearComponent`'s `isUniqueConstraintError` wiring is reached by NO test: on SQLite knex serializes the two transactions so the catch never runs, and CI pins `TEST_DATABASE_TYPE=sqlite`. This is a standing, disclosed gap — already stated in the test's comment and in the PR description. Not closed, and deliberately not overstated: the classifier has its own unit tests and the unique index is proved directly by the migration test, but the wiring between them is unexercised.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts
```diff
@@ -61,20 +64,21 @@ export const PATCH = traceApiRoute(
...
-          : (existing.removedAt ?? null)
-
-      if (addedAt !== null && removedAt !== null && removedAt <= addedAt) {
+      // `parsed.data` is handed over whole rather than rebuilt: the schema's
+      // `.nullish()` sits OUTERMOST, so a key absent from the body is absent
+      // here too and an explicit null arrives present-and-null — which is
+      // exactly the presence the check reads. Reconstructing that with
+      // conditional spreads only restates it, and the fields it would leave out
+      // are ones the check never looks at.
+      const boundsError = getComponentPeriodBoundsError(
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts:73]: **Finding (low, quality lens):** the route rebuilt a `changes` object with conditional spreads to preserve key presence, which `parsed.data` already carries — the premise `optionalText`'s own doc comment states and the same route relies on elsewhere via `'field' in parsed.data`. The spread performed no conversion, unlike the one below it that turns epochs into `Date`.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts
```diff
@@ -238,6 +256,129 @@ describe('Fitness gear component item API', () => {
...
+          method: 'PATCH',
+          headers: { Origin: 'https://llun.test' },
+          body: JSON.stringify({ removedAt: Date.UTC(2026, 1, 15) })
+        }),
+        params
+      )
+      const data = await response.json()
+
+      expect(response.status).toBe(422)
+      expect(data.error).toBe('removedAt must be after addedAt')
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts:307]: **Finding (low, testing lens):** this refit refusal test, and its mirror at line 329 ("refuses an added date that would invert the first period of a refitted component"), assert the 422 status but not its body, unlike the single-period case beside them — so the error string for the refit scenarios was pinned only at the unit level.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts
```diff
@@ -238,6 +256,129 @@ describe('Fitness gear component item API', () => {
       expect(mockDb.updateFitnessGearComponent).toHaveBeenCalled()
     })
 
+    // A refitted component, whose derived pair comes from two DIFFERENT periods:
```
Comment (llun) [inline app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts:259]: **Finding (quality lens, declined):** these route-level inversion tests — the two refit refusal cases below and the two acceptance cases that follow them, through "accepts moving both ends of a single-period component past the old window" — largely duplicate the new unit cases added in `lib/services/fitness-gears/gearRequests.test.ts` for `getComponentPeriodBoundsError`, since they share the same scenario values. Recommend trimming to avoid two copies of the same coverage.

lib/services/fitness-gears/gearRequests.test.ts
```diff
@@ -371,3 +372,121 @@ describe('getGearKindFieldError', () => {
     expect(getGearKindFieldError(kind, fields)).toBeNull()
   })
 })
+
+describe('getComponentPeriodBoundsError', () => {
+  // Only the two `refitted` refusals below pin the bug this function was
```
llun, line 377: **Finding (low, scope-regression lens):** round 3 added this comment saying only the two `refitted` refusal cases pin the extracted bug. That claim is TRUE — verified by hand-evaluating every case against the reconstructed derived-pair formula — but the comment's stated REASON ("a single-period component makes the two formulations agree by construction") accounts only for the single-period cases; the multi-period cases that also agree do so for their own reasons the comment did not give.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/components/[componentId]/route.ts:73]: Fixed in c9e44e78c: `parsed.data` is passed whole. Round 4's correctness lens then verified the presence equivalence empirically against the real Zod version rather than trusting the comment, and established that the one input shape where the two forms could diverge — an explicit `key: undefined` own-property — is unreachable through `req.json()`, since `JSON.parse` never produces one. Resolving.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts:307]: Fixed in c9e44e78c: both now assert `data.error`. Round 4's testing lens proved these are not vacuous by mutating the route's 422 body and confirming they fail. Resolving.

app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts
Comment (llun) [reply app/api/v1/fitness/gear/[id]/components/[componentId]/route.test.ts:259]: Declined, on evidence: the testing lens revert-proved that each route test also proves what the unit test cannot — that the route wires `existing.periods` and the parsed body into the check, maps the result to the right HTTP status, and leaves `updateFitnessGearComponent` uncalled on refusal. That's layered coverage, not duplication: the unit tests in `gearRequests.test.ts` pin the bounds logic itself, these pin the route's wiring of it. Two lenses disagreed here and the one that ran the proofs (testing) was followed. Keeping these tests as-is; resolving as wontfix.

lib/services/fitness-gears/gearRequests.test.ts
llun: Fixed in 34f43eb81: the comment now distinguishes the two groups — the single-period cases where the formulations agree by construction, and the other multi-period cases which agree for their own separate reasons (an edit the correct check accepts anyway, or a bound neither formulation looks at). Round 4's testing lens independently confirmed the underlying claim by reverting the function and observing exactly 2 of 10 cases fail. Resolving.

llun:
**Rounds 3-4 whole-PR checks (non-blocking summary, no thread to resolve):**

- Whole-PR file sweep: all 26 files in `git diff origin/main...HEAD` belong to the stated feature; no scratch files, `.env.local`, `dev.sqlite3` or agent revert-proof edits are committed.
- The branch is 4 commits behind `origin/main` and reports MERGEABLE. Deliberately NOT rebased: `main` advances after every feature PR here (a version-bump PR merges each time), so BEHIND is the normal state, and the remote branch carries no commits the author did not write. Say if you'd prefer it rebased before merge.
- Suite re-counted at 890 files / 11,743 tests; the PR description's stale 11,724 has been corrected.
- Standing disclosed gap, unchanged: `refitFitnessGearComponent`'s `isUniqueConstraintError` wiring is exercised by no test, because SQLite serializes the two racing transactions and CI pins `TEST_DATABASE_TYPE=sqlite`. The classifier has its own unit tests and the unique index is proved directly by the migration test; the wiring between them is not. Round 4's testing lens separately confirmed the route's presence-semantics assumption IS covered, by three pre-existing tests, found by patching Zod to materialize absent keys as `undefined`.

lib/database/sql/fitnessGear.ts
```diff
@@ -847,16 +979,127 @@ export const FitnessGearSQLDatabaseMixin = (
...
+              'id',
+              first.id
+            )
+            // Clearing a bound cannot invert anything.
+            const addedAt = params.addedAt
+            if (addedAt) {
+              query.where((builder) =>
+                builder
+                  .whereNull('removedAt')
+                  .orWhere('removedAt', '>', addedAt)
```
llun, line 1069: FINDING 5.1 (HIGH) — fixed in 16a27da5.

`updateFitnessGearComponent` carried no ordering check of its own: it resolved which period was first/last at WRITE time and wrote whatever date it was handed, while the route validates the periods it READ. A refit landing between the two appends a new last period, so the bound lands on a row nobody checked. Example: a component closed at 2026-03-01, refitted today, sent `PATCH { removedAt: 2026-02-15 }` — wrote the second period as `[today, 2026-02-15)`, a range no activity can satisfy, and a later refit buries it in the middle where nothing re-examines it.

Fixed: each bound write now carries an SQL ordering predicate against the other end of the row it lands on (see the `addedAt`/`removedAt` branches below). Reachable with no interleave at all, since the write had no check of its own — the regression test needs no concurrency.

lib/database/sql/fitnessGear.ts
```diff
@@ -847,16 +979,127 @@ export const FitnessGearSQLDatabaseMixin = (
...
+        const last = orderedPeriods[orderedPeriods.length - 1]
+
+        // Both bounds land on the SAME row when the component has one period,
+        // and then that row's stored state says nothing about whether the edit
+        // is ordered — both ends are being replaced. Writing them separately
+        // makes the first write's predicate compare the new `addedAt` against
+        // the OLD `removedAt` the second write is about to overwrite, so moving
+        // a window wholesale applied only half of itself and left the other
+        // half silently dropped behind a 200.
+        const writesBothBoundsOnOneRow =
```
llun, line 1020: FINDING 6.1 (HIGH) — fixed in e899da178, self-caught then confirmed by the round-6 correctness lens via the same trace.

Round 5's per-bound predicates read the row's STORED state, and both bounds land on the SAME row for any component that has never been refitted — every component in production. `PATCH { addedAt: Apr 1, removedAt: Sep 1 }` on `[Jan 1, Mar 1)` ran the `addedAt` write first, whose predicate compared Apr 1 against the stored `removedAt` of Mar 1 and refused, while the `removedAt` write landed. Result `[Jan 1, Sep 1)`: half the edit applied, half silently dropped behind a 200. Round 5's own test missed it because it used a refitted component, where the bounds land on different rows and never interact.

Fixed: when both bounds land on one row (`writesBothBoundsOnOneRow`), they are written together in a single statement with the ordering settled in code, since nothing stored bears on it. The per-bound SQL predicates remain for the different-rows case, which is where they were needed.

lib/database/sql/fitnessGear.ts
```diff
@@ -847,16 +979,127 @@ export const FitnessGearSQLDatabaseMixin = (
...
+          // rewrite is not equivalent: it skips the write whenever a bound is
+          // cleared, so clearing one while moving the other silently applies
+          // neither, behind a 200.
+          //
+          // The skip is a floor, not the guard callers should rely on: it
+          // refuses an inverted pair rather than writing one, but says nothing
+          // about it. The route answers 422 first, through the shared
+          // `getComponentPeriodBoundsError`, and a future caller reaching this
+          // method some other way needs to do the same.
+          if (!addedAt || !removedAt || removedAt > addedAt) {
```
llun, line 1044: FINDING 7.1 (medium) — fixed in b82325926.

Found by mutation testing, which is why earlier rounds' revert-proofs missed it. Rewriting `!addedAt || !removedAt || removedAt > addedAt` as the plausible-looking `addedAt && removedAt && removedAt > addedAt` passed the entire suite. They are not equivalent: the rewrite skips the write whenever either bound is cleared, so clearing one while moving the other silently applies neither — the same failure shape as 5.1 and 6.1.

Fixed: three cases added in `fitnessGear.test.ts` for a single-period component clearing one bound while moving the other, and clearing both — all of which fail against that mutation — plus a twice-refitted-component case pinning that the joint branch does NOT fire and the middle period comes through untouched. The comment above the guard now states why it is two negations, and that the skip is a floor rather than the guard callers should rely on (the route answers 422 first).

AGENTS.md
```diff
@@ -773,11 +773,78 @@ it; there is no legacy shape left to copy.
...
+  from making two periods OVERLAP. It does not on its own keep a period from
+  being INVERTED, and the route has to check each bound against the other end
+  of the period it lands on to do that: the component's own `addedAt` and
+  `removedAt` are derived from two DIFFERENT periods once a part has been
+  refitted, so validating the request against that derived pair compares bounds
+  that do not belong together and lets `[May 1, Feb 15)` through — a period no
+  activity can fall inside, which drops its distance silently and for good. The
+  bug reads as correct because for a single-period component the first and last
+  period are the same row and the two formulations coincide.
+  **The route's check is not sufficient on its own, and the write guards its
```
llun, line 815: FINDING 6.2 (low) — fixed in ac2788b78.

`AGENTS.md`, `CLAUDE.md` and `docs/fitness-file-storage.md` all said "each bound write carries its own ordering predicate" universally — true when written, one commit earlier, but no longer once the joint-write path (6.1's fix) landed. Third time in this loop a fix left its own documentation behind it.

Fixed: all three now state which guard applies where — per-bound SQL predicate when the bounds land on different rows, joint write with ordering settled in code when they land on the same row — and why getting that split backwards is what produced the half-applied edit (6.1).

lib/database/sql/fitnessGear.ts
llun: Fix already pushed as part of the same push (16a27da5). Resolving.

lib/database/sql/fitnessGear.ts
llun: Fix already pushed as part of the same push (e899da178). Resolving.

lib/database/sql/fitnessGear.ts
llun: Fix already pushed as part of the same push (b82325926, current head). Resolving.

AGENTS.md
llun: Fix already pushed as part of the same push (ac2788b78). Resolving.

llun:
**Review loop closing summary (rounds 1-7)**

Seven rounds. Findings per round: 5, 4, 2, 1, 1, 2, 1. Two consecutive fully clean rounds were never achieved, so `AGENTS.md`'s stopping condition was not met; the loop was stopped by an explicit decision after round 7, not by satisfying the rule. Stating that plainly rather than implying convergence.

Every defect found in rounds 5-7 was in ONE surface: the `addedAt`/`removedAt` handling of `PATCH /components/:componentId`, which no UI calls (the components card offers create, retire, refit and delete only). The feature itself — periods table, derived pair, `/refit` — went seven rounds without a lens questioning its shape, and round 5's quality pass found nothing in it that could be deleted without losing a guarantee.

The maintainer was offered the option of rejecting `addedAt`/`removedAt` on that PATCH until a UI needs them, which would delete the whole class of bug, and chose to keep the capability. Recorded here because it was a deliberate call, not an oversight.

Standing disclosed gap, unchanged: `refitFitnessGearComponent`'s `isUniqueConstraintError` wiring is exercised by no test — SQLite serializes the racing transactions and CI pins `TEST_DATABASE_TYPE=sqlite`. The classifier has unit tests and the unique index is proved by the migration test; the wiring between them is not.

Final suite: 890 files / 11,749 tests.

PR #1592, opened 2026-08-27
fix: normalize local actor usernames and fold casing on every lookup

llun:
## The bug

A local actor's `actors.id` **is** `https://<domain>/users/<username>` — `getLocalActorId` interpolates the username straight into it, and every local status id is `${actorId}/statuses/${n}`. So username casing was an identity question, not a cosmetic one, and nothing folded it.

Confirmed against a fresh database before writing any fix:

- `createAccount({ username: 'Alice' })` then `getActorFromUsername({ username: 'alice' })` returned **null**. So `acct:alice@host` 404'd at WebFinger for a user who registered as `Alice` — taking `/@alice`, mentions (which resolve through WebFinger), `/api/v1/accounts/lookup`, both search routes and `authorize_interaction` with it.
- `Alice` and `alice` could **both** register, becoming two distinct actors with two distinct ActivityPub ids. `isUsernameExists` and the `actors_username_domain_unique` index are both case-sensitive.

Emails already solved exactly this through `normalizeEmail`, and `domain` is lowercased everywhere. `username` was the one identifier left raw.

## The fix

**Mint side.** `normalizeUsername` (`lib/utils/normalizeUsername.ts`, the sibling of `normalizeEmail` and as deliberately minimal — trim then lowercase) runs in three layered places, the same way email normalization does: `localUsernameSchema`, `registerAccount`, and `createAccount`/`createActorForAccount`. The last is what makes it unbypassable — it is the only place a local actor row is written — and it normalizes into one variable that feeds both the `username` column and `getLocalActorId`, so the column and the id cannot drift.

The fold in the schema runs **before** the reserved-name refine, and that ordering is load-bearing: `isFederationSigningActorUsername` is a case-sensitive `startsWith('__instance__')`, so `__INSTANCE__` passed the check and minted a confusable neighbour of the instance actor at a *different* id. It is also before `.max()`, since a fold can change a string's length (`İ` folds to two code units).

**Lookup side.** `findActorRowByUsername` (`lib/database/sql/utils/usernameMatch.ts`) backs both `getActorFromUsername` and `isUsernameExists`. It is **exact-match first, then folded — not a single `lower(username) = ?`**, for two independent reasons:

1. **Correctness.** SQL `lower()` and JS `toLowerCase()` are not the same function. SQLite's builtin folds ASCII only, so a remote actor named `Фёдор` has its stored value fold to `Фёдор` while the JS side folds the identical input to `фёдор` — a fold-only query would stop finding a row that resolves today. Trying exact first means no lookup that works now can break, on any backend, for any charset.
2. **Precedence.** Existing local actors that differ only by case are deliberately **not** migrated (their ids are already federated), so an instance can hold both `Alice` and `alice`, and `/@Alice` must not start resolving to `alice`. "Exact wins" says that without a tiebreak clause. The folded arm orders by `createdAt`, `id` so an unmatched casing resolves to whoever claimed the name first rather than to whatever the index yields.

`isUsernameExists` folding is what refuses a new `alice` beside an existing `Alice`.

**A third case-sensitivity site.** `OnlyLocalUserGuard` fronts the *entire* ActivityPub surface — actor document, inbox, outbox, followers, following, statuses, collections — and resolved by rebuilding the actor id from the URL path segment, which could only ever match one spelling. It answered `/api/users/Alice` and 404'd `/api/users/alice` while every human-facing surface folded. It now resolves by username; the host binding is unchanged, since matching `domain` against `headerHost` is the same constraint as requiring the id to have been minted on this host.

## Migration

`20260826120000_add_actors_lower_username_index.js` adds the functional index the folded arm reads, `(lower(username), domain)`. Without it the folded arm sequentially scans `actors` on exactly the requests least able to afford it — an unknown handle (404 traffic), the first lookup of a remote actor, account search — on a table that grows with every remote account the instance has ever seen.

It is **not unique**: an instance may already hold a case-colliding pair, and a unique index would refuse to build there. Collision is refused at the application layer instead.

**MySQL is skipped in both halves** — no index, and the folded arm never runs there. Its default collations (`utf8mb4_0900_ai_ci` and friends) already fold, and so does `actors_username_domain_unique`, which means a colliding pair cannot exist on MySQL at all; a second query would find nothing the first did not and would find it by scanning. The DDL is also the least portable of the three (`((lower(username)), domain)`, no `IF NOT EXISTS` on `CREATE INDEX`, backtick quoting, functional indexes only from 8.0.13, and never in MariaDB, which the `mysql2` client also reaches). An operator who puts a `_bin`/`_cs` collation on `actors` gets case-sensitive usernames there — the behaviour that backend had before this change, not a new regression.

Both schema dumps regenerated against fresh local databases; each differs by exactly the new index.

## Scope deliberately not taken

**Existing rows are left alone.** Lowercasing stored local usernames would only be meaningful if the ids moved too, and `actors.id` is referenced as a plain string by 49 `actorId` + 10 `targetActorId` columns, by every local status id, by the `settings` JSON, and by `${actorId}#main-key` for signature verification. Remote servers hold those ids and there is no ActivityPub mechanism to announce a rename (`Move` is a user-initiated account migration, not an id rewrite). The case-insensitive lookup makes those actors reachable by any casing, which is the actual user-visible fix.

**Remote usernames are stored verbatim.** A remote server mints its own ids and chooses its own casing; we fold when matching, never when persisting. WebFinger answers with the *stored* casing, so a client that echoes `subject` back gets the canonical handle.

## Verification

Gate green in order: `prettier` → `lint` → `typecheck` → `build` → `test` (**890 files / 11,750 tests**), re-run after rebasing onto `main`.

Index usage confirmed on a throwaway local PostgreSQL 17 seeded to 50k actor rows and on SQLite, using SQL captured from knex rather than hand-rebuilt — both arms are index scans with **both** columns as index conditions.

Every part of the fix is pinned by a test that fails when it is reverted:

| Revert | Failing tests |
| --- | --- |
| Drop the exact-match arm | 2 (non-ASCII lookup, exact-case precedence) |
| Exact-only (pre-fix behaviour) | 12 |
| Move the fold after the reserved-name refine | 3 |
| Index the raw column instead of `lower()` | 1 |

The index migration is pinned by reading the index definition out of `sqlite_master`, because no result-based test can see it — every functional test passes against a table with no index at all, and `PRAGMA index_info` reports a null column name for an expression.

## Known gap, not introduced here

`POST /api/v1/actors` pre-checks `isUsernameExists` and then calls `createActorForAccount` with no unique-violation catch, so a concurrent duplicate surfaces as a 500 rather than a 4xx — the TOCTOU pattern `REVIEW.md` flags. `registerAccount` already handles this correctly. Pre-existing and unrelated to casing, so it is left for its own change.

---

## Review loop

Seven rounds of the sub-agent review loop ran on this PR. **24 findings, all posted as inline threads, all addressed, replied to and resolved (0 unresolved).** Round 7 surfaced nothing, which is what closes the loop.

| Round | Findings | Nature |
| --- | --- | --- |
| 1 | 8 | 2 behavioural, 6 prose |
| 2 | 4 | 1 behavioural regression, 3 prose |
| 3 | 4 | 1 latent regression, 3 prose |
| 4 | 3 | prose only |
| 5 | 2 | prose only |
| 6 | 3 | 2 coverage gaps (mutation audit), 1 prose |
| 7 | 0 | clean |

Every finding was in this PR's own work, and most were in the *previous round's fix* — which is the argument for pointing each round at the last round's diff rather than re-reading the original.

**The two that mattered most were defects the fix itself introduced.** Resolving actors by username instead of by rebuilding the id widened what a URL can match, and that had consequences in both directions: the folded lookup arm trimmed its input, so `/users/%20alice%20` served a whole actor surface; and the reserved-name check first reserved the entire `__instance__` prefix, which would have de-federated a legacy `__instance__archive` account — 404ing an actor document and inbox that work on `main`. Neither was visible reviewing the first commit.

**Round 6's mutation audit was the highest-yield pass**: 24 mutations against every behavioural guarantee, 22 caught, 2 survived — `getMastodonActorFromUsername`'s fold and the deliberate decision *not* to fold `domain` were both promises with no test behind them. Both are now covered, verified by reproducing the exact mutations (0 failures before, 6 after).

Three tests in this PR were caught being **vacuous** before merge — two squatter fixtures that never reached the branch under test, and a drift pin whose own comment named a scenario none of its fixtures exercised. Each was rewritten and re-proven by mutation.

Verification standard used throughout: every fix is revert-proofed (reverting it must fail a named test), and every quantitative claim in code comments or docs was re-derived by construction rather than by reading — one call-site count in a comment was wrong precisely because it was produced by a buggy script and then trusted.


lib/database/sql/utils/usernameMatch.ts
```diff
@@ -0,0 +1,66 @@
...
+  { username, domain }: { username: string; domain: string }
+): Promise<SQLActor | undefined> => {
+  const exactMatch = await database<SQLActor>('actors')
+    .where('username', username)
+    .andWhere('domain', domain)
+    .first()
+  if (exactMatch || isMySQLClient(database)) return exactMatch
+
+  return database<SQLActor>('actors')
+    .whereRaw('lower(??) = ?', ['username', normalizeUsername(username)])
```
llun: **[correctness] The folded arm trims the input, so this folds whitespace as well as case.**

`normalizeUsername` is `trim().toLowerCase()`, so the folded arm compares a *trimmed* input against an untrimmed `lower(username)`. Confirmed by test: `getActorFromUsername({ username: '  alice  ' })` resolves `alice`, where it returned null before this PR.

`GET /users/%20alice%20` percent-decodes to `' alice '`, the exact arm misses, and the folded arm matches — so `OnlyLocalUserGuard` serves alice's full actor document, `/inbox`, `/outbox` and `/followers` at an unbounded family of URLs. That matters beyond aesthetics because the outbox root sends `Cache-Control: public, max-age=60, s-maxage=60`, so each spelling is a distinct shared-cache key for one identical body. It is also asymmetric: the input is trimmed but the column is not, so a lookup for `alice` still cannot find a stored `'alice '`.

Every comment here, plus AGENTS.md and REVIEW.md, describes this arm as folding *casing*. Fix: use `username.toLowerCase()` in the folded arm and keep `normalizeUsername` (trim included) for the MINT paths, where trimming is correct and mirrors `normalizeEmail`.

lib/services/guards/OnlyLocalUserGuard.ts
```diff
@@ -36,8 +35,21 @@ export const OnlyLocalUserGuard =
...
+    // mentions and account lookup now do. Rebuilding the id could only ever
+    // match one spelling, which left the entire ActivityPub surface — actor
+    // document, inbox, outbox, followers, following, statuses, collections —
+    // answering `/api/users/Alice` and 404ing `/api/users/alice`.
+    //
+    // The host binding is unchanged: matching `domain` against `headerHost` is
+    // the same constraint as requiring the id to have been minted on this host,
+    // because both are written from one value in `createAccount` /
+    // `createActorForAccount`. A remote actor still cannot match it.
+    const actor = await database.getActorFromUsername({
```
llun, line 53: **[correctness] `/users/__instance__` can now resolve to a user-owned actor.**

The folded arm makes this guard resolve `__instance__` to any local actor whose username folds to it. Preconditions are both reachable: a legacy account actor named `__INSTANCE__` (registerable before this PR — the reserved-name refine was a case-sensitive `startsWith`, which this PR's own `localUsername.ts` comment documents as having been live), and no `username = '__instance__'` row yet, i.e. `getFederationSigningActor()` has not run on that domain.

Confirmed by test: with a legacy `__INSTANCE__` row present, `getActorFromUsername({ username: '__instance__' })` returns it.

In that state `GET /users/__instance__` returns 200 with the squatter's Person document — `actor?.account` is truthy, so it passes even without `allowFederationSigningActor` — where it previously 404'd. That URI is exactly `getFederationSigningActorId(domain)`. Impact is bounded (every response body carries the squatter's own canonical `id`, and `isFederationSigningActor` stays false so no relay handshake is processed), so this is confusability at a reserved URI rather than privilege escalation — but it is a state the old id-rebuild could not reach.

Fix: refuse in this guard when the requested segment folds to a reserved username but the resolved actor is not the genuine federation signing actor.

AGENTS.md
```diff
@@ -167,6 +167,19 @@ ActivityPub objects are **JSON-LD**, so the same logical object can arrive in ma
...
 
+## Actor Usernames Are Case-Insensitive
+
+A local actor's username is not a label on its identity, it **is** its identity: `getLocalActorId` builds `https://<domain>/users/<username>` and every ActivityPub id downstream is derived from that string (`getLocalStatusId` is literally `${actorId}/statuses/${n}`). So casing is a correctness question on two separate axes, and both are handled.
+
+- **Every local username this instance MINTS is lowercased**, by `normalizeUsername` (`lib/utils/normalizeUsername.ts`, the sibling of `normalizeEmail` and deliberately as minimal: trim then lowercase). It runs in three places, layered the same way email normalization is: `localUsernameSchema` (both creation routes), `registerAccount` (so a direct service call cannot bypass the schema), and `createAccount`/`createActorForAccount` (the one place a local actor row is written, which is what makes it unbypassable). The DB layer normalizes into a local variable that feeds BOTH the `username` column and `getLocalActorId`, so the column and the id are derived from one value and cannot drift.
```
llun: **[correctness] "`normalizeUsername` runs in `localUsernameSchema`" contradicts the code.**

`localUsername.ts` imports only `zod` and `isFederationSigningActorUsername`. The fold there is a hand-rolled Zod `.trim()` + `.toLowerCase()`, so there are **two independent spellings** of the mint-time rule, not one — and `normalizeUsername`'s own JSDoc ("deliberately one function so they cannot disagree") is not what the code implements. Nothing pins them together.

Concrete drift: change `normalizeUsername` to strip a trailing dot or NFKC-normalize, and `registerAccount`/`createAccount` follow while `localUsernameSchema` does not — a name the schema validated as `alice.` gets stored as `alice`.

Fix: state accurately that the schema uses Zod's own chain (mirroring how the email schemas relate to `normalizeEmail`), and add a test pinning `localUsernameSchema.parse(x) === normalizeUsername(x)` so the two cannot drift.

llun: **[correctness] "`createAccount`/`createActorForAccount` … the one place a local actor row is written" is false.**

`getFederationSigningActor` (`lib/database/sql/actor.ts`) inserts a local actor row directly — non-empty `privateKey`, so `whereLocalActor` counts it — with its own `username` and its own `getFederationSigningActorId(domain, username)`, touching neither method. It is safe today only because both derive from the same `username` variable, which is exactly the argument the doc makes for the other two paths but not for this one. As written the doc tells a future author this class of path does not exist.

AGENTS.md
```diff
@@ -167,6 +167,19 @@ ActivityPub objects are **JSON-LD**, so the same logical object can arrive in ma
...
+- **Folding in `localUsernameSchema` runs BEFORE the reserved-name refine, and that ordering is load-bearing.** `isFederationSigningActorUsername` is a case-sensitive `startsWith('__instance__')`, so `__INSTANCE__` passed the check and minted a confusable neighbour of the instance actor at a _different_ id. It is also before `.max()`, so the length check reads the value actually stored — a fold can change a string's length. `LOCAL_USERNAME_PATTERN` still names `A-Z` on purpose: it describes what a caller may SEND, and by the time it runs there is none left.
+- **Every username LOOKUP folds**, through `findActorRowByUsername` (`lib/database/sql/utils/usernameMatch.ts`), which backs both `getActorFromUsername` and `isUsernameExists`. That covers `/@user`, WebFinger, mentions (which go through WebFinger), `/api/v1/accounts/lookup`, both search routes, `authorize_interaction` and the actor archive. `isUsernameExists` folding is what stops a second `Alice` from being minted beside an existing `alice`.
```
llun: **[quality + correctness] "That covers … both search routes" is inaccurate, and an existing lookup violates the "never a lone `lower(username) = ?`" rule.**

`getExactAccountIds` (`lib/database/sql/search/account.ts`) resolves an exact handle with a lone `LOWER(actors.username) = ?` plus `LOWER(actors.domain) = ?` — no exact arm, no precedence ordering, and `domain` folded, which these same docs say does not happen. It runs on **every** call to `searchAccountIds`, backing both search routes, not just their `resolve=true` branch.

So two endpoints disagree: `GET /api/v1/accounts/search?q=alice@Example.COM` can resolve a local `alice@example.com`, while `GET /api/v1/accounts/lookup?acct=alice@Example.COM` cannot. Pre-existing and not made worse by this PR, but the new rule is written as universal, so a reviewer applying it to that file gets a false positive — and the SQLite `Фёдор` failure mode the rule exists to prevent is live there.

lib/database/sql/utils/usernameMatch.ts
```diff
@@ -0,0 +1,66 @@
...
+ * functional `actors_lower_username_domain_idx` from
+ * `20260826120000_add_actors_lower_username_index.js`. The folded arm runs only
+ * on a miss, so the hot path is exactly the query this replaced.
+ *
+ * Ordering on the folded arm is `createdAt` then `id`: with the exact row
+ * already ruled out, more than one case-variant is possible, and the account
+ * that claimed the name first is both a defensible winner and a stable one.
+ * `.first()` alone would be neither.
+ *
+ * `domain` matching is left case-sensitive. Callers already lowercase it
```
llun: **[quality] "Callers already lowercase it (`parseAccountHandle`)" is false for two of this helper's own callers.**

- `app/api/v1/accounts/lookup/route.ts` defines its **own locally-shadowed** `parseAccountHandle(value, localDomain)` — same name as the shared one, different signature — which does not lowercase `domain`, then passes it straight to `getActorFromUsername`.
- `app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts` splits the actor segment inline (`decodedActor.split('@')`) with no normalization at all.

Not a regression (domain matching was already exact), but this blanket invariant is stated as settled fact in four places — here, AGENTS.md, CLAUDE.md and REVIEW.md — and it is not true. A future engineer relying on it to explain away `Alice@Example.COM` failing to resolve would be misled. Keep domain matching exact, but replace the justification with the accurate one.

lib/services/accounts/localUsername.ts
```diff
@@ -45,6 +45,19 @@ export const LOCAL_USERNAME_MAX_LENGTH = 50
...
+  // Lowercased rather than refused, so `Alice` registers as `alice` instead of
+  // becoming a second actor indistinguishable from `alice` to every
+  // case-insensitive client. The pattern below still names `A-Z` because it
+  // describes what a caller may SEND; by the time it runs there is none left.
+  //
+  // Ordering is load-bearing twice over. Folding before the reserved-name
+  // refine closes a bypass: `isFederationSigningActorUsername` is a
+  // case-sensitive `startsWith('__instance__')`, so `__INSTANCE__` passed the
+  // check and minted a confusable neighbour of the instance actor. And folding
+  // before `.max()` is what keeps the length check on the value actually
```
llun: **[correctness] The `.max()` half of "ordering is load-bearing twice over" overstates.**

The only Unicode lowercase mapping that lengthens a string is U+0130 `İ` → `i` + U+0307, and U+0307 is not in `LOCAL_USERNAME_PATTERN` — so any input whose fold changes length is refused by the regex regardless of where `.max()` sits, and ASCII folding is length-preserving. The reserved-name half **is** load-bearing (verified: with the fold before the refine, `__INSTANCE__` is rejected). Defensive ordering is fine; calling it load-bearing is a claim the code does not support.

lib/services/accounts/registerAccount.ts
```diff
@@ -49,6 +50,10 @@ export const registerAccount = async ({
   // same canonical (lowercased) address — even when this service is called
   // directly rather than through the request schema. See normalizeEmail.
   const email = normalizeEmail(rawEmail)
+  // Same reasoning for the username, and the same reason to do it here rather
+  // than trust the request schema: the taken-check, the insert and the actor id
+  // returned below must all see one canonical handle. See normalizeUsername.
+  const username = normalizeUsername(rawUsername)
```
llun, line 56: **[testing] This `normalizeUsername` call is untested — deleting it keeps every test green.**

Verified: reverting this line to `const username = rawUsername` leaves all 479 tests in `registerAccount.test.ts` and `app/api/v1/accounts/*` passing, because the production route always validates through `localUsernameSchema` first, which already lowercases. This line is real defense-in-depth for a direct service caller — the comment says exactly that — but nothing exercises it.

The email equivalent **is** tested directly in the same file (mixed-case/whitespace email asserted through to `createAccount`), which makes the missing username analogue a visible asymmetry. Add the mirror test.

lib/database/sql/utils/usernameMatch.ts
llun: Fixed in `a732eb5a8`. The folded arm now uses a bare `username.toLowerCase()`; `normalizeUsername` keeps the trim and stays on the MINT paths, where trimming is correct and mirrors `normalizeEmail`. Added five `it.each` cases in `usernameMatch.test.ts` covering leading/trailing spaces, a tab and padding around a different casing — all asserting the lookup returns null. Reverting the arm to `normalizeUsername` fails all five. The docblock now states why the two spellings differ, and AGENTS.md/REVIEW.md/CLAUDE.md say the arm folds case only.

lib/services/guards/OnlyLocalUserGuard.ts
llun: Fixed in `a732eb5a8`. Confirmed your analysis by test before fixing: with a legacy `__INSTANCE__` row present, `getActorFromUsername({ username: '__instance__' })` returned it. The guard now 404s any requested segment that folds to a reserved username unless the resolved actor IS the genuine signing actor — gated on the REQUESTED segment, not the resolved row, because a legacy row is exactly what cannot be trusted to describe itself at that URI.

Worth recording how nearly this test was useless: the first version seeded the squatter with `database.createActor()`, which produces an actor with no account, so the guard 404'd at the pre-existing `actor?.account` check and the test passed with the fix reverted. It now inserts through raw knex with an `accountId`, and carries a "confirms the squatter is a local account actor" assertion so that failure mode cannot come back silently.

AGENTS.md
llun: Fixed in `a732eb5a8`. You are right, and the JSDoc was the worst of it — it claimed the two could not disagree while the code had two independent spellings. AGENTS.md, CLAUDE.md and REVIEW.md now say there are TWO spellings (`normalizeUsername` and the schema's Zod chain), exactly as the email schemas relate to `normalizeEmail`, and `localUsername.test.ts` now asserts `localUsernameSchema.parse(x) === normalizeUsername(x)` over a table — so your trailing-dot scenario fails a test instead of shipping.

llun: Fixed in `a732eb5a8`. Verified: `getFederationSigningActor` inserts its own row with a non-empty `privateKey`. The docs now call `createAccount`/`createActorForAccount` the two ACCOUNT-facing mint paths and name that third writer explicitly, including the point that it is safe for the same reason — it derives its username and its id from one variable — rather than leaving that argument unstated for it.

AGENTS.md
llun: Fixed in `a732eb5a8`. The rule no longer reads as universal. AGENTS.md now scopes the coverage to the `resolve=true` branch of the search routes and names `getExactAccountIds` as the exception — including that it folds `domain` too, so search and lookup genuinely disagree on `alice@Example.COM`, and that the SQLite ASCII-fold gap is live in that one path. Unifying it needs its own change: it would alter domain-matching semantics and the index only covers `(lower(username), domain)`.

lib/database/sql/utils/usernameMatch.ts
llun: Fixed in `a732eb5a8`. Verified both call sites you named. All four copies of the claim are replaced: domain matching stays exact because that is what it already was and folding it is a separate change, NOT because callers normalize first — and the docblock now names the shadowed `parseAccountHandle` in `accounts/lookup/route.ts` and the inline split in `resolveStatusFromPath.ts` as counter-examples, plus the point that WebFinger carries its own domain fallback precisely because it is not a guarantee.

lib/services/accounts/localUsername.ts
llun: Fixed in `a732eb5a8`. Correct — `U+0307` is not in `LOCAL_USERNAME_PATTERN`, so the regex refuses any input whose fold changes length wherever `.max()` sits. The comment and AGENTS.md/REVIEW.md now call that ordering defensive rather than load-bearing, and keep "load-bearing" for the reserved-name half, which is verified.

lib/services/accounts/registerAccount.ts
llun: Fixed in `a732eb5a8`. Added the mirror of the email test: a mixed-case, whitespace-padded username asserted through to `isUsernameExists` and `createAccount`, and to the returned `actorId` as well as the column — since the username is interpolated into the id, normalizing one and not the other is the actual failure mode. Reverting the line fails it.

lib/services/guards/OnlyLocalUserGuard.ts
```diff
@@ -36,15 +39,51 @@ export const OnlyLocalUserGuard =
...
+    // itself, until `getFederationSigningActor()` first runs on that domain and
+    // the exact arm takes the id back. The old id-rebuild could not reach that
+    // state, and `actor.account` being truthy means the check above waves the
+    // squatter through even without `allowFederationSigningActor`.
+    //
+    // Gated on the REQUESTED segment, not on the resolved row: the question is
+    // who may answer at this URI, and a legacy row is exactly what cannot be
+    // trusted to describe itself here.
+    if (
+      isFederationSigningActorUsername(normalizeUsername(username)) &&
```
llun: **[correctness] The reserved-name check is a `startsWith`, and de-federates legacy accounts that can never own a signing-actor id.**

`isFederationSigningActorUsername` is `username.startsWith('__instance__')`, but `getFederationSigningActorUsername` only ever mints `__instance__` and `__instance__<digits>`. So a legacy local account named `__instance__archive` owns an id `getFederationSigningActorId` cannot produce at any index — and this check 404s its actor document, its inbox (so remote Follow/Create/Delete deliveries fail; only the shared inbox still works), its outbox, followers, following, collections, and every `/users/__instance__archive/statuses/<id>` dereference, which is how remote servers verify already-federated posts. Silently, with no migration or operator signal.

Such a name was registerable: the reserved-name refine first appeared in #612 (2026-05-01); before that `CreateAccountRequest` used an unanchored `z.string().regex(/\w+/)` with no reserved check. It returned 200 both on `main` and at `7f27287fb`.

The predicate matching the documented threat is "the folded segment equals `__instance__` or `__instance__<digits>`", not `startsWith`. Note the loose form is correct for its other callers: the mint refine (reserving the prefix stops a new confusable), and `isValidFederationSigningSQLActor`/`getExistingHeadlessActor`, which additionally require `accountId == null` so a user account can never be adopted as the signing actor. Only the guard applies the loose form to an account-bearing row.

lib/utils/normalizeUsername.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Canonicalizes a username for minting and comparison.
+ *
+ * Two separate jobs, deliberately one function so they cannot disagree:
+ *
+ *  - Every local username this instance MINTS is stored in this form, so
+ *    `Alice` and `alice` can never become two actors with two ActivityPub ids.
+ *    A local actor's id is `https://<domain>/users/<username>`, so the username
+ *    is the identity, not a label on it.
+ *  - Every username LOOKUP folds through it, so a handle typed, mentioned or
```
llun: **[correctness] This docblock still claims it backs every lookup — round 1 removed its last lookup caller.**

Round 1 replaced `normalizeUsername(username)` with `username.toLowerCase()` in `findActorRowByUsername`. After that this function has ZERO lookup callers: grep gives only the two mint paths in `account.ts`, `registerAccount.ts`, and a reserved-name check on a requested segment in `OnlyLocalUserGuard.ts`. The "deliberately one function so they cannot disagree" guarantee is precisely what round 1 deliberately broke.

This framing was copied from `normalizeEmail`, where "the one primitive every touchpoint routes through" IS true. A developer adding a rule here — the docs' own example is stripping a trailing dot — will believe from this file that lookups follow. They don't: `createAccount` would store `alice` and mint `.../users/alice` while `findActorRowByUsername` compares the caller's unstripped `alice.`. At `app/api/v1/actors/route.ts` that reaches `isUsernameExists({username: 'alice.'})` → miss → `createActorForAccount` normalizes to `alice` → duplicate `actors.id`, and that route has no `isUniqueConstraintError` handling, so a 500.

lib/database/sql/account.ts
```diff
@@ -106,6 +106,12 @@ export const AccountSQLDatabaseMixin = (database: Knex): AccountDatabase => ({
     publicKey
   }: CreateAccountParams) {
     const normalizedEmail = normalizeEmail(email)
+    // Normalized here as well as in the request schema for the same reason
+    // emails are: this is the one place a local actor is minted, so a caller
```
llun: **[correctness] This comment still says "the one place a local actor is minted" — the claim round 1 corrected everywhere else.**

Introduced by `7f27287fb`, and the origin of the claim round 1 went and fixed in AGENTS.md, REVIEW.md and CLAUDE.md. It is wrong twice over: `createActorForAccount` is a second mint path, and `getFederationSigningActor` inserts a local actor row bypassing both. The docs and the code now contradict each other on the one point round 1 set out to correct — and this is the comment a developer editing `createAccount` actually reads.

lib/database/sql/utils/usernameMatch.ts
```diff
@@ -0,0 +1,83 @@
...
+ * that claimed the name first is both a defensible winner and a stable one.
+ * `.first()` alone would be neither.
+ *
+ * `domain` matching is left case-sensitive, which is what it already was before
+ * this helper existed — folding it is a separate change, with its own index
+ * implications, and it is not this helper's question. Do NOT read that as
+ * "callers normalize it first": `parseAccountHandle` does, but
+ * `app/api/v1/accounts/lookup/route.ts` has its own locally-shadowed
+ * `parseAccountHandle` that does not, and
+ * `app/(timeline)/[actor]/[status]/resolveStatusFromPath.ts` splits the segment
```
llun, line 40: **[quality] The stated rationale for dropping the trim is not the load-bearing one.**

The comment justifies the fix with the outbox root's `Cache-Control: public, max-age=60, s-maxage=60` turning each spelling into its own shared-cache key. True, but not load-bearing: case folding is this helper's whole point, so `/users/AlIcE/outbox` and its 2^n siblings remain distinct cacheable 200s with identical bodies regardless. The fix is still right, for a different reason — a trimmed input compared against an untrimmed `lower(username)` is asymmetric and can only ADD matches the exact arm would not have found. Origin cost per variant is small anyway because `getCachedActorPublicStatusesCount` memoizes per `actorId`, not per URL.

lib/services/guards/OnlyLocalUserGuard.ts
llun: Fixed in `5b8e18722`. Confirmed `getFederationSigningActorUsername` mints only `__instance__` and `__instance__<digits>`, so this was a real regression introduced by round 1 — it 404'd an actor that works on `main`.

Added `isFederationSigningActorIdUsername` (`/^__instance__\d*$/`) to `instanceActor.ts` and wired the guard to it; the loose `startsWith` keeps its other callers, with a comment on each explaining which question it answers. Two `it.each` cases now assert a legacy `__instance__archive` / `__instance___backup` account still returns 200, and a third asserts `__instance__1` still 404s. Reverting the predicate to `startsWith` fails both 200 cases.

lib/utils/normalizeUsername.ts
llun: Fixed in `5b8e18722`. The docblock now says plainly that this is a MINT-side primitive and, unlike `normalizeEmail`, is NOT what every touchpoint routes through — naming both exceptions (the lookup folds with a bare `toLowerCase()`; the schema spells the rule as its own Zod chain) and stating that a new rule added here reaches the mint paths and nothing else, so the fold and the schema must be updated deliberately rather than by inheritance.

lib/database/sql/account.ts
llun: Fixed in `5b8e18722`. The comment now states why normalization is here (a direct caller cannot produce a mixed-case handle, and the column and id come from one variable) without the false exclusivity claim, and explicitly names `createActorForAccount` and `getFederationSigningActor` as the other writers, each of which has to make the one-variable argument for itself.

lib/database/sql/utils/usernameMatch.ts
llun: Fixed in `5b8e18722`. The comment now leads with the asymmetry — `' alice '` found `alice` while `alice` could never find a stored `'alice '`, so the relation only ever adds rows — and keeps the cache point in parentheses explicitly marked as NOT load-bearing, with your `getCachedActorPublicStatusesCount` note on why the origin cost is small. Better to record why the earlier rationale was wrong than to quietly delete it.

lib/services/federation/instanceActor.ts
```diff
@@ -7,9 +7,27 @@ export const FEDERATION_SIGNING_ACTOR_TYPE = 'Service'
...
+// and one call site needs this one. `OnlyLocalUserGuard` asks "may a non-signing
+// actor answer at this URI", and the prefix test over-answers it: a legacy
+// account named `__instance__archive` — registerable before the reserved-name
+// refine landed in #612, when the username schema was an unanchored `/\w+/`
+// with no reserved check — owns an id `getFederationSigningActorId` cannot
+// produce at any index, yet a `startsWith` guard 404s its actor document,
+// inbox (so remote deliveries fail), outbox, followers, following, collections
+// and every already-federated status dereference. Silently, with no migration.
+export const isFederationSigningActorIdUsername = (username: string) =>
+  /^__instance__\d*$/.test(username)
```
llun: **[correctness] `/^__instance__\d*$/` is not the mintable set in either direction, and the comment claims it is.**

**Over-accepts.** The minter emits `__instance__` for index 0 and `` `__instance__${index}` `` for integer index >= 1, so the index never carries a leading zero and `__instance__0` is never spelled. The regex matches `__instance__0`, `__instance__00`, `__instance__007`. So a pre-#612 legacy account at `https://<host>/users/__instance__0` — `\w+` accepted that name exactly as it accepted `__instance__archive` — gets 404 on its actor document, inbox, outbox, followers and every already-federated status dereference, where `main` serves 200. That is the same de-federation this commit's subject says it removes, left in place for a name the minter provably cannot produce.

**Over-rejects the "ONLY … can ever be" claim.** `getExistingHeadlessActor` selects `WHERE domain = ? AND type = 'Service' AND username LIKE '\_\_instance\_\_%' AND accountId IS NULL` and validates the winner with `isValidFederationSigningSQLActor`, which uses the LOOSE predicate. `createSigningActor` consults it before minting, so a headless Service row named `__instance__archive` is ADOPTED and its id becomes the live signing keyId used by every `send*Job`. A signing-actor id outside `__instance__\d*` is reachable by construction.

Nothing exploitable follows (an id is unique, so an account-owning actor can never occupy an adopted signer's URI, and the mint refine still blocks new `__instance__*` names) — but this comment is the stated justification for narrowing an authorization boundary, and it invites the follow-up that would break it: unifying `isValidFederationSigningSQLActor`/`isFederationSigningActor` onto the precise form would stop an adopted signer validating at all, silently ending federation signing on that instance.

lib/utils/normalizeUsername.ts
```diff
@@ -0,0 +1,28 @@
...
+ * every touchpoint routes through.** Do not copy that framing across:
+ *
+ *  - LOOKUP does not use it. `findActorRowByUsername` folds with a bare
+ *    `toLowerCase()`, because the trim here would compare a trimmed input
+ *    against an untrimmed `lower(username)` — see the comment there.
+ *  - The request schema does not call it either; `localUsernameSchema` spells
+ *    the same rule as a Zod `.trim().toLowerCase()` chain, and
+ *    `localUsername.test.ts` pins the two against each other.
+ *
+ * So a new rule added here (the standing example is stripping a trailing dot)
```
llun: **[correctness] "reaches the mint paths and NOTHING else" is false — this is round 2's own correction re-introducing the class of claim it was fixing.**

`OnlyLocalUserGuard` calls `normalizeUsername(username)` directly on the incoming request's path segment, immediately before testing it against `isFederationSigningActorIdUsername`. That is a lookup/access-control path, not a mint path, and that call site predates this commit. The round-2 commit message makes the same claim explicitly, so it is the PR's stated reasoning, not a stray docblock artifact.

Scenario: an engineer trusting this docblock adds the trailing-dot rule it names as "the standing example", and — following its instructions — updates the fold in `findActorRowByUsername` and the Zod schema, believing that is exhaustive. They never touch the guard, so `__instance__.` silently starts folding to `__instance__` before the reserved-name test, changing which requests are treated as addressing the reserved slot.

lib/services/guards/OnlyLocalUserGuard.test.ts
```diff
@@ -103,6 +104,182 @@ describe('OnlyLocalUserGuard', () => {
...
+      }
+    )
+
+    it.each([
+      { description: 'requested in lowercase', requested: '__instance__' },
+      {
+        description: 'requested in the stored casing',
+        requested: '__INSTANCE__'
+      },
+      { description: 'an indexed signing username', requested: '__instance__1' }
```
llun: **[testing] The `__instance__1` case is vacuous.**

No `__instance__1` actor exists on `llun.test` in this suite, so the 404 comes from the pre-existing "actor not found" branch, not the reserved-name check the describe block implies. Proven by mutation: replacing the predicate with `/^__instance__$/` leaves 17/17 passing, and neutering it entirely (`/^__never_matches__$/`) still leaves this row passing.

AGENTS.md
```diff
@@ -167,6 +167,20 @@ ActivityPub objects are **JSON-LD**, so the same logical object can arrive in ma
...
+- **A reserved username is reserved case-INSENSITIVELY, and `OnlyLocalUserGuard` enforces that at the URI.** Resolving by username is what made this necessary: `__INSTANCE__` was registerable before the refine folded casing, and the folded arm answers a request for `__instance__` with that account's actor — a user-owned Person document, inbox, outbox and followers served at `getFederationSigningActorId(domain)` itself, past the `actor.account` check and without `allowFederationSigningActor`, until `getFederationSigningActor()` first runs on that domain. The guard therefore 404s any request whose segment folds to a reserved name unless the resolved actor IS the genuine signing actor, gated on the REQUESTED segment rather than the resolved row.
```
llun: **[correctness] AGENTS.md and REVIEW.md still describe the round-1 guard, not the one at HEAD.**

Both say the guard "404s any request whose segment folds to a reserved name", and "reserved name" is defined two bullets earlier as what `isFederationSigningActorUsername` refuses — the whole `__instance__` prefix. At HEAD the guard refuses only `/^__instance__\d*$/`. Concretely, a legacy `__instance__archive` account returns 200 (that is the new test), where both docs say 404.

REVIEW.md is the checklist reviewers apply, so as written it marks the round-2 fix as the defect; an implementer "restoring" the rule re-widens the predicate and re-breaks the actor document and inbox of every `__instance__*` legacy account. Neither doc mentions the new predicate split at all, even though which one a call site takes is now load-bearing.

Same area, minor: both still lead the trim rationale with the shared-cache argument that the round-2 code comment explicitly demotes to "true but NOT load-bearing", so code and docs disagree about which argument is the real one.

lib/services/federation/instanceActor.ts
llun: Fixed in `6927881af`. Both halves confirmed.

The predicate is now `/^__instance__([1-9]\d*)?$/` — verified against the minter that this is exactly the emitted set, and that `__instance__0`/`__instance__00`/`__instance__007` are excluded. Two `it.each` cases assert a legacy `__instance__0` and `__instance__007` account still returns 200; reverting to `\d*` fails both.

The comment no longer claims to be every possible signer id. It now states the adoption path explicitly and carries a "these two predicates are NOT interchangeable and must not be unified" warning naming what breaks in each direction — widening the guard de-federates legacy accounts, narrowing the loose form kills an adopted signer. AGENTS.md, REVIEW.md and CLAUDE.md carry the same warning.

lib/utils/normalizeUsername.ts
llun: Fixed in `6927881af`. You are right, and it is the third round in a row where the defect was prose over-claiming rather than logic.

The docblock now enumerates all three callers — two mint paths and the guard's reserved-name check — spells out that a rule added here changes an access-control decision (with `__instance__.` folding to `__instance__` as the worked example), and ends by telling the reader to enumerate the call sites rather than trust any summary, including that one.

lib/services/guards/OnlyLocalUserGuard.test.ts
llun: Fixed in `6927881af`. Confirmed — and this is the second vacuous test in this PR, both times a fixture that did not reach the branch under test, so I am treating it as a pattern rather than an accident.

Replaced with a dedicated describe that seeds an `__instance__2` account actor through raw knex and asserts it really resolves with a truthy `account` before asserting the 404, so the "actor not found" branch cannot supply the result. Deleting the reserved-name check now fails 3 tests where it previously failed 2.

AGENTS.md
llun: Fixed in `6927881af`. The REVIEW.md point is the sharpest version of this — a checklist that marks the fix as the defect is worse than no checklist entry.

All three docs now describe the precise predicate, state that it is deliberately narrower than the mint refine's, and carry the same "not interchangeable" warning with what breaks in each direction. The trim rationale now leads with the asymmetry in all three, with the cache point marked explicitly as not load-bearing.

lib/utils/normalizeUsername.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * And it has exactly one caller that is neither a mint nor a lookup:
+ * `OnlyLocalUserGuard` folds the requested path segment with it before asking
+ * `isFederationSigningActorIdUsername` whether that segment names the instance
+ * actor's URI. That is an ACCESS-CONTROL decision, so a rule added here changes
+ * it. The standing example makes that concrete: teach this function to strip a
+ * trailing dot and `__instance__.` starts folding to `__instance__`, changing
+ * which requests are treated as addressing the reserved slot.
+ *
+ * So the callers are: two mint paths, one reserved-name check, and nothing
```
llun: **[correctness] The docblock enumerates 3 callers; there are 4.**

`grep -rn normalizeUsername` over the repo (excluding tests) returns four production call sites: `lib/database/sql/account.ts:119` (`createAccount`), `lib/database/sql/account.ts:446` (`createActorForAccount`), **`lib/services/accounts/registerAccount.ts:56`**, and `lib/services/guards/OnlyLocalUserGuard.ts:87`. `registerAccount` is a third mint caller not covered by "two mint paths … and nothing else", and AGENTS.md names it as its own layer for a specific reason ("so a direct service call cannot bypass the schema").

Concrete failure: the docblock's standing example is "teach this function to strip a trailing dot." A maintainer doing that follows the instruction, audits "those three", and never opens `registerAccount.ts:56` — where the newly-folded value is what feeds `database.isUsernameExists({ username, domain })` and the `ERR_TAKEN` decision. The closing "do not trust a summary of them, including this one" softens it but does not repair the positive claim "and nothing else", which is what a reader acts on.

This is the sentence round 3 added specifically to fix an over-claim, so it is the highest-value one to correct.

lib/services/guards/OnlyLocalUserGuard.ts
```diff
@@ -36,15 +39,57 @@ export const OnlyLocalUserGuard =
...
+    // the exact arm takes the id back. The old id-rebuild could not reach that
+    // state, and `actor.account` being truthy means the check above waves the
+    // squatter through even without `allowFederationSigningActor`.
+    //
+    // Gated on the REQUESTED segment, not on the resolved row: the question is
+    // who may answer at this URI, and a legacy row is exactly what cannot be
+    // trusted to describe itself here.
+    //
+    // `isFederationSigningActorIdUsername`, NOT the loose prefix test the mint
+    // refine uses. Only `__instance__` and `__instance__<digits>` can ever be a
```
llun: **[correctness] The call-site comment restates the predicate, and both halves are now false.**

> `Only __instance__ and __instance__<digits> can ever be a signing-actor id, so the prefix form would de-federate a legacy __instance__archive account`

- "**Only** … can ever be a signing-actor id" is exactly what round 3 disclaimed in the helper it calls: *"This one is deliberately NOT 'every id that can ever be a signing actor' — that set is larger. `getExistingHeadlessActor` ADOPTS any pre-existing headless Service row matching `__instance__%`…"* On an instance that adopted `__instance__archive`, that URI **is** a signing-actor id. The guard comment asserts the opposite of the helper comment, AGENTS.md, REVIEW.md and CLAUDE.md.
- "`__instance__<digits>`" describes `/^__instance__\d*$/` — the regex round 3 *replaced*. `__instance__0` and `__instance__007` are `__instance__<digits>` and are now deliberately not reserved, with tests asserting they return 200.

Failure: a maintainer triaging "why isn't `__instance__007` reserved?" reads only the call-site comment, sees the predicate documented as `__instance__<digits>`, and widens the regex back to `\d*` — re-de-federating the legacy accounts round 3 fixed.

Same phrasing also sits in `OnlyLocalUserGuard.test.ts` three lines above the `it.each` table that disproves it, and in that file's `404s it, because __instance__<digits> is a mintable signing id` test title.

lib/services/federation/instanceActor.ts
```diff
@@ -7,9 +7,43 @@ export const FEDERATION_SIGNING_ACTOR_TYPE = 'Service'
...
+// status dereference. Silently, with no migration.
+//
+// **These two predicates are NOT interchangeable and must not be unified.**
+// This one is deliberately NOT "every id that can ever be a signing actor" —
+// that set is larger. `getExistingHeadlessActor` ADOPTS any pre-existing
+// headless Service row matching `__instance__%` with a null `accountId`, and
+// validates it with the LOOSE predicate, so an instance can legitimately sign
+// as `__instance__archive`. Narrowing `isValidFederationSigningSQLActor` or
+// `isFederationSigningActor` onto this form would stop such an adopted signer
+// validating at all — silently ending federation signing on that instance. The
```
llun, line 40: **[quality] "allowed by `isFederationSigningActor` before this test is reached" names the wrong mechanism.**

Under a charitable reading this is true, and I could not construct a wrong behaviour from it — but it misdescribes twice. (i) 7 of the guard's 14 invocations do not pass `allowFederationSigningActor`, and on those an adopted signer is *not* allowed — it 404s earlier. (ii) What actually keeps an adopted `__instance__2`-style signer from being 404'd is the `!isFederationSigningActor(actor)` conjunct **inside** the test, not something before it. A reader who believes that second call is redundant "because it was already allowed above" and deletes it would 404 the genuine signer's actor document and inbox on all 7 opt-in routes.

llun (commented):
Round 4 review findings.

lib/utils/normalizeUsername.ts
llun: Fixed in `3d9787fb0`. Confirmed by grep — four callers, and I had missed `registerAccount`.

The paragraph now names all four explicitly (three mint-side, including `registerAccount` and why it is its own layer, plus the guard's reserved-name check) and records that an earlier revision of itself made exactly this mistake, so the next reader sees the failure mode rather than another confident summary.

Worth noting the two round-4 lenses disagreed here: the scope lens reported "grep confirms exactly these three non-test call sites". I checked by hand and you were right. Recorded, because it means a clean report from one lens is not confirmation.

lib/services/guards/OnlyLocalUserGuard.ts
llun: Fixed in `3d9787fb0`. The call site no longer restates the definition at all — it points at the helper and names the two misreadings to avoid ("every signing-actor id", and `__instance__<digits>`), which is the failure you described. The test comment and the test title are corrected the same way; the only remaining occurrences of `__instance__<digits>` in either file are the explicit "NOT that" disclaimers.

lib/services/federation/instanceActor.ts
llun: Fixed in `3d9787fb0` in both `instanceActor.ts` and AGENTS.md. The comment now names the `!isFederationSigningActor(actor)` conjunct as the thing doing the work, says explicitly not to delete it as redundant, and records that the two checks are not interchangeable because the 7 invocations without `allowFederationSigningActor` refuse an adopted signer earlier anyway.

I re-derived the 7-of-14 figure rather than copying it, and initially got 5 by counting FILES instead of invocations — 14 invocations, 7 with the option, 7 without. Your number was right.

lib/services/guards/OnlyLocalUserGuard.ts
```diff
@@ -36,15 +39,66 @@ export const OnlyLocalUserGuard =
...
+    // account — 404ing an actor document and inbox that work on `main` today.
+    // See that helper for the exact set and for why the two predicates must not
+    // be unified; do NOT restate it here, and in particular do not read it as
+    // "every signing-actor id" (an ADOPTED headless signer can sit outside it)
+    // or as `__instance__<digits>` (`__instance__0` and `__instance__007` are
+    // deliberately not reserved — the minter cannot emit them).
+    //
+    // The second conjunct is what keeps the GENUINE signer served here, on the
+    // routes that opt into it. It is not redundant with the `isAllowedActor`
+    // check above: 7 of this guard's call sites do not pass
```
llun: **[correctness] "7 of this guard's call sites do not pass `allowFederationSigningActor`" — it is 5, of 14.**

Counted by invocation, not by file. There are exactly 14 non-test `OnlyLocalUserGuard(` invocations. **9 pass** `allowFederationSigningActor: true` (`route.ts`, `followers`, `outbox`, `inbox`, `following`, `collections/tags`, `collections/featured`, `collections/featured-collections/route.ts`, `collections/featured-collections/[id]/route.ts`). **5 do not**: `quote_authorizations/[id]/route.ts`, `statuses/[statusId]/route.ts`, `statuses/[statusId]/replies/route.ts`, `statuses/[statusId]/likes/route.ts`, `statuses/[statusId]/shares/route.ts`.

The PR touches no route file, so 5 is also the number on `main` — the figure was never right, not stale. Nothing in the repo produces 7 (the test file has 15 invocations, 13 without the option). The same sentence appears in `lib/services/federation/instanceActor.ts:44`.

Not a functional defect — the surrounding reasoning holds — but it is exactly the kind of checkable number the next reader will grep, and it fails on the first grep. It was introduced by the round whose stated purpose was fixing prose accuracy.

lib/services/federation/instanceActor.ts
```diff
@@ -7,9 +7,46 @@ export const FEDERATION_SIGNING_ACTOR_TYPE = 'Service'
...
+// guard stays correct across that split because its reserved-name test carries
```
llun: **[correctness] "that, not anything earlier in the guard, is what keeps a genuine signer served" attributes the wrong mechanism.**

Two inaccuracies, both in the replacement for round 3's wrong story:

**(a) For the case this paragraph is about, the conjunct is not the reason.** The subject is the adopted headless signer (`getExistingHeadlessActor` adopting `__instance__archive`). For that name `isFederationSigningActorIdUsername('__instance__archive')` is **false**, so `&&` short-circuits and `!isFederationSigningActor(actor)` is never evaluated. What keeps an adopted signer past that test is the *first* conjunct's precision — the property the preceding sentences establish. The conjunct is load-bearing only for a signer whose name the minter CAN emit.

**(b) "not anything earlier in the guard" is false even for the bare `__instance__` signer.** `isAllowedActor`'s `options.allowFederationSigningActor && isFederationSigningActor(actor)` is earlier and equally necessary: an accountless signer has a falsy `actor.account`, so deleting that disjunct 404s the genuine signer before the reserved-name test runs, on all 9 opt-in routes. Both checks are required; the sentence tells a maintainer one of them isn't.

The guard's own copy is correctly scoped ("here, on the routes that opt into it"); it is the `instanceActor.ts` and AGENTS.md copies that drop the scope and add the exclusive claim.

lib/services/guards/OnlyLocalUserGuard.ts
llun: Fixed in `2913d5efc`. Confirmed 5 of 14 by paren-balancing each invocation; both round-5 lenses independently got the same.

Worth recording how this happened, because it is worse than a typo. The number came from a counting script of mine that sampled the text after the FIRST `OnlyLocalUserGuard(` in a file for every match in that file, so any multi-guard file was miscounted — and I then used that broken output to *confirm* the figure rather than to check it, and reported it as verified. The lesson is not "count more carefully" but "a script written to check a claim needs checking before its output is evidence."

The count now appears once, in the guard, instead of being restated in `instanceActor.ts` — a checkable number in two places is two chances to be wrong.

lib/services/federation/instanceActor.ts
llun: Fixed in `2913d5efc`. Both halves confirmed directly rather than by reading: `isFederationSigningActorIdUsername('__instance__archive')` is false so the `&&` short-circuits, and `isAllowedActor` is `actor?.account || (allowFederationSigningActor && isFederationSigningActor(actor))`, which an accountless signer can only clear via the second disjunct.

The guard comment now sets out both checks, says which case each carries, and states explicitly that the conjunct does no work for an adopted signer outside the mintable set. `instanceActor.ts` no longer restates the reasoning at all — it points at the guard — and AGENTS.md says both are required with neither subsuming the other.

This is the third consecutive round where a correction of mine introduced the next finding. That pattern is now itself recorded in the `normalizeUsername` docblock, which tells the reader to grep the call sites rather than trust its own summary.

llun (commented):
Round 6 mutation-sweep and doc-drift findings.

lib/database/sql/utils/usernameMatch.test.ts
```diff
@@ -0,0 +1,379 @@
+import { getTestSQLDatabase } from '@/lib/database/testUtils'
```
llun, line 1: **[testing] Mutation sweep: `getMastodonActorFromUsername`'s fold was entirely unguarded.**

A 24-mutation audit of the whole PR — breaking each behavioural guarantee in turn and recording which tests fail — caught 22. This was one of two survivors.

Reverting `getMastodonActorFromUsername` (`lib/database/sql/actor.ts`) to its pre-PR raw exact-match query, bypassing `findActorRowByUsername`, left the entire 300-test targeted suite green. The method's own comment says it folds "for the same reason `getActorFromUsername` does… which makes it the cheapest possible moment to stop them diverging" — but its only callers are pre-existing exact-case fixtures in `actor.test.ts`/`account.test.ts`, so the divergence could come straight back with nothing failing. Low severity (it has no production caller today) but the stated guarantee had zero coverage.

lib/database/sql/utils/usernameMatch.ts
```diff
@@ -0,0 +1,93 @@
+import { Knex } from 'knex'
```
llun, line 1: **[testing] Mutation sweep: domain case-sensitivity was unguarded.**

The second survivor of the same audit. This helper's docblock states that leaving `domain` case-sensitive is a deliberate boundary ("folding it is a separate change… not this helper's question"). I folded `domain` alongside `username` in the folded arm's `whereRaw` and the full 300-test suite still passed.

The existing `does not fold across domains` cases cannot detect this: they use an entirely different domain string (`somewhere.else.test` vs `TEST_DOMAIN`), never a same-domain case variant, so they pass whether or not domain is folded. Given AGENTS.md notes `getExactAccountIds` already disagrees with this project on domain-folding, the boundary is worth pinning explicitly.

REVIEW.md
```diff
@@ -101,6 +101,84 @@ change doesn't touch.
...
+  (`lib/utils/normalizeUsername.ts`) and every lookup folds through
+  `findActorRowByUsername` (`lib/database/sql/utils/usernameMatch.ts`).
+- Normalization is layered like email's: `localUsernameSchema`,
+  `registerAccount`, **and** `createAccount`/`createActorForAccount`. The last is
+  the one that matters — it is where the column and the id are
+  derived from one variable and so cannot drift. It is NOT the only place a local
+  actor row is written — `getFederationSigningActor` inserts its own — and the
+  schema's fold is a SECOND spelling of the rule (Zod's `.trim().toLowerCase()`,
+  not a `normalizeUsername` call), pinned against it by `localUsername.test.ts`.
+- The fold in `localUsernameSchema` runs **before** the reserved-name refine and
```
llun, line 119: **[correctness] The `.max()` ordering claim here is the one copy round 1 missed.**

REVIEW.md says a fold can change a string's length "so `.max()` must read the stored value". `AGENTS.md` and the code comment both say the opposite — "defensive only, NOT load-bearing" — corrected in round 1, whose commit message reads "`.max()` ordering is defensive, not load-bearing, since the regex refuses any input whose fold changes length". `git log -S` confirms this checklist still carries the pre-round-1 text.

`İ` (U+0130) is the only lengthening `toLowerCase()` mapping and yields `i` + U+0307, which is outside `LOCAL_USERNAME_PATTERN`, so `.regex()` refuses any such input wherever `.max()` sits.

REVIEW.md is the checklist a reviewer applies, so this is the copy most likely to cause harm: whoever next reconciles the three documents sees two disagreeing with the checklist and "fixes" the correct two back to the false claim.

lib/database/sql/utils/usernameMatch.test.ts
llun: Fixed in `437b931f5`. Added a `getMastodonActorFromUsername` describe with mismatched-case `it.each` cases mirroring the `getActorFromUsername` block, plus an absent-username case. Reproduced your exact mutation against them: 0 failures before, 2 after.

lib/database/sql/utils/usernameMatch.ts
llun: Fixed in `437b931f5`. Added a `domain matching stays case-sensitive` describe: uppercased and capitalised domain variants against `getActorFromUsername`, the same for `isUsernameExists`, and — the case that actually matters — one where the USERNAME case differs too, so the exact arm misses and the FOLDED arm is what answers, which is the arm a domain fold would live in. It asserts the same lookup with the exact domain still resolves, so the miss is provably the domain and not the casing.

Reproduced your mutation: 0 failures before, 4 after (6 across both findings).

REVIEW.md
llun: Fixed in `340045ec8`. Verified by execution before changing it — both `İ` and a 50-character name plus `İ` fail `LOCAL_USERNAME_PATTERN` raw and folded, so the ordering really is defensive-only.

The bullet now states that, and explicitly warns against reconciling the other two copies back to the load-bearing claim, naming why this one survived: round 1 fixed two of the three places. That is the third time in this loop a correction reached some copies of a claim and not all of them, so the warning is aimed at the pattern rather than just this sentence.

PR #1597, opened 2026-08-27
fix: repair blurhashes stored before the write path validated them

llun:
## Why

A blurhash is the one piece of media metadata a remote actor hands us directly: it rides on a federated note's attachment and `createNoteJob` persists it.

Before `normalizeBlurhash`, the validator compared `hash.trim()` while the caller stored the **untrimmed original**, so a whitespace-padded hash was approved on the trimmed copy and written in a form `decode` throws on (`blurhash length mismatch: length is 29 but it should be 28`). A structurally invalid one got through too — right alphabet, right length, but not the length the size flag in its own first character demands (`'aaaaaa'`) — because the check never ran the blurhash package's `isBlurhashValid`.

That fix covered the **write path only**:

1. Existing rows are not repaired, and nothing re-validates on read.
2. `scripts/maintenance/backfillMediaBlurhash.ts` cannot reach them. Its non-force selection is `blurhash IS NULL OR thumbnailUrl LIKE '/api/v1/files/%'`; a federated attachment with a bad blurhash has a **non-null** `blurhash` and a **null** `thumbnailUrl`, so it matches neither branch. Only `--force` reaches it, and that re-downloads and recomputes every attachment in the instance.
3. `lib/types/domain/attachment.ts` re-serves the stored value verbatim to third-party clients (`getMastodonAttachment`, `getDocumentFromAttachment`), so the bad value keeps leaving this instance.

## What changed

`--revalidate` is a new **mode** of `backfillMediaBlurhash.ts`. It selects `attachments` where `blurhash IS NOT NULL`, runs each value through `normalizeBlurhash`, and writes back the normalized form — or nulls the column when the hash is unsalvageable.

**It reads no image bytes and makes no network call.** A padded hash only has to be trimmed; an unsalvageable one only has to be cleared. So the mode is reached *before* the storage and host checks in `main()` and stays usable on an instance whose storage backend is unreachable or unconfigured (verified below).

```text
[attachments att-padded] blurhash "  L6PZfSi_.AyE_3t7t7R**0o#DgR4\n" is stored padded; rewriting it as "L6PZfSi_.AyE_3t7t7R**0o#DgR4"
[attachments att-structural] blurhash "aaaaaa" is not one `decode` can read; clearing it
Blurhash revalidation complete: scanned 5, repaired 2, cleared 2, left 1 untouched
```

### Decisions worth reviewing

- **A mode, not a wider default selection.** Folding `blurhash IS NOT NULL` into the default predicate would make the ordinary sweep scan every attachment in the instance, which is exactly what the narrow default predicate exists to avoid.
- **`--revalidate` and `--force` refuse to run together** (exit 1 with a clear message) rather than one silently winning. They answer opposite questions — "recompute this from the image" against "repair what is already stored, touching no image".
- **They compose across runs, in one direction.** `--revalidate` clears what it cannot fix; those rows then match the default run's `blurhash IS NULL` selection, so a later ordinary run is what recomputes them from the bytes where the bytes still exist. Confirmed on the local fixture: the ordinary sweep saw 1 candidate before and 3 after.
- **Three counts, reported separately** — repaired / cleared / untouched — following the script's existing habit of not merging distinct causes into one `processed` number. Noted in the code and in AGENTS.md that, *unlike* `backfillAttachments`' two counts, these three **do** partition the scan.
- **No `mediaType` filter.** A video attachment carries its poster frame's blurhash, and the default sweep's analysis step only ever reads `image/*`, so this is the only pass that reaches one.
- **`medias.blurhash` is out of scope.** Every value in that column comes from `computeBlurhash` → the blurhash package's `encode`, so it is canonical by construction; no path stores a peer-supplied hash there. Checked `createMedia`/`updateMedia`'s callers (`localFile.ts`, `S3StorageFile.ts`, both media routes).

### Clearing is safe

`lib/components/posts/media.tsx` gates the `` at `opacity-0` until `onLoad` behind a blurhash canvas **only when `blurhash` is truthy**; a falsy one falls through to a bare ``. `BlurhashCanvas` swallows a failed `decode` into an empty canvas. So an undecodable hash renders an empty box until the image loads, and a `NULL` one renders the image immediately — clearing is strictly better than leaving the value.

Per **Deleting Media a Post Uses** in `AGENTS.md`: this does not weaken the deleted-media placeholder promise. That promise rests on the attachment carrying a blurhash a client can actually **paint**, and a value `decode` refuses never painted one. Nothing here touches `attachments.mediaId`.

## Verification

Full gate green: `prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test` (894 files, 11904 tests).

**Ran the script for real** against a worktree-local SQLite DB (`ACTIVITIES_DATABASE_CLIENT=better-sqlite3`, `dev.sqlite3`, 160 migrations applied), seeded with one attachment of each shape — padded-but-valid, structurally invalid, out-of-charset, canonical, `NULL`, and a `video/mp4` with a padded hash:

- `--revalidate --dry-run` → correct plan, database unchanged byte-for-byte.
- `--revalidate --batch-size 2` → repaired 2, cleared 2, 1 untouched; the canonical row survived byte-for-byte; every surviving hash then `decode`s cleanly. Batch size 2 exercises the keyset paging across a predicate the pass invalidates as it writes.
- Second run → `scanned 3, repaired 0, cleared 0, left 3 untouched` (idempotent).
- `--revalidate --force` → exits 1 with the refusal message.
- With `ACTIVITIES_MEDIA_STORAGE_*` removed → `--revalidate` still runs; the ordinary sweep correctly fails with `Media storage backend is not configured`.

**Tests** (`scripts/maintenance/backfillMediaBlurhash.test.ts`, 64 passing) use the **real** `normalizeBlurhash` — the module mock now spreads `importOriginal()` and replaces only `analyzeImageBuffer` — so they assert what `decode` can actually read rather than what a stub was told to say. Mutation-tested; each of these makes at least one test fail:

| Mutation | Tests failed |
| --- | --- |
| never write | 7 |
| `--dry-run` writes anyway | 1 |
| hardcoded batch limit | 1 |
| summary logged per row instead of once | 3 |
| drop the charset half of the check | 1 |
| selection includes `NULL` blurhashes | 1 |
| offset paging instead of keyset | 2 |
| `update` without the row filter | 3 |
| rewrite even a canonical hash | 1 |

## Docs

`docs/maintenance.md` gains a **Repairing blurhashes already stored** section; `AGENTS.md` gains **A federated blurhash is untrusted input, and the residue is repaired by a mode of its own** under Security & Configuration Tips, with the matching condensed bullet in `CLAUDE.md`.

## Note on incidental changes

`main()` was restructured so `getKnex()` is inside the `try` whose `finally` destroys it. Previously a throw from the database/storage/host checks skipped `db.destroy()`. Same checks, same order, same messages.

---

## Review loop

**Round 1** — five lenses: correctness, scope/regression, quality, testing, and a destructive-operation lens. Correctness, scope/regression and quality returned no findings; each independently verified something I could not fully prove alone (attachment ids are always `crypto.randomUUID()`, so the `id > ''` keyset cursor orders identically on both backends; `decode`'s own length check is intrinsic to the format's one-byte header, so nothing this pass clears is renderable by a spec-compliant decoder; `JSON.stringify` escapes every C0 control character, so the untrusted stored value cannot inject terminal escapes).

Three findings, all posted inline and reproduced before posting. Fixed in `6069fbc7b`:

- **The summary-counts test could not see a repaired/cleared swap.** Its fixture was 1 repaired + 1 cleared, so the summary read the same either way round. Now 1 repaired, 2 cleared, 3 untouched — both swap mutations fail it where both previously passed.
- **`main()`'s two new promises had zero test signal.** Deleting the `--revalidate`/`--force` guard outright left all 64 tests passing. Three tests now cover the refusal, the short-circuit above the storage/host checks, and the other direction. Also pinned: the incidental `try`/`finally` move — restoring the startup checks to their old position fails two of them.
- **No optimistic-concurrency guard on the per-row UPDATE** — deliberate wontfix, reasoned on the thread. `backfillAttachments` has the identical unguarded pattern, so guarding only the new pass would read as a difference that does not exist; and a bare `where('blurhash', stored)` would make the counts report a clear that did not happen. The reviewer's own trace found the blast radius empty: the one live path that overwrites an existing attachment's `blurhash` is owner-scoped local editing writing canonical values, which this pass skips without issuing an UPDATE.

Also in that commit: `docs/maintenance.md` now says the per-row log line is the only record of a cleared value (recoverable only if the attachment URL is still reachable) and pipes the apply example through `tee`; and `REVIEW.md` gained the matching checklist entries, which its own docs-hygiene rule asks for and my first commit missed.

Gate re-run green after the fixes: 894 files, 11907 tests.

**Round 2** — testing and scope/regression, pointed at the round-1 fixes. **Clean, no findings.** The scope lens validated the docs change from an angle I had not checked: the `2>&1` in the `tee` example is *necessary*, not merely correct, because the "clearing" line goes to stderr via `console.warn` while "repaired" goes to stdout — a plain `| tee` would drop exactly the line the prose calls the only record.

**Round 3** — correctness and quality over the full diff, since neither had seen the round-1 commit. Two findings, fixed in `ba7191cfe`:

- **A comment of mine was wrong in the dangerous direction.** It claimed `vi.importMock` returns an *empty namespace* for the partially-mocked module. Measured: it returns the **original** module — real, callable, a different object from the mock the module under test received — so `vi.mocked()` on it configures nothing and asserts nothing, and calling it runs the real implementation. That also made a standing repo-wide claim stale: `AGENTS.md` and `REVIEW.md` both said `vi.importMock` "always returns the mock".
- **The count signpost was asymmetric.** Round 1 added the two-vs-three disambiguation to `CLAUDE.md` but not to `AGENTS.md`'s canonical section or `docs/maintenance.md` — so the condensed pointer carried a clarification its own sources lacked.

**Round 4** — correctness and scope/regression on the round-3 fixes. Both lenses independently caught that my *correction* stated the wrong mechanism. Fixed in `e745adcc6`:

| factory | `vi.importMock` returns the mock? |
| --- | --- |
| sync | yes |
| async, no `importOriginal` parameter | yes |
| async, takes `importOriginal`, never calls it | yes |
| async, **awaits** `importOriginal()` | **no — returns the original** |

Awaiting the call is the trigger; async is only its precondition. The loose phrasing mattered because it over-reports: the scope lens grepped every other `vi.importMock` site in the repo and found three files that pair it with an `async` factory — none reaching `importOriginal` — which are correct as written and would have been flagged wrongly.

No other call site in the repo pairs `vi.importMock` with a factory that awaits `importOriginal`, so this PR surfaced no live bug elsewhere.

**Round 5** — correctness and quality on the round-4 fixes. Two low findings, fixed in `89a871a70`: `AGENTS.md` said the example file has "three awaiting" factories when it has four (wrong when first written), and the import-block comment enumerated two factory shapes that appear nowhere in that file, against the file's own convention of stating a rule once and pointing back.

That round also isolated the last two shapes I had asked about — calling `importOriginal()` without awaiting still returns the mock, and awaiting it returns the original whether the result is spread or discarded — so "awaits" is exactly the right word. A third shape exists (`vi.importMock` on a module nothing in the file's graph imports returns an empty stub regardless of factory) and is deliberately left undocumented: it applies to no call site in this repo.

The quality lens also answered whether five rounds of review had bloated the prose, by measuring: the new `AGENTS.md` blurhash section is ~3.6KB against the ~14.2KB storage-root rule and the ~28KB host-says-so rule it sits beside — terser than the repo's own bar for a comparable rule.

**Round 6** — scope/regression clean; the testing lens ran a fresh mutation set aimed at what earlier rounds had not tried and found **the best finding of the loop**: advancing the keyset cursor to `rows[0]` instead of `rows[rows.length - 1]` survived the entire suite for `backfillMedias` and `backfillAttachments`. Every paging test in the file ran at `batchSize: 1`, where both ends of a batch are the same row, so a trailing cursor was invisible to all of them. Fixed in `d0b5e898d` — see the [comment](https://github.com/llun/activities.next/pull/1597#issuecomment-5438947719) for the fixtures and why each is needed. The cursor was never wrong in shipped code; this is coverage the loops never had.

**Round 7** — testing and scope/regression on those fixes. **Clean.** The testing lens confirmed all four mutations now fail, that no new test passes for a stated-wrong reason (it checked each fixture claim against the source), and that removing the cursor advance entirely *hangs* rather than fails, as the file documents. It also ran the file under `--sequence.shuffle` to rule out state leaks. The scope lens was asked to challenge the decision to widen into two functions this PR does not otherwise change, and argued for keeping it here on grounds sharper than my own: the `lastId = rows[rows.length - 1]` idiom is shared across all three loops, and pinning the invariant on the new copy while leaving the two it was copied from unguarded would leave an inconsistent net over one idiom meant to behave identically everywhere.

**Round 8** — final correctness and quality passes over the whole PR. **Clean.** The correctness lens settled the one open question about the new pass having no per-row `try`/`catch` unlike its two siblings: `normalizeBlurhash` provably cannot throw for any input reaching it (`isBlurhashValid` wraps its own length check and returns `{result: false}`), so the only statement in the loop that can throw is the UPDATE itself — a systemic failure, not a row-specific one, where stopping is correct. It also confirmed the pass is safe to re-run after a kill: already-fixed rows re-classify as `untouched`, already-cleared rows re-clear to the same `null`.

**Two consecutive clean rounds (7 and 8) — loop complete**, at 8 rounds of the 20 cap. Seven commits; 11911 tests green.

## Final verification

Re-ran the script end to end against the local SQLite database on the finished code, re-seeded with one row of each shape:

```text
[DRY RUN] [attachments att-charset] blurhash "L6PZfSi_.AyE_3t7t7R**0o#Dg!4" is not one `decode` can read; would clear it
[DRY RUN] [attachments att-padded] blurhash "  L6PZfSi_.AyE_3t7t7R**0o#DgR4\n" is stored padded; would rewrite it as "L6PZfSi_.AyE_3t7t7R**0o#DgR4"
[DRY RUN] [attachments att-structural] blurhash "aaaaaa" is not one `decode` can read; would clear it
[DRY RUN] [attachments att-video] blurhash "\tL6PZfSi_.AyE_3t7t7R**0o#DgR4" is stored padded; would rewrite it as "L6PZfSi_.AyE_3t7t7R**0o#DgR4"
Blurhash revalidation complete: scanned 5, repaired 2, cleared 2, left 1 untouched
```

Dry run left the database byte-for-byte unchanged; the apply at `--batch-size 2` produced the same counts; the re-run reported `scanned 3, repaired 0, cleared 0, left 3 untouched` (idempotent); every surviving hash `decode`s cleanly; `--revalidate --force` exits 1.


scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1301,3 +1323,265 @@ describe('backfillMedias', () => {
...
+    await insertAttachment({ id: 'att-2', blurhash: STRUCTURAL_BLURHASH })
+    await insertAttachment({ id: 'att-3', blurhash: VALID_BLURHASH })
+    // Never selected: a missing blurhash is the backfill's job, and this pass
+    // has no way to produce one.
+    await insertAttachment({ id: 'att-4', blurhash: null })
+
+    await revalidateAttachmentBlurhashes(db, options({ batchSize: 1 }))
+
+    const expectedSummary =
+      'Blurhash revalidation complete: scanned 3, repaired 1, cleared 1, left 1 untouched'
```
llun: **testing lens — the fixture is symmetric, so this test cannot see a repaired/cleared swap.**

The fixture is 1 repaired + 1 cleared + 1 untouched, so `repaired 1, cleared 1` reads the same whichever way round the two counters are. Swapping `totalRepaired`/`totalCleared` — at the per-row increment or in the summary template — leaves this test green, even though its own comment says the three counts are the whole point of the pass.

Reproduced: the swap is caught today only incidentally, by `does not skip rows as it clears them out of its own predicate`, whose fixture happens to be all-clearable (0 repaired, 4 cleared). That is a paging test; generalise its fixture later and a swap ships silently.

Fix: make this fixture asymmetric so the counts alone distinguish the two.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -528,28 +541,161 @@ export const backfillAttachments = async (
...
   )
+}
+
+export const main = async () => {
+  const options = parseArgs(process.argv.slice(2))
+  // The two modes answer opposite questions — "recompute this from the image"
+  // against "repair what is already stored, touching no image" — and serving
+  // both off one invocation would make the cheap one's promise untrue. Refused
+  // rather than silently ordered, so an operator who asked for both finds out.
+  if (options.revalidate && options.force) {
```
llun, line 661: **testing lens — new documented behaviour in `main()` with zero test signal.**

This PR adds two user-facing promises to `main()`: the `--revalidate`/`--force` refusal (also stated in the script header, "Refuses to run together with --force") and the deliberate ordering that lets `--revalidate` short-circuit *before* the storage and host checks, which is what makes it usable on an instance whose storage is unreachable.

Reproduced independently: deleting this entire guard block leaves all 64 tests passing. An inverted condition, a dropped guard, or a reordering that makes revalidate require storage config again would all ship with no signal. `main()` being untested predates this PR, but the promises being tested are new.

Fix: a unit test around `main()` — stub the module boundaries and assert it throws on `{revalidate: true, force: true}`, and that the revalidate path never reaches `getMediaStorage`.

scripts/maintenance/backfillMediaBlurhash.ts
```diff
@@ -528,28 +541,161 @@ export const backfillAttachments = async (
...
+        totalRepaired += 1
+        console.log(
+          `${prefix}[attachments ${row.id}] blurhash ${JSON.stringify(stored)} is stored padded; ${options.dryRun ? 'would rewrite' : 'rewriting'} it as ${JSON.stringify(normalized)}`
+        )
+      }
+
+      if (!options.dryRun) {
+        await db('attachments')
+          .where('id', row.id)
+          .update({ blurhash: normalized })
```
llun, line 645: **destructive-op lens — no optimistic-concurrency guard on the per-row UPDATE.**

The write is keyed on `id` alone, with no `where('blurhash', stored)`, so nothing detects the row changing between the SELECT and this UPDATE while the instance is live.

The reviewer traced the blast radius and found it empty today: the only live path that overwrites an existing attachment's `blurhash` is `lib/database/sql/status.ts` via `updateNote`, which is owner-scoped local editing and always writes a canonical value copied from the media row — which this pass reads as `normalized === stored` and skips without issuing an UPDATE at all. `createNoteJob` only inserts for a new status, and `updateNoteJob` passes no attachments.

Raising it for an explicit decision rather than silent omission.

llun (commented):
Sub-agent review loop, round 1. Five lenses ran: correctness, scope/regression, quality, testing, and a destructive-operation lens. Correctness, scope/regression and quality returned no findings. The three below are from the testing and destructive-op lenses; each was reproduced before posting.

scripts/maintenance/backfillMediaBlurhash.test.ts
llun: Fixed in 6069fbc7b.

The fixture is now asymmetric — 1 repaired, 2 cleared, 3 untouched — so the three numbers are all different and the summary alone distinguishes them. Verified by mutation: swapping the counters in the summary template, and swapping the per-row increments, each now fail 2 tests where both previously passed.

The comment above the fixture now says why the numbers differ, so a future edit does not quietly make it symmetric again.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Fixed in 6069fbc7b — three tests in a new `main` block.

- `refuses --revalidate beside --force, before opening a connection` — also asserts `getKnex` was never called, so the refusal cannot regress into one that opens a pool first.
- `revalidates without asking for a storage backend or a host` — asserts neither `getMediaStorage` nor `getConfig` is reached, which is the actual promise, and reads the outcome off the summary line since `main` has already destroyed the connection by then.
- `still requires the storage backend when not revalidating` — the other direction, so a short-circuit that swallowed the ordinary run cannot pass.

Verified by mutation, each failing at least one test where all 64 previously passed: guard deleted outright, `&&` inverted to `||`, and the revalidate branch made unreachable so it falls through to the storage/host checks.

The third test also pins the incidental `try`/`finally` move that the scope lens flagged as untested on both sides of the diff — it asserts `db.destroy()` runs when a startup check throws. Restoring those checks to their old position outside the `try` fails two of the three.

`main`'s three module boundaries are mocked with the async-`importOriginal` spread already used for `imageAnalysis`, so only the entry points `main` calls are replaced and nothing outside the new block changes behaviour.

scripts/maintenance/backfillMediaBlurhash.ts
llun: Wontfix, deliberately — thank you for tracing the blast radius rather than asserting it.

Three reasons, in order of weight:

1. **It would make the file inconsistent in a way that reads as significant.** `backfillAttachments`' own final `.update()` has the identical unguarded pattern, over four columns rather than one. Guarding only the new pass leaves a reader to infer a difference between the two that does not exist — the "edited one at a time and drift silently" failure mode `REVIEW.md` warns about for the two storage drivers. Guarding both is scope creep into code this PR does not otherwise touch.

2. **A bare `where('blurhash', stored)` would make the counts lie.** The counters increment on the decision, before the write. If a guarded UPDATE then matched nothing, a row counted as `cleared` would not have been cleared, and the summary an operator reads would be wrong — worse than the race it guards. Doing it properly means a fourth count for a case with no live trigger, which is machinery the problem does not have.

3. **The file's convention is to add a count for an observed cause.** The two never-summed counts in `backfillAttachments` exist because both causes were seen in production data.

Worth restating your own finding for the record, since it is the part that makes this safe: the only live path that overwrites an existing attachment's `blurhash` is owner-scoped local editing via `updateNote`, and it always writes a canonical value copied from the media row — which this pass reads as `normalized === stored` and skips without issuing an UPDATE at all. `createNoteJob` only inserts for a new status, and `updateNoteJob` passes no attachments. So the rows this pass actually writes are federated ones no live path touches.

If a "re-sync attachments on Update" feature is ever built, this becomes reachable, and the right fix then is to guard both call sites together.

CLAUDE.md
```diff
@@ -54,7 +54,8 @@
...
... when the attachment already carries a BlurHash**: `lib/components/posts/media.tsx` gates the `<img>` at `opacity-0` until `onLoad` only inside `if (blurhash)`, and a falsy one renders a bare `<img>` — a broken-image icon nothing can repair once the bytes are gone. Nulling `mediaId` would be a regression, not a cleanup: a NULL marks a FEDERATED attachment and `isReplaceableMediaAttachment` reads it to decide the attachment is not the owner's to replace, so a broken attachment the author can still remove by editing the post would become permanent (pinned by `clears only editable media while preserving legacy and fitness attachments` in `lib/database/sql/status.test.ts`). Deleting the attachment row instead removes the promised placeholder and silently rewrites a published status whose federated copies keep it. `scripts/maintenance/backfillMediaBlurhash.ts` reports the residue as **two counts that are never summed** — a gone media row is nothing to act on, an unparseable `mediaId` is a bad write — and neither partitions `processed`. (Not to be confused with that script's `--revalidate` mode, whose three counts DO partition its scan.) See **Deleting Media a Post Uses** in `AGENTS.md`.
```
llun, line 57: **quality lens (round 3) — the disambiguation landed on the pointer but not on the sources it points to.**

Round 1 added a parenthetical to this bullet distinguishing `backfillAttachments`' two never-summed counts from `--revalidate`'s three partitioning ones. It was not added to either place this bullet points at:

- `AGENTS.md` → "Deleting Media a Post Uses", the canonical text this bullet cites, still says only "two counts, never summed".
- `docs/maintenance.md` → "Attachments whose `mediaId` resolves to nothing", which is the operator-facing one and the likeliest place to actually cause the confusion — an operator reading that page top to bottom meets the two-count explanation first and the three-count one about twenty lines later, with no signpost between them.

That inverts the usual relationship: `CLAUDE.md` is the condensed pointer, the other two are the sources of truth. Nothing is factually wrong; the add was just asymmetric.

Confirmed: `grep -c "Not to be confused with that script"` returns 1 for `CLAUDE.md`, 0 for `AGENTS.md` and `docs/maintenance.md`.

llun (commented):
Sub-agent review loop, round 3 (correctness + quality over the full two-commit diff, since neither had seen the round-1 fix commit). Round 2 was clean. One finding.

CLAUDE.md
llun: Fixed in ba7191cfe. Both `AGENTS.md` → "Deleting Media a Post Uses" and `docs/maintenance.md` → "Attachments whose `mediaId` resolves to nothing" now carry the signpost, the latter as a linked cross-reference to the `--revalidate` section further down the same page.

Round 3 also produced a second finding, on the same commit, which I'll record here since it has no thread of its own — the correctness lens probed a comment I wrote in `backfillMediaBlurhash.test.ts` and found it false in the more dangerous direction.

The comment claimed `vi.importMock` returns an *empty namespace* for the partially-mocked `imageAnalysis` module. Measured across all four mocks in that file:

| factory | `vi.importMock` returns | is it the mock? | same object as the static import? |
| --- | --- | --- | --- |
| sync (`safeImageFetch`) | the factory's result | yes | yes |
| async + `importOriginal` (the other three) | the **original** module | **no** | **no** |

So with an async factory the export comes back real, callable, and a different object from the one the module under test was given: `vi.mocked()` on it configures nothing and asserts nothing, and calling it runs the real implementation. Worse than a missing binding, because it looks right.

That also makes a standing project claim stale — `AGENTS.md` and `REVIEW.md` both say `vi.importMock` "always returns the mock", which holds for a sync factory only, and a partial mock always has an async one. Both now carry the caveat, since reaching `importOriginal` is the normal way to keep one export real.

AGENTS.md
```diff
@@ -2048,6 +2051,22 @@ preserving legacy and fitness attachments` pins the surviving-null behaviour.
   is separately `vi.mock`'d, and forces a type-erasing double-cast. `vi.importMock`
   **is** a valid, documented Vitest API — some review bots incorrectly claim it
   does not exist; do not "fix" it on their say-so.
+- **"Always returns the mock" stops holding the moment a factory AWAITS
```
llun, line 2054: **correctness lens (round 4) — my round-3 correction stated the wrong mechanism, and was fixed in e745adcc6.**

I had written the rule as sync-vs-async. That is wrong, and wrong in a way that matters: it points the rule away from the thing that actually breaks and sends a reader to a static import they do not need.

Re-measured here across all four shapes:

| factory | `vi.importMock` returns the mock? |
| --- | --- |
| sync | yes |
| async, no `importOriginal` parameter | yes |
| async, takes `importOriginal`, never calls it | yes |
| async, **awaits** `importOriginal()` | **no — returns the original module** |

So awaiting the call is the trigger; being async is only its precondition. All three places now say that, and `REVIEW.md` says explicitly not to flag a factory merely for being async, because that heuristic over-reports.

The scope lens independently confirmed there is no live bug elsewhere: no other `vi.importMock` call site in the repo pairs with a factory that awaits `importOriginal`. Three files (`lib/components/post-box/reply-preview.test.tsx`, `lib/services/accounts/registerAccount.test.ts`, `lib/services/notifications/sendNotificationAlerts.test.ts`) do combine `vi.importMock` with an `async` factory, but none reaches `importOriginal`, so they are correct as written — which is exactly why the loose "async" phrasing had to go rather than stay as a conservative over-warning.

One correction to that lens's own report, for the record: it stated `CLAUDE.md` does not mention `backfillMediaBlurhash` or `--revalidate`. The project `CLAUDE.md` does, four times, including the count signpost added in the previous round — it appears to have read the global file. Its conclusion (nothing there to contradict) is unaffected.

llun (commented):
Sub-agent review loop, round 4 (correctness + scope/regression on the round-3 fixes). One finding, from both lenses independently — fixed in e745adcc6 and recorded below.

AGENTS.md
```diff
@@ -2048,6 +2051,22 @@ preserving legacy and fitness attachments` pins the surviving-null behaviour.
...
+  it all return the factory's result; only awaiting it diverges. Then
+  `vi.importMock` hands back the **original** module — the export comes back
+  real, callable, and a DIFFERENT object from the one the module under test was
+  given, so `vi.mocked()` on it configures nothing and asserts nothing, and
+  calling it runs the real implementation. Worse than a missing binding,
+  because it looks right. Read a partial mock — the
+  `{ ...(await importOriginal()), fn: vi.fn() }` shape — through a plain static
+  import, which does resolve to the mock.
+  `scripts/maintenance/backfillMediaBlurhash.test.ts` has one of each: a sync
+  factory read with `vi.importMock`, and four awaiting ones read statically.
```
llun, line 2069: **correctness lens (round 5) — the example file has four awaiting factories, not three.**

`@/lib/config`, `@/lib/database`, `@/lib/services/medias` and `@/lib/services/medias/imageAnalysis` all use `async (importOriginal) => ({ ...(await importOriginal()), ... })`. `grep -c "await importOriginal"` on that file returns 4, and the miscount contradicted the file's own "These four are read back through the static import" comment. Wrong when first written, not introduced by the previous round.

Fixed in 89a871a70.

The round also strengthened the measurement behind the rule without changing it — two shapes I had asked to be isolated:

| factory | `vi.importMock` returns the mock? |
| --- | --- |
| async, **calls** `importOriginal()` but does not await | yes |
| async, **awaits** it and discards the result | no — original |
| async, awaits it and spreads the result | no — original |

So "awaits" is exactly the right word: not merely calling, and it makes no difference whether the awaited value is used. The lens also validated the "do not flag merely for async" instruction against a live example rather than a synthetic one — `app/api/v1/statuses/[id]/route.test.ts` pairs `vi.importMock` with an `async` factory that never reaches `importOriginal`, and reconfiguring the returned `getConfig` demonstrably drives the route handler, which only works if the mock came back.

One further shape it surfaced is deliberately **not** documented: `vi.importMock` on a module nothing in the file's graph ever imports returns an empty stub regardless of factory. It applies to no call site in this repo, and growing the rule to cover it is the accretion the quality lens is warning about in the next thread.

scripts/maintenance/backfillMediaBlurhash.test.ts
```diff
@@ -1,6 +1,20 @@
 import knex, { Knex } from 'knex'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
+// These four are read back through the static import, never `vi.importMock`:
```
llun, line 4: **quality lens (round 5) — this comment enumerated factory shapes that do not exist in this file.**

It listed "an async one with no `importOriginal` parameter" and "an async one that takes it and never calls it". This file has exactly two shapes: one sync factory (`safeImageDownload`) and four that await `importOriginal`. That taxonomy is the canonical `AGENTS.md` bullet's job, and the file's own convention two lines below — "for the reason given there" — is to state a rule once and point back. A newcomer had to read past cases irrelevant to this file to reach the part that matters.

Trimmed in 89a871a70 to the local reason plus a citation, 11 lines to 9.

Recording the rest of that lens's verdict, since it is the answer to whether five rounds of review have bloated the prose: it measured the new `AGENTS.md` blurhash section at ~3.6KB against the two comparable rules it sits beside — the storage-root rule at ~14.2KB and the host-says-so rule at ~28KB — and found the new one markedly terser than the repo's own bar. The `revalidateAttachmentBlurhashes` JSDoc matches the density of `downloadRemoteImage`'s, which predates this PR. It also confirmed no orphaned copies of the old "sync vs async" framing survive anywhere, and found nothing load-bearing left undocumented.

llun (commented):
Sub-agent review loop, round 5 (correctness + quality on the round-4 fixes). Two low findings, both real, both fixed in 89a871a70.

llun:
### Review loop, round 6 — a surviving mutation, and the best finding of the loop

Scope/regression: clean. The testing lens ran a fresh mutation set aimed at what earlier rounds had not tried, and found a genuine coverage hole. Fixed in d0b5e898d.

**The mutation.** Advancing the keyset cursor to `rows[0]` instead of `rows[rows.length - 1]` survived the **entire** suite — at `backfillMediaBlurhash.ts:265` (`backfillMedias`) and `:392` (`backfillAttachments`). I reproduced it before acting.

**Why nothing saw it.** Every paging test in the file ran at `batchSize: 1`, where `rows[0]` and `rows[rows.length - 1]` are the same row. The multi-row tests use the default batch size of 50, so they never cross a batch boundary. A trailing cursor was invisible to all of them.

**The fixture that separates them** is three rows at a batch size of two — and it has to be chosen so nothing *else* masks the bug, which is the part worth writing down:

| loop | what keeps the fixture honest |
| --- | --- |
| `backfillMedias`, `backfillAttachments` | `--force`, which removes the selection predicate — otherwise a row's own write drops it out of the next batch's `WHERE` and hides the trailing cursor |
| `revalidateAttachmentBlurhashes` | all-canonical hashes, because an untouched row is never written, so only the cursor decides what the next batch reads |

A trailing cursor then re-reads the middle row and the count reads 5 for 3 rows.

`revalidateAttachmentBlurhashes` did already catch this — but only by accident of the `--dry-run` test never writing, the same caught-by-luck shape round 1 flagged in the counts test. It gets an aimed test rather than keeping the incidental one.

Also closed: `backfillMedias` never had the SELECT-counting limit test its two sibling loops carry, so a hardcoded batch size survived there too — three rows being indistinguishable in one batch or two by result.

All four mutations now fail, individually and applied together. **The cursor was never wrong in shipped code** — this is coverage the loops never had.

**On scope:** two of the three fixes are in functions this PR does not otherwise change. I judged that in scope rather than deferring it — it is test-only, it is the file this PR already substantially reshapes, and the repro was in hand. Leaving a proven, reproduced hole behind in a file I had just spent six rounds on seemed worse than the small widening.

PR #1606, opened 2026-08-27
feat: link the boosted-by username to the booster's profile

llun:
## What

The `Boosted by <name>` line above a boost rendered the booster's name as plain text. That made the one actor named on that row the only actor in the post header with no way to open their profile — the avatar and the display name beside it have both been links since the post components were written.

The name is now a link to `/@user@domain`.

## How

- `getActorProfileHref` is extracted from `ActorAvatar` in `lib/components/posts/actor.tsx` so the href has one derivation. An actor profile resolves to its mention; an actor whose username normalises to empty, or no actor at all, falls through to the handle recovered from the actor id; an id carrying no usable handle either (an opaque `did:`/UUID username) resolves to nothing — which `BoostStatus` renders as **plain text** rather than as a link to nowhere.
- `getActorDisplayName` sits beside it, so the link's text and its destination come from the same place. A federated `preferredUsername` is untrusted (`z.string()`, stored verbatim), so `@alice` and `booster@attacker.example` both render as `alice`/`booster` rather than putting an attacker-controlled string in a link's accessible name.
- `prefetch={false}`, like its two neighbours: a feed renders this row once per boost, and `/@user@domain` is a fully dynamic route that WebFingers and signs an actor fetch out to a remote server for an actor this instance has not persisted (see **Link prefetching in feeds** in `AGENTS.md`).
- `font-medium text-foreground hover:underline` — the row is `text-muted-foreground`, so without an explicit colour the link had no at-rest affordance at all. This matches `CollectionDetail.tsx:437` and `NotificationItem.tsx:134`, the two other places a person's name sits inside a muted line.

## Tests

`lib/components/posts/boost-status.test.tsx` (new) plus additions to `actor.test.tsx` and rewritten boost assertions in `post.test.tsx`. Every production change in this PR was put through a revert matrix — each is caught by a named test that fails for the right reason:

| Reverted | Caught by |
| --- | --- |
| the `<Link>` wrapper | 9 tests |
| `prefetch={false}` | 2 |
| `text-foreground` | 1 |
| `getActorDisplayName`'s normalisation | 3 |
| `getActorProfileHref`'s empty-username guard | 2 |

The negative assertion this PR replaced in `post.test.tsx` (`queryByRole('link', { name: 'Booster' })` absent, from #621) was guarding that the post **header** belongs to the boosted post's author rather than the booster; that intent is preserved by asserting the two links resolve to two different profiles.

## Verification

Full gate green: `prettier --check`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` (898 files / 12,000 tests).

Checked in a browser against a local SQLite instance with the mock user, in both themes. Light: label `rgb(110,110,110)`, link `rgb(10,10,10)`. Dark: label `rgb(163,163,163)`, link `rgb(250,250,250)` on `rgb(10,10,10)`. Underlines on hover, navigates on click, no console errors.

## Docs

`AGENTS.md` → **Link prefetching in feeds**: the boosted-by link and the new test file are added to the opt-out list and its named regression tests.

## Review

The sub-agent code-review loop ran: **6 rounds**, ending on two consecutive clean rounds.

Five lenses (correctness, design-system, quality, scope-regression, testing) reviewed each round's diff, every finding was put through an adversarial verifier before being posted, and each surviving finding was fixed, replied to and resolved. The verifier refuted 5 findings that would otherwise have been acted on — including a proposal to swap `parentElement` for `container` in three assertions, which was shown to pass wrongly when the link is detached from its label.

What the loop actually caught, in order: no at-rest contrast on the link; the raw federated username used as link text while the href normalised it; the wrong layer shared (`getDisplayUsername` rather than the `getActorDisplayName` composition); an href built from a username that normalised to empty, which made a broken destination *look* correct; a test exercising only half of `getDisplayUsername`; two untested changes; and three comments that were confidently wrong about their own truth table.

Three round-6 findings were recorded as explicit wontfix with justification on their threads (a redundant-but-named prefetch test, an unreachable `!actorId` branch, and British/American spelling in prose).

**Deferred to a follow-up:** `ActorInfo`'s actor-present branch still links unconditionally, so a degenerate username renders a broken empty-text link there. That branch is byte-for-byte unchanged from `main` — the bug predates this PR — and converging `ActorInfo`/`ActorAvatar` onto the shared helpers needs a third fallback branch and its own test matrix.


lib/components/posts/post.tsx
```diff
@@ -91,11 +97,35 @@ export const BoostStatus: FC<BoostStatusProps> = ({ status }) => {
...
+          // in `actor.tsx` carry: `prefetch={false}`, because a feed renders
+          // this row once per boost and `/@user@domain` is a fully dynamic
+          // route that federates out for a remote actor this instance has not
+          // persisted (see "Link prefetching in feeds" in AGENTS.md), and the
+          // same click guard, so opening a profile never doubles as a click on
+          // whatever surface embeds the post.
+          <Link
+            href={profileHref}
+            prefetch={false}
+            className="font-medium hover:underline"
```
llun: **The booster link has no at-rest visual affordance.**

The `<Link>` is `font-medium hover:underline` and sits inside a row that is `text-sm text-muted-foreground` (post.tsx:105). No colour is set on the link, and `app/globals.css` has no base `a { color: … }` rule outside `.markdown-content a`, so it inherits the muted grey and is distinguished at rest only by a weight bump from 400 to 500. `hover:underline` gives nothing at all on touch.

The codebase resolves this exact shape — a person's name embedded in a muted line — by adding `text-foreground` on the link:

- `app/(timeline)/collections/[id]/CollectionDetail.tsx:432-442` — "Curated by {'{'}<Link className="font-medium text-foreground hover:underline">{'}'}", inside `<p className="px-1 text-xs text-muted-foreground">`.
- `app/(timeline)/notifications/NotificationItem.tsx:130-139` — `<Link … className="font-semibold text-foreground hover:underline">` inside `<span className="text-muted-foreground">`.

The two other links in `post.tsx` that sit in a muted row (the fitness-file download link, line ~297, and the fitness source link, line ~387) instead use `text-muted-foreground underline-offset-2 hover:text-foreground hover:underline`. Either treatment gives an at-rest cue; the shipped one gives none.

Suggested fix: `className="font-medium text-foreground hover:underline"`.

(Verified against the actual cascade, not just class names. Counter-example noted for completeness: the relative-timestamp button at post.tsx:449 is also muted with only `hover:underline` — but it is a metadata token, not an identity, and the two closest structural analogues both add `text-foreground`.)

lib/components/posts/post.tsx
```diff
@@ -91,11 +97,35 @@ export const BoostStatus: FC<BoostStatusProps> = ({ status }) => {
     status.actor?.name ||
     status.actor?.username ||
```
llun: **The link's text uses the raw `username`; its href normalises it. They can disagree.**

`actorName` is `status.actor?.name || status.actor?.username || getActorIdMention(status.actorId)` — the RAW `username`. `ActorInfo` uses `actor.name || getDisplayUsername(actor.username)` for the identical field (`actor.tsx:244`), and the third fallback here, `getActorIdMention`, normalises too (via `getActorIdParts` → `getDisplayUsername`). The middle branch is the only unnormalised one.

`ActorProfile.username` is federated input stored verbatim: `lib/types/activitypub/actor.ts:31` declares `preferredUsername: z.string()` with no shape check, `recordActorIfNeeded` (`lib/actions/utils.ts:126`) writes it straight through `createActor`, and nothing normalises on read. `name` is genuinely optional upstream. So a remote actor with `preferredUsername: '@alice'` (or `'alice@example.com'`) and no display name renders as `Boosted by @alice` — as the visible text AND the accessible name of a focusable link — while `getActorProfileHref` two lines below builds the href correctly through `getActorMention`, which does call `getDisplayUsername`.

This predates the PR, but the PR is what promotes the string from inert text into a link's accessible name, and it undercuts the PR's own stated goal of one shared derivation. In scope: it is the same variable, in the same function, that the new link consumes.

Suggested fix — export `getDisplayUsername` from `lib/components/posts/actor.tsx` (module-private today, line 33) and spell the chain so the existing fallback semantics survive an empty result:

```ts
const actorName =
  status.actor?.name ||
  (status.actor && getDisplayUsername(status.actor.username)) ||
  getActorIdMention(status.actorId)
```

`getDisplayUsername` is a no-op for every well-formed handle (Mastodon handles never contain `@`), so this only changes the malformed case. Add a `boost-status.test.tsx` case for "actor present, no `name`, `username: '@booster'`" — that branch is currently untested.

llun (commented):
Automated review: 2 verified findings on the new boosted-by profile link.

lib/components/posts/post.tsx
llun: Fixed in c4ce33532. The link is now `font-medium text-foreground hover:underline`, matching `CollectionDetail.tsx:437` and `NotificationItem.tsx:134` — the two other places a person's name sits inside a `text-muted-foreground` line. Confirmed first that `app/globals.css` carries no base `a { color: … }` rule outside `.markdown-content a`, so nothing else was supplying a resting colour.

lib/components/posts/post.tsx
llun: Fixed in c4ce33532. `getDisplayUsername` is now exported from `actor.tsx` and the chain reads `status.actor?.name || (status.actor && getDisplayUsername(status.actor.username)) || getActorIdMention(status.actorId)`, so an empty result still falls through to the id-derived mention exactly as before. Added the missing case to `boost-status.test.tsx` (actor present, no `name`, `username: '@booster'`) asserting the link text is `booster` and the href `/@booster@remote.example`. Proved it guards the fix: reverting the `actorName` line alone fails that test with the only link present carrying the accessible name `@booster`.

llun (commented):
Round 2 review: 4 verified findings inline.

lib/components/posts/actor.tsx
```diff
@@ -149,6 +149,21 @@ const getActorIdParts = (
...
+// avatar below, and the boosted-by line in `post.tsx` — share this instead of
+// re-deriving it. Undefined when the actor id carries no usable handle (an
+// opaque `did:`/UUID username), which is the case those callers render as
+// plain text rather than as a link to nowhere.
+export const getActorProfileHref = (
+  actor?: ActorProfile | null,
+  actorId?: string,
+  statusUrl?: string | null
+): string | undefined => {
+  if (actor) return `/${getActorMention(actor)}`
```
llun: **The round-2 fix made the text/href mismatch harder to spot for a degenerate username.**

`getActorProfileHref` is unconditional on the actor being usable — `if (actor) return '/' + getActorMention(actor)` — and never falls back to the actor-id path, so it will build an href out of a username that normalised to nothing. `getDisplayUsername` collapses to `''` only when the username is entirely `@` characters.

For `actor = { name: '', username: '@', domain: 'remote.example' }` with `actorId = 'https://remote.example/users/booster'`, verified by running both commits:

- Before this commit: text `@`, href `/@@remote.example`. Broken, but self-evidently so.
- After this commit: `getDisplayUsername('@')` is `''`, falsy, so `actorName` falls through to `getActorIdMention` → text `@booster@remote.example`, which looks like a perfectly good mention, while the href is still `/@@remote.example`. `parseAccountHandle` rejects an empty username, so that href 404s.

So the fix removed the visible signal while leaving the destination as broken as it was. `preferredUsername` is `z.string()` with no shape check (`lib/types/activitypub/actor.ts:31`) and `recordActorIfNeeded` writes it verbatim (`lib/actions/utils.ts:126`), so the input is reachable — and this commit's own rationale is that malformed `preferredUsername` is a realistic case.

Fix: make `getActorProfileHref` fall through to the actor-id path when the actor's username normalises to empty, so the href derives from the same place the text does. Then both read `/@booster@remote.example` for this input.

lib/components/posts/post.tsx
```diff
@@ -89,13 +96,37 @@ export const BoostStatus: FC<BoostStatusProps> = ({ status }) => {
   if (status.type !== StatusType.enum.Announce) return null
   const actorName =
     status.actor?.name ||
-    status.actor?.username ||
+    (status.actor && getDisplayUsername(status.actor.username)) ||
```
llun: **The fix exported the wrong layer, leaving a third copy of one composition.**

What every caller wants is the composition `name || getDisplayUsername(username)`, not the primitive. That composition already existed twice in `actor.tsx` — `getInitials(actor.name || getDisplayUsername(actor.username) || '')` (~line 187) and `{actor.name || getDisplayUsername(actor.username)}` (~line 244) — and this is now a third copy. The `(status.actor && …)` operand exists only because `status.actor?.name` does not narrow `status.actor` for the next operand; confirmed under this repo's `tsc --strict`, where the un-guarded spelling is `TS18047: 'myStatus.actor' is possibly 'null'`.

`actor.tsx` already has the idiom for this: `getActorProfileHref` exists precisely so callers do not re-derive the href. Its sibling is missing.

Fix — add beside `getActorProfileHref`:

```ts
export const getActorDisplayName = (
  actor?: ActorProfile | null
): string | undefined =>
  actor ? actor.name || getDisplayUsername(actor.username) : undefined
```

and let `BoostStatus` read `getActorDisplayName(status.actor) || getActorIdMention(status.actorId)`. Verified behaviourally identical to the shipped chain in both branches (associativity of `||`), type-clean at all three call sites, and a net-smaller diff — three changed lines in `post.tsx` become one. Retrofitting `ActorInfo`/`ActorAvatar` onto it is a legitimate but separable follow-up; do not do it here.

lib/components/posts/boost-status.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+  it('does not prefetch the booster profile link', () => {
+    render(<BoostStatus status={boost} />)
+
+    expect(screen.getByRole('link', { name: 'Booster' })).toHaveAttribute(
+      'data-prefetch',
+      'false'
+    )
+  })
+
+  it('normalises a malformed federated username before rendering it as the link text', () => {
```
llun: **The new case exercises only half of `getDisplayUsername`.**

Proven by mutation: weakening the helper to `username.replace(/^@+/, '')` — dropping `.split('@')[0]` — leaves this file plus `actor.test.tsx` at 10/10 passing and all 353 tests under `lib/components/posts/` green. The fixture `'@booster'` has no embedded `@`, so the split half is unguarded, and no fixture anywhere under `lib/components/posts/*.test.tsx` has a non-leading `@`.

It matters: with the split half regressed, `preferredUsername: 'booster@attacker.example'` renders an attacker-controlled string verbatim as the link text (`booster@attacker.example`) with href `/@booster@attacker.example@remote.example` — a 404 rather than a cross-domain navigation, so the harm is UI text-spoofing, not an open redirect. That is exactly the disagreement between text and destination this commit set out to close.

Fix: add a case (or make it a two-row `it.each` with the existing one) using `username: 'booster@attacker.example'` and asserting the link text is `booster` and the href `/@booster@remote.example`.

lib/components/posts/post.tsx
```diff
@@ -89,13 +96,37 @@ export const BoostStatus: FC<BoostStatusProps> = ({ status }) => {
...
+          // in `actor.tsx` carry: `prefetch={false}`, because a feed renders
+          // this row once per boost and `/@user@domain` is a fully dynamic
+          // route that federates out for a remote actor this instance has not
+          // persisted (see "Link prefetching in feeds" in AGENTS.md), and the
+          // same click guard, so opening a profile never doubles as a click on
+          // whatever surface embeds the post.
+          <Link
+            href={profileHref}
+            prefetch={false}
+            className="font-medium text-foreground hover:underline"
```
llun, line 121: **The at-rest contrast fix has no test.**

Reverting this line to `font-medium hover:underline` leaves all 353 tests under `lib/components/posts/` passing. Adding `text-foreground` was the whole point of the round-1 fix, and nothing guards it: `app/globals.contrast.test.ts` pins token VALUES, never their usage, and jsdom computes no cascaded style from the Tailwind build, so `toHaveClass` is the only available mechanism.

Asserting a class token is the established convention in this exact file — `post.test.tsx` already asserts `-ml-13`, `ml-auto`, `gap-1`, `right-0` and others as guards for load-bearing layout decisions. Unlike `ImageUploadField.test.tsx:109-111` there is no cascade collision here, so a single positive `expect(link).toHaveClass('text-foreground')` in `boost-status.test.tsx` is enough.

lib/components/posts/actor.tsx
llun: Fixed in bb5fb0cd8. `getActorProfileHref`'s first branch is now `if (actor && getDisplayUsername(actor.username))`, so an actor whose username normalises to empty falls through to the actor-id path and the href derives from the same handle the label does. For `username: '@'` both now read `/@booster@remote.example`. Pinned by a new test; reverting the guard alone fails it with `href="/@@remote.example"`.

lib/components/posts/post.tsx
llun: Fixed in bb5fb0cd8. `getActorDisplayName(actor)` now lives beside `getActorProfileHref`, and `BoostStatus` reads `getActorDisplayName(status.actor) || getActorIdMention(status.actorId)` — three lines become one, and the `(status.actor && …)` type-checker workaround is gone. `getDisplayUsername` is un-exported again, since nothing outside `actor.tsx` needs the primitive now. `ActorInfo`/`ActorAvatar` are deliberately left alone, as agreed.

lib/components/posts/boost-status.test.tsx
llun: Fixed in bb5fb0cd8. The case is now a two-row `it.each` — `'@booster'` (leading `@`) and `'booster@attacker.example'` (embedded `@`) — both expecting link text `booster` and href `/@booster@remote.example`. Reverting the normalisation fails both rows.

lib/components/posts/post.tsx
llun: Fixed in bb5fb0cd8. `boost-status.test.tsx` now asserts `toHaveClass('text-foreground')` on the booster link. Reverting the class fails it with `received: "font-medium hover:underline"`.

lib/components/posts/post.tsx
```diff
@@ -88,14 +95,36 @@ interface BoostStatusProps {
 export const BoostStatus: FC<BoostStatusProps> = ({ status }) => {
   if (status.type !== StatusType.enum.Announce) return null
   const actorName =
-    status.actor?.name ||
-    status.actor?.username ||
-    getActorIdMention(status.actorId)
+    getActorDisplayName(status.actor) || getActorIdMention(status.actorId)
+  // Undefined when the boost actor id carries no usable handle (an opaque
+  // `did:`/UUID username) — the same case `ActorInfo` renders as plain text.
```
llun: **This comment is now false, and it was this PR that made it so.**

It reads: "Undefined when the boost actor id carries no usable handle (an opaque `did:`/UUID username) — the same case `ActorInfo` renders as plain text."

The `ActorInfo` half is wrong. Verified by rendering it: for `actor = { name: '', username: '@', domain: 'remote.example' }` with an opaque actorId, `getActorProfileHref` returns `undefined` (so `BoostStatus` renders plain text), while `ActorInfo` given the same actor takes its actor-present branch — which renders a `<Link>` unconditionally — and produces:

```html
<a href="/@@bsky.brid.gy"></a><span>@@bsky.brid.gy</span>
```

a broken link with EMPTY visible text, not plain text. The claim is true only for `actor === null` (which is what the existing `renders plain text when the actor id carries no usable handle` test covers, with `actor: null`).

Round 2's guard is what opened this second route to `undefined`, so the comment written before it is no longer accurate.

Fix: drop the `ActorInfo` cross-reference, or scope it explicitly to the no-actor-row case. Do not "fix" it by changing `ActorInfo` — see the companion finding on `actor.tsx`'s `getActorDisplayName` doc comment.

lib/components/posts/actor.tsx
```diff
@@ -157,6 +178,17 @@ export const getActorIdMention = (
   return `${handle}${domain}`
 }
 
+// The name a caller shows beside (or instead of) `getActorProfileHref`'s
+// destination — the one composition (`name || getDisplayUsername(username)`)
+// `ActorAvatar`'s initials and `ActorInfo`'s display name already use inline.
+// Undefined when there is no actor to name, mirroring `getActorProfileHref`'s
+// own `undefined` for "nothing to link"; a caller with a fallback of its own
+// (`BoostStatus` reads the actor id instead) chains it with `||`.
```
llun, line 189: **The comment promises a consolidation the PR deliberately does not make.**

It says the helper is "the one composition (`name || getDisplayUsername(username)`) `ActorAvatar`'s initials and `ActorInfo`'s display name already use inline" — but neither was converted, so `actor.tsx` carries that composition three times (this helper, `ActorAvatar`'s initials at ~line 204, `ActorInfo`'s label at ~line 261).

That deferral is the right call and should stand. Verified: `ActorInfo`'s actor-present branch is byte-for-byte unchanged from `main` — the divergence it causes for a degenerate username predates this PR entirely, so "fix what you broke" does not apply. And migrating it is not a substitution: `getActorDisplayName(actor)` alone still returns `''` for that actor, so a real fix needs a third branch falling through to the actor-id-derived handle, in a component rendered on every post surface, with its own test matrix. That is its own PR.

Fix here: reword the comment so it describes what the helper IS rather than implying the other two already call it. The migration should be filed as a follow-up.

llun (commented):
Round 3 review findings (2 inline; a third targets an unchanged file and is posted as a top-level comment separately).

llun:
**Round 3 finding — `lib/components/posts/actor.test.tsx` (posted here as a top-level comment because this PR does not modify that file, so no inline diff anchor is available on it).**

**`ActorAvatar`'s behaviour under the new guard is untested.**

Round 2 changed the SHARED `getActorProfileHref`, and `ActorAvatar` calls it. Confirmed by running both versions with a realistic `actor` + `actorId` pair and `username: '@'`:

- pre-guard: avatar href `/@@remote.example`
- post-guard: avatar href `/@booster@remote.example`

Nothing covers it. `actor.test.tsx` has four tests, all using the ordinary `username: 'test'` fixture, and `boost-status.test.tsx` only exercises `BoostStatus`. Unlike the two inline findings in this round, this gap is NOT pre-existing — the guard is new in this PR, `ActorAvatar` observably changed because of it, and a revert of the guard would be caught only through the boosted-by line.

Fix: add a case to `actor.test.tsx` — `ActorAvatar` with `actor` (username `'@'`) plus an `actorId`, asserting the href falls back to the actor-id-derived path.

llun:
Addressed in e97ff9a3b. `actor.test.tsx` now covers `ActorAvatar` with `actor` (`username: '@'`) plus an `actorId`, asserting the href falls back to `/@booster@remote.example`. Proved it guards the guard: reverting `getActorProfileHref`'s condition to a bare `actor` fails the new case with `href="/@@llun.test"`.

lib/components/posts/post.tsx
llun: Fixed in e97ff9a3b. The comment now scopes the `ActorInfo` parallel to the no-actor-row case and states explicitly that `ActorInfo`'s actor-present branch renders a link unconditionally, so it does not share this fallback. `ActorInfo` itself is untouched — see the thread on `getActorDisplayName` for why that migration is a follow-up.

lib/components/posts/actor.tsx
llun: Fixed in e97ff9a3b. The comment now says `ActorAvatar` and `ActorInfo` compute the same composition inline today rather than calling this helper, and names converting them as a separate, larger change — a degenerate username needs a third fallback branch there plus its own test matrix. Filing that as a follow-up rather than doing it at round 3 of this PR: `ActorInfo`'s actor-present branch is byte-for-byte unchanged from `main`, so the divergence predates this PR.

lib/components/posts/post.tsx
```diff
@@ -88,14 +95,41 @@ interface BoostStatusProps {
...
-    status.actor?.username ||
-    getActorIdMention(status.actorId)
+    getActorDisplayName(status.actor) || getActorIdMention(status.actorId)
+  // Undefined when there is no boost actor row and the actor id also carries
+  // no usable handle (an opaque `did:`/UUID username) — the case `ActorInfo`
+  // renders as plain text too, when it likewise has no actor to link.
+  // `ActorInfo`'s actor-present branch renders a link unconditionally, so it
+  // does not share this fallback: a present actor whose username normalises
+  // to empty still links from here (see `getActorProfileHref`'s own comment)
+  // but renders a broken link there.
```
llun: **Comment above `profileHref` is inaccurate — second time — and doesn't need to restate the truth table.**

Reproduced by exhaustively running `getActorProfileHref` over the actor/actorId matrix:

```
actor absent, actorId absent          => undefined
actor absent, actorId opaque          => undefined
actor absent, actorId usable          => "/@alice@remote.example"
actor empty-username, actorId absent  => undefined
actor empty-username, actorId opaque  => undefined
actor empty-username, actorId usable  => "/@alice@remote.example"
actor usable-username, actorId absent => "/@bob@bob.example"
actor usable-username, actorId opaque => "/@bob@bob.example"
```

Two claims are false against that:
- "Undefined when there is no boost actor row …" — the actor row CAN be present (e.g. `{ username: '@' }` is truthy) and `profileHref` is still `undefined`, provided the actorId is also opaque. The real condition is `(actor absent OR its username normalises empty) AND (actorId absent OR its username is opaque)`.
- "a present actor whose username normalises to empty still links from here" — true only when the actorId ALSO yields a usable handle. With an opaque actorId it renders plain text — the opposite of what's claimed.

Also, 4 of the 7 lines describe `ActorInfo`, which `BoostStatus` never calls and which does not share `getActorProfileHref`. The function's own doc comment is already the accurate source for this; paraphrasing it here is what produced a wrong comment twice.

Suggested replacement:
```
// Can be undefined — see `getActorProfileHref`'s doc comment for exactly
// when, since it is not simply "no boost actor row". `ActorInfo`'s
// actor-present branch links unconditionally instead of falling back, so
// it does not share this behaviour.
```

lib/components/posts/actor.tsx
```diff
@@ -157,6 +178,20 @@ export const getActorIdMention = (
   return `${handle}${domain}`
 }
 
+// The name a caller shows beside (or instead of) `getActorProfileHref`'s
+// destination: `name || getDisplayUsername(username)`. `ActorAvatar`'s
+// initials and `ActorInfo`'s display name compute this same composition
+// inline today rather than calling this helper — converting them is a
+// separate, larger change (a degenerate username needs a third fallback
+// branch there, plus its own test matrix) and is out of scope here.
```
llun: **"out of scope here" is PR-review language that shouldn't ship in the code.**

"converting them is a separate, larger change (…) and is out of scope here" answers this review round, not a future reader — "here" has no referent once this merges. No other comment in `post.tsx` or `actor.tsx` uses in-scope/out-of-scope phrasing; every other comment in this function (and `getActorProfileHref`'s, and the prefetch opt-out block below) states durable technical fact only.

Lines 181-183 (what the composition is, and that two call sites duplicate it inline) and 187-189 (the `undefined` contract) are accurate and should stay — only the middle block needs to change.

Suggested replacement for lines 184-186:
```
// inline today rather than calling this helper: converging them needs a
// third fallback branch there for a degenerate username, plus its own
// test matrix, so they still duplicate it.
```

llun (commented):
Round 4 review: two verified findings on the doc comments touched by the previous round's fixes.

lib/components/posts/post.tsx
llun: Replaced the comment block above `profileHref` with the exact text you verified, at lib/components/posts/post.tsx:99-102 in commit c52a69796. It now points to `getActorProfileHref`'s own doc comment for the exact truth table instead of restating (and mis-stating) it, and drops the `ActorInfo` paraphrase.

lib/components/posts/actor.tsx
llun: Replaced the middle block (previously lines 184-186) with the exact text you verified, at lib/components/posts/actor.tsx:184-186 in commit c52a69796. Lines 181-183 and 187-189 are untouched. The "out of scope here" phrasing is gone; the comment now states the durable fact (converging `ActorAvatar`/`ActorInfo` needs a third fallback branch plus its own test matrix, so they still duplicate this helper) without PR-review language.

lib/components/posts/boost-status.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+    const link = screen.getByRole('link', { name: 'Booster' })
+    expect(link).toHaveAttribute('href', '/@booster@remote.example')
+    expect(link.parentElement).toHaveTextContent('Boosted by Booster')
+  })
+
+  // A feed renders this row once per boost, and `<Link>` prefetches on viewport
+  // entry, so scrolling would fire an RSC request per boosted row against a
+  // fully dynamic route that federates out for unpersisted remote actors.
+  it('does not prefetch the booster profile link', () => {
```
llun, line 104: **Redundant with another test — keeping it deliberately.**

Found in the final sweep: `BoostStatus` renders exactly one `<Link prefetch={false}>` literal, reached by both the actor-present and actor-absent branches, so this test's `data-prefetch="false"` assertion is also made by `links the handle recovered from the actor id when the profile is absent`. Proved removable: deleting it and re-running the full five-mutation revert matrix still caught all five, including the `prefetch={false}` removal.

Decision: keep it. `actor.test.tsx` sets the precedent in this exact area — `ActorAvatar`'s single `<Link prefetch={false}>` literal is covered by two separately-named tests, one per call path, redundant with each other by the identical argument. `AGENTS.md`'s "Link prefetching in feeds" section names both files as the regression guard without drawing a mutation-minimality distinction. The convention here is a named, discoverable test per call path, not the minimum set a mutation matrix requires; dropping this one would remove the only test whose name states the guarantee for the actor-present path.

lib/components/posts/actor.tsx
```diff
@@ -149,6 +149,27 @@ const getActorIdParts = (
...
+// to nowhere.
+export const getActorProfileHref = (
+  actor?: ActorProfile | null,
+  actorId?: string,
+  statusUrl?: string | null
+): string | undefined => {
+  if (actor && getDisplayUsername(actor.username)) {
+    return `/${getActorMention(actor)}`
+  }
+  if (!actorId) return undefined
```
llun, line 169: **Untested branch — deliberately not testing it.**

Mutating this line to fall through leaves all 88 tests passing; the fallthrough would produce `href: '/@'`.

Decision: leave untested. The state it guards — actor present, `actorId` absent — has no live constructor: both call sites (`BoostStatus` and `ActorAvatar`) source `actor` and `actorId` from the same `Status`, whose `actorId` is a required `z.string()` (`lib/types/domain/status.ts:150`), and `ActorAvatar` short-circuits the no-identity case before reaching this line. That makes it a type-narrowing guard for the function's own optional signature, not a defence against untrusted federated data like the malformed-`preferredUsername` cases that ARE tested. Worth a test if a future caller ever decouples the two props.

lib/components/posts/actor.tsx
```diff
@@ -149,6 +149,27 @@ const getActorIdParts = (
 const getActorIdHandle = (actorId: string, statusUrl?: string | null) =>
   getActorIdParts(actorId, statusUrl).handle
 
+// The profile page an actor links to. Callers that need only the href — the
+// avatar below, and the boosted-by line in `post.tsx` — share this instead of
+// re-deriving it. Falls through to the actor-id path when the actor is
+// present but its username normalises to empty (e.g. a federated
```
llun, line 155: **British "normalises" vs the codebase's "normalizes" — leaving it.**

The new tests and comments use "normalises"; `post.test.tsx` and every identifier (`normalizeBlurhash`, `normalizeUsername`, `normalizeHost`) use the American spelling.

Decision: leave. The repo already mixes the two in prose, pre-existing to this PR — British spellings appear in `lib/types/activitypub/actor.ts`, `lib/types/activitypub/objects.ts`, `lib/utils/host.ts`, `lib/components/fitness/useHeatmapTiles.ts`, `lib/activities/jsonld/index.ts`, `lib/services/mastodon/getMastodonFilter.ts`, `lib/database/sql/linkPreview.ts`, and in `AGENTS.md` and `REVIEW.md` themselves. No style rule or lint check pins prose to American English, and identifiers — the load-bearing spelling — are unaffected.

llun (commented):
Round-6 sweep: three WONTFIX decisions recorded inline below.

lib/components/posts/boost-status.test.tsx
llun: Decided as stated above; no change in this PR.

lib/components/posts/actor.tsx
llun: Decided as stated above; no change in this PR.

lib/components/posts/actor.tsx
llun: Decided as stated above; no change in this PR.

PR #1608, opened 2026-08-28
fix: link ActorInfo's display name through the shared profile href

llun:
## What

`ActorInfo`'s actor-present branch built both halves of the author link straight from the actor:

```
href = `/${getActorMention(actor)}`
text = actor.name || getDisplayUsername(actor.username)
```

For a federated actor whose `preferredUsername` normalises to empty — the literal `@` — both collapse at once. The href becomes `/@@domain`, which `parseAccountHandle` rejects (it strips one leading `@`, splits on `@`, and refuses an empty username), so it 404s. The text becomes `''`, so the link is invisible: a zero-width anchor with an empty accessible name.

`preferredUsername` is a bare `z.string()` with no shape check (`lib/types/activitypub/actor.ts`) and `recordActorIfNeeded` writes it verbatim, so any remote server can produce this.

#1606 fixed exactly this for the **sibling** element: `getActorProfileHref` falls through to the actor-id-derived href when the actor's username normalises to empty, and `ActorAvatar` calls it. `ActorInfo` did not, so the avatar and the display name in the same post header linked to two different places for such an actor. Converging them was deferred there because it is not a pure substitution.

## How

- `ActorInfo` reads `getActorProfileHref(actor, actorId, statusUrl)` and renders a `<span>` rather than a `<Link>` when it answers `undefined` — the way its own no-actor branch already did.
- The **third fallback branch**: `getActorDisplayName(actor)` is `''`, not `undefined`, for the degenerate actor, so the name falls through to the actor-id-derived `handle`. The muted text beside it falls through to that same derivation's `domain`, because the actor's own mention is `@@domain` there too — and that fallback is gated on the **username**, not on the display name, so an actor that has a name but a degenerate username keeps its name and still gets a usable mention beside it.
- Those two conditions are now the same condition, which collapses `ActorInfo`'s two branches into one: `getActorDisplayName(actor) || idParts?.handle` covers "no actor" and "degenerate actor" identically, since the helper answers `undefined` for the first and `''` for the second.
- `getUsableActorMention` (added in review, see below) is the one spelling of "does this actor have a usable mention". `getActorProfileHref` and `ActorInfo` both read it, so the predicate that decides the fallback exists once rather than twice.
- `ActorAvatar`'s initials call `getActorDisplayName(actor) || ''` instead of recomposing `name || getDisplayUsername(username)` inline. Pure substitution — the `|| ''` is kept because the helper's return type is `string | undefined`.
- `statusUrl` now reaches the actor-present path, so a degenerate username whose actor id is *also* opaque (`did:`/UUID) can still be rescued by the handle in the status URL.

## Tests

`lib/components/posts/actor.test.tsx` gains a matrix over (actor present/absent) × (username degenerate/normal) × (actor id opaque/usable), asserting the href **and** the exact rendered text (`container.textContent`, not a substring match — `toHaveTextContent` would pass on `Test Actor@@remote.example`).

Written first and watched fail. Against `main`, 6 of the 20 cases fail and each fails for the right reason:

| Case | Failure on `main` |
| --- | --- |
| degenerate username, usable actor id | `href="/@@remote.example"`, expected `/@booster@remote.example` |
| degenerate username + name, usable actor id | same href; text `Test Actor@@remote.example` |
| degenerate username, status url rescue | same href |
| degenerate username, opaque actor id (×2, named and unnamed) | link rendered where plain text is expected |
| degenerate username, no actor id | link to `/@` rendered where plain text is expected |

The other 14 pass on both sides, and that is deliberate: they pin the behaviour this change must **not** move (a normal actor, an unnamed actor, an opaque actor id that the actor mention wins over, the no-actor branch, and `ActorAvatar`'s three initials paths, which are a pure refactor).

The matrix also proved sufficient under mutation: dropping `getUsableActorMention`'s `actor &&` guard, weakening it to `actor ? … : null`, inverting its ternary, and dropping `getActorProfileHref`'s early return each fail 3–17 of the 20 cases. No mutant survived.

## Verification

Full gate green at every commit: `yarn run prettier --write .`, `yarn lint`, `yarn typecheck`, `yarn build`, `yarn test` (898 files / 12,015 tests).

Checked in a browser against a local SQLite instance, light and dark, with two seeded remote actors carrying `preferredUsername: '@'` — one named, one not — posting into the home timeline.

Before, on `main`:

| element | href | text |
| --- | --- | --- |
| avatar | `/@namedbooster@named.example` | `NB` |
| display name | `/@@named.example` | `Named Booster` |
| avatar | `/@booster@degenerate.example` | — |
| display name | `/@@degenerate.example` | *(empty)* |

`http://localhost:3100/@@degenerate.example` renders the app's 404 page, confirming the destination.

After:

| element | href | text |
| --- | --- | --- |
| avatar | `/@namedbooster@named.example` | `NB` |
| display name | `/@namedbooster@named.example` | `Named Booster` |
| avatar | `/@booster@degenerate.example` | — |
| display name | `/@booster@degenerate.example` | `@booster` |

Every ordinary actor on the same timeline (local `testuser`, remote `mira` and `rico`) renders byte-identically before and after — avatar and display name both pointing at the same `/@user@domain`. No console errors, no hydration warnings.

## Known gap, deliberately not widened here

`ActorAvatar`'s initials stay a pure substitution, so the degenerate unnamed actor's avatar still falls back to an empty string rather than to the actor-id handle. That is unchanged from `main` and from #1606; giving the avatar the same third fallback is a separate decision about the avatar's own fallback text, not about this link.

## Docs

`AGENTS.md` → **Status Posts & Actions**: a bullet recording that all three author links in a post derive their href from `getActorProfileHref`, and that each degrades to **unlinked content** where it answers `undefined` — plain text for `ActorInfo` and `BoostStatus`, an unlinked avatar for `ActorAvatar`. It also records why `ActorInfo` cannot simply render `getActorDisplayName` + `getActorMention`, and that the name and the mention are separate fallback chains that can disagree.

Stale comments corrected in the same PR: `getActorDisplayName`'s doc said `ActorAvatar` and `ActorInfo` still duplicate its composition; `BoostStatus`'s said `ActorInfo` links unconditionally; and `getActorProfileHref`'s tail said all its callers degrade to "plain text", which was never true of the avatar.

## Review

The sub-agent code-review loop ran **5 rounds**, ending on a clean one. Six findings across rounds 1–4, each confirmed by an adversarial verifier before being posted, each fixed, replied to and resolved; round 5 surfaced nothing. One further finding was refuted and never posted.

| Round | Found |
| --- | --- |
| 1 | `AGENTS.md` overclaimed that a degenerate actor is named from the actor id — untrue when it has a `name`; "degrades to plain text" was false for `ActorAvatar`; the usable-mention predicate was spelled twice |
| 2 | round 1's own fix had put one explanation in three places — the `parseAccountHandle` sentence in two code comments, and the name-fallback paragraph in both the code and `AGENTS.md` |
| 3 | the surviving comment still restated the bullet it pointed at, and "its mention and href come from the actor id" was loose — `mutedHandle` is `idParts.domain` alone |
| 4 | the count in this very section: it read "5 findings" while six threads stood resolved |
| 5 | nothing — the round that stopped the loop |

Round 3's thread bundles two findings — the scope and correctness lenses reached the same clause by different routes — so six threads carry seven lens findings.

The refuted one: `getActorProfileHref` short-circuits on `!actorId` before consulting `statusUrl`, while `ActorInfo`'s `idParts` consults it regardless — so in principle a degenerate actor with no `actorId` but a rescuable `statusUrl` would render a handle with no link. Unreachable: `Status.actor` is only non-null when it was looked up *by* that same `actorId`, so `actor` truthy with `actorId` falsy cannot occur.

Rounds 2 and 3 are worth noting as a pattern — a round whose job is fixing duplicated prose is the round most likely to create more of it, because the fixer is holding the corrected wording and looking for somewhere to put it. Both were settled by deleting rather than rewording.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1721,23 +1721,31 @@ consistency is enforced by keeping the wiring in one place rather than per page.
   own. There is deliberately no prop for hiding one of the menu's own items;
   that is the per-page drift this whole section exists to prevent.
 - **Every author link in a post derives its href from `getActorProfileHref`
-  (`lib/components/posts/actor.tsx`), and each one degrades to plain text
-  where that answers `undefined`.** The three are the avatar (`ActorAvatar`),
-  the display name (`ActorInfo`) and the boosted-by line (`BoostStatus` in
-  `post.tsx`). A federated `preferredUsername` is a bare `z.string()` that
-  `recordActorIfNeeded` writes verbatim, so it can normalise to nothing — and
-  the mention built from one that does, `@@domain`, is a handle
-  `parseAccountHandle` rejects. `ActorInfo` therefore does **not** simply
-  render `getActorDisplayName(actor)` and `getActorMention(actor)`: when the
-  username normalises to empty it names the actor from the actor id instead,
-  the same three-part `handle`/`domain`/`href` the no-actor case has always
-  used, so its text and its destination stay in step with the avatar beside
-  it. Before this was shared, the avatar linked to `/@booster@domain` while
-  the display name right next to it linked to `/@@domain` — a 404 — with
-  **empty** link text, because `name || getDisplayUsername(username)` is `''`
-  for the same actor. Covered by the matrix in
-  `lib/components/posts/actor.test.tsx`; every case there passes against the
-  old code except the degenerate-username ones.
+  (`lib/components/posts/actor.tsx`), and each one degrades to unlinked
+  content where that answers `undefined`.** The three are the avatar
+  (`ActorAvatar`), the display name (`ActorInfo`) and the boosted-by line
+  (`BoostStatus` in `post.tsx`) — `ActorInfo` and `BoostStatus` fall back to
+  plain text, `ActorAvatar` falls back to an unlinked wrapper around the same
+  avatar (image or initials) it would otherwise link. A federated
+  `preferredUsername` is a bare `z.string()` that `recordActorIfNeeded` writes
+  verbatim, so it can normalise to nothing — and the mention built from one
+  that does, `@@domain`, is a handle `parseAccountHandle` rejects. `ActorInfo`
+  therefore does **not** simply render `getActorDisplayName(actor)` and
+  `getActorMention(actor)`: when the username normalises to empty, the
+  mention — and so both the href and the muted handle beside the name — falls
+  back to the actor id's `handle`/`domain`/`href` tuple, the same one the
+  no-actor case has always used, so the link's destination stays in step with
+  the avatar beside it. The name is a separate fallback chain,
+  `getActorDisplayName(actor) || idParts?.handle || ''`, that checks the
+  actor's own `name` first: a named actor with a degenerate username keeps its
+  name as the link text even though the mention under it switched to the
+  actor id, and only an actor with neither a name nor a usable username is
+  named from the actor-id handle too. Before this was shared, the avatar
+  linked to `/@booster@domain` while the display name right next to it linked
+  to `/@@domain` — a 404 — with **empty** link text, because
+  `name || getDisplayUsername(username)` is `''` for the same actor. Covered
+  by the matrix in `lib/components/posts/actor.test.tsx`; every case there
+  passes against the old code except the degenerate-username ones.
 
 ### Post media layout (`attachments.tsx`)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1720,6 +1720,24 @@ consistency is enforced by keeping the wiring in one place rather than per page.
   in the row a moment ago, so they stay nearest it) and **before** the menu's
   own. There is deliberately no prop for hiding one of the menu's own items;
   that is the per-page drift this whole section exists to prevent.
+- **Every author link in a post derives its href from `getActorProfileHref`
+  (`lib/components/posts/actor.tsx`), and each one degrades to plain text
+  where that answers `undefined`.** The three are the avatar (`ActorAvatar`),
+  the display name (`ActorInfo`) and the boosted-by line (`BoostStatus` in
+  `post.tsx`). A federated `preferredUsername` is a bare `z.string()` that
+  `recordActorIfNeeded` writes verbatim, so it can normalise to nothing — and
+  the mention built from one that does, `@@domain`, is a handle
+  `parseAccountHandle` rejects. `ActorInfo` therefore does **not** simply
+  render `getActorDisplayName(actor)` and `getActorMention(actor)`: when the
+  username normalises to empty it names the actor from the actor id instead,
+  the same three-part `handle`/`domain`/`href` the no-actor case has always
+  used, so its text and its destination stay in step with the avatar beside
+  it. Before this was shared, the avatar linked to `/@booster@domain` while
+  the display name right next to it linked to `/@@domain` — a 404 — with
+  **empty** link text, because `name || getDisplayUsername(username)` is `''`
+  for the same actor. Covered by the matrix in
+  `lib/components/posts/actor.test.tsx`; every case there passes against the
+  old code except the degenerate-username ones.
 
 ### Post media layout (`attachments.tsx`)
 
--- a/lib/components/posts/actor.tsx
+++ b/lib/components/posts/actor.tsx
@@ -149,12 +149,12 @@ const getActorIdParts = (
 const getActorIdHandle = (actorId: string, statusUrl?: string | null) =>
   getActorIdParts(actorId, statusUrl).handle
 
-// The profile page an actor links to. Callers that need only the href — the
-// avatar below, and the boosted-by line in `post.tsx` — share this instead of
-// re-deriving it. Falls through to the actor-id path when the actor is
-// present but its username normalises to empty (e.g. a federated
-// `preferredUsername` of just `@` characters), so the href is never built
-// from a mention with an empty local part — otherwise undefined when the
+// The profile page an actor links to, shared by every author link in a post —
+// the avatar below, `ActorInfo`'s display name, and the boosted-by line in
+// `post.tsx` — instead of each re-deriving it. Falls through to the actor-id
+// path when the actor is present but its username normalises to empty (e.g. a
+// federated `preferredUsername` of just `@` characters), so the href is never
+// built from a mention with an empty local part — otherwise undefined when the
 // actor id also carries no usable handle (an opaque `did:`/UUID username),
 // which is the case those callers render as plain text rather than as a link
 // to nowhere.
@@ -179,14 +179,13 @@ export const getActorIdMention = (
 }
 
 // The name a caller shows beside (or instead of) `getActorProfileHref`'s
-// destination: `name || getDisplayUsername(username)`. `ActorAvatar`'s
-// initials and `ActorInfo`'s display name compute this same composition
-// inline today rather than calling this helper: converging them needs a
-// third fallback branch there for a degenerate username, plus its own
-// test matrix, so they still duplicate it.
+// destination: `name || getDisplayUsername(username)`.
 // Undefined when there is no actor to name, mirroring `getActorProfileHref`'s
-// own `undefined` for "nothing to link"; a caller with a fallback of its own
-// (`BoostStatus` reads the actor id instead) chains it with `||`.
+// own `undefined` for "nothing to link". It is empty — not undefined — for an
+// actor that exists but has neither a name nor a username that normalises to
+// anything, so every caller needs a fallback of its own and chains it with
+// `||`: `BoostStatus` and `ActorInfo` read the actor id, `ActorAvatar` draws
+// no initials.
 export const getActorDisplayName = (
   actor?: ActorProfile | null
 ): string | undefined =>
@@ -204,7 +203,7 @@ export const ActorAvatar: FC<Props> = ({ actor, actorId, statusUrl }) => {
 
   const href = getActorProfileHref(actor, actorId, statusUrl)
   const initials = actor
-    ? getInitials(actor.name || getDisplayUsername(actor.username) || '')
+    ? getInitials(getActorDisplayName(actor) || '')
     : getInitials(getActorIdHandle(actorId || '', statusUrl))
 
   const avatar = (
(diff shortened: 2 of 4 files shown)


AGENTS.md
```diff
@@ -1720,6 +1720,24 @@ consistency is enforced by keeping the wiring in one place rather than per page.
...
+- **Every author link in a post derives its href from `getActorProfileHref`
+  (`lib/components/posts/actor.tsx`), and each one degrades to plain text
+  where that answers `undefined`.** The three are the avatar (`ActorAvatar`),
+  the display name (`ActorInfo`) and the boosted-by line (`BoostStatus` in
+  `post.tsx`). A federated `preferredUsername` is a bare `z.string()` that
+  `recordActorIfNeeded` writes verbatim, so it can normalise to nothing — and
+  the mention built from one that does, `@@domain`, is a handle
+  `parseAccountHandle` rejects. `ActorInfo` therefore does **not** simply
+  render `getActorDisplayName(actor)` and `getActorMention(actor)`: when the
+  username normalises to empty it names the actor from the actor id instead,
```
llun: **AGENTS.md overclaims that the name comes from the actor id**

The sentence "when the username normalises to empty it names the actor from the actor id instead, the same three-part `handle`/`domain`/`href` the no-actor case has always used" is wrong for a degenerate-username actor that **has** a `name`.

`name = getActorDisplayName(actor) || idParts?.handle || ''` short-circuits on the actor's own `name`, so only the href and the muted mention move to the actor-id-derived tuple — the display text stays `actor.name`. The id-derived handle is reached only when the actor has neither a name nor a usable username.

The bullet's own cited test contradicts it: `keeps the name but takes the mention from the actor id` expects text `'Test Actor@remote.example'` — the actor's real name, not `@booster`.

Scope the sentence to the mention/href, and state the name's precedence separately.

AGENTS.md
```diff
@@ -1720,6 +1720,24 @@ consistency is enforced by keeping the wiring in one place rather than per page.
   in the row a moment ago, so they stay nearest it) and **before** the menu's
   own. There is deliberately no prop for hiding one of the menu's own items;
   that is the per-page drift this whole section exists to prevent.
+- **Every author link in a post derives its href from `getActorProfileHref`
+  (`lib/components/posts/actor.tsx`), and each one degrades to plain text
```
llun: **"degrades to plain text" is false for one of the three links it enumerates**

The bullet asserts "each one degrades to plain text" and then names three: the avatar (`ActorAvatar`), the display name (`ActorInfo`) and the boosted-by line (`BoostStatus`). It holds for the latter two — `ActorInfo` renders `<span className="font-semibold truncate">{name}</span>` and `BoostStatus` renders the bare `actorName` string. It does not hold for `ActorAvatar`, whose `if (!href)` branch returns `<div onClick={…}>{avatar}</div>` — an unlinked `<Avatar>`, i.e. an image or an initials badge, never a text node.

Say "degrades to unlinked content" and spell out which is which.

Related, and now in scope because this PR rewrote the opening of that comment: `getActorProfileHref`'s doc comment in `lib/components/posts/actor.tsx` ends "...which is the case those callers render as plain text rather than as a link to nowhere". That tail was already imprecise about the avatar on `origin/main`, but the caller list immediately above it is one this PR edited, so fix both in the same change.

lib/components/posts/actor.tsx
```diff
@@ -228,44 +227,36 @@ export const ActorAvatar: FC<Props> = ({ actor, actorId, statusUrl }) => {
...
-    )
-  }
+  const href = getActorProfileHref(actor, actorId, statusUrl)
+  // The actor's own mention is only usable while its username normalises to
+  // something: `@@domain` is a handle `parseAccountHandle` rejects, and the
+  // name beside it would be empty too. So an actor with a degenerate
+  // `preferredUsername` is named from the actor id instead — the same place
+  // `getActorProfileHref` has already sent the link, and the same parts the
+  // no-actor case has always used.
+  const mention =
```
llun: **the "does this actor have a usable mention" predicate is now spelled twice**

`getActorProfileHref` has `if (actor && getDisplayUsername(actor.username)) return `/${getActorMention(actor)}``, and `ActorInfo` re-evaluates the identical `actor && getDisplayUsername(actor.username) ? getActorMention(actor) : null` to get the text of that same mention.

That is the shape this PR exists to remove. The bug being fixed was two places deriving one actor's mention independently and drifting; leaving a second spelling of the very predicate that decides "does this actor have a usable mention" invites the same drift back.

Extract a private helper:
```ts
const getUsableActorMention = (actor?: ActorProfile | null): string | null =>
  actor && getDisplayUsername(actor.username) ? getActorMention(actor) : null
```
`getActorProfileHref` becomes `const mention = getUsableActorMention(actor); if (mention) return `/${mention}`` and `ActorInfo` reads it directly.

Verified behaviour-preserving: same functions, same arguments, same call counts, and it does **not** touch either short-circuit that keeps `getActorIdParts` (and its unguarded `decodeURIComponent` in `getActorIdUsername`) from running — those guards are unchanged. Nice-to-have, not blocking.

AGENTS.md
llun: Fixed in 14047a19e. The bullet now states the two fallbacks separately: the *mention* (and with it the href and the muted handle) falls back to the actor-id tuple whenever the username normalises to empty, regardless of whether the actor has a name; the *name* is its own chain, `getActorDisplayName(actor) || idParts?.handle || ''`, which checks `actor.name` first — so a named actor with a degenerate username keeps its name as the link text even though the mention under it switched to the actor id, and only an actor with neither a name nor a usable username is named from the actor-id handle too. That now agrees with the `keeps the name but takes the mention from the actor id` case it cites.

AGENTS.md
llun: Fixed in 14047a19e. The bullet now says "degrades to unlinked content" and names which is which: plain text for `ActorInfo` and `BoostStatus`, an unlinked wrapper around the same avatar (image or initials) for `ActorAvatar`. The same correction is applied to the tail of `getActorProfileHref`'s doc comment in `lib/components/posts/actor.tsx`, which this PR had already rewritten above that sentence.

lib/components/posts/actor.tsx
llun: Fixed in 14047a19e. `getUsableActorMention` is extracted next to `getActorMention`; `getActorProfileHref` and `ActorInfo` both read it, so the predicate exists once. Behaviour-preserving — all 20 cases in `lib/components/posts/actor.test.tsx` still pass with that file unedited, and neither short-circuit that keeps `getActorIdParts` (and the unguarded `decodeURIComponent` in `getActorIdUsername`) from running was disturbed: `idParts` is still computed only when `mention` is falsy. Full gate re-run green independently of the fixer: prettier:check, lint, typecheck, build, and 898 files / 12,015 tests.

The fix also removed a clause on the same line — `ActorInfo`'s comment said "and the name beside it would be empty too", which carried the same overclaim as thread PRRT_kwDOIZZzx86dFsMm.

lib/components/posts/actor.tsx
```diff
@@ -228,44 +233,39 @@ export const ActorAvatar: FC<Props> = ({ actor, actorId, statusUrl }) => {
...
-        ) : (
-          <span className="font-semibold truncate">{handle}</span>
-        )}
-        <span className="text-muted-foreground truncate">{domain}</span>
-      </div>
-    )
-  }
+  const href = getActorProfileHref(actor, actorId, statusUrl)
+  // The actor's own mention is only usable while its username normalises to
+  // something: `@@domain` is a handle `parseAccountHandle` rejects. So an
```
llun: **Title: round 1's fix put the same explanation in two code comments**

`getUsableActorMention`'s own doc comment (around line 39) already says: "The actor's own mention is only usable while its username normalises to something: `@@domain` is a handle `parseAccountHandle` rejects, so an actor with a degenerate `preferredUsername` has no mention to offer here."

`ActorInfo`'s comment now opens with a near-verbatim restatement of the same two clauses. Before round 1 extracted the helper there was one copy; now there are two in this file, plus the `AGENTS.md` bullet.

The helper is the right owner — it is the smallest unit adjacent to the actual check, and both call sites reach the fact through it. Delete the restatement from `ActorInfo` rather than rewording it again.

lib/components/posts/actor.tsx
```diff
@@ -228,44 +233,39 @@ export const ActorAvatar: FC<Props> = ({ actor, actorId, statusUrl }) => {
...
+  // actor with a degenerate `preferredUsername` gets its mention (and href)
+  // from the actor id instead — the same tuple `getActorProfileHref` has
+  // already sent the link to, and the same parts the no-actor case has
+  // always used. The *name* is a separate fallback chain that checks the
```
llun: **Title: the "name is a separate fallback chain" paragraph is duplicated verbatim into AGENTS.md**

`AGENTS.md` (around line 1738) now says: "The name is a separate fallback chain, `getActorDisplayName(actor) || idParts?.handle || ''`, that checks the actor's own `name` first: a named actor with a degenerate username keeps its name as the link text even though the mention under it switched to the actor id, and only an actor with neither a name nor a usable username is named from the actor-id handle too."

`ActorInfo`'s comment says the same thing, same two clauses, same order, reworded. Round 1 wrote both in the same commit — so a round whose whole job was fixing duplicated prose introduced a fresh instance of it.

This repo's convention is a terse code comment paired with the narrative `AGENTS.md` entry. Let `AGENTS.md` keep the full case breakdown and cut the code comment to the one thing a reader cannot get from the code or the two neighbouring doc comments — that `name` and `mention` are independent chains and can disagree — with a pointer to the section, the way `ActorAvatar`'s comment points at "Link prefetching in feeds".

Net effect: a ~10-line comment above a 4-line derivation comes down to about 4.

lib/components/posts/actor.tsx
llun: Fixed in d965ec801. Deleted, not reworded. `getUsableActorMention`'s doc comment is untouched and remains the single owner of the `@@domain` / `parseAccountHandle` fact; `ActorInfo`'s restatement of it is gone.

lib/components/posts/actor.tsx
llun: Fixed in d965ec801. The full case breakdown stays in the `AGENTS.md` bullet, and the code comment is cut to the one thing neither that bullet nor the neighbouring helper comments state at this spot — that `name` and `mention` are independent chains and can disagree — with a pointer to the section, matching how `ActorAvatar`'s comment points at "Link prefetching in feeds". Ten comment lines became four above the same four-line derivation.

The commit changes one file and one hunk: `AGENTS.md`, the tests and `post.tsx` are untouched. Gate re-run independently of the fixer: prettier:check, lint, typecheck, build, and 898 files / 12,015 tests, with `lib/components/posts/actor.test.tsx` unedited and all 20 cases passing.

lib/components/posts/actor.tsx
```diff
@@ -228,44 +233,31 @@ export const ActorAvatar: FC<Props> = ({ actor, actorId, statusUrl }) => {
...
-          </Link>
-        ) : (
-          <span className="font-semibold truncate">{handle}</span>
-        )}
-        <span className="text-muted-foreground truncate">{domain}</span>
-      </div>
-    )
-  }
+  const href = getActorProfileHref(actor, actorId, statusUrl)
+  // `name` and `mention` are independent fallback chains and can disagree.
```
llun, line 237: Two round-3 lenses flagged the same clause of this comment for different reasons.

The scope-regression lens: round 2 was told to delete rather than reword, but the 4-line remainder still restated the `AGENTS.md` bullet it points at — that bullet already says "a named actor with a degenerate username keeps its name as the link text even though the mention under it switched to the actor id".

The correctness lens, independently: "its mention and href come from the actor id" is loose. `href` is unambiguously `idParts.href`, but the mention that actually renders is `mutedHandle` = `idParts.domain` alone (`@remote.example`) — never the full `handle`+`domain` the actor id resolves to. A reader could take it to mean the whole mention is shown.

Both are settled by deleting the clause rather than rewording it a fourth time.

llun: Fixed in 94bf7a219. The clause is deleted; what remains states only that the two chains can disagree and points at the section that spells out when — which is what a call-site comment should do when the conditions live elsewhere. A deletion also cannot introduce a fifth wrong claim, which is the failure mode this comment has been circling for three rounds.

Gate re-run: prettier, lint, typecheck, build, 898 files / 12,015 tests. Comment-only — no executable token changed.

llun:
The PR body was written against commit 1 of 4 and had gone stale: its **Docs** section still said the `AGENTS.md` bullet records "degrades to plain text", which round 1 corrected to "unlinked content", and it did not mention `getUsableActorMention` or the two later comment commits. Refreshed against the final four-commit state, with a **Review** section summarising the three rounds.

PR #1627, opened 2026-08-29
fix: consume verification code before redirecting authenticated session

llun:
## Summary

`app/(nosidebar)/auth/confirmation/page.tsx` was redirecting to `/` whenever an active cookie session existed *before* checking and consuming the `verificationCode` query param.

### Reachability Analysis
While `databaseHooks.session.create.before` (`canCreateSessionForAccount`) prevents creating new sessions for unconfirmed accounts, hitting `/auth/confirmation?verificationCode=...` while holding a session occurs in reachable scenarios:
1. **Same-account email repoint/rotation**: When an account updates its email via `POST /api/v1/emails/confirmations`, `repointUnconfirmedAccountEmail` marks the account unconfirmed (`emailVerified: false`, `verifiedAt: null`, `verificationCode: '...'`), but does not delete existing sessions from the database (unlike password reset). The account is now pending confirmation while retaining an active cookie session. Clicking the confirmation link redirected to `/`, where `AuthenticatedGuard` 403s on every route, leaving the account locked with no way to confirm.
2. **Cross-account browser session**: If a user is already signed in (e.g. User A) in a browser and opens a confirmation link for User B (or an alt/test account), the previous code redirected User A to `/` without consuming User B's verification code.

### Fix
- Reordered operations in `app/(nosidebar)/auth/confirmation/page.tsx` so the verification code is extracted and consumed first via `database.verifyAccount({ verificationCode: code })`.
- Retained the redirect to `/` when an active session exists and no verification code query param is present.
- Added unit tests pinning this behavior in `app/(nosidebar)/auth/confirmation/page.test.tsx`.

### Cache / State Invalidation
No session or actor cache invalidation is required: Better Auth sessions store token metadata rather than verification state; React `cache()` is per-request scoped; and guards reload the account freshly from the database on subsequent requests.

## Checklist

### DIFF
--- a/app/(nosidebar)/auth/confirmation/page.test.tsx
+++ b/app/(nosidebar)/auth/confirmation/page.test.tsx
@@ -1,0 +1,143 @@
+import { getDatabase } from '@/lib/database'
+import { Database } from '@/lib/database/types'
+import { getServerAuthSession } from '@/lib/services/auth/getSession'
+
+import Page from './page'
+
+vi.mock('@/lib/database', () => ({
+  getDatabase: vi.fn()
+}))
+
+vi.mock('@/lib/services/auth/getSession', () => ({
+  getServerAuthSession: vi.fn()
+}))
+
+const redirectMock = vi.fn((path: string) => path)
+vi.mock('next/navigation', () => ({
+  redirect: (path: string) => redirectMock(path)
+}))
+
+const aSession = { user: { email: 'rider@example.com' } } as Awaited<
+  ReturnType<typeof getServerAuthSession>
+>
+
+describe('/auth/confirmation page', () => {
+  const verifyAccountMock = vi.fn()
+  const mockDatabase = {
+    verifyAccount: verifyAccountMock
+  } as unknown as Database
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


llun:
## Code Review Loop

Sub-agent code review completed (1 round). The diff was reviewed against `REVIEW.md` and `AGENTS.md` for correctness, security, import style, logging, API patterns, test quality, and documentation.

**Result: No actionable findings.**

Items verified:
- Reorder is correct: verification code is consumed before the session redirect check
- Imports use absolute `@/lib/...` paths correctly
- Tests pin the order of operations and cover the regression case (session + code), the redirect case (session + no code), logged-out paths, array params, empty strings, and database unavailability
- No `console.log` or unauthorized logging
- No new attack surface introduced
- `verifyAccount` called with correct params
- `isVerify` helper handles undefined/empty codes correctly
- jsdom docblock not needed (server component function returns plain JS objects, no DOM rendering)

Definition of Done validation passed:
- `yarn lint` ✅ (0 errors)
- `yarn typecheck` ✅
- `yarn build` ✅
- `yarn test` ✅ (903 files, 12,125 tests)

PR #1700, opened 2026-09-01
fix(otel): remove KnexInstrumentation to resolve Cloud SQL Query span collision

llun:
### Summary

When `KnexInstrumentation` from `@opentelemetry/instrumentation-knex` is enabled, it creates a client-side database span (`select.actors`) with its own `spanId` for every query and sets it as the active span context.

When `attachSqlcommenter` subsequently read the active span context to generate `/*traceparent='00-traceId-spanId-01'*/`, it passed that client span's ID to Cloud SQL. Google Cloud SQL Query Insights uses the `parent_id` from the comment as the `spanId` of its `Cloud SQL Query` span. Because both the application's Knex span and Cloud SQL's span shared the identical `spanId`, Cloud Trace encountered a span ID collision where Cloud SQL's span showed up as a root span without parentage.

### Changes
1. Removed `KnexInstrumentation` so the application's true request / service span (e.g. `api.getInstance`) remains active when SQL queries are prepared.
2. Cloud SQL Query Insights now receives the application's request span ID in `traceparent`, attaching the database query execution plans directly as children under the application span in Google Cloud Trace.
3. Enhanced `attachSqlcommenter` to ensure client prototype-level wrapping.

### Verification
- Prettier, Oxlint, TypeScript typecheck, Next.js build, and Vitest suite (12,453 tests) pass.

### DIFF
--- a/instrumentation.node.ts
+++ b/instrumentation.node.ts
@@ -7,7 +7,6 @@ import { OTLPTraceExporter as GrpcOLTPTraceExporter } from '@opentelemetry/expor
 import { OTLPTraceExporter as HttpOLTPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http'
 import { OTLPTraceExporter as ProtoOLTPTraceExporter } from '@opentelemetry/exporter-trace-otlp-proto'
 import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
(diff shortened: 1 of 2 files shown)


llun:
The main issue is not code but the configration in CloudSQL here

Screenshot 2026-09-01 at 16 15 01

 Select `Store application tags` and the trace is now connected

PR #1701, opened 2026-09-01
fix: attach traceparent sqlcomment to queries inside knex transactions

llun:
## Summary

Follow-up to #1699. Cloud SQL Query Insights spans were appearing in Cloud Trace as isolated root traces (`parentSpanId: null`) instead of nesting under the application spans that triggered the queries.

Investigation found **two independent causes**:

1. **Instance configuration (primary, fixed operationally):** the Cloud SQL instance had Query Insights enabled but not `recordApplicationTags`, which is what makes Query Insights parse sqlcommenter tags — `traceparent` included. Without it, Query Insights discards the comment and mints its own trace IDs for sampled query plans, which is exactly the observed symptom. This is a [community-confirmed failure mode](https://discuss.google.dev/t/query-insights-query-traces-not-linking-up-with-application-traces/98302), and the setting has now been enabled on the instance (no restart required — the update completed in ~2 minutes with the instance available).

2. **Transactions bypassed the sqlcommenter wrapper (this PR):** `attachSqlcommenter` overrode `query` on the knex client *instance*, but knex builds each transaction client via `Object.create(client.constructor.prototype)` ([knex 3.3.0 `makeTxClient`](https://github.com/knex/knex/blob/master/lib/execution/transaction.js)) and captures `const _query = trxClient.query` from the prototype — the instance-level override is never consulted. Every statement inside every `db.transaction()` (83 call sites in `lib/database/sql` plus the better-auth `knexAdapter` — essentially the whole write path) reached PostgreSQL with no traceparent comment, so its Query Insights samples could never link even with the instance flag enabled.

## Fix

Install the wrapper on the dialect class prototype instead of the instance, guarded by a symbol against double attachment. The transaction client inherits the wrapped `query`, and `makeTxClient`'s own wrapper captures and calls through it. Verified mechanically: the comment is appended before knex's `enrichQueryObject`, contains no `?` or `\` (so `positionBindings` cannot mangle it), and node-postgres sends `queryConfig.text` verbatim.

## Verification of the wider diagnosis

- The app side was already attaching correct comments for non-transactional queries: production revision `activitiesnext-01261-qcr` runs exactly `cb500118d` (#1699), deployed ~1h before the sampled example trace — yet that trace (`819965a52958355770648998f47148c0`) contains only Query Insights spans with no application tags at all, confirming the parsing was disabled instance-side.
- sqlcommenter format checked against the spec: trailing comment, single-quoted value, `00-<32hex>-<16hex>-01` — all correct (hex/dash values need no URL encoding).
- `AlwaysOnSampler` means every locally-started span carries sampled flags (`-01`), including under an unsampled Cloud Run LB parent.
- Docs note: query-string truncation (`queryStringLength=1024`) does **not** drop tags ("You can still add tags to queries that exceed the length limit").

## Test

New regression test: a query inside `db.transaction()` must carry the traceparent comment — fails against the old instance-level wrapper (`SELECT 1` reaches the driver bare), passes with the prototype-level attachment. The existing capture hook also moved to the class prototype, since transaction clients never consult instance properties.

A second test (from review round 1) pins the double-attachment guard: attaching sqlcommenter to a second knex instance sharing the dialect class must not stack the wrapper — with the guard removed, queries carry the traceparent comment twice. The `Symbol.for` (vs `Symbol()`) choice is now documented in code: the global-registry key survives module re-evaluation under Next.js dev Fast Refresh, where a per-module symbol would re-wrap `query` on every reload.

## Test results

- `yarn lint` ✅ (pre-existing warnings only)
- `yarn typecheck` ✅
- `yarn build` ✅
- `yarn test` ✅ 932 files / 12,455 tests (re-run in full after each review round)

## Known remaining gaps (out of scope, documented for follow-up)

- 6 of 295 route files don't wrap handlers in `traceApiRoute` (`oauth/token`, `api/auth/[...all]`, `api/oauth/authorize`, `v1/trends`, two fitness-heatmap routes) — queries there rely on the HttpInstrumentation span's context.
- knex streams (`client.stream`) bypass the wrapper — zero production callers today.
- Operator scripts and knex-CLI migrations run uninstrumented by design.

### DIFF
--- a/lib/database/index.ts
+++ b/lib/database/index.ts
@@ -11,6 +11,10 @@ interface DatabaseInstance {
   knex: Knex
 }
 
+// Symbol.for (global registry), not Symbol(): Next.js dev Fast Refresh
+// re-evaluates this module while the cached knex dialect class survives, so a
+// per-module Symbol() would never match on reload and query would be
+// re-wrapped every time, stacking duplicate traceparent comments.
 const SQLCOMMENTER_ATTACHED = Symbol.for('activities.next.sqlcommenter')
 
 export const attachSqlcommenter = (db: Knex): Knex => {
--- a/lib/database/index.ts
+++ b/lib/database/index.ts
@@ -11,10 +11,17 @@ interface DatabaseInstance {
   knex: Knex
 }
 
+const SQLCOMMENTER_ATTACHED = Symbol.for('activities.next.sqlcommenter')
+
 export const attachSqlcommenter = (db: Knex): Knex => {
-  const origQuery = db.client?.query
-  if (typeof origQuery === 'function') {
-    db.client.query = function (
+  // Patch the dialect class prototype, not the knex instance: knex builds each
+  // transaction client via Object.create(client.constructor.prototype), so an
+  // instance-level override is invisible to every query inside a transaction.
+  const proto = db.client?.constructor?.prototype
+  const origQuery = proto?.query
+  if (typeof origQuery === 'function' && !proto[SQLCOMMENTER_ATTACHED]) {
+    proto[SQLCOMMENTER_ATTACHED] = true
+    proto.query = function (
       connection: unknown,
       obj: { sql: string; [key: string]: unknown } | string
     ) {
--- a/lib/database/sqlcommenter.test.ts
+++ b/lib/database/sqlcommenter.test.ts
@@ -13,9 +13,13 @@ describe('attachSqlcommenter', () => {
     })
   )
 
+  // Capture on the dialect class prototype, not the knex instance: a
+  // transaction client is built via Object.create(client.constructor.prototype)
+  // and never consults instance-level properties.
   let executedSql: string[] = []
-  const origDriverQuery = db.client._query
-  db.client._query = function (
+  const clientProto = db.client.constructor.prototype
+  const origDriverQuery = clientProto._query
+  clientProto._query = function (
     conn: unknown,
     obj: { sql: string; [key: string]: unknown }
   ) {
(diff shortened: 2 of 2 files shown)


lib/database/sqlcommenter.test.ts
```diff
@@ -46,6 +50,28 @@ describe('attachSqlcommenter', () => {
...
+    await db.transaction(async (trx) => {
+      await trx.raw('SELECT 1')
+    })
+
+    const selects = executedSql.filter((sql) => sql.startsWith('SELECT 1'))
+    expect(selects).toHaveLength(1)
+    expect(selects[0]).toContain(
+      "/*traceparent='00-02dad5002ab305f1fd75ae8bd0d46e94-3ca938408dc381d3-01'*/"
+    )
+  })
```
llun, line 73: The double-attachment guard in `attachSqlcommenter` (the `!proto[SQLCOMMENTER_ATTACHED]` check plus the `proto[SQLCOMMENTER_ATTACHED] = true` write, `lib/database/index.ts:22-23`) has no committed test. It is load-bearing: with the guard removed, calling `attachSqlcommenter` on two knex instances sharing a dialect class stacks the wrapper and every query carries the traceparent comment twice (verified by running exactly that mutation). A future refactor that drops the guard would regress silently. Add a test that calls `attachSqlcommenter` on a second knex instance of the same dialect (or twice on the same instance) and asserts the comment appears exactly once.

lib/database/index.ts
```diff
@@ -11,10 +11,17 @@ interface DatabaseInstance {
   knex: Knex
 }
 
+const SQLCOMMENTER_ATTACHED = Symbol.for('activities.next.sqlcommenter')
```
llun, line 18: `Symbol.for` (global registry) vs a plain `Symbol()` is load-bearing here and nothing says so. Under Next.js dev-server Fast Refresh this module is re-evaluated while the knex dialect class survives in the module cache; a per-module `Symbol()` would never match on the re-run, so the guard would pass and `proto.query` would be wrapped again on top of itself — queries would then carry a compounding duplicate traceparent comment that grows with every reload, and no test can catch it (Vitest does not simulate HMR). Verified empirically with a module re-evaluation reproduction. Add a one/two-line comment stating this constraint, in the same style as the existing prototype-vs-instance comment above it.

lib/database/sqlcommenter.test.ts
llun: Added in cddd3da54: a new test attaches sqlcommenter to a second better-sqlite3 knex instance (same module-cached dialect class, so same prototype) and asserts the executed SQL carries the traceparent comment exactly once. Proven load-bearing: with the guard removed the test fails with two /*traceparent= markers.

lib/database/index.ts
llun: Added in cddd3da54: a two-line comment above SQLCOMMENTER_ATTACHED stating that Symbol.for's global-registry key survives module re-evaluation under Next.js dev Fast Refresh, where a per-module Symbol() would never match and query would be re-wrapped on every reload, stacking duplicate comments.

PR #1702, opened 2026-09-01
fix: make profile and actor loading skeletons visibly shimmer

llun:
## Summary

The `(timeline)/[actor]` loading skeletons — profile, followers, following, and status detail — were effectively invisible in light mode. They used Tailwind's stock `animate-pulse` over `bg-muted`, and the light-theme `--muted` (96.1% lightness) sits ~4 lightness points off the white background; since `animate-pulse` fades the block *toward* the background, the entire animated range stayed within ~2 points of lightness. (An earlier shimmer attempt, `383a73fca`, shipped invisible for this same reason and was reverted twice — the guards below exist so this one can't quietly regress the same way.)

This PR replaces that treatment with a shared `skeleton` utility in `app/globals.css`:

- **New `--skeleton` / `--skeleton-highlight` tokens, declared in both themes.** The base tone is clearly distinct from the surfaces it loads over (90% light, 20% dark). `--muted` is untouched — it is a hover/track surface used all over the app.
- **A shimmer sweep**: a `::after` gradient band translated across each block. The animation is a `transform`, not an animated `background-position`, so it stays compositor-friendly, and `overflow: hidden` clips the band to each block's own border radius.
- **`prefers-reduced-motion: reduce` removes the band entirely**, leaving the static tinted block. Because the base is now visible at rest, reduced-motion users get a real placeholder where the old treatment gave them near-white nothing.
- The four `[actor]` loading files swap `animate-pulse bg-muted` for `skeleton`; every shape and layout class is unchanged.

## Tests

- New `app/globals.skeleton.test.ts` pins the stylesheet itself: the `.skeleton` rule exists (guarding the `no-scrollbar` applied-while-defined-nowhere failure mode), both themes declare the tokens, the base tone stays ≥8 lightness points off its surface in each theme, the composited shimmer band stays ≥5 lightness points off the base in each theme (a highlight blending to the base's own lightness renders the sweep invisible), and the reduced-motion block stills the band.
- Each of the four loading tests asserts **every leaf `<div>` carries `skeleton`** (jsdom paints no CSS, so class presence is the guard) — a bare `length > 0` was shown by mutation to pass with a single styled placeholder — plus that `animate-pulse` is gone.

## Verification

- `yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test` all green locally (932 files / 12,451 tests).
- Verified in a real browser against a seeded local SQLite instance (AGENTS.md recipe): light and dark skeletons are clearly visible with the band sweeping (two frames show it at different positions); computed styles confirm `skeleton-shimmer` animates the `::after` band; under emulated `prefers-reduced-motion: reduce` the band computes to `display: none`. Light / dark / reduced-motion screenshots were captured in the working session (`gh` cannot upload images to PR bodies — they can be dragged in from there).

## Review

The sub-agent review loop ran 3 rounds (5 lenses per full round + adversarial verification of each finding): round 1 produced 4 verified findings (the two test-coverage gaps above, found by mutation; an over-claiming CSS comment, trimmed; two redundant `relative` tokens, dropped), round 2 produced 2 minor consistency findings in the new test code, round 3 was clean. All 6 inline threads are replied-to and resolved.

### DIFF
--- a/app/(timeline)/[actor]/loading.test.tsx
+++ b/app/(timeline)/[actor]/loading.test.tsx
@@ -21,14 +21,18 @@ describe('[actor] loading', () => {
     expect(screen.getByLabelText('Loading profile')).toBeInTheDocument()
   })
 
-  it('renders placeholders with the shimmer skeleton utility', () => {
+  it('renders every placeholder with the shimmer skeleton utility', () => {
+    // jsdom paints no CSS so classes are the observable; every leaf <div> in
+    // this skeleton is a placeholder (containers always hold further divs),
+    // and a bare `length > 0` missed both a partial strip and a future
+    // unstyled row; the `.skeleton` definition itself is guarded by
+    // app/globals.skeleton.test.ts.
     const { container } = render(<Loading />)
-
-    // jsdom paints no CSS, so the class is the observable: the shimmer lives
-    // on the `skeleton` utility (app/globals.css, guarded by
-    // app/globals.skeleton.test.ts), and the old animate-pulse-on-bg-muted
-    // treatment — near-invisible in light mode — must not come back.
-    expect(container.querySelectorAll('.skeleton').length).toBeGreaterThan(0)
+    const leaves = Array.from(container.querySelectorAll('div')).filter(
+      (el) => el.children.length === 0
+    )
+    expect(leaves.length).toBeGreaterThan(0)
+    leaves.forEach((el) => expect(el).toHaveClass('skeleton'))
     expect(container.querySelector('.animate-pulse')).toBeNull()
   })
 })
--- a/app/(timeline)/[actor]/loading.test.tsx
+++ b/app/(timeline)/[actor]/loading.test.tsx
@@ -20,4 +20,15 @@ describe('[actor] loading', () => {
 
     expect(screen.getByLabelText('Loading profile')).toBeInTheDocument()
   })
+
+  it('renders placeholders with the shimmer skeleton utility', () => {
+    const { container } = render(<Loading />)
+
+    // jsdom paints no CSS, so the class is the observable: the shimmer lives
+    // on the `skeleton` utility (app/globals.css, guarded by
+    // app/globals.skeleton.test.ts), and the old animate-pulse-on-bg-muted
+    // treatment — near-invisible in light mode — must not come back.
+    expect(container.querySelectorAll('.skeleton').length).toBeGreaterThan(0)
+    expect(container.querySelector('.animate-pulse')).toBeNull()
+  })
 })
--- a/app/(timeline)/[actor]/loading.tsx
+++ b/app/(timeline)/[actor]/loading.tsx
@@ -4,56 +4,56 @@ export const ProfileLoading: FC = () => {
   return (
     <div aria-busy="true" aria-label="Loading profile" className="space-y-6">
       <section className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
-        <div className="relative h-36 animate-pulse bg-muted md:h-52" />
+        <div className="skeleton relative h-36 md:h-52" />
 
         <div className="relative px-6 pb-6">
-          <div className="relative -mt-10 h-20 w-20 animate-pulse rounded-full border-4 border-background bg-muted" />
+          <div className="skeleton relative -mt-10 h-20 w-20 rounded-full border-4 border-background" />
 
           <div className="mt-4 flex flex-col gap-3 sm:flex-row sm:items-start sm:justify-between">
             <div className="space-y-2">
-              <div className="h-7 w-48 animate-pulse rounded-md bg-muted" />
-              <div className="h-4 w-32 animate-pulse rounded-md bg-muted" />
+              <div className="skeleton h-7 w-48 rounded-md" />
+              <div className="skeleton h-4 w-32 rounded-md" />
             </div>
-            <div className="h-9 w-28 animate-pulse rounded-md bg-muted" />
+            <div className="skeleton h-9 w-28 rounded-md" />
           </div>
 
           <div className="mt-4 space-y-2">
-            <div className="h-4 w-3/4 animate-pulse rounded bg-muted" />
-            <div className="h-4 w-1/2 animate-pulse rounded bg-muted" />
+            <div className="skeleton h-4 w-3/4 rounded" />
+            <div className="skeleton h-4 w-1/2 rounded" />
           </div>
 
           <div className="mt-5 flex flex-wrap gap-6">
-            <div className="h-4 w-16 animate-pulse rounded bg-muted" />
-            <div className="h-4 w-20 animate-pulse rounded bg-muted" />
-            <div className="h-4 w-20 animate-pulse rounded bg-muted" />
+            <div className="skeleton h-4 w-16 rounded" />
+            <div className="skeleton h-4 w-20 rounded" />
+            <div className="skeleton h-4 w-20 rounded" />
           </div>
         </div>
       </section>
 
       <div className="space-y-4">
         <div className="flex gap-2">
-          <div className="h-9 w-20 animate-pulse rounded-lg bg-muted" />
-          <div className="h-9 w-20 animate-pulse rounded-lg bg-muted" />
-          <div className="h-9 w-20 animate-pulse rounded-lg bg-muted" />
+          <div className="skeleton h-9 w-20 rounded-lg" />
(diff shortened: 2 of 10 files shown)


app/(timeline)/[actor]/loading.test.tsx
```diff
@@ -20,4 +20,15 @@ describe('[actor] loading', () => {
...
   })
+
+  it('renders placeholders with the shimmer skeleton utility', () => {
+    const { container } = render(<Loading />)
+
+    // jsdom paints no CSS, so the class is the observable: the shimmer lives
+    // on the `skeleton` utility (app/globals.css, guarded by
+    // app/globals.skeleton.test.ts), and the old animate-pulse-on-bg-muted
+    // treatment — near-invisible in light mode — must not come back.
+    expect(container.querySelectorAll('.skeleton').length).toBeGreaterThan(0)
```
Comment (llun) [inline app/(timeline)/[actor]/loading.test.tsx]: **[round 1 · testing — verified]** `length > 0` cannot see a partial regression: stripping `skeleton` from all but one placeholder leaves this green (mutation-tested), and a future copy-pasted row without the class is equally invisible to it. In all four loading skeletons every leaf `<div>` is a placeholder and every container `<div>` contains further divs (verified mechanically), so assert the invariant instead: collect `container.querySelectorAll('div')` with `children.length === 0`, expect the list non-empty and each element `toHaveClass('skeleton')`. Catches both the subtractive and the additive case with no count to maintain. Applies to all four loading.test.tsx files.

app/globals.skeleton.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  it('defines the .skeleton class the loading skeletons apply', () => {
+    expect(css).toMatch(/\.skeleton\s*\{/)
+    expect(css).toMatch(/\.skeleton::after\s*\{/)
+    expect(css).toMatch(/@keyframes skeleton-shimmer/)
+  })
+
+  it.each([':root', '.dark'])('declares the skeleton tokens in %s', (theme) => {
+    const block = blockOf(theme)
+    expect(tokenOf(block, '--skeleton')).toBeTruthy()
+    expect(tokenOf(block, '--skeleton-highlight')).toBeTruthy()
```
llun, line 72: **[round 1 · testing — verified]** The highlight token is only checked for existence, and the highlight IS the shimmer: setting dark's `--skeleton-highlight` to `hsl(0 0% 20% / 0.75)` — the dark base's own lightness — makes the band invisible in dark mode while all six tests stay green (mutation-tested). Add a per-theme guard: parse the highlight's lightness and alpha (alpha defaults to 1), compute the composited band lightness over the base (`alpha·Lh + (1−alpha)·Lb`, exact for these grey tokens), and assert `|L_eff − L_base| ≥ 5` (today ≈7.5 light / 7.2 dark, matching this file's few-points-of-slack convention; the parser must keep throwing loudly on non-hsl values).

app/globals.css
```diff
@@ -166,6 +177,13 @@
      lightness, not shadow). */
   --control-active: hsl(0 0% 30%);
 
+  /* Loading skeleton: on the dark ramp visibility comes from going UP — the
+     base sits above the 9% card and the 17% muted track. The highlight is a
```
llun: **[round 1 · design-system — verified]** This comment claims the base clears both the card and the 17% muted track, but only the card half is guarded by app/globals.skeleton.test.ts, and the muted margin is only 3 points. Verification found skeletons never render on a muted surface (every call site sits on the page background or a `bg-background/80` card), and this block's sibling elevation comments are likewise unguarded narrative — so rather than adding an arbitrary assertion, trim the unguarded clause: '…the base sits above the 9% card.'

app/(timeline)/[actor]/loading.tsx
```diff
@@ -4,56 +4,56 @@ export const ProfileLoading: FC = () => {
   return (
     <div aria-busy="true" aria-label="Loading profile" className="space-y-6">
       <section className="overflow-hidden rounded-2xl border bg-background/80 shadow-sm">
-        <div className="relative h-36 animate-pulse bg-muted md:h-52" />
+        <div className="skeleton relative h-36 md:h-52" />
```
Comment (llun) [inline app/(timeline)/[actor]/loading.tsx]: **[round 1 · quality — verified]** `relative` here and on the avatar div (line 10) is now redundant: `.skeleton` itself declares `position: relative` for its `::after` band, both rules land in the same `utilities` cascade layer with equal specificity (confirmed against the real Tailwind compile), and these are the only 2 of 49 skeleton call sites carrying the token. Drop `relative` from both classNames.

llun (commented):
Round 1 of the sub-agent review loop: 4 verified findings (2 test-coverage gaps found by mutation, 1 comment/guard mismatch, 1 redundant class).

app/(timeline)/[actor]/loading.test.tsx
Comment (llun) [reply app/(timeline)/[actor]/loading.test.tsx]: Fixed in 0d37a40d5: all four loading tests now collect every leaf <div> and assert each carries `skeleton` (plus the animate-pulse null check). Verified non-vacuous by stripping the class from a single leaf — the test fails.

app/globals.skeleton.test.ts
llun: Fixed in 0d37a40d5: `lightnessAndAlphaOf` parses lightness+alpha, and a per-theme `it.each` asserts the composited band lightness differs from the base by ≥5 (today ≈7.5 light / 7.2 dark). Verified non-vacuous with a same-lightness highlight — the test fails.

app/globals.css
llun: Fixed in 0d37a40d5: trimmed the clause to 'the base sits above the 9% card' so the comment claims exactly what the tests guard.

app/(timeline)/[actor]/loading.tsx
Comment (llun) [reply app/(timeline)/[actor]/loading.tsx]: Fixed in 0d37a40d5: dropped the redundant `relative` token from both classNames; `.skeleton` supplies `position: relative`.

app/globals.skeleton.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  it.each([':root', '.dark'])(
+    'keeps the %s shimmer band visible against the base',
+    (theme) => {
+      // The highlight IS the shimmer: a band whose composited lightness matches
+      // the base renders the sweep invisible while every other assertion stays
+      // green. Lightness-space compositing is exact for these grey (0% sat)
+      // tokens, and the parser throws loudly on any other value form.
+      const block = blockOf(theme)
+      const base = lightnessAndAlphaOf(tokenOf(block, '--skeleton'))
```
llun: **[round 2 · quality]** `base` is parsed with `lightnessAndAlphaOf` but only `.lightness` is ever read — `base.alpha` is dead. The file already has `lightnessOf` for the lightness-only case and its two other callers use it; make this the third consistent call site (`const base = lightnessOf(...)`, then use `base` directly in the composite and the assertion). Same math, and the helper name states the opaque-base assumption a reader currently has to infer from the dead property.

app/globals.skeleton.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    /hsl\(\s*[\d.]+\s+[\d.]+%\s+([\d.]+)%(?:\s*\/\s*([\d.]+))?\s*\)/
+  )
+  if (!match) throw new Error(`Not a space-separated hsl() value: ${value}`)
+  return {
+    lightness: Number(match[1]),
+    alpha: match[2] === undefined ? 1 : Number(match[2])
+  }
+}
+
+const lightnessOf = (value: string): number =>
```
llun, line 59: **[round 2 · quality]** `lightnessOf` lost its doc comment when the JSDoc moved onto `lightnessAndAlphaOf`; every other helper in the file carries a one-liner. Restore one saying why the two closely-named helpers coexist, e.g. 'Lightness only, for callers that never need the alpha channel.'

llun (commented):
Round 2 of the sub-agent review loop: 4 lenses clean, 2 minor consistency findings from the quality lens.

app/globals.skeleton.test.ts
llun: Fixed in 99bfd58e1: the base is now `lightnessOf(tokenOf(block, '--skeleton'))` and the composite/assertion use it directly — third consistent lightness-only call site, no dead property.

app/globals.skeleton.test.ts
llun: Fixed in 99bfd58e1: restored a one-liner on `lightnessOf` ('Lightness only, for callers that never need the alpha channel.').

PR #1703, opened 2026-09-01
feat: add Mastodon-style RE: fallback link to federated quote posts

llun:
## Summary

Quote posts now federate with Mastodon's legacy fallback line: `<p class="quote-inline">RE: <a href="…">…</a></p>` prepended to the Note's `content`, so servers that do not understand the FEP-044f quote fields still show a link to the quoted post. Matches current Mastodon behavior (`TextFormatter#add_quote_fallback`): prepended, only while the quote fields themselves are emitted (live `pending`/`accepted` edge — the gate is shared **by construction** with `getQuoteNoteFields`), and skipped when the content already contains the quoted url (which also keeps a boosted remote quote post from being double-prefixed).

- **Outbound**: new `addQuoteFallbackToContent` in `lib/activities/quoteNoteFields.ts`, wired into both Note emitters — `getNoteFromStatus` (Create/Update delivery and the QuoteRequest `instrument`) and `toActivityPubObject` (AP status GET, both outboxes, `/replies`, actor archive). The link target is `quotedStatusId`, the same value already emitted as `quote`/`quoteUrl`/`quoteUri`/`_misskey_quote`. The url is HTML-escaped via the existing `escapeHtml`.
- **Inbound (fixes a live rendering bug)**: our sanitizer stripped `class` from `<p>` entirely, so a Mastodon 4.5 quote post received today shows its RE: line as a *visible duplicate* above our QuoteCard. `quote-inline` is now allowlisted (on `p` — filtered to that one class — plus `span` and `a`, covering the older appended-span convention), and `cleanClassName` hides it when — and only when — `post.tsx` renders a quote card. With no resolved edge the fallback stays visible, per Mastodon's client guidance: it is then the reader's only clue. Since remote HTML is sanitized at render, already-stored posts are repaired retroactively.
- **Link previews**: the extractor classifies `quote-inline` as a hiding class (its coupling test forces the decision), so a fallback link can never win a preview card — erring toward no card, the direction that module already commits to.

## Deliberate non-goals

- No fallback in the Mastodon REST `content` for local statuses — the API already serializes the structured `quote` entity, and injecting into the shared render pipeline would leak the line into the web UI, notifications, and emails. Legacy *servers* get the fallback; legacy *clients* of our own API are unchanged.
- `QuoteCard` excerpts deliberately keep an inner RE: line: the card renders the quoted status at depth 1 with no nested quote card, and per Mastodon's client docs the fallback is only redundant when the quote it describes is displayed.
- Polls untouched (composer forbids quote+poll; the `Question` branch emits no quote fields).
- `toActivityPubObject` still emits raw markdown for local text — pre-existing asymmetry, out of scope.

## Test results

Full Definition of Done gate green: `prettier` → `lint` → `typecheck` → `build` → `test` (931 files, 12,467 tests). New coverage: helper unit table in `quoteNoteFields.test.ts` (state gate, skip-if-linked, escaping, empty content), fallback assertions in `getNoteFromStatus.test.ts` and `status.test.ts` (`toActivityPubObject`), sanitizer keeps/strips cases in `sanitizeText.test.ts`, extractor coupling buckets + ancestor-inheritance case in `extractUrl.test.ts`, `cleanClassName` option behavior, and `post.test.tsx` hidden-with-edge / visible-without-edge component tests.

## Browser verification

Verified against a local SQLite instance with seeded mock data (AGENTS.md recipe): quoting a post from the timeline renders the QuoteCard with **no** RE: line in the body, and the AP GET for the same status answers:

```json
"content": "<p class=\"quote-inline\">RE: <a href=\"https://localhost:3000/users/testuser/statuses/01a05cfd-…\">https://localhost:3000/users/testuser/statuses/01a05cfd-…</a></p>This photo set is fantastic — quoting it for later."
```

with `quote`/`quoteUrl`/`quoteUri`/`_misskey_quote` beside it. (Screenshots were taken during in-browser verification; this environment cannot attach images to the PR body — the timeline rendering is also pinned by the two new `post.test.tsx` cases.)

## Notes

- No migrations, no config changes, `package.json` version untouched.
- Docs updated in the same PR: AGENTS.md (`ALLOWED_CONTENT_CLASSES` enumeration, extractor hidden-class phrasing, quote-emitter description) and REVIEW.md (hidden-class bullet, FEP-044f entry now covers the content fallback).


lib/activities/quoteNoteFields.ts
```diff
@@ -44,6 +45,33 @@ export const getQuoteNoteFields = (
...
+ * `post.tsx` does whenever it renders a quote card.
+ */
+export const addQuoteFallbackToContent = (
+  content: string,
+  quoteEdge: StatusQuote | null | undefined
+): string => {
+  const fields = getQuoteNoteFields(quoteEdge)
+  if (!fields) return content
+  const url = fields.quote
+  if (!url || content.includes(url)) return content
```
llun: **[correctness, confirmed] The dedup check misses the HTML-escaped form of the url, so a boosted remote quote post can be double-prefixed.**

`content.includes(url)` tests the RAW id, but stored remote content carries the ESCAPED rendering (`&` → `&amp;`) — the only form any fallback-emitting sender (this codebase included) ever writes into HTML. For a quoted id containing `&` (e.g. `…/statuses/9?ref=abc&v=2`), a remote quote post that already carries its origin's fallback misses the check and gets a second `<p class="quote-inline">RE: …</p>` prepended when re-served through `toActivityPubObject` — reachable via a local boost's AP GET (`getOriginalStatus` unwraps to the remote original, no locality guard) and via the `/replies` collection embedding a remote quote-post reply. Verified by running the real module at this commit: `quote-inline` occurrence count comes out 2, not 1; non-quote-aware receivers (the fallback's whole audience) render the RE: line twice. The doc comment's "stops a boosted remote quote post from being double-prefixed" only holds for ids with no escapable characters. (Mastodon's own `add_quote_fallback` shares the raw-url check, so this is an ecosystem edge case — but our comment claims more than that check delivers.)

Fix: compute `escaped` first and skip when the content contains either form: `if (content.includes(url) || content.includes(escaped)) return content`. Validated against the existing 21 tests plus the repro.

lib/activities/quoteNoteFields.ts
```diff
@@ -44,6 +45,33 @@ export const getQuoteNoteFields = (
   }
 }
 
+export const QUOTE_INLINE_CLASS = 'quote-inline'
```
llun: **[quality, confirmed-minor] `QUOTE_INLINE_CLASS` is exported but nothing imports it** — even this module's own test spells the literal. Verification adjudicated the alternatives: wiring the sanitizer/extractor/cleanClassName to import it would invert dependency direction (client-rendered text utilities reaching into the AP emission module) against the repo's "dependency-free module" doctrine, and the drift hazard is largely theoretical — `quote-inline` is Mastodon's wire-format marker (not unilaterally renameable) and every consumer file pins its own literal with tests, plus the sanitizer↔extractor coupling test. The established house style in this exact file trio (`invisible`, `ellipsis`, `mention`, `hashtag`) is independent literals.

Fix: drop the `export` keyword; keep the private const for the template.

lib/utils/text/sanitizeText.ts
```diff
@@ -52,12 +57,18 @@ export const SANITIZED_OPTION = {
   allowedAttributes: {
     a: ['href', 'rel', 'class', 'translate', 'target'],
     span: ['class', 'translate'],
+    p: ['class'],
```
llun, line 60: **[testing, confirmed — no change needed] `p: ['class']` in `allowedAttributes` is redundant**: sanitize-html 2.17.7 implicitly allows `class` on any tag with an `allowedClasses` entry (`index.js` lines 184–191, and documented in its README), so `allowedClasses.p` alone carries the behavior — removing this line leaves all 48 sanitizer tests green.

Resolved as keep-as-is after verification: `a` and `span` have carried the identical redundant `class` entry since before this PR (the `allowedClasses` map that made them redundant arrived in #1469), so removing only the new `p` entry would be incoherent, and stripping all three is a separate whole-file cleanup, not this PR's scope. The real gate (`allowedClasses.p: ['quote-inline']`) is present, working, and test-pinned; no comment added per the repo's comment-churn precedent.

llun (commented):
Sub-agent review round 1: five lenses (correctness, design-system, quality, scope-regression, testing) ran in parallel; every finding below survived an adversarial verification pass that reproduced it against the real module. Two need changes, one is a reasoned keep-as-is.

lib/activities/quoteNoteFields.ts
llun: Fixed in 4b20b4c9f: `escaped` is computed before the skip, which now tests both forms — `content.includes(url) || content.includes(escaped)` — plus a regression test with a quoted id carrying `&` against content holding only the `&amp;` rendering (fails on the pre-fix code). All 22 helper tests green.

lib/activities/quoteNoteFields.ts
llun: Fixed in 4b20b4c9f: dropped the `export` keyword — `QUOTE_INLINE_CLASS` stays as the module-private const backing the template. Consumers keep their tested literals per the file trio's established style (`invisible`/`ellipsis`/`mention`/`hashtag` precedent).

lib/utils/text/sanitizeText.ts
llun: No change, per the verification in the thread: the working gate is `allowedClasses.p`, and `a`/`span` have carried the same redundant `class` entry since #1469 — removing only `p` would be incoherent, and a whole-file cleanup of all three is out of this PR's scope. Resolving as addressed-by-decision.

lib/activities/quoteNoteFields.test.ts
```diff
@@ -69,6 +70,72 @@ describe('getQuoteNoteFields', () => {
...
+    { label: 'null', value: null },
+    { label: 'undefined', value: undefined }
+  ])('leaves content alone when the edge is $label', ({ value }) => {
+    expect(addQuoteFallbackToContent('<p>hi</p>', value)).toEqual('<p>hi</p>')
+  })
+
+  it('skips when the content already contains the quoted url', () => {
+    const linked = `<p>see <a href="${QUOTED_STATUS_ID}">this</a></p>`
+    expect(addQuoteFallbackToContent(linked, edge())).toEqual(linked)
+  })
```
llun, line 105: **[testing, confirmed by mutation run] The raw-form half of the dedup skip was unguarded.** A mutant reduced to `if (content.includes(escaped)) return content` survived the entire suite — every existing fixture's url contains none of the five characters `escapeHtml` touches, so raw and escaped are byte-identical there and the raw branch is indistinguishable from the escaped one. The raw branch has a traced production path: a LOCAL status stores raw author text (`toActivityPubObject` does no markdown conversion), so a hand-pasted link keeps its literal `&` and only the raw check can match it.

Fixed in 0fea883ef: added `skips when the content contains only the RAW form of an escapable quoted url` — a raw `&`-carrying mention. Verified both directions: 23/23 green on the real code, and exactly this test fails against the escaped-only mutant.

llun (commented):
Sub-agent review round 2 (scoped to round 1's fix, `0c9e313a1..4b20b4c9f`): correctness and scope-regression lenses came back clean; the testing lens surfaced one mutation-coverage gap, self-verified by running the mutant.

llun:
Sub-agent review loop concluded. Round 1: five lenses (correctness, design-system, quality, scope-regression, testing) + adversarial verification per finding → 2 fixes (escaped-url dedup in 4b20b4c9f, un-export in the same commit) and 1 reasoned keep-as-is, all threads resolved. Round 2 (scoped to round 1's fix): correctness and scope-regression clean; testing found the raw-branch mutation-coverage gap → killed by the test in 0fea883ef, verified in both directions. Round 3 (scoped to round 2's fix): clean, one comment-precision nit addressed in the final commit. Full gate green at head: prettier, lint, typecheck, build, 12,469 tests.

PR #1716, opened 2026-09-01
minor: list remote accounts' followers and following over the Mastodon API

llun:
## Summary

`GET /api/v1/accounts/:id/followers` and `/following` answered from the `follows` table, which only holds relationships this instance took part in. Opening a remote account (e.g. The Verge) in a Mastodon client therefore showed one local follower under **Followers** and "No People Found" under **Following**. Stock Mastodon behaves the same way; this PR is an enhancement over it.

For a **remote** actor the two routes now read the actor's own `followers` / `following` ActivityPub collections and serialize each listed actor as a Mastodon `Account`:

- **`lib/services/mastodon/remoteFollowCollection.ts`** (new) — fetches the actor's Person document and the collection page (via the existing `getActorCollections`, signed by the federation signing actor) inside a 60 s **promise cache** keyed per actor, field and page, so a re-opened sheet or two concurrent viewers share one round of fetches and a failure is never cached. Stored actors are hydrated in one `getMastodonActorsFromIds` batch; at most **20 unknown actors per page** are resolved through `recordActorIfNeeded` (5 at a time) and the rest of that page's unknowns are dropped rather than stalling the request. A page cursor is validated against the actor's own collection with `isCollectionPageUrl` (the same check `remote-statuses` applies to `page_url`) and refused otherwise.
- **`app/api/v1/accounts/[id]/followCollectionHandler.ts`** (new) — one handler for both routes, which were near-identical copies. Local actors keep the exact previous behaviour. For remote actors the remote page URLs ride in the Mastodon cursors (`max_id` = page `next`, `min_id` = page `prev`), so an unmodified client paginates by sending the `Link` header's value straight back. A URL cursor that is not a page of that actor's collection is a `400`.
- **Fallbacks keep today's behaviour as the floor.** The remote read is only made for a signed-in viewer (the route is public and each page can cost a signed fetch per unknown actor — the same line `search` / `accounts/lookup` draw around `resolve=true`). An anonymous caller, a blocked domain (`canFederateWithDomain`, checked per request rather than inside the cache), an unreachable server, an actor document with no such collection, and a hidden collection (Mastodon's "hide your social graph", which publishes a size but no page) all serve the locally-known rows.
- Docs: new bullet in `docs/mastodon-api-compatibility.md`.

Known limitation, unchanged here: this instance's own ActivityPub `followers`/`following` collections still publish only `totalItems` with no page, so another Activity.next instance cannot list our users' follows this way. That is a separate change.

### Tests

- `lib/services/mastodon/remoteFollowCollection.test.ts` — page order, unknown-actor recording (once, in place), the per-page cap and its ordering, item shapes (object with id, blank node, no id, duplicates), cursor validation (same collection / other collection / other host), every `unavailable` cause, concurrent-miss collapsing including the Person fetch, TTL expiry, and failure non-caching.
- Both `route.test.ts` files — remote branch: served page + `Link` cursors carrying the remote page URLs, `max_id`/`min_id` forwarded as `pageUrl`, `400` on a refused cursor, anonymous → local rows with no remote call, fallback to local rows on `unavailable` and on a throw, blocked domain → no remote call, and a URL cursor ignored on the local fallback path. Existing local-actor tests untouched.

Full gate run locally in order: `yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test` (935 files, 12593 tests) — all green.

## Checklist

https://claude.ai/code/session_01M3nuwhCuUMGhpSPAM3oJBf

---
_Generated by [Claude Code](https://claude.ai/code/session_01M3nuwhCuUMGhpSPAM3oJBf)_

### DIFF
--- a/app/api/v1/accounts/[id]/followCollectionHandler.ts
+++ b/app/api/v1/accounts/[id]/followCollectionHandler.ts
@@ -1,0 +1,280 @@
+import { NextRequest } from 'next/server'
+import { z } from 'zod'
+
+import { Database } from '@/lib/database/types'
+import { canFederateWithDomain } from '@/lib/services/federation/domainPolicy'
+import { getFederationSigningActorSafe } from '@/lib/services/federation/getFederationSigningActor'
+import { headerHost } from '@/lib/services/guards/headerHost'
+import {
+  RemoteFollowField,
+  getRemoteFollowCollectionPage
+} from '@/lib/services/mastodon/remoteFollowCollection'
+import { resolveActorIdParam } from '@/lib/services/mastodon/resolveClientId'
+import { Actor } from '@/lib/types/domain/actor'
+import { clampedLimit } from '@/lib/utils/clampedLimit'
+import { HttpMethod } from '@/lib/utils/http-headers'
+import { logger } from '@/lib/utils/logger'
+import { buildPaginationLinkHeader } from '@/lib/utils/paginationLinkHeader'
+import { ERROR_400, apiCorsError, apiResponse } from '@/lib/utils/response'
+import { toLoggableError } from '@/lib/utils/toLoggableError'
+
+// Shared handler for GET /api/v1/accounts/:id/following and /followers.
+// https://docs.joinmastodon.org/methods/accounts/#following
+// https://docs.joinmastodon.org/methods/accounts/#followers
+//
+// Public with optional auth. Paginated with Mastodon id cursors + Link
+// headers, mirroring the accounts/:id/statuses route.
+//
+// A LOCAL actor's list is the `follows` table, and the cursor is the follow-row
+// id (the column getFollowing/getFollowers paginate on), not the account id.
+//
+// A REMOTE actor's list is read live from the collection its own server
+// publishes — see `lib/services/mastodon/remoteFollowCollection.ts` — because
+// the `follows` table only holds relationships this instance took part in,
+// which for a remote actor is one local follower and nothing followed. On that
+// path the cursor is the remote page URL: `max_id` carries the page's `next`,
+// `min_id` its `prev`, so an unmodified Mastodon client paginates by sending
+// the Link header's value straight back. A cursor that parses as a URL is
+// accepted only for a page of THAT actor's collection (the service applies
+// the same `isCollectionPageUrl` check `remote-statuses` applies to its
+// `page_url`), and is a 400 otherwise. A cursor that is not a URL is a follow-row id from the local
+// path and is answered from local rows.
+//
+// The remote read is only made for a signed-in viewer. The route is public,
+// and each page can cost a signed fetch per unknown actor on it, so an
+// anonymous caller gets the local rows exactly as before — the same line
+// `search` and `accounts/lookup` draw around `resolve=true`. Every failure on
+// the remote path (blocked domain, unreachable server, hidden collection)
+// also falls back to the local rows, so today's behaviour is the floor.
+
+export const FOLLOW_COLLECTION_CORS_HEADERS = [
+  HttpMethod.enum.OPTIONS,
+  HttpMethod.enum.GET
+]
+
+const CORS_HEADERS = FOLLOW_COLLECTION_CORS_HEADERS
+
+const FollowCollectionQueryParams = z.object({
+  max_id: z.string().optional(),
+  since_id: z.string().optional(),
+  min_id: z.string().optional(),
+  limit: clampedLimit(80, 40)
+})
+
+interface HandleParams {
+  req: NextRequest
+  database: Database
+  currentActor: Actor | null
+  encodedAccountId: string | undefined
+  field: RemoteFollowField
+}
+
+const isUrlCursor = (cursor: string | undefined): cursor is string => {
+  if (!cursor) return false
+  try {
+    const url = new URL(cursor)
+    return url.protocol === 'https:' || url.protocol === 'http:'
+  } catch {
+    return false
+  }
+}
+
+const getLocalFollows = async (
+  database: Database,
+  field: RemoteFollowField,
+  actorId: string,
+  cursors: {
+    limit: number
+    maxId?: string
+    minId?: string
+    sinceId?: string
+  }
+) => {
+  const follows =
+    field === 'following'
+      ? await database.getFollowing({ actorId, ...cursors })
+      : await database.getFollowers({ targetActorId: actorId, ...cursors })
+  return follows.map((follow) => ({
+    id: follow.id,
+    accountId: field === 'following' ? follow.targetActorId : follow.actorId
+  }))
+}
+
+export const handleFollowCollectionRequest = async ({
+  req,
+  database,
+  currentActor,
+  encodedAccountId,
+  field
+}: HandleParams) => {
+  if (!encodedAccountId) return apiCorsError(req, CORS_HEADERS, 400)
+
+  const id = await resolveActorIdParam(database, encodedAccountId)
+  const actor = await database.getActorFromId({ id })
+  if (!actor) return apiCorsError(req, CORS_HEADERS, 404)
+
+  const url = new URL(req.url)
+  const parsed = FollowCollectionQueryParams.safeParse(
+    Object.fromEntries(url.searchParams.entries())
+  )
+  if (!parsed.success) {
+    return apiResponse({
+      req,
+      allowedMethods: CORS_HEADERS,
+      data: ERROR_400,
+      responseStatusCode: 400
+    })
+  }
+  const { limit, max_id: maxId, min_id: minId, since_id: sinceId } = parsed.data
+
+  // Percent-encoded: the router hands the id over already decoded, and the
+  // resolver accepts a raw http(s) URI as an id form, so the raw value would
+  // emit a Link URL that does not route back here.
+  const path = `/api/v1/accounts/${encodeURIComponent(encodedAccountId)}/${field}`
+  const host = headerHost(req.headers)
+
+  const isLocalActor = Boolean(actor.privateKey)
+  const pageUrl = isUrlCursor(maxId)
+    ? maxId
+    : isUrlCursor(minId)
+      ? minId
+      : undefined
+
+  if (!isLocalActor && currentActor) {
+    const remote = await serveRemoteCollection({
+      req,
+      database,
+      actorId: actor.id,
+      field,
+      pageUrl,
+      limit,
+      host,
+      path
+    })
+    if (remote) return remote
+  }
+
+  // A remote page URL is meaningless against the follow-row ids the local path
+  // paginates on: it is either a client re-sending a Link cursor to a
+  // now-anonymous session, or one for a collection this instance could not
+  // read this time. Serve the first local page rather than compare a URL
+  // against row ids.
+  const localCursors = {
+    limit,
+    maxId: isUrlCursor(maxId) ? undefined : maxId,
+    minId: isUrlCursor(minId) ? undefined : minId,
+    sinceId: isUrlCursor(sinceId) ? undefined : sinceId
+  }
+  const orderedFollows = await getLocalFollows(
+    database,
+    field,
+    actor.id,
+    localCursors
+  )
+
+  // Batch-hydrate the accounts in a single query, then re-order to match
+  // orderedFollows (getMastodonActorsFromIds does not guarantee order).
+  const accountsById = new Map(
+    (
+      await database.getMastodonActorsFromIds({
+        ids: orderedFollows.map((follow) => follow.accountId)
+      })
+    ).map((account) => [account.url, account])
+  )
+  const accounts = orderedFollows
+    .map((follow) => accountsById.get(follow.accountId))
+    .filter((account): account is NonNullable<typeof account> =>
+      Boolean(account)
+    )
+
+  const additionalHeaders = buildPaginationLinkHeader({
+    host,
+    path,
+    limit,
+    nextMaxId:
+      orderedFollows.length > 0
+        ? orderedFollows[orderedFollows.length - 1].id
+        : null,
+    prevMinId: orderedFollows.length > 0 ? orderedFollows[0].id : null
+  })
+
+  return apiResponse({
+    req,
+    allowedMethods: CORS_HEADERS,
+    data: accounts,
+    additionalHeaders
+  })
+}
+
+// The remote branch. Answers null wherever the local rows should be served
+// instead; the one non-null failure is a URL cursor that does not belong to
+// this actor's collection, which is a client error rather than a fallback.
+const serveRemoteCollection = async ({
+  req,
+  database,
+  actorId,
+  field,
+  pageUrl,
+  limit,
+  host,
+  path
+}: {
+  req: NextRequest
+  database: Database
+  actorId: string
+  field: RemoteFollowField
+  pageUrl: string | undefined
+  limit: number
+  host: string | null | undefined
+  path: string
+}): Promise<Response | null> => {
+  try {
+    // Server-to-server fetches are signed by the headless instance actor,
+    // never the viewer's actor; a missing signer degrades to an unsigned
+    // fetch. Domain policy applies to every live-fetch surface, and is checked
+    // here per request rather than inside the cached read so a newly blocked
+    // domain is refused at once.
+    const [canFederate, signingActor] = await Promise.all([
+      canFederateWithDomain(database, actorId),
+      getFederationSigningActorSafe(database, `for remote account ${field}`)
+    ])
+    if (!canFederate) return null
+
+    const result = await getRemoteFollowCollectionPage({
+      database,
+      actorId,
+      field,
+      signingActor,
+      pageUrl
+    })
+    if (result.status === 'unavailable') return null
+    if (result.status === 'invalid-page') {
+      return apiResponse({
+        req,
+        allowedMethods: CORS_HEADERS,
+        data: ERROR_400,
+        responseStatusCode: 400
+      })
+    }
+
+    return apiResponse({
+      req,
+      allowedMethods: CORS_HEADERS,
+      data: result.page.accounts,
+      additionalHeaders: buildPaginationLinkHeader({
+        host,
+        path,
+        limit,
+        nextMaxId: result.page.nextPageUrl,
+        prevMinId: result.page.prevPageUrl
+      })
+    })
+  } catch (error) {
+    logger.warn({
+      message: `Failed to read remote actor ${field} collection; serving local rows`,
+      actorId,
+      err: toLoggableError(error)
+    })
+    return null
+  }
+}
--- a/app/api/v1/accounts/[id]/followCollectionHandler.ts
+++ b/app/api/v1/accounts/[id]/followCollectionHandler.ts
@@ -34,11 +34,14 @@ import { toLoggableError } from '@/lib/utils/toLoggableError'
 // which for a remote actor is one local follower and nothing followed. On that
 // path the cursor is the remote page URL: `max_id` carries the page's `next`,
 // `min_id` its `prev`, so an unmodified Mastodon client paginates by sending
-// the Link header's value straight back. A cursor that parses as a URL is
-// accepted only for a page of THAT actor's collection (the service applies
-// the same `isCollectionPageUrl` check `remote-statuses` applies to its
-// `page_url`), and is a 400 otherwise. A cursor that is not a URL is a follow-row id from the local
-// path and is answered from local rows.
+// the Link header's value straight back. When the remote is consulted, a
+// cursor that parses as a URL is accepted only for a page of THAT actor's
+// collection (the service applies the same `isCollectionPageUrl` check
+// `remote-statuses` applies to its `page_url`) and is a 400 otherwise. On the
+// local path — a local actor, or any of the fallbacks below — a URL cursor is
+// ignored and the first local page served, because it names nothing among the
+// follow-row ids that path paginates on; a cursor that is not a URL is such a
+// row id.
 //
 // The remote read is only made for a signed-in viewer. The route is public,
 // and each page can cost a signed fetch per unknown actor on it, so an
@@ -47,12 +50,7 @@ import { toLoggableError } from '@/lib/utils/toLoggableError'
 // the remote path (blocked domain, unreachable server, hidden collection)
 // also falls back to the local rows, so today's behaviour is the floor.
 
-export const FOLLOW_COLLECTION_CORS_HEADERS = [
-  HttpMethod.enum.OPTIONS,
-  HttpMethod.enum.GET
-]
-
-const CORS_HEADERS = FOLLOW_COLLECTION_CORS_HEADERS
+export const CORS_HEADERS = [HttpMethod.enum.OPTIONS, HttpMethod.enum.GET]
 
 const FollowCollectionQueryParams = z.object({
   max_id: z.string().optional(),
--- a/app/api/v1/accounts/[id]/followCollectionHandler.ts
+++ b/app/api/v1/accounts/[id]/followCollectionHandler.ts
@@ -1,3 +1,4 @@
+import { SpanStatusCode, trace } from '@opentelemetry/api'
 import { NextRequest } from 'next/server'
 import { z } from 'zod'
 
@@ -268,6 +269,13 @@ const serveRemoteCollection = async ({
       })
     })
   } catch (error) {
+    const span = trace.getActiveSpan()
+    if (span) {
+      span.recordException(
+        error instanceof Error ? error : new Error(String(error))
+      )
+      span.setStatus({ code: SpanStatusCode.ERROR })
+    }
     logger.warn({
       message: `Failed to read remote actor ${field} collection; serving local rows`,
       actorId,
--- a/app/api/v1/accounts/[id]/followers/route.test.ts
+++ b/app/api/v1/accounts/[id]/followers/route.test.ts
@@ -6,7 +6,7 @@ import { getFederationSigningActorSafe } from '@/lib/services/federation/getFede
 import { getRemoteFollowCollectionPage } from '@/lib/services/mastodon/remoteFollowCollection'
 import { seedDatabase } from '@/lib/stub/database'
 import { actorPublicId } from '@/lib/stub/publicIds'
-import { seedActor1 } from '@/lib/stub/seed/actor1'
+import { ACTOR1_ID, seedActor1 } from '@/lib/stub/seed/actor1'
 import { ACTOR2_ID } from '@/lib/stub/seed/actor2'
 import { EXTERNAL_ACTOR1 } from '@/lib/stub/seed/external1'
 import { urlToId } from '@/lib/utils/urlToId'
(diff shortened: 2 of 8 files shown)


lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,305 @@
...
+ * policy is deliberately NOT part of the cached work: the caller checks
+ * `canFederateWithDomain` per request, so a newly blocked domain is refused at
+ * once rather than served from a warm entry.
+ */
+export const getRemoteFollowCollectionPage = (
+  params: Params
+): Promise<RemoteFollowCollectionResult> => {
+  const { database, actorId, field, pageUrl } = params
+  const entries = getCacheEntries(database)
+  const key = `${field}|${actorId}|${pageUrl ?? ''}`
```
llun: **should-fix (abuse bound):** the cache key includes the raw `pageUrl`, and `isCollectionPageUrl` only compares protocol/host/path prefix, so `?page=2&n=1`, `?page=2&n=2`, … are all "valid" pages of the collection and all cache misses. A signed-in viewer can therefore issue unbounded parallel misses, each costing a Person fetch + collection root + page + up to 20 × `recordActorIfNeeded` (which is itself a Person fetch plus three collection roots) — roughly 80 signed outbound requests per request — and each nonce also drives `MAX_CACHED_PAGES` insertion-order eviction, flushing other viewers' entries.

Suggested fix: normalize the key (drop the fragment, sort the query) so the cache is not trivially bypassed by ordering, AND add a small process-wide in-flight cap on uncached remote reads that answers `unavailable` (local fallback) when exhausted, so adversarial use degrades to today's behaviour instead of fanning out. State in the module comment that the cache bounds repeat opens, not adversarial use.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,305 @@
...
+    pageUrl
+  })
+  // No collection, or a collection the remote publishes a size for but no
+  // page of — Mastodon's "hide your social graph" — is the caller's cue to
+  // fall back to what this instance knows locally.
+  if (!collection?.page) return { status: 'unavailable' }
+
+  const actorIds = Array.from(
+    new Set(
+      collection.page.orderedItems
```
llun: **should-fix (negative cache defeated):** `collection.page.orderedItems` is read straight off a `JSON.parse`d remote body with no validation. A page carrying `items` instead of `orderedItems` (Friendica/Pleroma shapes), or a non-array value, throws a `TypeError` here; that rejection hits the `.catch` on the cached promise, evicts the entry, and is retried — three signed fetches plus a warn with a stack — on every open, unlike an `unavailable` result which is cached for the TTL. Separately, a 5,000-item page is hydrated and serialized in full: `getMastodonActors` runs one unbatched `whereIn`, which throws "too many SQL variables" past 999 ids on SQLite, and the response ignores `limit` entirely.

Fix: `Array.isArray(page.orderedItems)` else `{ status: 'unavailable' }`, and slice the ids to a constant (the route's 80 max) before the lookup.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+  }
+  await Promise.all(
+    Array.from({ length: Math.min(concurrency, items.length) }, worker)
+  )
+}
+
+// The stored actor ids a batch of Mastodon accounts answers for. `uri` is the
+// actor id the row was looked up by; `url` is the profile URL, which for a
+// local actor is the same string and for a remote one usually is not. Keying
```
llun: **nit (comment is wrong):** the serializer emits both `url` and `uri` as `sqlActor.id` (`getMastodonAccountFromSQLActor`), so "for a remote one usually is not" is false. Keying on both is redundant rather than harmful — either key on `uri` alone or fix the comment.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,305 @@
...
+        // actor cannot be fetched; both simply leave the entry out of the page.
+        const actor = await recordActorIfNeeded({
+          actorId,
+          database,
+          signingActor
+        })
+        if (actor) recordedActorIds.push(actor.id)
+      } catch (error) {
+        logger.warn({
+          message: 'Failed to record remote follow collection actor',
```
llun: **nit ×2:** (a) `BlockedFederationDomainError` for a listed actor is expected policy, not a failure, but it is logged at `warn` with a stack once per blocked actor per page — log that class at `debug`. (b) The same unknown actor can be recorded concurrently from the `followers|X` and `following|X` cache entries; the loser's unique-constraint violation lands here and the actor is dropped from that page for the TTL. Re-read `getActorFromId` on failure before giving up.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,305 @@
...
+// deleting the entry on rejection so a failure is retried rather than served
+// for a minute. The actor's own Person document is fetched INSIDE the cached
+// work, because the collection URL lives on it: fetched outside, a cache hit
+// would still cost one signed remote request per page open.
+
+export type RemoteFollowField = 'followers' | 'following'
+
+export const REMOTE_FOLLOW_PAGE_TTL_MS = 60_000
+export const MAX_UNKNOWN_ACTORS_PER_PAGE = 20
+export const UNKNOWN_ACTOR_CONCURRENCY = 5
```
llun, line 66: **should-fix (unpinned constant):** `UNKNOWN_ACTOR_CONCURRENCY` is documented ("five at a time" in the docs bullet and the module header) but nothing tests it — set it to 1 or 100 and every test stays green. Hold `recordActorIfNeeded` open with `createDeferred` and assert at most 5 are in flight. `MAX_CACHED_PAGES` eviction is likewise untested, and both constants are exported but referenced nowhere outside this module.

---

app/api/v1/accounts/[id]/followCollectionHandler.ts
```diff
@@ -0,0 +1,280 @@
...
+//
+// A REMOTE actor's list is read live from the collection its own server
+// publishes — see `lib/services/mastodon/remoteFollowCollection.ts` — because
+// the `follows` table only holds relationships this instance took part in,
+// which for a remote actor is one local follower and nothing followed. On that
+// path the cursor is the remote page URL: `max_id` carries the page's `next`,
+// `min_id` its `prev`, so an unmodified Mastodon client paginates by sending
+// the Link header's value straight back. A cursor that parses as a URL is
+// accepted only for a page of THAT actor's collection (the service applies
+// the same `isCollectionPageUrl` check `remote-statuses` applies to its
```
Comment (llun) [inline app/api/v1/accounts/[id]/followCollectionHandler.ts]: **nit (comment overstates the 400):** a URL cursor "is a 400 otherwise" only holds when the remote read actually runs. An anonymous caller, a blocked domain, or an actor document that cannot be fetched all answer 200 with the first local page for the same foreign URL (the second comment further down describes this correctly). Reword to "when the remote is consulted, a 400 otherwise; on the local fallback the URL cursor is ignored and the first page served".

---

app/api/v1/accounts/[id]/followCollectionHandler.ts
```diff
@@ -0,0 +1,280 @@
...
+// `search` and `accounts/lookup` draw around `resolve=true`. Every failure on
+// the remote path (blocked domain, unreachable server, hidden collection)
+// also falls back to the local rows, so today's behaviour is the floor.
+
+export const FOLLOW_COLLECTION_CORS_HEADERS = [
+  HttpMethod.enum.OPTIONS,
+  HttpMethod.enum.GET
+]
+
+const CORS_HEADERS = FOLLOW_COLLECTION_CORS_HEADERS
```
Comment (llun) [inline app/api/v1/accounts/[id]/followCollectionHandler.ts]: **nit:** `CORS_HEADERS` is a pure alias of the export defined two lines above; use one name.

---

docs/mastodon-api-compatibility.md
```diff
@@ -395,6 +395,28 @@ are not part of the Mastodon API and are safe for Mastodon clients to ignore.
...
+  answers `GET /api/v1/accounts/:id/followers` and `/following` for a remote
+  account from the relationships it already knows about, so a small instance
+  shows one local follower and nothing followed. For a remote actor these two
+  routes instead read the actor's own `followers`/`following` ActivityPub
+  collections (`lib/services/mastodon/remoteFollowCollection.ts`), signed by
+  the federation signing actor, and serialize each listed actor as an Account
+  entity. The remote page URLs ride in the Mastodon cursors — the Link header's
+  `max_id` carries the page's `next` and `min_id` its `prev` — so an unmodified
+  client paginates by sending them back; a URL cursor is accepted only for a
+  page of that actor's collection and is a `400` otherwise, and `limit` is
```
llun: **nit (docs accuracy):** same as the handler comment — the `400` fires only when the remote read runs; an anonymous caller, a blocked domain or an unreachable actor document gets the first local page for the same URL cursor. Say so here.

---

app/api/v1/accounts/[id]/following/route.test.ts
```diff
@@ -113,3 +129,183 @@ describe('GET /api/v1/accounts/:id/following', () => {
...
+    })
+    vi.mocked(canFederateWithDomain).mockResolvedValue(true)
+    vi.mocked(getFederationSigningActorSafe).mockResolvedValue(
+      signingActor as never
+    )
+    vi.mocked(getRemoteFollowCollectionPage).mockResolvedValue({
+      status: 'ok',
+      page: {
+        accounts: await database.getMastodonActorsFromIds({
+          ids: [ACTOR2_ID]
```
Comment (llun) [inline app/api/v1/accounts/[id]/following/route.test.ts]: **should-fix (vacuous body assertion):** the seed has `EXTERNAL_ACTOR1` following exactly `ACTOR2_ID`, so the LOCAL following list for this actor is `[ACTOR2_ID]` — the same value the mocked remote page returns. The `toEqual([ACTOR2_ID])` body assertion in the headline remote test therefore passes with the whole remote branch deleted (only the `toHaveBeenCalledWith` and Link assertions still fail). Mock the remote page with `ACTOR3_ID` so the served body differs from the local rows. (The followers twin is fine: local followers = actor1, remote mock = actor2.)

---

app/api/v1/accounts/[id]/following/route.test.ts
```diff
@@ -113,3 +129,183 @@ describe('GET /api/v1/accounts/:id/following', () => {
...
+      params: Promise.resolve({ id: urlToId(EXTERNAL_ACTOR1) })
+    })
+
+    expect(response.status).toBe(200)
+    const data = await response.json()
+    const localFollows = await database.getFollowing({
+      actorId: EXTERNAL_ACTOR1,
+      limit: 40
+    })
+    expect(data.length).toBe(localFollows.length)
```
Comment (llun) [inline app/api/v1/accounts/[id]/following/route.test.ts]: **should-fix (length-only assertion):** this compares 1 to 1, so a remote page served by mistake and a local fallback are indistinguishable. Assert the account `uri`s (`[ACTOR2_ID]`, the local following row) instead.

---

app/api/v1/accounts/[id]/followers/route.test.ts
```diff
@@ -143,3 +158,183 @@ describe('GET /api/v1/accounts/:id/followers', () => {
...
+      params: Promise.resolve({ id: urlToId(EXTERNAL_ACTOR1) })
+    })
+
+    expect(response.status).toBe(200)
+    const data = await response.json()
+    const localFollows = await database.getFollowers({
+      targetActorId: EXTERNAL_ACTOR1,
+      limit: 40
+    })
+    expect(data.length).toBe(localFollows.length)
```
Comment (llun) [inline app/api/v1/accounts/[id]/followers/route.test.ts]: **should-fix (length-only assertion):** compares 1 to 1; assert the account `uri`s (actor1's id, the local follower row) so a wrongly served remote page cannot pass as a fallback.

---

lib/services/mastodon/remoteFollowCollection.test.ts
```diff
@@ -0,0 +1,409 @@
...
+    })
+    await getRemoteFollowCollectionPage({
+      database,
+      actorId: REMOTE_ACTOR_ID,
+      field: 'following'
+    })
+    expect(getActorCollections).toHaveBeenCalledTimes(3)
+  })
+
+  it('refetches once the TTL has passed', async () => {
```
llun, line 551: **nit:** the fake-timer test relies on the NEXT test's `beforeEach` (`vi.useRealTimers()`) for cleanup; put it in an `afterEach` so the last test in the file is protected from a future reorder.

---

llun (commented):
Sub-agent review round 1 of 2 (two parallel reviewers: correctness/security, and tests/conventions/docs). Findings are inline. Clean categories, verified: no SSRF/open-proxy path (target is a stored actor row, collection URL comes from the fetched Person, page URL is prefix-checked before any page fetch, every fetch goes through `safeRemoteFetch`); no regression on the local-actor path (byte-for-byte the previous logic); no thrown error escapes to a 500; mocks are reset per test; naming, import and API-response conventions satisfied.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 7bba009. The key is now normalized (fragment dropped, query sorted — `normalizePageUrlKey`), and `MAX_IN_FLIGHT_REMOTE_READS` (4, per process) caps uncached reads: past it the service answers `unavailable` uncached and the route serves the local rows. The module comment now says the cache bounds repeat opens only and the in-flight cap is what bounds a flood. Pinned by `answers unavailable, uncached, past MAX_IN_FLIGHT_REMOTE_READS uncached reads`, `keys the cache on the page, not on its spelling` and `does not cache a failure and frees its in-flight slot`.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 7bba009. A page whose `orderedItems` is missing or not an array now answers `{ status: 'unavailable' }`, which is cached for the TTL like any other unreadable collection (test: `answers unavailable when the page carries orderedItems that is not an array`, which also asserts the second open fetches nothing). The ids hydrated from one page are cut at `MAX_PAGE_ITEMS` (80) before the batch lookup (test: `serves at most MAX_PAGE_ITEMS actors from one remote page`, asserting one lookup of exactly 80 ids).

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 7bba009: `indexAccountsByActorId` keys on `uri` only and the comment now says why (`uri` is `sqlActor.id`, the key the page's actor ids resolve against). The handler's local path still keys on `url` — that code is the old routes' logic moved verbatim and kept byte-for-byte to avoid a regression there.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 7bba009 (`recordUnknownActor`): a `BlockedFederationDomainError` is logged at `debug`, and any other failure re-reads `getActorFromId` before giving the actor up, so a lost insert race keeps the actor on the page. Tests: `drops an actor that cannot be recorded instead of failing the page` (asserts one debug + one warn, for the right actor) and `keeps an actor whose insert lost a race but whose row now exists`.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 7bba009. `records unknown actors at most UNKNOWN_ACTOR_CONCURRENCY at a time` holds `recordActorIfNeeded` open with `createDeferred`, asserts exactly 5 in flight, then that settling one admits exactly one more; `evicts the oldest entry past MAX_CACHED_PAGES` fills 257 entries and proves page 0 refetches while page 1 does not. Both constants are now referenced by the test file.

---

app/api/v1/accounts/[id]/followCollectionHandler.ts
Comment (llun) [reply app/api/v1/accounts/[id]/followCollectionHandler.ts]: Reworded in 7bba009: the 400 applies when the remote is consulted; on the local path (a local actor or any fallback) a URL cursor is ignored and the first local page served, because it names nothing among the follow-row ids that path paginates on.

---

app/api/v1/accounts/[id]/followCollectionHandler.ts
Comment (llun) [reply app/api/v1/accounts/[id]/followCollectionHandler.ts]: Fixed in 7bba009: the handler exports `CORS_HEADERS` directly and both routes import that name; the alias is gone.

---

docs/mastodon-api-compatibility.md
llun: Fixed in 7bba009: the docs bullet now says the 400 applies when the remote is consulted, lists every fallback (anonymous caller, blocked domain, unreachable server or actor document, unreadable page shape, hidden collection), states that on the fallback a URL cursor is ignored and the first local page served, and describes the 80-item cap and the in-flight cap.

---

app/api/v1/accounts/[id]/following/route.test.ts
Comment (llun) [reply app/api/v1/accounts/[id]/following/route.test.ts]: Fixed in 7bba009: the mocked remote page now serves `ACTOR3_ID`, which the local following row for `EXTERNAL_ACTOR1` (`ACTOR2_ID`) can never produce, with a comment recording why.

---

app/api/v1/accounts/[id]/following/route.test.ts
Comment (llun) [reply app/api/v1/accounts/[id]/following/route.test.ts]: Fixed in 7bba009: the fallback test asserts the account uris are exactly `[ACTOR2_ID]`, the seed's local following row, which the mocked remote page (`ACTOR3_ID`) cannot produce.

---

app/api/v1/accounts/[id]/followers/route.test.ts
Comment (llun) [reply app/api/v1/accounts/[id]/followers/route.test.ts]: Fixed in 7bba009: the fallback test asserts the account uris are exactly `[ACTOR1_ID]`, the seed's local follower row, which the mocked remote page (`ACTOR2_ID`) cannot produce.

---

lib/services/mastodon/remoteFollowCollection.test.ts
llun: Fixed in 7bba009: `vi.useRealTimers()` (and `vi.restoreAllMocks()`) moved to an `afterEach`.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,396 @@
...
+    database,
+    actorIds,
+    signingActor
+  })
+
+  return {
+    status: 'ok',
+    page: {
+      accounts,
+      nextPageUrl: getPageUrl(collection.page.next),
```
llun: **should-fix:** `collection.page.next`/`prev` are emitted into the Link header UNCHECKED, but the same value is refused with `invalid-page` → 400 when the client sends it back if it fails `isCollectionPageUrl` (same protocol + host, path equal to or under the collection's). A remote whose pages live beside the collection rather than under it (`/users/x/followers_page?…`, a CDN host, `http` pages on an `https` collection) hands the client a 400 on a URL this server minted. Before this PR `getActorCollections` silently substituted the first page for such a URL; now it is a hard error. Fix: run `isCollectionPageUrl(next, collectionUrl)` when building `nextPageUrl`/`prevPageUrl` and null the ones that fail, so pagination ends cleanly, and add a test with a `next` on another path.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,396 @@
...
+      return null
+    }
+    // The same unknown actor can be recorded at the same moment from the
+    // other collection's entry (or any other path that records actors); the
+    // loser's insert fails on the id's unique constraint after the row exists.
+    // Re-read before giving the actor up for the whole TTL.
+    const existing = await database
+      .getActorFromId({ id: actorId })
+      .catch(() => null)
+    if (existing) return existing.id
```
llun: **nit:** when the re-read finds the row, the original error is dropped silently. On `recordActorIfNeeded`'s stale-refresh branch that hides a persistent `updateActor`/`setActorCounters` failure. Log it at `debug` with `err` before returning.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,396 @@
...
+  if (!person || !collectionUrl) return { status: 'unavailable' }
+
+  // The same check `remote-statuses` applies to its `page_url`: a cursor is
+  // only ever a page of THIS actor's collection, never an arbitrary URL this
+  // instance would then fetch on the client's behalf.
+  if (pageUrl && !isCollectionPageUrl(pageUrl, collectionUrl)) {
+    return { status: 'invalid-page' }
+  }
+
+  const collection = await getActorCollections({
```
llun: **nit (documented gap):** the page body is never `compactActivityPub`-ed (AGENTS: new entry points parsing remote AP docs must compact). A page spelling its items as `as:orderedItems` reads as `unavailable` (the safe direction), and a `next` given as `{ id }` yields `getPageUrl → null`, silently ending pagination. `totalItems` is unvalidated JSON typed as `number` and unused by the handler.

---

lib/services/mastodon/remoteFollowCollection.ts
```diff
@@ -0,0 +1,396 @@
...
+// is the trade the fallback exists for.
+
+export type RemoteFollowField = 'followers' | 'following'
+
+export const REMOTE_FOLLOW_PAGE_TTL_MS = 60_000
+// The route's own maximum page size; a remote page is never served past it.
+export const MAX_PAGE_ITEMS = 80
+export const MAX_UNKNOWN_ACTORS_PER_PAGE = 20
+export const UNKNOWN_ACTOR_CONCURRENCY = 5
+// Uncached page reads in flight at once, across every actor and viewer this
```
llun: **nit ×2:** (a) "across every actor and viewer this process serves" claims more than the code guarantees — `inFlightReads` lives on `CacheState`, keyed per `Database` in the `WeakMap`; it is process-wide in effect only because there is one database singleton today. Say "per database instance (one per process today)". (b) A read holds its slot through up to 20 × (Person + 3 collection roots) fetches with `request`'s retry/backoff, so one slow attacker-controlled domain can pin all four slots and degrade the feature instance-wide for minutes. Worth a per-read deadline or `withNetworkRetry: false` inside the cached work in a follow-up; at minimum note it beside the constant.

---

docs/mastodon-api-compatibility.md
```diff
@@ -395,6 +395,33 @@ are not part of the Mastodon API and are safe for Mastodon clients to ignore.
...
+  remote read: an anonymous caller, a blocked domain, an unreachable server or
+  actor document, a page in a shape the reader does not handle, and a hidden
+  collection (Mastodon's "hide your social graph", which publishes a size but
+  no page) all fall back to the locally-known rows, so the old behaviour is the
+  floor — and on that fallback a URL cursor is ignored and the first local page
+  served. Each page resolves at most 20 actors this instance has never stored
+  (five at a time, each a `recordActorIfNeeded`) and drops the rest of that
+  page's unknowns; a resolved page is cached for 60 seconds per actor, field
+  and page, with concurrent misses sharing one fetch. The cache bounds repeat
+  opens only, since a page URL is client-chosen; a process-wide cap of four
```
llun: **nit:** "process-wide" — the cap is per database instance (one per process today); say that.

---

docs/mastodon-api-compatibility.md
```diff
@@ -395,6 +395,33 @@ are not part of the Mastodon API and are safe for Mastodon clients to ignore.
 
 - **Remote statuses** — `/api/v1/accounts/:id/remote-statuses` exposes cached
   remote posts for an actor.
+- **Remote follow lists are read live from the remote collections.** Mastodon
+  answers `GET /api/v1/accounts/:id/followers` and `/following` for a remote
+  account from the relationships it already knows about, so a small instance
+  shows one local follower and nothing followed. For a remote actor these two
+  routes instead read the actor's own `followers`/`following` ActivityPub
+  collections (`lib/services/mastodon/remoteFollowCollection.ts`), signed by
+  the federation signing actor, and serialize each listed actor as an Account
```
llun: **nit:** "signed by the federation signing actor" — the handler uses `getFederationSigningActorSafe`, and a missing signer degrades to an unsigned fetch. Say "when one is available".

---

app/api/v1/accounts/[id]/following/route.test.ts
Comment (llun) [inline app/api/v1/accounts/[id]/following/route.test.ts:1]: **nit (pre-existing line, outside the diff):** the clamp test's `getFollowingSpy` (around line 120) is never restored, while the followers twin calls `mockRestore()` on its `getFollowers` spy, and this file has no `restoreAllMocks`. Harmless today (`vi.spyOn` on an already-spied method reuses the spy and `clearAllMocks` clears its history), but the two files should match.

---

llun (commented):
Sub-agent review round 2 of 2 (the last round for this PR), on 7bba009. No blockers. One should-fix (an unchecked remote `next`/`prev` page URL can be refused with a 400 when sent back) and six nits, inline. Verified clean this round: the in-flight counter on every path (rejection, `finally`, cache hit, refusal uncached), key normalization keeps distinct pages distinct and fetches the raw URL, `MAX_PAGE_ITEMS` after dedup, no legitimate page shape now reads as unavailable, local path unchanged, no new 500 paths, all fetches via `safeRemoteFetch`, domain policy outside the cache. Every new test was mutation-checked: each fails for the stated reason when its guard is loosened (concurrency at 6, eviction direction, cap ahead of cache lookup, cached refusal, leaked slot, race re-read, key sort/hash, log level, both slices). Docs constants and wording match the code apart from the two nits.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Not changing this in the PR, deliberately. The page is read through the pre-existing `getActorCollections` reader, which `getActorPosts` and the follow-timeline backfill also use without compacting the page envelope (they compact the embedded objects per item). Here the items are actor ids, so there is nothing per-item to compact, and every shape this reader cannot read (`as:orderedItems`, an object-valued `next`) degrades in the safe direction: `unavailable` → local rows, or pagination simply ending. `totalItems` is carried for the service's own result and never emitted. Compacting the collection envelope is a change to the shared reader and belongs in its own PR; noted as a known gap.

---

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 2aa0fa9: `getPageUrl` now validates `next` and `prev` URLs using `isCollectionPageUrl(url, collectionUrl)` and nulls URLs that do not match the collection path/host/protocol, preventing 400 errors when a client paginates. Added a test in `remoteFollowCollection.test.ts` verifying that out-of-collection `next`/`prev` URLs are nulled.

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 2aa0fa9: when the re-read finds the existing row after a failed insert, the error is now logged at `debug` level with `err` and `actorId` before returning `existing.id`. Added an assertion in `remoteFollowCollection.test.ts` verifying the debug log.

lib/services/mastodon/remoteFollowCollection.ts
llun: Fixed in 2aa0fa9: updated the comment above `MAX_IN_FLIGHT_REMOTE_READS` to state that the cap is scoped per database instance (one per process today), and noted that slots are held across unknown-actor fetches under retry/backoff.

docs/mastodon-api-compatibility.md
llun: Fixed in 2aa0fa9: updated `docs/mastodon-api-compatibility.md` to state that the in-flight remote read cap is per database instance (one per process today).

docs/mastodon-api-compatibility.md
llun: Fixed in 2aa0fa9: updated `docs/mastodon-api-compatibility.md` to say "signed by the federation signing actor when one is available".

app/api/v1/accounts/[id]/following/route.test.ts
Comment (llun) [reply app/api/v1/accounts/[id]/following/route.test.ts:1]: Fixed in 2aa0fa9: added `getFollowingSpy.mockRestore()` to the clamp test in `following/route.test.ts` to match `followers/route.test.ts`.

PR #1820, opened 2026-09-06
fix: support actor and profile URLs in webfinger and account lookups

llun:
## Summary

- Resolves 404 error on WebFinger (`/.well-known/webfinger`) when the `resource` query parameter is an ActivityPub actor URL (e.g. `https://llun.dev/users/null` per RFC 7033 §4.1/§4.4) or profile URL (e.g. `https://llun.dev/@null`).
- Previously, `getAccountFromResource` strictly split on `@` expecting `[acct:]username@domain`, which yielded length 1 for actor URLs and caused WebFinger to immediately return 404 without querying the database.
- Added URL handle extraction via `parseAccountUrlHandle` (supporting `/users/:username` and `/@:username`), plus fallback to `database.getActorFromId({ id: resource })` in `getWebFingerResponse` for custom actor URLs.
- Injected `fallbackDomain` from request `Host` header for bare handles (e.g. `acct:username` or `username`).
- Fixed similar handle-parsing assumptions in other lookup endpoints:
  - `/api/v1/accounts/lookup`: `parseAccountHandle` now recognizes actor and profile URLs via `parseAccountUrlHandle`.
  - `/api/v1/accounts/search`: When searching accounts with `resolve=true` and query `q` is an actor or profile URL, exact actor resolution now uses `parseAccountUrlHandle(query)`.
- Added unit and route tests covering actor URL parsing, WebFinger responses, and Mastodon API account lookups/searches.

## Checklist

- [x] If migrations changed: BOTH `migrations/schema.sql` and `migrations/schema.sqlite.sql` are regenerated in this PR (No migrations)

### DIFF
--- a/app/api/v1/accounts/lookup/route.test.ts
+++ b/app/api/v1/accounts/lookup/route.test.ts
@@ -131,6 +131,44 @@ describe('GET /api/v1/accounts/lookup', () => {
     })
   })
 
+  it('resolves lookup when acct is an ActivityPub actor URL', async () => {
+    const actor = { id: 'https://llun.test/users/test1' }
+    const account = { id: 'test1', username: 'test1', acct: 'test1' }
+    mockGetActorFromUsername.mockResolvedValue(actor)
+    mockGetMastodonActorFromId.mockResolvedValue(account)
+
+    const response = await GET(
+      new NextRequest(
+        'https://llun.test/api/v1/accounts/lookup?acct=https%3A%2F%2Fllun.test%2Fusers%2Ftest1'
+      )
+    )
+
+    expect(response.status).toBe(200)
+    expect(mockGetActorFromUsername).toHaveBeenCalledWith({
+      username: 'test1',
+      domain: 'llun.test'
+    })
+  })
+
+  it('resolves lookup when acct is a profile URL', async () => {
+    const actor = { id: 'https://llun.test/users/test1' }
+    const account = { id: 'test1', username: 'test1', acct: 'test1' }
+    mockGetActorFromUsername.mockResolvedValue(actor)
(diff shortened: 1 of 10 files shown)


llun:
### Code Review Round 1 - Resolved

Addressed findings from code review:
1. **Prevent URL Fall-Through in WebFinger**: When a resource is an absolute HTTP(S) URL but not recognized by `parseAccountUrlHandle` (e.g. custom actor paths), `getAccountFromResource` immediately returns `null` rather than falling through to handle-splitting under `fallbackDomain`.
2. **Prevent URL Fall-Through in Account Lookup**: In `parseAccountHandle`, HTTP(S) URLs that fail `parseAccountUrlHandle` return `null` directly (triggering 400 Bad Request) rather than falling through and being treated as a username under `localDomain`.
3. **Validate HTTP/HTTPS Protocol**: Ensured `parseProfileUrlAccountHandle` and `parseActorUrlAccountHandle` only accept `http:` and `https:` protocols.
4. **Enhanced Test Coverage**:
   - Added test in `index.test.ts` verifying that custom actor URLs with `fallbackDomain` fall back to `getActorFromId` without invoking `getActorFromUsername`.
   - Added test in `lookup/route.test.ts` asserting that non-account URLs return 400 Bad Request.
   - Added tests in `accountHandle.test.ts` for non-HTTP protocol rejection and profile URLs with explicit ports.

llun:
### Code Review Round 2 - Clean

Sub-agent code review round 2 completed cleanly.
- All round 1 findings verified as resolved.
- Pre-commit verification gate (`prettier` → `lint` → `typecheck` → `build` → `test`) passed cleanly with 0 errors and all 13,048 tests passing.
- PR is ready.

PR #1893, opened 2026-09-07
fix(jobs): use getters in JOBS registry to avoid circular dependency TDZ

llun:
## Summary
When importing `createNoteJob` directly (such as in `scripts/maintenance/importRemoteStatus.ts`), a circular dependency chain occurred during module loading:
`importRemoteStatus.ts` -> `createNoteJob.ts` -> `syncStatusLinkPreview.ts` -> `lib/services/queue` -> `lib/services/queue/base.ts` -> `lib/jobs/index.ts` -> `createNoteJob.ts`.

Because `createNoteJob.ts` was mid-evaluation when `lib/jobs/index.ts` was evaluated, constructing the `JOBS` dictionary with eager property values (`[CREATE_NOTE_JOB_NAME]: createNoteJob`) threw `ReferenceError: Cannot access 'createNoteJob' before initialization` due to the temporal dead zone (TDZ).

This PR converts the properties in `JOBS` to property getters (`get [JOB_NAME]() { return job }`). This defers accessing the job handlers until runtime lookup rather than module evaluation time, safely eliminating circular initialization TDZ errors while preserving own property enumerability and `hasOwnProperty` semantics.

## Testing
- `yarn run prettier --write .` passed.
- `yarn lint` passed with 0 errors.
- `yarn typecheck` passed.
- `yarn build` passed.
- `yarn test` passed (990 test files, 13,576 tests).
- Verified `node scripts/run.cjs scripts/maintenance/importRemoteStatus.ts` loads successfully without TDZ ReferenceError.

### DIFF
--- a/lib/jobs/index.test.ts
+++ b/lib/jobs/index.test.ts
@@ -1,3 +1,5 @@
+import { execFileSync } from 'node:child_process'
+
 import { JOBS } from '@/lib/jobs'
 import * as jobNames from '@/lib/jobs/names'
 
@@ -12,9 +14,17 @@ describe('JOBS registry', () => {
     }
   })
 
-  it('avoids TDZ when a job module that depends on queue is imported directly', async () => {
-    const { createNoteJob } = await import('./createNoteJob')
-    expect(createNoteJob).toBeDefined()
-    expect(typeof createNoteJob).toBe('function')
+  it('avoids TDZ when createNoteJob is the initial entry point in a fresh process', () => {
+    expect(() => {
+      execFileSync(
+        process.execPath,
+        ['-r', 'tsx/cjs', '-e', 'require("./lib/jobs/createNoteJob.ts")'],
+        {
+          cwd: process.cwd(),
+          encoding: 'utf-8',
+          stdio: 'pipe'
+        }
+      )
+    }).not.toThrow()
   })
 })
--- a/lib/jobs/index.test.ts
+++ b/lib/jobs/index.test.ts
@@ -1,0 +1,20 @@
+import { JOBS } from '@/lib/jobs'
+import * as jobNames from '@/lib/jobs/names'
+
+describe('JOBS registry', () => {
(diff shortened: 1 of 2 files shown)


lib/jobs/index.test.ts
```diff
@@ -0,0 +1,20 @@
...
+    expect(names.length).toBeGreaterThan(0)
+
+    for (const name of names) {
+      expect(Object.prototype.hasOwnProperty.call(JOBS, name)).toBe(true)
+      expect(typeof JOBS[name]).toBe('function')
+    }
+  })
+
+  it('avoids TDZ when a job module that depends on queue is imported directly', async () => {
+    const { createNoteJob } = await import('./createNoteJob')
```
llun: [P2] Exercise the direct import in a fresh module graph

The top-level `import { JOBS } from '@/lib/jobs'` has already loaded and initialized `createNoteJob` before this dynamic import runs, so this test only reads a cached module and cannot catch the TDZ regression. I verified under Node 24.15.0 that both new tests still pass after restoring the entire pre-PR eager registry, while a fresh `node -r tsx/cjs -e 'require("./lib/jobs/createNoteJob.ts")'` fails with `Cannot access 'createNoteJob' before initialization`; that same command succeeds with this PR. Please add a fresh-process regression test using the scripts' tsx/CJS loader (or another verified isolated import path), with `createNoteJob` as the first entry point, and confirm it fails against the original registry. The current tests would allow this exact production failure to return unnoticed.

llun: Updated in 749a12f71: Replaced the dynamic import test with a fresh-process regression test using Node and the `tsx/cjs` loader (`node -r tsx/cjs -e "require(\"./lib/jobs/createNoteJob.ts\")"`). Verified that it fails against the original eager registry with `ReferenceError: Cannot access 'createNoteJob' before initialization` and passes cleanly with the getter registry.

PR #2005, opened 2026-09-08
fix: remove nested boost rows by original status

llun:
Deleting the root Note of a nested boost left the outer card in Favorites/Bookmarks/home after the shared status-array migration. Resolve the root original during removal while preserving row IDs and the previous immediate-wrapper match; interaction patching and pagination cursors remain unchanged.

Validation: Node 24; `yarn run prettier --write .` → `yarn lint` → `yarn typecheck` → `yarn build` → `yarn test --maxWorkers=4` all pass. Four focused regressions cover nested root removal, intermediate wrapper removal, and Favorites/Bookmarks interaction callbacks; focused suites pass 29 tests. The initial three regressions failed before the fix.

Browser verification: local SQLite test account on the production build; deleting the root Note from the nested Favorites card immediately showed the empty state, and removing a bookmark immediately cleared Bookmarks. Bookmark creation normalizes to the original server-side, so the nested bookmark callback is additionally covered by the component regression. Screenshots waived by the user.

Independent Luna max review will run against this PR head before merge. No package version change.

### DIFF
--- a/app/(timeline)/bookmarks/BookmarksTimeline.test.tsx
+++ b/app/(timeline)/bookmarks/BookmarksTimeline.test.tsx
@@ -101,6 +101,12 @@ const boostedBookmarkedStatus: StatusAnnounce = {
   originalStatus: bookmarkedStatus
 }
 
+const nestedBoostedBookmarkedStatus: StatusAnnounce = {
+  ...boostedBookmarkedStatus,
+  id: 'https://activities.local/users/booster/statuses/boost-2',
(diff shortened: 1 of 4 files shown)


llun (commented):
Independent Luna max review, round 1, examined the complete four-file diff at `aa35951d3e32c5e81778c0259cc98f103d852d82`. No actionable findings. Focused review verification: 29 tests passed. Codex also confirmed ordered full local gates, local browser removal flows, and all required CI on this exact head.

PR #2010, opened 2026-09-08
test: require the headless signer before exclusion checks

llun:
The headless-signer exclusion test could pass when signer creation returned null because not.toContain(signer?.id) reduced to checking for undefined. Require the signer fixture before asserting its absence from the admin account IDs.

The original account-list assertions remain unchanged. The 38-test moderation suite passes; a temporary missing-signer mutation failed at the new guard and was restored. Ordered Node24 prettier → lint → typecheck → build → full tests pass. No production changes, exclusions, or package version edits. Fresh independent Luna max review will complete before merge.

### DIFF
--- a/lib/database/sql/moderation.test.ts
+++ b/lib/database/sql/moderation.test.ts
@@ -346,6 +346,7 @@ describe('ModerationDatabase', () => {
         // Materialize the headless federation signing actor (accountId null on
         // the configured host) — it must never appear in the admin listing.
         const signer = await database.getFederationSigningActor()
+        if (!signer) throw new Error('Expected a federation signing actor')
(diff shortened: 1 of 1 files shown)


llun (commented):
Independent Luna Max review, round 1: clean. The explicit required-signer guard prevents the exclusion assertion from passing vacuously when the fixture is missing. The complete diff is test-only. All 38 focused tests and ordered full local gates passed; a temporary missing-signer mutation fails at the guard as intended.

Reviewed head: `2633666632f07863c4eed994e5627ee838a2730e`. Codex verified the full ordered local gates and required CI on this exact head before merge.

PR #2014, opened 2026-09-08
fix: upgrade Better Auth to 1.7.3

llun:
## Summary

Upgrade Better Auth, core, OAuth provider, and passkey to 1.7.3. Account writes now use the provider/account key without populating the historical issuer column, so new and existing credentials remain usable under the restored 1.6 identity semantics. Authenticator setup explicitly requests TOTP and handles the new discriminated response.

Align better-call and @better-auth/utils with the published peer requirements and deduplicate Nano Stores to 1.5.3; its duplicate versions otherwise create incompatible copies of the core plugin types. No release-age override was needed.

## Database and rollout

Follows the [1.7 upgrade guide](https://better-auth.com/docs/guides/1-7-upgrade-guide#account-identity-keeps-the-provider-key). Retains the already-nullable issuer column and historical migrations; no schema migration or dump change is necessary. Check deployed constraints and duplicate mapped provider keys before cutover. Drain every 1.7.2 authentication-serving instance before 1.7.3 accepts authentication traffic or writes; mixed-version auth traffic can reject newly created credentials. Rollback to 1.7.2 requires restoring issuer values for rows created during 1.7.3 before returning authentication traffic to the older release. Architecture guidance documents these requirements.

## Validation

- All five local gates passed in order: formatter, lint, typecheck, production build, and 14,368 tests across 1,033 files.
- Fresh replay of all 167 migrations on disposable SQLite and PostgreSQL 17 databases; nullable issuer and absence of issuer indexes confirmed.
- PostgreSQL: 39 account tests and 10 real-handler credential identity tests passed.
- Browser: registration and sign-in with NULL issuer; OAuth consent, code exchange, and authenticated verify_credentials; TOTP setup renders QR code, manual key, and verification form.
- All CI checks passed on final commit 78d006fea, including Docker smoke tests and both schema-sync checks.
- Immutable Yarn install passed. Existing SQLite 13/Vitest 5 peer-range warnings remain unchanged from 1.7.2.

Implementation used Luna at max effort. Two independent Sol/high reviewers completed two review rounds. The final round was clean; adapter fallback and coordinated-cutover guidance were corrected, and all inline findings were answered and resolved. The historical migration comment was retained with an explicit no-change rationale.

## Checklist

- [x] Five local gates pass in the required order
- [x] Stale documentation updated and references checked
- [x] No migrations changed; both schema dumps remain unchanged
- [x] Application package version untouched


docs/architecture.md
```diff
@@ -1177,7 +1177,9 @@ it; there is no legacy shape left to copy.
...
 - **`knexAdapter` must implement `consumeOne` and `incrementOne` natively — the factory throws rather than synthesising them.** Both are race-safe primitives better-auth relies on for correctness, not speed: `consumeOne` backs single-use credentials (verification tokens, OAuth authorization codes, device codes) and `incrementOne` backs guarded counters (2FA failed attempts, rate limits, an authorization code's remaining uses). What makes each safe is that the mutating statement **re-applies the caller's predicate** alongside the row's primary key, so a racing caller re-evaluates it after the winner commits and comes away with nothing; a plain read-then-write would let both callers through. An empty predicate returns `null` rather than consuming or mutating an arbitrary row — the factory does not guard that itself.
```
llun: Review round 1 (Sol/high): Better Auth 1.7.3 restores guarded fallbacks for `consumeOne` and `incrementOne`, so the statement that its factory throws rather than synthesising them is stale. Keep our native implementation requirement, but explain the current upstream contract. The matching comment in `lib/services/auth/knexAdapter.test.ts` should be updated too.

llun:
Review round 1 (Sol/high), P3: The header comment in `migrations/20260821120000_better_auth_17_columns.js:3–9` describes issuer as load-bearing for sign-in. That describes 1.7.0–1.7.2, whereas 1.7.3 uses the provider key. Consider qualifying the comment as historical to avoid confusing future maintainers.

llun:
Disposition for the historical migration comment finding: no change. The approved upgrade retains the historical migrations unchanged; their issuer backfill records the requirements of 1.7.0–1.7.2. The current architecture guide explicitly states that 1.7.3 restores provider-key identity and retains those historical backfills only for compatibility. Editing a historical migration for commentary would also require regenerating both dumps under repository rules, without changing runtime behavior or the schema. This historical-context comment is therefore resolved as a documented scope decision.

docs/architecture.md
```diff
@@ -1177,7 +1177,9 @@ it; there is no legacy shape left to copy.
...
+- **Preflight the real database before a 1.7.3 rollout.** The committed schema already has a nullable `issuer` and no issuer-based unique index, but a deployment that was manually migrated through 1.7.0–1.7.2 may still carry the old constraint or index. Inspect the actual constraints and indexes, and check for duplicate mapped provider keys `(provider, providerId)` before new pods serve traffic; resolve ambiguous rows before deployment because better-auth rejects lookups with more than one match. If the preflight finds a constraint or duplicate, stop the rollout and keep the current release serving until the data and schema change is complete and reversible. Keep operational SQL in the deployment runbook, not this guide.
```
llun: Review round 1 (Sol/high), P2: Document a coordinated authentication cutover. A 1.7.3 instance creates credential rows with NULL issuer, which a still-serving 1.7.2 instance cannot authenticate. A rolling deployment with overlapping old/new authentication traffic therefore risks intermittent sign-in failures. Drain all 1.7.2 authentication instances before 1.7.3 accepts authentication traffic or writes. The current constraint/duplicate preflight and rollback guidance do not cover this overlap.

docs/architecture.md
llun: Fixed in 78d006fea: current guidance and the adapter test comment now describe 1.7.3 guarded fallbacks accurately while preserving our native implementation requirement.

docs/architecture.md
llun: Fixed in 78d006fea: rollout guidance now requires draining all 1.7.2 auth-serving instances before 1.7.3 accepts authentication traffic or writes. The PR description includes the same coordinated cutover requirement.

llun:
Review round 2 complete at 78d006fea: both independent Sol/high reviewers re-read the full PR and found no new actionable issues. Round 1 findings are addressed or explicitly dispositioned; both inline discussions are replied to and resolved. All five local gates passed again, including 14,368 tests across 1,033 files.

PR #2028, opened 2026-09-09
none: require human approval for major releases

llun:
## Summary

- Delete the accidental `v2.0.0` and `v2.0.1` tags, then restore the pending release line to `v1.162.0`.
- Require a human maintainer to apply `release:major` before a `major:` marker can produce a major release; unapproved markers produce a minor release.
- Verify the approval again before any major tag is created, including the direct-tag fallback, and document that agents must never apply the approval label.

## Validation

- `yarn lint`
- `yarn test .github/workflows/version-bump.test.ts`
- `yarn typecheck` is blocked by pre-existing generated Next route-type errors outside this diff.
- `yarn build` is blocked in this workspace because Turbopack cannot bind its required local helper port; this occurs with and without sandbox escalation.
- Full CI is required before merge.

### DIFF
--- a/.github/scripts/determine-version-bump.mjs
+++ b/.github/scripts/determine-version-bump.mjs
@@ -2,6 +2,7 @@ import { execFileSync } from 'node:child_process'
 import { appendFileSync } from 'node:fs'
 import process from 'node:process'
 
+import { hasHumanMajorApproval } from './major-release-approval.mjs'
 import { nextVersion, selectVersionBump } from './version-bump-policy.mjs'
 
 const git = (...args) => execFileSync('git', args, { encoding: 'utf8' }).trim()
@@ -9,27 +10,6 @@ const output = (name, value) =>
   appendFileSync(process.env.GITHUB_OUTPUT, `${name}=${value}\n`)
 const majorMarker = (subject, body) =>
   /^major:/.test(subject) || /(^|\n)\s*[-*]?\s*major:/.test(body)
-
-const hasHumanMajorApproval = (hash) => {
-  try {
-    const pullRequests = JSON.parse(
-      execFileSync(
-        'gh',
-        ['api', `repos/${process.env.GITHUB_REPOSITORY}/commits/${hash}/pulls`],
-        { encoding: 'utf8' }
-      )
-    )
-    return pullRequests.some(
-      (pullRequest) =>
-        pullRequest.merged_at &&
-        pullRequest.labels.some((label) => label.name === 'release:major')
-    )
-  } catch {
-    // Fail closed: without the maintainer approval record, a major marker is
-    // intentionally treated as a minor release request.
-    return false
-  }
-}
 
 const baseSha = git('rev-parse', 'origin/main')
 output('base_sha', baseSha)
--- a/.github/scripts/determine-version-bump.mjs
+++ b/.github/scripts/determine-version-bump.mjs
@@ -1,0 +1,104 @@
+import { execFileSync } from 'node:child_process'
+import { appendFileSync } from 'node:fs'
+import process from 'node:process'
+
+import { nextVersion, selectVersionBump } from './version-bump-policy.mjs'
+
+const git = (...args) => execFileSync('git', args, { encoding: 'utf8' }).trim()
+const output = (name, value) =>
+  appendFileSync(process.env.GITHUB_OUTPUT, `${name}=${value}\n`)
+const majorMarker = (subject, body) =>
+  /^major:/.test(subject) || /(^|\n)\s*[-*]?\s*major:/.test(body)
+
+const hasHumanMajorApproval = (hash) => {
+  try {
+    const pullRequests = JSON.parse(
+      execFileSync(
+        'gh',
+        ['api', `repos/${process.env.GITHUB_REPOSITORY}/commits/${hash}/pulls`],
+        { encoding: 'utf8' }
+      )
+    )
+    return pullRequests.some(
+      (pullRequest) =>
+        pullRequest.merged_at &&
+        pullRequest.labels.some((label) => label.name === 'release:major')
+    )
+  } catch {
+    // Fail closed: without the maintainer approval record, a major marker is
+    // intentionally treated as a minor release request.
+    return false
+  }
+}
+
+const baseSha = git('rev-parse', 'origin/main')
+output('base_sha', baseSha)
+
+if (process.env.GITHUB_SHA !== baseSha) {
+  output('new_version', '')
+  process.exit(0)
+}
+
+const headMessage = git('log', '-1', '--pretty=format:%s', baseSha)
+if (/^Bump version to v\d+\.\d+\.\d+/.test(headMessage)) {
+  output('new_version', '')
+  process.exit(0)
+}
+
+git('checkout', '--detach', baseSha)
+const latestTag = git(
+  'tag',
+  '--merged',
+  'HEAD',
+  '--list',
+  'v[0-9]*',
+  '--sort=-v:refname'
+).split('\n')[0]
+
+if (!latestTag) {
+  output('bump', 'initial')
+  output('commit_count', '0')
+  output('latest_tag', '')
+  output('new_version', '1.0.0')
+  process.exit(0)
+}
+
+output('latest_tag', latestTag)
+const hashes = git(
+  'log',
+  `${latestTag}..HEAD`,
+  '--pretty=format:%H',
+  '--no-merges'
+)
+  .split('\n')
+  .filter(Boolean)
+
+if (hashes.length === 0) {
+  output('new_version', '')
+  process.exit(0)
+}
+
+const commits = hashes.map((hash) => {
+  const subject = git('log', '-1', '--pretty=format:%s', hash)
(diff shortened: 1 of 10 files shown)


.github/workflows/version-bump.test.ts
```diff
@@ -0,0 +1,14 @@
+import { readFileSync } from 'node:fs'
+
+const WORKFLOW_PATH = '.github/workflows/version-bump.yml'
+
+describe('automatic version-bump policy', () => {
+  const workflow = readFileSync(WORKFLOW_PATH, 'utf8')
+
+  it('caps major-prefixed commits at a minor release', () => {
```
llun: P1: The current workflow test only checks source text, so a future change could restore major arithmetic while preserving those strings. Extract or directly exercise the version-selection behavior with fixtures for title/body `major:` inputs and the v1.161.52 → v1.162.0 recovery path.

llun:
P1 review finding: `.github/workflows/tag-version.yml` still tags any merged `Bump version to vX.Y.Z` commit. An accidentally titled PR could therefore recreate v2.0.0 even if `version-bump.yml` caps prefixes. Add the same fail-closed human approval gate for major tag transitions, with executable regression coverage.

llun:
P1 review finding: the implementation does not actually prove that `release:major` was applied by a human maintainer. Both approval checks only test current label presence, while this workflow itself applies that label to the generated bump PR with `gh pr edit --add-label`. Any bot, agent, or other write-capable actor that applies the label satisfies the same check, and the tag workflow then creates the major tag. Please enforce provenance at the release boundary: for example, gate the major-tag job with a protected GitHub environment requiring maintainer review, or validate the label event actor and maintainer permission rather than treating label presence as human approval. Add an integration-level regression for a bot-applied label being rejected.

llun:
P1 review finding: the stale-run fallback at `.github/workflows/version-bump.yml:460-508` relies on branch protection being `strict`, but the live `required_status_checks` API currently reports `strict: false`. If main moves while a bump PR is checking, `arm_auto_merge` can therefore merge the stale version without recomputing against the new commits; the queued workflow then sees a version-bump head and skips. A later approved major commit can consequently be included beneath an already-calculated minor tag, and a stale approved major PR can merge after its source state changes. Either enable strict required-status checks on main as part of this change and verify it, or stop arming stale bump PRs and let the newer queued run force-push the recomputed version. Add coverage for main moving after PR creation.

.github/scripts/determine-version-bump.mjs
```diff
@@ -0,0 +1,104 @@
...
+        ['api', `repos/${process.env.GITHUB_REPOSITORY}/commits/${hash}/pulls`],
+        { encoding: 'utf8' }
+      )
+    )
+    return pullRequests.some(
+      (pullRequest) =>
+        pullRequest.merged_at &&
+        pullRequest.labels.some((label) => label.name === 'release:major')
+    )
+  } catch {
```
llun: P1: Fail the run when the approval lookup errors instead of silently converting it to an unapproved request. This catch handles API outages, auth failures, and malformed responses by returning false, so a genuinely human-approved `major:` commit can be released as a minor version. After that minor tag becomes the new baseline, the approved commit is no longer reconsidered and the intended major release is lost. Keep a successful lookup with no `release:major` label as the minor fallback, but let lookup/parsing failures stop the workflow for retry.

.github/workflows/version-bump.test.ts
llun: Addressed in efb415f2: version selection is executable policy code with title/body fixtures, including the v1.161.52 to v1.162.0 recovery case.

.github/scripts/determine-version-bump.mjs
llun: Addressed in efb415f2: approval API and JSON errors now fail the workflow for retry; only a successful lookup without qualifying approval downgrades to minor.

.github/scripts/verify-major-tag-approval.mjs
```diff
@@ -0,0 +1,54 @@
...
+
+const git = (...args) => execFileSync('git', args, { encoding: 'utf8' }).trim()
+const nextTag = process.argv[2]
+const majorMarker = (subject, body) =>
+  /^major:/.test(subject) || /(^|\n)\s*[-*]?\s*major:/.test(body)
+const getSourceMainSha = () => {
+  const pullRequest = getPullRequestsForCommit(process.env.GITHUB_SHA).find(
+    (candidate) => candidate.merged_at
+  )
+  const match = pullRequest?.body?.match(/^- Base main SHA: `([0-9a-f]{40})`$/m)
```
llun: P1: Do not derive the authorization range from the PR body. `Base main SHA` is editable PR text, so a manually created or edited `Bump version to vX.Y.Z` PR can point this guard at an older range containing some human-approved `major:` commit. Once the repository has one approved major release, that old approval can then be replayed to authorize a later major tag even though the current source range has no approved feature PR. For the tag workflow, the trusted source-main SHA is the bump squash commit parent (`HEAD^`); for the direct-tag fallback, pass the already computed `BASE_SHA` explicitly. Bind the range to those git/workflow values and add a regression showing that a forged body SHA cannot reuse an earlier approval.

.github/scripts/major-release-approval.mjs
```diff
@@ -0,0 +1,38 @@
...
+      !pullRequest.labels.some((label) => label.name === 'release:major')
+    ) {
+      return false
+    }
+
+    const events = githubApi(
+      `repos/${process.env.GITHUB_REPOSITORY}/issues/${pullRequest.number}/events?per_page=100`
+    )
+    return hasHumanMajorLabelEvent(pullRequest, events)
+  })
```
llun, line 56: P2: Paginate the issue-event lookup before deciding that no human approval exists. `per_page=100` returns only the first page, so on a busy PR where `release:major` was applied after the first 100 events, the current label check succeeds but the human label event is missed. The approved breaking change is then silently downgraded to minor and will fall behind that new tag. Use `gh api --paginate --slurp` (flattening the pages) or another complete provenance query, and cover an approval event beyond page one.

llun:
P1 review finding: `hasHumanMajorLabelEvent` does not prove that the currently present `release:major` label was applied by a User; it accepts any historical User `labeled` event. A sequence of User labels → label is removed → Bot/agent re-adds leaves the label currently present and the old User event in the event list, so this helper returns true and permits the major tag. Replay the `labeled`/`unlabeled` events for `release:major` in order and require the last effective application to be a User event (and paginate the events endpoint so the decisive event cannot fall beyond the first 100). Add that exact human-label/remove/bot-relabel sequence as a false regression case.

llun:
P2 review finding: the unapproved-major downgrade no longer fails closed when approval lookup is unavailable. `hasHumanMajorApproval` now lets either `gh api` call throw, and `determine-version-bump.mjs` does not catch it, so a transient API error, pagination/permission issue, or malformed response aborts the whole bump instead of turning `major:` into the documented minor release. The earlier implementation explicitly returned false on lookup failure. Catch approval lookup failures in the determine path and treat them as `majorApproved: false`; keep the tag guard throwing/failing closed. Add a test for lookup failure producing the minor calculation.

llun:
Correction: source approval requires a current release:major label and a GitHub label event whose actor is a user account. API failures stop the workflow. The workflow neither copies approval labels nor arms auto-merge; it merges only after checks and base-SHA revalidation.

.github/scripts/major-release-approval.mjs
```diff
@@ -0,0 +1,56 @@
...
+
+  const latestLabelEvent = events
+    .filter(
+      (event) =>
+        ['labeled', 'unlabeled'].includes(event.event) &&
+        event.label?.name === 'release:major'
+    )
+    .sort(
+      (left, right) =>
+        Date.parse(right.created_at) - Date.parse(left.created_at)
```
llun: P1: Break same-second label-event ties using event order/id, not only `created_at`. GitHub issue-event timestamps have one-second precision, and JavaScript sort is stable, so a User label followed by Bot unlabel/relabel events within the same second leaves the earlier User event first here. With the label currently present, `hasHumanMajorLabelEvent` then returns true even though the latest transition was Bot-applied. I reproduced this by passing three events with identical timestamps and increasing ids; the helper returns true. Preserve the API sequence and take the last transition, or sort by `created_at` plus the monotonic event id, and add this same-second regression.

.github/scripts/verify-major-tag-approval.mjs
llun: Addressed in 833eb14b: the tag workflow derives the source range from the bump commit parent, and the direct-tag fallback receives BASE_SHA explicitly.

.github/scripts/major-release-approval.mjs
llun: Addressed in 833eb14b: issue-event lookup now paginates with gh api --paginate --slurp and flattens all pages before selecting the latest transition.

.github/scripts/major-release-approval.mjs
llun: Addressed in bfda66fd: equal timestamps now break by descending GitHub event id, with a same-second User then Bot regression.

PR #2089, opened 2026-09-19
minor: generate alt text for uploaded videos

llun:
## Summary

Extends the existing automated image alt-text pipeline to uploaded videos, in both the synchronous and direct-to-S3 presigned upload paths, and adds screen-reader labelling for the rendered video.

- **Representative frame extraction** — `extractVideoImage` now runs ffmpeg with `-vf thumbnail -frames:v 1 -update 1`, so the stored poster and the frame sent for description is a visually representative color frame rather than frame 0 (often black or blank).
- **Sync pipeline** — `handleSyncMediaUpload` generates alt text for `type === 'video'`: it extracts the preview frame from the uploaded buffer and calls `generateAltText(..., 'image/jpeg')`. Buffer-read, extraction and DB-write failures are caught and logged; the upload still succeeds.
- **Presigned pipeline** — `S3StorageFile.completePresignedUpload` analyzes video objects within `PRESIGNED_ANALYSIS_MAX_BYTES`, derives blurhash/focus from the preview frame, generates alt text when `media.description == null`, and still completes verification on extraction/vision failure.
- **Prompt** — the system prompt covers images and video preview frames and forbids conversational meta-preambles such as "This video shows", "This image depicts" or "Screenshot of".
- **UI** — the ordinary `<video>` gets `aria-label={caption ?? name ?? undefined}`, matching `` alt-text behaviour.

## Files

- `lib/services/medias/extractVideoImage.ts`
- `lib/services/medias/handleSyncMediaUpload.ts`
- `lib/services/medias/S3StorageFile.ts`
- `lib/services/altText/openai.ts`
- `lib/components/posts/media.tsx`
- Tests: new `lib/services/medias/extractVideoImage.test.ts`, plus `app/api/v2/media/route.test.ts`, `lib/services/medias/S3StorageFile.test.ts`, `lib/services/altText/openai.test.ts`, `lib/components/posts/media.test.tsx`
- Docs: `docs/environment-variables.md`, `docs/features.md`, `docs/maintenance.md`, `.env.example`

## Real-model verification

Read `/Users/llun/Desktop/P9190255.mov` (102.9 MB), extracted the representative preview frame via `extractVideoPreviewFrame`, and called `generateAltText` against the configured `gemini-3.5-flash-lite` endpoint. Produced frame was a clear aircraft shot (not blank) and the generated description was:

> A gray military transport aircraft, identified as a Luftwaffe Airbus A400M, banks sharply to the left against an overcast sky.

## Test plan

- [x] `yarn run prettier --write .`
- [x] `yarn lint` (0 errors)
- [x] `yarn typecheck`
- [x] `yarn build`
- [x] `yarn test` — 1065 files / 14821 tests passed

## Notes

- No migration; no new environment variables.
- No operational SQL.


llun:
## Sub-agent review round 1

Three review agents ran in parallel over the full diff (correctness/storage, tests/typing, docs/UI/scope). Findings and disposition below.

### Addressed

- **Prompt regression** (`lib/services/altText/openai.ts`) — the rewrite dropped the old `"A photo of"` ban, changing existing image behavior. Restored it, and the prompt test now asserts the instruction as a negation (`/Never start with phrases such as "This video shows"/`) instead of merely checking the forbidden strings appear, so a contrary prompt can no longer pass.
- **Weak sync-video failure test** (`app/api/v2/media/route.test.ts`) — the extraction-failure case didn't prove the video branch ran. Added `toHaveBeenCalledTimes(1)` and renamed it to make clear it covers alt-text extraction specifically (poster extraction happens in `saveMedia` and remains fatal).
- **Weak success assertions** (`app/api/v2/media/route.test.ts`) — now pins the exact uploaded bytes (`Buffer.from([1, 2, 3])`) and the call counts of `extractVideoPreviewFrame` / `generateAltText`.
- **No image-side negative** (`lib/services/medias/S3StorageFile.test.ts`) — the image analysis test now asserts `extractVideoImage` is not called.
- **Stale/overstated docs** (`docs/environment-variables.md`, `docs/features.md`, `docs/maintenance.md`) — clarified that alt-text extraction is best-effort, but a synchronous video upload still requires a decodable poster frame, and scoped the "extract before storage, never after" invariant to the `saveFile` path (the presigned completion path is the documented after-store exception). Softened the "black frame still gets a description" claim to "more representative than frame 0".
- **Misleading helper name** (`extractVideoImage.test.ts`) — `lastOutputPath` → `onlyOutputPath`.
- **PR prefix** — retitled to `minor:` per the version-bump policy (new backwards-compatible user-facing feature).

### Not changed (with rationale)

- **Double frame extraction on sync uploads** — `saveMedia` stores a poster and then the handler extracts again for alt text. This is the explicitly requested shape (extract from the uploaded buffer in `handleSyncMediaUpload`), and sharing the frame would require widening `MediaStorageSaveFileOutput` / `saveMediaFile`, which is out of scope. The extra ffmpeg pass is bounded by the vision request that follows it.
- **`thumbnail` default batch under the 30s timeout** — the filter analyses its first batch (default 100 frames) rather than the whole clip; this is exactly the requested representative-frame behavior, and the 30s budget covers decoding 100 frames in normal cases. No upload path previously guaranteed success on a clip that needs more than that.
- **Presigned videos above 50 MiB skip analysis** — intentional and specified (`PRESIGNED_ANALYSIS_MAX_BYTES`), now stated in the docs.
- **`aria-label` terminal `?? undefined` branch untested** — `Attachment.name` is a required `z.string()`, so the branch is unreachable through the typed API; omitting the attribute matches how `` degrades.
- **`as unknown as` at the `execFile` mock boundary** — required by `execFile`'s overloads; no `any` is used.
- **`S3StorageFile.test.ts` `clearAllMocks` vs `mockReset`** — each `beforeEach` explicitly re-assigns the base implementations (`updateMedia`, `getMediaByIdForAccount`, `markMediaUploadVerified`, `extractVideoImage`), so no implementation leaks between tests.

### Verified clean

Path/SSRF safety (server-generated temp names, `wx` open, `finally` cleanup), error semantics (non-fatal analysis vs fatal save-time poster extraction), driver parity on the sync path, and the ffmpeg argument ordering (empirically produced a non-black frame with `-vf thumbnail -frames:v 1 -update 1`).

llun:
## Sub-agent review rounds 2–4

- **Round 2** — verified the round-1 fixes were non-vacuous (the prompt regex fails if the ban is removed; the call-count assertions cannot legitimately double-count). Two low-severity documentation inaccuracies in the new `docs/maintenance.md` subsection: `-update 1` was described as "required", and the `thumbnail` filter as globally "most representative".
- **Round 3** — verified those corrections, but caught that the source comment in `extractVideoImage.ts` still repeated the "would refuse" claim, and that the "silence the warning" rationale was moot because `-loglevel error` already suppresses the image2 warning.
- **Round 4** — **clean, no actionable findings.** All gates re-run green on the final state.

Final state: `-update 1` is documented as defensive/explicit rather than required, the `thumbnail` batch window (first 100 frames) is stated, and the code comment matches.

